PR #9014, opened 2020-07-13
KAFKA-10156 [WIP] Fix flaky testCloseOldestConnection

serjchebotarev:
In `testCloseOldestConnection` flaky behavior can be caused by the code path when in `selector.poll(0)` the connection is not closed immediately. This can be due to `Selector#maybeReadFromClosingChannel` returning `true`. To avoid this, also check `selector.closingChannel(id)` in addition to `selector.disconnected().containsKey(id)`.

### DIFF
--- a/clients/src/test/java/org/apache/kafka/common/network/SelectorTest.java
+++ b/clients/src/test/java/org/apache/kafka/common/network/SelectorTest.java
@@ -463,8 +463,13 @@ public class SelectorTest {
         time.sleep(6000); // The max idle time is 5000ms
         selector.poll(0);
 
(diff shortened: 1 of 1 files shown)


serjchebotarev:
@chia7712, @hachikuji & @ijuma could you please have a look at this one - if the chosen approach is okay or should this be done some other way?

chia7712:
> This can be due to Selector#maybeReadFromClosingChannel returning true

Is it an issue in this case?

serjchebotarev:
The problem is that locally I was not able to reproduce this behavior, AFAIK it was only observed in some CI builds. Maybe the way to check this is to implement fix and observe the test behavior in several CI builds... At the moment I do not have any better idea on how to check this.

PR #9017, opened 2020-07-14
KAFKA-8582 [WIP] Add ability to handle late messages in streams-aggregation

piddubnyi:
KAFKA-8582 Indicates that there is a need to handle *late* messages, which arrive when respective aggregation window (including grace period) is already closed.
There is also related [SO question](https://stackoverflow.com/questions/62370180/kafka-stream-time-and-window-expire-kstreamsessionwindowaggregate-skipping-rec), where @mjsax mentions following options for handling late messages:
- start new app to reset stream time
- fall back to the Processor API and basically re-implement aggregation logic manually.

This PR tries to add an ability to forward such *late* messages to a configurable deadLetterTopic.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedKStream.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedKStream.java
@@ -421,6 +421,70 @@ public interface TimeWindowedKStream<K, V> {
                                            final Materialized<K, VR, WindowStore<Bytes, byte[]>> materialized);
 
     /**
+     * Aggregate the values of records in this stream by the grouped key and defined windows.
+     * Records with {@code null} key or value are ignored.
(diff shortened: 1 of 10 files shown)


mjsax:
@piddubnyi Thanks for the PR -- it seems this PR includes public API changes and thus requires a KIP: https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Improvement+Proposals

Do you have a Jira account so we can assign the ticket to you -- btw: it's always best to _first_ assign a ticket before starting working on it; otherwise, it could happen that two people work on the same thing in parallel what is of course not desirable.

Let us know if you have any questions about the KIP process.

piddubnyi:
Hi @mjsax, I commented the ticket to get it assigned.
Regarding the KIP, I just created a fresh account in Confluence, however, it looks like there are no rights to trigger the KIP. Please suggest the best way to proceed.

mjsax:
I found your wiki account and granted write access. You should be all set.

PR #9020, opened 2020-07-14
KAFKA-10271 Performance regression while fetching a key from a single partition

dima5rr:
StreamThreadStateStoreProvider excessive loop over calling internalTopologyBuilder.topicGroups(), which is synchronized, thus causing significant performance degradation to the caller, especially when store has many partitions.

https://issues.apache.org/jira/browse/KAFKA-10271


abbccdda (commented):
Thanks for the PR, could we add some unit test coverage?

vvcephei:
Hi @dima5rr , thanks for the PR! I was looking into the context of this ticket and noticed that we're essentially just duplicating here the logic in `org.apache.kafka.streams.state.internals.QueryableStoreProvider#getStore`, which also loops over all stores of all providers as a sanity check before even returning the WrappingStoreProvider, which will loop over all stores of all providers _again_ before returning any results.

Anecdotally, I think that most people would actually just query immediately and then discard their store reference, for example like `value = kafkaStreams.store("storeName", partition=2).get("key")`. In that case, this double-iteration is pretty expensive, and the up-front sanity check doesn't seem to provide any value.

In fact, the only value it could provide is if you _do_ plan to save the store reference and re-use it for multiple queries. But in that case, there could be a rebalance at any time, so checking up front probably doesn't help much no matter what the use case is.

What do you think about removing the loop https://github.com/apache/kafka/blob/f699bd98c14e31f07c5a3f6ba9ce2c4b441e7fdb/streams/src/main/java/org/apache/kafka/streams/state/internals/QueryableStoreProvider.java#L59-L77 in favor of the one in here?

Of course, the short-circuit you're providing here is valuable in any case.

Thanks!
-John

vvcephei:
Alternatively, as you observed, if the parameters contain a single partition, then there should just be one specific store that matches. Instead of returning a `new WrappingStoreProvider` at all, maybe the QueryableStoreProvider should just have a high-level branch that, if the partition is speficied, then find the store and return a `new SingleStoreProvider`, otherwise return a `new WrappingStoreProvider`.

dima5rr:
Hi @vvcephei, thank you for the input.

After profiling under load it looks like problem in excessive loops over calling internalTopologyBuilder.topicGroups()
which is synchronized.
Though short-circuit WrappingStoreProvider can help, but will not solve root cause.

WDYT?

https://github.com/apache/kafka/blob/f699bd98c14e31f07c5a3f6ba9ce2c4b441e7fdb/streams/src/main/java/org/apache/kafka/streams/state/internals/StreamThreadStateStoreProvider.java#L113

https://github.com/apache/kafka/blob/f699bd98c14e31f07c5a3f6ba9ce2c4b441e7fdb/streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopologyBuilder.java#L1030

dima5rr:
Indeed StreamThreadStateStoreProvider does not need InternalTopologyBuilder in order to find stream task, it has all required data in StreamThread.

dima5rr:
> Hi @dima5rr , thanks for the PR! I was looking into the context of this ticket and noticed that we're essentially just duplicating here the logic in `org.apache.kafka.streams.state.internals.QueryableStoreProvider#getStore`, which also loops over all stores of all providers as a sanity check before even returning the WrappingStoreProvider, which will loop over all stores of all providers _again_ before returning any results.
> 
> Anecdotally, I think that most people would actually just query immediately and then discard their store reference, for example like `value = kafkaStreams.store("storeName", partition=2).get("key")`. In that case, this double-iteration is pretty expensive, and the up-front sanity check doesn't seem to provide any value.
> 
> In fact, the only value it could provide is if you _do_ plan to save the store reference and re-use it for multiple queries. But in that case, there could be a rebalance at any time, so checking up front probably doesn't help much no matter what the use case is.
> 
> What do you think about removing the loop
> 
> https://github.com/apache/kafka/blob/f699bd98c14e31f07c5a3f6ba9ce2c4b441e7fdb/streams/src/main/java/org/apache/kafka/streams/state/internals/QueryableStoreProvider.java#L59-L77
> 
> in favor of the one in here?
> Of course, the short-circuit you're providing here is valuable in any case.
> 
> Thanks!
> -John

Agree, it can be removed

dima5rr:
In addition, all synchronized methods in StreamsMetadataState.getKeyQueryMetadataForKey are also have performance implications under load.

https://github.com/apache/kafka/blob/598a0d16fa779cba9bb9ad0354236f269b2ff13d/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsMetadataState.java#L213

streams/src/main/java/org/apache/kafka/streams/state/internals/WrappingStoreProvider.java
```diff
@@ -46,11 +46,22 @@ public void setStoreQueryParameters(final StoreQueryParameters storeQueryParamet
...
+                allStores.addAll(stores);
+                if (storeQueryParameters.partition() != null) {
+                    break;
+                }
+            }
         }
         if (allStores.isEmpty()) {
+            if (storeQueryParameters.partition() != null) {
+                throw new InvalidStateStoreException(
+                        String.format("The specified partition %d for store %s does not exist.",
```
vvcephei, line 61: Is this really a different condition than the one on L65? It seems like the failure is still probably that the store "migrated" instead of "doesn't exist", right?

streams/src/main/java/org/apache/kafka/streams/state/internals/StreamThreadStateStoreProvider.java
```diff
@@ -20,62 +20,49 @@
...
-                }
-            } else {
-                for (final Task streamTask : tasks.values()) {
-                    final T store = validateAndListStores(streamTask.getStore(storeName), queryableStoreType, storeName, streamTask.id());
-                    if (store != null) {
-                        stores.add(store);
-                    }
-                }
+                final T store = validateAndListStores(streamTask.getStore(storeName), queryableStoreType, storeName, streamTask.id());
+                return store != null ? Collections.singletonList(store) : Collections.emptyList();
```
vvcephei: The nested early-return pattern is pretty hard to follow. Do you mind rewriting it to use if/else blocks? I know it was previously doing some early returns; it'd be better to migrate to a more maintainable style when we update the code, though.

vvcephei (commented):
Thanks for the update, and for adding the tests, and especially for profiling it!

I just had a couple of minor comments. Do you mind running the profiler again, just to make sure it lines up with your expectations?

Thanks,
-John

streams/src/main/java/org/apache/kafka/streams/state/internals/StreamThreadStateStoreProvider.java
dima5rr: sure

streams/src/main/java/org/apache/kafka/streams/state/internals/WrappingStoreProvider.java
dima5rr: L65 catches on rebalancing, while L60 is parameter validation for incorrect partition case.

streams/src/main/java/org/apache/kafka/streams/state/internals/StreamThreadStateStoreProvider.java
```diff
@@ -51,18 +51,22 @@ public StreamThreadStateStoreProvider(final StreamThread streamThread) {
...
+                tasks.values().stream().
+                        map(streamTask -> validateAndListStores(streamTask.getStore(storeName), queryableStoreType, storeName, streamTask.id())).
+                        filter(Objects::nonNull).
+                        forEach(stores::add);
             }
-            return tasks.values().stream().
-                    map(streamTask -> validateAndListStores(streamTask.getStore(storeName), queryableStoreType, storeName, streamTask.id())).
-                    filter(Objects::nonNull).
-                    collect(Collectors.toList());
+            return Collections.unmodifiableList(stores);
```
vvcephei: Ah, sorry, I can see that my prior comment was ambiguous. This is what I meant:
```java
            if (storeQueryParams.partition() == null) {
                return tasks.values().stream().
                        map(streamTask -> validateAndListStores(streamTask.getStore(storeName), queryableStoreType, storeName, streamTask.id())).
                        filter(Objects::nonNull).
                        collect(Collectors.toList());
            } else {
                final Task streamTask = findStreamTask(tasks, storeName, storeQueryParams.partition());
                if (streamTask == null) {
                    return Collections.emptyList();
                } else {
                    final T store = validateAndListStores(streamTask.getStore(storeName), queryableStoreType, storeName, streamTask.id());
                    return store == null ? Collections.emptyList() : Collections.singletonList(store);
                }
            }
```

The reason this is better for maintenence is that you only have to trace a path through the nested conditionals into a single inner block to understand what gets returned. I.e., code comprehension complexity is only the depth of the conditional tree.

In contrast, if we do early returns, you have to fully read all the conditional blocks that lead up to the one you're interested (depth-first traversal), so code comprehension is linear instead of logarithmic. If we mutate the collection, you actually have to read _all_ the conditionals to understand what is going to happen, so code comprehension is also linear instead of logarithmic.

dima5rr: Will concise it into functional way.

streams/src/main/java/org/apache/kafka/streams/state/internals/WrappingStoreProvider.java
guozhangwang: Could you elaborate a bit more about this? If `allStores.isEmpty()` is empty, it is always possible that the specified store-partition or just store-"null" does not exist in this client. Why they are different failure cases?

streams/src/test/java/org/apache/kafka/streams/integration/StreamStreamJoinIntegrationTest.java
```diff
@@ -60,6 +60,34 @@ public void prepareTopology() throws InterruptedException {
         rightStream = builder.stream(INPUT_TOPIC_RIGHT);
     }
 
+    @Test
+    public void shouldNotAccessJoinStoresWhenGivingName() throws InterruptedException {
```
guozhangwang: A good coverage improvement! Thanks.

streams/src/main/java/org/apache/kafka/streams/state/internals/StreamThreadStateStoreProvider.java
```diff
@@ -104,19 +95,11 @@ public StreamThreadStateStoreProvider(final StreamThread streamThread,
...
-        final Set<String> sourceTopicsSet = new HashSet<>(sourceTopics);
-        final Map<Integer, InternalTopologyBuilder.TopicsInfo> topicGroups = internalTopologyBuilder.topicGroups();
-        for (final Map.Entry<Integer, InternalTopologyBuilder.TopicsInfo> topicGroup : topicGroups.entrySet()) {
-            if (topicGroup.getValue().sourceTopics.containsAll(sourceTopicsSet)) {
-                return new TaskId(topicGroup.getKey(), partition);
-            }
-        }
-        throw new InvalidStateStoreException("Cannot get state store " + storeName + " because the requested partition " +
-            partition + " is not available on this instance");
+    private Optional<Task> findStreamTask(final Collection<Task> tasks, final String storeName, final int partition) {
```
guozhangwang, line 98: This is a great find, thanks!

streams/src/main/java/org/apache/kafka/streams/state/internals/QueryableStoreProvider.java
```diff
@@ -56,25 +55,6 @@ public QueryableStoreProvider(final List<StreamThreadStateStoreProvider> storePr
         if (!globalStore.isEmpty()) {
             return queryableStoreType.create(globalStoreProvider, storeName);
         }
-        final List<T> allStores = new ArrayList<>();
```
guozhangwang, line 59: LGTM.

guozhangwang (commented):
Thanks a great find, thanks @dima5rr ! I just had a question about one failure handling case, other wise it LGTM! cc @vvcephei

streams/src/main/java/org/apache/kafka/streams/state/internals/WrappingStoreProvider.java
vvcephei: Hey @dima5rr , I think Guozhang's question was hidden because the conversation was already "resolved". Do you mind answering this concern?

dima5rr: Hey @guozhangwang, you're right, this check is ambiguous, it's more likely parameter sanity validation when user explicitly specify a single partition.

guozhangwang: Got it, in that case how about we just encode the partition in the thrown's message so that upon throwing, people can still check if the partition is `null` or not when debugging?

Otherwise, this PR all LGTM :)

dima5rr: Hey @guozhangwang, I am just care that in case of partition is null, the error message is referenced in official FAQ.

https://docs.confluent.io/current/streams/faq.html#handling-invalidstatestoreexception-the-state-store-may-have-migrated-to-another-instance

guozhangwang: That's a fair point, let's just merge it as is then.

guozhangwang:
test this please

guozhangwang:
test this please

guozhangwang:
test this

guozhangwang:
test this

guozhangwang:
@dima5rr I tried to compile your branch but got a few compilation error like the following:

```
/Users/guozhang/Workspace/github/guozhangwang/kafka-work/streams/src/test/java/org/apache/kafka/streams/integration/StreamStreamJoinIntegrationTest.java:68: error: cannot find symbol
        final KStream<String, Integer> left = builder.stream(INPUT_TOPIC_LEFT, Consumed.with(Serdes.String(), Serdes.Integer()));
                                                                                             ^
  symbol:   variable Serdes
  location: class StreamStreamJoinIntegrationTest
/Users/guozhang/Workspace/github/guozhangwang/kafka-work/streams/src/test/java/org/apache/kafka/streams/integration/StreamStreamJoinIntegrationTest.java:68: error: cannot find symbol
        final KStream<String, Integer> left = builder.stream(INPUT_TOPIC_LEFT, Consumed.with(Serdes.String(), Serdes.Integer()));
```

dima5rr:
Hi @guozhangwang can you trigger new build, looks like flaky tests?

guozhangwang:
test this please

guozhangwang:
test this please

guozhangwang:
test this

guozhangwang:
test this

dima5rr:
test this please

guozhangwang:
Test passed, merged to trunk.

Thanks @dima5rr for your great contribution!

guozhangwang:
Cherry-picked to 2.6 as well.

PR #9022, opened 2020-07-14
KAFKA-10158: Fix flaky testDescribeUnderReplicatedPartitionsWhenReassignmentIsInProgress

bdbyrne:
Set `replica.fetch.max.bytes` to `1` and produce multiple record batches to allow
for throttling to take place. This helps avoid a race condition where the
reassignment would complete more quickly than expected causing an
assertion to fail some times.

### DIFF
--- a/core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
+++ b/core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
@@ -56,11 +56,16 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
     zkConnect = zkConnect,
     rackInfo = Map(0 -> "rack1", 1 -> "rack2", 2 -> "rack2", 3 -> "rack1", 4 -> "rack3", 5 -> "rack3"),
     numPartitions = numPartitions,
-    defaultReplicationFactor = defaultReplicationFactor
+    defaultReplicationFactor = defaultReplicationFactor,
+    replicaFetchMaxBytes = replicaFetchMaxBytes(),
   ).map(KafkaConfig.fromProps)
 
   private val numPartitions = 1
   private val defaultReplicationFactor = 1.toShort
+
+  private def replicaFetchMaxBytes() =
+    if (testName.getMethodName == "testDescribeUnderReplicatedPartitionsWhenReassignmentIsInProgress") Some(1)
+    else None
 
   private var topicService: AdminClientTopicService = _
   private var adminClient: Admin = _
@@ -672,11 +677,9 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
     adminClient.createTopics(
       Collections.singletonList(new NewTopic(testTopicName, partitions, replicationFactor).configs(configMap))).all().get()
     waitForTopicCreated(testTopicName)
-    TestUtils.generateAndProduceMessages(servers, testTopicName, numMessages = 10, acks = -1)
+    TestUtils.generateAndProduceMessages(servers, testTopicName, numMessages = 1000, acks = -1)
 
     val brokerIds = servers.map(_.config.brokerId)
-    TestUtils.setReplicationThrottleForPartitions(adminClient, brokerIds, Set(tp), throttleBytes = 1)
-
     val testTopicDesc = adminClient.describeTopics(Collections.singleton(testTopicName)).all().get().get(testTopicName)
     val firstPartition = testTopicDesc.partitions().asScala.head
 
@@ -703,7 +706,10 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
       topicService.describeTopic(new TopicCommandOptions(Array("--under-replicated-partitions"))))
     assertEquals(s"--under-replicated-partitions shouldn't return anything: '$underReplicatedOutput'", "", underReplicatedOutput)
 
-    TestUtils.removeReplicationThrottleForPartitions(adminClient, brokerIds, Set(tp))
+    // Verify reassignment is still ongoing.
+    val reassignments = adminClient.listPartitionReassignments(Collections.singleton(tp)).reassignments().get().get(tp)
+    assertFalse(Option(reassignments).forall(_.addingReplicas().isEmpty))
+
     TestUtils.waitForAllReassignmentsToComplete(adminClient)
   }
 
--- a/core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
+++ b/core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
@@ -712,12 +712,11 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
       topicService.describeTopic(new TopicCommandOptions(Array("--under-replicated-partitions"))))
     assertEquals(s"--under-replicated-partitions shouldn't return anything: '$underReplicatedOutput'", "", underReplicatedOutput)
 
-    TestUtils.removeReplicationThrottleForPartitions(adminClient, brokerIds, Set(tp))
-
     // Verify reassignment is still ongoing.
     val reassignments = adminClient.listPartitionReassignments(Collections.singleton(tp)).reassignments().get().get(tp)
     assertFalse(Option(reassignments).forall(_.addingReplicas().isEmpty))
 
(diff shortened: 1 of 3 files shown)


core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
```diff
@@ -672,11 +677,9 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
     adminClient.createTopics(
       Collections.singletonList(new NewTopic(testTopicName, partitions, replicationFactor).configs(configMap))).all().get()
     waitForTopicCreated(testTopicName)
-    TestUtils.generateAndProduceMessages(servers, testTopicName, numMessages = 10, acks = -1)
+    TestUtils.generateAndProduceMessages(servers, testTopicName, numMessages = 1000, acks = -1)
```
bdbyrne: Note that although we're bumping this to 1000, it should be a single produce batch with really small data, so on the order of a couple KBs.

bdbyrne:
Please review, @lbradstreet - thanks!

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
lbradstreet: It might be worth adding a comment somewhere in `testDescribeUnderReplicatedPartitionsWhenReassignmentIsInProgress` to describe what was done in the replicaFetchMaxBytes def and why.

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
```diff
@@ -672,11 +677,9 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
...
 
     val brokerIds = servers.map(_.config.brokerId)
-    TestUtils.setReplicationThrottleForPartitions(adminClient, brokerIds, Set(tp), throttleBytes = 1)
```
lbradstreet, line 678: Why did we remove this? We still need the throttle to be set so that the replica doesn't join the ISR. It could easily do this even if it only fetches a message at a time. The idea with the new replica fetch max bytes setting is to give the broker a chance to throttle itself before joining the ISR.

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
lbradstreet: With the throttle approach we don't want it to be produced in a single batch. If all messages are produced in a single batch then you will still fetch the entire batch in a single request even if replica.fetch.max.bytes is 1, as we allow this limit to be broken to fetch at least one message.

bdbyrne: So each message creates its own `ProducerRecord`, I was under the assumption that the limit to fetch expands to a single record, and not the batch itself. Or am I misunderstanding?

bdbyrne: Done.

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
bdbyrne: Added back, but to clarify - are you suggesting the high watermark could still be at 0 (or relatively low) even after producing the messages? Or just that both combined are a more effective throttle, essentially putting it at 1 record/sec?

bdbyrne (commented):
Thanks for the review, a couple questions.

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
lbradstreet: Each message creates its own ProducerRecord but these get collected into batches by the producer. Fetch requests will not break up a batch unless down-converting to the old record batch format, so you could end up with 1000 messages in a single batch being fetched in a single fetch request.

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
```diff
@@ -673,10 +678,14 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
       Collections.singletonList(new NewTopic(testTopicName, partitions, replicationFactor).configs(configMap))).all().get()
     waitForTopicCreated(testTopicName)
     TestUtils.generateAndProduceMessages(servers, testTopicName, numMessages = 10, acks = -1)
+    Thread.sleep(10)
```
lbradstreet: Do we want to leave a comment here to say that we are trying to ensure that messages are produced in multiple batches? I do wonder if there's a better way to achieve this via a producer setting, e.g. `batch.size`?

lbradstreet (approved):
The fix works and the test is no longer flaky with the changes made. Someone may have a better idea about how to avoid this kind of check which will be brittle if someone renames the test https://github.com/apache/kafka/pull/9022/files#diff-0bc85b210490e2fc83d3b11c6a9d83dfR67

hachikuji:
ok to test

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
```diff
@@ -56,12 +56,17 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
...
-    defaultReplicationFactor = defaultReplicationFactor
+    defaultReplicationFactor = defaultReplicationFactor,
+    replicaFetchMaxBytes = replicaFetchMaxBytes(),
   ).map(KafkaConfig.fromProps)
 
   private val numPartitions = 1
   private val defaultReplicationFactor = 1.toShort
 
+  private def replicaFetchMaxBytes() =
+    if (testName.getMethodName == "testDescribeUnderReplicatedPartitionsWhenReassignmentIsInProgress") Some(1)
```
hachikuji: It's always an annoyance with Junit that there is no way for the test case to override initialization in a `@Before`. What we often end up doing is removing the annotation and calling the setup method explicitly in each test case. I slightly prefer that option since it is easier to understand, but not sure it is possible since we're extending `KafkaServerTestHarness`, which has its own initialization logic. For the sake of argument, would it be possible to set max fetch bytes to 1 for all tests? Either that or maybe we should just produce more data in the test case.

bdbyrne: Agreed, the `KafkaServerTestHarness` makes it more difficult. I've updated the test to set max fetch bytes to 1 for all tests, which is fine given none of the other tests produce data.

hachikuji:
retest this please

chia7712 (approved):

omkreddy:
ok to test

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
lbradstreet: I simply meant that we still need the replication throttle so the reassignment doesn't complete before we perform our checks.

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
ijuma: The broker only works in terms of record batches, not individual records.

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
ijuma: We can call `flush` on the producer to force it to send the messages.

ijuma: Looking a bit more, we actually close the producer after we generate the messages, so I don't think this sleep is needed at all.

ijuma (approved):
I pushed some minor fixes. LGTM now.

ijuma:
ok to test

ijuma:
https://builds.apache.org/job/kafka-pr-jdk11-scala2.13/

ijuma:
retest this please

ijuma:
https://builds.apache.org/job/kafka-pr-jdk11-scala2.13/

ijuma:
Green build, merging to trunk, 2.6 and 2.5.

PR #9030, opened 2020-07-16
MINOR: Filter out quota configs for ConfigCommand using --bootstrap-server

rajinisivaram:

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/requests/DescribeClientQuotasResponse.java
+++ b/clients/src/main/java/org/apache/kafka/common/requests/DescribeClientQuotasResponse.java
@@ -66,10 +66,11 @@ public class DescribeClientQuotasResponse extends AbstractResponse {
     }
 
     public DescribeClientQuotasResponse(int throttleTimeMs, Throwable e) {
+        ApiError apiError = ApiError.fromThrowable(e);
         this.data = new DescribeClientQuotasResponseData()
                 .setThrottleTimeMs(throttleTimeMs)
-                .setErrorCode(Errors.forException(e).code())
-                .setErrorMessage(e.getMessage())
+                .setErrorCode(apiError.error().code())
+                .setErrorMessage(apiError.message())
                 .setEntries(null);
     }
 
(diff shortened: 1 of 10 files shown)


clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -197,18 +197,20 @@
 
 public class RequestResponseTest {
 
+    private UnknownServerException unknownServerException = new UnknownServerException("secret");
```
dajac: nit: I would add a small comment to explain why we do this here.

dajac (commented):
Thanks for the PR. I just left one small suggestion.

clients/src/main/java/org/apache/kafka/common/requests/DescribeClientQuotasResponse.java
```diff
@@ -66,10 +66,11 @@ public DescribeClientQuotasResponse(Map<ClientQuotaEntity, Map<String, Double>>
     }
 
     public DescribeClientQuotasResponse(int throttleTimeMs, Throwable e) {
+        ApiError apiError = ApiError.fromThrowable(e);
```
rondagostino, line 71: Should the same thing happen in `public AlterClientQuotasResponse(Collection<ClientQuotaEntity> entities, int throttleTimeMs, Throwable e)` for consistency?  And also maybe in `IncrementalAlterConfigsResponse` (`public static IncrementalAlterConfigsResponseData toResponseData`)?

rondagostino (commented):
LGTM, left one question.

clients/src/main/java/org/apache/kafka/common/requests/DescribeClientQuotasResponse.java
rajinisivaram: @rondagostino Thanks for the review. Updated `AlterClientQuotasResponse`. I think IncrementalAlterConfigsResponse is ok since it uses ApiError, we create the error with the right message when converting from Throwable. Can you just verify that the method with ApiError was the one you meant? Thanks,

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
rajinisivaram: @dajac Thanks for the review, added comment.

rondagostino (commented):
LGTM!

omkreddy (approved):

rajinisivaram:
@dajac @rondagostino @omkreddy Thanks for the review, merging to trunk, 2.6 and 2.5.

PR #8864, opened 2020-06-13
KAFKA-9274: Mark `retries` config as deprecated and add new `task.timeout.ms` config

mjsax:
First PR for KIP-572 for public changes only. Producer/Admin client won't have any functional change as they still will respect `retries`. For KS, we will get rid or the usage or `retries` in follow up PRs.

Call for review: @cmccabe @hachikuji @abbccdda @vvcephei

PR #4855, opened 2018-04-12
KAFKA-5697: prevent poll() from blocking forever

vvcephei:
Add the new stricter-timeout version of `poll` proposed in KIP-266.

The pre-existing variant `poll(long timeout)` would block indefinitely for metadata
updates if they were needed, then it would issue a fetch and poll for `timeout` ms 
for new records. The initial indefinite metadata block caused applications to become
stuck when the brokers became unavailable. The existence of the timeout parameter
made the indefinite block especially unintuitive.

This PR adds `poll(Duration timeout)` with the semantics:
1. iff a metadata update is needed:
    1. send (asynchronous) metadata requests
    2. poll for metadata responses (counts against timeout)
        - if no response within timeout, **return an empty collection immediately**
2. if there is fetch data available, **return it immediately**
3. if there is no fetch request in flight, send fetch requests
4. poll for fetch responses (counts against timeout)
    - if no response within timeout, **return an empty collection** (leaving async fetch request for the next poll)
    - if we get a response, **return the response**

The old method, `poll(long timeout)` is deprecated, but we do not change its semantics, so it remains:
1. iff a metadata update is needed:
    1. send (asynchronous) metadata requests
    2. poll for metadata responses *indefinitely until we get it*
2. if there is fetch data available, **return it immediately**
3. if there is no fetch request in flight, send fetch requests
4. poll for fetch responses (counts against timeout)
    - if no response within timeout, **return an empty collection** (leaving async fetch request for the next poll)
    - if we get a response, **return the response**

One notable usage is prohibited by the new `poll`: previously, you could call `poll(0)` to block for metadata updates, for example to initialize the client, supposedly without fetching records. Note, though, that this behavior is not according to any contract, and there is no guarantee that `poll(0)` won't return records the first time it's called. Therefore, it has always been unsafe to ignore the response.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9047, opened 2020-07-21
KAFKA-9274: Remove `retries` for global task

mjsax:
- part of KIP-572
 - removed the usage of `retries` in `GlobalStateManger`
 - instead of retries the new `task.timeout.ms` config is used

Second PR for KIP-572 (cf. #8864)

Call for review @vvcephei 

Docs Screenshots:

Main upgrade guide:

Screen Shot 2020-08-04 at 10 26 48 AM

Streams upgrade guide:

Screen Shot 2020-08-04 at 10 27 24 AM

Snippet from StreamConfig:

Screen Shot 2020-08-04 at 10 28 37 AM


streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
```diff
@@ -523,6 +524,8 @@
     public static final String STATE_DIR_CONFIG = "state.dir";
     private static final String STATE_DIR_DOC = "Directory location for state store. This path must be unique for each streams instance sharing the same underlying filesystem.";
 
+    public static final String TASK_TIMEOUT_MS_CONFIG = "task.timeout.ms";
```
mjsax, line 531: This is just added until the first PR is merged to unblock the work on this PR.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -184,32 +217,21 @@ public void registerStore(final StateStore store, final StateRestoreCallback sta
...
+            highWatermarks = globalConsumer.endOffsets(topicPartitions);
+        } catch (final TimeoutException retryableException) {
+            log.debug(
+                "Failed to get end offsets for partitions {}. The broker may be transiently unavailable at the moment. Will retry.",
+                topicPartitions,
+                retryableException
+            );
+
+            // handled in `GlobalStateMangerImpl#initialize()`
+            throw retryableException;
```
vvcephei: Minor note: it's confusing to track down exceptions when they are re-thrown like this, because the stacktrace would only reference L223. Even though there is a small performance penalty, it's better for maintainability to always throw a new exception like `new RetryableException(timeoutException)`.

In this particular case, it may be more appropriate just to remove the try-catch and add a `throws` declaration. There are two things that make me think this:
1. The log message here says, "will retry", but this method can have no idea whether or not it'll be retried
2. The calling method has a comment that says this method logs the error, which is also an assumption that may not survive refactoring

It seems like we can resolve all three of these maintenence problems by just moving the log message to the caller.

This feedback also applies elsewhere.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -184,32 +217,21 @@ public void registerStore(final StateStore store, final StateRestoreCallback sta
...
-                Utils.sleep(retryBackoffMs);
-            }
+        final Map<TopicPartition, Long> highWatermarks;
+        try {
+            highWatermarks = globalConsumer.endOffsets(topicPartitions);
+        } catch (final TimeoutException retryableException) {
+            log.debug(
+                "Failed to get end offsets for partitions {}. The broker may be transiently unavailable at the moment. Will retry.",
+                topicPartitions,
+                retryableException
```
vvcephei: The varargs version of `debug` does _not_ take a cause at the end. This exception will not be logged. You have to use the version that only takes `(String, Exception)`.

streams/src/test/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImplTest.java
```diff
@@ -613,69 +617,262 @@ public boolean lockGlobalState() throws IOException {
...
+                throw new TimeoutException("KABOOM!");
+            }
+        };
+        consumer.updatePartitions(t1.topic(), Collections.singletonList(new PartitionInfo(t1.topic(), t1.partition(), null, null, null)));
+        consumer.updatePartitions(t2.topic(), Collections.singletonList(new PartitionInfo(t2.topic(), t2.partition(), null, null, null)));
+        consumer.updatePartitions(t3.topic(), Collections.singletonList(new PartitionInfo(t3.topic(), t3.partition(), null, null, null)));
+        consumer.updatePartitions(t4.topic(), Collections.singletonList(new PartitionInfo(t4.topic(), t4.partition(), null, null, null)));
+
+        streamsConfig = new StreamsConfig(new Properties() {
+            {
```
vvcephei: Double-brace initialization is an anti-pattern. It would be preferable to use `mkProperties`.

vvcephei:
Test failures:
```
kafka.api.PlaintextAdminIntegrationTest.testAlterReplicaLogDirs failed, log available in /home/jenkins/jenkins-slave/workspace/kafka-pr-jdk11-scala2.13/core/build/reports/testOutput/kafka.api.PlaintextAdminIntegrationTest.testAlterReplicaLogDirs.test.stdout

kafka.api.PlaintextAdminIntegrationTest > testAlterReplicaLogDirs FAILED
    org.scalatest.exceptions.TestFailedException: only 0 messages are produced within timeout after replica movement. Producer future Some(Failure(java.util.concurrent.TimeoutException: Timeout after waiting for 10000 ms.))
        at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:530)
        at org.scalatest.Assertions.newAssertionFailedException$(Assertions.scala:529)
        at org.scalatest.Assertions$.newAssertionFailedException(Assertions.scala:1389)
        at org.scalatest.Assertions.fail(Assertions.scala:1091)
        at org.scalatest.Assertions.fail$(Assertions.scala:1087)
        at org.scalatest.Assertions$.fail(Assertions.scala:1389)
        at kafka.api.PlaintextAdminIntegrationTest.testAlterReplicaLogDirs(PlaintextAdminIntegrationTest.scala:289)
```

```
org.apache.kafka.connect.integration.SourceConnectorsIntegrationTest.testTopicsAreCreatedWhenAutoCreateTopicsIsEnabledAtTheBroker failed, log available in /home/jenkins/jenkins-slave/workspace/kafka-pr-jdk8-scala2.12/connect/runtime/build/reports/testOutput/org.apache.kafka.connect.integration.SourceConnectorsIntegrationTest.testTopicsAreCreatedWhenAutoCreateTopicsIsEnabledAtTheBroker.test.stdout

org.apache.kafka.connect.integration.SourceConnectorsIntegrationTest > testTopicsAreCreatedWhenAutoCreateTopicsIsEnabledAtTheBroker FAILED
    java.lang.NullPointerException
        at java.util.concurrent.ConcurrentHashMap.get(ConcurrentHashMap.java:936)
        at org.reflections.Store.getAllIncluding(Store.java:82)
        at org.reflections.Store.getAll(Store.java:93)
        at org.reflections.Reflections.getSubTypesOf(Reflections.java:404)
        at org.apache.kafka.connect.runtime.isolation.DelegatingClassLoader.getPluginDesc(DelegatingClassLoader.java:355)
        at org.apache.kafka.connect.runtime.isolation.DelegatingClassLoader.scanPluginPath(DelegatingClassLoader.java:340)
        at org.apache.kafka.connect.runtime.isolation.DelegatingClassLoader.scanUrlsAndAddPlugins(DelegatingClassLoader.java:268)
        at org.apache.kafka.connect.runtime.isolation.DelegatingClassLoader.initPluginLoader(DelegatingClassLoader.java:216)
        at org.apache.kafka.connect.runtime.isolation.DelegatingClassLoader.initLoaders(DelegatingClassLoader.java:209)
        at org.apache.kafka.connect.runtime.isolation.Plugins.<init>(Plugins.java:61)
        at org.apache.kafka.connect.cli.ConnectDistributed.startConnect(ConnectDistributed.java:91)
        at org.apache.kafka.connect.util.clusters.WorkerHandle.start(WorkerHandle.java:50)
        at org.apache.kafka.connect.util.clusters.EmbeddedConnectCluster.addWorker(EmbeddedConnectCluster.java:167)
        at org.apache.kafka.connect.util.clusters.EmbeddedConnectCluster.startConnect(EmbeddedConnectCluster.java:253)
        at org.apache.kafka.connect.util.clusters.EmbeddedConnectCluster.start(EmbeddedConnectCluster.java:137)
        at org.apache.kafka.connect.integration.SourceConnectorsIntegrationTest.testTopicsAreCreatedWhenAutoCreateTopicsIsEnabledAtTheBroker(SourceConnectorsIntegrationTest.java:104)
```

vvcephei (commented):
Thanks for this PR, @mjsax !

The main behavioral change looks good to me. I had a couple of small suggestions.

docs/streams/developer-guide/config-streams.html
```diff
@@ -203,7 +203,7 @@ <h4><a class="toc-backref" href="#id5">bootstrap.servers</a><a class="headerlink
           </tr>
           <tr class="row-even"><td>commit.interval.ms</td>
             <td>Low</td>
-            <td colspan="2">The frequency with which to save the position (offsets in source topics) of tasks.</td>
+            <td colspan="2">The frequency in milliseconds with which to save the position (offsets in source topics) of tasks.</td>
```
mjsax, line 206: Added a couple of side fixed for the docs.

docs/streams/developer-guide/config-streams.html
```diff
@@ -308,15 +308,10 @@ <h4><a class="toc-backref" href="#id5">bootstrap.servers</a><a class="headerlink
             <td colspan="2">The replication factor for changelog topics and repartition topics created by the application.</td>
             <td>1</td>
           </tr>
-          <tr class="row-odd"><td>retries</td>
```
mjsax, line 311: Forgot to remove `retries` in the first PR.

docs/streams/developer-guide/config-streams.html
```diff
@@ -326,13 +321,18 @@ <h4><a class="toc-backref" href="#id5">bootstrap.servers</a><a class="headerlink
...
             <td colspan="2">The amount of time in milliseconds to wait before deleting state when a partition has migrated.</td>
-            <td>600000 milliseconds</td>
+            <td>600000 milliseconds (10 minutes)</td>
           </tr>
           <tr class="row-odd"><td>state.dir</td>
             <td>High</td>
             <td colspan="2">Directory location for state stores.</td>
             <td><code class="docutils literal"><span class="pre">/tmp/kafka-streams</span></code></td>
           </tr>
+          <tr class="row-odd"><td>task.timeout.ms</td>
```
mjsax, line 331: Forgot to add the new config in the first PR

docs/streams/upgrade-guide.html
```diff
@@ -95,11 +95,12 @@ <h3><a id="streams_api_changes_270" href="#streams_api_changes_270">Streams API
     </p>
 
     <p>
-        The configuration parameter <code>retries</code> is deprecated in favor of a the new parameter <code>task.timeout.ms</code>.
```
mjsax, line 98: Fixed some typos. And added reference to `max.block.ms` config

docs/upgrade.html
```diff
@@ -23,8 +23,8 @@ <h5><a id="upgrade_270_notable" href="#upgrade_270_notable">Notable changes in 2
 <ul>
     <li>The configuration parameter <code>retries</code> is deprecated for the producer, admin, and Kafka Streams clients
         via <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-572%3A+Improve+timeouts+and+retries+in+Kafka+Streams">KIP-572</a>.
-        You should use the producer's <code>delivery.timeout.ms</code>, admin's <code>default.api.timeout.ms</code>, and
-        Kafka Streams' new <code>task.timeout.ms</code> parameters instead.
+        You should use the producer's <code>delivery.timeout.ms</code> and <code>max.block.ms</code>, admin's
```
mjsax, line 26: As above.

streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
```diff
@@ -357,7 +357,7 @@
     /** {@code commit.interval.ms} */
     @SuppressWarnings("WeakerAccess")
     public static final String COMMIT_INTERVAL_MS_CONFIG = "commit.interval.ms";
-    private static final String COMMIT_INTERVAL_MS_DOC = "The frequency with which to save the position of the processor." +
+    private static final String COMMIT_INTERVAL_MS_DOC = "The frequency in milliseconds with which to save the position of the processor." +
```
mjsax, line 360: Same fixes as in the docs.

streams/src/main/java/org/apache/kafka/streams/errors/RetryableErrorException.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.errors;
+
+public class RetryableErrorException extends StreamsException {
```
mjsax: Added new exception type as requested.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -58,30 +60,33 @@
...
-    private final long retryBackoffMs;
     private final Duration pollTime;
+    private final long taskTimeoutMs;
     private final Set<String> globalNonPersistentStoresTopics = new HashSet<>();
     private final OffsetCheckpoint checkpointFile;
     private final Map<TopicPartition, Long> checkpointFileCache;
     private final Map<String, String> storeToChangelogTopic;
     private final List<StateStore> globalStateStores;
 
-    @SuppressWarnings("deprecation") // TODO: remove in follow up PR when `RETRIES` is removed
```
mjsax, line 77: This PR fixed 3 TODOs form the first PR. (The other two are in the test -- also added a comment to the original PR that links to this PR as reference.)

streams/src/test/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImplTest.java
```diff
@@ -612,72 +617,521 @@ public boolean lockGlobalState() throws IOException {
         }
     }
 
-    @SuppressWarnings("deprecation") // TODO revisit in follow up PR
```
mjsax, line 599: Second TODO

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -131,11 +135,40 @@ public void setGlobalProcessorContext(final InternalProcessorContext globalProce
...
             final String sourceTopic = storeToChangelogTopic.get(stateStore.name());
             changelogTopics.add(sourceTopic);
-            stateStore.init(globalProcessorContext, stateStore);
+
+            try {
+                stateStore.init(globalProcessorContext, stateStore);
+                deadlineMs = NO_DEADLINE;
+            } catch (final RetryableErrorException retryableException) {
+                if (taskTimeoutMs == 0L) {
+                    throw new StreamsException(retryableException.getCause());
```
mjsax: If we rethrow, we get rid of the `RetryableErrorException` and pass in the original root cause, ie, the `TimeoutException`.

streams/src/test/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImplTest.java
```diff
@@ -612,72 +617,521 @@ public boolean lockGlobalState() throws IOException {
...
+            () -> stateManager.initialize()
+        );
+        final Throwable cause = expected.getCause();
+        assertThat(cause, instanceOf(TimeoutException.class));
+        assertThat(cause.getMessage(), equalTo("KABOOM!"));
+
+        assertEquals(numberOfCalls.get(), 1);
     }
 
-    @SuppressWarnings("deprecation") // TODO revisit in follow up PR
```
mjsax, line 633: Third TODO

streams/src/test/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImplTest.java
```diff
@@ -612,72 +617,521 @@ public boolean lockGlobalState() throws IOException {
...
+            TimeoutException.class,
+            () -> stateManager.initialize()
+        );
+        assertThat(expected.getMessage(), equalTo("Global task did not make progress to restore state within 1100 ms. Adjust `task.timeout.ms` if needed."));
+
+        assertEquals(numberOfCalls.get(), 12);
+    }
+
+    @Test
+    public void shouldNotFailOnSlowProgressWhenEndOffsetsThrowsTimeoutException() {
```
mjsax, line 734: This test is new (also added it for `partitionFor()` case).

streams/src/test/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImplTest.java
```diff
@@ -612,72 +617,521 @@ public boolean lockGlobalState() throws IOException {
...
+            streamsConfig
+        );
+        processorContext.setStateManger(stateManager);
+        stateManager.setGlobalProcessorContext(processorContext);
+
+        stateManager.initialize();
+    }
+
+    @Test
+    public void shouldNotRetryWhenPositionThrowsTimeoutExceptionAndTaskTimeoutIsZero() {
```
mjsax, line 942: Replicated the tests from above (`endOffset` and `partitionFor`) for the `position` call.

mjsax:
@vvcephei Updated this PR.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -299,7 +318,17 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
...
+                        // a successful `position()` call above for the requested partition and thus the consumer
+                        // should have a valid local position that it can return immediately
+
+                        // hence, a `TimeoutException` indicates a bug and thus we rethrow it as fatal `IllegalStateException`
+                        throw new IllegalStateException(error);
+                    }
+
                     stateRestoreAdapter.restoreBatch(restoreRecords);
                     stateRestoreListener.onBatchRestored(topicPartition, storeName, offset, restoreRecords.size());
                     restoreCount += restoreRecords.size();
```
mjsax: @vvcephei I actually had a follow up thought: given that we fetch data for a single partition only, should we trigger a timeout within this loop if `records` is empty (ie, `poll() did not return anything)? Otherwise, this restore loop might "hang" forever if we lose the connection to the broker -- IMHO, `task.timeout.ms` should cover this case? Otherwise, we block on startup as the initial global store loading would never finish and we don't even start the actual `StreamThreads`.

mjsax:
Flaky tests only.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: Why not just preserve the whole story of what happened, like `throw new StreamsException("Couldn't retry because timeout is set to zero", retryableException)`?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -131,11 +135,40 @@ public void setGlobalProcessorContext(final InternalProcessorContext globalProce
...
+                    final long newDeadlineMs = currentWallClockMs + taskTimeoutMs;
+                    deadlineMs = newDeadlineMs < 0L ? Long.MAX_VALUE : newDeadlineMs;
+                } else if (currentWallClockMs > deadlineMs) {
+                    throw new TimeoutException(String.format(
+                        "Global task did not make progress to restore state within %d ms. Adjust `task.timeout.ms` if needed.",
+                        currentWallClockMs - deadlineMs + taskTimeoutMs
+                    ));
+                }
+
+                log.debug(retryableException.getMessage() + " Will retry. Remaining time in milliseconds: {}", deadlineMs - currentWallClockMs);
```
vvcephei:
```suggestion
                log.debug(String.format("Got an exception in store.init(). Will retry. Remaining time in milliseconds: %d",  deadlineMs - currentWallClockMs), retryableException);
```

Sorry, my prior feedback was ambiguous. This is what I meant.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -299,7 +318,17 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
                             restoreRecords.add(recordConverter.convert(record));
                         }
                     }
-                    offset = globalConsumer.position(topicPartition);
+                    try {
+                        offset = globalConsumer.position(topicPartition);
+                    } catch (final TimeoutException error) {
+                        // the `globalConsumer.position()` call should never block, because we know that we did
+                        // a successful `position()` call above for the requested partition and thus the consumer
+                        // should have a valid local position that it can return immediately
```
vvcephei: I think this makes sense, but it seems to make some assumptions about the internal implementation of the consumer that doesn't seem necessary here. Is it important to assert that the consumer can't possibly get a timeout exception here?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -299,7 +318,17 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
                             restoreRecords.add(recordConverter.convert(record));
                         }
                     }
-                    offset = globalConsumer.position(topicPartition);
+                    try {
```
vvcephei: I can't comment above, but can the `poll` call itself throw a timeout exception? Or does it always just return no results in that case?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: Huh, interesting thought. Just to be clear, it looks like it would already block forever in this case today, right?
Yeah, it does seem like we should implement a similar non-progress timeout for this loop in that case.

vvcephei (commented):
Thanks for the update @mjsax , very nice!

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: Ack

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: I understood what you meant, but I thought that because we actual do retry, it's better/cleaner/less-noisy to not log the full stack trace?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: > I can't comment above, but can the poll call itself throw a timeout exception?

No, it cannot. It would just return an empty result.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: I guess strictly speaking, it's not necessary, but I personally prefer this strict style as it surfaces bugs quicker. If we write "robust" code, it could hide a bug. -- It's just a personal preference and we have similar code elsewhere. If you feel very strong about it, I am also ok to change it.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: Yes, it would block forever today. Curious to hear what @guozhangwang thinks.

guozhangwang: My understanding is that for non-global state stores, we would also start ticking if we cannot make progress either due to exceptions or `poll()` returned no data, is that right? If yes, I'm +1 on covering the same for global state store here.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: Ah, good, at least it wasn't lack of clarity. It's your call. I've wasted quite a bit of time trying to track down root causes of things specifically because of stuff like this. It really doesn't seem like saving ten log lines is worth the extra hassle.

I'm not saying it's "bad" the way you proposed. But it is unusual, and unusual code is more often than not a bummer in the long run. Exceptions in java take a "cause" for a reason, and they print the stacktraces of those causes in a well understood order and format that integrates well with all kinds of tools from IDEs to analytics and log aggregation. I agree that the stacktrace isn't that useful in this case. Still, specifically breaking this whole ecosystem because we don't think the stacktrace is likely useful seems a poor tradeoff.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: Thanks; yes, this is a very good habit. In this specific case, I wasn't sure what the "bug" would be, though. Just that the implementation of the consumer does something different, but equally correct, from our current understanding of how it works?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: Thanks!

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: The bug would be in the consumer: it forgets its position and needs to fetch it from the broker again (of course, we would only expose this bug, if there is also a broker/network issue and the request actually times out... what makes it a rather weak "check").

mjsax:
@vvcephei I updated the PR and needed to do quite some changes to get it into reasonable shape. Instead of throwing a new `RetryableException` that we catch in the outer layer, I just retry in each step directly. This make reasoning about the timeout easier.

streams/src/main/java/org/apache/kafka/streams/internals/QuietStreamsConfig.java
```diff
@@ -1,33 +0,0 @@
...
-
-import java.util.Map;
-
-/**
- * A {@link StreamsConfig} that does not log its configuration on construction.
- *
- * This producer cleaner output for unit tests using the {@code test-utils},
- * since logging the config is not really valuable in this context.
- */
-public class QuietStreamsConfig extends StreamsConfig {
```
mjsax, line 29: Moved this class into `ClientUtils.java`

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -275,31 +259,70 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
                               final RecordConverter recordConverter) {
         for (final TopicPartition topicPartition : topicPartitions) {
             globalConsumer.assign(Collections.singletonList(topicPartition));
+            long offset;
             final Long checkpoint = checkpointFileCache.get(topicPartition);
             if (checkpoint != null) {
                 globalConsumer.seek(topicPartition, checkpoint);
+                offset = checkpoint;
```
mjsax, line 259: Some side improvement: if we `seek` base on the checkpoint, there is no reason to call `position()` because we know what our offset is. -- Only if we `seekToBeginning()` we need to get the current offset from the consumer itself (instead of `position()` we could also call `beginningOffsets` but `position` it the easer to use API.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -275,31 +259,70 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
...
             stateRestoreListener.onRestoreStart(topicPartition, storeName, offset, highWatermark);
             long restoreCount = 0L;
 
+            long deadlineMs = NO_DEADLINE;
             while (offset < highWatermark) {
                 try {
                     final ConsumerRecords<byte[], byte[]> records = globalConsumer.poll(pollTime);
+                    if (records.isEmpty()) {
+                        if (taskTimeoutMs == 0L) {
+                            deadlineMs = maybeUpdateDeadlineOrThrow(
```
mjsax: For `poll()`, even if retrying is disabled, we need to retry to give a fetch request (with default request timeout of 30 seconds but only a default `pollTime` of 100ms) a fair change to actually return.

At least I believe this makes sense. Also \cc @guozhangwang

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: We don't need to call `retryUntilSuccessOrThrowOnTaskTimeout` now, so we actually get cleaner code :)

docs/streams/developer-guide/config-streams.html
```diff
@@ -326,13 +321,18 @@ <h4><a class="toc-backref" href="#id5">bootstrap.servers</a><a class="headerlink
...
+            <td>Medium</td>
```
abbccdda, line 332: nit: for a doc clean-up, it is helpful to include a screenshot of updated paragraph. The changes starting at L331 should be suffice.

streams/src/main/java/org/apache/kafka/streams/internals/QuietStreamsConfig.java
abbccdda: What's the purpose for this move?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -318,6 +341,72 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
...
 
+    private void retryUntilSuccessOrThrowOnTaskTimeout(final Runnable runnable,
+                                                       final String errorMessage) {
+        long deadlineMs = NO_DEADLINE;
+
+        do {
+            try {
+                runnable.run();
+                return;
+            } catch (final TimeoutException retryableException) {
```
abbccdda: `retriableException` to be consistent with the defined exception type in AK.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -318,6 +341,72 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
...
+                            "Retrying is disabled. You can enable it by setting `%s` to a value larger than zero.",
+                            StreamsConfig.TASK_TIMEOUT_MS_CONFIG
+                        ),
+                        retryableException
+                    );
+                }
+
+                deadlineMs = maybeUpdateDeadlineOrThrow(deadlineMs);
+
+                log.debug(errorMessage, retryableException);
```
abbccdda: Should we log warning here instead? At least this is a timeout.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
abbccdda: If we do have a valid position from previous `position` call, do we still need to update the position here again?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -275,31 +259,70 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
...
-                    offset = globalConsumer.position(topicPartition);
+                    try {
+                        offset = globalConsumer.position(topicPartition);
+                    } catch (final TimeoutException error) {
+                        // the `globalConsumer.position()` call should never block, because we know that we did
+                        // a successful `position()` call above for the requested partition and thus the consumer
+                        // should have a valid local position that it can return immediately
+
+                        // hence, a `TimeoutException` indicates a bug and thus we rethrow it as fatal `IllegalStateException`
+                        throw new IllegalStateException(error);
```
abbccdda: We should add some explanation in the illegal state exception for why such timeout is fatal

streams/src/test/java/org/apache/kafka/test/NoOpReadOnlyStore.java
```diff
@@ -78,6 +78,7 @@ public void init(final ProcessorContext context, final StateStore root) {
             new File(context.stateDir() + File.separator + name).mkdir();
         }
         this.initialized = true;
+        context.register(root, (k, v) -> { });
```
abbccdda, line 81: Why do we need this?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -275,31 +259,70 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
...
         for (final TopicPartition topicPartition : topicPartitions) {
             globalConsumer.assign(Collections.singletonList(topicPartition));
+            long offset;
             final Long checkpoint = checkpointFileCache.get(topicPartition);
             if (checkpoint != null) {
                 globalConsumer.seek(topicPartition, checkpoint);
+                offset = checkpoint;
             } else {
                 globalConsumer.seekToBeginning(Collections.singletonList(topicPartition));
+                final AtomicLong position = new AtomicLong();
```
abbccdda: For the above call, was curious why we couldn't seek for all the topic partitions that are missing positions here, instead of doing one by one look-up?

abbccdda (commented):
Thanks for the PR, got a couple of comments.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: Man, this is confusing. I think I see what you're getting at, but it seems pretty strange to have to go to all that trouble to extract the consumer config so that we can apply a shorter timeout on each poll, but then loop around until the originally configured client timeout passes.

Can you explain how the outcome is different than just calling `globalConsumer.poll(requestTimeoutMs)`?

It also seems strange to extract a consumer timeout configuration that specifically does not apply to `poll` and apply it to `poll`. This seems like it would violate users' expectations when they set that configuration value.

Why wouldn't we instead apply the non-progress timeout (`taskTimeoutMs`), since it seems like that's exactly what it's for? I.e., it seems like `globalConsumer.poll(pollTime + taskTimeoutMs)` might be the most appropriate choice here?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: It seems like what we're doing here is syncing our `offset` variable with the internal position cursor of the client. Although @mjsax is claiming the client never needs to update its position from the consumer group, the position would advance as a consequence of polling records earlier in this loop. Right?

vvcephei: Hey @mjsax , it sounds like you have an opinion about what would be unexpected internal behavior of the client, based on your knowledge of how it is currently implemented. There doesn't seem to be anything in the contract of the client that says we should rely on this implementation detail, so it seems inappropriate to build in a fatal exception for this case.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -185,32 +200,16 @@ public void registerStore(final StateStore store, final StateRestoreCallback sta
...
-                    retryBackoffMs,
-                    attempts,
-                    retries,
-                    retryableException);
-                Utils.sleep(retryBackoffMs);
-            }
-        }
+        final Map<TopicPartition, Long> highWatermarks = new HashMap<>();
+        retryUntilSuccessOrThrowOnTaskTimeout(
+            () -> highWatermarks.putAll(globalConsumer.endOffsets(topicPartitions)),
```
vvcephei: No need to change anything, just a note: It's mildly concerning to see side-effecting operations inside a retry loop, since there's no guarantee that failed attempts won't leave any garbage in the system, for example leaving a couple of things in highWatermarks on each attempt. I've read through the code, and I don't think this could actually happen right now, but it could after refactoring.

Since we're really just retrying simple API calls, something like this would be safer:
```java
    private <R> R retryUntilSuccessOrThrowOnTaskTimeout(final Supplier<R> supplier,
                                                        final String errorMessage) {
        long deadlineMs = NO_DEADLINE;

        do {
            try {
                return supplier.get();
            } catch (final TimeoutException retryableException) {
                if (taskTimeoutMs == 0L) {
                    throw new StreamsException(
                        String.format(
                            "Retrying is disabled. You can enable it by setting `%s` to a value larger than zero.",
                            StreamsConfig.TASK_TIMEOUT_MS_CONFIG
                        ),
                        retryableException
                    );
                }

                deadlineMs = maybeUpdateDeadlineOrThrow(deadlineMs);

                log.debug(errorMessage, retryableException);
            }
        } while (true);
    }
```

Which would support:
```java
        final Map<TopicPartition, Long> highWatermarks = new HashMap<>(
            retryUntilSuccessOrThrowOnTaskTimeout(
                () -> globalConsumer.endOffsets(topicPartitions),
                String.format(
                    "Failed to get offsets for partitions %s. The broker may be transiently unavailable at the moment.",
                    topicPartitions
                )
            )
        );
```

Incidentally, this seems to be exactly the same as:
```java

        final Map<TopicPartition, Long> highWatermarks = new HashMap<>(
            globalConsumer.endOffsets(topicPartitions, Duration.ofMillis(taskTimeoutMs))
        );
```

right? That might be preferable because we're not implementing any backoff, whereas retrying internally in the client, it can do exponential backoff, etc. (I think).

streams/src/main/java/org/apache/kafka/streams/internals/QuietStreamsConfig.java
vvcephei: I guessing because there are other "quiet" configs there. On the other hand, the "quiet *Streams* config" isn't really a "*client* util"...

vvcephei (commented):
Thanks for the update, @mjsax ! I had some trouble wrapping my mind around the retry stuff. Sorry :/

streams/src/main/java/org/apache/kafka/streams/internals/QuietStreamsConfig.java
mjsax: Well, `KafkaStreams` is the "streams client" IMHO :)

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: Correct, we rely on the clients internal backoff behavior here. Happy to switch to a `Supplier`

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: Well, I guess we want to restore partition by partition because the "restore listener" callback is easier to handle this way?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: > Can you explain how the outcome is different than just calling globalConsumer.poll(requestTimeoutMs)?

I did consider it, but was not 100% sure if this might be better. Was also concerned about "miss using" request timeout (as you already mentioned) as we have `StreamsConfig.POLL_TIMEOUT_MS`.

I like the proposal to use `pollTime + taskTimeoutMs` though!

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: What John says: We need to track the restore progress to eventually break the outer while loop. We don't use `offset` to `seek()`, so it's not about updating the consumer's position but knowing when we are done.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: As this would expose a bug, is does not seem useful to users -- and for us, we have the comment in the code.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: Not sure if I can follow?

streams/src/test/java/org/apache/kafka/test/NoOpReadOnlyStore.java
mjsax: That is what the real code does, too. Otherwise we need to call `register` manually in the tests what is annoying.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: Actually, we need to use all three: `pollTime + requestTimeout + taskTimeoutMs`

If we don't include `requestTimeout` and `taskTimeoutMs` is set to `0` (or a smaller value than `requestTimeout` in general), the first `poll()` would return no data with very high probability and we would fail what seems not desirable.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: As requested by John, I'll drop this and just call `retryUntilSuccessOrThrowOnTaskTimeout` here, too.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: I guess with the new `retryUntilSuccessOrThrowOnTaskTimeout` the code is still clean is we just use it here, too.

mjsax:
Thanks for the review @abbccdda @vvcephei -- updated the PR.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: Btw: just for completeness: we will still use `pollTime` unmodified in `globalConsumer.poll(...)` during regular processing, so the config has still value. Only in the startup bootstrapping phase, we would use the sum of poll/request/task-timeout to not fail unnecessarily.

Also note that during normal processing, we don't apply `taskTimeoutMs` atm. Not sure if we should, but I would not want to include it in this PR. -- It won't block the "main" processing if the global `poll()` loop would not make progress because the `StreamThreads` are already running, and thus it's impact is different -- also, during normal processing the global `highWatermark` may move and thus we need a different strategy anyway compared to the bootstrap phase.

vvcephei: Ah, that last point sounds good.

I'm not opposed, but it still seems strange to grab a random timeout from the client config and apply it to the `poll` method. It kind of makes me wonder what we're really trying to do here. If we want to give poll at least 30 seconds to return data, then we can just give it at least 30 seconds, right? No reason to abuse the client configuration.

On the other hand, the pollTime may not even appropriate for the global thread, right? It seems more like it was meant for StreamThread to make sure we don't block too long and violate the max poll interval. But since the global thread is only assigned and not subscribed, it has no constraint on how long it should block, except for the taskTimeoutMs config, right?

Clearly, we need to set some kind of lower bound on it, though, but it's not clear that it needs to be configurable. Anyway, just food for thought. Feel free to keep the requestTimeout if you really think it's appropriate.

streams/src/main/java/org/apache/kafka/streams/internals/QuietStreamsConfig.java
vvcephei: Haha, that's kind of a stretch, but it's not a big deal :)

vvcephei:
I left one more comment about using the request timeout config, but I'm really fine with whatever you decide. LGTM! Thanks again.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: Well, the requestTimeout is no really "random" -- if we send a fetch request, it's the maximum time until the producer would abort the request and retry. -- _If_ poll() would throw a TimeoutException, this would be the point when we get a timeout if there are no retries. -- To be fair, maybe using the consumer's _default_api_timeout_ms_ config might be better than using requestTimeout though? -- The point being is, that picking a consumer config seems to be the best thing we can do for this case instead of a hard-coded number (or introducing a new config for this specific corner case).

> pollTime may not even appropriate for the global thread

I think it does (even if differently). At least for regular processing: (1) we interleave polling and flushing, and also when we stop the client, we want it to responsive and not being blocked in poll() for 5 minutes (taskTimeoutMs default) for this case. (What of course implies that one cannot really "intercept" the startup phase atm -- something we might want to address, but maybe not in this PR.)

> Clearly, we need to set some kind of lower bound on it, though, but it's not clear that it needs to be configurable.

I think the worst we can do is to make it not configurable. :)

streams/src/main/java/org/apache/kafka/streams/internals/QuietStreamsConfig.java
mjsax: Is it? KIP-28 suggest to add Kafka Streams -- a processor client :D

https://cwiki.apache.org/confluence/display/KAFKA/KIP-28+-+Add+a+processor+client

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: Thanks, @mjsax , that's fair.

I promise I'm not equivocating here; I'm just trying to figure out what my intuition is trying to tell me.

It seems like maybe the fundamental problem here is that we can't distinguish among a successful poll that returns no data, a failure to poll, and a pending async fetch request. The one thing we know is that the end offset is beyond our current position, so there _should_ be data to poll, so we can assume that an empty return means either that the fetch failed internally or it hasn't completed yet.

Stepping back, this seems to be related to the problem of task idling, in which it's pointless to "idle" for a time so short that we have no chance to actually get a response back from the broker.

I feel like this is substantially my fault from #4855 / KIP-266. The purpose of making this API completely async was to avoid harming liveness in situations where we might have a relatively strict deadline. But that's not the case here.

I guess the "poor man's" solution we're going for here is to block poll for at least long enough to allow for a complete fetch round-trip from the broker. If we know that there was a round-trip, and we didn't get any data, then we can conclude that there was an error (since we know there is data to get). Since we can't know that there was a round trip, we weaken the condition to: if we know it's been long enough that there should have been a round-trip and we don't get data, we conclude there was probably an error.

In your KIP, we specified we would start the task timer _after_ the first error, so it seems like we really want to just block the poll for the round-trip time, and then apply your "update deadline, etc." function. I'm with you now that to get the round-trip time, we have to extract some config(s) from the Consumer. This is a pretty awkward hack, but now that I've thought it through, it seems the best we can do. Maybe we can mull it over and file an improvement jira for the Consumer to improve use cases like this.

Anyway, it seems like the "poll time" config is irrelevant, we just need to know what config to grab that corresponds to completing a fetch request with high probability. It seems like we shouldn't need to update metadata, so we would send a fetch request on the first poll call, and we just need to block for whatever time bounds the fetch response time.

I'm honestly not sure what timeout would be best here. It looks like the ConsumerNetworkClient will just wait for a response until it gets a "disconnect" (L598). Is that a socket timeout? I'm not sure.

mjsax: I agree. The "issue" is really that `poll()` does _not_ throw a `TimeoutException`... Also, because we do manual assignment, `poll()` would never return "early" as it never need to wait for joining a consumer group. -- However, compare to `max.task.idle.ms`, we are in a better situation here, because we poll() for only a single partition at a time.

I also agree, that applying `task.timeout.ms` should start _after_ we got a first timeout -- this was how the original code worked that you criticized as:
> Man, this is confusing.

And I agree, that the code was not straightforward to understand. But if we think it's the right thing to do, I am also happy to add it back :)

I am also not an expert on all consumer internals, but from my understanding, fetch requests are send async in general, and if a fetch request fails, the consumer would actually not retry it but a retry would be triggered by the next `poll()` call. If there is no data available (ie, fetch request did not return yet) when `poll()` is called, the consumer would block internally until `poll(Duration)` timeout expires or until a fetch request returns (whatever comes first).

Furthermore, before `poll()` returns, it always check if a fetch request is in-flight or not, and sends one if not.

Thus, on the verify first call to `poll()` we know that no fetch request can be in-flight and we also know that `poll()` would send one, and block until it returns or `poll(Duration)` expires. Thus, if `poll()` does not block for at least `request.timeout.ms`, and we get an empty result back we don't know which case holds, however, if we use the request timeout, it seems that we _know_ if the fetch was successful or did time out? We also know, that a fetch request will be inflight after `poll()` returns. Thus, for any consecutive `poll()` applying request timeout also ensures that we know if the request was successful or not.

I guess the only difference to what I just described to my original code was, that I uses `pollTime + requestTimeout`.

Bottom line: I am not 100% sure what you propose? Should we go with the original design? Or with the new design? -- In the end, I think we don't need a follow up PR, and we can just try to get it right in this PR. I don't see any benefit in splitting it up into 2 PRs (because, as mentioned above, we fetch for a single partitions and thus it's a different case compared to `max.task.idle.ms` scenario).

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -292,21 +278,36 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
 
             while (offset < highWatermark) {
                 try {
-                    final ConsumerRecords<byte[], byte[]> records = globalConsumer.poll(pollTime);
+                    final ConsumerRecords<byte[], byte[]> records =
+                        globalConsumer.poll(pollTimePlusRequestTimeoutPlusTaskTimeout);
```
hachikuji: It does seem a bit weird here to add in the request timeout. Not sure I follow the reasoning behind that...

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: Ok, thanks @mjsax . I just traced though the consumer code again, and have finally been able to see what you already knew: that `request.timeout.ms` is indeed the correct amount of time to wait.

Namely, we send a fetch here: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1285

Which calls through to client.send here: https://github.com/apache/kafka/blob/659ca8f089c6b1c3a53036a8f96de7f763971fd1/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L263

Which fills in the `request.timeout.ms` config value here: https://github.com/apache/kafka/blob/659ca8f089c6b1c3a53036a8f96de7f763971fd1/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerNetworkClient.java#L106

Which uses it to construct a ClientRequest here: https://github.com/apache/kafka/blob/659ca8f089c6b1c3a53036a8f96de7f763971fd1/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerNetworkClient.java#L129-L130

Which then gets used to create an InFlightRequest when it gets sent here: https://github.com/apache/kafka/blob/659ca8f089c6b1c3a53036a8f96de7f763971fd1/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L1248

Which is later used to detect expired requests here: https://github.com/apache/kafka/blob/659ca8f089c6b1c3a53036a8f96de7f763971fd1/clients/src/main/java/org/apache/kafka/clients/InFlightRequests.java#L162

Which is used to list nodes (brokers) for which there is an expired request here: https://github.com/apache/kafka/blob/659ca8f089c6b1c3a53036a8f96de7f763971fd1/clients/src/main/java/org/apache/kafka/clients/InFlightRequests.java#L179

Which is then processed as a "disconnection" here: https://github.com/apache/kafka/blob/659ca8f089c6b1c3a53036a8f96de7f763971fd1/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L803

It also looks like the KafkaClient just does a tight-loop checking for a network response, so we don't really need any extra time to account for sampling errors. Also, it still seems like using the sum as the poll duration is just as good as using your retry logic, so I think the duration parameter is fine.

My only remaining question, which maybe doesn't really matter one way or another, is whether `poll.ms` really belongs here or not. It seems like the desired semantics are accomplished by just waiting `request.timeout.ms` for the initial failure, and then an extra `task.timeout.ms` for any retries.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
vvcephei: Unfortunately, Github ate the very extensive thread about this: https://github.com/apache/kafka/pull/9047/files/0e07109bb0cfe87c85e76fbab3b50e9274300388#r460449901

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: Thanks for digging into it @vvcephei -- The question about `pollTimeout` is fair. I guess for default configs it might not matter much, as the default is 100ms IIRC.

At the same time, we apply the same `pollTimeout` during regular processing, and as a matter of fact, for this case, a use might want to use a longer poll-timeout, as otherwise, the thread would just "busy wait" anyway (only the responsiveness for a shutdown of the app should be considered).

Thus, it might actually make sense to exclude `pollTimeout` completely and only use `requestTimeout + taskTimeout`. Again, using `taskTimeout` in `poll()` reduces the responsiveness of a shutdown -- however, atm during bootstrapping we ignore a shutdown signal anyway, hence, for now we don't make the situation worse. I create a ticket to fix this: https://issues.apache.org/jira/browse/KAFKA-10317 and will add a comment for now.

Short related note: actually using `requestTimeout` seems to be a conservative upper bound for poll(). A request could fail with a different error before `requestTimeout` hits and would be retried internally for this case -- if this happens, we might want to start the `taskTimeout` earlier. However, we don't have any means atm to detect this case. Thus, using `requestTimeout` is the best option we have right now (because triggering `taskTimeout` too early seems to be worse than triggering it too late). I created a ticket though that might allow us to improve the code later: https://issues.apache.org/jira/browse/KAFKA-10315

mjsax:
@vvcephei @abbccdda @guozhangwang @hachikuji -- I updated this PR according to your discussions. Needed to squash for rebasing to resolve conflicts.

docs/streams/developer-guide/config-streams.html
abbccdda: @mjsax could we do a screenshot to make sure it looks good on the web-page?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -274,30 +252,74 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
...
+                //   -> do a more precise timeout handling if `poll` would throw an exception if a fetch request fails
+                //      (instead of letting the consumer retry fetch requests silently)
+                //
+                // TODO https://issues.apache.org/jira/browse/KAFKA-10317 and
+                //      https://issues.apache.org/jira/browse/KAFKA-7380
+                //  -> don't pass in `task.timeout.ms` to stay responsive if `KafkaStreams#close` gets called
+                final ConsumerRecords<byte[], byte[]> records = globalConsumer.poll(requestTimeoutPlusTaskTimeout);
+                if (records.isEmpty()) {
+                    // this will always throw
+                    maybeUpdateDeadlineOrThrow(time.milliseconds());
```
abbccdda, line 305: Could we just throw here?

streams/src/test/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImplTest.java
```diff
@@ -671,19 +1211,21 @@ private void writeCorruptCheckpoint() throws IOException {
...
+    private void initializeConsumer(final long numRecords, final long startOffset, final TopicPartition... topicPartitions) {
+        consumer.assign(Arrays.asList(topicPartitions));
+
         final HashMap<TopicPartition, Long> startOffsets = new HashMap<>();
-        startOffsets.put(topicPartition, startOffset);
         final HashMap<TopicPartition, Long> endOffsets = new HashMap<>();
-        endOffsets.put(topicPartition, startOffset + numRecords);
-        consumer.updatePartitions(topicPartition.topic(), Collections.singletonList(new PartitionInfo(topicPartition.topic(), topicPartition.partition(), null, null, null)));
-        consumer.assign(Collections.singletonList(topicPartition));
+        for (final TopicPartition topicPartition : topicPartitions) {
```
abbccdda, line 1219: nit: we could just use `Map` for `startOffsets` and `endOffsets`

abbccdda (commented):
Thanks for the update. Just a couple of more comments, btw, it would be good to add test for IOException cases, such as:
```
  try {
            checkpointFile.write(filteredOffsets);
            log.warn("Failed to write offset checkpoint file to {} for global stores: {}", checkpointFile, e);
        }
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
mjsax: We could, but this implies redundant code to "assemble" the error message, and I prefer to reuse the existing code for it.

mjsax:
> btw, it would be good to add test for IOException cases, such as

Seems out of scope for this PR. We should rather to a new PR. I created https://issues.apache.org/jira/browse/KAFKA-10351

docs/streams/developer-guide/config-streams.html
mjsax: Done.

@guozhangwang @abbccdda -- I am just wondering if we should mark to parameter with priority "Low" though?

abbccdda (approved):
LGTM for the added screenshot and test coverage

docs/streams/developer-guide/config-streams.html
guozhangwang: "medium" looks fine to me. But I'm not feeling strong against it either.

guozhangwang (approved):
LGTM!

PR #9050, opened 2020-07-22
KAFKA-10193: Add preemption for controller events that have callbacks

jeffkbkim:
JIRA: https://issues.apache.org/jira/browse/KAFKA-10193
* add `preempt(): Unit` method for all `ControllerEvent` so that all events (and future events) must implement it
* for events that have callbacks, move the preemption from individual methods to `preempt()`
* add preemption for `ApiPartitionReassignment` and `ListPartitionReassignments`
* add integration tests:
1. test whether `preempt()` is called when controller shuts down
2. test whether the events with callbacks have the correct error response (`NOT_CONTROLLER`)
* explicit typing for `ControllerEvent` methods


core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
```diff
@@ -598,6 +603,86 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
...
+      override def preempt(): Unit = count.decrementAndGet()
+    }
+
+    controller.eventManager.put(processedEvent)
+    controller.eventManager.put(preemptedEvent)
+    controller.eventManager.put(preemptedEvent)
+
+    doAnswer((_: InvocationOnMock) => {
+      latch.countDown()
+    }).doCallRealMethod().when(spyThread).awaitShutdown()
```
jsancio, line 628: I see that you made the thread variable a `var` to be able to do this. Would this work instead?

```scala
controller.eventManager.put(processedEvent)
controller.eventManager.put(preemptedEvent)
controller.eventManager.put(preemptedEvent)

controller.shutdown()
latch.countDown()
TestUtils.waitUntilTrue(() => { 
  !controller.eventManager.thread.isAlive
}, "...")

assertEquals(0, count.get())
```

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
```diff
@@ -598,6 +603,86 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
...
+    verify(spyThread).awaitShutdown()
+  }
+
+  @Test
+  def testPreemptionWithCallbacks(): Unit = {
+    servers = makeServers(1, enableControlledShutdown = false)
+    val controller = getController().kafkaController
+    val latch = new CountDownLatch(1)
+    val spyThread = spy(controller.eventManager.thread)
+    controller.eventManager.setControllerEventThread(spyThread)
```
jsancio: See my other comment about possibly removing this method.

jsancio (commented):
Thanks for the PR @jeffkbkim. Looks good in general just some comments on the changes to the controller thread variable and the addition of the `setControllerEventThread`.

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
jeffkbkim: @jsancio In the suggested code, `latch.await()` blocks the event thread from shutting down while `thread.awaitShutdown()` (called in `controller.shutdown()`) blocks the main thread from reaching `latch.countDown()` causing a deadlock. The fix is to call `latch.countDown()` before `thread.awaitShutdown()` which has been implemented using `spy`.

Please let me know if there's another way to keep the `thread` field as a `val`.

stanislavkozlovski: Could we have the test call `latch.countDown()` in a background thread with a delay, right before we call controller.shutdown? I guess it's possible to have race conditions with that solution

stanislavkozlovski:
> add preempt(): Unit method for all ControllerEvent so that all events (and future events) must implement it
> for events that have callbacks, move the preemption from individual methods to preempt()
> add preemption for ApiPartitionReassignmentand ListPartitionReassignments

Great idea

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
jeffkbkim: @stanislavkozlovski thanks for the comment. i've tried this approach before and the test passes but sometimes output:
```
[2020-07-23 11:12:43,316] ERROR [RequestSendThread controllerId=0] Controller 0 fails to send a request to broker localhost:51542 (id: 0 rack: null) (kafka.controller.RequestSendThread:76)
java.lang.InterruptedException
```
i also see a bit of flakiness with this approach as we cannot exactly time when `latch.countDown()` is called.

jsancio: Got it. Sounds good.

core/src/main/scala/kafka/controller/ControllerEventManager.scala
```diff
@@ -77,7 +77,7 @@ class ControllerEventManager(controllerId: Int,
   private val putLock = new ReentrantLock()
   private val queue = new LinkedBlockingQueue[QueuedEvent]
   // Visible for test
-  private[controller] val thread = new ControllerEventThread(ControllerEventThreadName)
+  private[controller] var thread = new ControllerEventThread(ControllerEventThreadName)
```
jsancio, line 81: This comment applies to `clearAndPut`:

```scala
  def clearAndPut(event: ControllerEvent): QueuedEvent = inLock(putLock) {
    queue.forEach(_.preempt(processor))
    queue.clear()
    put(event)
  }
```
I think there is a bug here where at most one event will be process twice. Once by `_preempt(processor)` and once by `doWork`. I think we can fix this concurrency bug if we use `LinkedBlockingQueue::drainTo`. E.g.

```scala
  def clearAndPut(event: ControllerEvent): QueuedEvent = {
    val preemptedEvents = ...;

    inLock(putLock) {
      queue.drainTo(preemptedEvents)
      put(event)
    }

    preemtedEvents.forEach(_.preempt(processor))
  }
```

core/src/main/scala/kafka/controller/KafkaController.scala
```diff
@@ -1986,101 +1965,125 @@ private[controller] class ControllerStats extends KafkaMetricsGroup {
 
 sealed trait ControllerEvent {
   def state: ControllerState
+  def preempt(): Unit
```
jsancio, line 1969: I would add a documentation comment to this method explaining that this method will not be executed by the controller thread but instead it will be executed by some other thread.

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
```diff
@@ -598,6 +603,86 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
...
+    val processedEvent = new MockEvent(ControllerState.TopicChange) {
+      override def process(): Unit = latch.await()
+      override def preempt(): Unit = {}
+    }
+    val preemptedEvent = new MockEvent(ControllerState.TopicChange) {
+      override def process(): Unit = {}
+      override def preempt(): Unit = count.decrementAndGet()
```
jsancio: This method should be executed by the thread that is running this test. If you agree, no need to use an `AtomicInteger`.

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
jsancio: The important part is that we want to do `latch.countDown` after `initiateShutdown` has been called so that the controller thread doesn't pick up a new event because `isRunning` is `false`.

jsancio (commented):
@jeffkbkim, I left a few more comments.

core/src/main/scala/kafka/controller/KafkaController.scala
```diff
@@ -1986,101 +1965,125 @@ private[controller] class ControllerStats extends KafkaMetricsGroup {
...
 }
 
 case object ControllerChange extends ControllerEvent {
-  override def state = ControllerState.ControllerChange
+  override def state: ControllerState = ControllerState.ControllerChange
+  override def preempt(): Unit = {}
```
mumrah, line 1974: Can we define a default implementation on the trait rather than overriding all these with empty functions?

core/src/main/scala/kafka/controller/KafkaController.scala
```diff
@@ -2090,6 +2093,12 @@ case class ReplicaLeaderElection(
   callback: ElectLeadersCallback = _ => {}
 ) extends ControllerEvent {
   override def state: ControllerState = ControllerState.ManualLeaderBalance
+
+  override def preempt(): Unit = callback(
+    partitionsFromAdminClientOpt.fold(Map.empty[TopicPartition, Either[ApiError, Int]]) { partitions =>
```
mumrah, line 2099: nit: I think you can use `foreach` instead of fold + default value

core/src/main/scala/kafka/controller/KafkaController.scala
jeffkbkim: @mumrah one of the issues was that people are able to create controller events without implementing preemption. In most cases, this is fine because most events don't have callbacks. But for events that do have callbacks, someone can forget to implement preemption and this won't be caught during compile time. This is to force an implementation of preemption which can be empty when designing a new event.

core/src/main/scala/kafka/controller/KafkaController.scala
jeffkbkim: @mumrah i believe the initial `Map.empty[TopicPartition, Either[ApiError, Int]]` was used to structure the return type (`ElectLeadersCallback: Map[TopicPartition, Either[ApiError, Int]] => Unit`) which cannot be achieved using `foreach`.

core/src/main/scala/kafka/controller/ControllerEventManager.scala
jeffkbkim: @jsancio updated the PR. i wrapped the entire method in a lock because the return type `QueuedEvent` can be retrieved by `put(event)`. the suggested code would require another variable to be initialized, then set inside the lock, then returned at the end. I thought it would be best to prefer simplicity and readability over performance as this only occurs in the start and end of `ControllerEventManager`'s lifecycle.

core/src/main/scala/kafka/controller/ControllerEventManager.scala
```diff
@@ -140,6 +143,9 @@ class ControllerEventManager(controllerId: Int,
     }
   }
 
+  // for testing
+  private[controller] def setControllerEventThread(thread: ControllerEventThread): Unit = this.thread = thread
```
jsancio: We can remove this since we have `private[controller] var thread` with the same visibility.

core/src/main/scala/kafka/controller/KafkaController.scala
jsancio: Yeah. The value returned by the `fold` is passed to `callback`. `foreach` would return `Unit`.

jsancio (approved):
LGTM. Minor nits.

Thanks for the PR @jeffkbkim !

omkreddy:
ok to test

omkreddy:
retest this please

omkreddy:
ok to test

core/src/main/scala/kafka/controller/KafkaController.scala
omkreddy: @jeffkbkim I think, we can add default implementation on the trait and add a comment saying that to implement preempt() method for events which have callbacks.  I hope, PR reviewers will catch if someone forget to implement preemption methods in future

omkreddy (approved):
@jeffkbkim Thanks for the PR. LGTM. left minor comment

PR #9052, opened 2020-07-22
MINOR: TopologyTestDriver should not require dummy parameters

vvcephei:
TopologyTestDriver comes with a paper cut that it passes through a
config requirement that application.id and bootstrap.servers must be
configured. But these configs are _not_ required in the context of
TopologyTestDriver specifically. This change relaxes the requirement.

### DIFF
--- a/streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
+++ b/streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
@@ -413,6 +413,12 @@ public class TopologyTestDriverTest {
     }
 
     @Test
+    public void shouldNotRequireParameters() {
+        new TopologyTestDriver(setupSingleProcessorTopology(), new Properties());
+        // should not throw exceptions
+    }
+
+    @Test
     public void shouldInitProcessor() {
         testDriver = new TopologyTestDriver(setupSingleProcessorTopology(), config);
         assertTrue(mockProcessors.get(0).initialized);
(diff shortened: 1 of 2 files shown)


abbccdda (approved):
LGTM

streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
```diff
@@ -412,6 +412,12 @@ private Topology setupTopologyWithInternalTopic(final String firstTableName,
         return builder.build(config);
     }
 
+    @Test
+    public void shouldNotRequireParameters() {
+        new TopologyTestDriver(setupSingleProcessorTopology(), new Properties());
+        // should not throw exceptions
```
mjsax, line 418: Don't think we need this comment.

mjsax (commented):
Should be back this up with a Jira because it seems to be a nice improvement and thus it would be easier to discover (also would end up in the release notes that way)?

Btw: can you double check the docs and update them accordingly? IIRC we do have examples that explain how the TTD works.

Last, should we update our own tests right away? I assume we often specify the dummy values in our test code and it might be best to "clean it up" with this PR.

mjsax:
Only flaky tests failed. I updated the corresponding Jira tickets.

vvcephei:
Thanks @mjsax . Your suggestions sound good to me. I'll circle back on this PR in a bit.

abbccdda:
retest this please

mjsax:
@vvcephei -- Random thought: with this change, we still require people to pass in `Properties` object. As it can be empty, I am wondering if we should do a small KIP and add a constructor that does not require to pass in a `Properties` object?

vvcephei:
Closing in favor of https://github.com/apache/kafka/pull/9477, since I've recycled my fork since opening this PR and can no longer update the PR.

PR #8913, opened 2020-06-23
KAFKA-10191 fix flaky StreamsOptimizedTest

chia7712:
We have to call ```KafkaStreams#cleanUp``` to reset local state before starting application up the second run.

issue: https://issues.apache.org/jira/browse/KAFKA-10191

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9053, opened 2020-07-22
MINOR: Revert KAFKA-10191 fix flaky StreamsOptimizedTest (#8913)

vvcephei:
This PR depends on features of the reset tool that are not present in 2.5

### DIFF
--- a/core/src/main/scala/kafka/tools/StreamsResetter.java
+++ b/core/src/main/scala/kafka/tools/StreamsResetter.java
@@ -162,6 +162,7 @@ public class StreamsResetter {
             consumerConfig.putAll(properties);
             exitCode = maybeResetInputAndSeekToEndIntermediateTopicOffsets(consumerConfig, dryRun);
             maybeDeleteInternalTopics(adminClient, dryRun);
+
         } catch (final Throwable e) {
             exitCode = EXIT_CODE_ERROR;
             System.err.println("ERROR: " + e);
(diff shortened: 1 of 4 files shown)


vvcephei:
Hey @abbccdda @ableegoldman @chia7712 ,

You were all involved in the original PR #8913 , so I figured you should be aware of this and have the opportunity to comment.

This PR is incompatible with 2.5. When the system test runs, it always fails with `force is not a recognized option`

I tried removing it, but then I just get `java.lang.IllegalStateException: Consumer group 'StreamsOptimizedTest' is still active and has following members`.

I'm thinking we should just revert this commit to 2.5 to unblock the release, and then @chia7712 can send a new PR targeted to 2.5 if he wants to.

Thoughts?

chia7712:
> I'm thinking we should just revert this commit to 2.5 to unblock the release, and then @chia7712 can send a new PR targeted to 2.5 if he wants to.

+1 and sorry to use the nonexistent option in 2.5 :(

vvcephei:
No worries, it wasn't your fault @chia7712 . It was mine, since I cherry-picked it and didn't run the test again before pushing.

abbccdda (approved):
LGTM, thanks for the fix! @vvcephei

chia7712 (approved):

chia7712:
It seems we have to set lower session timeout and then wait for group expiration if the ```--force``` is not backport to 2.5. Or we can fix the system test by calling ```cleanup``` only. Not sure whether this (simple) workaround is good enough for branch 2.5?

ableegoldman (approved):
Sounds good, this test wasn't actively failing in 2.5 (although it wasn't technically doing the right thing either) so we may as well revert it for the release and then fix it after

vvcephei:
Thanks, all. I'll just revert (it's a cleaner way to unblock the release), and if the test becomes flaky, then we'll deal with that separately.

vvcephei:
Tests passed. Merging.

PR #9054, opened 2020-07-22
KAFKA-10282: Remove Log metrics immediately when deleting log

bob-barrett:
Currently, we remove the Log metrics when asynchronous deletion of the log is triggered. However, we attempt to register the metrics immediately upon log creation. If a Log object is re-created for a partition that is pending deletion (because a topic was quickly re-created or because a partition was moved off and back onto a broker), the registration of the new metrics can happen before the asyncrhonous deletion. In this case, the metrics are removed after the second registration, leading to missing Log metrics.

To fix this, this patch changes the log deletion behavior to remove the metrics when the log is first marked for deletion, rather than when the files are deleted. This removes the window in which metrics registration can occur before metrics removal. This is justifiable because the log should be logically deleted when a delete request or partition movement finishes, rather than when the files are actually removed. Tested with unit tests.

### DIFF
--- a/core/src/main/scala/kafka/log/LogManager.scala
+++ b/core/src/main/scala/kafka/log/LogManager.scala
@@ -199,27 +199,22 @@ class LogManager(logDirs: Seq[File],
       if (cleaner != null)
         cleaner.handleLogDirFailure(dir)
 
-      val offlineCurrentTopicPartitions = currentLogs.collect {
-        case (tp, log) if log.parentDir == dir => tp
-      }
-      offlineCurrentTopicPartitions.foreach { topicPartition => {
-        val removedLog = currentLogs.remove(topicPartition)
-        if (removedLog != null) {
-          removedLog.closeHandlers()
-          removedLog.removeLogMetrics()
+      def removeOfflineLogs(logs: Pool[TopicPartition, Log]): Iterable[TopicPartition] = {
+        val offlineTopicPartitions: Iterable[TopicPartition] = logs.collect {
+          case (tp, log) if log.parentDir == dir => tp
         }
-      }}
+        offlineTopicPartitions.foreach { topicPartition => {
+          val removedLog = removeLogAndMetrics(logs, topicPartition)
+          if (removedLog != null) {
+            removedLog.closeHandlers()
+          }
+        }}
 
-      val offlineFutureTopicPartitions = futureLogs.collect {
-        case (tp, log) if log.parentDir == dir => tp
+        offlineTopicPartitions
       }
-      offlineFutureTopicPartitions.foreach { topicPartition => {
-        val removedLog = futureLogs.remove(topicPartition)
-        if (removedLog != null) {
-          removedLog.closeHandlers()
-          removedLog.removeLogMetrics()
-        }
-      }}
+
+      val offlineCurrentTopicPartitions = removeOfflineLogs(currentLogs)
+      val offlineFutureTopicPartitions = removeOfflineLogs(futureLogs)
 
       warn(s"Logs for partitions ${offlineCurrentTopicPartitions.mkString(",")} are offline and " +
            s"logs for future partitions ${offlineFutureTopicPartitions.mkString(",")} are offline due to failure on log directory $dir")
@@ -932,6 +927,7 @@ class LogManager(logDirs: Seq[File],
         val logsToCheckpoint = logsInDir(logDir)
         checkpointRecoveryOffsetsAndCleanSnapshotsInDir(logDir, logsToCheckpoint, ArrayBuffer.empty)
         checkpointLogStartOffsetsInDir(logDir, logsToCheckpoint)
+        sourceLog.removeLogMetrics()
         addLogToBeDeleted(sourceLog)
       } catch {
         case e: KafkaStorageException =>
@@ -959,10 +955,7 @@ class LogManager(logDirs: Seq[File],
                   isFuture: Boolean = false,
                   checkpoint: Boolean = true): Log = {
     val removedLog: Log = logCreationOrDeletionLock synchronized {
-      if (isFuture)
-        futureLogs.remove(topicPartition)
-      else
-        currentLogs.remove(topicPartition)
+      removeLogAndMetrics(if (isFuture) futureLogs else currentLogs, topicPartition)
     }
     if (removedLog != null) {
       // We need to wait until there is no more cleaning task on the log to be deleted before actually deleting it.
@@ -978,7 +971,6 @@ class LogManager(logDirs: Seq[File],
         checkpointRecoveryOffsetsAndCleanSnapshotsInDir(logDir, logsToCheckpoint, ArrayBuffer.empty)
         checkpointLogStartOffsetsInDir(logDir, logsToCheckpoint)
       }
-      removedLog.removeLogMetrics()
       addLogToBeDeleted(removedLog)
       info(s"Log for partition ${removedLog.topicPartition} is renamed to ${removedLog.dir.getAbsolutePath} and is scheduled for deletion")
     } else if (offlineLogDirs.nonEmpty) {
@@ -1151,6 +1143,12 @@ class LogManager(logDirs: Seq[File],
           error(s"Error flushing topic ${topicPartition.topic}", e)
       }
     }
+  }
+
+  private def removeLogAndMetrics(logs: Pool[TopicPartition, Log], tp: TopicPartition): Log = {
+    val removedLog = logs.remove(tp)
+    if (removedLog != null) removedLog.removeLogMetrics()
+    removedLog
   }
 }
 
--- a/core/src/main/scala/kafka/log/LogManager.scala
+++ b/core/src/main/scala/kafka/log/LogManager.scala
@@ -205,8 +205,8 @@ class LogManager(logDirs: Seq[File],
         }
         offlineTopicPartitions.foreach { topicPartition => {
           val removedLog = removeLogAndMetrics(logs, topicPartition)
-          if (removedLog != null) {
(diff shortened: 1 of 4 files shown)


dajac (approved):
LGTM, thanks for the fix.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -978,6 +978,7 @@ class LogManager(logDirs: Seq[File],
         checkpointRecoveryOffsetsAndCleanSnapshotsInDir(logDir, logsToCheckpoint, ArrayBuffer.empty)
         checkpointLogStartOffsetsInDir(logDir, logsToCheckpoint)
       }
+      removedLog.removeLogMetrics()
```
dhruvilshah3: We use the same pattern of removing the log from `futureLogs` or `currentLogs` and then removing the corresponding log metrics in `handleLogDirFailure`. Do you think it'd be useful to turn this into a helper method and reuse the logic in both places?

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
```diff
@@ -563,4 +565,39 @@ class LogManagerTest {
...
+    assertTrue(logMetrics.isEmpty)
+
+    logManager.getOrCreateLog(tp, () => logConfig)
+
+    assertEquals(LogMetricNames.allMetricNames.size, logMetrics.size)
+    logMetrics.foreach { metric =>
+      assertTrue(metric.getMBeanName.contains(metricTag))
+    }
+
+    time.sleep(logConfig.fileDeleteDelayMs + 1)
```
dhruvilshah3, line 600: Why is this sleep needed?

bob-barrett: It causes enough "time" to elapse that the actual file deletion is triggered. Having it here tests that we still have metrics after the deletion occurs. I'll add a comment to that effect.

core/src/main/scala/kafka/log/LogManager.scala
bob-barrett: That's a good catch, thanks. I'll add that.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
```diff
@@ -563,4 +565,39 @@ class LogManagerTest {
     logManager.topicConfigUpdated("test-topic")
     assertTrue(logManager.partitionsInitializing.isEmpty)
   }
+
+  @Test
+  def testMetricsExistWhenLogIsRecreatedBeforeDeletion(): Unit = {
+    val topicName = "metric-test"
+    def logMetrics = KafkaYammerMetrics.defaultRegistry.allMetrics.keySet.asScala.
+      filter(_.getType == "Log").filter(_.getScope.contains(topicName))
```
soondenana: nit: maybe combine consecutive filters.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
```diff
@@ -563,4 +565,39 @@ class LogManagerTest {
...
+    logManager.asyncDelete(tp)
+
+    assertTrue(logMetrics.isEmpty)
+
+    logManager.getOrCreateLog(tp, () => logConfig)
+
+    assertEquals(LogMetricNames.allMetricNames.size, logMetrics.size)
+    logMetrics.foreach { metric =>
+      assertTrue(metric.getMBeanName.contains(metricTag))
+    }
```
soondenana: duplicate code, its being scala maybe define an internal helper method. There is another instance of this below.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
soondenana: Is it deterministic? I realize its mocktime, but threads can be scheduled in any order. Or using `MockTime` means all schedulers will get triggered before next line in test will get to run?

soondenana (commented):
Thanks for fixing it. Don't have much to add to what @dhruvilshah3 said. Left couple of comments in test.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
bob-barrett: Yes, `MockTime` contains a `MockScheduler`, which explicitly checks for any scheduled tasks that need to be run whenever we sleep. And we pass that `MockScheduler` into the `Log` and `LogManager` constructors in the tests. See https://github.com/apache/kafka/blob/trunk/core/src/test/scala/unit/kafka/utils/MockScheduler.scala#L59.

bob-barrett:
While addressing the feedback, I realized I had missed the case when replacing a current log with a future log as part of altering a partition's log dir. We also deferred metrics removal until the file deletion in that case. That has been fixed now.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -1151,6 +1144,12 @@ class LogManager(logDirs: Seq[File],
       }
     }
   }
+
+  private def removeLogAndMetrics(logs: Pool[TopicPartition, Log], tp: TopicPartition): Log = {
+    val removedLog = logs.remove(tp)
+    if (removedLog != null) removedLog.removeLogMetrics()
+    removedLog
```
soondenana: nit: Lets return Option(removedLog) to ease null checking by clients.

Seems like the same object gets returned by `asyncDelete` but is only used in one place in test code, so many want to change the return value of that too. The less "null" we have the better.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -199,27 +199,22 @@ class LogManager(logDirs: Seq[File],
...
 
-      val offlineCurrentTopicPartitions = currentLogs.collect {
-        case (tp, log) if log.parentDir == dir => tp
-      }
-      offlineCurrentTopicPartitions.foreach { topicPartition => {
-        val removedLog = currentLogs.remove(topicPartition)
-        if (removedLog != null) {
-          removedLog.closeHandlers()
-          removedLog.removeLogMetrics()
+      def removeOfflineLogs(logs: Pool[TopicPartition, Log]): Iterable[TopicPartition] = {
```
soondenana, line 202: Thanks for deduping this code.

soondenana (approved):
LGTM

Left one nit.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -1151,6 +1149,16 @@ class LogManager(logDirs: Seq[File],
...
   }
+
+  private def removeLogAndMetrics(logs: Pool[TopicPartition, Log], tp: TopicPartition): Option[Log] = {
+    val removedLog = logs.remove(tp)
+    if (removedLog != null) {
+      removedLog.removeLogMetrics()
+      Some(removedLog)
+    } else {
+      None
+    }
```
soondenana, line 1160: nit: I think just returning `Option(removedLog)` should take care of Some/None part.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -199,27 +199,22 @@ class LogManager(logDirs: Seq[File],
...
+      def removeOfflineLogs(logs: Pool[TopicPartition, Log]): Iterable[TopicPartition] = {
+        val offlineTopicPartitions: Iterable[TopicPartition] = logs.collect {
+          case (tp, log) if log.parentDir == dir => tp
         }
-      }}
+        offlineTopicPartitions.foreach { topicPartition => {
+          val removedLog = removeLogAndMetrics(logs, topicPartition)
+          removedLog.foreach {
+            log => log.closeHandlers()
+          }
```
soondenana, line 210: nit: as its one line, can be abbreviated as `_.closeHandlers()`

soondenana (approved):
Thanks for fixing it.

Couple more nits, feel free to ignore if its too much work (or you don't see value)

ijuma:
ok to test

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -932,6 +927,7 @@ class LogManager(logDirs: Seq[File],
         val logsToCheckpoint = logsInDir(logDir)
         checkpointRecoveryOffsetsAndCleanSnapshotsInDir(logDir, logsToCheckpoint, ArrayBuffer.empty)
         checkpointLogStartOffsetsInDir(logDir, logsToCheckpoint)
+        sourceLog.removeLogMetrics()
```
dhruvilshah3, line 930: This seems a bit odd but looks like an existing issue. We are removing metrics for the `current` log but we leave the metrics for the `future` log as is, which means it will retain the `is-future` tag. Maybe worth filing a JIRA if you think it is a problem too.

dhruvilshah3 (approved):
Thanks for the PR, LGTM. I left a comment about a potential existing issue.

core/src/main/scala/kafka/log/LogManager.scala
bob-barrett: Good catch. Opened https://issues.apache.org/jira/browse/KAFKA-10320 to track it.

gwenshap:
retest this please

gwenshap:
Test failures:
org.apache.kafka.connect.mirror.MirrorConnectorsIntegrationTest.testOneWayReplicationWithAutorOffsetSync1
in both JDK 11 and JDK 14. My understanding is that this isn't related to this PR.

bob-barrett:
Test failure looks unrelated and passes locally, probably a flake

gwenshap (commented):
post-facto LGTM 👍

PR #9060, opened 2020-07-23
KAFKA-9274: Remove `retries` from InternalTopicManager

mjsax:
- part of KIP-572
 - replace `retries` in InternalTopicManager with infinite retires plus
   a new timeout, based on consumer config MAX_POLL_INTERVAL_MS
- if the new timeout hits, we don't throw `StreamsException` any longer (as we did when exceeding retries), but send `INCOMPLETE_SOURCE_TOPIC_METADATA` error code to let all instances shut down

Third PR for KIP-572 (cf #8864 and #9047)

Call for review @vvcephei

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -25,12 +25,11 @@ import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.common.Metric;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.common.metrics.JmxReporter;
-import org.apache.kafka.common.metrics.MetricsContext;
 import org.apache.kafka.common.metrics.KafkaMetricsContext;
 import org.apache.kafka.common.metrics.MetricConfig;
 import org.apache.kafka.common.metrics.Metrics;
+import org.apache.kafka.common.metrics.MetricsContext;
 import org.apache.kafka.common.metrics.MetricsReporter;
 import org.apache.kafka.common.metrics.Sensor;
 import org.apache.kafka.common.metrics.Sensor.RecordingLevel;
@@ -1249,11 +1248,7 @@ public class KafkaStreams implements AutoCloseable {
 
         log.debug("Current changelog positions: {}", allChangelogPositions);
         final Map<TopicPartition, ListOffsetsResultInfo> allEndOffsets;
-        try {
-            allEndOffsets = fetchEndOffsets(allPartitions, adminClient);
-        } catch (final TimeoutException e) {
-            throw new StreamsException("Timed out obtaining end offsets from kafka", e);
-        }
+        allEndOffsets = fetchEndOffsets(allPartitions, adminClient);
         log.debug("Current end offsets :{}", allEndOffsets);
 
         for (final Map.Entry<TopicPartition, ListOffsetsResultInfo> entry : allEndOffsets.entrySet()) {
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
@@ -33,6 +33,8 @@ import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.errors.StreamsException;
+import org.apache.kafka.streams.errors.TaskAssignmentException;
+import org.apache.kafka.streams.processor.internals.ClientUtils.QuietConsumerConfig;
 import org.slf4j.Logger;
 
 import java.util.HashMap;
@@ -98,7 +100,7 @@ public class InternalTopicManager {
      * If a topic exists already but has different number of partitions we fail and throw exception requesting user to reset the app before restarting again.
      * @return the set of topics which had to be newly created
      */
-    public Set<String> makeReady(final Map<String, InternalTopicConfig> topics) {
+    public Set<String> makeReady(final Map<String, InternalTopicConfig> topics) throws TaskAssignmentException {
         // we will do the validation / topic-creation in a loop, until we have confirmed all topics
         // have existed with the expected number of partitions, or some create topic returns fatal errors.
         log.debug("Starting to validate internal topics {} in partition assignor.", topics);
@@ -178,9 +180,7 @@ public class InternalTopicManager {
                     final String timeoutError = String.format("Could not create topics within %d milliseconds. " +
                         "This can happen if the Kafka cluster is temporary not available.", retryTimeoutMs);
                     log.error(timeoutError);
-                    // TODO: should we throw a different exception instead and catch it, to return a `INCOMPLETE_SOURCE_TOPIC_METADATA` error code
-                    throw new StreamsException(timeoutError);
-
+                    throw new TaskAssignmentException(timeoutError);
                 }
                 log.info(
                     "Topics {} could not be made ready. Will retry in {} milliseconds. Remaining time in milliseconds: {}",
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
@@ -33,7 +33,6 @@ import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.errors.StreamsException;
-import org.apache.kafka.streams.errors.TaskAssignmentException;
 import org.apache.kafka.streams.processor.internals.ClientUtils.QuietConsumerConfig;
 import org.slf4j.Logger;
 
@@ -100,7 +99,7 @@ public class InternalTopicManager {
      * If a topic exists already but has different number of partitions we fail and throw exception requesting user to reset the app before restarting again.
      * @return the set of topics which had to be newly created
      */
-    public Set<String> makeReady(final Map<String, InternalTopicConfig> topics) throws TaskAssignmentException {
+    public Set<String> makeReady(final Map<String, InternalTopicConfig> topics) {
         // we will do the validation / topic-creation in a loop, until we have confirmed all topics
         // have existed with the expected number of partitions, or some create topic returns fatal errors.
         log.debug("Starting to validate internal topics {} in partition assignor.", topics);
@@ -178,9 +177,9 @@ public class InternalTopicManager {
 
                 if (currentWallClockMs >= deadlineMs) {
                     final String timeoutError = String.format("Could not create topics within %d milliseconds. " +
-                        "This can happen if the Kafka cluster is temporary not available.", retryTimeoutMs);
+                        "This can happen if the Kafka cluster is temporarily not available.", retryTimeoutMs);
                     log.error(timeoutError);
-                    throw new TaskAssignmentException(timeoutError);
+                    throw new TimeoutException(timeoutError);
                 }
                 log.info(
                     "Topics {} could not be made ready. Will retry in {} milliseconds. Remaining time in milliseconds: {}",
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
@@ -17,16 +17,19 @@
 package org.apache.kafka.streams.processor.internals;
 
 import org.apache.kafka.clients.admin.Admin;
-import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.admin.CreateTopicsResult;
 import org.apache.kafka.clients.admin.DescribeTopicsResult;
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.admin.TopicDescription;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.common.KafkaFuture;
 import org.apache.kafka.common.errors.LeaderNotAvailableException;
+import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.common.errors.TopicExistsException;
 import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
+import org.apache.kafka.common.serialization.ByteArrayDeserializer;
 import org.apache.kafka.common.utils.LogContext;
+import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.errors.StreamsException;
@@ -45,17 +48,21 @@ public class InternalTopicManager {
         "Please report at https://issues.apache.org/jira/projects/KAFKA or dev-mailing list (https://kafka.apache.org/contact).";
 
     private final Logger log;
-    private final long windowChangeLogAdditionalRetention;
-    private final Map<String, String> defaultTopicConfigs = new HashMap<>();
 
-    private final short replicationFactor;
+    private final Time time;
     private final Admin adminClient;
 
-    private final int retries;
+    private final short replicationFactor;
+    private final long windowChangeLogAdditionalRetention;
     private final long retryBackOffMs;
+    private final long retryTimeoutMs;
 
-    @SuppressWarnings("deprecation") // TODO: remove in follow up PR when `RETRIES` is removed
-    public InternalTopicManager(final Admin adminClient, final StreamsConfig streamsConfig) {
+    private final Map<String, String> defaultTopicConfigs = new HashMap<>();
+
+    public InternalTopicManager(final Time time,
+                                final Admin adminClient,
+                                final StreamsConfig streamsConfig) {
+        this.time = time;
         this.adminClient = adminClient;
 
         final LogContext logContext = new LogContext(String.format("stream-thread [%s] ", Thread.currentThread().getName()));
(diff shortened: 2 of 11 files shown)


streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1249,11 +1248,7 @@ public void cleanUp() {
 
         log.debug("Current changelog positions: {}", allChangelogPositions);
         final Map<TopicPartition, ListOffsetsResultInfo> allEndOffsets;
-        try {
-            allEndOffsets = fetchEndOffsets(allPartitions, adminClient);
-        } catch (final TimeoutException e) {
-            throw new StreamsException("Timed out obtaining end offsets from kafka", e);
-        }
+        allEndOffsets = fetchEndOffsets(allPartitions, adminClient);
```
mjsax, line 1251: Minor side fix: `fetchEndOffset` can never throw `TimeoutException` because it catches all `RuntimeException` and convert them into `StreamsException` already

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
```diff
@@ -46,33 +49,38 @@
...
 
-    private final int retries;
+    private final short replicationFactor;
+    private final long windowChangeLogAdditionalRetention;
     private final long retryBackOffMs;
+    private final long retryTimeoutMs;
+
+    private final Map<String, String> defaultTopicConfigs = new HashMap<>();
 
-    @SuppressWarnings("deprecation") // TODO: remove in follow up PR when `RETRIES` is removed
```
mjsax, line 57: first TODO removed

streams/src/test/java/org/apache/kafka/streams/StreamsConfigTest.java
```diff
@@ -159,7 +159,6 @@ public void testGetGroupInstanceIdConfigs() {
         assertNull(returnedProps.get(ConsumerConfig.GROUP_INSTANCE_ID_CONFIG));
     }
 
-    @SuppressWarnings("deprecation") // TODO revisit in follow up PR
```
mjsax, line 162: Second TODO removed

streams/src/test/java/org/apache/kafka/streams/StreamsConfigTest.java
```diff
@@ -186,25 +183,9 @@ public void consumerConfigMustContainStreamPartitionAssignorConfig() {
         );
         assertEquals(7L, returnedProps.get(StreamsConfig.WINDOW_STORE_CHANGE_LOG_ADDITIONAL_RETENTION_MS_CONFIG));
         assertEquals("dummy:host", returnedProps.get(StreamsConfig.APPLICATION_SERVER_CONFIG));
-        assertNull(returnedProps.get(StreamsConfig.RETRIES_CONFIG));
-        assertEquals(5, returnedProps.get(StreamsConfig.adminClientPrefix(StreamsConfig.RETRIES_CONFIG)));
         assertEquals(100, returnedProps.get(StreamsConfig.topicPrefix(TopicConfig.SEGMENT_BYTES_CONFIG)));
     }
 
-    @SuppressWarnings("deprecation") // TODO revisit in follow up PR
```
mjsax, line 194: Third TODO removed -- as we don't pass `retires` and `retry.backoff.ms` via admin config any longer, this test is not needed any more

streams/src/test/java/org/apache/kafka/streams/processor/internals/InternalTopicManagerTest.java
```diff
@@ -71,21 +73,20 @@
...
     private final String topic3 = "test_topic_3";
     private final List<Node> singleReplica = Collections.singletonList(broker1);
-    private final int numRetries = 1;
 
     private String threadName;
 
     private MockAdminClient mockAdminClient;
     private InternalTopicManager internalTopicManager;
 
-    @SuppressWarnings("deprecation") // TODO revisit in follow up PR
```
mjsax, line 81: Fourth TODO removed

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
```diff
@@ -154,27 +164,34 @@ public InternalTopicManager(final Admin adminClient, final StreamsConfig streams
...
             if (!topicsNotReady.isEmpty()) {
-                log.info("Topics {} can not be made ready with {} retries left", topicsNotReady, remainingRetries);
+                currentWallClockMs = time.milliseconds();
 
-                Utils.sleep(retryBackOffMs);
+                if (currentWallClockMs >= deadlineMs) {
+                    final String timeoutError = String.format("Could not create topics within %d milliseconds. " +
+                        "This can happen if the Kafka cluster is temporary not available.", retryTimeoutMs);
+                    log.error(timeoutError);
+                    // TODO: should we throw a different exception instead and catch it, to return a `INCOMPLETE_SOURCE_TOPIC_METADATA` error code
```
mjsax: @vvcephei this is an open question

\cc @guozhangwang

mjsax:
Triggered system tests: https://jenkins.confluent.io/job/system-test-kafka-branch-builder/4071/

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
```diff
@@ -96,13 +104,15 @@ public InternalTopicManager(final Admin adminClient, final StreamsConfig streams
         // have existed with the expected number of partitions, or some create topic returns fatal errors.
         log.debug("Starting to validate internal topics {} in partition assignor.", topics);
 
-        int remainingRetries = retries;
+        long currentWallClockMs = time.milliseconds();
+        final long deadlineMs = currentWallClockMs + retryTimeoutMs;
```
dajac, line 108: If we want to guarantee that the `deadlineMs` is respected, I think that we must set the timeout of the AdminClient's call accordingly: `CreateTopicsOptions.timeoutMs`. With the default, I think that the call could be longer than half of `MAX_POLL_INTERVAL_MS_CONFIG`.

mjsax: Good question. Default `max.poll.interval.ms` is 5 minutes (ie, the deadline is set to 2.5 minutes by default) while default `api.default.timeout.ms` is 1 minutes? Thus we might be ok?

dajac: That's right. I misread the default value of `max.poll.interval.ms`, too many zeros for my eyes ;). The default works fine then. Do we want to protect ourselves if the user changes the default? Or shall we just call out that `api.default.timeout.ms` should be lower than `max.poll.interval.ms` somewhere?

mjsax:
Only the `StreamsStandbyTask.test_standby_tasks_rebalance` system test failed and it's know to be buggy.

mjsax:
Jenkins failed on know flaky tests only.

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
mjsax: I am happy to add a check in `StreamsConfig` and either throw or log a WARN depending how strict we want to be.

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
guozhangwang: I'm thinking that moving forward we should try to not create internal topics during rebalance but try pre-create in starting, but for now assuming this is still the case I think letting the whole application to die is fine --- i.e. treat it the same as source topics. Hence I'm leaning towards encoding INCOMPLETE_SOURCE_TOPIC_METADATA to shutdown the whole app, across all clients.

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
dajac: Thinking a bit more about this, with the default, you may end up not honouring the deadline. `createTopics` can take up to 1m so if you invoke one when less than 1m is reaming before the deadline, you may not honour the deadline. It may not be that important though.

If we want to strictly enforce it, we could calculate the maximum timeout for each call, something like deadline - now, and set it with `CreateTopicsOptions.timeoutMs`.

streams/src/test/java/org/apache/kafka/streams/processor/internals/InternalTopicManagerTest.java
```diff
@@ -71,21 +73,20 @@
...
     private final Map<String, Object> config = new HashMap<String, Object>() {
         {
             put(StreamsConfig.APPLICATION_ID_CONFIG, "app-id");
             put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, broker1.host() + ":" + broker1.port());
             put(StreamsConfig.REPLICATION_FACTOR_CONFIG, 1);
             put(StreamsConfig.producerPrefix(ProducerConfig.BATCH_SIZE_CONFIG), 16384);
-            put(StreamsConfig.adminClientPrefix(StreamsConfig.RETRIES_CONFIG), numRetries);
+            put(StreamsConfig.consumerPrefix(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG), 100);
```
abbccdda, line 88: Why do we reduce max poll interval?

abbccdda (commented):
Overall LGTM

abbccdda:
Also, do we have new unit test coverage for the changes?

streams/src/test/java/org/apache/kafka/streams/processor/internals/InternalTopicManagerTest.java
mjsax: In this PR, we change the "deadline" in the group leader to create/verify all internal topics from "counting retries" to a timeout of `max.poll.interval.ms / 2` and we reduce the default of 5 minutes to speed up this test.

cf https://github.com/apache/kafka/pull/9060/files#diff-d3963e433c59b08688bb4481faa20e97R79

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
mjsax: Ack

mjsax:
Updated. Call for review @abbccdda 

> Also, do we have new unit test coverage for the changes?

We did not really change much, only switching from `retries` to a "timeout" and thus existing unit tests (that I updated accordingly, eg, different configs and/or different exception type) should be sufficient?

The only other thing is the new error code and I just added 2 unit tests for this case.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -373,8 +373,15 @@ public GroupAssignment assign(final Cluster metadata, final GroupSubscription gr
 
         final Set<TaskId> statefulTasks = new HashSet<>();
 
-        final boolean probingRebalanceNeeded =
-            assignTasksToClients(fullMetadata, allSourceTopics, topicGroups, clientMetadataMap, partitionsForTask, statefulTasks);
+        final boolean probingRebalanceNeeded;
+        try {
+            probingRebalanceNeeded = assignTasksToClients(fullMetadata, allSourceTopics, topicGroups, clientMetadataMap, partitionsForTask, statefulTasks);
```
abbccdda, line 378: Is this part of the initiative to throw a different exception? Could we update the summary of this PR?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -477,7 +484,7 @@ private boolean checkMetadataVersions(final int minReceivedMetadataVersion,
      * @return map from repartition topic to its partition info
      */
     private Map<TopicPartition, PartitionInfo> prepareRepartitionTopics(final Map<Integer, TopicsInfo> topicGroups,
-                                                                           final Cluster metadata) {
+                                                                           final Cluster metadata) throws TaskAssignmentException {
```
abbccdda: Could we update the meta comment to explain when a `TaskAssignmentException` is thrown?

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
```diff
@@ -91,18 +100,20 @@ public InternalTopicManager(final Admin adminClient, final StreamsConfig streams
      * If a topic exists already but has different number of partitions we fail and throw exception requesting user to reset the app before restarting again.
      * @return the set of topics which had to be newly created
      */
-    public Set<String> makeReady(final Map<String, InternalTopicConfig> topics) {
+    public Set<String> makeReady(final Map<String, InternalTopicConfig> topics) throws TaskAssignmentException {
```
abbccdda: Comments for thrown exception

streams/src/test/java/org/apache/kafka/streams/processor/internals/InternalTopicManagerTest.java
```diff
@@ -383,18 +405,25 @@ public void shouldThrowExceptionWhenKeepsTopicLeaderNotAvailable() {
...
-            () -> topicManager.makeReady(Collections.singletonMap(topic, internalTopicConfig)));
+        final TaskAssignmentException exception = assertThrows(
+            TaskAssignmentException.class,
+            () -> topicManager.makeReady(Collections.singletonMap(topic, internalTopicConfig))
+        );
+        assertNull(exception.getCause());
+        assertThat(
+            exception.getMessage(),
+            equalTo("Could not create topics within 50 milliseconds." +
+                " This can happen if the Kafka cluster is temporary not available.")
```
abbccdda: s/temporary/temporarily

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
mjsax: `TaskAssignmentException` is not a checked exception and it's just a curtesy declaration... (I can also remove `throws TaskAssignmentException` if you prefer).

Don't see any need to document anything further -- the code makes it clear (in fact, this method does not even throw the exception itself, but it just bubbles up from `internalTopicManager.makeReady` and the code documents itself:
https://github.com/apache/kafka/pull/9060/files#diff-d3963e433c59b08688bb4481faa20e97R179-R184

mjsax:
@abbccdda I updated the PR. I also realized, that it might be better to throw a `TimeoutException` instead of a `TaskAssignmentException` if we hit the new timeout. Updated the code and tests accordingly.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -373,8 +373,15 @@ public GroupAssignment assign(final Cluster metadata, final GroupSubscription gr
 
         final Set<TaskId> statefulTasks = new HashSet<>();
 
-        final boolean probingRebalanceNeeded =
-            assignTasksToClients(fullMetadata, allSourceTopics, topicGroups, clientMetadataMap, partitionsForTask, statefulTasks);
+        final boolean probingRebalanceNeeded;
+        try {
+            probingRebalanceNeeded = assignTasksToClients(fullMetadata, allSourceTopics, topicGroups, clientMetadataMap, partitionsForTask, statefulTasks);
+        } catch (final TaskAssignmentException | TimeoutException e) {
+            return new GroupAssignment(
```
abbccdda, line 380: Since we could throw different exceptions here, would be good to add a log to indicate which type of exception is thrown.

abbccdda (commented):
Just one minor comment.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
mjsax: There are already corresponding log.error statement before those exceptions are thrown. No need to double log IMHO?

abbccdda: Cool, sg.

abbccdda (approved):
LGTM

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
```diff
@@ -45,33 +49,38 @@
...
         windowChangeLogAdditionalRetention = streamsConfig.getLong(StreamsConfig.WINDOW_STORE_CHANGE_LOG_ADDITIONAL_RETENTION_MS_CONFIG);
-        final AdminClientConfig adminConfigs = new ClientUtils.QuietAdminClientConfig(streamsConfig);
-        retries = adminConfigs.getInt(AdminClientConfig.RETRIES_CONFIG);
-        retryBackOffMs = adminConfigs.getLong(AdminClientConfig.RETRY_BACKOFF_MS_CONFIG);
+        retryBackOffMs = streamsConfig.getLong(StreamsConfig.RETRY_BACKOFF_MS_CONFIG);
+        final Map<String, Object> consumerConfig = streamsConfig.getMainConsumerConfigs("dummy", "dummy", -1);
+        // need to add mandatory configs; otherwise `QuietConsumerConfig` throws
+        consumerConfig.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, ByteArrayDeserializer.class);
+        consumerConfig.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, ByteArrayDeserializer.class);
+        retryTimeoutMs = new QuietConsumerConfig(consumerConfig).getInt(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG) / 2L;
```
ableegoldman, line 79: Hey @mjsax , am I reading this PR correctly? Do we now only allow a single member to retry topic creation/validation for up to half of the poll interval, after which we shut down the entire application? That sounds like the opposite of resiliency...what if the brokers are temporarily unavailable? Before this we would just let the single thread die, and the internal topic creation/validation would be retried on the subsequent rebalance. That wasn't ideal, but given the upcoming work to allow reviving/recreating a death thread, that seems to be preferable to permanently ending the application? 

Sorry if I'm misreading this, was just going over all the PRs in the last month or so to produce a diff+summary of the important ones, and want to make sure I actually understand all the changes we've made

ableegoldman: Apologies if this was touched on in the KIP, it's been a while and the discussion thread was quite long so I may have missed something there

mjsax: Note that the previous default was "zero retries" and thus the new default is more resilient with a 5 minute default max.poll.interval. -- But yes, we shutdown the whole app for this case now as proposed by @guozhangwang (IIRC).

guozhangwang: Today since we do not have ways to partially create tasks we'd have to create all topics to make sure all tasks are "complete" within each rebalance, if we cannot successfully create the topics within the poll.interval (i.e. we'd need to complete that rebalance with the poll.interval, and I guess halving it is to be more conservative), then killing that thread is not very useful anyways since we cannot proceed with the initializable tasks anyways.

That being said, with the upcoming work I'd agree that just shutdown the thread and allow users to optionally retry rebalance with new threads would be preferrable.

guozhangwang: cc @cadonna @wcarlson5 to bring to your attention.

ableegoldman: Yeah I think we should remove the shutdown error code in case of TimeoutException during internal topic validation before 2.7. I'll create a ticket so we don't lose track -- I think even just letting it kill the one thread is better than killing all of them

mjsax: SGTM.

PR #9064, opened 2020-07-23
KAFKA-10205: Documentation and handling of non deterministic Topologies

soarez:
Kafka Streams topologies must be defined in a deterministic order, otherwise users can run into confusing NPE.
This patch adds a more descriptive exception and documentation clarifying the deterministic requirement.

https://issues.apache.org/jira/browse/KAFKA-10205

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/StreamsBuilder.java
+++ b/streams/src/main/java/org/apache/kafka/streams/StreamsBuilder.java
@@ -51,6 +51,14 @@ import java.util.regex.Pattern;
 /**
  * {@code StreamsBuilder} provide the high-level Kafka Streams DSL to specify a Kafka Streams topology.
  *
+ * <p>
+ * It is a requirement that the processing logic ({@link Topology}) be defined in a deterministic way,
+ * as in, the order in which all operators are added must be predictable and the same across all application
+ * instances.
+ * Topologies are only identical if all operators are added in the same order.
+ * If different {@link KafkaStreams} instances of the same application build different topologies the result may be
+ * incompatible runtime code and unexpected results or errors
+ *
  * @see Topology
  * @see KStream
  * @see KTable
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
@@ -31,6 +31,7 @@ import org.apache.kafka.streams.errors.DeserializationExceptionHandler;
 import org.apache.kafka.streams.errors.LockException;
 import org.apache.kafka.streams.errors.StreamsException;
 import org.apache.kafka.streams.errors.TaskMigratedException;
+import org.apache.kafka.streams.errors.TopologyException;
 import org.apache.kafka.streams.processor.Cancellable;
 import org.apache.kafka.streams.processor.PunctuationType;
 import org.apache.kafka.streams.processor.Punctuator;
@@ -1109,6 +1110,14 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
 
         public RecordQueue createQueue(final TopicPartition partition) {
             final SourceNode<?, ?, ?, ?> source = topology.source(partition.topic());
+            if (source == null) {
+                throw new TopologyException(
+                        "Topic is unkown to the topology. " +
+                                "This may happen if different KafkaStreams instances of the same application execute different Topologies. " +
+                                "Note that Topologies are only identical if all operators are added in the same order."
+                );
+            }
+
             final TimestampExtractor sourceTimestampExtractor = source.getTimestampExtractor();
             final TimestampExtractor timestampExtractor = sourceTimestampExtractor != null ? sourceTimestampExtractor : defaultTimestampExtractor;
             return new RecordQueue(
--- a/streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
@@ -46,6 +46,7 @@ import org.apache.kafka.streams.errors.LockException;
 import org.apache.kafka.streams.errors.ProcessorStateException;
 import org.apache.kafka.streams.errors.StreamsException;
 import org.apache.kafka.streams.errors.TaskMigratedException;
+import org.apache.kafka.streams.errors.TopologyException;
 import org.apache.kafka.streams.processor.PunctuationType;
 import org.apache.kafka.streams.processor.Punctuator;
 import org.apache.kafka.streams.processor.StateStore;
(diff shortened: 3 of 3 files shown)


streams/src/main/java/org/apache/kafka/streams/StreamsBuilder.java
```diff
@@ -49,6 +49,9 @@
 /**
  * {@code StreamsBuilder} provide the high-level Kafka Streams DSL to specify a Kafka Streams topology.
  *
+ * It is a requirement that the processing logic (Topology) be defined in a deterministic way.
```
mjsax: Nit: insert `<p>` tag to actually get the new paragraph rendered.

Nit: `Topology -> `{@link Topology}`

It's not really clear what "deterministic" means. We should elaborate more.

streams/src/main/java/org/apache/kafka/streams/StreamsBuilder.java
```diff
@@ -49,6 +49,9 @@
...
+ * If different instances build different runtime code logic the resulting behavior may be unexpected.
```
mjsax: "different" for sure, but this implies that one might have an operator the other does not. The observed issue is, that even if both contain the same operator, they might be added in different order (and thus be named differently) to the `Topology`, thus we should stretch that order matters.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -181,8 +182,16 @@ public StreamTask(final TaskId id,
 
         final TimestampExtractor defaultTimestampExtractor = config.defaultTimestampExtractor();
         final DeserializationExceptionHandler defaultDeserializationExceptionHandler = config.defaultDeserializationExceptionHandler();
+        final Set<String> sourceTopics = topology.sourceTopics();
         for (final TopicPartition partition : partitions) {
-            final SourceNode source = topology.source(partition.topic());
+            final String topicName = partition.topic();
+            if (!sourceTopics.contains(topicName)) {
+                throw new TopologyException(
+                        "Topic not found " + topicName + ". Is the Streams Topology built in a deterministic way?"
```
mjsax: `Topic not found` sounds like as-if the topic was not found in the cluster -- however, what actually happened is that we received a record but the record's topic is unknown in the sub-topology.

Similar to above, "deterministic" is not really easy to understand. I would also not phrase it as a question, but as a statement:
```
... This may happen if different KafkaStreams instances of the same application execute different Topologies. Note that Topologies are only identical if all operators are added in the same order.
```

Or similar.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -1984,6 +1985,24 @@ public void shouldThrowWakeupExceptionOnInitializeOffsetsWhenWakeupException() {
         task.initializeStateStores();
     }
 
+    @Test(expected = TopologyException.class)
```
mjsax: We should not use this annotation but rather use `assertThrows` (we still have some code that does not use `assertThrows` but we try to lazily migrate our tests, as it provides a better test pattern).

mjsax (commented):
Thanks for the PR. Couple of comments/suggestions.

mjsax:
Btw: The PR should be against `trunk`.

soarez:
Thanks for review @mjsax. I've changed the base branch to `trunk` and believe I have addressed all your feedback so far. Please take another look.

streams/src/main/java/org/apache/kafka/streams/StreamsBuilder.java
```diff
@@ -50,6 +50,14 @@
 /**
  * {@code StreamsBuilder} provide the high-level Kafka Streams DSL to specify a Kafka Streams topology.
  *
+ * <p>
+ * It is a requirement that the processing logic ({@link Topology}) be defined in a deterministic way,
+ * as in, the order in which all operators are added must be predictable and the same across all application
+ * instances.
+ * Topologies are only identical if all operators are added in the same order.
+ * If different KafkaStream instances of the same application build different topologies the result may be
```
mjsax: {@link KafkaStreams}

streams/src/main/java/org/apache/kafka/streams/StreamsBuilder.java
```diff
@@ -50,6 +50,14 @@
...
+ * incompatible runtimes and unexpected results.
```
mjsax: `incompatible runtimes and unexpected results` -> `incompatible runtime code and unexpected results or errors.`

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -1916,6 +1917,40 @@ public void shouldAlwaysSuspendRunningTasks() {
...
+                null
+        );
+        final StreamsMetricsImpl metrics = new StreamsMetricsImpl(this.metrics, "test", StreamsConfig.METRICS_LATEST);
+        EasyMock.expect(stateManager.changelogPartitions()).andReturn(Collections.emptySet());
+        EasyMock.replay(stateManager);
+
+        // The processor topology is missing the topics
+        final ProcessorTopology topology = withSources(asList(), mkMap());
+
+        assertThrows(TopologyException.class, () ->
```
mjsax: nit: formatting: (we should also get the exception an verify the error message)
```
final TopologyException  exception = assertThrows(
    TopologyException.class,
    () -> new StreamTask(
        ...
    )
);

assertThat(exception.getMessage(), equalTo("..."));
```

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -1916,6 +1917,40 @@ public void shouldAlwaysSuspendRunningTasks() {
         assertThat(task.state(), equalTo(SUSPENDED));
     }
 
+    @Test
+    public void szTest() {
```
mjsax: `szTest` is a terrible test name: `shouldThrowTopologyExceptionIfTaskCreatedForUnknownTopic`

mjsax (commented):
Thanks for the PR @soarez -- couple of comments.

soarez:
Thanks for the review @mjsax. I've addressed your feedback. Please take another look.

soarez:
Ping @mjsax - a reminder to review this

mjsax:
Retest this please.

mjsax (approved):
LGTM.

vvcephei (approved):
This is great. Thanks @soarez !

mjsax:
Compile error:
```
/home/jenkins/jenkins-slave/workspace/kafka-pr-jdk11-scala2.13/streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java:1989: error: constructor StreamsMetricsImpl in class StreamsMetricsImpl cannot be applied to given types;
20:24:26         final StreamsMetricsImpl metrics = new StreamsMetricsImpl(this.metrics, "test", StreamsConfig.METRICS_LATEST);
20:24:26                                            ^
20:24:26   required: Metrics,String,String,Time
20:24:26   found: Metrics,String,String
20:24:26   reason: actual and formal argument lists differ in length
20:24:27 1 error
```

Seems you need to rebase?

soarez:
@mjsax rebased and fixed the error, now `./gradlew  :streams:unitTest` seems to run without errors over here.

mjsax:
Thanks for the PR @soarez! Merged to `trunk`.

PR #9067, opened 2020-07-23
MINOR: Streams integration tests should not call exit

vvcephei:
Fixes this issue:
```
Execution failed for task ':streams:unitTest'.
> Process 'Gradle Test Executor 69' finished with non-zero exit value 134
```

### DIFF
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -104,6 +104,7 @@
     </module>
 
     <module name="Regexp">
+      <property name="id" value="dontUseSystemExit"/>
       <property name="format" value="System\.exit"/>
       <property name="illegalPattern" value="true"/>
       <property name="ignoreComments" value="true"/>
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -103,6 +103,13 @@
       <property name="ignoreComments" value="true"/>
     </module>
 
+    <module name="Regexp">
+      <property name="format" value="System\.exit"/>
+      <property name="illegalPattern" value="true"/>
+      <property name="ignoreComments" value="true"/>
+      <property name="message" value="'System.exit': Should not directly call System.exit, but Exit.exit instead."/>
+    </module>
+
     <!-- code quality -->
     <module name="MethodLength"/>
     <module name="ParameterNumber">
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -17,8 +17,14 @@
               files="(ApiMessageType).java|MessageDataGenerator.java"/>
     <suppress checks="MethodLength"
               files="MessageDataGenerator.java"/>
+    <suppress checks="Regexp"
+              files="MessageGenerator.java"
+              lines="285,288"/>
 
     <!-- Clients -->
+    <suppress checks="Regexp"
+              files="Exit.java"
+              lines="43"/>
     <suppress checks="ClassFanOutComplexity"
               files="(Fetcher|Sender|SenderTest|ConsumerCoordinator|KafkaConsumer|KafkaProducer|Utils|TransactionManager|TransactionManagerTest|KafkaAdminClient|NetworkClient|Admin).java"/>
     <suppress checks="ClassFanOutComplexity"
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -17,14 +17,12 @@
               files="(ApiMessageType).java|MessageDataGenerator.java"/>
     <suppress checks="MethodLength"
               files="MessageDataGenerator.java"/>
-    <suppress checks="Regexp"
-              files="MessageGenerator.java"
-              lines="285,288"/>
+    <suppress id="dontUseSystemExit"
+              files="MessageGenerator.java"/>
 
     <!-- Clients -->
-    <suppress checks="Regexp"
-              files="Exit.java"
-              lines="43"/>
+    <suppress id="dontUseSystemExit"
+              files="Exit.java"/>
     <suppress checks="ClassFanOutComplexity"
               files="(Fetcher|Sender|SenderTest|ConsumerCoordinator|KafkaConsumer|KafkaProducer|Utils|TransactionManager|TransactionManagerTest|KafkaAdminClient|NetworkClient|Admin).java"/>
     <suppress checks="ClassFanOutComplexity"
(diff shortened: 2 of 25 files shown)


vvcephei:
Hey @mjsax , can you review this?

It's weird that the error is reported from `unitTest`, but the only call path to the `Exit.exit()` method is though this integration test. It may not completely fix the issue, but it doesn't hurt to put this in for now and see what happens.

mjsax:
Is this the right fix? I thought we should replace `System.exit()` with `Exit.exit()` to address the isssu? -- We call `System.exit()` in multiple system tests.

vvcephei:
Hey @mjsax ,

I double-checked, and it does seem to be the right fix. I confirmed that no matter whether you run the test with IDEA or gradle, there's nothing to automatically set the exit procedure, so `Exit.exit` is equivalent to `System.exit` unless a test specifically overrides the exit procedure, as I have done here.

Nevertheless, I see no reason not to also take your advice and just replace all calls to `System.exit` with `Exit.exit`. Since `Exit` is provided as a utility in the main Client module, it's available in every context in which Streams is available. Note, this by itself won't fix anything. If a test calls `Exit.exit`, it will fail with the reported error unless it overrides the exit procedure.

I did my best to trace all call paths that end in `exit()`, and still `SmokeTestDriverIntegrationTest` is the only one that I found.

vvcephei:
Hey @mjsax ,

Since Streams accounted for almost all of the `System.exit` calls, and since `Exit.exit` is available almost everywhere in AK, I'm proposing to add a checkstyle rule forbidding it. The violation will look like this:

```
[ant:checkstyle] [ERROR] /home/confluent/kafka/clients/src/main/java/org/apache/kafka/some/BadClass.java:43: Line matches the illegal pattern ''System.exit': Should not directly call System.exit, but Exit.exit instead.'. [Regexp]
```

lbradstreet:
> Hey @mjsax ,
> 
> I double-checked, and it does seem to be the right fix. I confirmed that no matter whether you run the test with IDEA or gradle, there's nothing to automatically set the exit procedure, so `Exit.exit` is equivalent to `System.exit` unless a test specifically overrides the exit procedure, as I have done here.
> 
> Nevertheless, I see no reason not to also take your advice and just replace all calls to `System.exit` with `Exit.exit`. Since `Exit` is provided as a utility in the main Client module, it's available in every context in which Streams is available. Note, this by itself won't fix anything. If a test calls `Exit.exit`, it will fail with the reported error unless it overrides the exit procedure.
> 
> I did my best to trace all call paths that end in `exit()`, and still `SmokeTestDriverIntegrationTest` is the only one that I found.

Right, you'll need ```Exit.setExitProcedure((_, _) => doSomething())``` to be setup somewhere to override the Exit.exit behavior, otherwise it will still do the same thing as it did before.

checkstyle/checkstyle.xml
```diff
@@ -103,6 +103,13 @@
       <property name="ignoreComments" value="true"/>
     </module>
 
+    <module name="Regexp">
+      <property name="format" value="System\.exit"/>
+      <property name="illegalPattern" value="true"/>
+      <property name="ignoreComments" value="true"/>
+      <property name="message" value="'System.exit': Should not directly call System.exit, but Exit.exit instead."/>
```
mjsax, line 111: Nice one!

mjsax (approved):
LGTM.

vvcephei:
Hmm. It looks like the Jenkins agent was shut down in the middle of the java 8 run. Retrying...

Retest this please.

vvcephei:
Thanks, @mjsax !

checkstyle/checkstyle.xml
lbradstreet: Thank you for this!

checkstyle/suppressions.xml
```diff
@@ -17,8 +17,14 @@
...
     <suppress checks="MethodLength"
               files="MessageDataGenerator.java"/>
+    <suppress checks="Regexp"
+              files="MessageGenerator.java"
+              lines="285,288"/>
 
     <!-- Clients -->
+    <suppress checks="Regexp"
+              files="Exit.java"
+              lines="43"/>
```
ijuma: It seems really brittle to have line numbers in the suppression. Could we:
1. Remove the `lines` part.
2. Give a name to the `Regexp` so that we only exclude the one related to `System.exit` (not sure if this is possible)

streams/examples/src/main/java/org/apache/kafka/streams/examples/wordcount/WordCountTransformerDemo.java
```diff
@@ -153,8 +154,8 @@ public void run() {
             streams.start();
             latch.await();
         } catch (final Throwable e) {
-            System.exit(1);
+            Exit.exit(1);
         }
-        System.exit(0);
+        Exit.exit(0);
```
ijuma: I think it was intentional not to use an internal class in demos that others may copy.

tools/src/main/java/org/apache/kafka/tools/VerifiableProducer.java
```diff
@@ -241,7 +240,7 @@ public static VerifiableProducer createFromArgs(ArgumentParser parser, String[]
         } else {
             parser.printHelp();
             // Can't use `Exit.exit` here because it didn't exist until 0.11.0.0.
-            System.exit(0);
+            Exit.exit(0);
```
ijuma: Did you read the comment immediately above this line? Also applies to `VerifiableConsumer`.

checkstyle/suppressions.xml
vvcephei: I agree, @ijuma , but this is the best I could come up with. Apparently, it's not possible to give it a name, so the only alternative would be to suppress _all_ Regexp checks for this file.

tools/src/main/java/org/apache/kafka/tools/VerifiableProducer.java
vvcephei: Haha, no I didn't. Sorry about that.

streams/examples/src/main/java/org/apache/kafka/streams/examples/wordcount/WordCountTransformerDemo.java
vvcephei: Ok, I can back this change out of the demos. I'd taken an "agricultural fire" approach to this, since it seemed like the error was coming from a source I couldn't identify. Namely, the reported error is from the "unitTest" task, whereas the only test that would have called "Sytem.exit" I could find was in the JUnit IntegrationTest category.

lbradstreet (approved):
LGTM

ijuma (approved):
LGTM

mjsax:
Merged to `trunk` and cherry-picked to `2.6` branch.

ijuma:
@mjsax did you run checkstyle and spotBugs before pushing the 2.6 change? It seems like it broke the build:

https://ci-builds.apache.org/job/Kafka/job/kafka-2.6-jdk8/3/console

mjsax:
Seems I made a mistake cherry-picking to `2.6` -- I remember that there was a conflict; guess I resolved it incorrectly. Did a HOTFIX PR: #9198

PR #5374, opened 2018-07-16
MINOR: update release.py

mjsax:
*More detailed description of your change,
if necessary. The PR title and PR message become
the squashed commit message, so use a separate
comment to ping reviewers.*

*Summary of testing strategy (including rationale)
for the feature or bug fix. Unit and/or integration
tests are expected for any behaviour change and
system tests should be considered for larger changes.*

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9070, opened 2020-07-24
MINOR: speed up release script

vvcephei:
Fixes slow release due to establishing a 
separate SSH connection per file to copy.

### DIFF
--- a/release.py
+++ b/release.py
@@ -567,8 +567,9 @@ print("Temporary build working director:", work_dir)
 kafka_dir = os.path.join(work_dir, 'kafka')
 streams_quickstart_dir = os.path.join(kafka_dir, 'streams/quickstart')
 print("Streams quickstart dir", streams_quickstart_dir)
-cmd("Creating staging area for release artifacts", "mkdir kafka-" + rc_tag, cwd=work_dir)
-artifacts_dir = os.path.join(work_dir, "kafka-" + rc_tag)
+artifact_name = "kafka-" + rc_tag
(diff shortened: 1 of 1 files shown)


vvcephei:
Hey all, after waiting at least 10 hours cumulatively for file uploads (so far) for the 2.5.1 release candidates, I've been feeling motivated to try and improve the situation.

Currently, it takes about 2 hours to upload the release artifacts, and almost all of that time is just establishing an ssh connection over and over again. Each connection takes about 6 seconds to establish, and then a few milliseconds to transfer the file.

I was a little surprised to see this since it looked like the code was supposed to be using sftp in batch mode, but was actually doing a batch of one command each time.

I found in https://github.com/apache/kafka/pull/5374 that this was actually intentional, because the full batch of commands was taking too long and making people think the command had hung.

I'm proposing to instead expand the toolchain to use `tar`, `rsync`, and `ssh`. These seem like reasonable requirements, and I've double-checked the commands on a Mac, and they seem to do the same thing as they do on Linux.

I don't know how long the batch upload was taking before, but the rsyncing the compressed artifact is taking me about one minute. Considering the compilation step a little earlier "hangs" with no output for at least two minutes, this seems to be ok.

With this change, the release script takes about five minutes to run instead of two hours.

WDYT?

omkreddy (approved):
@vvcephei  Thanks for the PR. LGTM.

ijuma:
Nice improvement!

rhauch:
Agreed! This significantly sped up putting together 2.6.0RC2.

PR #9072, opened 2020-07-24
KAFKA-10162; Make the rate based quota behave more like a Token Bucket (KIP-599, Part III)

dajac:
* Introduce a new SampledStat named TokenBucket
* Change the ControllerMutationManager to use it by default

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
@@ -22,14 +22,44 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 import org.apache.kafka.common.metrics.MetricConfig;
 
+/**
+ * The {@link TokenBucket} is a {@link SampledStat} implementing a token bucket that can be used
+ * in conjunction with a {@link Rate} to enforce a quota.
+ *
+ * A token bucket accumulates tokens with a constant rate R, one token per 1/R second, up to a
+ * maximum burst size B. The burst size essentially means that we keep permission to do a unit of
+ * work for up to B / R seconds.
+ *
+ * The {@link TokenBucket} adapts this to fit within a {@link SampledStat}. It accumulates tokens
+ * in chunks of Q units by default (sample length * quota) instead of one token every 1/R second
+ * and expires in chunks of Q units too (when the oldest sample is expired).
+ *
+ * Internally, we achieve this behavior by not completing the current sample until we fill that
+ * sample up to Q (used all credits). Samples are filled up one after the others until the maximum
+ * number of samples is reached. If it is not possible to created a new sample, we accumulate in
+ * the last one until a new one can be created. The over used credits are spilled over to the new
+ * sample at when it is created. Every time a sample is purged, Q credits are made available.
+ *
+ * It is important to note that the maximum burst is not enforced in the class and depends on
+ * how the quota is enforced in the {@link Rate}.
+ */
 public class TokenBucket extends SampledStat {
 
     private final TimeUnit unit;
 
+    /**
+     * Instantiates a new TokenBucket that works by default with a Quota {@link MetricConfig#quota()}
+     * in {@link TimeUnit#SECONDS}.
+     */
     public TokenBucket() {
         this(TimeUnit.SECONDS);
     }
 
+    /**
+     * Instantiates a new TokenBucket that works with the provided time unit.
+     *
+     * @param unit The time unit of the Quota {@link MetricConfig#quota()}
+     */
     public TokenBucket(TimeUnit unit) {
         super(0);
         this.unit = unit;
@@ -42,19 +72,19 @@ public class TokenBucket extends SampledStat {
             return;
         }
 
-        final double quota = quota(config);
+        final double sampleWindowQuota = sampleWindowQuota(config);
         final long firstTimeWindowMs = firstTimeWindowMs(config, timeMs);
 
-        // Get current sample or create one if empty
+        // Get current sample or create one if empty.
         Sample sample = current(firstTimeWindowMs);
 
         // Verify that the current sample was not reinitialized. If it was the case,
-        // restart from the first time window
+        // restart from the first time window.
         if (sample.eventCount == 0) {
             sample.reset(firstTimeWindowMs);
         }
 
-        // Add the value to the current sample
+        // Add the value to the current sample.
         sample.value += value;
         sample.eventCount += 1;
 
@@ -63,8 +93,8 @@ public class TokenBucket extends SampledStat {
         // new sample. Repeat until either the sample is not complete
         // or no new sample can be created.
         while (sample.isComplete(timeMs, config)) {
-            double extra = sample.value - quota;
-            sample.value = quota;
+            double extra = sample.value - sampleWindowQuota;
+            sample.value = sampleWindowQuota;
 
             sample = advance(config, sample.lastWindowMs + config.timeWindowMs());
             sample.value = extra;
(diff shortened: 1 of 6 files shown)


clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,150 @@
...
+ */
+package org.apache.kafka.common.metrics.stats;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import org.apache.kafka.common.metrics.MetricConfig;
+
+public class TokenBucket extends SampledStat {
```
dajac, line 46: Note to myself: Add javadoc.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private final TimeUnit unit;
+
+    public TokenBucket() {
+        this(TimeUnit.SECONDS);
+    }
+
+    public TokenBucket(TimeUnit unit) {
```
apovzner, line 63: This should be the same unit as the unit in `Rate`, right? If so, I think someone could create Rate as:
```new Rate(TimeUnit.MILLISECONDS, new TokenBucket())```
Or

```new Rate(new TokenBucket(TimeUnit.MILLISECONDS))```

File: clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java [hunk 3]
```diff
@@ -0,0 +1,150 @@
...
+            if (this.current <= 0)
+                this.current = this.samples.size() - 1;
+        }
+    }
+
+    private static double quota(MetricConfig config) {
+        if (config.quota() == null)
+            throw new IllegalStateException("TokenBucket can't be used without a quota.");
+
+        return config.quota().bound();
```
apovzner: I think this should return `config.quota().bound() * convert(config.timeWindowMs())`, because we are using the value returned from this method to compare to a value in a sample window. Also could be useful to add one unit test where MetricConfig.timeWindowMs() is not 1 second.

apovzner: and maybe consider renaming it to `sampleQuota` or `sampleWindowQuota` or something that hints that this method returns quota over a sample time window?

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,150 @@
...
+            if (value <= sample.value) {
+                sample.value -= value;
+                return;
+            }
+
+            sample.reset(timeMs);
+            value -= quota;
+
+            this.current = this.current - 1;
+            if (this.current <= 0)
```
apovzner: should this be `< 0`? Otherwise, we skip index 0

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: That is correct. This is bit unfortunate but as the two are independent from each others, we can't enforce using the same timeunit nor reuse the time unit of the rate.

I will make this clear in the javadoc.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: Good catch!

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: You're absolutely right.

dajac:
@apovzner Thanks for your comments. I have updated the PR to incorporate your feedback. I have also fixed a bug in unrecord and added the javadoc.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,179 @@
...
+     *
+     * @param unit The time unit of the Quota {@link MetricConfig#quota()}
+     */
+    public TokenBucket(TimeUnit unit) {
+        super(0);
+        this.unit = unit;
+    }
+
+    @Override
+    public void record(MetricConfig config, double value, long timeMs) {
```
junrao, line 69: Just a high level comment on the approach. This approach tries to spread a recorded value to multiple samples if the sample level quota is exceeded. While this matches the token bucket behavior for quota, it changes the behavior when we observe the value of the measurable. Currently, we record a full value in the current Sample. When we observe the value of the measurable, the effect of this value will last for the number of samples. After which, this value rolls out and no longer impacts the observed measurable. With this change, if we record a large value, the observed effect of the value could last much longer than the number of samples. For example, if we have 10 1-sec samples each with a quota of 1 and we record a value of 1000, the effect of this value will last for 1000 secs instead of 10 secs from the observability perspective. This may cause some confusion since we don't quite know when an event actually occurred.

I was thinking about an alternative approach that decouples the recording/observation of the measurable from quota calculation. Here is the rough idea. We create a customized SampledStat that records new values in a single sample as it is.  In addition, it maintains an accumulated available credit. As time advances, we add new credits based on the quota rate, capped by samples * perSampleQuota. When a value is recorded, we deduct the value from the credit and allow the credit to go below 0. We change the Sensor.checkQuotas() logic such that if the customized SampledStat is used, we throw QuotaViolationException if credit is < 0. This preserves the current behavior for observability, but allows quota to be enforced based on token bucket.

junrao (commented):
@dajac : Thanks for the PR. One high level comment below.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
apovzner: @junrao Regarding "With this change, if we record a large value, the observed effect of the value could last much longer than the number of samples. " -- This will not happen with this approach. If we record a very large value, we never move to the bucket with timestamp > current timestamp (of the recording time). This approach can only add the value to older buckets, which did not expire, but never to the buckets "in the future". 

For example, if we only have 2 samples, and perSampleQuota  = 5, and say we already filled in both buckets up to quota: [5, 5]. If new requests arrive but the timestamp did not move past the last bucket, we are going to be adding this value to the last bucket, for example getting to [5, 20] if we recorded 15. If the next recording happens after the time moved passed the last bucket, say we record 3, then buckets will look like [20, 3].

apovzner: @junrao If I understood your proposal correctly, we will keep Rate calculation the same, but additionally implement TokenBucket (traditional way) which will tell us when quota is violated. This would be much easier. However, I think, it would not fix our issue (that we are trying to fix) of too large throttle times during bursty workload. ClientQuotaManager calculates throttle times by comparing rate (based on how we record Rate) to quota, which I think would result in the same behavior as before unless we change the way we calculate throttle time as well.

dajac: @junrao Thanks for your comment. Your observation is correct. This is because we spill over the amount above the quota into any newly created sample. So @apovzner's example would actually end up like this: [5, 18]. @apovzner I think that you have mentioned once another way to do this that may not require to spill over the remainder.

Your remark regarding the observability is right. This is true as soon as we deviate from a sampled rate, regardless of the implementation that we may choose eventually. Separating concerns is indeed a good idea if we can. Regarding your suggested approach, I think that may works. As @apovzner said, we will need to change the way the throttle time is computed so that it does not rely on the rate but on the amount of credits. Another concern is that the credits won't be observable neither so we may keep the correct rate but still may not understand why one is throttled or not if we can't observe the amount of credits used/left.

If we want to separate concerns, having both a sampled Rate and a Token Bucket working side by side would be the best. The Rate would continue to provide the actual Rate as of today and the Token Bucket would be used to enforced the quota. We could expose the amount of credits in the bucket via a new metric. One way to achieve this would be to have two MeasurableStats within the Sensor: the Rate and the Token Bucket. We would still need to have an adapted version of the Token Bucket that works with our current configs (quota, samples, window). I implemented one to compare with the current approach few weeks ago: https://github.com/dajac/kafka/blob/19e7180105c366d245183644c823135fb2872571/clients/src/main/java/org/apache/kafka/common/metrics/stats/RealTokenBucket.java. This one works within a Rate though.

junrao: @apovzner @dajac : Currently, we calculate the delayed time based on QuotaViolationException.value and QuotaViolationException.bound. I was thinking that we could pass some additional info in QuotaViolationException so that we could calculate the delayed time properly.

Overall, it seems that decoupling the observation from token bucket based quota might be easier to understand. As for monitoring the remaining credit, we could add a separate metric.

Also, it seems that quota only makes sense for rates. So, instead of making quota available on arbitrary measurable, we could just make it work for Rate.

PR #9073, opened 2020-07-24
MINOR: add task ':streams:testAll'

vvcephei:
Adds a new task to Streams to run all the tests for all sub-projects.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1266,6 +1266,27 @@ project(':streams') {
     if( !generatedDocsDir.exists() ) { generatedDocsDir.mkdirs() }
     standardOutput = new File(generatedDocsDir, "streams_config.html").newOutputStream()
   }
+
+  task testAll(
+    dependsOn: [
+            ':streams:test',
+            ':streams:test-utils:test',
+            ':streams:streams-scala:test',
+            ':streams:upgrade-system-tests-0100:test',
+            ':streams:upgrade-system-tests-0101:test',
+            ':streams:upgrade-system-tests-0102:test',
+            ':streams:upgrade-system-tests-0110:test',
+            ':streams:upgrade-system-tests-10:test',
+            ':streams:upgrade-system-tests-11:test',
+            ':streams:upgrade-system-tests-20:test',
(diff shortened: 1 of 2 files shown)


abbccdda (approved):
Ran on local to verify it works.

abbccdda:
One more thought, it might be useful to add this to the repo `README` if a user is only doing stream side changes and want to test it out on local @vvcephei.

vvcephei:
Thanks @abbccdda ! There was only one flaky test, and the last change was only to the README, so I'll merge now.

build.gradle
```diff
@@ -1266,6 +1266,27 @@ project(':streams') {
...
+            ':streams:upgrade-system-tests-0102:test',
+            ':streams:upgrade-system-tests-0110:test',
+            ':streams:upgrade-system-tests-10:test',
+            ':streams:upgrade-system-tests-11:test',
+            ':streams:upgrade-system-tests-20:test',
+            ':streams:upgrade-system-tests-21:test',
+            ':streams:upgrade-system-tests-22:test',
+            ':streams:upgrade-system-tests-23:test',
+            ':streams:upgrade-system-tests-24:test',
+            ':streams:upgrade-system-tests-25:test',
```
mjsax, line 1286: I don't think we have actual unit/integration tests in those packages? Also seems a pain to keep it updates (and I am sure we will forget to update it). Should we just skip those?

mjsax: Oh. Already merged. Never mind.

vvcephei: Thanks for the comment anyway.

I was on the fence about it, but I went with including these projects just to be sure that `testAll` really tests all of the components of `:streams`. Even though they don't have tests, the `:test` target is convenient because it runs all of the compilation targets, as well as spotbugs and checkstyle. The idea is just to flush out any/all possible sources of failure so that we can be pretty sure that if `:streams:testAll` passes, then so will Jenkins.

I agree it's a pain. Maybe we can consider either contributing a script to keep this stuff updated or writing some gradle code to build the list automatically.

PR #9074, opened 2020-07-24
KAFKA-10287: PROPOSAL: safe offset tracking

vvcephei:
This PR should be targeted to trunk, but it's not a trivial task. I just wanted to propose this approach.


vvcephei:
Hey @cadonna , I was reflecting on your fix for KAFKA-10287, and it seems like dealing with those sentinel values is quite risky. We've had several bugs in the past, I don't feel certain that we don't have bugs today, and any refactoring could easily introduce bugs in the future.

What do you think about taking a structural approach to offset sentinel management, like in this proposal?

This wouldn't be for 2.6, but it's targeted at that branch because I built upon your PR.

PR #9075, opened 2020-07-24
KAFKA-10306: GlobalThread should fail on InvalidOffsetException

mjsax:
Fix for `2.6` blocker bug.

Call for review @guozhangwang @vvcephei (\cc @rhauch)

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateMaintainer.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateMaintainer.java
@@ -31,7 +31,7 @@ interface GlobalStateMaintainer {
 
     void flushState();
 
-    void close() throws IOException;
+    void close(final boolean wipeStateStore) throws IOException;
 
     void update(ConsumerRecord<byte[], byte[]> record);
 }
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
@@ -19,7 +19,6 @@ package org.apache.kafka.streams.processor.internals;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
-import org.apache.kafka.clients.consumer.InvalidOffsetException;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.TimeoutException;
@@ -291,27 +290,17 @@ public class GlobalStateManagerImpl implements GlobalStateManager {
             long restoreCount = 0L;
 
             while (offset < highWatermark) {
-                try {
-                    final ConsumerRecords<byte[], byte[]> records = globalConsumer.poll(pollTime);
-                    final List<ConsumerRecord<byte[], byte[]>> restoreRecords = new ArrayList<>();
-                    for (final ConsumerRecord<byte[], byte[]> record : records.records(topicPartition)) {
-                        if (record.key() != null) {
-                            restoreRecords.add(recordConverter.convert(record));
-                        }
+                final ConsumerRecords<byte[], byte[]> records = globalConsumer.poll(pollTime);
+                final List<ConsumerRecord<byte[], byte[]>> restoreRecords = new ArrayList<>();
+                for (final ConsumerRecord<byte[], byte[]> record : records.records(topicPartition)) {
+                    if (record.key() != null) {
+                        restoreRecords.add(recordConverter.convert(record));
                     }
-                    offset = globalConsumer.position(topicPartition);
-                    stateRestoreAdapter.restoreBatch(restoreRecords);
-                    stateRestoreListener.onBatchRestored(topicPartition, storeName, offset, restoreRecords.size());
-                    restoreCount += restoreRecords.size();
-                } catch (final InvalidOffsetException recoverableException) {
-                    log.warn("Restoring GlobalStore {} failed due to: {}. Deleting global store to recreate from scratch.",
-                        storeName,
-                        recoverableException.toString());
-
-                    // TODO K9113: we remove the re-init logic and push it to be handled by the thread directly
-
-                    restoreCount = 0L;
                 }
+                offset = globalConsumer.position(topicPartition);
+                stateRestoreAdapter.restoreBatch(restoreRecords);
+                stateRestoreListener.onBatchRestored(topicPartition, storeName, offset, restoreRecords.size());
+                restoreCount += restoreRecords.size();
             }
             stateRestoreListener.onRestoreEnd(topicPartition, storeName, restoreCount);
             checkpointFileCache.put(topicPartition, offset);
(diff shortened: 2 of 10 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -291,27 +290,17 @@ private void restoreState(final StateRestoreCallback stateRestoreCallback,
...
+                final List<ConsumerRecord<byte[], byte[]>> restoreRecords = new ArrayList<>();
+                for (final ConsumerRecord<byte[], byte[]> record : records.records(topicPartition)) {
+                    if (record.key() != null) {
+                        restoreRecords.add(recordConverter.convert(record));
                     }
-                    offset = globalConsumer.position(topicPartition);
-                    stateRestoreAdapter.restoreBatch(restoreRecords);
-                    stateRestoreListener.onBatchRestored(topicPartition, storeName, offset, restoreRecords.size());
-                    restoreCount += restoreRecords.size();
-                } catch (final InvalidOffsetException recoverableException) {
```
mjsax, line 306: This is the actual bug: we swallow the exception. However, because we don't do any "seek", we just hit the same exception in `poll()` over and over and never recover but loop forever.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateMaintainer.java
```diff
@@ -31,7 +31,7 @@
 
     void flushState();
 
-    void close() throws IOException;
+    void close(final boolean wipeStateStore) throws IOException;
```
mjsax, line 34: We could also not wipe out the store and let the user do it manually. A manual cleanup is actually required nowadays, thus, it's actually a small side "improvement".

Note that we wipe out the whole global task dir here -- in contrast, users could do a manual per-store wipe out... But as we do the same coarse grained wipe out for all tasks and we have already a ticket for "per store cleanup" I though it would be ok for now.

Let me know what you think.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStreamThread.java
```diff
@@ -234,24 +234,18 @@ void initialize() {
...
-                final ConsumerRecords<byte[], byte[]> received = globalConsumer.poll(pollTime);
-                for (final ConsumerRecord<byte[], byte[]> record : received) {
-                    stateMaintainer.update(record);
-                }
-                final long now = time.milliseconds();
-                if (now >= lastFlush + flushInterval) {
-                    stateMaintainer.flushState();
-                    lastFlush = now;
-                }
-            } catch (final InvalidOffsetException recoverableException) {
```
mjsax, line 247: We just let the original exception bubble up, to be able to wipe out the store. -- This is also just a side "improvement"; we could also just die and let users cleanup the state directory manually. However, it seems better to wipe it out directly.

streams/src/test/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImplTest.java
```diff
@@ -323,21 +322,6 @@ public void shouldRestoreRecordsUpToHighwatermark() {
         assertEquals(2, stateRestoreCallback.restored.size());
     }
 
-    @Test
-    public void shouldRecoverFromInvalidOffsetExceptionAndRestoreRecords() {
```
mjsax, line 327: This test was broken: it throw the `InvalidOffsetException` only once and thus the second `poll()` succeeds -- however, this is not how a real consumer behalves.

streams/src/test/java/org/apache/kafka/streams/processor/internals/GlobalStreamThreadTest.java
```diff
@@ -128,10 +130,9 @@ public void shouldThrowStreamsExceptionOnStartupIfThereIsAStreamsException() {
         assertFalse(globalStreamThread.stillRunning());
     }
 
-    @SuppressWarnings("unchecked")
```
mjsax, line 131: Just a little side cleanup :)

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateMaintainer.java
vvcephei: Thanks @mjsax , this sounds perfect to me.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateUpdateTask.java
```diff
@@ -114,8 +118,16 @@ public void flushState() {
...
 
-    public void close() throws IOException {
+    public void close(final boolean wipeStateStore) throws IOException {
         stateMgr.close();
+        if (wipeStateStore) {
+            try {
+                log.error("Wiping state stores for global task.");
+                Utils.delete(stateMgr.baseDir());
+            } catch (final IOException e) {
+                log.error("Failed to wiping state stores for global task.", e);
```
vvcephei:
```suggestion
                log.error("Failed to delete global task directory after detecting corruption.", e);
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateUpdateTask.java
```diff
@@ -114,8 +118,16 @@ public void flushState() {
         stateMgr.checkpoint(offsets);
     }
 
-    public void close() throws IOException {
+    public void close(final boolean wipeStateStore) throws IOException {
         stateMgr.close();
+        if (wipeStateStore) {
+            try {
+                log.error("Wiping state stores for global task.");
```
vvcephei: This doesn't seem to be an error. Maybe info would be better? Also, I think "wipe state stores" might be confusing for a user looking at the log messages with no context. "Deleting the task directory" seems to be a more context-free statement of what we're doing.

```suggestion
                log.info("Deleting global task directory after detecting corruption.");
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStreamThread.java
```diff
@@ -284,10 +278,21 @@ public void run() {
...
 
+        boolean wipeStateStore = false;
         try {
             while (stillRunning()) {
                 stateConsumer.pollAndUpdate();
             }
+        } catch (final InvalidOffsetException recoverableException) {
+            wipeStateStore = true;
+            log.error(
+                "Updating global state failed. You can restart KafkaStreams to recover from this error.",
```
vvcephei:
```suggestion
                "Updating global state failed due to inconsistent local state. Will attempt to clean up the local state. You can restart KafkaStreams to recover from this error.",
```

Just a thought to indicate why just restarting would recover anything.

vvcephei (commented):
Thanks for the PR, @mjsax ! Just a few comments.

vvcephei:
Otherwise, it LGTM. Feel free to merge whether or not you like my suggestions ;)

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStreamThread.java
```diff
@@ -331,17 +336,36 @@ private StateConsumer initialize() {
...
                 Duration.ofMillis(config.getLong(StreamsConfig.POLL_MS_CONFIG)),
                 config.getLong(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG)
             );
-            stateConsumer.initialize();
+
+            try {
+                stateConsumer.initialize();
+            } catch (final InvalidOffsetException recoverableException) {
+                log.error(
+                    "Bootstrapping global state failed. You can restart KafkaStreams to recover from this error.",
```
mjsax:
```suggestion
                    "Bootstrapping global state failed due to inconsistent local state. Will attempt to clean up the local state. You can restart KafkaStreams to recover from this error.",
```

vvcephei (approved):
Thanks, @mjsax !

mjsax:
Retest this please.

mjsax:
Retest this please.

mjsax:
Merged to `trunk` and cherry-picked to `2.6` branch.

PR #9076, opened 2020-07-25
MINOR - fix typo

wangxianghu:
*FIx typo in ProducerConfig*


rondagostino (commented):
I think it was grammatically correct as it stood before this change.  This change, if it were to occur, would have to state `if the leader fails` (the current change says `if the leader fail`, which is incorrect grammar).  So I believe this is a choice between two grammatically correct possibilities.  I personally don't see the need for a change.

wangxianghu:
> I think it was grammatically correct as it stood before this change. This change, if it were to occur, would have to state `if the leader fails` (the current change says `if the leader fail`, which is incorrect grammar). So I believe this is a choice between two grammatically correct possibilities. I personally don't see the need for a change.

Reasonable

PR #9078, opened 2020-07-25
KAFKA-10132: Return correct value types for MBean attributes

rgroothuijsen:
Currently, JMX outputs all metrics as having type `double`, even if they are strings or other types of numbers. This PR gets the type from the metric's value if possible, using `double` as a fallback if the type can't be determined.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/metrics/JmxReporter.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/JmxReporter.java
@@ -272,7 +272,14 @@ public class JmxReporter implements MetricsReporter {
             for (Map.Entry<String, KafkaMetric> entry : this.metrics.entrySet()) {
                 String attribute = entry.getKey();
                 KafkaMetric metric = entry.getValue();
-                String metricType = metric.metricValue() != null ? metric.metricValue().getClass().getName() : null;
+                String metricType = null;
+
+                try {
+                    metricType = metric.metricValue().getClass().getName();
+                } catch (NullPointerException e) {
+                    log.debug("Could not determine metric value type for attribute {}", attribute);
+                }
+
                 attrs[i] = new MBeanAttributeInfo(attribute,
                                                   metricType,
                                                   metric.metricName().description(),
--- a/clients/src/main/java/org/apache/kafka/common/metrics/JmxReporter.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/JmxReporter.java
@@ -272,7 +272,7 @@ public class JmxReporter implements MetricsReporter {
             for (Map.Entry<String, KafkaMetric> entry : this.metrics.entrySet()) {
                 String attribute = entry.getKey();
                 KafkaMetric metric = entry.getValue();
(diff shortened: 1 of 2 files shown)


rgroothuijsen:
I'm not entirely sure about returning raw nulls as a fallback, however, or how permissive it should be about null values in general. Thoughts?

clients/src/main/java/org/apache/kafka/common/metrics/JmxReporter.java
```diff
@@ -272,8 +272,16 @@ public MBeanInfo getMBeanInfo() {
             for (Map.Entry<String, KafkaMetric> entry : this.metrics.entrySet()) {
                 String attribute = entry.getKey();
                 KafkaMetric metric = entry.getValue();
+                String metricType = double.class.getName();
+
+                try {
+                    metricType = metric.metricValue().getClass().getName();
+                } catch (NullPointerException e) {
```
abbccdda, line 279: It is weird to catch NPE, could we just check whether `metricValue` is defined?

rgroothuijsen: @abbccdda That's what I figured as well at first, and it passes in unit tests, but there's a strange side effect. Upon starting `connect-distributed`, a whole bunch of NPEs appear for various metrics. It looks like when I call `metricValue()`, it eventually arrives [here](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L1811), and throws an error because `member` is null. I could also do a null check there, though I'm not sure what value the method would return in that case.

jeqo: @rgroothuijsen this could be a bug on the DistributedHerder side. Seems that `DistributedHerder#herderMetrics` is initialized [too early](https://github.com/apache/kafka/blob/8e211eb72f9a45897cc37fed394a38096aa47feb/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L216). Could you give a try moving it later in the constructor, [maybe after config](https://github.com/apache/kafka/blob/8e211eb72f9a45897cc37fed394a38096aa47feb/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L249), to check if the same exception happens again?

rgroothuijsen: @jeqo Thanks, moving the initialization did make a difference, specifically after moving it past [this point](https://github.com/apache/kafka/blob/8e211eb72f9a45897cc37fed394a38096aa47feb/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L232).

This did lead to an opposite problem, though, as now other exceptions start appearing. From what I can tell, `WorkerGroupMember` and `HerderMetrics` are dependent on each other, so I'm getting errors in one or the other when I change the ordering around.

PR #9083, opened 2020-07-27
KAFKA-9450: Follow-up; Forbid process after closed

guozhangwang:
A few cleanup and tighten screws:

* When a processor is closed (due to topology-closure), we should not allow processing more records.
* Let all built-in processors to extend from AbstractProcessor.
* Remove duplicated override functions.


guozhangwang:
cc @ableegoldman @mjsax

guozhangwang:
@ableegoldman @abbccdda This is a minor cleanup PR that I found while working on KAFKA-9450.

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
```diff
@@ -37,6 +37,7 @@
 
     /**
      * If the checkpoint has not been loaded from the file yet (null), then we should not overwrite the checkpoint;
+     * If the checkpoint has been loaded form the file and has never been re-written (empty map), then we should re-write the checkpoint;
```
ableegoldman:
```suggestion
     * If the checkpoint has been loaded from the file and has never been re-written (empty map), then we should re-write the checkpoint;
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextImpl.java
```diff
@@ -217,7 +217,7 @@ public StateStore getStateStore(final String name) {
                     forward((ProcessorNode<K, V, ?, ?>) child, key, value);
                 }
             } else {
-                final ProcessorNode<K, V, ?, ?> child = currentNode().getChild(sendTo);
+                final ProcessorNode child = currentNode().getChild(sendTo);
```
ableegoldman: Rebasing error?

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorNode.java
```diff
@@ -149,10 +155,20 @@ public void close() {
         } catch (final Exception e) {
```
ableegoldman, line 160: Should we `throwIfClosed()` at the beginning of `close()` too?

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -1310,9 +1310,11 @@ public void shouldNotCheckpointOffsetsAgainOnCommitIfSnapshotNotChangedMuch() {
         EasyMock.expect(recordCollector.offsets()).andReturn(Collections.singletonMap(changelogPartition, offset)).anyTimes();
         stateManager.checkpoint();
         EasyMock.expectLastCall().once();
+        EasyMock.expect(stateManager.changelogPartitions()).andReturn(Collections.singleton(changelogPartition));
```
ableegoldman: Why does this need to change?

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorNode.java
```diff
@@ -118,6 +120,10 @@ public void init(final InternalProcessorContext context) {
...
             throw new StreamsException(String.format("failed to initialize processor %s", name), e);
         }
+
+        // revived tasks could re-initialize the topology,
+        // in which case we should reset the flag
+        closed = false;
```
ableegoldman, line 129: Should we assert that it's in `closed` when `init` is called?

ableegoldman (commented):
Thanks for finishing this up, some minor suggestions but overall LGTM

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextImpl.java
guozhangwang: This is because I added the typing in line 98 of ProcessorNode below, I will revert it back and update the line below in `forward`.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorNode.java
guozhangwang: SG.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
guozhangwang: My bad, will revert.

guozhangwang:
cc @abbccdda for a final review.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextImpl.java
```diff
@@ -217,12 +217,12 @@ public StateStore getStateStore(final String name) {
                     forward((ProcessorNode<K, V, ?, ?>) child, key, value);
                 }
             } else {
-                final ProcessorNode<K, V, ?, ?> child = currentNode().getChild(sendTo);
```
ableegoldman, line 220: Why change the `K, V` to `?, ?` -- and if we don't do that, do we still need the cast in the `forward` call down below?

ableegoldman (commented):
> Execution failed for task ':streams:checkstyleTest'.

Just FYI

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -1566,6 +1566,7 @@ public void shouldCheckpointForSuspendedTask() {
         EasyMock.verify(stateManager);
     }
 
+
```
abbccdda: nit: not necessary

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextImpl.java
abbccdda: +1

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamFlatTransformValues.java
```diff
@@ -44,19 +45,18 @@ public KStreamFlatTransformValues(final ValueTransformerWithKeySupplier<KIn, VIn
         return valueTransformerSupplier.stores();
     }
 
-    public static class KStreamFlatTransformValuesProcessor<KIn, VIn, VOut> implements Processor<KIn, VIn> {
+    public static class KStreamFlatTransformValuesProcessor<KIn, VIn, VOut> extends AbstractProcessor<KIn, VIn> {
```
abbccdda, line 48: Could you elaborate why this is better than Processor?

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextImpl.java
guozhangwang: I made the change in ProcessorNode to add back the template types: https://github.com/apache/kafka/pull/9083/files/82b6f6f5d238401097e0906c8135c5c189524666#diff-705bfd0ed3f214048b76d775708cc7d2L96

But since `currentNode()`'s template is `<?, ?, ?, ?>` its templated `getChild` and that's why I need to weaken it here --- as you can see from the above `if` branch, it now aligns consistently on the typing.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamFlatTransformValues.java
guozhangwang: As described in at the top, `Let all built-in processors to extend from AbstractProcessor.` The main reason is that AbstractProcessor provides some basic functionalities and hence it's better to let our own impl to base on them.

abbccdda (approved):
LGTM

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextImpl.java
```diff
@@ -217,12 +217,12 @@ public StateStore getStateStore(final String name) {
...
+                final ProcessorNode<?, ?, ?, ?> child = currentNode().getChild(sendTo);
                 if (child == null) {
                     throw new StreamsException("Unknown downstream node: " + sendTo
                         + " either does not exist or is not connected to this processor.");
                 }
-                forward(child, key, value);
+                forward((ProcessorNode<K, V, ?, ?>) child, key, value);
```
mjsax, line 225: @guozhangwang It's weird to remove the types above and add a cast here. Seems like a step backward with regard to type safety. Not sure if @vvcephei's changes would fix it?

PR #9084, opened 2020-07-27
MINOR: Preserve Kafka exception from RebalanceListener

guozhangwang:
Some of the rebalance listener may be implemented by Kafka as well, e.g. Connect and Streams, and if the exception thrown is actually a KafkaException, then we should not wrap it but directly throw the exception from the listener.

Unit tests to be added; cc @abbccdda to review.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java
@@ -728,7 +728,11 @@ public final class ConsumerCoordinator extends AbstractCoordinator {
         subscriptions.resetGroupSubscription();
 
         if (exception != null) {
-            throw new KafkaException("User rebalance callback throws an error", exception);
+            if (exception instanceof KafkaException) {
+                throw (KafkaException) exception;
+            } else {
+                throw new KafkaException("User rebalance callback throws an error", exception);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java
```diff
@@ -728,7 +728,11 @@ protected void onJoinPrepare(int generation, String memberId) {
         subscriptions.resetGroupSubscription();
 
         if (exception != null) {
-            throw new KafkaException("User rebalance callback throws an error", exception);
+            if (exception instanceof KafkaException) {
+                throw (KafkaException) exception;
```
abbccdda, line 732: Why do we need this cast?

abbccdda (commented):
I'm not fully convinced that this is necessary, either way the thrown exception would be of type `KafkaException` and no information gets lost with the current behavior.

guozhangwang:
@abbccdda The exception is thrown from the consumer.poll, and the caller of the consumer could be expecting specific exceptions, e.g. TaskMigratedException will be handled specifically in Streams, but a general KafkaException would be treated as a fatal error.

We're already doing this e.g. in ConsumerCoordinator line 432, but it is not done in all occasions, so I'm just trying to make the behavior to be consistent across all callers here.

guozhangwang:
@abbccdda @cadonna ping for another review again. I think it would be related to our exception categorization work.

guozhangwang:
test this

PR #8530, opened 2020-04-22
KAFKA-9388: Refactor integration tests to always use different application ids

guozhangwang:
When debugging KAFKA-9388, I found the reason that the second test method test takes much longer (10s) than the previous one (~500ms) is because they used the same app.id. When the previous clients are shutdown, they would not send leave-group and hence we are still depending on the session timeout (10s) for the members to be removed out of the group.

When the second test is triggered, they will join the same group because of the same application id, and the `prepare-rebalance` phase would would for the full rebalance timeout before it kicks out the previous members.

Setting different application ids could resolve such issues for integration tests --- I did a quick search and found some other integration tests have the same issue. And after this PR my local unit test runtime reduced from about 14min to 7min.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9087, opened 2020-07-27
HOTFIX: Set session timeout and heartbeat interval to default to decrease flakiness

cadonna:

### DIFF
--- a/streams/src/test/java/org/apache/kafka/streams/integration/RestoreIntegrationTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/integration/RestoreIntegrationTest.java
@@ -117,8 +117,6 @@ public class RestoreIntegrationTest {
         streamsConfiguration.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.Integer().getClass());
         streamsConfiguration.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, 1000);
         streamsConfiguration.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
-        streamsConfiguration.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, 1000);
-        streamsConfiguration.put(ConsumerConfig.HEARTBEAT_INTERVAL_MS_CONFIG, 300);
         return streamsConfiguration;
     }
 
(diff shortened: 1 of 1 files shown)


cadonna:
Call for review: @guozhangwang @ableegoldman

ableegoldman (approved):
Thanks @cadonna , nice find

abbccdda (commented):
Instead of removing the set config value in the test, what about we just do the config change on demand so that we don't need to remove them in the first place?

cadonna:
@abbccdda Looking at this test class again, I even think that we could set session timeout and heartbeat interval to default for all tests. The main motivation to reduce them was to not need to change the application ID for each test (see https://github.com/apache/kafka/pull/8530#discussion_r413170584). However, in the meanwhile each test has its own application ID, so I guess it would be fine to use again the defaults here. Is this correct @guozhangwang ?

guozhangwang:
For correctness I agree; though for some tests we cover the rebalance triggered by member dropping out of the group and hence reducing the session timeout also largely reduces the overall run time of the test case, so I'd say for testing runtime we may still want to check case-by-case.

cadonna:
I did some runs of the `RestoreIntegrationTest` and the runtime does not significantly change between runs with default values for session timeout and heartbeat interval and reduced values for session timeout and heartbeat interval. Hence, I set the two configs to their default values for the entire test.

guozhangwang:
test this please

guozhangwang:
LGTM. @abbccdda could help merging after green builds.

cadonna:
Java 11 failed due to the following failing test:

```
kafka.api.ConsumerBounceTest.testClose
```

Java 8 failed due to the following failing tests:

```
org.apache.kafka.connect.mirror.MirrorConnectorsIntegrationTest.testReplication
org.apache.kafka.connect.mirror.MirrorConnectorsIntegrationTest.testOneWayReplicationWithAutorOffsetSync1
```

cadonna:
@chia7712 Would you be so kind and trigger jenkins on this PR, please?

chia7712:
retest this please

cadonna:
Java 14 failed due to the following failing test:

```
org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest > shouldUpgradeFromEosAlphaToEosBeta[true]
```

chia7712:
```EosBetaUpgradeIntegrationTest``` pass on my local. retest this please

abbccdda (approved):
LGTM

PR #9090, opened 2020-07-28
toString pattern change like ProducerRecord.java

codelabor:
In most of 'toString()' methods,'=' is used instead of ' = ', so I followed the convention.
Actually, if you print 'ProducerRecord' and 'ConsumerRecord', the pattern looks different and looks strange.
---------------------------------
producerRecord: ProducerRecord(topic=xxxx, partition=null, headers= ...)
consumerRecord: ConsumerRecord(topic = xxxx, partition = 0, ..., headers = ...)


abbccdda (commented):
I'm actually neutral to this change, maybe @hachikuji could provide some thoughts on whether we should consolidate and which pattern we should keep, either consumer or producer.

PR #9092, opened 2020-07-28
KAFKA-10163; Define `controller_mutation_rate` as a Double instead of a Long

dajac:
First tests have shown that `controller_mutation_rate` can be quite low (e.g. around 1) in clusters with multiple tenants. At the moment, the rate is defined as a Long which limits the possible low values. Using a Double seems more appropriate.

### DIFF
--- a/core/src/main/scala/kafka/server/DynamicConfig.scala
+++ b/core/src/main/scala/kafka/server/DynamicConfig.scala
@@ -103,7 +103,7 @@ object DynamicConfig {
       .define(ProducerByteRateOverrideProp, LONG, DefaultProducerOverride, MEDIUM, ProducerOverrideDoc)
       .define(ConsumerByteRateOverrideProp, LONG, DefaultConsumerOverride, MEDIUM, ConsumerOverrideDoc)
       .define(RequestPercentageOverrideProp, DOUBLE, DefaultRequestOverride, MEDIUM, RequestOverrideDoc)
-      .define(ControllerMutationOverrideProp, LONG, DefaultConsumerOverride, MEDIUM, ControllerMutationOverrideDoc)
+      .define(ControllerMutationOverrideProp, DOUBLE, DefaultConsumerOverride, MEDIUM, ControllerMutationOverrideDoc)
 
     def configKeys = clientConfigs.configKeys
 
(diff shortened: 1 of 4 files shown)


rajinisivaram:
ok to test

rajinisivaram (commented):
@dajac Thanks for the PR, looks good. Perhaps worth adding a test with quotas that are fractions like 0.5 to make sure they don't get accidentally converted to long somewhere and end up becoming zero.

dajac:
@rajinisivaram I have updated an existing test to use 1.5 instead of 2 to verify this. https://github.com/apache/kafka/pull/9092/files#diff-296d7b93103356535b8b891f019e4651R113.

core/src/main/scala/kafka/server/DynamicConfig.scala
```diff
@@ -103,7 +103,7 @@ object DynamicConfig {
       .define(ProducerByteRateOverrideProp, LONG, DefaultProducerOverride, MEDIUM, ProducerOverrideDoc)
       .define(ConsumerByteRateOverrideProp, LONG, DefaultConsumerOverride, MEDIUM, ConsumerOverrideDoc)
       .define(RequestPercentageOverrideProp, DOUBLE, DefaultRequestOverride, MEDIUM, RequestOverrideDoc)
-      .define(ControllerMutationOverrideProp, LONG, DefaultConsumerOverride, MEDIUM, ControllerMutationOverrideDoc)
+      .define(ControllerMutationOverrideProp, DOUBLE, DefaultConsumerOverride, MEDIUM, ControllerMutationOverrideDoc)
```
chia7712: why not changing the type of default value from ```Long``` to ```Double```?

dajac: good point. i have changed this to use `Int.MaxValue.toDouble` by default like we do for the request quota.

rajinisivaram (approved):
@dajac Thanks for the PR, LGTM

rajinisivaram:
ok to test

rajinisivaram:
retest this please

rajinisivaram:
ok to test

rajinisivaram:
@dajac Thanks for the PR, test failures not related, merging to trunk.

PR #9094, opened 2020-07-29
KAFKA-10054: KIP-613, add TRACE-level e2e latency metrics

ableegoldman:
Adds avg, min, and max e2e latency metrics at the new TRACE level. Also adds the missing `avg` task-level metric at the INFO level.

I think where we left off with the KIP, the TRACE-level metrics were still defined to be "stateful-processor-level". I realized this doesn't really make sense and would be pretty much impossible to define given the DFS processing approach of Streams, and felt that store-level metrics made more sense to begin with. I haven't updated the KIP yet so I could get some initial feedback on this

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/admin/AdminClientConfig.java
+++ b/clients/src/main/java/org/apache/kafka/clients/admin/AdminClientConfig.java
@@ -198,7 +198,7 @@ public class AdminClientConfig extends AbstractConfig {
                                 .define(METRICS_RECORDING_LEVEL_CONFIG,
                                         Type.STRING,
                                         Sensor.RecordingLevel.INFO.toString(),
-                                        in(Sensor.RecordingLevel.INFO.toString(), Sensor.RecordingLevel.DEBUG.toString()),
+                                        in(Sensor.RecordingLevel.INFO.toString(), Sensor.RecordingLevel.DEBUG.toString(), Sensor.RecordingLevel.TRACE.toString()),
                                         Importance.LOW,
                                         METRICS_RECORDING_LEVEL_DOC)
                                 .define(CLIENT_DNS_LOOKUP_CONFIG,
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
@@ -24,7 +24,6 @@ import org.apache.kafka.common.KafkaException;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.common.metrics.Sensor;
-import org.apache.kafka.common.metrics.Sensor.RecordingLevel;
 import org.apache.kafka.common.utils.LogContext;
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.streams.StreamsConfig;
@@ -151,7 +150,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
         for (final String terminalNodeName : topology.terminalNodes()) {
             e2eLatencySensors.put(
                 terminalNodeName,
-                TaskMetrics.e2ELatencySensor(threadId, taskId, terminalNodeName, RecordingLevel.INFO, streamsMetrics)
+                TaskMetrics.e2ELatencySensor(threadId, taskId, terminalNodeName, streamsMetrics)
             );
         }
 
@@ -159,7 +158,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
             final String sourceNodeName = sourceNode.name();
             e2eLatencySensors.put(
                 sourceNodeName,
-                TaskMetrics.e2ELatencySensor(threadId, taskId, sourceNodeName, RecordingLevel.INFO, streamsMetrics)
+                TaskMetrics.e2ELatencySensor(threadId, taskId, sourceNodeName, streamsMetrics)
             );
         }
 
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
@@ -36,6 +36,7 @@ import org.apache.kafka.streams.processor.PunctuationType;
 import org.apache.kafka.streams.processor.Punctuator;
 import org.apache.kafka.streams.processor.TaskId;
 import org.apache.kafka.streams.processor.TimestampExtractor;
+import org.apache.kafka.streams.processor.internals.metrics.ProcessorNodeMetrics;
 import org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl;
 import org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.Version;
 import org.apache.kafka.streams.processor.internals.metrics.TaskMetrics;
@@ -150,7 +151,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
         for (final String terminalNodeName : topology.terminalNodes()) {
             e2eLatencySensors.put(
                 terminalNodeName,
-                TaskMetrics.e2ELatencySensor(threadId, taskId, terminalNodeName, streamsMetrics)
+                ProcessorNodeMetrics.e2ELatencySensor(threadId, taskId, terminalNodeName, streamsMetrics)
             );
         }
 
@@ -158,7 +159,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
             final String sourceNodeName = sourceNode.name();
             e2eLatencySensors.put(
                 sourceNodeName,
-                TaskMetrics.e2ELatencySensor(threadId, taskId, sourceNodeName, streamsMetrics)
+                ProcessorNodeMetrics.e2ELatencySensor(threadId, taskId, sourceNodeName, streamsMetrics)
             );
         }
 
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
@@ -478,14 +478,14 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
     @Override
     public void closeClean() {
         validateClean();
-        streamsMetrics.removeAllTaskLevelSensors(Thread.currentThread().getName(), id.toString());
+        removeAllSensors();
         close(true);
         log.info("Closed clean");
     }
 
     @Override
     public void closeDirty() {
-        streamsMetrics.removeAllTaskLevelSensors(Thread.currentThread().getName(), id.toString());
+        removeAllSensors();
         close(false);
         log.info("Closed dirty");
     }
@@ -499,7 +499,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
     @Override
     public void closeCleanAndRecycleState() {
         validateClean();
-        streamsMetrics.removeAllTaskLevelSensors(Thread.currentThread().getName(), id.toString());
+        removeAllSensors();
         switch (state()) {
             case SUSPENDED:
                 stateMgr.recycle();
@@ -548,6 +548,13 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
             log.debug("Tried to close clean but there was pending uncommitted data, this means we failed to"
                           + " commit and should close as dirty instead");
             throw new TaskMigratedException("Tried to close dirty task as clean");
+        }
+    }
+
+    private void removeAllSensors() {
+        streamsMetrics.removeAllTaskLevelSensors(Thread.currentThread().getName(), id.toString());
+        for (final String nodeName : e2eLatencySensors.keySet()) {
+            streamsMetrics.removeAllNodeLevelSensors(Thread.currentThread().getName(), id.toString(), nodeName);
         }
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/metrics/ProcessorNodeMetrics.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/metrics/ProcessorNodeMetrics.java
@@ -299,7 +299,7 @@ public class ProcessorNodeMetrics {
                                           final String processorNodeId,
                                           final StreamsMetricsImpl streamsMetrics) {
         final String sensorName = processorNodeId + "-" + RECORD_E2E_LATENCY;
-        final Sensor sensor = streamsMetrics.taskLevelSensor(threadId, taskId, sensorName, RecordingLevel.INFO);
+        final Sensor sensor = streamsMetrics.nodeLevelSensor(threadId, taskId, processorNodeId, sensorName, RecordingLevel.INFO);
         final Map<String, String> tagMap = streamsMetrics.nodeLevelTagMap(threadId, taskId, processorNodeId);
         addAvgAndMinAndMaxToSensor(
             sensor,
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/metrics/ProcessorNodeMetrics.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/metrics/ProcessorNodeMetrics.java
@@ -24,10 +24,15 @@ import java.util.Map;
 
 import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.LATENCY_SUFFIX;
 import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.PROCESSOR_NODE_LEVEL_GROUP;
+import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.RECORD_E2E_LATENCY;
+import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.RECORD_E2E_LATENCY_AVG_DESCRIPTION;
+import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.RECORD_E2E_LATENCY_MAX_DESCRIPTION;
+import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.RECORD_E2E_LATENCY_MIN_DESCRIPTION;
 import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.ROLLUP_VALUE;
 import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.TASK_LEVEL_GROUP;
 import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.TOTAL_DESCRIPTION;
 import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.addAvgAndMaxToSensor;
+import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.addAvgAndMinAndMaxToSensor;
 import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.addInvocationRateAndCountToSensor;
 
 public class ProcessorNodeMetrics {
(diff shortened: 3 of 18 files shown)


clients/src/main/java/org/apache/kafka/clients/admin/AdminClientConfig.java
```diff
@@ -198,7 +198,7 @@
                                 .define(METRICS_RECORDING_LEVEL_CONFIG,
                                         Type.STRING,
                                         Sensor.RecordingLevel.INFO.toString(),
-                                        in(Sensor.RecordingLevel.INFO.toString(), Sensor.RecordingLevel.DEBUG.toString()),
+                                        in(Sensor.RecordingLevel.INFO.toString(), Sensor.RecordingLevel.DEBUG.toString(), Sensor.RecordingLevel.TRACE.toString()),
```
ableegoldman, line 201: It's kind of a bummer that we can't just add the new TRACE level for Streams only; we have to add it to all the clients that Streams passes its configs down to. We could check for the new TRACE level and strip it off before passing the configs on to the clients, but that just seems like asking for trouble.

streams/src/main/java/org/apache/kafka/streams/processor/internals/metrics/TaskMetrics.java
```diff
@@ -88,13 +92,6 @@ private TaskMetrics() {}
     private static final String NUM_BUFFERED_RECORDS_DESCRIPTION = "The count of buffered records that are polled " +
         "from consumer and not yet processed for this active task";
 
-    private static final String RECORD_E2E_LATENCY = "record-e2e-latency";
```
ableegoldman, line 91: Moved the common descriptions to StreamsMetricsImpl

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -227,6 +234,14 @@ protected Bytes keyBytes(final K key) {
         return byteEntries;
     }
 
+    private void maybeRecordE2ELatency() {
+        if (e2eLatencySensor.shouldRecord() && e2eLatencySensor.hasMetrics()) {
```
ableegoldman: For KV stores, we just compare the current time with the current record's timestamp

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredSessionStore.java
```diff
@@ -248,4 +253,12 @@ public void close() {
     private Bytes keyBytes(final K key) {
         return Bytes.wrap(serdes.rawKey(key));
     }
+
+    private void maybeRecordE2ELatency() {
+        if (e2eLatencySensor.shouldRecord() && e2eLatencySensor.hasMetrics()) {
```
ableegoldman: For session and window stores, we also just compare the current time with the current record's timestamp when `put` is called. This can mean the e2e latency is measured several times on the same record, for example in a windowed aggregation.
At first I thought that didn't make sense, but now I think it's actually exactly what we want. First of all, it means we can actually account for the latency between calls to `put` within a processor. For simple point inserts this might not be a huge increase on the scale of ms, but more complex processing may benefit from seeing this granularity of information. If they don't want it, well, that's why we introduced `TRACE`

Second, while it might seem like we're over-weighting some records by measuring the e2e latency on them more than others, I'm starting to think this actually makes more sense than not: the big picture benefit/use case for the e2e latency metric is less "how long for this record to get sent downstream" and more "how long for this record to be reflected in the state store/IQ results". Given that, each record should be weighted by its actual proportion of the state store. You aren't querying individual records (in a window store), you're querying the windows themselves

I toyed around with the idea of measuring the e2e latency relative to the window time, instead of the record timestamp, but ultimately couldn't find any sense in that. 
Thoughts?

streams/src/test/java/org/apache/kafka/streams/integration/MetricsIntegrationTest.java
```diff
@@ -668,6 +671,9 @@ private void checkKeyValueStoreMetrics(final String group0100To24,
         checkMetricByName(listMetricStore, SUPPRESSION_BUFFER_SIZE_CURRENT, 0);
         checkMetricByName(listMetricStore, SUPPRESSION_BUFFER_SIZE_AVG, 0);
         checkMetricByName(listMetricStore, SUPPRESSION_BUFFER_SIZE_MAX, 0);
+        checkMetricByName(listMetricStore, RECORD_E2E_LATENCY_AVG, expectedNumberofE2ELatencyMetrics);
+        checkMetricByName(listMetricStore, RECORD_E2E_LATENCY_MIN, expectedNumberofE2ELatencyMetrics);
+        checkMetricByName(listMetricStore, RECORD_E2E_LATENCY_MAX, expectedNumberofE2ELatencyMetrics);
```
ableegoldman: Ok there's something I'm not understanding about this test and/or the built-in metrics version. For some reason, the KV-store metrics are 0 when `METRICS_0100_TO_24` is used, and 1 (as expected) when the latest version in used. I feel like this is wrong, and it should always be 1, but I need some clarify on how this config is supposed to be used
What makes me pretty sure there's something actually wrong here is that for the window/session store metrics, they are actually always at 1. But I can't figure out why the KV store metrics would be any different than the others. Any ideas @cadonna ?

ableegoldman:
call for review @vvcephei @cadonna @guozhangwang

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
cadonna: I think, you do not need to check for metrics with `e2eLatencySensor.hasMetrics()`. There should always be metrics within this sensor.  
`hasMetrics()` is used in `StreamsMetricsImpl#maybeMeasureLatency()` because some sensors may not contain any metrics due to the built-in metrics version. For instance, the destroy sensor exists for built-in metrics version 0.10.0-2.4 but not for latest. To avoid version checks in the record processing code, we just create an empty sensor and call record on it effectively not recording any metrics for this sensor for version latest.
We do not hide newly added metrics if the built-in version is set to an older version.
Same applies to the other uses of `hasMetrics()` introduced in this PR.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredSessionStore.java
cadonna: Your approach makes sense to me. I agree that the latency should refer to the update in the state store and not to record itself. If a record updates the state more than once then latency should be measured each time.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -468,38 +468,44 @@ public void shouldRecordE2ELatencyOnProcessForSourceNodes() {
...
+        final Metric avgMetric = getProcessorMetric("record-e2e-latency", "%s-avg", task.id().toString(), sourceNode, StreamsConfig.METRICS_LATEST);
         final Metric minMetric = getProcessorMetric("record-e2e-latency", "%s-min", task.id().toString(), sourceNode, StreamsConfig.METRICS_LATEST);
+        final Metric maxMetric = getProcessorMetric("record-e2e-latency", "%s-max", task.id().toString(), sourceNode, StreamsConfig.METRICS_LATEST);
 
+        assertThat(avgMetric.metricValue(), equalTo(Double.NaN));
         assertThat(minMetric.metricValue(), equalTo(Double.NaN));
         assertThat(maxMetric.metricValue(), equalTo(Double.NaN));
 
         // e2e latency = 10
         task.maybeRecordE2ELatency(0L, 10L, sourceNode);
```
cadonna: Could you test `maybeRecordE2ELatency()` through `process()` and `forward()`? Although you test `maybeRecordE2ELatency()`, you do not test if the recording is done during processing, but that is the crucial thing, IMO.

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetrics.java
```diff
@@ -443,6 +447,25 @@ public static Sensor suppressionBufferSizeSensor(final String threadId,
...
+    public static Sensor e2ELatencySensor(final String threadId,
+                                          final String taskId,
+                                          final String storeType,
+                                          final String storeName,
+                                          final StreamsMetricsImpl streamsMetrics) {
+        final Sensor sensor = streamsMetrics.storeLevelSensor(threadId, taskId, storeName, RECORD_E2E_LATENCY, RecordingLevel.TRACE);
+        final Map<String, String> tagMap = streamsMetrics.storeLevelTagMap(threadId, taskId, storeType, storeName);
+        addAvgAndMinAndMaxToSensor(
+            sensor,
+            STATE_STORE_LEVEL_GROUP,
```
cadonna, line 459: You need to use the `stateStoreLevelGroup()` here instead of `STATE_STORE_LEVEL_GROUP` because the group name depends on the version and the store type.

streams/src/test/java/org/apache/kafka/streams/integration/MetricsIntegrationTest.java
cadonna: I agree with you, it should always be 1. It is the group of the metrics. See my comment in `StateStoreMetrics`. I am glad this test served its purpose, because I did not notice this in the unit tests!

streams/src/test/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetricsTest.java
```diff
@@ -327,6 +327,38 @@ public void shouldGetExpiredWindowRecordDropSensor() {
...
+        final String descriptionOfAvg = "The average " + e2eLatencyDescription;
+        final String descriptionOfMin = "The minimum " + e2eLatencyDescription;
+        final String descriptionOfMax = "The maximum " + e2eLatencyDescription;
+
+        expect(streamsMetrics.storeLevelSensor(THREAD_ID, TASK_ID, STORE_NAME, metricName, RecordingLevel.TRACE))
+            .andReturn(expectedSensor);
+        expect(streamsMetrics.storeLevelTagMap(THREAD_ID, TASK_ID, STORE_TYPE, STORE_NAME)).andReturn(storeTagMap);
+        StreamsMetricsImpl.addAvgAndMinAndMaxToSensor(
+            expectedSensor,
+            STORE_LEVEL_GROUP,
```
cadonna, line 346: You need to make this dependent on the built-in metrics version by using instance variable `storeLevelGroup`.

cadonna (commented):
Thank you for the PR @ableegoldman 

Here my feedback:

streams/src/test/java/org/apache/kafka/streams/integration/MetricsIntegrationTest.java
cadonna: Sorry, I did a mistake here. We should not give new metrics old groups. I think to fix this test you need to adapt the filter on line 618 to let all metrics with groups that relate to KV state stores pass. See `checkWindowStoreAndSuppressionBufferMetrics()` for an example.

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetrics.java
cadonna: I just realized that we should not put new metrics into old groups. Your code is fine. Do not use `stateStoreLevelGroup()`! Sorry for the confusion.

streams/src/test/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetricsTest.java
cadonna: I just realized that we should not put new metrics into old groups. Your code is fine. Do not use instance variable `storeLevelGroup`! Sorry for the confusion.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
ableegoldman: Ah, ok that makes sense. Thanks for clarifying

streams/src/test/java/org/apache/kafka/streams/integration/MetricsIntegrationTest.java
ableegoldman: Hm, if I let additional metrics through (by allowing all `STATE_STORE_LEVEL_GROUP` metrics regardless of buildInMetricsVersion) then the test fails much earlier. For example
```java.lang.AssertionError: Size of metrics of type:'put-latency-avg' must be equal to 1 but it's equal to 3 ```

I think I could just do the check for the e2e latency before filtering, but I'm worried there's something else wrong here. Why do we get 3 `put-latency-avg` metrics if we let all store metrics through in `checkKeyValueStoreMetrics` but not when we let all metrics through in `checkWindowStoreAndSuppressionBufferMetrics`?

File: streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java (hunk 6 above)
Comment (ableegoldman) [reply streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java]: Ack

**Comment by ableegoldman** [comment]
Addressed all your comments @cadonna, ready for another pass. One open question about the MetricsIntegrationTest

**Comment by abbccdda** [comment]
retest this please

File: streams/src/test/java/org/apache/kafka/streams/integration/MetricsIntegrationTest.java (hunk 5 above)
Comment (cadonna) [reply streams/src/test/java/org/apache/kafka/streams/integration/MetricsIntegrationTest.java]: You can use the following to get it right without the need to do the check for the e2e latency before filtering

```
.filter(m -> m.metricName().tags().containsKey(tagKey) && 
    (m.metricName().group().equals(group0100To24) || m.metricName().group().equals(STATE_STORE_LEVEL_GROUP))
).collect(Collectors.toList());

```

The reason for the difference between the KV store and the window store is that they are used in different tests with different number of state stores.

The test that uses the KV stores tests three different types of KV stores, namely in-memory, rocksdb, and in-memory-lru-cache. For each of this types the old group name changes. That is also the reason we need to pass the parameter `group0100To24` to `checkKeyValueStoreMetrics()`.

In `checkWindowStoreAndSuppressionBufferMetrics()` we need to filter for four groups, because the corresponding test uses suppression and window state store. Suppression buffers had their own group in the old version. In the new version they moved into the state store group. Those groups are `BUFFER_LEVEL_GROUP_0100_TO_24` and `STATE_STORE_LEVEL_GROUP`. The window state store had their own group in the old version, i.e., `STATE_STORE_LEVEL_GROUP_ROCKSDB_WINDOW_STORE_0100_TO_24` (we are only using RocksDB-based window stores in the test). Finally, during the implementation of KIP-444, we discovered that we named a group incorrectly. That's why we filter also for group `stream-rocksdb-window-metrics`.

So to sum up, it is hard to compare the verifications for KV stores and window stores, because they are used in different tests. Sorry, I should have been clearer on that before.

Comment (ableegoldman) [reply streams/src/test/java/org/apache/kafka/streams/integration/MetricsIntegrationTest.java]: I see, thanks for the explanation.  The suggestion worked

**Comment by guozhangwang** [comment]
test this please

**Comment by guozhangwang** [comment]
test this please

**Comment by ableegoldman** [comment]
`EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta` failed, all other tests passed

**Comment by ableegoldman** [comment]
cc @guozhangwang for 2nd review & merge

File: streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java [hunk 9]
```diff
@@ -154,15 +153,15 @@ public StreamTask(final TaskId id,
         for (final String terminalNodeName : topology.terminalNodes()) {
             e2eLatencySensors.put(
                 terminalNodeName,
-                TaskMetrics.e2ELatencySensor(threadId, taskId, terminalNodeName, RecordingLevel.INFO, streamsMetrics)
+                TaskMetrics.e2ELatencySensor(threadId, taskId, terminalNodeName, streamsMetrics)
```
guozhangwang: This is not introduced by this PR, but I'm wondering why we record the `maybeRecordE2ELatency` twice, once before the record is being processed and once when it reaches the sink node?

guozhangwang (commented):
On the implementation side the PR looks good to me. I've a minor questions regarding the task level metrics recording.

I agree generally speaking it makes more sense to record at store level than at processor node level, but I still have a meta comment: we will decouple the caching / emitting for state stores soon and that means the only scenario that task-level e2e latency be different from store level is would be suppression itself.

So what about we only define it at the suppression node as a node-level metric still, and we record it whenever a record is suppressed for emitting at the moment? I know we've sort of discussed about this, but I just want to make sure we do not make metrics that are soon not useful at all. Instead, I'm a bit leaning towards adding the task level metrics first, and then consider if we add it at store or node level later after we've done the caching / suppression refactoring.

ableegoldman:
Hey @guozhangwang 
Thanks for the review. You have some high-level questions so I'll try to answer them here but let me know if you want to sync offline.

> the only scenario that task-level e2e latency be different from store level is would be suppression itself

Well not necessarily; for one thing, we introduced the TRACE level metrics so we could get the actual (not cached) system time without feeling guilty about the perf hit. So the store-level metrics give a finer-grained look into the subtopology for users who really care (while those who just want a rough ballpark can look at task-level only).

In general, it's likely that suppression will incur the more significant latency, but there are plenty of things that we know users do today that can incur noticeable latency. For better or worse they might make remote API or other potentially long-blocking calls, or have a slow state store causing a bottleneck, or scan large parts of the state, etc.

> I'm a bit leaning towards adding the task level metrics first

The task level metrics have already been added (min and max, just not avg). They're in 2.6 so that ship has sailed 🙂 

> decouple the caching / emitting for state stores soon 

By the way, do you have a ticket or quick writeup for this idea anywhere? You've referenced this in a few PRs now and I'd like to better understand your grand plan.

ableegoldman:
Oh I think I misread your original comment, editing my response above

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
guozhangwang: NVM, I realized it is at the per-node metrics level, we just have two types of nodes (source and termination node) for each task.

guozhangwang (commented):
@ableegoldman Sorry for the late reply!

Cool, I think that's a valid argument. But regarding the recording place, your argument makes me thinking that store-level trace may not cover other delay scenarios that are not caused by caching / suppression. So maybe we'd still prefer to record at the other nodes (than the source / termination nodes, since they are measured at info level anyways).

What about just call the measuring in each `ProcessorContext#forward` for all children, while the only difference is that some nodes are INFO while others are TRACE? wdyt @cadonna

streams/src/main/java/org/apache/kafka/streams/processor/internals/metrics/ProcessorNodeMetrics.java
```diff
@@ -289,6 +294,25 @@ public static Sensor processorAtSourceSensorOrForwardSensor(final String threadI
         return processAtSourceSensor(threadId, taskId, processorNodeId, streamsMetrics);
     }
 
+    public static Sensor e2ELatencySensor(final String threadId,
+                                          final String taskId,
+                                          final String processorNodeId,
+                                          final StreamsMetricsImpl streamsMetrics) {
+        final String sensorName = processorNodeId + "-" + RECORD_E2E_LATENCY;
+        final Sensor sensor = streamsMetrics.taskLevelSensor(threadId, taskId, sensorName, RecordingLevel.INFO);
```
ableegoldman: There's a slight hiccup with moving the INFO metrics from task to node level:
We get the current sensor from `StreamsMetrics#taskLevelSensor` which computes the `fullSensorName` with the `#taskSensorPrefix`
If we instead use `StreamsMetrics#nodeLevelSensor` then the` fullSensorName` is constructed from the `#nodeSensorPrefix`, which is obviously different. So moving this to a “true” node level sensor would be a breaking change, IIUC
I think the best we can do is just move this from TaskMetrics to ProcessorNodeMetrics, but still leave it as a taskLevelSensor. Let me know if I'm missing something though cc @guozhangwang @cadonna

cadonna: I do not think that using `StreamsMetrics#nodeLevelSensor()` is a breaking change. Sensors are not exposed publicly. They are merely containers for metrics. Metrics are exposed publicy. The full sensor name is just the key to retrieve sensors in the metrics map. It is also stored in `StreamsMetricsImpl` in the `*LevelSensor` data structures to know what sensors need to be removed from the metrics map when `removeAll*LevelSensors()` is called. So, the full sensor name is used for internal house keeping and hidden from the users. I think it is fine to change the `fullSensorName`.

cadonna (approved):
LGTM! Thank you, @ableegoldman!

streams/src/main/java/org/apache/kafka/streams/processor/internals/metrics/ProcessorNodeMetrics.java
ableegoldman: Oh, awesome, that's what I was hoping to hear! I'll move it to use the `nodeLevelSensor` then

ableegoldman:
@guozhangwang can we finally merge this? 🙂

guozhangwang:
LGTM.

PR #9095, opened 2020-07-29
KAFKA-10321: fix infinite blocking for global stream thread startup

abbccdda:
In the unit test `shouldDieOnInvalidOffsetExceptionDuringStartup` for JDK 11, we spotted a case where a global stream thread startup would stall if it fails immediately upon the first poll. The reason is that `start()` function only checks whether the thread is *not running*, as it needs to block until it finishes the initialization. However, if the thread transits to `DEAD` immediately, the `start()` call would block forever.

Use the failed unit test to verify it works.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStreamThread.java
@@ -44,6 +44,7 @@ import java.util.Set;
 
 import static org.apache.kafka.streams.processor.internals.GlobalStreamThread.State.DEAD;
 import static org.apache.kafka.streams.processor.internals.GlobalStreamThread.State.PENDING_SHUTDOWN;
+import static org.apache.kafka.streams.processor.internals.GlobalStreamThread.State.RUNNING;
 
 /**
  * This is the thread responsible for keeping all Global State Stores updated.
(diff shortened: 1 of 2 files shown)


vvcephei (commented):
Thanks, @abbccdda !

It looks good to me, but I just have a couple of small questions:
1. Should we throw an exception instead of just completing the `start()` method if we failed to go to RUNNING?
2. It seems like we could also check that the thread isn't pending shutdown, just to detect a little earlier that there was some error.

Thanks!

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStreamThread.java
```diff
@@ -173,6 +178,18 @@ public boolean stillRunning() {
...
 
+    public boolean inErrorState() {
+        synchronized (stateLock) {
+            return state.inErrorState();
+        }
+    }
+
+    public boolean stillInitializing() {
+        synchronized (stateLock) {
+            return !state.isRunning() && !state.inErrorState();
```
guozhangwang: Why not just `state.CREATED` as we are excluding three out of four states here?

guozhangwang (commented):
Just a minor comment, otherwise LGTM!

guozhangwang:
LGTM. Please feel free to merge and cherry-pick.

abbccdda:
1/3 green with two hanging tests:
shouldMigratePersistentWindowStoreToTimestampedWindowStoreUsingPapi
testReprocessingFromScratchAfterResetWithIntermediateUserTopic

abbccdda:
retest this please

abbccdda:
retest this please

PR #9098, opened 2020-07-29
KAFKA-9924: Prepare RocksDB and metrics for RocksDB properties recording

cadonna:
This PR refactors the RocksDB store and the metrics infrastructure in Streams 
in preparation of the recordings of the RocksDB properties specified in KIP-607.

The refactoring includes:
- wrapper around `BlockedBasedTableConfig` to make the cache accessible to the 
   RocksDB metrics recorder
- RocksDB metrics recorder now takes also the DB instance and the cache in addition
   to the statistics
- The value providers for the metrics are added to the RockDB metrics recorder also if
   the recording level is INFO.
- The creation of the RocksDB metrics recording trigger is moved to `StreamsMetricsImpl`   

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
@@ -213,7 +213,10 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
         } else {
             metricsRecorder.addValueProviders(name, db, null, statistics);
             log.warn("A table format configuration is used that does not expose the block cache. This means " +
-                "that metrics that relate to the block cache may be wrong if the block cache is shared.");
+                "that metrics that relate to the block cache may be wrong if the block cache is shared. " +
+                "Please use the BlockBasedTableConfig instance provided by Options#tableFormatConfig() to configure " +
+                "the internal table format of RocksDB. Do not provide a new instance of BlockBasedTableConfig to " +
+                "the RocksDB options.");
         }
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
@@ -188,7 +188,7 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
         openRocksDB(dbOptions, columnFamilyOptions);
         open = true;
 
-        addValueProvidersToMetricsRecorder(configs);
+        addValueProvidersToMetricsRecorder();
     }
 
     private void maybeSetUpStatistics(final Map<String, Object> configs) {
@@ -204,7 +204,7 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
         }
     }
 
-    private void addValueProvidersToMetricsRecorder(final Map<String, Object> configs) {
+    private void addValueProvidersToMetricsRecorder() {
         final TableFormatConfig tableFormatConfig = userSpecifiedOptions.tableFormatConfig();
         final Statistics statistics = userSpecifiedStatistics ? null : userSpecifiedOptions.statistics();
         if (tableFormatConfig instanceof BlockBasedTableConfigWithAccessibleCache) {
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
@@ -207,17 +207,14 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
     private void addValueProvidersToMetricsRecorder() {
         final TableFormatConfig tableFormatConfig = userSpecifiedOptions.tableFormatConfig();
         final Statistics statistics = userSpecifiedStatistics ? null : userSpecifiedOptions.statistics();
-        if (tableFormatConfig instanceof BlockBasedTableConfigWithAccessibleCache) {
-            final Cache cache = ((BlockBasedTableConfigWithAccessibleCache) tableFormatConfig).blockCache();
-            metricsRecorder.addValueProviders(name, db, cache, statistics);
-        } else {
-            metricsRecorder.addValueProviders(name, db, null, statistics);
-            log.warn("A table format configuration is used that does not expose the block cache. This means " +
-                "that metrics that relate to the block cache may be wrong if the block cache is shared. " +
-                "Please use the BlockBasedTableConfig instance provided by Options#tableFormatConfig() to configure " +
+        if (!(tableFormatConfig instanceof BlockBasedTableConfigWithAccessibleCache)) {
+            throw new ProcessorStateException("A table format configuration is used that does not expose the block cache. " +
+                "Use the BlockBasedTableConfig instance provided by Options#tableFormatConfig() to configure " +
                 "the internal table format of RocksDB. Do not provide a new instance of BlockBasedTableConfig to " +
                 "the RocksDB options.");
         }
+        final Cache cache = ((BlockBasedTableConfigWithAccessibleCache) tableFormatConfig).blockCache();
+        metricsRecorder.addValueProviders(name, db, cache, statistics);
     }
 
     void openRocksDB(final DBOptions dbOptions,
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
@@ -31,7 +31,6 @@ import org.apache.kafka.streams.state.KeyValueIterator;
 import org.apache.kafka.streams.state.KeyValueStore;
 import org.apache.kafka.streams.state.RocksDBConfigSetter;
 import org.apache.kafka.streams.state.internals.metrics.RocksDBMetricsRecorder;
-import org.rocksdb.BlockBasedTableConfig;
 import org.rocksdb.BloomFilter;
 import org.rocksdb.Cache;
 import org.rocksdb.ColumnFamilyDescriptor;
@@ -48,6 +47,7 @@ import org.rocksdb.RocksDB;
 import org.rocksdb.RocksDBException;
 import org.rocksdb.RocksIterator;
 import org.rocksdb.Statistics;
+import org.rocksdb.TableFormatConfig;
 import org.rocksdb.WriteBatch;
 import org.rocksdb.WriteOptions;
 import org.slf4j.Logger;
@@ -100,9 +100,9 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
     private BloomFilter filter;
 
     private RocksDBConfigSetter configSetter;
+    private boolean userSpecifiedStatistics = false;
 
     private final RocksDBMetricsRecorder metricsRecorder;
-    private boolean isStatisticsRegistered = false;
 
     ProcessorContext internalProcessorContext;
     // visible for testing
@@ -131,7 +131,7 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
         final ColumnFamilyOptions columnFamilyOptions = new ColumnFamilyOptions();
         userSpecifiedOptions = new RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapter(dbOptions, columnFamilyOptions);
 
-        final BlockBasedTableConfig tableConfig = new BlockBasedTableConfig();
+        final BlockBasedTableConfigWithAccessibleCache tableConfig = new BlockBasedTableConfigWithAccessibleCache();
         cache = new LRUCache(BLOCK_CACHE_SIZE);
         tableConfig.setBlockCache(cache);
         tableConfig.setBlockSize(BLOCK_SIZE);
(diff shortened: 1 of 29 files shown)


cadonna:
Call for review: @vvcephei @guozhangwang

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -181,23 +181,39 @@ void openDB(final ProcessorContext context) {
...
+        final TableFormatConfig tableFormatConfig = userSpecifiedOptions.tableFormatConfig();
+        final Statistics statistics = userSpecifiedStatistics ? null : userSpecifiedOptions.statistics();
+        if (tableFormatConfig instanceof BlockBasedTableConfigWithAccessibleCache) {
+            final Cache cache = ((BlockBasedTableConfigWithAccessibleCache) tableFormatConfig).blockCache();
+            metricsRecorder.addValueProviders(name, db, cache, statistics);
+        } else {
+            metricsRecorder.addValueProviders(name, db, null, statistics);
+            log.warn("A table format configuration is used that does not expose the block cache. This means " +
+                "that metrics that relate to the block cache may be wrong if the block cache is shared.");
         }
```
vvcephei: Ah, after reading your test, I now see the issue. I'd overlooked the fact that users would independently construct the table config object AND the cache. I see now that this makes it impossible to reliably capture the cache, since users have to actually choose to pass our special table config to the Options and then pass the Cache to that table config.

This doesn't seem ideal. What do you think about just using reflection instead?

```suggestion
        if (tableFormatConfig instanceof BlockBasedTableConfig) {
            final BlockBasedTableConfig blockBasedTableConfig = (BlockBasedTableConfig) tableFormatConfig;
            try {
                final Field blockCacheField = BlockBasedTableConfig.class.getDeclaredField("blockCache_");
                blockCacheField.setAccessible(true);
                final Cache nullableBlockCache = (Cache) blockCacheField.get(blockBasedTableConfig);
                metricsRecorder.addValueProviders(name, db, nullableBlockCache, statistics);
            } catch (final NoSuchFieldException | IllegalAccessException | ClassCastException e) {
                log.warn("Expected to find and access field 'blockCache_' in BlockBasedTableConfig. " +
                             "Probably, an incompatible version of RocksDB is being used. " +
                             "Cache will be missing from memory metrics.", e);
                metricsRecorder.addValueProviders(name, db, null, statistics);
            }
        } else {
            metricsRecorder.addValueProviders(name, db, null, statistics);
        }
```

We would obviously test all the branches here to de-risk the reflection. We can also add a test that searches the classpath for implementations of TableFormatConfig to ensure we don't miss the memo if RocksDB adds a new TableFormatConfig implementation.

Alternative thought, if you don't like the reflection: We would _also_ subclass Options and override `org.rocksdb.Options#setTableFormatConfig` to check if the passed `TableFormatConfig` is a `BlockBasedTableConfig`, and if so, then _we_ wrap it with `BlockBasedTableConfigWithAccessibleCache`.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/MockProcessorContext.java
```diff
@@ -227,7 +228,8 @@ public MockProcessorContext(final Properties config, final TaskId taskId, final
         this.metrics = new StreamsMetricsImpl(
             new Metrics(metricConfig),
             threadId,
-            streamsConfig.getString(StreamsConfig.BUILT_IN_METRICS_VERSION_CONFIG)
+            streamsConfig.getString(StreamsConfig.BUILT_IN_METRICS_VERSION_CONFIG),
+            Time.SYSTEM
```
vvcephei, line 232: Interesting... Should we add a `Time` argument to the constructor? It would be a minor amendment to the KIP.

vvcephei (commented):
Thanks, @cadonna !

This PR looks great, especially the tests.

I had a couple of thoughts.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/MockProcessorContext.java
cadonna: I would prefer to postpone that, because currently it is not strictly needed and the time is only used in the RocksDB recording trigger that records only internal RocksDB metrics. I do not see how exposing time would be useful for users during testing. If anybody complains, we can still do it in a future KIP.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
cadonna: I agree with you that it is not ideal and thank you for this lesson on reflection. 

Indeed, I do not like reflection in this case, because it makes the code too much dependent on RocksDB internals. We should use reflection to check if the public API to configure RocksDB changed in a newer version, but that is another story.

I do not understand how the alternative of wrapping `BlockBasedTableConfig` into `BlockBasedTableConfigWithAccessibleCache` should work. Since the cache is not accessible in `BlockBasedTableConfig` it will also not be accessible when it is wrapped in `BlockBasedTableConfigWithAccessibleCache` (despite the name). We need to get the reference to the cache when the cache is set in `BlockBasedTableConfig`. If the cache is already set we can only use reflection.

Since the block based table format is the only format in RocksDB that uses the cache, I do not see why a user absolutely needs to pass a new `BlockBasedTableConfig` object. I think for now it is OK to log a warning, and clearly document that the provided `BlockBasedTableConfig` object should be used.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -204,17 +204,17 @@ private void maybeSetUpStatistics(final Map<String, Object> configs) {
...
         final Statistics statistics = userSpecifiedStatistics ? null : userSpecifiedOptions.statistics();
-        if (tableFormatConfig instanceof BlockBasedTableConfigWithAccessibleCache) {
-            final Cache cache = ((BlockBasedTableConfigWithAccessibleCache) tableFormatConfig).blockCache();
-            metricsRecorder.addValueProviders(name, db, cache, statistics);
-        } else {
-            metricsRecorder.addValueProviders(name, db, null, statistics);
-            log.warn("A table format configuration is used that does not expose the block cache. This means " +
-                "that metrics that relate to the block cache may be wrong if the block cache is shared.");
-        }
+        if (!(tableFormatConfig instanceof BlockBasedTableConfigWithAccessibleCache)) {
```
vvcephei: It seems like we would now be forbidding the use of `PlainTableConfig`. Is that intentional?

vvcephei (commented):
Thanks for the update, @cadonna ! I had a follow-on question.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -181,26 +181,42 @@ void openDB(final ProcessorContext context) {
...
         openRocksDB(dbOptions, columnFamilyOptions);
         open = true;
+
+        addValueProvidersToMetricsRecorder();
     }
 
-    private void maybeSetUpMetricsRecorder(final Map<String, Object> configs) {
-        if (userSpecifiedOptions.statistics() == null &&
+    private void maybeSetUpStatistics(final Map<String, Object> configs) {
+        if (userSpecifiedOptions.statistics() != null) {
```
guozhangwang, line 196: Could this ever be null actually? I think even in unit tests the DBOptions would contain stats?

guozhangwang (approved):
Just one minor question, otherwise lgtm. I think the preparation makes sense for the follow-up PR.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
cadonna: The statistics are `null` when it is not set in the options. If it were not `null` unit test `RocksDBTest#shouldAddValueProvidersWithStatisticsToInjectedMetricsRecorderWhenRecordingLevelDebug()` would fail. I also verified with a debug run.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
cadonna: No, it was not intentional. I simply did not think hard enough.  
I agree with you that we should also allow `PlainTableConfig`. If users use the PlainTable format they should already expect that block cache specific metrics do not show up or report constant zero. 
Thank you for pointing this out!

vvcephei (approved):
Thanks, @cadonna !

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
guozhangwang: I was thinking that, if we know `maybeSetUpStatistics` is only called once in lifetime of the store then here we could check that `userSpecifiedOptions.statistics() == null` and otherwise throw illegal-state exception. And then in `addValueProvidersToMetricsRecorder` we check that ``userSpecifiedOptions.statistics() != null` and otherwise throw illegal-state exception. This is to make the call-trace assumptions more specific.

cadonna: I see what you are aiming at. However, why should we throw an `IllegalStateException` when `userSpecifiedOptions.statistics() != null`? The user could provide a statistics object through the config setter and use it to read the statistics externally to our metric framework. Such a pattern was supported before and was not changed in this PR or the PR that introduced the statistics-based RocksDB metrics.
Anyways, it is good that you mentioned this, because I checked again the code in the metrics recorder and a `null` check was missing. Plus I included some checks to ensure that either all statistics of a metrics recorder are`null` or all statistics of a metrics recorder are NOT `null`.

vvcephei:
Ok to test

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/RocksDBMetricsRecorder.java
```diff
@@ -187,37 +206,39 @@ public void record(final long now) {
...
-            blockCacheIndexMisses += valueProviders.statistics.getAndResetTickerCount(TickerType.BLOCK_CACHE_INDEX_MISS);
-            blockCacheFilterHits += valueProviders.statistics.getAndResetTickerCount(TickerType.BLOCK_CACHE_FILTER_HIT);
-            blockCacheFilterMisses += valueProviders.statistics.getAndResetTickerCount(TickerType.BLOCK_CACHE_FILTER_MISS);
-            writeStallDuration += valueProviders.statistics.getAndResetTickerCount(TickerType.STALL_MICROS);
-            bytesWrittenDuringCompaction += valueProviders.statistics.getAndResetTickerCount(TickerType.COMPACT_WRITE_BYTES);
-            bytesReadDuringCompaction += valueProviders.statistics.getAndResetTickerCount(TickerType.COMPACT_READ_BYTES);
-            numberOfOpenFiles += valueProviders.statistics.getAndResetTickerCount(TickerType.NO_FILE_OPENS)
-                - valueProviders.statistics.getAndResetTickerCount(TickerType.NO_FILE_CLOSES);
-            numberOfFileErrors += valueProviders.statistics.getAndResetTickerCount(TickerType.NO_FILE_ERRORS);
+        if (storeToValueProviders.values().stream().anyMatch(valueProviders -> valueProviders.statistics != null)) {
```
vvcephei: I think it's implied in `addValueProviders` that if any of the valueProviders' statistics are non-null, then they are all non-null, in which case, this makes sense as a guard. Still, it's kind of subtle.

Why not just put a guard inside the loop instead to `continue` or `break` if it turns out that `valueProviders.statistics == null`?

vvcephei (commented):
Hey @cadonna , just one small piece of feedback.

vvcephei (approved):
beautiful. Thanks, @cadonna !

vvcephei:
Test this please

vvcephei:
Ok to test

vvcephei:
Test this please

vvcephei:
Retest this please

vvcephei:
Test this please

vvcephei:
Well, I guess I'll just run them locally.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
guozhangwang: Hmm, that means if a user calls `setStatistics` externally then streams itself would not provide the built-in metrics, is that right? I was not aware of this, and if it is by-design then that's fine.

cadonna: Actually, Streams would provide the built-in metrics, but not update them. I also thought about not providing them at all, but that is a bit tricky and I did not want spent too much time on such a rare case. I will open a Jira and defer this to later.

vvcephei:
Test this please

vvcephei:
Test this please

vvcephei:
Test this please

vvcephei:
@cadonna : 

```
> Task :streams:checkstyleMain
[ant:checkstyle] [ERROR] /home/confluent/kafka/streams/src/generated/java/org/apache/kafka/streams/InteractiveQueryPartitionResponse.java:1: Line does not match expected header line of '/*'. [Header]
```

cadonna:
Test this please

cadonna:
ok to test

cadonna:
Test this please

vvcephei:
Test this please

vvcephei:
\o/

vvcephei:
Retest this please

vvcephei:
Looks like the tests never actually ran. I'm running them locally now.

vvcephei:
Build passed:
```
$ test-kafka :streams:testAll
...
BUILD SUCCESSFUL in 9m 57s
94 actionable tasks: 94 executed
```

PR #9099, opened 2020-07-29
KAFKA-6733: Printing additional ConsumerRecord fields in DefaultMessageFormatter

badaiaqrandista:
Implementation of KIP-431 - Support of printing additional ConsumerRecord fields in DefaultMessageFormatter

https://cwiki.apache.org/confluence/display/KAFKA/KIP-431%3A+Support+of+printing+additional+ConsumerRecord+fields+in+DefaultMessageFormatter

comment to ping reviewers.*


core/src/main/scala/kafka/tools/ConsoleConsumer.scala
```diff
@@ -459,48 +466,32 @@ class DefaultMessageFormatter extends MessageFormatter {
...
-  private def propertiesWithKeyPrefixStripped(prefix: String, props: Properties): Properties = {
-    val newProps = new Properties()
-    props.asScala.foreach { case (key, value) =>
-      if (key.startsWith(prefix) && key.length > prefix.length)
-        newProps.put(key.substring(prefix.length), value)
-    }
-    newProps
+  var headersDeserializer: Option[Deserializer[_]] = None
+
+  override def init(props: Properties): Unit = {
```
dajac: `init(props: Properties)` has been deprecated. It would be great if we could keep using `configure(configs: Map[String, _])` as before. I think that we should also try to directly extract the values from the `Map` instead of using a `Properties`.

core/src/test/scala/kafka/tools/DefaultMessageFormatterTest.scala
```diff
@@ -0,0 +1,235 @@
...
+        "h1:v1,h2:v2\n"),
+      Array(
+        "print empty headers",
+        consumerRecord(headers = Nil),
+        Map("print.headers" -> "true", "print.value" -> "false"),
+        "NO_HEADERS\n"),
+      Array(
+        "print all possible fields with default delimiters",
+        consumerRecord(),
+        Map("print.key" -> "true",
```
dajac, line 121: nit: Move `print.key` to next line to remain consistent with the formatting of the other Maps.

dajac (commented):
@badaiaqrandista Thanks for the PR. It looks good overall. I left few comments.

core/src/main/scala/kafka/tools/ConsoleConsumer.scala
badaiaqrandista: @dajac 

I have replaced `init` with `configure` and changed code to extract the values directly from `Map`. Please review again.

core/src/test/scala/kafka/tools/DefaultMessageFormatterTest.scala
badaiaqrandista: @dajac 

I've re-indent these as well.

badaiaqrandista:
@dajac Can you please re-review the changes I've made?

bbejeck:
Ok to test

badaiaqrandista:
@bbejeck Thank you for including the changes in the test. I've fixed the errors found by the test. Please test again.

bbejeck:
Ok to test

bbejeck:
ok to test

bbejeck:
retest this please.

badaiaqrandista:
@bbejeck 

One question: As part of my PR, I add a new test for DefaultMessageFormatter class: core/src/test/scala/kafka/tools/DefaultMessageFormatterTest.scala

But I found that there is another unit test that include a test for DefaultMessageFormatter class:
https://github.com/apache/kafka/blob/trunk/core/src/test/scala/unit/kafka/tools/ConsoleConsumerTest.scala#L501

This was the cause of the system test failure yesterday. I changed the latter test to work with the updated DefaultMessageFormatter. But after thinking about it further, I should just delete the test for DefaultMessageFormatter in ConsoleConsumerTest.scala.

What do you think?

bbejeck:
> I should just delete the test for DefaultMessageFormatter in ConsoleConsumerTest.scala.

@badaiaqrandista

I'm not sure.  My vote would be to keep the test but move it over to the `DefaultMessageFormatterTest.scala` class.  

But I'm not familiar enough with this code to say for sure.  From a quick look at the old test, it's not clear to me why it failed.  I guess the `Partition number` gets printed by default now?

\cc @dajac

badaiaqrandista:
@bbejeck 

It failed because it print the partition number as a single integer after the value. I moved the partition to be before the key (if printed) and value, and also added prefix "Partition:" to differentiate it from "Offset:".

It's only printed if "print.partition=true".

I will keep the tests as is then. ConsoleConsumerTest is a unit test for the class while DefaultMessageFormatterTest is more of an integration test.

bbejeck:
@badaiaqrandista can you rebase this PR?   We have a little bit of time to get this into 2.7 before the code freeze on 10/21

dajac:
Sorry guys. I have completely forgot this one. I think that it is fine to keep both tests as suggested by @badaiaqrandista.

dajac (commented):
@badaiaqrandista Thanks for the PR. The code LGTM pending Jenkins and the rebase.

bbejeck (commented):
Thanks for the PR @badaiaqrandista this LGTM modulo rebasing and a clean build.

bbejeck:
retest this please

bbejeck:
ok to test

bbejeck:
Java 11 and Java 15 passed.

Java 8 failed with `org.apache.kafka.streams.integration.PurgeRepartitionTopicIntegrationTest.shouldRestoreState` known flaky test.

badaiaqrandista:
@bbejeck Do I need to do anything on this?

bbejeck:
>Do I need to do anything on this?

@badaiaqrandista, nope as soon as we can get a green build, I'll merge it.

bbejeck:
Merged #9099 into trunk

bbejeck:
Thanks for the contribution @badaiaqrandista!

PR #9101, opened 2020-07-29
KAFKA-10325: KIP-649 implementation

ryandielhenn:
This is the initial implementation of [KIP-649](https://cwiki.apache.org/confluence/display/KAFKA/KIP-649%3A+Dynamic+Client+Configuration) which implements dynamic client configuration. This work is still in progress.

TODO: 

1. Add supporting types to org.apache.kafka.common.config for the new network protocol instead of reusing the types in org.apache.kafka.common.quota.

2. Have clients register supported configs to an internal topic so that compatibility information and static client configs can be returned to the administrator or processed with an external tool (e.g. Apache Druid). Set a log retention time on the topic and send the retention time back to each client that requests dynamic configs. Have clients register supported configs every time they pull their dynamic configs on an interval equal to the log retention time so that only compatibility information of active clients is kept in the internal topic.


clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -356,6 +362,11 @@ boolean ensureActiveGroup(final Timer timer) {
         }
 
         startHeartbeatThreadIfNeeded();
+        if (!dynamicConfig.shouldDisable()) {
+            // This will only return a future and block for it if this is before the first JoinGroupRequest being sent
+            RequestFuture<ClientResponse> configsFuture = dynamicConfig.maybeFetchInitialConfigs();
+            dynamicConfig.maybeWaitForInitialConfigs(configsFuture);
```
jsancio: There are 3 consecutive calls to methods for `dynamicConfig`. We can move all of this code to `DynamicConsumerConfig`.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -504,6 +515,7 @@ public void onSuccess(ByteBuffer value) {
                             log.info("Successfully joined group with generation {}", generation.generationId);
                             state = MemberState.STABLE;
                             rejoinNeeded = false;
+                            rebalanceConfig.coordinatorUpdated();
```
jsancio: I think we have a race between the `sendJoinGroupRequest` completing with some session timeout X and getting another value Y for the session timeout in `DescribeReponse`.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -155,6 +157,10 @@ public AbstractCoordinator(GroupRebalanceConfig rebalanceConfig,
         this.time = time;
         this.heartbeat = new Heartbeat(rebalanceConfig, time);
         this.sensors = new GroupCoordinatorMetrics(metrics, metricGrpPrefix);
+        this.dynamicConfig = new DynamicConsumerConfig(client, this, rebalanceConfig, time, logContext);
+        if (!rebalanceConfig.enableDynamicConfig()) {
+            dynamicConfig.disable();
+        }
```
jsancio: Two consecutive calls to `DynamicConsumerConfig`. How about:
```java
this.dynamicConfig = new DynamicConsumerConfig(
  client, this, rebalanceConfig, time, logContext, rebalanceConfig.enableDynamicConfig()
); 
```

clients/src/main/java/org/apache/kafka/clients/consumer/internals/DynamicConsumerConfig.java
```diff
@@ -0,0 +1,168 @@
...
+ */
+public class DynamicConsumerConfig extends DynamicClientConfigUpdater {
+    /* Client to use */
+    private ConsumerNetworkClient client;
+
+    /* Configs to update */
+    private GroupRebalanceConfig rebalanceConfig;
+
+    /* Object to synchronize on when response is recieved */
+    Object lock;
```
jsancio: This makes the concurrency requirement very hard to read and validate. This could be any object since it is pass through the constructor.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/DynamicConsumerConfig.java
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * Send a {@link DescribeConfigsRequest} to a node specifically for dynamic client configurations
+     *
+     * @return {@link RequestFuture} 
+     */ 
+    public RequestFuture<ClientResponse> maybeFetchInitialConfigs() {
+        if (!initialConfigsFetched) {
+            Node node = null;
+            while (node == null) {
+                node = client.leastLoadedNode();
```
jsancio: Is this a livelock? How will this return non-null? What thread is calling this method and what thread is responsible for polling the client until there is a node?

clients/src/main/java/org/apache/kafka/clients/consumer/internals/DynamicConsumerConfig.java
```diff
@@ -0,0 +1,168 @@
...
+     * Block for a {@link DescribeConfigsResponse} and process it. Used to fetch the initial dynamic configurations synchronously before sending the initial
+     * {@link org.apache.kafka.common.requests.JoinGroupRequest}. Since this join RPC sends the group member's session timeout
+     * to the group coordinator, we should check if a dynamic configuration for session timeout is set before joining.
+     * If we do not do this initial fetch synchronously, then we could possibly trigger an unnecessary group rebalance operation by 
+     * sending a second join request after the dynamic configs are recieved asynchronously.
+     *
+     * @param responseFuture - future to block on
+     * @return true if responseFuture was blocked on and a response was recieved
+     */
+    public boolean maybeWaitForInitialConfigs(RequestFuture<ClientResponse> responseFuture) {
```
jsancio: This method and the method above can be merged and simplify the API. Every call to `maybeFetchInitialConfigs` is followed by a call to `maybeWaitForInitialConfigs`.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/DynamicConsumerConfig.java
```diff
@@ -0,0 +1,168 @@
...
+        if (responseFuture != null) {
+            client.poll(responseFuture);
+            if (responseFuture.isDone()) {
+                DescribeConfigsResponse configsResponse = (DescribeConfigsResponse) responseFuture.value().responseBody();
+                handleSuccessfulResponse(configsResponse);
+                this.initialConfigsFetched = true;
+                return true;
+            }
+        }
+        return false;
```
jsancio: Why return a boolean? It looks like the client of this method never uses it.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/DynamicConsumerConfig.java
```diff
@@ -0,0 +1,168 @@
...
+        // We only want to process them if they have changed since the last time they were fetched.
+        if (!dynamicConfigs.equals(previousDynamicConfigs)) {
+            previousDynamicConfigs = dynamicConfigs;
+            try {
+                rebalanceConfig.setDynamicConfigs(dynamicConfigs);
+            } catch (IllegalArgumentException e) {
+                log.info("Rejecting dynamic configs: {}", e.getMessage());
+            }
+        }
+        update();
```
jsancio: `update` is called twice in a row. Here and on line 132 on this file.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
```diff
@@ -148,6 +148,10 @@ public Sender(LogContext logContext,
         this.apiVersions = apiVersions;
         this.transactionManager = transactionManager;
         this.inFlightBatches = new HashMap<>();
+        this.dynamicConfig = new DynamicProducerConfig(client, config, time, logContext, requestTimeoutMs);
+        if (!config.getBoolean(CommonClientConfigs.ENABLE_DYNAMIC_CONFIG_CONFIG)) {
+            this.dynamicConfig.disable();
```
jsancio: Two consecutive calls to `DynamicProducerConfig`. We are already passing `config` in the constructor.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
```diff
@@ -321,11 +325,15 @@ void runOnce() {
         }
 
         long currentTimeMs = time.milliseconds();
-        long pollTimeout = sendProducerData(currentTimeMs);
+        if (!dynamicConfig.shouldDisable()) {
+            dynamicConfig.maybeFetchConfigs(currentTimeMs);
```
jsancio: Two consecutive calls to `dynamicConfig` this can be done in a method for `dynamicConfig` and simplify the API.

core/src/main/scala/kafka/server/AdminManager.scala
```diff
@@ -431,6 +433,26 @@ class AdminManager(val config: KafkaConfig,
...
                   .setIsSensitive(false).setReadOnly(false).setSynonyms(List.empty.asJava))
+          case ConfigResource.Type.CLIENT =>
+            val clientId = resource.resourceName
+            val defaultProps = adminZkClient.fetchEntityConfig(ConfigType.Client, ConfigEntityName.Default)
+            val clientProps = adminZkClient.fetchEntityConfig(ConfigType.Client, if (clientId.isEmpty) ConfigEntityName.Default else clientId)
+            val overlayedProps = new Properties()
+            overlayedProps.putAll(defaultProps)
+            overlayedProps.putAll(clientProps)
+            val configMap = overlayedProps.stringPropertyNames.asScala
+              .filter(ClientConfigs.isClientConfig).map{key => (key -> overlayedProps.getProperty(key))}.toMap
```
jsancio: Java's `Properties` is a `HashTable`. This can be done by using `flatMap`. E.g.:

```scala
overlayedProps.asScala.flatMap { (key, value) =>
  val key = key.toString
  val value = value.toString

  if (ClientConfigs.isClientConfig(key) {
    None
  } else {
    Some(key -> value)
  }
}

File: core/src/main/scala/kafka/server/AdminManager.scala [hunk 12]
```diff
@@ -431,6 +433,26 @@ class AdminManager(val config: KafkaConfig,
...
+            // Resort to default dynamic client config if configs are not specified for the client-id
+            if (clientId.nonEmpty) {
+              createResponseConfig(configMap,
+                createClientConfigEntry(clientId, clientProps, defaultProps, perClientIdConfig = true, 
+                  includeSynonyms, includeDocumentation))
+            } else {
+              createResponseConfig(configMap,
+                createClientConfigEntry(clientId, clientProps, defaultProps, perClientIdConfig = false, 
+                  includeSynonyms, includeDocumentation))
+            }
```
jsancio:
```scala
createResponseConfig(
  configMap,
  createClientConfigEntry(
    clientId,
    clientProps,
    defaultProps, 
    perClientIdConfig = clientId.nonEmpty, 
    includeSynonyms,
    includeDocumentation
  )
)
```

core/src/main/scala/kafka/server/AdminManager.scala
```diff
@@ -597,6 +619,15 @@ class AdminManager(val config: KafkaConfig,
             if (!validateOnly)
               alterLogLevelConfigs(alterConfigOps)
             resource -> ApiError.NONE
+          case ConfigResource.Type.CLIENT =>
+            val (configType, configKeys) = (ConfigType.Client, DynamicConfig.Client.configKeys)
```
jsancio: You are constructing a tuple and pattern matching against it. Not sure if the compiler optimizes this way. It looks these are all alias to known symbols. Do we need this?

core/src/main/scala/kafka/server/DynamicConfig.scala
```diff
@@ -78,23 +78,42 @@ object DynamicConfig {
     def isQuotaConfig(name: String): Boolean = configNames.contains(name)
   }
 
+  object ClientConfigs {
+    val AcksOverrideProp = "acks"
+    val SessionTimeoutOverrideProp = "session.timeout.ms"
+    val HeartbeatIntervalOverrideProp = "heartbeat.interval.ms"
```
jsancio: Duplicate strings. These are specified somewhere else. Those symbols should be accessible from there.

core/src/main/scala/kafka/server/DynamicConfig.scala
```diff
@@ -103,12 +122,25 @@ object DynamicConfig {
...
       .define(RequestPercentageOverrideProp, DOUBLE, DefaultRequestOverride, MEDIUM, RequestOverrideDoc)
+      .define(AcksOverrideProp, STRING, DefaultAcksOverride, ConfigDef.ValidString.in("all", "-1", "0", "1"), HIGH, AcksOverrideDoc)
+      .define(
+        SessionTimeoutOverrideProp, 
+        INT, 
+        DefaultSessionTimeoutOverride, 
+        ConfigDef.Range.between(Defaults.GroupMinSessionTimeoutMs, Defaults.GroupMaxSessionTimeoutMs), 
+        HIGH, 
+        SessionTimeoutOverrideDoc)
+      .define(HeartbeatIntervalOverrideProp, INT, DefaultHeartbeatIntervalOverride, HIGH, HeartbeatIntervalOverrideDoc)
```
jsancio, line 142: For all of the properties that you have added: didn't we decide to not validate the values here and to instead let the client do the validation?

jsancio (commented):
Thanks for the PR. First pass. Partial review.

core/src/main/scala/kafka/server/AdminManager.scala
ryandielhenn:
```
  if (ClientConfigs.isClientConfig(key) {
    Some(key -> value)
  } else {
    None
  }
```

Want to keep the client configs and discard the quotas.

core/src/main/scala/kafka/server/DynamicConfig.scala
ryandielhenn: This prevents key-value pairs that the broker knows are invalid from being persisted to zookeeper e.g. acks=2. The group coordinator also requires that a group members session timeout is in between GroupMinSessionTimeoutMs and GroupMaxSessionTimeoutMs, so the broker can also prevent invalid session timeouts from being persisted to zk.

The motivation behind doing validation on the broker and even more validation on the client is that some clients could reject the dynamic configs because of conflicts with user-provided configs while other clients accept the dynamic config. As long as the broker sanity-checks them before persisting to zk this is possible. If sanity checking is not done, and for example acks=2 is persisted, it is invalid for all clients not a subset of them.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/DynamicConsumerConfig.java
ryandielhenn: I used this to test that the RPC was only being sent periodically but this can be done in a different way... I will make the return type void.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/DynamicConsumerConfig.java
ryandielhenn: I think I forgot to remove this after starting to use atomics to track the interval and in-progress updates.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/DynamicConsumerConfig.java
ryandielhenn: The main thread is calling the method and the main thread is polling. This was a hack to try to force the initial describe configs before the join group request. This returns non-null when there is a connected node, connecting node, or node with no connection. It may be best to just fetch the configs and poll for them if the node is non-null and ready to send a request to.  If it is null or not ready then maybe skip the first synchronous DescribeConfigs RPC and fetch the configs asynchronously on the next try.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
dielhennr: I think fixing the synchronization also fixed this race.

ryandielhenn: The latest commit tries to address this.

ryandielhenn:
Hey @jsancio , 

I added some work in progress to this branch including a new RPC for this feature. Fitting user and client-id into the `DescribeConfigs` API was awkward, so I thought that the next best step would be to create a new RPC similar to  [KIP-546](https://cwiki.apache.org/confluence/display/KAFKA/KIP-546%3A+Add+Client+Quota+APIs+to+the+Admin+Client). These changes allow for a more expressive and well-defined interface. 

I am also working on having the clients register the configs that they support with the brokers. I tried tying the registration to connectionId in the hopes that this would give a unique identifier to each running application. However, this will not work since the connectionId can change while a client is active. Similarly, tying registration to ip:port will not work because a client talks to different brokers on different ports. Would it be safe to assume that clients with the same ip address are all the same version? Do you have any suggestions for what identifier config registration should be tied to if this assumption cannot be made?

EDIT: I updated the KIP’s rejected alternatives section with a design that ties config registration to <user,client-id> along with ClientInformation. This update would allow the user to see what dynamic configs are supported for each combination of client software name and version that is requesting the configs associated with a <user, client-id>.

mimaison:
KIP-649 was never voted and is now abandoned so I'll close this PR. Feel free to restart the discussion on the KIP if it's still something you want to do.

mimaison:
KIP-649 was never voted and is now abandoned so I'll close this PR. Feel free to restart the discussion on the KIP if it's still something you want to do.

PR #9102, opened 2020-07-30
KAFKA-10326 Both serializer and deserializer should be able to see th…

chia7712:
https://issues.apache.org/jira/browse/KAFKA-10326

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java
+++ b/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java
@@ -228,6 +228,13 @@ public class AbstractConfig {
         return copy;
     }
 
+    public Map<String, Object> originals(Map<String, Object> configOverrides) {
+        Map<String, Object> copy = new RecordingMap<>();
+        copy.putAll(originals);
+        copy.putAll(configOverrides);
+        return copy;
+    }
+
     /**
      * Get all the original settings, ensuring that all values are of type String.
      * @return the original settings
--- a/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java
+++ b/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java
@@ -228,7 +228,7 @@ public class AbstractConfig {
         return copy;
     }
 
-    public Map<String, Object> originals(Map<String, Object> configOverrides) {
+    public Map<String, Object> copyWithOverride(Map<String, Object> configOverrides) {
         Map<String, Object> copy = new RecordingMap<>();
         copy.putAll(originals);
         copy.putAll(configOverrides);
--- a/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java
+++ b/clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java
@@ -228,7 +228,7 @@ public class AbstractConfig {
         return copy;
     }
 
-    public Map<String, Object> copyWithOverride(Map<String, Object> configOverrides) {
+    public Map<String, Object> originals(Map<String, Object> configOverrides) {
         Map<String, Object> copy = new RecordingMap<>();
         copy.putAll(originals);
         copy.putAll(configOverrides);
--- a/clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
@@ -2535,4 +2535,31 @@ public class KafkaConsumerTest {
 
         assertEquals(countingRebalanceListener.revokedCount, 1);
     }
+
+    @Test
+    public void deserializerShouldSeeGeneratedClientId() {
+        Properties props = new Properties();
+        props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9999");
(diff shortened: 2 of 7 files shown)


chia7712:
```EosBetaUpgradeIntegrationTest``` is flaky... retest this please

**Review by abbccdda — COMMENTED** [review]
Are there any serde use cases that require client id?

**Comment by chia7712** [comment]
> Are there any serde use cases that require client id?

I observed this issue when we are removing custom client id config. It causes error as the metrics of our serde is encoded with client id.

File: clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java [hunk 1]
```diff
@@ -228,6 +228,13 @@ public Password getPassword(String key) {
         return copy;
     }
 
+    public Map<String, Object> originals(Map<String, Object> configOverrides) {
```
abbccdda, line 231: s/originals/copyWithOverride?

abbccdda (commented):
The change makes sense to me. Will also let @guozhangwang take a look. Btw, I don't think this is an improvement rather than a bug, as we don't have any guarantee to see client id in serdes before.

chia7712:
> Btw, I don't think this is an improvement rather than a bug, as we don't have any guarantee to see client id in serdes before.

There is another reason. We do pass generated client id to metric reporter.  It seems to me all plugins should see consistent props.

clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java
chia7712: Will copy that

guozhangwang:
This change makes sense to me, thanks @chia7712 

One (very) minor thing is that, `enableAutoCommit` may also be overridden when passing into the coordinator object, i.e. the value read from ConsumerConfig may be different from what's actually used. I'm not too worry that this would be leveraged by serde but just want to bring this up.

chia7712:
@guozhangwang thanks for reviews

> One (very) minor thing is that,  may also be overridden when passing into the coordinator object, i.e. the value read from ConsumerConfig may be different from what's actually used. I'm not too worry that this would be leveraged by serde but just want to bring this up.

Let me address it to make consistent configs for all plugins.

clients/src/main/java/org/apache/kafka/common/config/AbstractConfig.java
chia7712: I have updated the method name.

chia7712:
```org.apache.kafka.clients.admin.KafkaAdminClientTest.testMetadataRetries``` is traced by #9091 and ```EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]``` is flaky. retest this please

File: clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java [hunk 2]
```diff
@@ -2511,4 +2511,39 @@ public void testEnforceRebalanceTriggersRebalanceOnNextPoll() {
...
+        props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, DeserializerForClientId.class.getName());
+
+        KafkaConsumer<byte[], byte[]> consumer = new KafkaConsumer<>(props);
+
+        assertEquals(2, DeserializerForClientId.CLIENT_IDS.size());
+        assertEquals(DeserializerForClientId.CLIENT_IDS.get(0), consumer.getClientId());
+        assertEquals(DeserializerForClientId.CLIENT_IDS.get(1), consumer.getClientId());
+
+        assertEquals(2, DeserializerForClientId.AUTO_COMMIT_FLAGS.size());
+        assertEquals(DeserializerForClientId.AUTO_COMMIT_FLAGS.get(0), consumer.isAutoCommitEnabled());
```
abbccdda: This seems beyond the scope of this PR. Why do we also need to let deserializer see whether given consumer instance is auto commit enabled?

chia7712: This flag is included by reason of this comment https://github.com/apache/kafka/pull/9102#issuecomment-667734629

I’m ok to revert it if it is overkill

chia7712:
@guozhangwang @abbccdda More comments?

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
abbccdda: Sorry for the delay, I suggest we revert this part to properly scope this PR.

chia7712: @abbccdda thanks for suggestion. I have reverted it.

chia7712:
@guozhangwang @abbccdda ping ~

guozhangwang:
I took a quick look and it lgtm. @abbccdda If your pass is good please feel free to merge.

chia7712:
@abbccdda ping~

chia7712:
rebase to trigger new QA

guozhangwang:
I've merged it to trunk and marked fix version as 2.7, thanks a lot for your contribution @chia7712 !

PR #8850, opened 2020-06-11
KAFKA-10141: Add more detail to log segment delete messages

skaundinya15:
As specified in https://issues.apache.org/jira/browse/KAFKA-10141, it would be helpful to include as much information as possible when deleting log segments. This patch introduces log messages that give more specific details as to why the log segment was deleted and the specific metadata regarding that log segment.

*More detailed description of your change,
if necessary. The PR title and PR message become
the squashed commit message, so use a separate
comment to ping reviewers.*

*Summary of testing strategy (including rationale)
for the feature or bug fix. Unit and/or integration
tests are expected for any behaviour change and
system tests should be considered for larger changes.*

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9110, opened 2020-08-01
MINOR: Ensure a reason is logged for all segment deletion operations

dhruvilshah3:
This PR improves the logging for segment deletion to ensure that a reason is logged for segment deletions via all code paths. It also updates the logging so we can log a reason for an entire batch of deletions instead of logging one message per segment.

Sample log output:
```
[2020-08-05 11:56:35,826] INFO [Log partition=foo-0, dir=/tmp/kafka-logs] Deleting segment LogSegment(baseOffset=219030, size=1042252, lastModifiedTime=1596653795000, largestRecordTimestamp=Some(1596653795374)) due to retention time 1ms breach based on the largest record timestamp in the segment (kafka.log.Log)
[2020-08-05 11:56:35,826] INFO [Log partition=foo-0, dir=/tmp/kafka-logs] Deleting segment LogSegment(baseOffset=220035, size=1042252, lastModifiedTime=1596653795000, largestRecordTimestamp=Some(1596653795384)) due to retention time 1ms breach based on the largest record timestamp in the segment (kafka.log.Log)
...
```

### DIFF
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -2219,7 +2219,7 @@ class Log(@volatile private var _dir: File,
         // removing the deleted segment, we should force materialization of the iterator here, so that results of the
         // iteration remain valid and deterministic.
         val toDelete = segments.toList
-        info(s"${reason.logReason(this, toDelete)}")
+        reason.logReason(this, toDelete)
         toDelete.foreach { segment =>
           this.segments.remove(segment.baseOffset)
         }
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -1903,7 +1903,7 @@ class Log(@volatile private var _dir: File,
                  s"=max(provided offset = $expectedNextOffset, LEO = $logEndOffset) while it already " +
                  s"exists and is active with size 0. Size of time index: ${activeSegment.timeIndex.entries}," +
                  s" size of offset index: ${activeSegment.offsetIndex.entries}.")
-            removeAndDeleteSegments(Seq(activeSegment), asyncDelete = true, GenericDeletion)
+            removeAndDeleteSegments(Seq(activeSegment), asyncDelete = true, LogRollDeletion)
           } else {
             throw new KafkaException(s"Trying to roll a new log segment for topic partition $topicPartition with start offset $newOffset" +
                                      s" =max(provided offset = $expectedNextOffset, LEO = $logEndOffset) while it already exists. Existing " +
@@ -2086,7 +2086,7 @@ class Log(@volatile private var _dir: File,
             truncateFullyAndStartAt(targetOffset)
           } else {
             val deletable = logSegments.filter(segment => segment.baseOffset > targetOffset)
-            removeAndDeleteSegments(deletable, asyncDelete = true, LogTruncation)
+            removeAndDeleteSegments(deletable, asyncDelete = true, LogTruncateDeletion)
             activeSegment.truncateTo(targetOffset)
             updateLogEndOffset(targetOffset)
             updateLogStartOffset(math.min(targetOffset, this.logStartOffset))
@@ -2109,7 +2109,7 @@ class Log(@volatile private var _dir: File,
       debug(s"Truncate and start at offset $newOffset")
       lock synchronized {
         checkIfMemoryMappedBufferClosed()
-        removeAndDeleteSegments(logSegments, asyncDelete = true, LogTruncation)
+        removeAndDeleteSegments(logSegments, asyncDelete = true, LogTruncateDeletion)
         addSegment(LogSegment.open(dir,
           baseOffset = newOffset,
           config = config,
@@ -2720,14 +2720,14 @@ case object LogDeletion extends SegmentDeletionReason {
   }
 }
 
-case object LogTruncation extends SegmentDeletionReason {
+case object LogTruncateDeletion extends SegmentDeletionReason {
   override def reasonString(log: Log, segment: LogSegment): String = {
     s"Segment with base offset ${segment.baseOffset} will be deleted as part of log truncation"
   }
 }
 
-case object GenericDeletion extends SegmentDeletionReason {
+case object LogRollDeletion extends SegmentDeletionReason {
   override def reasonString(log: Log, segment: LogSegment): String = {
-    s"Segment with base offset ${segment.baseOffset} will be deleted"
+    s"Segment with base offset ${segment.baseOffset} will be deleted as part of log roll"
   }
 }
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -2219,7 +2219,7 @@ class Log(@volatile private var _dir: File,
         // removing the deleted segment, we should force materialization of the iterator here, so that results of the
         // iteration remain valid and deterministic.
         val toDelete = segments.toList
-        println(s"${reason.reasonString(this, toDelete)}")
+        info(s"${reason.reasonString(this, toDelete)}")
         toDelete.foreach { segment =>
           this.segments.remove(segment.baseOffset)
         }
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -797,7 +797,7 @@ class Log(@volatile private var _dir: File,
           warn(s"Corruption found in segment ${segment.baseOffset}, truncating to offset ${segment.readNextOffset}")
           removeAndDeleteSegments(unflushed.toList,
             asyncDelete = true,
-            reason = LogRecoveryDeletion)
+            reason = LogRecovery)
           truncated = true
         }
       }
@@ -810,7 +810,7 @@ class Log(@volatile private var _dir: File,
           "This could happen if segment files were deleted from the file system.")
         removeAndDeleteSegments(logSegments,
           asyncDelete = true,
-          reason = LogRecoveryDeletion)
+          reason = LogRecovery)
       }
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2227,14 +2210,17 @@ class Log(@volatile private var _dir: File,
...
+                                      asyncDelete: Boolean,
+                                      reason: SegmentDeletionReason): Unit = {
     if (segments.nonEmpty) {
       lock synchronized {
         // As most callers hold an iterator into the `segments` collection and `removeAndDeleteSegment` mutates it by
         // removing the deleted segment, we should force materialization of the iterator here, so that results of the
         // iteration remain valid and deterministic.
         val toDelete = segments.toList
         toDelete.foreach { segment =>
+          info(s"${reason.reasonString(this, segment)}")
```
kowshik: If we passed in the deletion reason further into the `deleteSegmentFiles` method, it seems we can print the reason string just once for a batch of segments being deleted. And within the reason string, we can provide the reason for deleting the batch:

https://github.com/confluentinc/ce-kafka/blob/master/core/src/main/scala/kafka/log/Log.scala#L2519
https://github.com/confluentinc/ce-kafka/blob/master/core/src/main/scala/kafka/log/Log.scala#L2526

ex: `info("Deleting segments due to $reason: ${segments.mkString(",")}"`

where `$reason` provides `due to retention time 1200000ms breach`.

The drawback is that sometimes we can not print segment-specific information since the error message is at a batch level. But generally it may be that segment-level deletion information could bloat our server logging, so it may be better to batch the logging instead.

What are your thoughts?

dhruvilshah3: While verbose, I think having the granularity of each segment is useful. This allows us to easily reason about why a particular segment was deleted. Note that we switched from a single log per batch to a log per segment in https://github.com/apache/kafka/pull/8850.

ijuma: Can we think about cases where this could be an issue? Say delete records is used, causing a large number of segments to be deleted Could that trigger excessive logging?

kowshik: @dhruvilshah3 Sounds good!

dhruvilshah3: @ijuma We log one message per deleted segment. This could cause temporary increase in log volume when DeleteRecords is used or when retention is lowered, for example.

Overall, we have a few options with different tradeoffs:

1. Log a common reason per batch being deleted, including base offsets of segments being deleted. eg.
```
Deleting segments due to retention time 999ms breach. BaseOffsets: (0,5,...).
```

2. Log a common reason per batch being deleted, including base offsets and metadata of segments. eg. 
```
Deleting segments due to retention time 999ms breach: LogSegment(baseOffset=0, size=360, lastModifiedTime=1596387738000, largestRecordTimestamp=Some(1596387737414)),LogSegment(baseOffset=5, size=360, lastModifiedTime=1596387738000, largestRecordTimestamp=Some(1596387737414)),...
```

3. Log one message per segment being deleted. This is the current behavior. eg.
```
Segment with base offset 0 will be deleted due to retention time 999ms breach based on the largest record timestamp from the segment, which is ...
Segment with base offset 5 will be deleted due to retention time 999ms breach based on the largest record timestamp from the segment, which is ...
...
```

Doing (2) may be a reasonable tradeoff. It eliminates some of the redundancy at the cost of making it to glean per segment metadata. Let me know what you think.

core/src/main/scala/kafka/log/LogSegment.scala
```diff
@@ -413,7 +413,7 @@ class LogSegment private[log] (val log: FileRecords,
   override def toString: String = "LogSegment(baseOffset=" + baseOffset +
     ", size=" + size +
     ", lastModifiedTime=" + lastModified +
-    ", largestTime=" + largestTimestamp +
+    ", largestRecordTimestamp=" + largestRecordTimestamp +
```
kowshik, line 416: Should we keep `largestTime` from LHS, and, print both `largestRecordTimestamp` and `largestTime` ?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2686,3 +2670,50 @@ object LogMetricNames {
...
+case object RetentionMsBreachDeletion extends SegmentDeletionReason {
+  override def reasonString(log: Log, toDelete: Iterable[LogSegment]): String = {
+    s"Deleting segments due to retention time ${log.config.retentionMs}ms breach: ${toDelete.mkString(",")}"
+  }
+}
+
+case object RetentionSizeBreachDeletion extends SegmentDeletionReason {
+  override def reasonString(log: Log, toDelete: Iterable[LogSegment]): String = {
+    s"Deleting segments due to retention size ${log.config.retentionSize} breach. " +
+      s"Current log size is ${log.size}. ${toDelete.mkString(",")}"
```
kowshik: nit: replace "." with ":" 

`s"Current log size is ${log.size}: ${toDelete.mkString(",")}"`

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2227,13 +2210,16 @@ class Log(@volatile private var _dir: File,
...
+  private def removeAndDeleteSegments(segments: Iterable[LogSegment],
+                                      asyncDelete: Boolean,
+                                      reason: SegmentDeletionReason): Unit = {
     if (segments.nonEmpty) {
       lock synchronized {
         // As most callers hold an iterator into the `segments` collection and `removeAndDeleteSegment` mutates it by
         // removing the deleted segment, we should force materialization of the iterator here, so that results of the
         // iteration remain valid and deterministic.
         val toDelete = segments.toList
+        println(s"${reason.reasonString(this, toDelete)}")
```
kowshik: Did you mean to use `info` level logging?

kowshik (approved):
Looks good from my side, small comments inline.
It may be good for others to have a look also. cc @ijuma

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2227,13 +2210,16 @@ class Log(@volatile private var _dir: File,
...
+  private def removeAndDeleteSegments(segments: Iterable[LogSegment],
+                                      asyncDelete: Boolean,
+                                      reason: SegmentDeletionReason): Unit = {
     if (segments.nonEmpty) {
       lock synchronized {
         // As most callers hold an iterator into the `segments` collection and `removeAndDeleteSegment` mutates it by
         // removing the deleted segment, we should force materialization of the iterator here, so that results of the
         // iteration remain valid and deterministic.
         val toDelete = segments.toList
+        info(s"${reason.reasonString(this, toDelete)}")
```
hachikuji: A little annoying to need to pass through segments just to be added to each log message individually. Maybe we could do it like this instead
```scala
info(s"Deleting segments due to ${reason.reasonString(this)}: $toDelete")
```

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2686,3 +2670,50 @@ object LogMetricNames {
     List(NumLogSegments, LogStartOffset, LogEndOffset, Size)
   }
 }
+
+sealed trait SegmentDeletionReason {
+  def reasonString(log: Log, toDelete: Iterable[LogSegment]): String
+}
+
+case object RetentionMsBreachDeletion extends SegmentDeletionReason {
```
hachikuji: nit: is it necessary to add `Deletion` to all of these? Maybe only `LogDeletion` needs it since it is referring to deletion of the log itself.

core/src/main/scala/kafka/log/LogSegment.scala
hachikuji: I'm ok with the change. I think it's better to reflect the underlying fields directly and redundant information just adds noise to the logs.

dhruvilshah3:
Thanks for the reviews @kowshik, @hachikuji, @ijuma. I have addressed the comments. Let me know what you think.

ijuma:
ok to test

core/src/main/scala/kafka/log/Log.scala
hachikuji: I think it is important to capture the segment level details. In the past, we have had trouble explaining precisely why a specific segment got deleted. For example, was it because of the last modified time or the record timestamp? When users are looking to understand why data is deleted, we should be able to provide a clear answer.

My personal preference is probably 3) because I hate dealing with lists of things in log messages. Simple grepping no longer work to extract the details. Big messages also messes up console scrolling and can choke downstream systems. For segments, I am not so worried about log noise because the rate of segment creation is not _that_ high.

dhruvilshah3: I think this is reasonable. Logging a segment per line will make it easier for us to diagnose issues. I made the change to log a segment per line for retention-related deletions. We still batch all segments in a single line for all other deletion events, eg. log deletion, truncation, etc.

hachikuji:
retest this please

hachikuji:
retest this please

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2227,13 +2210,16 @@ class Log(@volatile private var _dir: File,
...
+  private def removeAndDeleteSegments(segments: Iterable[LogSegment],
+                                      asyncDelete: Boolean,
+                                      reason: SegmentDeletionReason): Unit = {
     if (segments.nonEmpty) {
       lock synchronized {
         // As most callers hold an iterator into the `segments` collection and `removeAndDeleteSegment` mutates it by
         // removing the deleted segment, we should force materialization of the iterator here, so that results of the
         // iteration remain valid and deterministic.
         val toDelete = segments.toList
+        info(s"${reason.logReason(this, toDelete)}")
```
hachikuji: Don't we need to get rid of the call to `info` here?

dhruvilshah3: Thanks for catching. I fixed this.

hachikuji:
retest this please

hachikuji:
retest this please

hachikuji:
retest this please

hachikuji:
retest this please

hachikuji (approved):
LGTM. Thanks for the patch!

PR #9111, opened 2020-08-02
KAFKA-10337: await async commits in commitSync even if no offsets given

thomaslee:
The contract for `commitSync()` guarantees that the callbacks for all
prior async commits will be invoked before it (successfully?) returns.
Prior to this change the contract could be violated if an empty offsets
map were passed in to `commitSync()`.

Also added a unit test in `ConsumerCoordinatorTest` exercising this particular path.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/ConsumerCoordinator.java
@@ -36,11 +36,11 @@ import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.FencedInstanceIdException;
 import org.apache.kafka.common.errors.GroupAuthorizationException;
 import org.apache.kafka.common.errors.InterruptException;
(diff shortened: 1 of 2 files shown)


abbccdda (commented):
Thanks for the PR. I'm actually wondering whether it's valid to commit an empty map at any given time, cc @hachikuji

thomaslee:
It's somewhat easy to do it accidentally. The scenario I was concerned about looks something like this with auto offset commits disabled:

```java
KafkaConsumer<...> consumer = ...;
Map<TopicPartition, OffsetAndMetadata> offsets = new HashMap<>();

//
// in a message processing loop
//
offsets.put(topicPartition, new OffsetAndMetadata(offset));
consumer.commitAsync(offsets);
offsets = new HashMap<>();

//
// before a rebalance (e.g. onPartitionsRevoked)
//
consumer.commitSync(offsets);
```

Per the docs the `commitSync` call should ensure that the async commit finishes prior to the rebalance, but it may not if `offsets` is empty here.

PR #9112, opened 2020-08-02
KAFKA-10312 Fix error code returned by getPartitionMetadata

RamanVerma:
MetadataCache#getPartitionMetadata returns an error when the topic's leader Id
is present at MetadataCache but listener endpoint is not present for this leader.
For older version, LEADER_NOT_AVAILABLE is returned while LISTENER_NOT_FOUND is
returned for new metadata version.

getPartitionMetadata was looking up MetadataCache's host brokerId rather
than the topic's leader id while determining what error to return. This
could result in the call returning LISTENER_NOT_FOUND when it should
have returned LEADER_NOT_AVAILABLE. This commit corrects this behavior.

Unit tests were already present to test out the error codes returned
under different situations but they were giving out a false positive.
The test was using same broker id for both the MetadataCache's host as
well as for the topic's leader. Error manifests when the MetadataCache's
host id is changed. Improved the test.

This commit also consolidated couple of related tests to reduce code
duplication.

comment to ping reviewers.*

### DIFF
--- a/core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
+++ b/core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
@@ -154,7 +154,7 @@ class MetadataCacheTest {
         .setPort(9092)
         .setSecurityProtocol(securityProtocol.id)
         .setListener(listenerName.value)).asJava))
-    val metadataCacheBrokerId = 9
+    val metadataCacheBrokerId = 0
     // leader is not available. expect LEADER_NOT_AVAILABLE for any metadata version.
     verifyTopicMetadataPartitionLeaderOrEndpointNotAvailable(metadataCacheBrokerId, brokers, listenerName,
       leader = 1, Errors.LEADER_NOT_AVAILABLE, errorUnavailableListeners = false)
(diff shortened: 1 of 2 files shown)


RamanVerma:
Ran `./gradlew core:test` locally. All tests passed.

RamanVerma:
cc @hachikuji 
Not sure why I am not able to add anyone as a reviewer to the PR. I am not getting the add reviewers/recommended reviewers options.

hachikuji:
ok to test

core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
```diff
@@ -229,7 +205,7 @@ class MetadataCacheTest {
                                                                        errorUnavailableListeners: Boolean): Unit = {
     val topic = "topic"
 
-    val cache = new MetadataCache(1)
+    val cache = new MetadataCache(9)
```
hachikuji: nit: maybe it would be clearer if we passed `brokerId` as an argument? Then we could use broker 0 in the test case, for example, so that the UpdateMetadata request makes more sense.

hachikuji (commented):
Good catch. Left a minor suggestion.

core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
RamanVerma: @hachikuji done

core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
```diff
@@ -187,49 +191,24 @@ class MetadataCacheTest {
       new UpdateMetadataBroker()
         .setId(1)
         .setEndpoints(broker1Endpoints.asJava))
-    verifyTopicMetadataPartitionLeaderOrEndpointNotAvailable(brokers, sslListenerName,
+    val metadataCacheBrokerId = 9
```
hachikuji: Would 0 work just as well? It would make the test more coherent if we used one of the ids included in the UpdateMetadata request.

RamanVerma: Done

hachikuji:
retest this please

PR #9114, opened 2020-08-03
KAFKA-10162; Use Token Bucket algorithm for controller mutation quota (KIP-599, Part III)

dajac:
Based on the discussion in https://github.com/apache/kafka/pull/9072, I have put together an alternative way. This one does the following:
* Instead of changing the implementation of the Rate to behave like a Token Bucket, it actually use two different metrics: the regular Rate and a new Token Bucket. The latter is used to enforce the quota.
* The Token Bucket algorithm uses the rate of the quota as the refill rate for the credits and compute the burst based on the number of samples and their length (# samples * sample length * quota).
* The Token Bucket algorithm used can go under zero in order to handle unlimited burst (e.g. create topic with a number of partitions higher than the burst). Throttling kicks in when the number of credits is under zero.
* The throttle time is computed as credits under zero / refill rate (or quota).
* Only the controller mutation uses it for now.
* The remaining number of credits in the bucket is exposed with the `tokens` metrics per user/clientId.

The code can be improved and refactored. I just wanted to get out quickly to get feedback about the approach.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
@@ -16,8 +16,10 @@
  */
 package org.apache.kafka.common.metrics;
 
+import java.util.HashMap;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.metrics.CompoundStat.NamedMeasurable;
+import org.apache.kafka.common.metrics.stats.TokenBucket;
 import org.apache.kafka.common.utils.Time;
 
 import java.util.ArrayList;
@@ -44,6 +46,7 @@ public final class Sensor {
     private final String name;
     private final Sensor[] parents;
     private final List<Stat> stats;
+    private final Map<Stat, MetricConfig> configPerStat;
     private final Map<MetricName, KafkaMetric> metrics;
     private final MetricConfig config;
     private final Time time;
@@ -109,6 +112,7 @@ public final class Sensor {
         this.parents = parents == null ? new Sensor[0] : parents;
         this.metrics = new LinkedHashMap<>();
         this.stats = new ArrayList<>();
+        this.configPerStat = new HashMap<>();
         this.config = config;
         this.time = time;
         this.inactiveSensorExpirationTimeMs = TimeUnit.MILLISECONDS.convert(inactiveSensorExpirationTimeSeconds, TimeUnit.SECONDS);
@@ -199,8 +203,10 @@ public final class Sensor {
         synchronized (this) {
             synchronized (metricLock()) {
                 // increment all the stats
-                for (Stat stat : this.stats)
+                for (Stat stat : this.stats) {
+                    MetricConfig config = this.configPerStat.getOrDefault(stat, this.config);
                     stat.record(config, value, timeMs);
+                }
             }
             if (checkQuotas)
                 checkQuotas(timeMs);
@@ -223,8 +229,14 @@ public final class Sensor {
                 Quota quota = config.quota();
                 if (quota != null) {
                     double value = metric.measurableValue(timeMs);
-                    if (!quota.acceptable(value)) {
-                        throw new QuotaViolationException(metric, value, quota.bound());
+                    if (metric.measurable() instanceof TokenBucket) {
+                        if (value <= 0) {
+                            throw new QuotaViolationException(metric, value, quota.bound());
+                        }
+                    } else {
+                        if (!quota.acceptable(value)) {
+                            throw new QuotaViolationException(metric, value, quota.bound());
+                        }
                     }
                 }
             }
@@ -252,6 +264,8 @@ public final class Sensor {
             return false;
 
         this.stats.add(Objects.requireNonNull(stat));
+        if (config != null)
+            this.configPerStat.put(stat, config);
         Object lock = metricLock();
         for (NamedMeasurable m : stat.stats()) {
             final KafkaMetric metric = new KafkaMetric(lock, m.name(), m.stat(), config == null ? this.config : config, time);
@@ -297,6 +311,8 @@ public final class Sensor {
             registry.registerMetric(metric);
             metrics.put(metric.metricName(), metric);
             stats.add(stat);
+            if (config != null)
+                this.configPerStat.put(stat, config);
             return true;
         }
     }
--- a/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
@@ -16,7 +16,6 @@
  */
 package org.apache.kafka.common.metrics;
 
-import java.util.HashMap;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.metrics.CompoundStat.NamedMeasurable;
 import org.apache.kafka.common.metrics.stats.TokenBucket;
@@ -45,14 +44,23 @@ public final class Sensor {
     private final Metrics registry;
     private final String name;
     private final Sensor[] parents;
-    private final List<Stat> stats;
-    private final Map<Stat, MetricConfig> configPerStat;
+    private final List<StatAndConfig> stats;
     private final Map<MetricName, KafkaMetric> metrics;
     private final MetricConfig config;
     private final Time time;
     private volatile long lastRecordTime;
     private final long inactiveSensorExpirationTimeMs;
     private final Object metricLock;
+
+    private static class StatAndConfig {
+        public final Stat stat;
+        public final MetricConfig config;
+
+        StatAndConfig(Stat stat, MetricConfig config) {
+            this.stat = stat;
+            this.config = config;
+        }
+    }
 
     public enum RecordingLevel {
         INFO(0, "INFO"), DEBUG(1, "DEBUG");
@@ -112,7 +120,6 @@ public final class Sensor {
         this.parents = parents == null ? new Sensor[0] : parents;
         this.metrics = new LinkedHashMap<>();
         this.stats = new ArrayList<>();
-        this.configPerStat = new HashMap<>();
         this.config = config;
         this.time = time;
         this.inactiveSensorExpirationTimeMs = TimeUnit.MILLISECONDS.convert(inactiveSensorExpirationTimeSeconds, TimeUnit.SECONDS);
@@ -203,9 +210,8 @@ public final class Sensor {
         synchronized (this) {
             synchronized (metricLock()) {
                 // increment all the stats
-                for (Stat stat : this.stats) {
-                    MetricConfig config = this.configPerStat.getOrDefault(stat, this.config);
-                    stat.record(config, value, timeMs);
+                for (StatAndConfig statAndConfig : this.stats) {
+                    statAndConfig.stat.record(statAndConfig.config, value, timeMs);
                 }
             }
             if (checkQuotas)
@@ -230,7 +236,7 @@ public final class Sensor {
                 if (quota != null) {
                     double value = metric.measurableValue(timeMs);
                     if (metric.measurable() instanceof TokenBucket) {
-                        if (value <= 0) {
+                        if (value < 0) {
                             throw new QuotaViolationException(metric, value, quota.bound());
                         }
                     } else {
@@ -263,12 +269,11 @@ public final class Sensor {
         if (hasExpired())
             return false;
 
-        this.stats.add(Objects.requireNonNull(stat));
-        if (config != null)
-            this.configPerStat.put(stat, config);
+        final MetricConfig statConfig = config == null ? this.config : config;
+        stats.add(new StatAndConfig(Objects.requireNonNull(stat), statConfig));
         Object lock = metricLock();
         for (NamedMeasurable m : stat.stats()) {
-            final KafkaMetric metric = new KafkaMetric(lock, m.name(), m.stat(), config == null ? this.config : config, time);
+            final KafkaMetric metric = new KafkaMetric(lock, m.name(), m.stat(), statConfig, time);
             if (!metrics.containsKey(metric.metricName())) {
                 registry.registerMetric(metric);
                 metrics.put(metric.metricName(), metric);
@@ -301,18 +306,17 @@ public final class Sensor {
         } else if (metrics.containsKey(metricName)) {
             return true;
         } else {
+            final MetricConfig statConfig = config == null ? this.config : config;
             final KafkaMetric metric = new KafkaMetric(
                 metricLock(),
                 Objects.requireNonNull(metricName),
                 Objects.requireNonNull(stat),
-                config == null ? this.config : config,
+                statConfig,
                 time
             );
             registry.registerMetric(metric);
             metrics.put(metric.metricName(), metric);
-            stats.add(stat);
-            if (config != null)
-                this.configPerStat.put(stat, config);
+            stats.add(new StatAndConfig(Objects.requireNonNull(stat), statConfig));
             return true;
         }
     }
--- a/clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
@@ -33,6 +33,27 @@ import static org.apache.kafka.common.metrics.internals.MetricsUtils.convert;
  *
  * The quota is considered as exhausted when the amount of remaining credits in the bucket
  * is below zero. The enforcement is done by the {@link org.apache.kafka.common.metrics.Sensor}.
+ *
+ * Token Bucket vs Rate based Quota:
+ * The current sampled rate based quota does not cope well with bursty workloads. The issue is
+ * that a unique and large sample can hold the average above the quota and this until it is
+ * discarded. Practically, when this happens, one must wait until the sample is expired to
+ * bring the rate below the quota even though less time would be theoretically required. As an
+ * examples, let's imagine that we have:
+ * - Quota (Q)   = 5
+ * - Samples (S) = 100
+ * - Window (W)  = 1s
+ * A burst of 560 brings the average rate (R) to 5.6 (560 / 100). The throttle time is computed as
+ * follow: ((R - Q / Q * S * W)) = ((5.6 - 5) / 5 * 100 * 1) = 12 secs. In practice, the average
+ * rate won't go below the quota before the burst is dropped from the samples so one must wait
+ * 100s (S * W).
+ *
+ * The token bucket relies on continuously updated amount of credits. Therefore, it does not
+ * suffers from the above issue. The same example would work as follow:
+ * - Quota (Q) = 5
+ * - Burst (B) = 5 * 1 * 100 = 500 (Q * S * W)
+ * A burst of 560 brings the amount of credits to -60. One must wait 12s (-(-60)/5) to refill the
+ * bucket to zero.
  */
 public class TokenBucket implements MeasurableStat {
     private final TimeUnit unit;
--- a/clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
@@ -22,7 +22,7 @@ import org.apache.kafka.common.metrics.MetricConfig;
 
 public class TokenBucket implements MeasurableStat {
     private final TimeUnit unit;
-    private double credits;
+    private double tokens;
     private long lastUpdateMs;
 
     public TokenBucket() {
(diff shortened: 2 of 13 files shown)


clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,83 @@
...
+        refill(quota, burst, timeMs);
+        return this.credits;
+    }
+
+    @Override
+    public void record(final MetricConfig config, final double value, final long timeMs) {
+        if (config.quota() == null)
+            return;
+        final double quota = config.quota().bound();
+        final double burst = (config.samples() - 1) * convert(config.timeWindowMs()) * quota;
```
junrao: Should burst be computed from `#samples` or `#samples - 1` ?

core/src/main/scala/kafka/server/ControllerMutationQuotaManager.scala
```diff
@@ -131,6 +133,16 @@ class PermissiveControllerMutationQuota(private val time: Time,
 
 object ControllerMutationQuotaManager {
   val QuotaControllerMutationDefault = Int.MaxValue.toDouble
+
+  def throttleTime(e: QuotaViolationException, timeMs: Long): Long = {
+    e.metric().measurable() match {
+      case _: TokenBucket => Math.round(-e.value() * e.bound())
```
junrao: Hmm, assuming bound is the per sec rate, it seems that the throttleTimeMs should be `-e.value() / e.bound * 1000`?

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,83 @@
...
+    private double credits;
+    private long lastUpdateMs;
+
+    public TokenBucket() {
+        this(TimeUnit.SECONDS);
+    }
+
+    public TokenBucket(TimeUnit unit) {
+        this.unit = unit;
+        this.credits = 0;
```
junrao: Should we start with 0 credit or the full burst credits? The benefit of the latter is that during initialization, the requests won't be throttled as much due to a cold start.

junrao (commented):
@dajac : Thanks for the PR. Took a quick look. It does look like this approach is less intrusive and probably easier to understand. A few quick comments below.

clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
```diff
@@ -223,8 +229,14 @@ public void checkQuotas(long timeMs) {
                 Quota quota = config.quota();
                 if (quota != null) {
                     double value = metric.measurableValue(timeMs);
-                    if (!quota.acceptable(value)) {
-                        throw new QuotaViolationException(metric, value, quota.bound());
+                    if (metric.measurable() instanceof TokenBucket) {
+                        if (value <= 0) {
```
apovzner: we probably just want `value < 0` check here, right? Otherwise the throttle time will be 0 anyways.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
apovzner: If we want the burst to be more similar to original behavior, it seems like this should be `#samples`. With the current implementation, we can do 1 unit of work in the oldest window and then accept a burst right at the end of the last (not full yet) sample. Which means that the max burst size is almost at #samples * quota (if sample = 1 sec, quota is in units/second). Does this sound right to you?

Also, I think we should take into account `config.timeWindowMs`, because it could be something other than 1 second.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
apovzner: I also think it's better to start with burst size credits. We would normally create the sensor on the first byte/request/etc. and before that, the <user,client> is idle which means they were "accumulating credits".

apovzner:
I also took a quick look and I agree that this implementation is cleaner and definitely easier to understand.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: That's a very good question. I use `#samples - 1` here because we do so in `Rate#windowSize` to compute the total window so it is probably closer like this.

@apovzner `config.timeWindowMs` is already in the formula or did I misunderstand your comment?

core/src/main/scala/kafka/server/ControllerMutationQuotaManager.scala
dajac: Yes, that's right.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: The burst is not know when the `TokenBucket` is constructed as we must get the `MetricConfig` at least once to initialise it correctly. Note that, it will be initialised to the maximum burst when we refill the bucket when either `measure` or `record` is called for the first time. That works because `lastUpdateMs` is `0` so the delta between `now` and it will make the bucket full. So, the bucket actually stats from the maximum burst.

clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
dajac: Yes, that makes sense.

dajac:
@junrao @apovzner I have updated the PR to address your comments. I have also refactored the code a bit and updated all the tests of the controller quota manager to work with the Token Bucket.

From a public api perspective, we would have a new metric per user/clientId named: `tokens`. It represents the number of tokens in the bucket.

Please, pay close attention to the unit tests of the controller mutation quota manager. Most of the throttle times have been updated there to match the token bucket usage. This is due to the fact that the token bucket does not have to wait until old samples are purged to get it amount of tokens updated. It is updated continuously.

I wonder if we should use by default strait away of put it behind a config flag to let user opt in. It does not matter much for our new quota but that may matter if/when we migrate existing quotas. What do you think?

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,87 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.common.metrics.stats;
+
+import java.util.concurrent.TimeUnit;
+import org.apache.kafka.common.metrics.MeasurableStat;
+import org.apache.kafka.common.metrics.MetricConfig;
+
+public class TokenBucket implements MeasurableStat {
```
junrao, line 58: Could we add a high level description of the class?

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,87 @@
...
+        this.tokens = Math.min(burst, this.tokens - value);
+    }
+
+    private void refill(final double quota, final double burst, final long timeMs) {
+        this.tokens = Math.min(burst, this.tokens + quota * convert(timeMs - lastUpdateMs));
+        this.lastUpdateMs = timeMs;
+    }
+
+    private double burst(final MetricConfig config) {
+        return (config.samples() - 1) * convert(config.timeWindowMs()) * config.quota().bound();
```
junrao: Rate actually allows the windowSize to be close to the full samples * perSampleWindow. The logic around `config.samples() - 1` is just to make sure the windowSize contains at least that many full windows. So, to match that behavior, it seems that burst should use `config.samples()`.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    private double convert(final long timeMs) {
+        switch (unit) {
```
junrao: This code is duplicated from Rate. Could we reuse it somehow?

core/src/main/scala/kafka/server/ControllerMutationQuotaManager.scala
```diff
@@ -131,6 +133,22 @@ class PermissiveControllerMutationQuota(private val time: Time,
...
   val QuotaControllerMutationDefault = Int.MaxValue.toDouble
+
+  /**
+   * This calculates the amount of time needed to bring the TokenBucket within quota
+   * assuming that no new metrics are recorded.
+   *
+   * Basically, if a value < 0 is observed, the time required to bring it to zero is
+   * -value / refill rate (quota bound) * 1000.
+   */
+  def throttleTime(e: QuotaViolationException, timeMs: Long): Long = {
```
junrao: throttleTime => throttleTimeMs ?

core/src/main/scala/kafka/server/ControllerMutationQuotaManager.scala
```diff
@@ -156,6 +174,24 @@ class ControllerMutationQuotaManager(private val config: ClientQuotaManagerConfi
       quotaMetricTags.asJava)
   }
 
+  protected def clientTokenBucketMetricName(quotaMetricTags: Map[String, String]): MetricName = {
```
junrao: Could this be private?

clients/src/test/java/org/apache/kafka/common/metrics/SensorTest.java
```diff
@@ -209,4 +212,95 @@ public void shouldReturnPresenceOfMetrics() {
...
+        assertEquals(0, tkMetric.measurableValue(time.milliseconds()), 0.1);
+        strictRecord(sensor, 30, time.milliseconds());
+        assertEquals(-30, tkMetric.measurableValue(time.milliseconds()), 0.1);
+
+        metrics.close();
+    }
+
+    private void strictRecord(Sensor sensor, double value, long timeMs) {
+        synchronized (sensor) {
+            sensor.checkQuotas(timeMs);
```
junrao, line 275: sensor.record() always calls checkQuotas(). Why do we need to call it explicitly here?

clients/src/test/java/org/apache/kafka/common/metrics/SensorTest.java
```diff
@@ -209,4 +212,95 @@ public void shouldReturnPresenceOfMetrics() {
...
+        final MetricConfig stat2Config = new MetricConfig().quota(Quota.upperBound(10));
+        sensor.add(stat2Name, stat2, stat2Config);
+
+        sensor.record(10, 1);
+        Mockito.verify(stat1).record(stat1Config, 10, 1);
+        Mockito.verify(stat2).record(stat2Config, 10, 1);
+
+        Mockito.when(stat1.measure(stat1Config, 2)).thenReturn(2.0);
+        Mockito.when(stat2.measure(stat2Config, 2)).thenReturn(2.0);
+        sensor.checkQuotas(2);
```
junrao, line 302: Here, we are just verifying there is no quota exception?

clients/src/test/java/org/apache/kafka/common/metrics/TokenBucketTest.java
```diff
@@ -0,0 +1,93 @@
...
+            .quota(Quota.upperBound(5))
+            .timeWindow(2, TimeUnit.SECONDS)
+            .samples(11);
+
+        TokenBucket tk = new TokenBucket();
+
+        // Expect 100 credits at T
+        assertEquals(100, tk.measure(config, time.milliseconds()), 0.1);
+
+        // Record 60 at T, expect 100 credits
```
junrao: We are recording -60.

clients/src/test/java/org/apache/kafka/common/metrics/TokenBucketTest.java
```diff
@@ -0,0 +1,93 @@
...
+        tk.record(config, -60, time.milliseconds());
+        assertEquals(100, tk.measure(config, time.milliseconds()), 0.1);
+
+        // Advance by 2s, record 60, expect 40 credits
+        time.sleep(2000);
+        tk.record(config, 60, time.milliseconds());
+        assertEquals(40, tk.measure(config, time.milliseconds()), 0.1);
+
+        // Advance by 2s, record 60, expect 100 credits
```
junrao: We are recording -60.

junrao (commented):
@dajac : Thanks for the updated PR. Looks good overall. Just a few comments below.

clients/src/test/java/org/apache/kafka/common/metrics/SensorTest.java
dajac: In the above two tests, I simulate a "strict quotas" in the sense that recording is not allowed if the quota is already violated. Therefore, I check it before recording the value.

clients/src/test/java/org/apache/kafka/common/metrics/SensorTest.java
dajac: No. This test actually reproduce a bug that I have found. Basically, a stat can be added to the Sensor with a MetricsConfig but the Sensor was not using the provided one when recording a value but was using the one of the Sensor all the time. This test verifies that the correct config is used both for recording and measuring via calling checkQuota.

core/src/main/scala/kafka/server/ControllerMutationQuotaManager.scala
dajac: Yes.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: Sure, makes sense.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: Ack. I misunderstood it. `config.samples()` sounds good to me.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: Sure.

dajac:
@junrao Thanks for your comments. I just pushed an update which addressed them.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * The {@link TokenBucket} is a {@link MeasurableStat} implementing a token bucket algorithm
+ * that is usable within a {@link org.apache.kafka.common.metrics.Sensor}.
+ *
+ * The {@link Quota#bound()} defined the refill rate of the bucket while the maximum burst or
+ * the maximum number of credits of the bucket is defined by
+ * {@link MetricConfig#samples() * MetricConfig#timeWindowMs() * Quota#bound()}.
+ *
+ * The quota is considered as exhausted when the amount of remaining credits in the bucket
```
junrao, line 34: Could we document how this quota behaves differently from existing quota?

clients/src/test/java/org/apache/kafka/common/metrics/TokenBucketTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Before
+    public void setup() {
+        time = new MockTime(0, System.currentTimeMillis(), System.nanoTime());
+    }
+
+    @Test
+    public void testRecord() {
+        // Rate  = 5 unit / sec
+        // Burst = 2 * (11 - 1) = 20 units
```
junrao: we have 10 samples now.

clients/src/test/java/org/apache/kafka/common/metrics/TokenBucketTest.java
```diff
@@ -0,0 +1,93 @@
...
+        // Advance by 2s, record 60, expect -5 credits
+        time.sleep(2000);
+        tk.record(config, 60, time.milliseconds());
+        assertEquals(-5, tk.measure(config, time.milliseconds()), 0.1);
+    }
+
+    @Test
+    public void testUnrecord() {
+        // Rate  = 5 unit / sec
+        // Burst = 2 * (11 - 1) = 20 units
```
junrao: we have 10 samples now.

clients/src/test/java/org/apache/kafka/common/metrics/SensorTest.java
```diff
@@ -209,4 +212,95 @@ public void shouldReturnPresenceOfMetrics() {
...
     }
+
+    @Test
+    public void testStrictQuotaEnforcementWithRate() {
+        final Time time = new MockTime(0, System.currentTimeMillis(), 0);
+        final Metrics metrics = new Metrics(time);
+        final Sensor sensor = metrics.sensor("sensor", new MetricConfig()
+            .quota(Quota.upperBound(2))
+            .timeWindow(1, TimeUnit.SECONDS)
+            .samples(11));
```
junrao, line 223: Is the test based on 10 samples?

junrao (commented):
@dajac : Thanks for the updated PR. Just a few more minor comments below.

junrao:
ok to test

clients/src/test/java/org/apache/kafka/common/metrics/SensorTest.java
dajac: The config is correct. 11 samples. With the few samples in the test, the total window is actually 10s. This is why I use 10 in the formulas.

dajac:
@junrao Thanks. I have updated the PR.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,101 @@
...
+ * is below zero. The enforcement is done by the {@link org.apache.kafka.common.metrics.Sensor}.
+ *
+ * Token Bucket vs Rate based Quota:
+ * The current sampled rate based quota does not cope well with bursty workloads. The issue is
+ * that a unique and large sample can hold the average above the quota and this until it is
```
junrao: "this until it is" doesn't quite parse.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
```diff
@@ -0,0 +1,101 @@
...
+ * discarded. Practically, when this happens, one must wait until the sample is expired to
+ * bring the rate below the quota even though less time would be theoretically required. As an
+ * examples, let's imagine that we have:
+ * - Quota (Q)   = 5
+ * - Samples (S) = 100
+ * - Window (W)  = 1s
+ * A burst of 560 brings the average rate (R) to 5.6 (560 / 100). The throttle time is computed as
```
junrao: "The throttle time " : I guess this is the expected throttle time?

junrao (commented):
@dajac : Thanks for the updated PR. A couple of more minor comments.

junrao:
ok to test

junrao:
retest this please

junrao:
test this please

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: i have removed the "and this". Does it parse better?

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
dajac: Yes, that is correct. I have added "expected" to be clearer.

junrao (approved):
@dajac : Thanks for the latest update. LGTM

For your earlier question, if this works out well. We could just change existing quotas' implementation accordingly without necessarily a new flag to opt in.

Could you also update the KIP and the email thread with the newly added token metric to make sure there is no objection to this minor change?

junrao:
retest this please

junrao:
test this please

dajac:
Sure. I will update the KIP and the mailing list tomorrow. Thanks, Jun!

junrao:
The failed unit test seems unrelated.

@apovzner : Do you have any other comments on this? Thanks.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
apovzner: ah right, `lastUpdateMs` will make sure that bucket would be full on the first `record()`.

clients/src/main/java/org/apache/kafka/common/metrics/stats/TokenBucket.java
apovzner: hm.. maybe I just did not notice `config.timeWindowMs`. All good now.

apovzner (approved):
Looks great!

dajac:
@junrao I have updated the KIP and posted an update to the thread in the mailing list.

PR #9123, opened 2020-08-04
MINOR: optimize fetchablePartitions check by performing cheap check first

lbradstreet:
When the consumer fetches a lot of partitions at once the isAvailable
check can become expensive due to a high number of contains check. We
might as well perform the isFetchable check first as this is clearly much
cheaper.

This will help most with partitions that are paused or are otherwise not
in a state where they are fetchable.


lbradstreet:
image

From a profile fetching from 2500 partitions. This section of code takes 7.85%. In the happy case I would not expect this to improve, but we should see an improvement when partitions are unfetchable.

PR #9124, opened 2020-08-04
MINOR: add additional shutdown log info

abbccdda:
The purpose of this PR is to both enhance the visibility of shutdown progress for Kafka, as well as for the debugging purpose of a delayed process in between socket server close and eventually kafka server close.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/metrics/Metrics.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/Metrics.java
@@ -669,6 +669,7 @@ public class Metrics implements Closeable {
 
         for (MetricsReporter reporter : reporters) {
             try {
+                log.info("Closing reporter {}", reporter.getClass().getName());
                 reporter.close();
             } catch (Exception e) {
                 log.error("Error when closing " + reporter.getClass().getName(), e);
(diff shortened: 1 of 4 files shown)


clients/src/main/java/org/apache/kafka/common/metrics/Metrics.java
```diff
@@ -673,6 +673,8 @@ public void close() {
                 log.error("Error when closing " + reporter.getClass().getName(), e);
             }
         }
+
+        log.info("Metrics closed");
```
hachikuji: Would it be reasonable to add a message before closing each individual reporter?

core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
```diff
@@ -225,6 +225,7 @@ class DynamicBrokerConfig(private val kafkaConfig: KafkaConfig) extends Logging
     dynamicDefaultConfigs.clear()
     reconfigurables.clear()
     brokerReconfigurables.clear()
+    info(s"Dynamic configs cleared")
```
hachikuji: nit: this one is probably overkill since we're just clearing collections.

hachikuji (approved):
LGTM

PR #9128, opened 2020-08-05
KAFKA-7540 reduce session timeout to evict dead member in time and so…

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-7540

In short, the LEAVE_GROUP is sent to a shutdown broker so the dead member which is left in the group obstructs broker from completing rebalance.

The test case creates two consumers (with different group) and then shutdown their coordinator (broker). After first coordinator is shutdown, the heartbeat thread of first consumer gets timeout so it sends LEAVE_GROUP request to “next” coordinator. Unfortunately, the LEAVE_GROUP request can’t be processed successfully if the ”next” coordinator is the coordinator of second consumer (yep, it is shutdown also). 

The simple approach is that we can reduce the session timeout so coordinator can evict dead member in time.


chia7712:
I have looped the test 40 times. all pass

PR #9130, opened 2020-08-05
KAFKA-10492; Core Kafka Raft Implementation (KIP-595)

hachikuji:
This is the core Raft implementation specified by KIP-595: https://cwiki.apache.org/confluence/display/KAFKA/KIP-595%3A+A+Raft+Protocol+for+the+Metadata+Quorum. We have created a separate "raft" module where most of the logic resides. The new APIs introduced in this patch in order to support Raft election and such are disabled in the server until the integration with the controller is complete. Until then, there is a standalone server which can be used for testing the performance of the Raft implementation. See `raft/README.md` for details.

Co-authored-by: Boyang Chen <boyang@confluent.io>
Co-authored-by: Guozhang Wang <wangguoz@gmail.com> 


abbccdda (approved):
Double checked the last couple of commits and they make sense to me. Great work folks! @hachikuji @guozhangwang

guozhangwang:
Also browsed through the commit history and it matches my memory. Thanks for the awesome job @hachikuji

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -415,4 +444,10 @@ public void visit(Type field) {
         return hasBuffer.get();
     }
 
+    public static Set<ApiKeys> enabledApis() {
```
guozhangwang: nit: if this is only used for unit tests, better add a one-liner comment for that.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -182,10 +182,13 @@ class KafkaApis(val requestChannel: RequestChannel,
...
         case ApiKeys.DESCRIBE_USER_SCRAM_CREDENTIALS => handleDescribeUserScramCredentialsRequest(request)
         case ApiKeys.ALTER_USER_SCRAM_CREDENTIALS => handleAlterUserScramCredentialsRequest(request)
-        case ApiKeys.VOTE => handleVote(request)
-        case ApiKeys.BEGIN_QUORUM_EPOCH => handleBeginEpoch(request)
-        case ApiKeys.END_QUORUM_EPOCH => handleEndEpoch(request)
-        case ApiKeys.DESCRIBE_QUORUM => handleDescribeQuorum(request)
+
+        // Until we are ready to integrate the Raft layer, these APIs are treated as
+        // unexpected and we just close the connection.
+        case ApiKeys.VOTE => closeConnection(request, util.Collections.emptyMap())
```
guozhangwang, line 189: I feel this is not necessary since we would throw in `processCompletedReceives` at SocketServer already as `processChannelException`, but no harm guarding them as well.

hachikuji: Right. We should never reach here, but I thought we should make the `match` complete to avoid warnings.

hachikuji:
As long as there are no problems with the build, I am planning to merge this within the next couple days. For a history of the review comments, please see: https://github.com/confluentinc/kafka/commits/kafka-raft. 

Note that we have taken pains to ensure that the new APIs added for the Raft layer are not exposed so that we can continue iterating on them until they are ready to be integrated with the controller. There are also a few known performance issues which need to be sorted out. All of this will be tracked in https://issues.apache.org/jira/browse/KAFKA-9876 (in case anyone wants to help out).

PR #9132, opened 2020-08-06
[KAFKA-10422] [WIP] Introduce `timesForOffsets` in `KafkaConsumer`

guillaumebort:
The kafka consumer already provides an operation to quickly lookup the offsets by timestamp by using the `offsetsForTimes` operation.

However there are use cases where the inverse operation would be useful: having a set of offsets, I would like to lookup the ingestion timestamps for all these messages. Currently it would require fetching all these message by random access to retrieve the timestamps.

This add the `timesForOffsets` operation to the kafka consumer. The operation signature is equivalent to `offsetsForTimes`: given a mapping from partition to the offset to look up, it returns a mapping from partition to the timestamp and offset of the message at the requested offset. null is returned for the partition if there is no message at this offset.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/requests/ListOffsetRequest.java
+++ b/clients/src/main/java/org/apache/kafka/common/requests/ListOffsetRequest.java
@@ -43,6 +43,7 @@ import static org.apache.kafka.common.protocol.CommonFields.TOPIC_NAME;
 public class ListOffsetRequest extends AbstractRequest {
     public static final long EARLIEST_TIMESTAMP = -2L;
     public static final long LATEST_TIMESTAMP = -1L;
+    public static final long ANY_OFFSET = -1L;
 
     public static final int CONSUMER_REPLICA_ID = -1;
     public static final int DEBUGGING_REPLICA_ID = -2;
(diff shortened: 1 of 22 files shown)


clients/src/main/java/org/apache/kafka/common/requests/ListOffsetRequest.java
```diff
@@ -123,20 +126,36 @@
...
+            LIST_OFFSET_REQUEST_V3, LIST_OFFSET_REQUEST_V4, LIST_OFFSET_REQUEST_V5, LIST_OFFSET_REQUEST_V6};
     }
 
     private final int replicaId;
     private final IsolationLevel isolationLevel;
-    private final Map<TopicPartition, PartitionData> partitionTimestamps;
+    private final Map<TopicPartition, PartitionData> partitionsData;
     private final Set<TopicPartition> duplicatePartitions;
 
     public static class Builder extends AbstractRequest.Builder<ListOffsetRequest> {
```
Amanda1232, line 155: Our static analysis engine detects that your pull request includes a change that **class name shadows simple name of superclass**. According to [this description](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-shouldn-t-shadow-simple-name-of-superclass-nm-same-simple-name-as-superclass), it can be exceptionally confusing, creating many situations where you have to look at import statements to resolve references, and creating many opportunities to accidentally define methods that don't override the methods in their superclasses.

PR #9134, opened 2020-08-07
MINOR: set initial capacity for all array buffers in converting produ…

chia7712:
this is a bit optimization of converting producer responses. Those array buffers need initial capacity to avoid growing continually if the request carries a bunch of small messages and those messages are sent to different partitions.


chia7712:
test this patch with timed profiling recording (5 minutes) and the memory usage of array reallocation is reduced from 3.14 GB to 2.81 GB

PR #9136, opened 2020-08-07
KAFKA-10211: Add DirectoryConfigProvider

tombentley:
See KIP-632.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/config/provider/DirectoryConfigProvider.java
+++ b/clients/src/main/java/org/apache/kafka/common/config/provider/DirectoryConfigProvider.java
@@ -22,14 +22,16 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
-import java.io.FileFilter;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import static java.util.Collections.emptyMap;
 
 /**
  * An implementation of {@link ConfigProvider} based on a directory of files.
@@ -39,7 +41,7 @@ import java.util.Set;
  */
 public class DirectoryConfigProvider implements ConfigProvider {
 
-    private final Logger log = LoggerFactory.getLogger(getClass());
+    private static final Logger log = LoggerFactory.getLogger(DirectoryConfigProvider.class);
 
     @Override
     public void configure(Map<String, ?> configs) { }
@@ -55,7 +57,7 @@ public class DirectoryConfigProvider implements ConfigProvider {
      */
     @Override
     public ConfigData get(String path) {
-        return get(path, File::isFile);
+        return get(path, Files::isRegularFile);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tombentley:
@omkreddy, @mimaison, @rajinisivaram, @kkonstantine, @dajac thanks for voting on the KIP. I'd be grateful for review.

clients/src/main/java/org/apache/kafka/common/config/provider/DirectoryConfigProvider.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private ConfigData get(String path, FileFilter fileFilter) {
+        Map<String, String> map = new HashMap<>();
+        if (path != null && !path.isEmpty()) {
+            File dir = new File(path);
+            if (!dir.isDirectory()) {
+                log.warn("The path {} is not a directory", path);
+            } else {
+                for (File file : dir.listFiles(fileFilter)) {
```
mimaison: Should we use `java.nio.Files` here instead? That will also give us an Exception if an IO error happens instead of `null`.

mimaison: Actually it's worth checking if we can fully switch to `java.nio` classes.

clients/src/main/java/org/apache/kafka/common/config/provider/DirectoryConfigProvider.java
```diff
@@ -0,0 +1,98 @@
...
+                    } catch (IOException e) {
+                        throw new ConfigException("Could not read file " + file.getAbsolutePath() + " for property " + file.getName(), e);
+                    }
+                }
+            }
+        }
+        return new ConfigData(map);
+    }
+
+    private String read(Path path) throws IOException {
```
mimaison: This can be `static`

clients/src/main/java/org/apache/kafka/common/config/provider/DirectoryConfigProvider.java
```diff
@@ -0,0 +1,98 @@
...
+     * @return the configuration data.
+     */
+    @Override
+    public ConfigData get(String path, Set<String> keys) {
+        return get(path, pathname ->
+                pathname.isFile()
+                        && keys.contains(pathname.getName()));
+    }
+
+    private ConfigData get(String path, FileFilter fileFilter) {
```
mimaison: If we make `log` static, this can be `static` too

clients/src/test/java/org/apache/kafka/common/config/provider/DirectoryConfigProviderTest.java
```diff
@@ -0,0 +1,147 @@
...
+    private static File writeFile(File file) throws IOException {
+        Files.write(file.toPath(), file.getName().toUpperCase(Locale.ENGLISH).getBytes(StandardCharsets.UTF_8));
+        return file;
+    }
+
+    @Before
+    public void setup() throws IOException {
+        provider = new DirectoryConfigProvider();
+        provider.configure(Collections.emptyMap());
+        parent = TestUtils.tempDirectory();
```
mimaison, line 62: Should we delete anything that gets created in this directory?

mimaison (changes requested):
Thanks @tombentley. I've made a first pass and left a few comments

clients/src/main/java/org/apache/kafka/common/config/provider/DirectoryConfigProvider.java
tombentley: That's a good point about null being used for the error case as well as the directory case. Thanks.

tombentley:
@mimaison done.

clients/src/main/java/org/apache/kafka/common/config/provider/DirectoryConfigProvider.java
```diff
@@ -0,0 +1,108 @@
...
+                    map = Files.list(dir)
+                        .filter(fileFilter)
+                        .collect(Collectors.toMap(
+                            p -> p.getFileName().toString(),
+                            p -> {
+                                try {
+                                    return read(p);
+                                } catch (IOException e) {
+                                    throw new ConfigException("Could not read file " + p + " for property " + p.getFileName(), e);
+                                }
```
dajac: nit: Could we push the `try catch` block into the `read` method? That would streamline the code a little bit.

tombentley: Yes, good point. Done.

mimaison:
@tombentley The change looks good to me but there's a checkstyle issue:
```
[ant:checkstyle] [ERROR] /Users/mickael/github-ws/kafka/clients/src/test/java/org/apache/kafka/common/config/provider/DirectoryConfigProviderTest.java:103:85: ',' is not followed by whitespace. [WhitespaceAfter]
```

tombentley:
@mimaison fixed.

mimaison:
ok to test

mimaison (approved):

mimaison:
ok to test

mimaison:
retest this please

mimaison:
Test failures are not related, merging

PR #8976, opened 2020-07-02
KIP-617: Allow Kafka Streams State Stores to be iterated backwards

jeqo:
KIP reference implementation. <https://cwiki.apache.org/confluence/display/KAFKA/KIP-617%3A+Allow+Kafka+Streams+State+Stores+to+be+iterated+backwards>

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9137, opened 2020-08-07
KAFKA-9929: Support reverse iterator on KeyValueStore

jeqo:
Implements [KIP-617](https://cwiki.apache.org/confluence/display/KAFKA/KIP-617%3A+Allow+Kafka+Streams+State+Stores+to+be+iterated+backwards) on KeyValueStore.

Testing strategy: extend existing tests to validate reverse operations are supported.


jeqo:
cc @ableegoldman 

This PR is ready for review, covering related feedback from #8976

streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
```diff
@@ -38,35 +38,68 @@
...
     KeyValueIterator<K, V> range(K from, K to);
 
+    /**
+     * Get a reverse iterator over a given range of keys. This iterator must be closed after use.
+     * The returned iterator must be safe from {@link java.util.ConcurrentModificationException}s
+     * and must not return null values.
+     * Order is not guaranteed as bytes lexicographical ordering might not represent key order.
+     *
+     * @param from The first key that could be in the range, where iteration ends.
+     * @param to   The last key that could be in the range, where iteration starts from.
```
ableegoldman, line 67: Seems a bit tricky to say that _to_ is the variable where iteration starts _from_ 😉  But I can see it both ways, so being clear in the javadocs is good enough for me

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
```diff
@@ -110,7 +111,15 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
...
     public synchronized KeyValueIterator<Bytes, byte[]> range(final Bytes from, final Bytes to) {
+        return range(from, to, false);
+    }
+
+    @Override
+    public KeyValueIterator<Bytes, byte[]> reverseRange(final Bytes from, final Bytes to) {
+        return range(from, to, true);
+    }
 
+    KeyValueIterator<Bytes, byte[]> range(final Bytes from, final Bytes to, final boolean reverse) {
```
ableegoldman: Should be private

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
```diff
@@ -150,8 +166,9 @@ public void close() {
     private class InMemoryKeyValueIterator implements KeyValueIterator<Bytes, byte[]> {
         private final Iterator<Bytes> iter;
 
-        private InMemoryKeyValueIterator(final Set<Bytes> keySet) {
-            this.iter = new TreeSet<>(keySet).iterator();
+        private InMemoryKeyValueIterator(final Set<Bytes> keySet, final boolean reverse) {
+            if (reverse) this.iter = new TreeSet<>(keySet).descendingIterator();
+            else this.iter = new TreeSet<>(keySet).iterator();
```
ableegoldman: nit: Use braces & separate lines

streams/src/main/java/org/apache/kafka/streams/state/internals/NamedCache.java
```diff
@@ -281,15 +281,24 @@ public boolean isEmpty() {
...
     }
 
-    private Iterator<Bytes> keySetIterator(final Set<Bytes> keySet) {
-        return new TreeSet<>(keySet).iterator();
+    synchronized Iterator<Bytes> reverseKeyRange(final Bytes from, final Bytes to) {
+        return keySetIterator(cache.navigableKeySet().subSet(from, true, to, true), true);
+    }
+
+    private Iterator<Bytes> keySetIterator(final Set<Bytes> keySet, final boolean reverse) {
+        if (reverse) return new TreeSet<>(keySet).descendingIterator();
```
ableegoldman: nit: braces + separate lines

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBRangeIterator.java
```diff
@@ -29,32 +29,41 @@
...
+        this.reverse = reverse;
+        if (reverse) {
+            iter.seekForPrev(to.get());
+            rawLastKey = from.get();
+        } else {
+            iter.seek(from.get());
+            rawLastKey = to.get();
+        }
+        if (rawLastKey == null) {
             throw new NullPointerException("RocksDBRangeIterator: RawToKey is null for key " + to);
```
ableegoldman: nit: `RawToKey` --> `RawLastKey`

ableegoldman: Also it should be `from` for the reverse case, right?

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBRangeIterator.java
```diff
@@ -29,32 +29,41 @@
...
         final KeyValue<Bytes, byte[]> next = super.makeNext();
-
         if (next == null) {
             return allDone();
         } else {
-            if (comparator.compare(next.key.get(), rawToKey) <= 0) {
-                return next;
+            if (!reverse) {
+                if (comparator.compare(next.key.get(), rawLastKey) <= 0) return next;
+                else return allDone();
```
ableegoldman: nit: braces & separate lines

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
```diff
@@ -193,22 +193,26 @@ public void prepareBatch(final List<KeyValue<Bytes, byte[]>> entries,
...
+                to,
+                reverse);
         }
 
         @Override
-        public KeyValueIterator<Bytes, byte[]> all() {
+        public KeyValueIterator<Bytes, byte[]> all(final boolean reverse) {
             final RocksIterator innerIterWithTimestamp = db.newIterator(newColumnFamily);
-            innerIterWithTimestamp.seekToFirst();
+            if (reverse) innerIterWithTimestamp.seekToLast();
```
ableegoldman: braces & separate lines

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDbIterator.java
```diff
@@ -58,7 +61,8 @@ public synchronized boolean hasNext() {
             return allDone();
         } else {
             next = getKeyValue();
-            iter.next();
+            if (reverse) iter.prev();
```
ableegoldman: separate lines 🙂

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
```diff
@@ -306,26 +313,29 @@ public synchronized boolean hasNext() {
                 } else {
                     next = KeyValue.pair(new Bytes(nextWithTimestamp), iterWithTimestamp.value());
                     nextWithTimestamp = null;
-                    iterWithTimestamp.next();
+                    if (reverse) iterWithTimestamp.prev();
```
ableegoldman: braces & separate lines here and below

streams/src/test/java/org/apache/kafka/streams/state/internals/AbstractKeyValueStoreTest.java
```diff
@@ -422,6 +503,21 @@ public void shouldNotThrowInvalidRangeExceptionWithNegativeFromKey() {
                     " Note that the built-in numerical serdes do not follow this for negative numbers")
             );
         }
+    }
+
+    @Test
+    public void shouldNotThrowInvalidReverseRangeExceptionWithNegativeFromKey() {
```
ableegoldman, line 510: Can we add tests for some other invalid range cases? For example with both bounds positive but from > to

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingKeyValueStoreTest.java
```diff
@@ -294,13 +305,27 @@ public void shouldIterateOverRange() {
...
+    @Test
+    public void shouldReverseIterateOverRange() {
+        final int items = addItemsToCache();
+        final KeyValueIterator<Bytes, byte[]> range =
+            store.reverseRange(bytesKey(String.valueOf(0)), bytesKey(String.valueOf(items)));
+        final List<Bytes> results = new ArrayList<>();
+        while (range.hasNext()) {
+            results.add(range.next().key);
+        }
+        assertEquals(items, results.size());
```
ableegoldman, line 332: Can we add some tests that verify the actual contents + order of the reverse range?

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingKeyValueStoreTest.java
```diff
@@ -339,12 +366,24 @@ public void shouldThrowIfTryingToDoRangeQueryOnClosedCachingStore() {
         store.range(bytesKey("a"), bytesKey("b"));
     }
 
+    @Test(expected = InvalidStateStoreException.class)
```
ableegoldman: Use `assertThrows` -- we've been (slowly) migrating away from `@Test(expected)` in the Streams tests

streams/src/test/java/org/apache/kafka/streams/state/internals/CompositeReadOnlyKeyValueStoreTest.java
```diff
@@ -150,6 +161,17 @@ public void shouldThrowUnsupportedOperationExceptionWhileRemove() {
         } catch (final UnsupportedOperationException e) { }
     }
 
+    @Test
+    public void shouldThrowUnsupportedOperationExceptionWhileReverseRange() {
+        stubOneUnderlying.put("a", "1");
+        stubOneUnderlying.put("b", "1");
+        final KeyValueIterator<String, String> keyValueIterator = theStore.reverseRange("a", "b");
+        try {
+            keyValueIterator.remove();
```
ableegoldman: use `assertThrows` here as well

streams/src/test/java/org/apache/kafka/streams/state/internals/CompositeReadOnlyKeyValueStoreTest.java
```diff
@@ -199,11 +236,44 @@ public void shouldSupportRangeAcrossMultipleKVStores() {
...
+        cache.put("d", "d");
+        cache.put("x", "x");
+
+        final List<KeyValue<String, String>> results = toList(theStore.reverseRange("a", "e"));
         assertTrue(results.contains(new KeyValue<>("a", "a")));
         assertTrue(results.contains(new KeyValue<>("b", "b")));
         assertTrue(results.contains(new KeyValue<>("c", "c")));
         assertTrue(results.contains(new KeyValue<>("d", "d")));
         assertEquals(4, results.size());
+        //FIXME: order does not hold between stores, how to validate order here?
```
ableegoldman: I think the best we can do is just make sure that order is correct within a store. ie if `a`, `m` are all in `stubOneUnderling` then make sure the reverse range returns `m` before `a`.

I also think it would be fine to just make sure all the expected values are returned without checking the order, since there are other tests to verify that the order within a store is correct

ableegoldman (commented):
Thanks for splitting this up! Mostly minor comments on the KVStore PR, some general notes about the Streams style conventions:
1. always use braces for `if`/`else`/etc and put each condition on its own line
2. We're trying to move away from using `@Test(expected)` in the tests in favor of `assertThrows`. A lot of the tests haven't been migrated over yet, but just a heads up

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBRangeIterator.java
jeqo: right. I've moved this into each condition to set a correct exception message.

streams/src/test/java/org/apache/kafka/streams/state/internals/AbstractKeyValueStoreTest.java
jeqo: ack. added.

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingKeyValueStoreTest.java
jeqo: ack. added.

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingKeyValueStoreTest.java
jeqo: migrated the whole class. will apply this to the other PRs.

streams/src/test/java/org/apache/kafka/streams/state/internals/CompositeReadOnlyKeyValueStoreTest.java
jeqo: ack. make sense.

jeqo:
@ableegoldman this should be ready for another review :)

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
```diff
@@ -29,7 +29,7 @@
                           final RocksIterator newIterator,
                           final Set<KeyValueIterator<Bytes, byte[]>> openIterators,
                           final Bytes prefix) {
-        super(name, newIterator, openIterators);
+        super(name, newIterator, openIterators, false);
```
ableegoldman: This class is still unused so maybe we should just take it out. Someone's working on a prefix seek KIP at the moment so it'll be replaced soon anyway

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
```diff
@@ -374,7 +412,7 @@ public Bytes peekNextKey() {
             if (next == null) {
                 return allDone();
             } else {
-                if (comparator.compare(next.key.get(), upperBoundKey) <= 0) {
+                if (comparator.compare(next.key.get(), lastKey) <= 0) {
```
ableegoldman: We need to branch on `reverse` here too, right?

ableegoldman: Kind of unrelated, but WDYT about renaming `RocksDBDualCFIterator` to `RocksDBDualCFAllIterator` or something on the side? I feel like these iterators could be cleaned up a bit in general to be more understandable -- for example, it's weird that we do the `iterator#seek`-ing in the actual `all()` method but for range queries we do the seeking inside the iterator constructor. 
Just thinking out loud though, we can do some followup refactoring once this is merged

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
```diff
@@ -306,26 +319,41 @@ public synchronized boolean hasNext() {
...
                     next = KeyValue.pair(new Bytes(nextNoTimestamp), convertToTimestampedFormat(iterNoTimestamp.value()));
                     nextNoTimestamp = null;
-                    iterNoTimestamp.next();
+                    if (reverse) {
+                        iterNoTimestamp.prev();
+                    } else {
+                        iterNoTimestamp.next();
+                    }
                 } else {
                     if (comparator.compare(nextNoTimestamp, nextWithTimestamp) <= 0) {
```
ableegoldman: If `compare(noTimestamp, withTimestamp)` <= 0 then `withTimestamp >= noTimestamp`, so for the reverse case we would actually want to return `withTimestamp` next here (and vice versa in the else block)

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDbIterator.java
```diff
@@ -58,7 +61,11 @@ public synchronized boolean hasNext() {
             return allDone();
         } else {
             next = getKeyValue();
-            iter.next();
+            if (reverse) {
```
ableegoldman: Instead of storing and checking the `reverse` flag on every iteration, can we define something like
```
java.util.function.Consumer<RocksIterator> advanceIterator = reverse ? RocksIterator::prev : RocksIterator::next;
```
so then we can just blindly call `advanceIterator.accept(iter)`

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingKeyValueStoreTest.java
```diff
@@ -321,67 +368,94 @@ public void shouldClearNamespaceCacheOnClose() {
         assertEquals(0, cache.size());
     }
 
-    @Test(expected = InvalidStateStoreException.class)
```
ableegoldman, line 324: Awesome, thank you for cleaning up the whole class 🙏

streams/src/test/java/org/apache/kafka/streams/state/internals/AbstractKeyValueStoreTest.java
```diff
@@ -188,7 +188,55 @@ public void testPutGetRange() {
     }
 
     @Test
-    public void testPutGetRangeWithDefaultSerdes() {
+    public void testPutGetReverseRange() {
```
ableegoldman, line 191: I think we should add some tests to `RocksDBTimestampedStoreTest`. I thought it would extend `AbstractKeyValueStoreTest` and thus benefit from everything you added here, but doesn't seem to be the case :/ 
Personally I found the `RocksDBDualCFIterator` logic a bit difficult to follow even before the reverse iteration,  so it would be nice to have some tests specifically covering reverse iterators over multi-column-family timestamped stores

ableegoldman (commented):
Did a quick survey of everywhere that implements `range()` and `all()` to see if there might be anywhere else we need to add `reverseRange` and `reverseAll`. These guys seem to be missing their reverse counterparts: 
 - `KeyValueStoreReadWriteDecorator`
- `KeyValueToTimestampedKeyValueByteStoreAdapter`
- `Segment` (this one is not strictly necessary until the WindowStore PR though)

The fact that you've had to add these to about 25 different classes already really highlights the mess of the store hierarchy 😬

streams/src/test/java/org/apache/kafka/streams/state/internals/AbstractKeyValueStoreTest.java
jeqo: Just realized that, I also thought that path was tested. Good catch!

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
jeqo: Agree. I will continue the current approoach and create an issue to follow up this.

jeqo: https://issues.apache.org/jira/browse/KAFKA-10409

streams/src/test/java/org/apache/kafka/streams/state/internals/AbstractKeyValueStoreTest.java
jeqo: @ableegoldman I have extended `RocksDBTimestampedStoreTest` to use `reverseAll` and `reverseRange` as part of the current tests. 
Unfortunately, `AbstractKeyValueStoreTest` tests do not fit with the creation path of Timestamped stores as pre inserted data is required. 
Will add this to the same JIRA ticket to consider when refactoring iterators and tests.

ableegoldman: Ah, that makes sense I guess. Looks like the new additions to `RocksDBTimestampedStoreTest` cover the cross-column family code path so that's good enough for now

ableegoldman (commented):
Thanks for the PR! Things LGTM now, I'll ping a committer to give this another pass and get it merged.

Let me know when the WindowStore PR is ready and I'll start reviewing the next one 😄

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingKeyValueStore.java
```diff
@@ -249,7 +248,23 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
...
+        return new MergedSortedCacheKeyValueBytesStoreIterator(cacheIterator, storeIterator, false);
+    }
+
+    @Override
+    public KeyValueIterator<Bytes, byte[]> reverseRange(final Bytes from,
+                                                        final Bytes to) {
+        if (from.compareTo(to) > 0) {
+            LOG.warn("Returning empty iterator for fetch with invalid key range: from > to. "
+                + "This may be due to serdes that don't preserve ordering when lexicographically comparing the serialized bytes. " +
+                "Note that the built-in numerical serdes do not follow this for negative numbers");
```
vvcephei: This warning seems to miss the most likely scenario, that the user just passed the arguments in the wrong order.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
```diff
@@ -1,54 +0,0 @@
...
- */
-package org.apache.kafka.streams.state.internals;
-
-import org.apache.kafka.common.utils.Bytes;
-import org.apache.kafka.streams.state.KeyValueIterator;
-import org.rocksdb.RocksIterator;
-
-import java.util.Set;
-
-class RocksDBPrefixIterator extends RocksDbIterator {
```
vvcephei, line 25: Was this class unused or something?

vvcephei (commented):
Thanks for the PR, @jeqo ! I made a partial pass today. It looks really good overall.

A high-level comment: in several places in this code, we have a internal boolean flag to represent the direction. It seems odd that in these places, `true = reverse`. It seems like the "natural" iteration order is forward, so we should have blocks like `if (forward) /*forward logic*/ else /*reverse logic*/`. But what we have here is `if (not reverse) /*forward logic*/ else /*reverse logic*/`. It's not exactly hard to understand, but still it seems to be creating unnecessary hoops for people to jump through when they are reading the code.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
vvcephei: yes, it looks like it was. Thanks for cleaning it up!

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDbIterator.java
```diff
@@ -58,7 +62,7 @@ public synchronized boolean hasNext() {
             return allDone();
         } else {
             next = getKeyValue();
-            iter.next();
+            advanceIterator.accept(iter);
```
vvcephei, line 65: clever!

vvcephei (commented):
Hey @jeqo , I've completed my review. It looks good to me! Thanks so much for the KIP and PR.

Please let me know what you want to do about my prior comments, if anything :)

jeqo:
@vvcephei thanks for your feedback! Inverting flag actually make things more readable. Changing to `forward` and `if (forward) {} else {}`.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
ableegoldman: Yeah this was left over from the FKJ work IIRC

vvcephei (approved):
Thanks @jeqo !

vvcephei:
It looks like java 14 only failed while attempting to download some docs from Oracle, and java 8 failed on this unrelated test: `org.apache.kafka.clients.admin.KafkaAdminClientTest.testMetadataRetries`

PR #9139, opened 2020-08-07
KAFKA-9929: Support backward iterator on SessionStore

jeqo:
Depends on https://github.com/apache/kafka/pull/9138

comment to ping reviewers.*


jeqo:
cc @ableegoldman and team, this should be ready to review. 
Would also be great to look #9228 as part of KIP-666, to get the APIs aligned. 

Thanks!

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingSessionStore.java
```diff
@@ -359,7 +431,11 @@ private void getNextSegmentIterator() {
...
             current.close();
-            current = context.cache().range(cacheName, cacheKeyFrom, cacheKeyTo);
+            if (forward) {
+                current = context.cache().range(cacheName, cacheKeyFrom, cacheKeyTo);
+            } else {
+                current = context.cache().reverseRange(cacheName, cacheKeyFrom, cacheKeyTo);
+            }
         }
 
         private void setCacheKeyRange(final long lowerRangeEndTime, final long upperRangeEndTime) {
```
ableegoldman, line 480: Can you fix the `keyFrom == keyTo` to use `.equals` on the side (down on line 370)

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingSessionStore.java
```diff
@@ -359,7 +431,11 @@ private void getNextSegmentIterator() {
             setCacheKeyRange(currentSegmentBeginTime(), currentSegmentLastTime());
 
             current.close();
-            current = context.cache().range(cacheName, cacheKeyFrom, cacheKeyTo);
+            if (forward) {
```
ableegoldman: I think we're going to need some additional changes in this class similar to what we had in CachingWindowStore. Definitely at least in `getNextSegmentIterator()`. Let's make sure to have some cross-segment test coverage here as well, especially because the iteration logic of session store range queries is the hardest to wrap your head around out of all the stores (at least, it is for me)

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemorySessionStore.java
```diff
@@ -382,9 +478,20 @@ private boolean setInnerIterators() {
...
                 if (latestSessionStartTime == Long.MAX_VALUE) {
-                    recordIterator = nextKeyEntry.getValue().entrySet().iterator();
+                    final Set<Entry<Long, byte[]>> entries;
+                    if (forward) entries = nextKeyEntry.getValue().descendingMap().entrySet();
+                    else entries = nextKeyEntry.getValue().entrySet();
+                    recordIterator = entries.iterator();
                 } else {
-                    recordIterator = nextKeyEntry.getValue().headMap(latestSessionStartTime, true).entrySet().iterator();
+                    final Set<Entry<Long, byte[]>> entries;
+                    if (forward) entries = nextKeyEntry.getValue()
```
ableegoldman: If/else needs brackets

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemorySessionStore.java
```diff
@@ -201,7 +247,26 @@ public void remove(final Windowed<Bytes> sessionKey) {
 
         removeExpiredSegments();
 
-        return registerNewIterator(key, key, Long.MAX_VALUE, endTimeMap.entrySet().iterator());
+        return registerNewIterator(
+            key,
+            key,
+            Long.MAX_VALUE, endTimeMap.entrySet().iterator(),
```
ableegoldman: missing newline

ableegoldman (commented):
Hey @jeqo -- finally got around to giving this a first pass, thanks for being so patient 😄 

A few minor notes, mostly things we've covered in the KV/window store PRs already. I'm still going through the tests but things look good overall

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingSessionStore.java
```diff
@@ -270,25 +335,32 @@ public void close() {
...
+                                     final long latestSessionStartTime,
+                                     final boolean forward) {
             this.keyFrom = keyFrom;
             this.keyTo = keyTo;
             this.latestSessionStartTime = latestSessionStartTime;
             this.lastSegmentId = cacheFunction.segmentId(maxObservedTimestamp);
             this.segmentInterval = cacheFunction.getSegmentInterval();
+            this.forward = forward;
 
             this.currentSegmentId = cacheFunction.segmentId(earliestSessionEndTime);
```
ableegoldman: Ok I _think_ that for the reverse case, this should be initialized to `cacheFunction.segmentId(maxObservedTimestamp)` and `lastSegmentId` should be initialized to this (`segmentId(earliestSessionEndTime)`).

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingSessionStoreTest.java
```diff
@@ -301,6 +366,29 @@ public void shouldFetchRangeCorrectlyAcrossSegments() {
...
+        cachingStore.put(aa3, "3".getBytes());
+
+        final KeyValueIterator<Windowed<Bytes>, byte[]> rangeResults =
+            cachingStore.backwardFindSessions(keyA, keyAA, 0, SEGMENT_INTERVAL * 2);
+        final Set<Windowed<Bytes>> keys = new HashSet<>();
+        while (rangeResults.hasNext()) {
+            keys.add(rangeResults.next().key);
+        }
+        rangeResults.close();
+        assertEquals(mkSet(a1, a2, a3, aa1, aa3), keys);
```
ableegoldman: We're losing the ordering check by comparing this as a set, let's use a list (or whatever) to verify the actual order

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingSessionStoreTest.java
```diff
@@ -278,6 +326,23 @@ public void shouldFetchCorrectlyAcrossSegments() {
...
 
+    @Test
+    public void shouldBackwardFetchCorrectlyAcrossSegments() {
+        final Windowed<Bytes> a1 = new Windowed<>(keyA, new SessionWindow(SEGMENT_INTERVAL * 0, SEGMENT_INTERVAL * 0));
+        final Windowed<Bytes> a2 = new Windowed<>(keyA, new SessionWindow(SEGMENT_INTERVAL * 1, SEGMENT_INTERVAL * 1));
+        final Windowed<Bytes> a3 = new Windowed<>(keyA, new SessionWindow(SEGMENT_INTERVAL * 2, SEGMENT_INTERVAL * 2));
+        cachingStore.put(a1, "1".getBytes());
+        cachingStore.put(a2, "2".getBytes());
+        cachingStore.flush();
+        cachingStore.put(a3, "3".getBytes());
```
ableegoldman, line 372: Can we add a few more records that span multiple segments that don't get flushed as well?

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingSessionStoreTest.java
```diff
@@ -456,68 +562,88 @@ public void shouldClearNamespaceCacheOnClose() {
...
 
-    @Test(expected = NullPointerException.class)
+    @Test
     public void shouldThrowNullPointerExceptionOnPutNullKey() {
-        cachingStore.put(null, "1".getBytes());
+        assertThrows(NullPointerException.class, () -> cachingStore.put(null, "1".getBytes()));
+    }
+
+    @Test
+    public void shouldNotThrowInvalidBackwardRangeExceptionWithNegativeFromKey() {
```
ableegoldman: Technically "InvalidRangeException" was just the name of the exception that could get thrown, there is no "InvalidBackwardRangeException" that I know of 😛  But I think the meaning is clear enough lol

streams/src/test/java/org/apache/kafka/test/ReadOnlySessionStoreStub.java
```diff
@@ -82,11 +118,15 @@ public boolean hasNext() {
...
                     return it.next();
                 }
-
             }
         );
     }
 
+    @Override
+    public KeyValueIterator<Windowed<K>, V> backwardFetch(K from, K to) {
+        return null;
```
ableegoldman: I guess it probably doesn't matter since we presumably aren't using these backward methods of the ReadOnlySessionStoreStub, but it seems like it might result in some tricky NPEs to debug if ever someone does try to use it in a test. If you don't feel like implementing it I think it's fine to just throw UnsupportedOperationException and say that you'll have to implement this to use it.

Or just copy the code from the forward direction and flip it 🤷‍♀️  Same goes for all the methods in here that return null

ableegoldman (commented):
Just a few more minor things, should definitely be able to merge this by feature freeze 👍

streams/src/test/java/org/apache/kafka/test/ReadOnlySessionStoreStub.java
jeqo: mixed choice here. hope is good enough

streams/src/main/java/org/apache/kafka/streams/state/ReadOnlySessionStore.java
```diff
@@ -24,35 +24,156 @@
...
+     * @param key                    the key to return sessions for
+     * @param earliestSessionEndTime the end timestamp of the earliest session to search for, where iteration starts.
+     * @param latestSessionStartTime the end timestamp of the latest session to search for, where iteration ends.
+     * @return iterator of sessions with the matching key and aggregated values, from earliest to latest session time.
+     * @throws NullPointerException If null is used for key.
+     */
+    default KeyValueIterator<Windowed<K>, AGG> findSessions(final K key,
+                                                            final long earliestSessionEndTime,
+                                                            final long latestSessionStartTime) {
+        throw new UnsupportedOperationException("Moved from SessionStore");
```
vvcephei: It won't matter to users whether this method was moved from another interface or not. They just need to know why they're getting the exception. I.e., we just need to tell them that the store implementation they selected didn't implement the method.

```suggestion
        throw new UnsupportedOperationException("This API is not supported by this implementation of ReadOnlySessionStore.");
```

We should say the exact same thing in all default implementations. Right now, they're a bit inconsistent.

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingSessionStore.java
```diff
@@ -150,22 +149,44 @@ public void remove(final Windowed<Bytes> sessionKey) {
         validateStoreOpen();
 
         final PeekingKeyValueIterator<Bytes, LRUCacheEntry> cacheIterator = wrapped().persistent() ?
-            new CacheIteratorWrapper(key, earliestSessionEndTime, latestSessionStartTime) :
+            new CacheIteratorWrapper(key, earliestSessionEndTime, latestSessionStartTime, true) :
             context.cache().range(cacheName,
-                        cacheFunction.cacheKey(keySchema.lowerRangeFixedSize(key, earliestSessionEndTime)),
-                        cacheFunction.cacheKey(keySchema.upperRangeFixedSize(key, latestSessionStartTime))
+                cacheFunction.cacheKey(keySchema.lowerRangeFixedSize(key, earliestSessionEndTime)),
+                cacheFunction.cacheKey(keySchema.upperRangeFixedSize(key, latestSessionStartTime))
```
vvcephei: There are a lot of unnecessary whitespace changes in this PR. You don't need to back them all out right now, but in the future, please clean up the diff before submitting a PR. These extra changes make it harder to review.

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingSessionStore.java
jeqo: right! great catch! forgot to align this with WindowStore.

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingSessionStore.java
jeqo: One of the reasons these missing pieces are not throwing exceptions seems to be that Caching Store tests are only covering or InMemory or Persistent underlying stores. This means only one path of the caching logic is followed (e.g. `persistent() ? iteratorWraper : otherIterator`)

I rename the test classes to be explicit about what backing store is covered. We could probably cover this as another issue?

ableegoldman: Ohh yeah good catch. Kind of weird that we're inconsistent with which underlying store type is used in the existing caching store tests, too. But thanks for adding the in-memory + persistent flavors for the SessionStore, I think it sounds reasonable to file a ticket to follow up on the Window/KV flavors and clean up the caching tests

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemorySessionStore.java
```diff
@@ -382,9 +479,23 @@ private boolean setInnerIterators() {
                 currentKey = nextKeyEntry.getKey();
 
                 if (latestSessionStartTime == Long.MAX_VALUE) {
-                    recordIterator = nextKeyEntry.getValue().entrySet().iterator();
+                    final Set<Entry<Long, byte[]>> entries;
+                    if (forward) entries = nextKeyEntry.getValue().descendingMap().entrySet();
+                    else entries = nextKeyEntry.getValue().entrySet();
```
vvcephei: The code style discourages inline conditionals. It's more maintainable to always use blocks.

ableegoldman:
Retest this please

ableegoldman (approved):
LGTM! Thanks for the reverse iterator KIP 😄

ableegoldman:
test this please

ableegoldman:
Alright Jenkins is being wonky yet again, but I checked out this branch and ran the tests locally to verify.

vvcephei (approved):
Thanks, @jeqo !

I had a few complaints about formatting and one about the exception message. Since we're down to the limit, I'll go ahead and add a commit addressing these concerns, test it for myself (thanks for your efforts there, @ableegoldman ), and merge it.

Thanks so much for this PR.

vvcephei:
Ok, I've pushed a couple of commits addressing my feedback.

I absolutely share the desire to clean up bad formatting in the codebase, but when PRs are this extensive, I'd suggest eliminating absolutely all changes that aren't directly related to the change. The extra whitespace changes, etc., just add noise that makes it harder for reviewers to do their job. Plus, it increases the probability of merge conflicts.

I'd encourage sending a separate PR applying formatting changes or other style fixes. This is what I do myself.

Anyway, I rolled back the "extra" stuff. I also applied a couple of extra formatting changes to make the new code itself comply with the style guidelines.

And it looks like Jenkins finally woke up, so I'll let it go ahead and finish running.

vvcephei:
Oh, ironically, there is a merge conflict with trunk, and it's due to one of the only "extra" changes I left in: renaming Caching*StoreTest to Caching*StoreTest and in the Session variant, making `context` a local variable.

I didn't un-rename that class, but I did restore `context` as a field. I copied over the newly-added test in trunk just to the InMemory variants of the tests. They only mock the inner store and test the init method anyway.

vvcephei:
Jenkins is running again, so I'll let it finish and merge in the morning.

vvcephei:
Ok, it's merged. I'm waiting on permission from Bill before cherry-picking to 2.7

vvcephei:
cherry-picked to 2.7. Thanks @jeqo !

PR #8672, opened 2020-05-15
KAFKA-10002; Improve performances of StopReplicaRequest with large number of partitions to be deleted

dajac:
* Update checkpoint files once for all deleted partitions instead of updating them for each deleted partitions. With this, a stop replica requests with 2000 partitions to be deleted takes ~2 secs instead of ~40 secs previously.
* Refactor the checkpointing methods to not compute the `logsByDir` all the time. It is now reused as much as possible.
* Refactor the exception handling. Some checkpointing methods were handling `IOException` but the underlying write process already catches them and throws KafkaStorageException instead.
* Reduce the logging in the log cleaner manager. It does not log anymore when a partition is deleted as it is not a useful information.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9140, opened 2020-08-07
KAFKA-10371; Partition reassignments can result in crashed ReplicaFetcherThreads

dajac:
https://github.com/apache/kafka/pull/8672 introduced a bug leading to crashing the replica fetcher threads. The issue is that https://github.com/apache/kafka/pull/8672 deletes the Partitions prior to stopping the replica fetchers. As the replica fetchers relies access the Partition in the ReplicaManager, they crash with a NotLeaderOrFollowerException that is not handled.

This PR reverts the code to the original ordering to avoid this issue.

The regression has been caught by our system test: `kafkatest.tests.core.reassign_partitions_test`.

I have not managed to reproduce the issue in a unit test without reimplementing the entire system test in Java. I am not sure that makes sense as we already have it in Python.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -366,8 +366,7 @@ class ReplicaManager(val config: KafkaConfig,
       } else {
         this.controllerEpoch = controllerEpoch
 
-        val stoppedPartitions = mutable.Set.empty[TopicPartition]
-        val deletedPartitions = mutable.Set.empty[TopicPartition]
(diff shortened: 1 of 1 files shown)


hachikuji (approved):
LGTM. Let's create a JIRA to add concurrency tests so we can try to avoid this kind of problem in the future.

hachikuji:
ok to test

dajac:
Jira: https://issues.apache.org/jira/browse/KAFKA-10374

soondenana (approved):
LGTM

Thanks for fixing this.

soondenana:
I can confirm that this fixes the issue. I ran system test 18 times and didn't hit this issue. Without this fix, when I run the system test 10 times I hit the issue 2 times.

PR #9141, opened 2020-08-07
MINOR: Improve checks for CogroupedStreamAggregateBuilder

lct45:
Updated `CogroupedStreamAggregateBuilder` to have individual builders depending on the windowed aggregation, or lack thereof. This replaced passing in all options into the builder, with all but the current type of aggregation set to null and then checking to see which value was not null.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
@@ -47,20 +47,19 @@ class CogroupedStreamAggregateBuilder<K, VOut> {
     CogroupedStreamAggregateBuilder(final InternalStreamsBuilder builder) {
         this.builder = builder;
     }
-    <KR, VIn, W extends Window> KTable<KR, VOut> buildNotWindowed(final Map<KGroupedStreamImpl<K, ?>, Aggregator<? super K, ? super Object, VOut>> groupPatterns,
-                                                       final Initializer<VOut> initializer,
-                                                       final NamedInternal named,
-                                                       final StoreBuilder<?> storeBuilder,
-                                                       final Serde<KR> keySerde,
-                                                       final Serde<VOut> valueSerde,
-                                                       final String queryableName) {
-        build(groupPatterns, storeBuilder);
+    <KR> KTable<KR, VOut> processRepartitions(final Map<KGroupedStreamImpl<K, ?>, Aggregator<? super K, ? super Object, VOut>> groupPatterns,
+                                              final Initializer<VOut> initializer,
+                                              final NamedInternal named,
+                                              final StoreBuilder<?> storeBuilder,
+                                              final Serde<KR> keySerde,
+                                              final Serde<VOut> valueSerde,
+                                              final String queryableName) {
+        processRepartitions(groupPatterns, storeBuilder);
         final Collection<StreamsGraphNode> processors = new ArrayList<>();
         boolean stateCreated = false;
         int counter = 0;
         for (final Entry<KGroupedStreamImpl<K, ?>, Aggregator<? super K, Object, VOut>> kGroupedStream : groupPatterns.entrySet()) {
             final StatefulProcessorNode<K, ?> statefulProcessorNode = getStatefulProcessorNode(
-                    initializer,
                     named.suffixWithOrElseGet(
                             "-cogroup-agg-" + counter++,
                             builder,
@@ -75,22 +74,21 @@ class CogroupedStreamAggregateBuilder<K, VOut> {
         return createTable(processors, named, keySerde, valueSerde, queryableName);
     }
 
-    <KR, VIn, W extends Window> KTable<KR, VOut> buildTimeWindows(final Map<KGroupedStreamImpl<K, ?>, Aggregator<? super K, ? super Object, VOut>> groupPatterns,
-                                                              final Initializer<VOut> initializer,
-                                                              final NamedInternal named,
-                                                              final StoreBuilder<?> storeBuilder,
-                                                              final Serde<KR> keySerde,
-                                                              final Serde<VOut> valueSerde,
-                                                              final String queryableName,
-                                                                  final Windows<W> windows) {
-        build(groupPatterns, storeBuilder);
+    <KR, W extends Window> KTable<KR, VOut> processRepartitions(final Map<KGroupedStreamImpl<K, ?>, Aggregator<? super K, ? super Object, VOut>> groupPatterns,
+                                                                final Initializer<VOut> initializer,
+                                                                final NamedInternal named,
+                                                                final StoreBuilder<?> storeBuilder,
+                                                                final Serde<KR> keySerde,
+                                                                final Serde<VOut> valueSerde,
+                                                                final String queryableName,
+                                                                final Windows<W> windows) {
+        processRepartitions(groupPatterns, storeBuilder);
 
         final Collection<StreamsGraphNode> processors = new ArrayList<>();
         boolean stateCreated = false;
         int counter = 0;
         for (final Entry<KGroupedStreamImpl<K, ?>, Aggregator<? super K, Object, VOut>> kGroupedStream : groupPatterns.entrySet()) {
             final StatefulProcessorNode<K, ?> statefulProcessorNode = getStatefulProcessorNode(
-                    initializer,
                     named.suffixWithOrElseGet(
                             "-cogroup-agg-" + counter++,
                             builder,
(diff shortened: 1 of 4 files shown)


streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -47,18 +47,96 @@
     CogroupedStreamAggregateBuilder(final InternalStreamsBuilder builder) {
         this.builder = builder;
     }
-
-    <KR, VIn, W extends Window> KTable<KR, VOut> build(final Map<KGroupedStreamImpl<K, ?>, Aggregator<? super K, ? super Object, VOut>> groupPatterns,
+    <KR, VIn, W extends Window> KTable<KR, VOut> buildNotWindowed(final Map<KGroupedStreamImpl<K, ?>, Aggregator<? super K, ? super Object, VOut>> groupPatterns,
```
ableegoldman: WDYT about naming this just `build`? It's not as clear, but I think it's in line with the naming conventions elsewhere. For example we have `KStreamWindowAggregate` and `KStreamAggregate` (then maybe we can come up with a more descriptive name for the method currently called `build`

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -85,68 +163,39 @@
...
+                Collections.singleton(mergeNode.nodeName()),
+                queryableName,
+                passThrough,
+                mergeNode,
+                builder);
     }
 
-    private <W extends Window> StatefulProcessorNode<K, ?> getStatefulProcessorNode(final Aggregator<? super K, Object, VOut> aggregator,
-                                                                                    final Initializer<VOut> initializer,
+    private <W extends Window> StatefulProcessorNode<K, ?> getStatefulProcessorNode(final Initializer<VOut> initializer,
```
ableegoldman: We can remove the `<W extends Window>` now, right? Also it looks like the `initializer` input isn't needed anymore either

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -47,18 +47,96 @@
...
+                    new KStreamSessionWindowAggregate<>(sessionWindows, storeBuilder.name(), initializer, kGroupedStream.getValue(), sessionMerger));
+            stateCreated = true;
+            processors.add(statefulProcessorNode);
+            builder.addGraphNode(parentNodes.get(kGroupedStream.getKey()), statefulProcessorNode);
+        }
+        return createTable(processors, named, keySerde, valueSerde, queryableName);
+    }
+
+    private void build(final Map<KGroupedStreamImpl<K, ?>, Aggregator<? super K, ? super Object, VOut>> groupPatterns,
+                                                       final StoreBuilder<?> storeBuilder) {
```
ableegoldman: Can we call this something like `ensureCopartitioning` or `processRepartitions` or something? My take is that the copartitioning is the main point of this method so that's probably good to include in the name

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -85,68 +163,39 @@
...
-                windows,
-                sessionWindows,
-                sessionMerger);
-            stateCreated = true;
-            processors.add(statefulProcessorNode);
-            builder.addGraphNode(parentNodes.get(kGroupedStream.getKey()), statefulProcessorNode);
-        }
+    }
+
+    <KR, VIn, W extends Window> KTable<KR, VOut> createTable(final Collection<StreamsGraphNode> processors,
```
ableegoldman: Can remove the `W extends Window` generic

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -47,18 +47,96 @@
...
+                    storeBuilder,
+                    new KStreamAggregate<>(storeBuilder.name(), initializer, kGroupedStream.getValue()));
+            stateCreated = true;
+            processors.add(statefulProcessorNode);
+            builder.addGraphNode(parentNodes.get(kGroupedStream.getKey()), statefulProcessorNode);
+        }
+        return createTable(processors, named, keySerde, valueSerde, queryableName);
+    }
+
+    <KR, VIn, W extends Window> KTable<KR, VOut> buildTimeWindows(final Map<KGroupedStreamImpl<K, ?>, Aggregator<? super K, ? super Object, VOut>> groupPatterns,
```
ableegoldman: Don't need `Vin` here

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -47,18 +47,96 @@
...
+                    storeBuilder,
+                    new KStreamWindowAggregate<>(windows, storeBuilder.name(), initializer, kGroupedStream.getValue()));
+            stateCreated = true;
+            processors.add(statefulProcessorNode);
+            builder.addGraphNode(parentNodes.get(kGroupedStream.getKey()), statefulProcessorNode);
+        }
+        return createTable(processors, named, keySerde, valueSerde, queryableName);
+    }
+
+    <KR, VIn, W extends Window> KTable<KR, VOut> buildSessionWindows(final Map<KGroupedStreamImpl<K, ?>, Aggregator<? super K, ? super Object, VOut>> groupPatterns,
```
ableegoldman: Don't need `Vin` and `W extends Window` here

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
ableegoldman: Also you can remove the `Vin` and `W extends Window` generics on this method

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
ableegoldman: Honestly I think it's fine to just name all three of these `build`, since they accept different parameters and it should be pretty clear from the context whether it's windowed or not. But being more descriptive is never a bad thing either. Your call 🙂

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -47,18 +47,96 @@
...
+                            builder,
+                            CogroupedKStreamImpl.AGGREGATE_NAME),
+                    stateCreated,
+                    storeBuilder,
+                    new KStreamAggregate<>(storeBuilder.name(), initializer, kGroupedStream.getValue()));
+            stateCreated = true;
+            processors.add(statefulProcessorNode);
+            builder.addGraphNode(parentNodes.get(kGroupedStream.getKey()), statefulProcessorNode);
+        }
+        return createTable(processors, named, keySerde, valueSerde, queryableName);
```
ableegoldman: This is all the same for all three methods except for the `KStreamAggregate`/`KStreamWindowAggregate`/etc right? I think if you wanted to further deduplicate things you could factor this  out into a method that accepts a `Function< KGroupedStreamImpl, KStreamAggProcessorSupplier`>, and then each of the `build` methods can just pass in a function that returns `new KStreamWindowAggregate` or so on.
I'm not sure it's really worth it or not, but  it can be done in case you were wondering. Up to you whether you want to do it

ableegoldman (commented):
Nice! Just some minor cleaning up left over after you removed a bunch of stuff. (I just opened it in IDEA and looked for anything that was greyed out). Give it a pass to make sure the method parameters are all aligned properly after you remove the generics -- a tip from experience, I always forget to fix that after making changes.

Also, github won't let me comment on this line because it's too far from the PR, but can you fix the alignment of parameters down in [#createRepartitionSource](https://github.com/apache/kafka/pull/9141/files#diff-4286012c89d23871a643d157a651eb7bR220-R223)  as well?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
lct45: I went with processRepartitions and then updated all the other methods to just be `build` with different parameters since that seems cleaner

mjsax:
Retest this please.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -85,68 +160,38 @@
...
+                                           final Serde<KR> keySerde,
+                                           final Serde<VOut> valueSerde,
+                                           final String queryableName) {
         final String mergeProcessorName = named.suffixWithOrElseGet(
-            "-cogroup-merge",
-            builder,
-            CogroupedKStreamImpl.MERGE_NAME);
+                "-cogroup-merge",
+                builder,
+                CogroupedKStreamImpl.MERGE_NAME);
```
vvcephei: It seems like your indentation is set to 8 spaces instead of 4.

vvcephei (commented):
Hey @lct45, thanks for this improvement!

It LGTM, but can you fix the indentation before I merge this?

vvcephei:
All the tests were green except for the flaky
```
   org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[false]
    org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]
```

vvcephei:
Verified the last commit only changed whitespace before merging.

PR #9142, opened 2020-08-07
MINOR: Fix delete_topic for system tests

skaundinya15:
The current implementation of `delete_topic` in `kafka.py` doesn't take into account security credentials and thus fails when running tests with security enabled and trying to delete a topic. This PR is to fix this issue.

### DIFF
--- a/tests/kafkatest/services/kafka/kafka.py
+++ b/tests/kafkatest/services/kafka/kafka.py
@@ -503,7 +503,7 @@ class KafkaService(KafkaPathResolverMixin, JmxMixin, Service):
         self.logger.info("Running topic creation command...\n%s" % cmd)
         node.account.ssh(cmd)
 
-    def delete_topic(self, topic, node=None):
+    def delete_topic(self, topic, node=None, use_zk_to_delete_topic=False):
         """
         Delete a topic with the topics command
         :param topic:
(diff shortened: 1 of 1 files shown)


tests/kafkatest/services/kafka/kafka.py
```diff
@@ -503,7 +503,7 @@ def create_topic(self, topic_cfg, node=None, use_zk_to_create_topic=True):
         self.logger.info("Running topic creation command...\n%s" % cmd)
         node.account.ssh(cmd)
 
-    def delete_topic(self, topic, node=None):
+    def delete_topic(self, topic, node=None, use_zk_to_delete_topic=False):
```
rondagostino, line 506: create/describe/list use `use_zk_to_..._topic=True` so I'm wondering if it is best to remain consistent here.  For example, if running an older version, might `--bootstrap-server` not be available?

rondagostino (commented):
Left one comment, otherwise LGTM!

tests/kafkatest/services/kafka/kafka.py
skaundinya15: Yeah that's a good point, but in this scenario I was thinking to remain consistent with the original implementation. Also I'm not sure how far we want to backport this, but if we backport it past the version where `--bootstrap-server` is not available, then I think we would have to rewrite this function to account for that. Also since we are on our way to deprecate and remove ZK, this would be a good way to start pushing us towards doing that.

rajinisivaram (approved):
@skaundinya15 Thanks for the PR, LGTM

skaundinya15:
Link to system test run from this branch: http://confluent-kafka-branch-builder-system-test-results.s3-us-west-2.amazonaws.com/2020-08-07--001.1596866417--skaundinya15--minor-fix-delete-topic-implementation--387bd9ed9/report.html. The system test that uses `delete_topic` is 
```
Module: kafkatest.tests.core.replica_scale_test
Class:  ReplicaScaleTest
Method: test_clean_bounce
Arguments:
{
  "partition_count": 34,
  "replication_factor": 3,
  "topic_count": 500
}
```
And from the link above this test passed.

tests/kafkatest/services/kafka/kafka.py
rajinisivaram: @rondagostino are we ok with merging this to trunk? Since this is not required for existing tests which either use ZK or PLAINTEXT brokers, not planning to backport to older versions.

rondagostino: @rajinisivaram Yes, merging to just trunk seems fine to me.

rajinisivaram:
@skaundinya15 Thanks for the PR, merging to trunk

PR #9144, opened 2020-08-07
KAFKA-9705: (part-1) add redirection fields in the request header

abbccdda:
Add the redirection supporting fields, including:

1. initial principal name
2. initial client id
3. the flag to indicate whether a given request is coming from the control plane in a secured environment.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/ClientRequest.java
+++ b/clients/src/main/java/org/apache/kafka/clients/ClientRequest.java
@@ -44,8 +44,8 @@ public final class ClientRequest {
      * @param clientId The client ID to use for the header
      * @param createdTimeMs The unix timestamp in milliseconds for the time at which this request was created.
      * @param expectResponse Should we expect a response message or is this request complete once it is sent?
-     * @param initialPrincipalName The initial principal name if this is a redirect request
-     * @param initialClientId The initial client id if this is a redirect request
+     * @param initialPrincipalName The initial principal name if this is a redirect request, or null if this was not redirected
+     * @param initialClientId The initial client id if this is a redirect request, or null if this was not redirected
      * @param callback A callback to execute when the response has been received (or null if no callback is necessary)
      */
     public ClientRequest(String destination,
--- a/clients/src/main/java/org/apache/kafka/clients/ClientRequest.java
+++ b/clients/src/main/java/org/apache/kafka/clients/ClientRequest.java
@@ -34,6 +34,8 @@ public final class ClientRequest {
     private final boolean expectResponse;
     private final int requestTimeoutMs;
     private final RequestCompletionHandler callback;
+    private final String initialPrincipalName;
+    private final String initialClientId;
 
     /**
      * @param destination The brokerId to send the request to
@@ -42,6 +44,8 @@ public final class ClientRequest {
      * @param clientId The client ID to use for the header
      * @param createdTimeMs The unix timestamp in milliseconds for the time at which this request was created.
      * @param expectResponse Should we expect a response message or is this request complete once it is sent?
+     * @param initialPrincipalName The initial principal name if this is a redirect request
+     * @param initialClientId The initial client id if this is a redirect request
      * @param callback A callback to execute when the response has been received (or null if no callback is necessary)
      */
     public ClientRequest(String destination,
@@ -51,6 +55,8 @@ public final class ClientRequest {
                          long createdTimeMs,
                          boolean expectResponse,
                          int requestTimeoutMs,
+                         String initialPrincipalName,
+                         String initialClientId,
                          RequestCompletionHandler callback) {
         this.destination = destination;
         this.requestBuilder = requestBuilder;
@@ -60,6 +66,8 @@ public final class ClientRequest {
         this.expectResponse = expectResponse;
         this.requestTimeoutMs = requestTimeoutMs;
         this.callback = callback;
+        this.initialPrincipalName = initialPrincipalName;
+        this.initialClientId = initialClientId;
     }
 
     @Override
@@ -85,11 +93,13 @@ public final class ClientRequest {
     public RequestHeader makeHeader(short version) {
         short requestApiKey = requestBuilder.apiKey().id;
         return new RequestHeader(
-            new RequestHeaderData().
-                setRequestApiKey(requestApiKey).
-                setRequestApiVersion(version).
-                setClientId(clientId).
-                setCorrelationId(correlationId),
+            new RequestHeaderData()
+                .setRequestApiKey(requestApiKey)
+                .setRequestApiVersion(version)
+                .setClientId(clientId)
+                .setCorrelationId(correlationId)
+                .setInitialPrincipalName(initialPrincipalName)
+                .setInitialClientId(initialClientId),
             ApiKeys.forId(requestApiKey).requestHeaderVersion(version));
     }
 
--- a/clients/src/main/java/org/apache/kafka/clients/KafkaClient.java
+++ b/clients/src/main/java/org/apache/kafka/clients/KafkaClient.java
@@ -189,15 +189,17 @@ public interface KafkaClient extends Closeable {
      *                         cancelling the request. The request may get cancelled sooner if the socket disconnects
      *                         for any reason including if another pending request to the same node timed out first.
      * @param callback the callback to invoke when we get a response
+     * @param initialPrincipalName the initial client principal name, when building a forward request
+     * @param initialClientId the initial client id, when building a forward request
      */
     ClientRequest newClientRequest(String nodeId,
                                    AbstractRequest.Builder<?> requestBuilder,
                                    long createdTimeMs,
                                    boolean expectResponse,
                                    int requestTimeoutMs,
+                                   String initialPrincipalName,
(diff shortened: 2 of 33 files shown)


clients/src/main/java/org/apache/kafka/common/requests/RequestHeader.java
```diff
@@ -37,11 +37,22 @@ public RequestHeader(Struct struct, short headerVersion) {
...
     public RequestHeader(ApiKeys requestApiKey, short requestVersion, String clientId, int correlationId) {
-        this(new RequestHeaderData().
-                setRequestApiKey(requestApiKey.id).
-                setRequestApiVersion(requestVersion).
-                setClientId(clientId).
-                setCorrelationId(correlationId),
+        this(requestApiKey, requestVersion, clientId, correlationId, null, null);
+    }
+
+    public RequestHeader(ApiKeys requestApiKey,
```
cmccabe, line 43: I think it would be good to have a constructor that didn't have the two forwarding fields, to avoid changing a large number of tests.

clients/src/main/java/org/apache/kafka/server/authorizer/AuthorizableRequestContext.java
```diff
@@ -69,4 +69,11 @@
      * Returns the correlation id from the request header.
      */
     int correlationId();
+
+    /**
+     * Returns the initial principal name for a forwarded request.
+     */
+    default String initialPrincipalName() {
```
cmccabe, line 78: Does this need to be here?  I'm concerned that having it here will eventually lead to us using it for authorization, which it shouldn't be.

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -309,7 +309,10 @@ object RequestChannel extends Logging {
   }
 }
 
-class RequestChannel(val queueSize: Int, val metricNamePrefix : String, time: Time) extends KafkaMetricsGroup {
+class RequestChannel(val queueSize: Int,
+                     val metricNamePrefix : String,
+                     time: Time,
+                     val maybeFromControlPlane: Boolean) extends KafkaMetricsGroup {
```
cmccabe: Instead of `maybeFromControlPlane`, how about `fromControlPlaneListener`

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -88,15 +88,17 @@ class SocketServer(val config: KafkaConfig,
...
   private var controlPlaneProcessorOpt : Option[Processor] = None
   private[network] var controlPlaneAcceptorOpt : Option[Acceptor] = None
   val controlPlaneRequestChannelOpt: Option[RequestChannel] = config.controlPlaneListenerName.map(_ =>
-    new RequestChannel(20, ControlPlaneMetricPrefix, time))
+    new RequestChannel(20, ControlPlaneMetricPrefix, time, true))
+  // data-plane
+  private val dataPlaneProcessors = new ConcurrentHashMap[Int, Processor]()
+  private[network] val dataPlaneAcceptors = new ConcurrentHashMap[EndPoint, Acceptor]()
+  // If the control plane processor is not defined, just set the flag to true in data plane to bypass the check for whether a given
+  // request is from the control plane or not.
```
cmccabe: I don't think this is quite right.  The flag should be set based on whether we're in the control plane listener OR the inter-broker listener.  It's much more common to use a separate inter-broker listener, than to use a fully separate control plane listener.

clients/src/main/java/org/apache/kafka/server/authorizer/AuthorizableRequestContext.java
abbccdda: Yea, we need to have it here for audit logging. We could of course have the meta comment suggest "do not use for authorization"

clients/src/main/java/org/apache/kafka/common/requests/RequestHeader.java
abbccdda: We do have a constructor which doesn't contain the two fields above.

core/src/main/scala/kafka/network/SocketServer.scala
abbccdda: What if the user doesn't configure an inter-broker listener?

clients/src/main/java/org/apache/kafka/common/requests/RequestContext.java
```diff
@@ -74,7 +77,9 @@ public RequestAndSize parseRequest(ByteBuffer buffer) {
                         ", apiVersion: " + header.apiVersion() +
                         ", connectionId: " + connectionId +
                         ", listenerName: " + listenerName +
-                        ", principal: " + principal, ex);
+                        ", principal: " + principal +
+                        ", initial principal: " + initialPrincipalName() +
+                        ", initial client id: " + header.initialClientId(), ex);
```
dajac: nit: Could we use camel case like the others?

clients/src/main/java/org/apache/kafka/common/network/KafkaChannel.java
```diff
@@ -147,6 +153,7 @@ public KafkaChannel(String id, TransportLayer transportLayer, Supplier<Authentic
         this.disconnected = false;
         this.muteState = ChannelMuteState.NOT_MUTED;
         this.state = ChannelState.NOT_CONNECTED;
+        this.isInterBrokerListener = isInterBrokerListener;
```
cmccabe: How about `privilegedListener`?  Then we could add a comment explaining what that means:
1. if there is a control plane listener, only it is privileged
2. otherwise, if there is an inter broker listener, only it is privileged
3. otherwise, all listeners are privileged

clients/src/main/java/org/apache/kafka/clients/ClientRequest.java
```diff
@@ -42,6 +44,8 @@
      * @param clientId The client ID to use for the header
      * @param createdTimeMs The unix timestamp in milliseconds for the time at which this request was created.
      * @param expectResponse Should we expect a response message or is this request complete once it is sent?
+     * @param initialPrincipalName The initial principal name if this is a redirect request
```
cmccabe: "or null if this was not redirected"

clients/src/main/java/org/apache/kafka/clients/ClientRequest.java
```diff
@@ -42,6 +44,8 @@
...
+     * @param initialClientId The initial client id if this is a redirect request
```
cmccabe: "or null if this was not redirected"

clients/src/main/java/org/apache/kafka/clients/KafkaClient.java
```diff
@@ -189,16 +189,18 @@ ClientRequest newClientRequest(String nodeId, AbstractRequest.Builder<?> request
...
      * @param callback the callback to invoke when we get a response
+     * @param initialPrincipalName the initial client principal name, when building a forward request
+     * @param initialClientId the initial client id, when building a forward request
      */
     ClientRequest newClientRequest(String nodeId,
                                    AbstractRequest.Builder<?> requestBuilder,
                                    long createdTimeMs,
                                    boolean expectResponse,
                                    int requestTimeoutMs,
+                                   String initialPrincipalName,
```
cmccabe, line 200: can we have an overload for this that doesn't have these two fields, and defaults them to null?  It would really help avoid having so much churn in junit tests.

clients/src/main/java/org/apache/kafka/server/authorizer/AuthorizableRequestContext.java
cmccabe: Yes, can you add a comment here saying this should be used for logging, but not for authorization?

core/src/main/scala/kafka/network/RequestChannel.scala
cmccabe: I think `fromPrivilegedListener` is clearest

core/src/main/scala/kafka/network/SocketServer.scala
cmccabe: From the above:
1. if we have a control plane listener, the privileged listener is that one, and only that one
2. otherwise, if we have inter-broker listeners, only those are privileged,
3. otherwise, all listeners are privileged

We should have this in the JavaDoc for SocketServer

cmccabe:
We should also have a test of the privileged listener logic, to ensure that the right listeners are privileged in various configurations (not sure what junit test file that could go in ... socket server test maybe?)  Ideally we would do it as a unit test and not open sockets, etc.

clients/src/main/java/org/apache/kafka/clients/KafkaClient.java
abbccdda: I'm not sure it's meaningful to add an overload function just for the sake of minimizing code changes. Plus, we should not underestimate the more code changes potentially being added when we have 2 almost identical request builder function.

cmccabe: Fair enough

cmccabe:
LGTM

PR #9147, opened 2020-08-09
MINOR: supervise TransactionalMessageCopier producer

lbradstreet:
transactions_test and group_mode_transactions_test have proven quite brittle around timeouts (see 67f5b5de77d67c02edb335737215312d099a1cac, e099b58df5b3e4f87173fc55880f9c343308739f,
d9fe30dab0fc56318b012731c348ed1ddae2ec04, 07db26c20fcbccbf758591607864f7fd4bd8975f). This fights a losing battle, especially if we want to increase the types of nemesis checks that we want to perform on transaction support (e.g. iptables based partitioning). This PR creates a new producer when the prior one is in an unrecoverable state and still allows us to still test the EOS invariants required for this test.

### DIFF
--- a/tools/src/main/java/org/apache/kafka/tools/TransactionalMessageCopier.java
+++ b/tools/src/main/java/org/apache/kafka/tools/TransactionalMessageCopier.java
@@ -48,6 +48,7 @@ import java.util.Properties;
 import java.util.Random;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static java.util.Collections.singleton;
 import static net.sourceforge.argparse4j.impl.Arguments.store;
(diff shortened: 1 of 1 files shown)


tools/src/main/java/org/apache/kafka/tools/TransactionalMessageCopier.java
```diff
@@ -345,39 +357,55 @@ public void onPartitionsAssigned(Collection<TopicPartition> partitions) {
...
+                                    numMessagesProcessedSinceLastRebalance.get(), remainingMessages.get(), transactionalId, e));
+                            producer.get().abortTransaction();
+                            resetToLastCommittedPositions(consumer);
                         }
                     } catch (ProducerFencedException | OutOfOrderSequenceException e) {
-                        // We cannot recover from these errors, so just rethrow them and let the process fail
-                        throw e;
-                    } catch (KafkaException e) {
-                        producer.abortTransaction();
+                        // These failures are not recoverable with the same producer
```
lbradstreet, line 397: I think this is a reasonable way to deal with these issues as the main thing we are trying to test here is that we produce an exact copy from the input topic to output topic, however I do have questions about what kind of exceptions we should allow for. Originally my change allowed for supervision of ProducerFencedException(s) but not OutOfOrderSequenceException(s). If we do not expect those in any cases where we are messing with cluster behavior (broker bounces, network partitioning, etc) then I can take the handling out for the OoOS case.

tools/src/main/java/org/apache/kafka/tools/TransactionalMessageCopier.java
```diff
@@ -284,7 +296,7 @@ public static void main(String[] args) {
 
         String consumerGroup = parsedArgs.getString("consumerGroup");
 
-        final KafkaProducer<String, String> producer = createProducer(parsedArgs);
+        AtomicReference<KafkaProducer<String, String>> producer = new AtomicReference<>(createProducer(parsedArgs));
```
lbradstreet, line 299: volatiles can't be local. Arguably this could get refactored a little bit if we don't want to call get everywhere.

PR #9153, opened 2020-08-10
MINOR: Fix state transition diagram for stream threads

cadonna:


streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -93,7 +95,7 @@
      *          |      +-----+-------+      |
      *          +<---- | Partitions  |      |
      *          |      | Assigned (3)| <----+
-     *          |      +-----+-------+      |
+     *          |      +-----+-------+      ^
```
cadonna: Before this change a transition from `STARTING` to `RUNNING` was possible in the diagram. However, such a transition is not valid.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -80,11 +80,13 @@
...
-     *          |            v              |
-     *          |      +-----+-------+      |
-     *          +<---- | Partitions  |      |
+     *          |                           |
+     *          |                           |
+     *          |            +<----------+  |
+     *          |            |           |  |
+     *          |            v           |  |
+     *          |      +-----+-------+   |  |
+     *          +<---- | Partitions  | --+  |
```
cadonna: This valid loop was only mentioned in the notes but was not shown in the diagram.

cadonna:
Call for review: @guozhangwang @ableegoldman

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
ableegoldman: This seems a bit subtle. What if we instead break up the loop from Running --> Running below?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -93,7 +95,7 @@
      *          |      +-----+-------+      |
      *          +<---- | Partitions  |      |
      *          |      | Assigned (3)| <----+
-     *          |      +-----+-------+      |
+     *          |      +-----+-------+      ^
      *          |            |              |
      *          |            |--------------+
```
ableegoldman: It seems like this here is the problem...AFAICT this line is just to demonstrate that RUNNING -> RUNNING is valid, but it's tied in to the arrows connecting every other state that actually can't transition to Running. Can we change this to a self-contained loop like you added for PARTITIONS_REVOKED?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
ableegoldman: Nice catch

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
cadonna: I had the same feeling. I will change it.

ableegoldman (approved):
Thanks @cadonna ! LGTM

guozhangwang (approved):
LGTM

PR #9156, opened 2020-08-10
KAFKA-10077: Filter downstream of state-store results in spurious tombstones

big-andy-coates:
fixes: [KAFKA-10077](https://issues.apache.org/jira/browse/KAFKA-10077).

Enable sending old values on `KTable.filter` call to avoid the filter forwarding tombstones for rows that do not exist in the output.


big-andy-coates:
@mjsax as discussed. Please review.

I think this change makes the output from the table filter semantically correct, i.e. we no longer output tombstones for rows that didn't exist in the output to begin with. However, this comes at a cost!  Now the source table is being materialized, (as you can see from the changes needed to get some other tests to pass).

The cost of better semantics could be very high, and the user has no way of avoiding this. Where as previously the user could choose to 'fix' the bad semantics by manually calling `enableSendingOldValues` themselves, if they cared. I'm left feeling a little uneasy about _forcing_ users to pay the cost of materialization, even if they either don't care about the spurious tombstones, or their use case doesn't generate them.

This leads me to the following questions:

1. Wouldn't this be a breaking change for existing users of the library? If we stick with this solution, how would we handle this?
1. Might it be better to only enable the sending of old values _if_ the source table is already materialized? This would mean the fix only pays the cost of an additional rocksdb read, which is still not zero, but much lower that forced materialization, and it would also mean this isn't a breaking change.

Or maybe we choose to _not_ fix this. Preferring the current semantically incorrect, but better performing, solution with a known workaround for users that require correct semantics?  i.e. we could document the use of `enableSendingOldValues` in the `filter` method's java docs.

Your thoughts my good man?

mjsax:
> Where as previously the user could choose to 'fix' the bad semantics by manually calling enableSendingOldValues themselves, if they cared.

That is not entirely correct. `enableSendingOldValues` is _not_ part of the public `KTable` interface but only a method of the internal `KTableImpl` class. Thus, users don't not really have the ability to enable it.

> Might it be better to only enable the sending of old values if the source table is already materialized?

This seems to be the "right" fix from my point of view. Note, that people can force an upstream materialization via `Materialized.as()`, and thus the impact is quite clear and it's public API that allows user to opt-in for this optimization. (When we discussed in person, I alway assumed we would implement it this way). -- Doing it this way is also a non-breaking change and thus no backward compatibility concern raise.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
```diff
@@ -203,6 +203,10 @@ public String queryableStoreName() {
             processorSupplier,
             tableNode,
             builder);
+
+        kTable.enableSendingOldValues();
```
mjsax: Don't we need to call `this.enableSendingOldValues()` as "this" is the upstream KTable the filter is applied to? Why do we need to enable sending old values on the result of the filter() ?

streams/src/test/java/org/apache/kafka/streams/StreamsBuilderTest.java
```diff
@@ -123,7 +123,7 @@ public void shouldAllowJoinMaterializedFilteredKTable() {
 
         assertThat(
             topology.stateStores().size(),
-            equalTo(1));
+            equalTo(2));
```
mjsax: This is surprising to me -- need to dig into the code I guess, but I though enabling to send old values would _not_ trigger any upstream materialization but only take effect _if_ the store is materializes already (but would be ignored otherwise). Seems my assumption is wrong though.

mjsax: The issue seems to be this line: https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableSource.java#L67

In a source-KTable, if we enable sending old values we force an materialization. I guess, we should have this materialization optional?

I would recommend to rename `enableSendingOldValues()` to `maybeEnableSendingOldValues(final boolean forceMaterialization)` or similar?

\cc @guozhangwang WDYT?

guozhangwang: `enableSendingOldValues` is indicated for sending a pair of <new, old> values, and if it is not set the `old` value would always be null. If we want to `enableSendingOldValues` then we'd have to materialize the source node still?

Maybe I'm not fully understanding the context here. Could you bring me up to date?

mjsax: @guozhangwang Atm, when `enableSendingOldValues` is set and the upstream store is not materialized already we enforce a materialization. Thus, enabling `enableSendingOldValues` in `KTable#filter()` would be a breaking change as we would start to materialize state that did not exist before if one upgrades a topology.

Instead, we want to say, _iff_ the upstream store exist, please send me the old value, but if the upstream store does not exist, it's ok to just send `old=null` but don't force a materialization.

guozhangwang: Ah I see.. what about changing the function as `enableSendingOldValues(boolean enforced)` in that case? And then only ktable aggregations which would definitely need this would set it to `true`.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
big-andy-coates: Switching to enabling via the filter.

streams/src/test/java/org/apache/kafka/streams/StreamsBuilderTest.java
big-andy-coates: Sounds good.

big-andy-coates:
@guozhangwang @mjsax ready for another review.

clients/src/main/java/org/apache/kafka/server/authorizer/AuthorizableRequestContext.java
```diff
@@ -17,11 +17,12 @@
 
 package org.apache.kafka.server.authorizer;
 
-import java.net.InetAddress;
 import org.apache.kafka.common.annotation.InterfaceStability;
 import org.apache.kafka.common.security.auth.KafkaPrincipal;
 import org.apache.kafka.common.security.auth.SecurityProtocol;
 
+import java.net.InetAddress;
```
mjsax: Unrelated change: can we revert this?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableAggregate.java
```diff
@@ -47,8 +47,10 @@
     }
 
     @Override
-    public void enableSendingOldValues() {
+    public boolean enableSendingOldValues(final boolean onlyIfMaterialized) {
```
mjsax: Can we flip the logic of the boolean flag ? -- It makes a knob in my mind to say, `onlyIfMaterailized=false` implies that we need to (ie, enforce to) materialize...  I would prefer the suggest from Guozhang to use a flag like `enforceMaterialization` (or maybe better `materializeIfNeeded`).

Also, the method may or may not enable sending old values and thus, we might want to rename it to `maybeEnableSendOldValues` (or `enableSendOldValuesIfPossible` or similar).

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
```diff
@@ -182,6 +182,8 @@ public String queryableStoreName() {
         final KTableProcessorSupplier<K, V, V> processorSupplier =
             new KTableFilter<>(this, predicate, filterNot, queryableStoreName);
 
+        processorSupplier.enableSendingOldValues(true);
```
mjsax: Why does the `filter` need to send old values? I though the `filter` needs to _receive_ old values, and thus we should call `this. enableSendingOldValues(true)` to enable sending old values on the filter's input?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
```diff
@@ -832,18 +834,25 @@ public String queryableStoreName() {
...
     @SuppressWarnings("unchecked")
-    public void enableSendingOldValues() {
+    public boolean enableSendingOldValues(final boolean onlyIfMaterialized) {
         if (!sendOldValues) {
             if (processorSupplier instanceof KTableSource) {
                 final KTableSource<K, ?> source = (KTableSource<K, V>) processorSupplier;
+                if (onlyIfMaterialized && !source.materialized()) {
+                    return false;
+                }
                 source.enableSendingOldValues();
```
mjsax, line 842: Seems the boolean input parameter is missing?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableAbstractJoin.java
```diff
@@ -39,9 +39,15 @@
     }
 
     @Override
-    public final void enableSendingOldValues() {
-        table1.enableSendingOldValues();
-        table2.enableSendingOldValues();
+    public final boolean enableSendingOldValues(final boolean onlyIfMaterialized) {
+        if (!table1.enableSendingOldValues(onlyIfMaterialized)) {
```
mjsax: Instead of the check, should we pass in `false` to enforce a materialization if necessary?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableAbstractJoin.java
```diff
@@ -39,9 +39,15 @@
...
+            throw new IllegalStateException("Table-table joins should always be materialized");
+        }
+
+        if (!table2.enableSendingOldValues(onlyIfMaterialized)) {
```
mjsax: as above.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableJoinMerger.java
```diff
@@ -77,10 +77,16 @@ public String getQueryableName() {
     }
 
     @Override
-    public void enableSendingOldValues() {
-        parent1.enableSendingOldValues();
-        parent2.enableSendingOldValues();
+    public boolean enableSendingOldValues(final boolean onlyIfMaterialized) {
+        if (!parent1.enableSendingOldValues(onlyIfMaterialized)) {
```
mjsax: as above

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableJoinMerger.java
```diff
@@ -77,10 +77,16 @@ public String getQueryableName() {
...
+            throw new IllegalStateException("Table-table joins should always be materialized");
+        }
+
+        if (!parent2.enableSendingOldValues(onlyIfMaterialized)) {
```
mjsax: as above?

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableFilterTest.java
```diff
@@ -295,12 +295,39 @@ public void shouldNotSendOldValuesOnMaterialization() {
...
+    @Test
+    public void shouldNotSendOldValuesWithoutMaterializationIfOptionallyRequested() {
+        final StreamsBuilder builder = new StreamsBuilder();
+        final String topic1 = "topic1";
+
+        final KTableImpl<String, Integer, Integer> table1 =
+            (KTableImpl<String, Integer, Integer>) builder.table(topic1, consumed);
+        final KTableImpl<String, Integer, Integer> table2 = (KTableImpl<String, Integer, Integer>) table1.filter(predicate);
+
+        table2.enableSendingOldValues(true);
```
mjsax: Why do we need to call this (ie, why do we want the result of the filter to send old values)?

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableFilterTest.java
```diff
@@ -295,12 +295,39 @@ public void shouldNotSendOldValuesOnMaterialization() {
...
+    public void shouldNotSendOldValuesOnMaterializationIfOptionallyRequested() {
+        final StreamsBuilder builder = new StreamsBuilder();
+        final String topic1 = "topic1";
+
+        final KTableImpl<String, Integer, Integer> table1 =
+            (KTableImpl<String, Integer, Integer>) builder.table(topic1, consumed);
+        final KTableImpl<String, Integer, Integer> table2 =
+            (KTableImpl<String, Integer, Integer>) table1.filter(predicate, Materialized.as("store2"));
+
+        table2.enableSendingOldValues(true);
```
mjsax: As we enable this on the result `KTable`, and we enforce a materialization of the result, I would expect that we actually do get old-values sent?

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableFilterTest.java
```diff
@@ -295,12 +295,39 @@ public void shouldNotSendOldValuesOnMaterialization() {
...
+
+        doTestNotSendingOldValue(builder, table1, table2, topic1);
+    }
+
     private void doTestSendingOldValue(final StreamsBuilder builder,
                                        final KTableImpl<String, Integer, Integer> table1,
                                        final KTableImpl<String, Integer, Integer> table2,
                                        final String topic1) {
-        table2.enableSendingOldValues();
```
mjsax, line 304: Should we actually keep this call within this method and pass in the boolean flag as parameter to `doTestSendingOldValue` ?

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableImplTest.java
```diff
@@ -304,6 +363,30 @@ public void testStateStore() {
         }
     }
 
+    @Test
+    public void testSendingOldValuesNotSetIfNotMaterialized() {
```
mjsax: `shouldNotSetSendingOldValuesIfNotMaterialized`

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableImplTest.java
```diff
@@ -304,6 +363,30 @@ public void testStateStore() {
...
+        final KTableImpl<String, String, String> table =
+            (KTableImpl<String, String, String>) builder.table("topic1", consumed);
+
+        table.enableSendingOldValues(true);
+
+        assertThat(table.sendingOldValueEnabled(), is(false));
+    }
+
+    @Test
+    public void testSendingOldValuesSetIfMaterializedForced() {
```
mjsax: `shouldSetSendingOldValuesIfMaterializationForced`

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableImplTest.java
```diff
@@ -304,6 +363,30 @@ public void testStateStore() {
...
+        final StreamsBuilder builder = new StreamsBuilder();
+
+        final KTableImpl<String, String, String> table =
+            (KTableImpl<String, String, String>) builder.table("topic1", consumed);
```
mjsax, line 399: Seems the `Matererialized.as("fred")` parameter is missing?

streams/streams-scala/src/test/scala/org/apache/kafka/streams/scala/kstream/KTableTest.scala
```diff
@@ -39,29 +39,27 @@ class KTableTest extends FlatSpec with Matchers with TestDriver {
     val sinkTopic = "sink"
 
     val table = builder.stream[String, String](sourceTopic).groupBy((key, _) => key).count()
-    table.filter((_, value) => value > 1).toStream.to(sinkTopic)
+    table.filter((key, value) => key.equals("a") && value == 1).toStream.to(sinkTopic)
```
mjsax, line 42: Why do we need to change the filter condition?

mjsax (commented):
Couple of comments. Also, this PR needs to be rebased.

clients/src/main/java/org/apache/kafka/server/authorizer/AuthorizableRequestContext.java
big-andy-coates: Done. (Must sort out Kafka code style settings at some point)

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableAggregate.java
big-andy-coates: I did initially go with this. However, it seems really unintuitive. 

Given a param such as `forceMaterialization` then it's clear that `enableSendingOldValues(true)` will force materialization, but what does `enabledSendingOldValues(false)` mean? OK, we're not forcing materialization, but actually the method won't enable sending old values if the param is `false` and its not already materialized.

Even if we go with `maybeEnableSendingOldValues(enforceMaterialization)` I still would argue the semantics are not intuitive from the names alone.  (Though of course JavaDocs would help).

However, given `enableSendingOldValues(onlyIfMaterialized)` is, IMHO, very intuitive. `enableSendingOldValues(true)` will only enable sending old values if already materialized, whereas `enableSendingOldValues(false)` will always enable sending old values, materializing as necessary, must like the previous `enableSendingOldValues()` method did.

Likewise, if we stick with `enableSendingOldValues(onlyIfMaterialized)` then I don't think we need to include a `maybe` or `IfPossible` in the name as this is implied already by the `onlyIf`.

That said, this is not code I have to look at every day. If there's a consensus we should flip, then happy enough to do it.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
big-andy-coates: `KTableFilter` has a boolean flag internally that also needs to be set if it is to correctly handle old values, (existing code).  So if we don't call `enableSendingOldValues` on it, it won't swallow the output when things haven't changed.

to put it another way, the `sendOldValues` field of `KTableFilter` is used to both signify that the upstream is sending old values, and to control if the filter should forward old values. I'll split these into two variables.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
big-andy-coates: Nope.  The `if` above is handling the `boolean`, there is no need for the `source` to be aware of this.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableAbstractJoin.java
big-andy-coates: My view, [as outlined above](https://github.com/apache/kafka/pull/9156#discussion_r487057186), is to stick with the pattern that's here.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableAbstractJoin.java
big-andy-coates: as above :)

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableJoinMerger.java
big-andy-coates: as above :)

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableJoinMerger.java
big-andy-coates: as above. :p

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableFilterTest.java
big-andy-coates: See [my comment above](https://github.com/apache/kafka/pull/9156#discussion_r487057805)

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableAggregate.java
mjsax: I guess it's subjective. Personally, I would prefer to flip it.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
mjsax: SGTM.

IMHO, `sendOldValues` in `KTableFilter` should have only one meaning: do send old values downstream. If the filter result is materialized, we don't care if he upstream is sending old values or not. However, if the filter is stateless, as a side effect, we also need to tell upstream to send old values.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableAbstractJoin.java
mjsax: My comment was about your pattern... (if we flip the logic, we would need to pass in `true` to force a materialization). My point is, shouldn't we pass in a constant? For KTableKTableAbstractJoin we always want that the upstream is sending us old values.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
mjsax: Ah. I though you changes `ProcessorSuppler#enableSendOldValues` but that was incorrect. You change `KTableProcessorSuppler#enableSendOldValues`.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
big-andy-coates: > If the filter result is materialized, we don't care if he upstream is sending old values or not. However, if the filter is stateless, as a side effect, we also need to tell upstream to send old values.

That's not how the code was working. It always asked the parent to send old values:

```java
    @Override
    public void enableSendingOldValues() {
        parent.enableSendingOldValues();
        sendOldValues = true;
    }
```

Do we want to change this as part of this PR to only call `parent.enableSendingOldValues` is the filter is _not_ materialized?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableAbstractJoin.java
big-andy-coates: Feels a little off to me to ignore what the caller is asking.  If they pass `true` for `onlyIfMaterialized`, then we shouldn't silently ignore them IMHO.

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableImplTest.java
big-andy-coates: I prefer `should` too, but a lot of the code users `test`.  Happy to switch. Can I use `should` everywhere?

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableImplTest.java
big-andy-coates: Why? This test is about forced materialization. If we pass that parameter, then it would work even if not forced, right?

streams/streams-scala/src/test/scala/org/apache/kafka/streams/scala/kstream/KTableTest.scala
big-andy-coates: So that we can test an existing row being removed.

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KTableImplTest.java
big-andy-coates: Rename test to make it clear its force materialized due to sending old values, 

`shouldSetSendingOldValuesIfMaterializationForcedInternally`

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableAggregate.java
big-andy-coates: OK, will flip.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableImpl.java
big-andy-coates: As discussed offline - change the behaviour of `enableSendingOldValues` such that a node that is already materialized will not ask upstream nodes to also send old values, is deemed out of scope of this PR.

@mjsax has requested a ticket to track this work: https://issues.apache.org/jira/browse/KAFKA-10494

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTableKTableAbstractJoin.java
big-andy-coates: Anyway, done...

big-andy-coates:
@mjsax - ready for another review.

- boolean `onlyIfMaterialized` is now flipped to `forceMaterialization`.
- `KTableFilter` constructor initialized internal `sendOldValues` flag based on parent tables state. (Not perfect, but works).
- Removed test that was testing bad behaviour.

The only outstanding piece, as I see it, is the bad behaviour if you call `enableSendingOldValues` without forcing materialization on a table that is itself materialized, but who's upstream is not.  In such a situation, the table will _not_ enable sending old values. 

A similar bug existed before this PR:  if you called `enableSendingOldValues` on a table that is itself materialized, but who's upstream is not, then it will force materialization on its upstream table, which is unnecessary.

With the new code, both of the bad behaviours are true, i.e. the change introduces the first one and the second one is still happening.  

I've raised https://issues.apache.org/jira/browse/KAFKA-10494 to track this.

big-andy-coates:
retest this please

mjsax (approved):
Thanks for the fix @big-andy-coates! LGTM.

Will merge after Jenkins passed.

big-andy-coates:
test this please

mjsax:
Retest this please.

mjsax:
Thanks for the patch @big-andy-coates!

PR #9161, opened 2020-08-11
KAFKA-10385 - Remove no print stat when on detailed stat mode on Consumer Perf

sebadiaz:
The condition to see a detailed stats is conditionned by the time.
     if (currentTimeMillis - lastReportTime >= config.reportingInterval)
       if (config.showDetailedStats)

But when you finish the loop, you haven t the last stats of the performance tests. Just the previous status exists.

 

I propose just to remove the condition to show the final statistic result.


abbccdda (commented):
Could you update the description section?

sebadiaz:
@abbccdda update done

abbccdda:
@ijuma Do we have any possible dependency on the consumer performance stats? I personally think the change is safe, just want to double check.

PR #9162, opened 2020-08-11
MINOR: refactor Log to get rid of "return" in nested anonymous function

chia7712:
scala throws and then catches ```NonLocalReturnException``` to implement the control flow of returning from a nested anonymous function. That is anti-pattern and we should avoid using it in the hot methods.

I run the producer benchmark and the following screen snapshot show the improvement by this patch. The ```NonLocalReturnException``` is gone.

**BEFORE**

截圖 2020-09-10 上午11 53 30

**AFTER**

截圖 2020-09-10 上午11 52 26

### DIFF
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -1239,11 +1239,10 @@ class Log(@volatile private var _dir: File,
 
               // update the transaction index with the true last stable offset. The last offset visible
               // to consumers using READ_COMMITTED will be limited by this value and the high watermark.
-              completedTxns.foreach {
-                completedTxn =>
-                  val lastStableOffset = producerStateManager.lastStableOffset(completedTxn)
-                  segment.updateTxnIndex(completedTxn, lastStableOffset)
-                  producerStateManager.completeTxn(completedTxn)
+              completedTxns.foreach { completedTxn =>
+                val lastStableOffset = producerStateManager.lastStableOffset(completedTxn)
+                segment.updateTxnIndex(completedTxn, lastStableOffset)
+                producerStateManager.completeTxn(completedTxn)
               }
 
               // always update the last producer id map offset so that the snapshot reflects the current offset
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -1144,7 +1144,7 @@ class Log(@volatile private var _dir: File,
             // re-validate message sizes if there's a possibility that they have changed (due to re-compression or message
             // format conversion)
             if (!ignoreRecordSize && validateAndOffsetAssignResult.messageSizeMaybeChanged) {
-              for (batch <- validRecords.batches.asScala) {
+              validRecords.batches.forEach { batch =>
                 if (batch.sizeInBytes > config.maxMessageSize) {
                   // we record the original message set size instead of the trimmed size
                   // to be consistent with pre-compression bytesRejectedRate recording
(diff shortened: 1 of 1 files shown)


ijuma:
Are we using `return` inside an anonymous function here?

ijuma:
Ah, is `maybeHandleIOException` the issue?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1501,43 +1501,36 @@ class Log(@volatile private var _dir: File,
...
+      if (startOffset == maxOffsetMetadata.messageOffset)
+        emptyFetchDataInfo(maxOffsetMetadata, includeAbortedTxns)
+      else if (startOffset > maxOffsetMetadata.messageOffset)
+        emptyFetchDataInfo(convertToOffsetMetadataOrThrow(startOffset), includeAbortedTxns)
+      else {
+        // Do the read on the segment with a base offset less than the target offset
+        // but if that segment doesn't contain any messages with an offset greater than that
+        // continue to read from successive segments until we get some messages or we reach the end of the log
+        var currentEntry = segmentFloorEntry
+        Iterator.continually(currentEntry)
```
ijuma: Since this code is pretty critical, I'd keep the previous while loop and add a boolean to indicate a break or something like that.

chia7712: your are right!

chia7712:
I profiled one of our use case to trace the ```NonLocalReturnException```. ```Log#read``` and ```Log#fetchOffsetByTimestamp``` produces most of ```NonLocalReturnException``` so I only refactor that two methods to keep this patch small.

ijuma:
@chia7712 what are the other cases where we do this in `Log`?

chia7712:
> what are the other cases where we do this in Log?

The use case I mentioned is a heavy IO-load task of my (team) application. I didn't trace other potential hot methods for remaining use cases because it seems to me the IO-heavy is the most important case we have to care.

For another, ```Log``` has a bunch of anonymous functions which are using "return" to control flow. Maybe we should fix them at once even if they are not critical methods :)

ijuma:
Yeah, I would fix all the cases in this file at once if possible.

chia7712:
> Yeah, I would fix all the cases in this file at once if possible.

sure

chia7712:
```org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest > shouldUpgradeFromEosAlphaToEosBeta[true]``` failed. I test it on my local and it pass. retest this please

**Comment by chia7712** [comment]
only ```shouldUpgradeFromEosAlphaToEosBeta``` fails. retest this please

**Comment by chia7712** [comment]
retest this please

**Comment by chia7712** [comment]
```testMetadataRetries``` is traced by #9091

```DynamicConnectionQuotaTest.testDynamicListenerConnectionCreationRateQuota``` pass on my local

**Comment by chia7712** [comment]
retest this please

**Comment by chia7712** [comment]
@ijuma I have attached test result and ```NonLocalReturnException``` is gone. Please take a look.

**Comment by chia7712** [comment]
```
Build / JDK 11 / kafka.api.PlaintextConsumerTest.testLowMaxFetchSizeForRequestAndPartition
```
pass on my local

```
Build / JDK 15 / org.apache.kafka.clients.admin.KafkaAdminClientTest.testMetadataRetries
```
it is already fixed by https://github.com/apache/kafka/commit/e4eab377e1489fc0cc90edec3eeb382b1192a442

**Comment by chia7712** [comment]
```
Build / JDK 15 / org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]
Build / JDK 8 / kafka.api.ConsumerBounceTest.testClose
```

unrelated failure. rebase to trigger QA

**Comment by chia7712** [comment]
```
Build / JDK 8 / org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]
```
It is flaky on my local so it should be unrelated to this patch. @ijuma Could you take a look?

**Comment by chia7712** [comment]
```
Build / JDK 8 / org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]
```
unrelated error. rebase

**Comment by chia7712** [comment]
All green :)

@ijuma @hachikuji Could you take a look?

File: core/src/main/scala/kafka/log/Log.scala [hunk 2]
```diff
@@ -1535,42 +1533,48 @@ class Log(@volatile private var _dir: File,
...
+        while (!done) {
+          val segment = segmentEntry.getValue
+
+          val maxPosition = {
+            // Use the max offset position if it is on this segment; otherwise, the segment size is the limit.
+            if (maxOffsetMetadata.segmentBaseOffset == segment.baseOffset) {
+              maxOffsetMetadata.relativePositionInSegment
+            } else {
+              segment.size
+            }
```
ijuma: Nit: we can remove the braces from the if/else.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1535,42 +1533,48 @@ class Log(@volatile private var _dir: File,
...
-          if (maxOffsetMetadata.segmentBaseOffset == segment.baseOffset) {
-            maxOffsetMetadata.relativePositionInSegment
+          val fetchInfo = segment.read(startOffset, maxLength, maxPosition, minOneMessage)
+          if (fetchInfo == null) {
+            segmentEntry = segments.higherEntry(segmentEntry.getKey)
+            done = segmentEntry == null
           } else {
-            segment.size
+            fetchDataInfo = if (includeAbortedTxns) addAbortedTransactions(startOffset, segmentEntry, fetchInfo)
+            else fetchInfo
```
ijuma: The formatting here is a bit weird (the else is aligned the same as the variable. I think it's easier to understand if the if/else are aligned and indented.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1535,42 +1533,48 @@ class Log(@volatile private var _dir: File,
...
+            } else {
+              segment.size
+            }
+          }
 
-        val maxPosition = {
-          // Use the max offset position if it is on this segment; otherwise, the segment size is the limit.
-          if (maxOffsetMetadata.segmentBaseOffset == segment.baseOffset) {
-            maxOffsetMetadata.relativePositionInSegment
+          val fetchInfo = segment.read(startOffset, maxLength, maxPosition, minOneMessage)
```
ijuma: Maybe something like the following may be clearer:

```scala
fetchDataInfo = segment.read(startOffset, maxLength, maxPosition, minOneMessage)
if (fetchDataInfo != null) {
  if (includeAbortedTxns)
    fetchDataInfo = addAbortedTransactions(startOffset, segmentEntry, fetchDataInfo)
} else segmentEntry = segments.higherEntry(segmentEntry.getKey)

done = fetchDataInfo != null || segmentEntry == null
```

What do you think?

ijuma (commented):
Thanks. A few nits below.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1099,171 +1099,169 @@ class Log(@volatile private var _dir: File,
...
-        }
+          // now that we have valid records, offsets assigned, and timestamps updated, we need to
+          // validate the idempotent/transactional state of the producers and collect some metadata
+          val (updatedProducers, completedTxns, maybeDuplicate) = analyzeAndValidateProducerState(
+            logOffsetMetadata, validRecords, origin)
 
-        // always update the last producer id map offset so that the snapshot reflects the current offset
-        // even if there isn't any idempotent data being written
-        producerStateManager.updateMapEndOffset(appendInfo.lastOffset + 1)
+          if (maybeDuplicate.isDefined) {
```
ijuma: Seems like we should use a pattern match here instead of `isDefined` and `get`.

core/src/main/scala/kafka/log/Log.scala
chia7712: good style. Will copy that.

ijuma (approved):
LGTM once you correct the one nit comment in the last commit.

chia7712:
```
org.apache.kafka.streams.integration.StreamTableJoinTopologyOptimizationIntegrationTest.shouldDoStreamTableJoinWithDifferentNumberOfPartitions

org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta
```

unrelated error.

chia7712:
ok to test

chia7712:
add a trivial change to trigger QA again

chia7712:
```
org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta
```
it is traced by https://issues.apache.org/jira/browse/KAFKA-10017

PR #9166, opened 2020-08-11
KAFKA-10388; Fix struct conversion logic for tagged structures

hachikuji:
The message generator was missing conversion logic for tagged structures. This led to casting errors when either `fromStruct` or `toStruct` were invoked. In the process of fixing this, I also found that tagged byte arrays were missing null checks. I fixed this here so that the improved tests passed. 

I noticed as well that zero-copy byte arrays were not handled, but that can be addressed separately (https://issues.apache.org/jira/browse/KAFKA-10389).


cmccabe:
Thanks.  This is one of those rare PRs where I really don't have any comments on how the code could be better.

I do think we should do a round-trip test with the JSON converter as well, but that's out of scope for this change.

Also, I should note that I am working on separating out the JSON stuff into a helper class.  I think we may want to do this for the Struct stuff as well.  Or at least provide some option to not generate the Struct stuff in cases where we don't need the legacy methods (like in the KIP-500 generated messages).  But again, that's out of scope for this change.

LGTM.

hachikuji:
retest this please

PR #9168, opened 2020-08-12
MINOR: Ensure same version of scala library is used for compile and at runtime

rajinisivaram:
We are forcing runtime scala-library dependency to be the one used by Kafka in https://github.com/apache/kafka/commit/3fdf1523a369e9e2f03e215f39c0ed987b2a16c5, but we still seem to include the version in jackson for compile. This PR explicitly excludes it. 

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,8 +74,10 @@ allprojects {
         }
       }
     }
-    configurations {
-      runtime {
+    configurations.all {
+      // zinc is the Scala incremental compiler, it has a configuration for its own dependencies
+      // that are unrelated to the project dependencies, we should not change them
+      if (name != "zinc") {
         resolutionStrategy {
           force(
             // ensure we have a single version of scala jars in the classpath, we enable inlining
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -716,7 +716,9 @@ project(':core') {
   dependencies {
     compile project(':clients')
     compile libs.jacksonDatabind
-    compile libs.jacksonModuleScala
+    compile(libs.jacksonModuleScala) {
+      exclude module: 'scala-library'
```
ijuma: Can we double check that jackson doesn't have a dependency on scala-reflect?

rajinisivaram:
@ijuma I have updated the PR using `configurations.all` that you suggested. Thank you!

ijuma (approved):
LGTM. Thanks for the fix!

rajinisivaram:
@ijuma Thanks for the review, streams test failures not related. Merging to trunk, 2.6, 2,5 and 2.4.

PR #9169, opened 2020-08-12
MINOR: the scheduler used to perform rebalance should have thread prefix

chia7712:
There is already a scheduler (see https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/KafkaServer.scala#L264) using default thread prefix so the others should define different thread name. Otherwise, it is hard to distinguish them by JVM profiler.


abbccdda (commented):
Thanks for the change. I'm not sure whether this would break any metrics on broker side, might be good to have @ijuma also take a look.

chia7712:
@ijuma Does it need a KIP to resolve naming conflict in metrics? Although I'm wonder whether is there a use case using this conflicted name.

PR #9170, opened 2020-08-12
KAFKA-10391: Overwrite checkpoint in task corruption to remove corrupted partitions

guozhangwang:
In order to do this, I also removed the optimization such that once enforced checkpoint is set to true, we always checkpoint unless the state stores are not initialized at all (i.e. the snapshot is null).

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateManagerUtil.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateManagerUtil.java
@@ -58,6 +58,9 @@ final class StateManagerUtil {
             return false;
         }
 
+        if (enforceCheckpoint)
+            return true;
+
         // we can checkpoint if the the difference between the current and the previous snapshot is large enough
         long totalOffsetDelta = 0L;
         for (final Map.Entry<TopicPartition, Long> entry : newOffsetSnapshot.entrySet()) {
@@ -66,7 +69,7 @@ final class StateManagerUtil {
 
         // when enforcing checkpoint is required, we should overwrite the checkpoint if it is different from the old one;
         // otherwise, we only overwrite the checkpoint if it is largely different from the old one
-        return enforceCheckpoint ? totalOffsetDelta > 0 : totalOffsetDelta > OFFSET_DELTA_THRESHOLD_FOR_CHECKPOINT;
+        return totalOffsetDelta > OFFSET_DELTA_THRESHOLD_FOR_CHECKPOINT;
     }
 
     /**
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
@@ -193,13 +193,16 @@ public class TaskManager {
 
             try {
                 task.suspend();
+                task.postCommit(true);
             } catch (final RuntimeException swallow) {
                 log.error("Error suspending corrupted task {} ", task.id(), swallow);
             }
             task.closeDirty();
+
+            // For active tasks pause their input partitions so we won't poll any more records
+            // for this task until it has been re-initialized;
+            // Note, closeDirty already clears the partitiongroup for the task.
             if (task.isActive()) {
-                // Pause so we won't poll any more records for this task until it has been re-initialized
-                // Note, closeDirty already clears the partitiongroup for the task.
                 final Set<TopicPartition> currentAssignment = mainConsumer().assignment();
                 final Set<TopicPartition> taskInputPartitions = task.inputPartitions();
                 final Set<TopicPartition> assignedToPauseAndReset =
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
@@ -193,6 +193,7 @@ public class TaskManager {
 
             try {
                 task.suspend();
+                // we need to enforce a checkpoint that removes the corrupted partitions
                 task.postCommit(true);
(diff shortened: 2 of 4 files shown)


guozhangwang:
Successful system tests: 

https://jenkins.confluent.io/job/system-test-kafka-branch-builder/4121/
https://jenkins.confluent.io/job/system-test-kafka-branch-builder/4122/

guozhangwang:
cc @ableegoldman @abbccdda for reviews

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -1609,12 +1609,11 @@ public void shouldReturnStateManagerChangelogOffsets() {
     }
 
     @Test
-    public void shouldCheckpointWithCreatedStateOnClose() {
+    public void shouldNotCheckpointOnCloseCreated() {
```
ableegoldman, line 1612: Why did this test change?

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -1629,16 +1628,18 @@ public void shouldCheckpointWithCreatedStateOnClose() {
...
 
+        EasyMock.verify(stateManager, recordCollector);
+
         final double expectedCloseTaskMetric = 1.0;
         verifyCloseTaskMetric(expectedCloseTaskMetric, streamsMetrics, metricName);
     }
 
     @Test
-    public void shouldNotCheckpointOnCloseRestoringIfNoProgress() {
+    public void shouldCheckpointOnCloseRestoringIfNoProgress() {
```
ableegoldman, line 1638: Same with this one, why did it change? Or was the test just wrong to begin with..?

ableegoldman (commented):
Do you have any theories as to why this was causing `StreamsUpgradeTest.test_app_upgrade` to be flaky? Do we really get that many TaskCorruptedExceptions in this system test?

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -193,13 +193,17 @@ private void closeAndRevive(final Map<Task, Collection<TopicPartition>> taskWith
...
+                // we need to enforce a checkpoint that removes the corrupted partitions
+                task.postCommit(true);
             } catch (final RuntimeException swallow) {
                 log.error("Error suspending corrupted task {} ", task.id(), swallow);
             }
             task.closeDirty();
+
+            // For active tasks pause their input partitions so we won't poll any more records
+            // for this task until it has been re-initialized;
+            // Note, closeDirty already clears the partitiongroup for the task.
```
abbccdda, line 205: partition group

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -193,13 +193,17 @@ private void closeAndRevive(final Map<Task, Collection<TopicPartition>> taskWith
 
             try {
                 task.suspend();
+                // we need to enforce a checkpoint that removes the corrupted partitions
+                task.postCommit(true);
```
abbccdda, line 197: I'm not sure this is 100% ensuring the snapshot gets done, as we have branching logic in postCommit. Do you think we should just get a helper to cleanup instead?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateManagerUtil.java
```diff
@@ -58,6 +58,9 @@ static boolean checkpointNeeded(final boolean enforceCheckpoint,
             return false;
         }
 
+        if (enforceCheckpoint)
```
abbccdda, line 61: Could we move this logic to the `StandbyTask` only? It is the only case I have seen which could have null snapshot passed in, which could make this helper assume both snapshots are not null.
```
if (oldOffsetSnapshot == null) {
            return false;	          
}	        
```

guozhangwang:
> Do you have any theories as to why this was causing `StreamsUpgradeTest.test_app_upgrade` to be flaky? Do we really get that many TaskCorruptedExceptions in this system test?

I observed several of the failed tests are due to the endless loop of the task-corruption. Basically we shutdown all applications and then restart them, and in between the log truncation may trigger and hence cause restoration hit out-of-range. And due to this bug we would fall in the loop forever.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
guozhangwang: It was the same as the other one, that it was wrong to begin with.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
guozhangwang: Here since the task is guaranteed to be in SUSPENDED `postCommit()` is ensured to trigger checkpointing.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateManagerUtil.java
guozhangwang: The null snapshot is for both active / standby tasks before the initializeIfNeeded is triggered, in that case, we should not overwrite the checkpoint even if it was enforced.

ableegoldman (approved):
I see, I didn't realize the log truncation would trigger in the middle of a system test. Sounds like we have good test coverage for TaskCorruptedException after all

LGTM

abbccdda (approved):

PR #9172, opened 2020-08-12
KAFKA-10387: Fix inclusion of transformation configs when topic creation is enabled in Connect

kkonstantine:
Addition of configs for custom topic creation with KIP-158 created a regression when transformation configs are also included in the configuration of a source connector. 

To experience the issue, just enabling topic creation at the worker is not sufficient. A user needs to supply a source connector configuration that contains both transformations and custom topic creation properties. 

The issue is that the enrichment of configs in `SourceConnectorConfig` happens on top of an `AbstractConfig` rather than a `ConnectorConfig`. Inheriting from the latter allows enrichment to be composable for both topic creation and transformations. 

Unit tests and integration tests are written to test these combinations. 

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/SourceConnectorConfig.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/SourceConnectorConfig.java
@@ -130,8 +130,8 @@ public class SourceConnectorConfig extends ConnectorConfig {
                     .removeIf(e -> e.getKey().equals(DEFAULT_TOPIC_CREATION_PREFIX + INCLUDE_REGEX_CONFIG)
                             || e.getKey().equals(DEFAULT_TOPIC_CREATION_PREFIX + EXCLUDE_REGEX_CONFIG));
             enrichedSourceConfig = new EnrichedSourceConnectorConfig(plugins,
-                    enrich(defaultConfigDef, props,
-                    defaultGroup), propsWithoutRegexForDefaultGroup);
+                    enrich(defaultConfigDef, props, defaultGroup),
+                    propsWithoutRegexForDefaultGroup);
         } else {
             enrichedSourceConfig = null;
         }
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/SourceConnectorConfig.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/SourceConnectorConfig.java
@@ -47,9 +47,9 @@ public class SourceConnectorConfig extends ConnectorConfig {
             + "created by source connectors";
     private static final String TOPIC_CREATION_GROUPS_DISPLAY = "Topic Creation Groups";
 
-    private static class EnrichedSourceConnectorConfig extends AbstractConfig {
-        EnrichedSourceConnectorConfig(ConfigDef configDef, Map<String, String> props) {
-            super(configDef, props);
+    private static class EnrichedSourceConnectorConfig extends ConnectorConfig {
+        EnrichedSourceConnectorConfig(Plugins plugins, ConfigDef configDef, Map<String, String> props) {
+            super(plugins, configDef, props);
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


kkonstantine:
I'll be adding a few more tests.

kkonstantine:
cc @rhauch @wicknicks

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/SourceConnectorConfig.java
```diff
@@ -129,7 +129,8 @@ public SourceConnectorConfig(Plugins plugins, Map<String, String> props, boolean
             propsWithoutRegexForDefaultGroup.entrySet()
                     .removeIf(e -> e.getKey().equals(DEFAULT_TOPIC_CREATION_PREFIX + INCLUDE_REGEX_CONFIG)
                             || e.getKey().equals(DEFAULT_TOPIC_CREATION_PREFIX + EXCLUDE_REGEX_CONFIG));
-            enrichedSourceConfig = new EnrichedSourceConnectorConfig(enrich(defaultConfigDef, props,
+            enrichedSourceConfig = new EnrichedSourceConnectorConfig(plugins,
+                    enrich(defaultConfigDef, props,
                     defaultGroup), propsWithoutRegexForDefaultGroup);
```
rhauch: Super tiny nit: the wrapping even from the original was confusing, and maybe it's worth changing one more line and cleaning up the wrapping:
```suggestion
            enrichedSourceConfig = new EnrichedSourceConnectorConfig(plugins,
                    enrich(defaultConfigDef, props, defaultGroup), 
                    propsWithoutRegexForDefaultGroup);
```
WDYT?

connect/runtime/src/test/java/org/apache/kafka/connect/integration/TransformationIntegrationTest.java
```diff
@@ -318,4 +323,61 @@ public void testFilterOnHasHeaderKeyWithSourceConnector() throws Exception {
         // delete connector
         connect.deleteConnector(CONNECTOR_NAME);
     }
+
+    /**
+     * Test the {@link Filter} transformer with a
+     * {@link HasHeaderKey} predicate on a source connector.
+     */
+    @Test
+    public void testFilterOnHasHeaderKeyWithSourceConnectorAndTopicCreation() throws Exception {
```
rhauch: Do we need another test here, or could we essentially replace the previous test with this one?

Either way, the test that uses topic creation should probably have a comment (either in the method JavaDoc or in a line comment below) that explicitly mentions enabling topic creation. We don't want someone removing the topic creation configs at a later time.

WDYT?

rhauch (commented):
Thanks, @kkonstantine. It's nice to see a straightforward fix for this issue. A few nits, and one question below.

kkonstantine:
According to https://status.apache.org the jenkins queue has grown too much and the infra team is investigating how to resolve the issues. Thus, the tests haven't run yet.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/SourceConnectorConfig.java
kkonstantine: Good catch. The break was added by accident there.

connect/runtime/src/test/java/org/apache/kafka/connect/integration/TransformationIntegrationTest.java
kkonstantine: Since we have a few other tests replacing should be fine. 

I'll add comments. But also, the test will fail if we don't create it (I removed the admin command) because we disable auto topic creation at the broker.

rhauch: Sounds good. I just want to avoid someone trying to simplify the tests in the future without understanding that this test is verifying both features work together.

kkonstantine: I added the logic to the previous test. 

Added another unit test to test 2 groups with 2 transformations.

kkonstantine:
Thanks for the comments @rhauch !
The PR should be ready for another pass. Jenkins still has a long queue of jobs. Yesterday 2/3 builds were green. I'm expecting the new job to eventually run today too (Builds locally).

rhauch (approved):
LGTM, pending a green build. Thanks, @kkonstantine!

kkonstantine:
Thanks @rhauch ! 
Tested manually with transforms from the `plugin.path` as well. Merging on `trunk` and backporting to `2.6`.  

(2/3 builds were green)

PR #8834, opened 2020-06-09
KAFKA-10134: Enable heartbeat during PrepareRebalance and Depend On State For Poll Timeout

guozhangwang:
1. Split the consumer coordinator's `REBALANCING` state into `PREPARING_REBALANCE ` and `COMPLETING_REBALANCE`. The first is when the join group request is sent, and the second is after the join group response is received. During the first state we should still not send hb since it shares the same socket with the join group request and the group coordinator has disabled timeout, however when we transit to the second state we should start sending hb in case leader's `assign` takes long time. This is also for fixing KAFKA-10122.

2. When deciding `coordinator#timeToNextPoll`, do not count in timeToNextHeartbeat if the state is in `UNJOINED` or `PREPARING_REBALANCE` since we would disable hb and hence its timer would not be updated.

3. On the broker side, allow hb received during `PREPARING_REBALANCE`, return NONE error code instead of REBALANCE_IN_PROGRESS. However on client side, we still need to ignore REBALANCE_IN_PROGRESS if state is COMPLETING_REBALANCE in case it is talking to an old versioned broker.

4. Piggy-backing a log4j improvement on the broker coordinator for triggering rebalance reason, as I found it a bit blurred during the investigation. Also subsumed https://github.com/apache/kafka/pull/9038 with log4j improvements.

The tricky part for allowing hb during `COMPLETING_REBALANCE` is in two parts: 1) before the sync-group response is received, a hb response may have reset the generation; also after the sync-group response but before the callback is triggered, a hb response can still reset the generation, we need to handle both cases by checking the generation / state. 2) with the hb thread enabled, the sync-group request may be sent by the hb thread even if the caller thread did not call `poll` yet.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9173, opened 2020-08-12
KAFKA-10122: Consumer should allow heartbeat during rebalance

ableegoldman:
We launched a Streams application reading from a single 3000-partition topic and saw continuous rebalancing. Digging into the logs, every time the leader sent a SyncGroup request it would discover that it had dropped out of the group and needed to rejoin. The assignment seemed to take slightly longer than 10s, the session interval, so it seemed to be getting kicked due to heartbeat expiration while the heartbeat thread was disabled.

I redeployed the app with this exact patch and saw it stabilize at last. The application was left running for ~20 hours or so and never rebalanced again after the last pod was rolled


ableegoldman:
Closing this since I didn't see the existing PR for this bugfix: https://github.com/apache/kafka/pull/8834

PR #9174, opened 2020-08-13
KAFKA-10395: relax output topic check in TTD to work with dynamic routing

ableegoldman:
I went through all 5 stages of grief in thinking about what to do here and decided the best thing was to just relax the check after all. Hopefully users who find their output topic unexpectedly empty due to a typo in the topic name will be able to figure it out quickly from the warning we now log instead.

Not sure how rampant the problem of output-topic-typos is to begin with...

### DIFF
--- a/streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
+++ b/streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
@@ -808,7 +808,7 @@ public class TopologyTestDriver implements Closeable {
         if (outputRecords == null && !processorTopology.sinkTopics().contains(topicName)) {
             log.warn("Unrecognized topic: {}, this can occur if dynamic routing is used and no output has been "
                          + "sent to this topic yet. If not using a TopicNameExtractor, check that the output topic "
-                         + "is correct. ", topicName);
+                         + "is correct.", topicName);
         }
         return outputRecords;
     }
--- a/streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
+++ b/streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
@@ -805,10 +805,10 @@ public class TopologyTestDriver implements Closeable {
 
     private Queue<ProducerRecord<byte[], byte[]>> getRecordsQueue(final String topicName) {
         final Queue<ProducerRecord<byte[], byte[]>> outputRecords = outputRecordsByTopic.get(topicName);
-        if (outputRecords == null) {
-            if (!processorTopology.sinkTopics().contains(topicName)) {
-                throw new IllegalArgumentException("Unknown topic: " + topicName);
-            }
+        if (outputRecords == null && !processorTopology.sinkTopics().contains(topicName)) {
+            log.warn("Unrecognized topic: {}, this can occur if dynamic routing is used and no output has been "
+                         + "sent to this topic yet. If not using a TopicNameExtractor, check that the output topic "
+                         + "is correct. ", topicName);
         }
         return outputRecords;
     }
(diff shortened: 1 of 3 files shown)


ableegoldman:
@cadonna @guozhangwang

streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
```diff
@@ -805,10 +805,10 @@ public void advanceWallClockTime(final Duration advance) {
...
     private Queue<ProducerRecord<byte[], byte[]>> getRecordsQueue(final String topicName) {
         final Queue<ProducerRecord<byte[], byte[]>> outputRecords = outputRecordsByTopic.get(topicName);
-        if (outputRecords == null) {
-            if (!processorTopology.sinkTopics().contains(topicName)) {
-                throw new IllegalArgumentException("Unknown topic: " + topicName);
-            }
+        if (outputRecords == null && !processorTopology.sinkTopics().contains(topicName)) {
+            log.warn("Unrecognized topic: {}, this can occur if dynamic routing is used and no output has been "
+                         + "sent to this topic yet. If not using a TopicNameExtractor, check that the output topic "
+                         + "is correct. ", topicName);
```
abbccdda: nit: could avoid the last space after `is correct.`

streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
```diff
@@ -805,10 +805,10 @@ public void advanceWallClockTime(final Duration advance) {
 
     private Queue<ProducerRecord<byte[], byte[]>> getRecordsQueue(final String topicName) {
         final Queue<ProducerRecord<byte[], byte[]>> outputRecords = outputRecordsByTopic.get(topicName);
-        if (outputRecords == null) {
-            if (!processorTopology.sinkTopics().contains(topicName)) {
-                throw new IllegalArgumentException("Unknown topic: " + topicName);
-            }
+        if (outputRecords == null && !processorTopology.sinkTopics().contains(topicName)) {
```
abbccdda, line 808: Seems good enough as a bug fix, but I was wondering whether we could detect the dynamic topic is configured or not to make sure we are not actually allowing some other bugs to catch in TTD

cadonna (approved):
LGTM! Thanks, @ableegoldman

cadonna:
Failure seem related

streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
ableegoldman: What do you mean by "configured"? My understanding is that topics don't actually even "exist" in the TTD, we just use a MockProducer and grab output records from its history to enqueue in the `outputRecordsByTopic` map. So if an output topic hasn't been sent to yet, there's no way to say whether it does or does not exist

vvcephei: We could detect if the processorTopology contains only `StaticTopicNameExtractors` and still throw in that case if the topic name isn't in the topology.

ableegoldman: Yeah, that's a good point. Took a quick look at this and personally I feel it's not worth it to add even more things for the internal topology to track. But it wouldn't really add any complexity so if this is a big concern I can go ahead and add in a flag to watch out for any non-static TopicNameExtractors

vvcephei: Ok, let's just keep it in our back pocket for now.

vvcephei (approved):
Thanks, @ableegoldman !

mjsax:
Might it be worth to cherry-pick to `2.6` ?

ableegoldman:
@mjsax Be my guest 🙂

mjsax:
Cherry-picked to `2.6` branch.

PR #9176, opened 2020-08-13
Allow replace all for RegexRouter

carlos-verdes:
### Context 
I need to create topics on the fly from database table names (cdc) and I'm replacing all `_` with `-`.
With current RegexRouter I can't do this in one call (actually if I don't know the number of `_` I don't know how many times I need to call the transformer).

That's the reason why I implemented a new flag to allow replaceAll on RegexRouter.

### Change applied
I added a new flag `replaceAll` of type `Boolean` with default value `false` which allows to apply a regex replacement in all occurencces of the topic name.
I updated the description of the transformer and added 2 tests to cover previous and new behaviour.

### Testing strategy (Unit test)
I created a new topic name with two `_` symbols (`sample_table_name`).
I called a regex with current implementation expecting only the first `_` to be substituted.
I called a regex with new flag expecting all `_` to be replaced with `-`.

### DIFF
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/RegexRouter.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/RegexRouter.java
@@ -29,33 +29,43 @@ public class RegexRouter<R extends ConnectRecord<R>> implements Transformation<R
 
     public static final String OVERVIEW_DOC = "Update the record topic using the configured regular expression and replacement string."
             + "<p/>Under the hood, the regex is compiled to a <code>java.util.regex.Pattern</code>. "
-            + "If the pattern matches the input topic, <code>java.util.regex.Matcher#replaceFirst()</code> is used with the replacement string to obtain the new topic.";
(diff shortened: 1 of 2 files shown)


carlos-verdes:
@kkonstantine can you help me with this PR or point to me to someone who can help?
I really would appreciate :)

kkonstantine:
Thanks @carlos-verdes 
I'll try to take a look within the next couple of weeks (taking some time off in the meantime). 

Without commenting on the approach yet, I'm pretty sure this will require a short KIP to denote the change on the public API for this transformation. Should be straightforward to write given that the change is small. 

cc @rhauch

rhauch:
@kkonstantine is correct that this changes the public API and therefore requires a KIP. 

Please also create a Jira issue, and change the title of this PR so that it starts with the issue (e.g., `KAFKA-12345: ` but with your issue number).  This will auto-link this PR in the Jira issue.

carlos-verdes:
I will be happy to do but I may need help about the KIP.
Do you need me to open a Jira ticket to move forward?

rhauch:
Yes, please create an issue as described above.

PR #9183, opened 2020-08-14
KAFKA-10404; Use higher poll timeout to avoid rebalance in testCoordinatorFailover

rajinisivaram:
Tests use 6s poll timeout, which isn't sufficient to ensure that clients don't leave the group due to poll timeout. The PR increases poll timeout to 15s. Session timeout of 5s is also low, but leaving it as-is for now with lower heartbeat timeout to make sure we catch unexpected rebalances during the failover.


omkreddy (approved):

rajinisivaram:
@omkreddy Thanks for the review, merging to trunk, 2.6 and 2.5.

PR #9186, opened 2020-08-16
KAFKA-10277: Allow null keys with non-null mappedKey in KStreamKGlobalTable join

JoelWee:

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
@@ -25,6 +25,8 @@ import org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Optional;
+
 import static org.apache.kafka.streams.processor.internals.metrics.TaskMetrics.droppedRecordsSensorOrSkippedRecordsSensor;
 import static org.apache.kafka.streams.state.ValueAndTimestamp.getValueOrNull;
 
@@ -58,27 +60,44 @@ class KStreamKTableJoinProcessor<K1, K2, V1, V2, R> extends AbstractProcessor<K1
 
     @Override
     public void process(final K1 key, final V1 value) {
-        // we do join iff keys are equal, thus, if key is null we cannot join and just ignore the record
-        // If {@code keyMapper} returns {@code null} it implies there is no match,
-        // so ignore unless it is a left join
+        // we do join iff the joining keys are equal, thus, if the mappedKey is null we cannot join
+        // and just ignore the record.
         //
         // we also ignore the record if value is null, because in a key-value data model a null-value indicates
         // an empty message (ie, there is nothing to be joined) -- this contrast SQL NULL semantics
         // furthermore, on left/outer joins 'null' in ValueJoiner#apply() indicates a missing record --
         // thus, to be consistent and to avoid ambiguous null semantics, null values are ignored
-        if (key == null || value == null) {
+        final Optional<K2> maybeMappedKey = maybeExtractMappedKey(key, value);
+        if (!maybeMappedKey.isPresent()) {
             LOG.warn(
                 "Skipping record due to null key or value. key=[{}] value=[{}] topic=[{}] partition=[{}] offset=[{}]",
                 key, value, context().topic(), context().partition(), context().offset()
             );
             droppedRecordsSensor.record();
         } else {
-            final K2 mappedKey = keyMapper.apply(key, value);
-            final V2 value2 = mappedKey == null ? null : getValueOrNull(valueGetter.get(mappedKey));
+            final K2 mappedKey = maybeMappedKey.get();
+            final V2 value2 = getValueOrNull(valueGetter.get(mappedKey));
             if (leftJoin || value2 != null) {
                 context().forward(key, joiner.apply(value, value2));
             }
         }
+    }
+
+    private Optional<K2> maybeExtractMappedKey(final K1 key, final V1 value) {
+        if (value == null) {
+            return Optional.empty();
+        }
+
+        // we allow the case where the key is null but mappedKey is not null and thus
+        // we need to guard against nullPointerExceptions. This may happen for GlobalKTables.
+        // For KTables, the keyMapper simply returns the key, so this will never happen
+        Optional<K2> maybeMappedKey;
+        try {
+            maybeMappedKey = Optional.ofNullable(keyMapper.apply(key, value));
+        } catch (final NullPointerException e) {
+            maybeMappedKey = Optional.empty();
+        }
+        return maybeMappedKey;
     }
 
     @Override
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
@@ -25,8 +25,6 @@ import org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Optional;
-
 import static org.apache.kafka.streams.processor.internals.metrics.TaskMetrics.droppedRecordsSensorOrSkippedRecordsSensor;
 import static org.apache.kafka.streams.state.ValueAndTimestamp.getValueOrNull;
 
@@ -67,37 +65,19 @@ class KStreamKTableJoinProcessor<K1, K2, V1, V2, R> extends AbstractProcessor<K1
         // an empty message (ie, there is nothing to be joined) -- this contrast SQL NULL semantics
         // furthermore, on left/outer joins 'null' in ValueJoiner#apply() indicates a missing record --
         // thus, to be consistent and to avoid ambiguous null semantics, null values are ignored
-        final Optional<K2> maybeMappedKey = maybeExtractMappedKey(key, value);
-        if (!maybeMappedKey.isPresent()) {
+        final K2 mappedKey = keyMapper.apply(key, value);
+        if (mappedKey == null || value == null) {
             LOG.warn(
                 "Skipping record due to null key or value. key=[{}] value=[{}] topic=[{}] partition=[{}] offset=[{}]",
                 key, value, context().topic(), context().partition(), context().offset()
             );
             droppedRecordsSensor.record();
         } else {
-            final K2 mappedKey = maybeMappedKey.get();
             final V2 value2 = getValueOrNull(valueGetter.get(mappedKey));
             if (leftJoin || value2 != null) {
                 context().forward(key, joiner.apply(value, value2));
             }
         }
-    }
-
-    private Optional<K2> maybeExtractMappedKey(final K1 key, final V1 value) {
-        if (value == null) {
-            return Optional.empty();
-        }
-
-        // we allow the case where the key is null but mappedKey is not null and thus
-        // we need to guard against nullPointerExceptions. This may happen for GlobalKTables.
-        // For KTables, the keyMapper simply returns the key, so this will never happen
-        Optional<K2> maybeMappedKey;
-        try {
-            maybeMappedKey = Optional.ofNullable(keyMapper.apply(key, value));
-        } catch (final NullPointerException e) {
-            maybeMappedKey = Optional.empty();
-        }
-        return maybeMappedKey;
     }
 
     @Override
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
@@ -58,22 +58,24 @@ class KStreamKTableJoinProcessor<K1, K2, V1, V2, R> extends AbstractProcessor<K1
 
     @Override
     public void process(final K1 key, final V1 value) {
-        // we do join iff the joining keys are equal, thus, if the mappedKey is null we cannot join
-        // and just ignore the record.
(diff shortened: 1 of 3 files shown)


JoelWee:
[KAFKA-10277](https://issues.apache.org/jira/browse/KAFKA-10277?jql=project%20%3D%20KAFKA%20AND%20labels%20%3D%20newbie%20AND%20status%20%3D%20Open%20ORDER%20BY%20updated%20DESC)

Hi @mjsax, please could you have a look? :)

It feels like if implemented this way, we should have a NullPointerException test for the processor, but I'm not sure where that test should be put in. It fits best as a direct unit test for the processor but it doesn't look like any of those tests are done. And it's somewhat inconvenient to add it to the existing join tests because of the way they are set up

mjsax:
Thanks for the PR @JoelWee -- Not exactly sure what NPE test you have in mind, but feel free to add a new test class `src/test/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessorTest.java` if required.

mjsax:
@ableegoldman Can you help to review?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
```diff
@@ -58,29 +60,46 @@ public void init(final ProcessorContext context) {
...
+            return Optional.empty();
+        }
+
+        // we allow the case where the key is null but mappedKey is not null and thus
+        // we need to guard against nullPointerExceptions. This may happen for GlobalKTables.
+        // For KTables, the keyMapper simply returns the key, so this will never happen
+        Optional<K2> maybeMappedKey;
+        try {
+            maybeMappedKey = Optional.ofNullable(keyMapper.apply(key, value));
+        } catch (final NullPointerException e) {
```
ableegoldman: Should we really catch NPE here? It seems like if the user wants to return a non-null mapped key from a null key, then they should handle the null case specifically in their `keyMapper` and not just throw an NPE. In general, an NPE is a sign that something has gone wrong. I would be pretty surprised if I threw an NPE explicitly in my user code and it just got swallowed and interpreted as if I had actually returned null.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
```diff
@@ -58,29 +60,46 @@ public void init(final ProcessorContext context) {
...
+        final Optional<K2> maybeMappedKey = maybeExtractMappedKey(key, value);
+        if (!maybeMappedKey.isPresent()) {
             LOG.warn(
                 "Skipping record due to null key or value. key=[{}] value=[{}] topic=[{}] partition=[{}] offset=[{}]",
                 key, value, context().topic(), context().partition(), context().offset()
             );
             droppedRecordsSensor.record();
         } else {
-            final K2 mappedKey = keyMapper.apply(key, value);
-            final V2 value2 = mappedKey == null ? null : getValueOrNull(valueGetter.get(mappedKey));
```
ableegoldman, line 77: Why remove this check? The `valueGetter.get` does an actual table lookup, which would be wasteful if we're going to skip this record anyways because the mapped key is null. Also, I'm pretty sure the lookup would throw an NPE

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
```diff
@@ -58,29 +60,46 @@ public void init(final ProcessorContext context) {
...
+            final K2 mappedKey = maybeMappedKey.get();
+            final V2 value2 = getValueOrNull(valueGetter.get(mappedKey));
             if (leftJoin || value2 != null) {
```
ableegoldman, line 78: We need to remove this too, right? We shouldn't forward anything regardless of whether it's a left join, if the mapped key is null then there's nothing to map it to

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
JoelWee: `maybeExtractMappedKey` will return `Optional.empty()` if the `mappedKey` is null. So `mappedKey` will never be null in this portion of the code and we can skip the check I think

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
JoelWee: Yea I think what you say makes sense. When I wrote this I was thinking about backwards compatibility: previously all null keys would never be passed to keyMappers but now they are. It feels like existing users might be surprised if their existing keyMappers start throwing NPEs. But it would also be suprising if the NPE a user throws get swallowed as you mentioned. 

I agree that it's better not to catch the NPE and it possibly looks like users should have always been handling the case where the key is null anyway so it should be fine?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
ableegoldman: Oh yeah, duh. Nevermind this 🙂

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
```diff
@@ -58,29 +60,46 @@ public void init(final ProcessorContext context) {
...
                 context().forward(key, joiner.apply(value, value2));
             }
         }
     }
 
+    private Optional<K2> maybeExtractMappedKey(final K1 key, final V1 value) {
+        if (value == null) {
+            return Optional.empty();
```
ableegoldman: This seems a little subtle. Can we just return the actual mapped key (or `.empty()`) in this method, and keep the explicit null check for `value` up above?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
ableegoldman: Yeah the compatibility argument is reasonable, but you could say that users should have been handling the null case all along and it was just due to a bug in Streams that we never actually passed in a null key. If people _aren't_ handling null, we should try and alert them quickly (and nothing catches people's attention faster than an NPE)

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
JoelWee: Yep agreed – was originally worried about NPEs arising from the null value as well. Done now :)

mjsax:
Retest this please.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
```diff
@@ -58,23 +58,22 @@ public void init(final ProcessorContext context) {
 
     @Override
     public void process(final K1 key, final V1 value) {
-        // we do join iff keys are equal, thus, if key is null we cannot join and just ignore the record
-        // If {@code keyMapper} returns {@code null} it implies there is no match,
-        // so ignore unless it is a left join
```
ableegoldman, line 63: It looks like we removed this comment about not ignoring it if it's a left join, but we didn't actually remove the code for that (yet). Which one is right? It seems like the comment is correct, and we shouldn't ignore the null key regardless of whether it is a left join. In that case, we should remove the `leftJoin`  part of the condition on line 78 below

JoelWee: Yea I agree the comment is right and we need to allow the case of non-null `key` with null `mappedKey` when it's a left join, since we view a null `mappedKey` as equivalent to "key not found in GlobalKTable" (except if the `key` is null, in which it should just be invalid I think?). Have made the changes now.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
```diff
@@ -58,22 +58,23 @@ public void init(final ProcessorContext context) {
 
     @Override
     public void process(final K1 key, final V1 value) {
-        // we do join iff keys are equal, thus, if key is null we cannot join and just ignore the record
-        // If {@code keyMapper} returns {@code null} it implies there is no match,
+        // We allow null keys unless {@code keyMapper} returns {@code null} and we ignore it as invalid.
```
mjsax: I guess we don't care about the original `key` any longer and only consider if `keyMapper` returns `null` or not?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
```diff
@@ -58,22 +58,23 @@ public void init(final ProcessorContext context) {
...
+        // For non-null keys, if {@code keyMapper} returns {@code null} it implies there is no match,
         // so ignore unless it is a left join
         //
         // we also ignore the record if value is null, because in a key-value data model a null-value indicates
         // an empty message (ie, there is nothing to be joined) -- this contrast SQL NULL semantics
         // furthermore, on left/outer joins 'null' in ValueJoiner#apply() indicates a missing record --
         // thus, to be consistent and to avoid ambiguous null semantics, null values are ignored
-        if (key == null || value == null) {
+        final K2 mappedKey = keyMapper.apply(key, value);
+        if ((key == null && mappedKey == null) || (!leftJoin && mappedKey == null) || value == null) {
```
mjsax: This condition seems unnecessary complex. Should it not just be:
```
if (mappedKey == null || value == null) {
```

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
```diff
@@ -58,22 +58,23 @@ public void init(final ProcessorContext context) {
...
             LOG.warn(
                 "Skipping record due to null key or value. key=[{}] value=[{}] topic=[{}] partition=[{}] offset=[{}]",
                 key, value, context().topic(), context().partition(), context().offset()
             );
             droppedRecordsSensor.record();
         } else {
-            final K2 mappedKey = keyMapper.apply(key, value);
             final V2 value2 = mappedKey == null ? null : getValueOrNull(valueGetter.get(mappedKey));
```
mjsax: At this point, we know that `mappedKey != null`, otherwise, we would have dropped the record.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
ableegoldman: Sorry, I think my original comment here was a bit ambiguous & confusingly phrased. What I meant was that the _removal_ of the comment seemed correct to me, ie we should not make any special exceptions for the left join case and should remove the `leftJoin` part of the `if (leftJoin || value2 != null) ` check down on line 79

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
JoelWee: I think that makes sense - but that will mean removing the previous logic of "`mappedKey` is null implies key not found in global table"? (Original line 62)

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
JoelWee: Right, thanks for the clarification! Wouldn't we still want the` leftJoin` in that case though? When we reach the `leftJoin` in the [code](https://github.com/apache/kafka/pull/9186/commits/e9616c64dfdc33481d0b831f80ecd0385801c761), the `mappedKey` is never null but it might not exist in the GlobalKTable (and so `value2` is null). If we're doing a `leftJoin`, then we'll want to allow these null values? (If not, the `leftJoin` is just the same as the normal `join`?)

ableegoldman: Ah yeah sorry, I'm getting things mixed up here...this comment is referring to when the mappedKey is null while the condition I cited below now only applies to when the value is null.  Your reasoning sounds correct, we should still process the record in that case if it's a left join. But we should also remove this comment, since if the mappedKey is null then we drop it, regardless of if its a left join or any other

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamKTableJoinProcessor.java
ableegoldman: Well, if `mappedKey` is null then there can't be a match in the global table since we can't do a lookup with a null key. I think what @mjsax means here (correct me if wrong) is just that we could phrase it a bit differently to say something like
```
// If the mappedKey is null, we ignore it as invalid. This should never happen for KTables 
// since keyMapper just returns the key, but for GlobalKTables a non-null key can result 
// in a null mappedKey. There can't be a match for a null mappedKey, so we drop it
```

...or something. Thoughts?

JoelWee: Yep makes sense. Have updated it now. Just noting here that this means we're changing a [test](https://github.com/apache/kafka/pull/9186/files#diff-e3715715832b244da2d8787362b0c570R230) we previously had

ableegoldman (approved):
Looks like the build failed with just the usual flaky `EosBetaUpgradeIntegrationTest`

mjsax:
Thanks for the fix @JoelWee! Merged to `trunk`.

PR #9187, opened 2020-08-16
MINOR: bump mockito to 3.5.0

lbradstreet:
3.5.0 the inline mock maker no longer uses any reflection and 3.5.0 is backwards compatible.

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -95,7 +95,7 @@ versions += [
   lz4: "1.7.1",
   mavenArtifact: "3.6.3",
   metrics: "2.2.0",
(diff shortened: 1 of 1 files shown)


lbradstreet:
I thought this might be useful for jmh benchmarks as reflection can often throw us off. I'm not sure we use the inline mock maker anywhere though.

ijuma:
ok to test

ijuma:
I don't think we use the inline mock maker.

ijuma:
Looks like 3.5.2 is the latest version https://search.maven.org/artifact/org.mockito/mockito-core/3.5.2/jar

PR #5785, opened 2018-10-11
Resize buffer on OVERFLOW

gordonmessmer:
We're seeing some BUFFER_OVERFLOW errors when using this library.  I might be mistaken, but as far as I can tell, the buffer isn't actually resized in the current implementation.  The Java documentation suggests that the buffer size should be added to the buffer's current position when resizing the buffer:

https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html

PR #9188, opened 2020-08-17
break when dst is full so that unwrap isn't called when appreadbuffer…

Spatterjaaay:
… may have data

There are a couple of different situations which can result in
BUFFER_OVERFLOW on read with the current implementation, due to the
while loop structure (such as TLS compression with identical buffer
sizes, or buffers sizes that differ to optimize modes where the cipher
text is larger than the plain text.)  The JDK documentation indicates
that a buffer of getApplicationBufferSize() bytes will be enough for a
single unwrap operation, but the SslTransportLayer loop may call unwrap
with an application buffer which isn't empty.

The current implementation will check dst for space and then move data
from the application buffer.  It will then continue the loop and may try
to unwrap() again without verifying that there are
getApplicationBufferSize() bytes free in the application buffer. If,
instead, the loop moves data into dst, and then breaks the loop if dst
is full, then unwrap() should never be called with data in the
application buffer.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/network/SslTransportLayer.java
+++ b/clients/src/main/java/org/apache/kafka/common/network/SslTransportLayer.java
@@ -599,11 +599,11 @@ public class SslTransportLayer implements TransportLayer {
                     }
 
                     // appReadBuffer will extended upto currentApplicationBufferSize
-                    // we need to read the existing content into dst before we can do unwrap again. If there are no space in dst
-                    // we can break here.
-                    if (dst.hasRemaining())
-                        read += readFromAppBuffer(dst);
-                    else
+                    // we need to read the existing content into dst before we can do unwrap again.
+                    read += readFromAppBuffer(dst);
+                    // If there are no space in dst we can break here. Any data left in appReadBuffer
+                    // or netReadBuffer will be processed on the next call to the read method.
+                    if (!dst.hasRemaining())
                         break;
                 } else if (unwrapResult.getStatus() == Status.BUFFER_UNDERFLOW) {
                     int currentNetReadBufferSize = netReadBufferSize();


vineethgn:
Based on the discussion in https://github.com/apache/kafka/pull/5785/, i tried the fix in this PR, it didnt work for me. 

I tried to reproduce the issue and i can reproduce the issue consistently in the following conditions.

1. [Conscrypt](https://github.com/google/conscrypt) as ssl provider 
2. openjdk-11-jdk

I checked by adding additional logs and could find that the packet buffer size and app buffer returned by Conscrypt ssl engine mismatches. Below are the logs that i got

SSl engine class org.conscrypt.Java8EngineWrapper (org.apache.kafka.common.network.SslTransportLayer)
SSl engine session class org.conscrypt.Java8ExtendedSSLSession (org.apache.kafka.common.network.SslTransportLayer)
SSl engine session App Buffer Size 16384 (org.apache.kafka.common.network.SslTransportLayer)
SSl engine session Net Buffer Size 16709 (org.apache.kafka.common.network.SslTransportLayer) 

Since this was a blocker, i modified  the Ssltransportlayer to have appreadbuffer atleast the same size as netreadbuffer, before call to unwrap. Is it ok to have appReadBuffer atleast the same size as netReadBuffer ?

so far  the clusters are running smoothly without buffer overflow or underflow issue. 

below is the code fix that i made that seemed to be working fine so far without further issues. 

`while (netReadBuffer.position() > 0) {
             netReadBuffer.flip();
             SSLEngineResult unwrapResult;
             try {
                 appReadBuffer = Utils.ensureCapacity(appReadBuffer, netReadBufferSize());
                 unwrapResult = sslEngine.unwrap(netReadBuffer, appReadBuffer);`

PR #9191, opened 2020-08-17
KAFKA-10355: Throw error when source topic was deleted

cadonna:
Before this commit, Kafka Streams would gracefully shut down the whole application when a source topic is deleted. The graceful shutdown does not give the user the possibility to react on the deletion of the source topic in the uncaught exception handler.

This commit changes this behavior and throws an error when a source topic is deleted.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/errors/MissingSourceTopicException.java
+++ b/streams/src/main/java/org/apache/kafka/streams/errors/MissingSourceTopicException.java
@@ -1,0 +1,26 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.errors;
+
+public class MissingSourceTopicException extends StreamsException {
+
+    private final static long serialVersionUID = 1L;
+
+    public MissingSourceTopicException(final String message) {
+        super(message);
+    }
+}
--- a/streams/src/test/java/org/apache/kafka/streams/processor/internals/HandlingSourceTopicDeletionTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/processor/internals/HandlingSourceTopicDeletionTest.java
@@ -67,8 +67,7 @@ public class HandlingSourceTopicDeletionTest {
 
     @Before
     public void before() throws InterruptedException {
-        CLUSTER.createTopic(INPUT_TOPIC, 2, 1);
-        CLUSTER.createTopic(OUTPUT_TOPIC, 2, 1);
+        CLUSTER.createTopics(INPUT_TOPIC, OUTPUT_TOPIC);
 
         final String safeTestName = safeUniqueTestName(getClass(), testName);
         final String appId = "app-" + safeTestName;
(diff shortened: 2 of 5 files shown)


cadonna:
FYI: @guozhangwang @ableegoldman @vvcephei @mjsax @abbccdda

streams/src/main/java/org/apache/kafka/streams/errors/MissingSourceTopicException.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.errors;
+
+public class MissingSourceTopicException extends StreamsException {
+
+    private final static long serialVersionUID = 1L;
```
ableegoldman, line 21: What's this for (or going to be for)?

streams/src/main/java/org/apache/kafka/streams/errors/MissingSourceTopicException.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.errors;
+
+public class MissingSourceTopicException extends StreamsException {
```
ableegoldman, line 19: I'm not sure this should extend StreamsException, my understanding is that it's generally reserved for Streams internal errors and not user code/setup issues

ableegoldman (commented):
Some minor questions but overall I like this plan. I actually always thought it was odd that we just shutdown silently instead of throwing an exception here

streams/src/main/java/org/apache/kafka/streams/errors/MissingSourceTopicException.java
cadonna: That is a good question. I have to admit that I blindly copied it from another exception class. The field has to do with exceptions implementing the `Serializable` interface. This field tells the JVM whether the serialized object can be deserialized into an object of the class that it is available in the JVM. See https://stackoverflow.com/questions/7187302/what-is-serialversionuid-in-java-normally-in-exception-class

streams/src/main/java/org/apache/kafka/streams/errors/MissingSourceTopicException.java
cadonna: That is a good point. I thought all exceptions thrown from inside Streams (except for the `IllegalStateException`) should be `StreamsException`s. The `RecordDeserializer` throws a `StreamsException` when the deserialization exception handler -- which is user code -- throws any exception. Maybe @guozhangwang can help here.

guozhangwang: I agree with @cadonna here, that all exceptions from inside Streams should be inheriting from `StreamsException`, and that may include 1) environmental issues like timeout --- note that after @mjsax KIP we would not throw TimeoutException (which is a KafkaException not StreamsException) but a new exception inherited from StreamsException, and IOException which are also wrapped as StateStoreException inherited from StreamsException, and 2) user bug in `process` etc which get caught by streams code.

guozhangwang (commented):
My take from @cadonna 's proposal is that moving forward we will let streams to always throw StreamsException inherited exceptions, and different class type could indicate different origin / type / possible-impact (data corrupted, data lost, for example) combos. In that case, this introduced `MissingSourceTopicException extends StreamsException` makes sense to me as it aligns with the general framework.

ableegoldman:
Tests didn't run, possible checkstyle issue? Results already cleaned up unfortunately

cadonna:
@ableegoldman Yes, you are right there were checkstyle issues. Will fix them.

cadonna:
This is the implementation for KIP-662.

cadonna:
Test this please

streams/src/test/java/org/apache/kafka/streams/processor/internals/HandlingSourceTopicDeletionTest.java
```diff
@@ -0,0 +1,109 @@
...
+            () -> "Kafka Streams application did not reach state RUNNING"
+        );
+
+        CLUSTER.deleteTopicAndWait(INPUT_TOPIC);
+
+        TestUtils.waitForCondition(
+            () -> kafkaStreams.state() == State.ERROR,
+            TIMEOUT,
+            () -> "Kafka Streams application did not reach state ERROR"
+        );
```
vvcephei: What do you think about expanding the test to verify that all members really get shut down, not just all the ones in an instance? It seems simple enough: just create and start two instances and then verify they both wind up in ERROR state.

WDYT?

vvcephei (commented):
Hey @cadonna , 

This looks good to me. I had one suggestion about expanding the test a little, but I'm really happy either way. Thanks!

streams/src/test/java/org/apache/kafka/streams/processor/internals/HandlingSourceTopicDeletionTest.java
cadonna: I had that thought, too. But then I thought it is not required to test the code added in this PR. On the other hand, it does not harm to use two Streams clients and broaden the test scope since this is an integration test.

cadonna:
I added unit tests.

vvcephei (approved):
Thanks, @cadonna !

PR #9199, opened 2020-08-18
KAFKA-10418: alter topic configs via kafka-topics error text

rondagostino:
Changing a topic config with the `kafka-topics` command while connecting to Kafka via `--bootstrap-server` (rather than connecting to ZooKeeper via `--zookeeper`) is not supported. The desired functionality is available elsewhere, though: it is possible to change a topic config while connecting to Kafka rather than ZooKeeper via the `kafka-configs` command instead. However, neither the `kafka-topics` error message received nor the `kafka-topics` help information itself indicates this other possibility. For example:

```
$ kafka-topics.sh --bootstrap-server localhost:9092 --alter --topic test --config flush.messages=12345
Option combination "[bootstrap-server],[config]" can't be used with option "[alter]"
```

```
$ kafka-topics.sh
...
--config <String: name=value> A topic configuration override for the topic being created or altered...It is supported only in combination with – create if --bootstrap-server option is used.
```

Rather than simply saying that what you want to do isn't available, it would be better to say also that you can do it with the `kafka-configs` command.

### DIFF
--- a/core/src/main/scala/kafka/admin/TopicCommand.scala
+++ b/core/src/main/scala/kafka/admin/TopicCommand.scala
@@ -617,10 +617,13 @@ object TopicCommand extends Logging {
                          .describedAs("topic")
                          .ofType(classOf[String])
     private val nl = System.getProperty("line.separator")
-    private val configOpt = parser.accepts("config", "A topic configuration override for the topic being created or altered."  +
-                                             "The following is a list of valid configurations: " + nl + LogConfig.configNames.map("\t" + _).mkString(nl) + nl +
+    private val kafkaConfigsCanAlterTopicConfigsViaBootstrapServer =
+      " (the kafka-configs CLI supports altering topic configs with a --bootstrap-server option)"
+    private val configOpt = parser.accepts("config", "A topic configuration override for the topic being created or altered." +
+                                             " The following is a list of valid configurations: " + nl + LogConfig.configNames.map("\t" + _).mkString(nl) + nl +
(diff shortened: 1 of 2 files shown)


rondagostino:
@cmccabe Can you take a look?

cmccabe:
LGTM

rondagostino:
Note that KIP-377 landed in 2.2 and explicitly stated in the new code that altering topic configurations via `kafka-topics` was deprecated – even using the `--zookeeper` flag (setting topic configs using `kafka-topics` when using `--create` to create a topic is supported and works with --bootstrap-server; it’s just altering/deleting topic configs that was deprecated via `--zookeeper` and that never worked/never will work via `--bootstrap-server`).  Therefore, potentially, this could be backported as far back as 2.2.x.

rondagostino:
JDK 11 build failure was a n infrastructure issue.  JDK 8 test failure is an unrelated flaky test (`org.apache.kafka.streams.integration.EosIntegrationTest.shouldNotViolateEosIfOneTaskFailsWithState[exactly_once]`).  JDK 14 build was clean.

rondagostino:
Actually, it turns out that the ability to use `kafka-configs --bootstrap-server` was only added in 2.5.  So this could be backported to 2.5 and 2.6 (not all the way back to 2.2 as originally stated).

cmccabe:
Thanks, @rondagostino.

I think it's OK to just have in trunk for now.  We already did a 2.6 release and this is not a critical fix, just a help text change.  But if someone feels strongly I will backport.

PR #9200, opened 2020-08-19
MINOR: mirror integration tests should not call System.exit

lbradstreet:
406635bcc9f2a4c439d198ea0549170de331323c switched System.exit to using Exit.exit, however the integration tests did not choose to override the exit procedure.

This can cause invalid test runs with errors like this:
```
[2020-08-18T15:52:01.142Z] * What went wrong:
[2020-08-18T15:52:01.142Z] Execution failed for task ':connect:mirror:integrationTest'.
[2020-08-18T15:52:01.142Z] > Process 'Gradle Test Executor 192' finished with non-zero exit value 1
[2020-08-18T15:52:01.142Z]   This problem might be caused by incorrect test process configuration.
[2020-08-18T15:52:01.142Z]   Please refer to the test execution section in the User Manual at https://docs.gradle.org/6.6/userguide/java_testing.html#sec:test_execution
```


connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
```diff
@@ -185,10 +192,9 @@ public void close() {
         deleteAllTopics(backup.kafka());
         primary.stop();
         backup.stop();
+        assertFalse(exited.get());
```
rondagostino:
```suggestion
        assertFalse(exited.get());
        Exit.resetExitProcedure();
```

core/src/test/scala/integration/kafka/tools/MirrorMakerIntegrationTest.scala
```diff
@@ -29,14 +29,31 @@ import org.apache.kafka.clients.producer.{ProducerConfig, ProducerRecord}
...
+  @Before
+  override def setUp(): Unit = {
+    Exit.setExitProcedure((_, _) => exited.set(true))
+    super.setUp()
+  }
+
+  @After
+  override def tearDown(): Unit = {
+    super.tearDown()
+    assertFalse(exited.get())
```
rondagostino:
```suggestion
    assertFalse(exited.get())
    Exit.resetExitProcedure()
```

rondagostino (commented):
Thanks for the PR.  I believe we need to reset the exit procedure after the test.  Aside from that -- LGTM.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
lbradstreet: Ah, fancy GitHub stuff. I pushed up the fix independently.

core/src/test/scala/integration/kafka/tools/MirrorMakerIntegrationTest.scala
lbradstreet: Done

omkreddy:
ok to test

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
```diff
@@ -185,10 +192,10 @@ public void close() {
...
+        Exit.resetExitProcedure();
```
rajinisivaram: I am guessing we ran into an issue because of exit, so perhaps this reset ought to be in a finally block?

lbradstreet: @rajinisivaram good point. I'll fix that.

lbradstreet: I've pushed a fix. Thanks.

ijuma:
Ok to test

rajinisivaram:
ok to test

rajinisivaram (approved):
@lbradstreet Thanks for the update, LGTM

rajinisivaram:
@lbradstreet Thanks for the PR, merging to trunk.

C0urante:
@lbradstreet I think the changes to the MM2 tests may have been unnecessary or, at least, inadequate. Embedded Connect clusters automatically [mask exit procedures on startup](https://github.com/apache/kafka/blob/aeeb7b2f9a9abe8f49543a2278757722e5974cb3/connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedConnectCluster.java#L128-L138) by default, which should apply to the MM2 integration tests.

I believe the issue with the MM2 integration tests is that embedded Connect clusters also automatically [reset exit procedures on shutdown](https://github.com/apache/kafka/blob/aeeb7b2f9a9abe8f49543a2278757722e5974cb3/connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedConnectCluster.java#L156-L157). Since the tests use two embedded Connect clusters which are shut down sequentially at the end of each test case, the second cluster to be shut down does so without using the masked exit procedures, which can result in `System::exit` being invoked.

PR #9205, opened 2020-08-21
MINOR: Fix typo in example

jolshan:
The ducker-ak test example includes a file path that does not exist. I fixed to be the correct path.


omkreddy (approved):
@jolshan Thanks for the PR. LGTM.

PR #9211, opened 2020-08-24
KAFKA-10426: Deadlock on session key update.

xakassi:
DistributedHerder goes to updateConfigsWithIncrementalCooperative() synchronized method and called configBackingStore.snapshot() which take a lock on internal object in KafkaConfigBackingStore class.

Meanwhile KafkaConfigBackingStore in ConsumeCallback inside synchronized block on internal object gets SESSION_KEY record and calls updateListener.onSessionKeyUpdate() which take a lock on DistributedHerder.

So, we have a Deadlock.

To avoid this, updateListener with new session key should be called outside synchronized block as it's done, for example, for updateListener.onTaskConfigUpdate(updatedTasks).

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/storage/KafkaConfigBackingStore.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/storage/KafkaConfigBackingStore.java
@@ -699,46 +699,42 @@ public class KafkaConfigBackingStore implements ConfigBackingStore {
                 if (started)
                     updateListener.onTaskConfigUpdate(updatedTasks);
             } else if (record.key().equals(SESSION_KEY_KEY)) {
-                SessionKey newSessionKey;
-                synchronized (lock) {
-                    if (value.value() == null) {
-                        log.error("Ignoring session key because it is unexpectedly null");
-                        return;
-                    }
-                    if (!(value.value() instanceof Map)) {
-                        log.error("Ignoring session key because the value is not a Map but is {}", value.value().getClass());
-                        return;
-                    }
-
-                    Map<String, Object> valueAsMap = (Map<String, Object>) value.value();
-
-                    Object sessionKey = valueAsMap.get("key");
-                    if (!(sessionKey instanceof String)) {
-                        log.error("Invalid data for session key 'key' field should be a String but is {}", sessionKey.getClass());
-                        return;
-                    }
-                    byte[] key = Base64.getDecoder().decode((String) sessionKey);
-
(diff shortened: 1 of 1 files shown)


connect/runtime/src/main/java/org/apache/kafka/connect/storage/KafkaConfigBackingStore.java
```diff
@@ -733,10 +734,11 @@ public void onCompletion(Throwable error, ConsumerRecord<String, byte[]> record)
                         new SecretKeySpec(key, (String) keyAlgorithm),
                         (long) creationTimestamp
                     );
-
-                    if (started)
-                        updateListener.onSessionKeyUpdate(KafkaConfigBackingStore.this.sessionKey);
```
kkonstantine, line 738: `KafkaConfigBackingStore.this.sessionKey` is already a `volatile` member variable. 

So, given that no other common data structures are altered in this block, I'm suspecting that your change means that we might as well get rid off the `synchronized` block altogether in this `else if` branch here. 

I'd like to give it a second thought in the morning but lmk what you think.

kkonstantine (commented):
Thanks @xakassi 
One question inline.

kkonstantine:
ok to test

connect/runtime/src/main/java/org/apache/kafka/connect/storage/KafkaConfigBackingStore.java
xakassi: Hi, @kkonstantine !
Yes, indeed, it looks like we can get rid of the synchronized block here at all. I think it's safe.
I commited this change.

xakassi:
Hi, guys! Hi, @kkonstantine !
This is my first PR to Kafka and I would like to clarify if there is anything else I need to do? Or should I just wait and you will take care of this PR?

kkonstantine:
Hi @xakassi 
Thanks for addressing the comments. We had some issues with the PR builder at the time, so the tests didn't run. 
I'll try to figure out how I could run the tests for this PR and if there's no easy way I'll ask you to submit a new one. 
Thanks for your patience.

xakassi:
Hi @kkonstantine !
Any news about tests? Should I recreate a PR?

kkonstantine:
Yes, let's open a new one @xakassi 
I haven't found an easy way to reconnect an old PR with the build job. 
Please link to this PR for the comments from the new one. Thanks.

xakassi:
I have closed this PR and submitted a new one: https://github.com/apache/kafka/pull/9431
@kkonstantine Please, take a look at it. Thank you!

PR #9213, opened 2020-08-24
MINOR: add epoch lineage checks to system tests

lbradstreet:
This adds assertions to check that leader epoch lineages match between
replicas. These have been added to system tests that involve broker
restarts and that wait for replicas to rejoin the ISR by the end of the
test.

I also moved wait_until_rejoin_isr from downgrade_test and upgrade_test
 as the implementation was the same in each.


lbradstreet:
This PR was able to help find KAFKA-10432: https://github.com/apache/kafka/pull/9219

PR #9216, opened 2020-08-24
KAFKA-10428: Fix schema for header conversion in MirrorSourceTask.

jthompson6:
The addBytes method adds the header using Schema.BYTES, which results in base64 encoding when the record is stored.  SimpleHeaderConverter#toConnectHeader implements schema inference which we can use here, as the ConsumerRecord does not have header schema information.

Testing: I added schema verification to the existing MirrorSourceTaskTest#testSerde, with a string and int example.

### DIFF
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorSourceTask.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorSourceTask.java
@@ -117,14 +117,14 @@ public class MirrorSourceTask extends SourceTask {
         try {
             consumerAccess.acquire();
         } catch (InterruptedException e) {
-            log.warn("Interrupted waiting for access to consumer. Will try closing anyway.");
+            log.warn("Interrupted waiting for access to consumer. Will try closing anyway."); 
         }
         Utils.closeQuietly(consumer, "source consumer");
         Utils.closeQuietly(offsetProducer, "offset producer");
         Utils.closeQuietly(metrics, "metrics");
         log.info("Stopping {} took {} ms.", Thread.currentThread().getName(), System.currentTimeMillis() - start);
     }
-
+   
     @Override
     public String version() {
         return "1";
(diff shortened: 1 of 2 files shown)


jthompson6:
@ryannedolan are you the right person to tag here?

ryannedolan:
Thanks for finding this. Would it make sense to implement a new nop converter for this purpose? Not sure whether a KIP would be required -- maybe it can be private to the mirror package?

jthompson6:
I see the header converter class is configurable in https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerConfig.java#L90, but it appears to be a whole cluster setting, when really we want to only set it for this connector. 

Do you think it is a bug that SimpleHeaderConvertor#fromConnectHeader transforms a byte[] header, even is the schema is BYTES? Maybe that is a question for @rhauch ?

rhauch (commented):
I think the actual root cause here is pretty simple: the "deserialization" logic for the headers just uses the raw byte array header values from the `ConsumerRecord` (e.g., no conversion), while the `MirrorSourceConnector` is likely inheriting the Connect worker's default `SimpleHeaderConverter` (e.g., "simple" conversion) because the connector configuration does not define `header.converter`.

**The bottom line: the same header converter should be used on the deserialization and serialization paths.**

I would correct this by always configuring the `MirrorSourceConnector` with:
```
header.converter=org.apache.kafka.connect.converters.ByteArrayConverter
```
This tells Connect to use the `ByteArrayConverter`, which also implements `HeaderConverter` and which is effectively the no-op that @ryannedolan asked about. 

Doing that actually requires no code changes -- unless MirrorMaker2 code should be updated to always use `header.converter=org.apache.kafka.connect.converters.ByteArrayConverter` when it configures the connector.

I hope that helps.

jthompson6:
Sorry for the delay, I just tested adding `"header.converter": "org.apache.kafka.connect.converters.ByteArrayConverter"` to the connector config and it works.

jthompson6:
Should I make an update to the readme @ryannedolan?
https://github.com/apache/kafka/blob/trunk/connect/mirror/README.md doesn't have a section for running MirrorSourceConnector in a general connect cluster, so should I add one?

ryannedolan:
> Should I make an update to the readme @ryannedolan?
> https://github.com/apache/kafka/blob/trunk/connect/mirror/README.md doesn't have a section for running MirrorSourceConnector in a general connect cluster, so should I add one?

That'd be greatly appreciated!

jthompson6:
Cool. Closing this, will make a new PR.

PR #9219, opened 2020-08-25
KAFKA-10432: LeaderEpochCache is incorrectly recovered for leader epoch 0

lbradstreet:
The leader epoch cache is incorrectly recovered for epoch 0 as the
assignment is skipped when epoch == 0. This check was likely intended to
prevent negative epochs from being applied or there was an assumption
that epochs started at 1.

A test has been added to LogSegmentTest to show the LogSegment
recovery path works for the epoch cache. This was a test gap as none of the 
recover calls supply a leader epoch cache to recover.


core/src/test/scala/unit/kafka/log/LogSegmentTest.scala
```diff
@@ -367,6 +371,45 @@ class LogSegmentTest {
...
+    seg.append(largestOffset = 109L, largestTimestamp = RecordBatch.NO_TIMESTAMP,
+      shallowOffsetOfMaxTimestamp = 108L, MemoryRecords.withRecords(108L, CompressionType.NONE, 1,
+        new SimpleRecord("a".getBytes), new SimpleRecord("b".getBytes)))
+
+    seg.append(largestOffset = 111L, largestTimestamp = RecordBatch.NO_TIMESTAMP,
+      shallowOffsetOfMaxTimestamp = 110, MemoryRecords.withRecords(110L, CompressionType.NONE, 2,
+        new SimpleRecord("a".getBytes), new SimpleRecord("b".getBytes)))
+
+    seg.recover(new ProducerStateManager(topicPartition, logDir), Some(cache))
+    assertEquals(ArrayBuffer(EpochEntry(0, 104L), EpochEntry(epoch=1, startOffset=106), EpochEntry(epoch=2, startOffset=110)), cache.epochEntries)
```
hachikuji: nit: use arg names consistently?

hachikuji (approved):
LGTM. Just left a minor nitpick.

core/src/test/scala/unit/kafka/log/LogSegmentTest.scala
lbradstreet: Fixed

PR #9220, opened 2020-08-25
KAFKA-10433 Reuse the ByteBuffer in validating compressed records

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-10433

It is hot method so reusing the ByteBuffer can reduce a bunch of memory usage if the compression type supports BufferSupplier.

**experiment**
- duration: 1 minute
- compression: LZ4
- workload: 10 producers -> 1 broker 
- memory usage of byte array allocation: 4.52 GB -> 0.68 GB

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/record/BufferSupplier.java
+++ b/clients/src/main/java/org/apache/kafka/common/record/BufferSupplier.java
@@ -22,6 +22,9 @@ import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.ConcurrentMap;
 
 /**
  * Simple non-threadsafe interface for caching byte buffers. This is suitable for simple cases like ensuring that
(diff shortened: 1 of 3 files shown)


ijuma:
@chia7712 Thanks for the PR. The intent is good, but I think the approach should be a bit different. As it happens, I have implemented this other approach. Would you be OK if I submit that as a PR and we can compare?

chia7712:
> Would you be OK if I submit that as a PR and we can compare?

Please feel free to submit another PR. We all love to see the better solution :)

ijuma:
@chia7712 here's what I had in mind:
https://github.com/apache/kafka/pull/9229

What do you think?

chia7712:
close as there is a better approach (#9229)

junrao:
@chia7712 : We filed [KAFKA-15674](https://issues.apache.org/jira/browse/KAFKA-15674) to revisit this. The current implementation (https://github.com/apache/kafka/pull/9229) passes around a thread unsafe data structure to multiple places such as the ReplicaManager, GroupCoordinator, TxnCoordinator, etc. Each of those places has the logic to add callbacks. If one is not careful (as in [KAFKA-15674](https://issues.apache.org/jira/browse/KAFKA-15674)), the thread unsafe data structure could be included in those callbacks, which will lead to subtle concurrency issues.

It's probably safer to replace the thread unsafe data structure with the thread safe one as you had in this PR. Would you be interested in resurrecting this PR? If there is no significant performance impact, we could just take the approach in this PR.

cc @ijuma

PR #9221, opened 2020-08-26
KAFKA-10436: Implement KIP-478 Topology changes

vvcephei:
Converts `Topology#addProcessor` and `#addGlobalStore`
Also, convert some of the internals in support of `addProcessor`

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9222, opened 2020-08-26
KAFKA-10437: Implement test-utils and StateStore changes for KIP-478

vvcephei:
Propose a new init method for StateStore so that it works with the new ProcessorContext.
Convert the test-utils MockProcessorContext to the new API.

### DIFF
--- a/streams/examples/src/test/java/org/apache/kafka/streams/examples/wordcount/WordCountProcessorTest.java
+++ b/streams/examples/src/test/java/org/apache/kafka/streams/examples/wordcount/WordCountProcessorTest.java
@@ -18,8 +18,8 @@ package org.apache.kafka.streams.examples.wordcount;
 
 import org.apache.kafka.common.serialization.Serdes;
 import org.apache.kafka.streams.KeyValue;
-import org.apache.kafka.streams.processor.MockProcessorContext;
-import org.apache.kafka.streams.processor.Processor;
+import org.apache.kafka.streams.processor.api.MockProcessorContext;
+import org.apache.kafka.streams.processor.api.Processor;
 import org.apache.kafka.streams.state.KeyValueStore;
 import org.apache.kafka.streams.state.Stores;
 import org.junit.Test;
@@ -36,7 +36,7 @@ import static org.junit.Assert.assertTrue;
 public class WordCountProcessorTest {
     @Test
     public void test() {
-        final MockProcessorContext context = new MockProcessorContext();
+        final MockProcessorContext<String, String> context = new MockProcessorContext<>();
 
         // Create, initialize, and register the state store.
         final KeyValueStore<String, Integer> store =
@@ -44,11 +44,10 @@ public class WordCountProcessorTest {
                 .withLoggingDisabled() // Changelog is not supported by MockProcessorContext.
                 // Caching is disabled by default, but FYI: caching is also not supported by MockProcessorContext.
                 .build();
-        store.init(context, store);
-        context.register(store, null);
+        store.init(context.getStateStoreContext(), store);
 
         // Create and initialize the processor under test
-        final Processor<String, String> processor = new WordCountProcessorDemo.MyProcessorSupplier().get();
+        final Processor<String, String, String, String> processor = new WordCountProcessorDemo.MyProcessorSupplier().get();
         processor.init(context);
 
         // send a record to the processor
@@ -61,7 +60,7 @@ public class WordCountProcessorTest {
         context.scheduledPunctuators().get(0).getPunctuator().punctuate(0L);
 
         // finally, we can verify the output.
-        final Iterator<MockProcessorContext.CapturedForward> capturedForwards = context.forwarded().iterator();
+        final Iterator<MockProcessorContext.CapturedForward<String, String>> capturedForwards = context.forwarded().iterator();
         assertEquals(new KeyValue<>("alpha", "2"), capturedForwards.next().keyValue());
         assertEquals(new KeyValue<>("beta", "1"), capturedForwards.next().keyValue());
         assertEquals(new KeyValue<>("gamma", "1"), capturedForwards.next().keyValue());
--- a/streams/src/main/java/org/apache/kafka/streams/processor/StateStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/StateStore.java
@@ -72,7 +72,7 @@ public interface StateStore {
      * Initializes this state store.
      * <p>
      * The implementation of this function must register the root store in the context via the
-     * {@link ProcessorContext#register(StateStore, StateRestoreCallback)} function, where the
+     * {@link StateStoreContext#register(StateStore, StateRestoreCallback)} function, where the
      * first {@link StateStore} parameter should always be the passed-in {@code root} object, and
      * the second parameter should be an object of user's implementation
      * of the {@link StateRestoreCallback} interface used for restoring the state store from the changelog.
--- a/streams/src/main/java/org/apache/kafka/streams/processor/StateStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/StateStore.java
@@ -17,6 +17,8 @@
 package org.apache.kafka.streams.processor;
 
 import org.apache.kafka.streams.errors.StreamsException;
+import org.apache.kafka.streams.processor.api.ProcessorContext;
+import org.apache.kafka.streams.processor.internals.StoreToProcessorContextAdapter;
 
 /**
  * A storage engine for managing state maintained by a stream processor.
@@ -25,7 +27,7 @@ import org.apache.kafka.streams.errors.StreamsException;
  * all data into this store directory.
  * The store directory must be created with the state directory.
  * The state directory can be obtained via {@link ProcessorContext#stateDir() #stateDir()} using the
- * {@link ProcessorContext} provided via {@link #init(ProcessorContext, StateStore) init(...)}.
+ * {@link ProcessorContext} provided via {@link #init(StateStoreContext, StateStore) init(...)}.
  * <p>
  * Using nested store directories within the state directory isolates different state stores.
  * If a state store would write into the state directory directly, it might conflict with others state stores and thus,
@@ -49,6 +51,27 @@ public interface StateStore {
      * Initializes this state store.
      * <p>
      * The implementation of this function must register the root store in the context via the
+     * {@link org.apache.kafka.streams.processor.ProcessorContext#register(StateStore, StateRestoreCallback)} function,
+     * where the first {@link StateStore} parameter should always be the passed-in {@code root} object, and
+     * the second parameter should be an object of user's implementation
+     * of the {@link StateRestoreCallback} interface used for restoring the state store from the changelog.
+     * <p>
+     * Note that if the state store engine itself supports bulk writes, users can implement another
+     * interface {@link BatchingStateRestoreCallback} which extends {@link StateRestoreCallback} to
+     * let users implement bulk-load restoration logic instead of restoring one record at a time.
+     * <p>
+     * This method is not called if {@link StateStore#init(StateStoreContext, StateStore)}
+     * is implemented.
+     *
+     * @throws IllegalStateException If store gets registered after initialized is already finished
+     * @throws StreamsException if the store's change log does not contain the partition
+     */
+    void init(org.apache.kafka.streams.processor.ProcessorContext context, StateStore root);
+
+    /**
+     * Initializes this state store.
+     * <p>
+     * The implementation of this function must register the root store in the context via the
      * {@link ProcessorContext#register(StateStore, StateRestoreCallback)} function, where the
      * first {@link StateStore} parameter should always be the passed-in {@code root} object, and
      * the second parameter should be an object of user's implementation
@@ -61,7 +84,9 @@ public interface StateStore {
      * @throws IllegalStateException If store gets registered after initialized is already finished
      * @throws StreamsException if the store's change log does not contain the partition
      */
-    void init(ProcessorContext context, StateStore root);
+    default void init(final StateStoreContext context, final StateStore root) {
+        init(StoreToProcessorContextAdapter.adapt(context), root);
+    }
 
     /**
      * Flush any cached data
--- a/streams/src/main/java/org/apache/kafka/streams/processor/StateStoreContext.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/StateStoreContext.java
@@ -24,7 +24,7 @@ import java.io.File;
 import java.util.Map;
 
 /**
- * Processor context interface.
+ * State store context interface.
  */
 public interface StateStoreContext {
 
@@ -86,7 +86,7 @@ public interface StateStoreContext {
      * Returns all the application config properties as key/value pairs.
      *
      * <p> The config properties are defined in the {@link org.apache.kafka.streams.StreamsConfig}
-     * object and associated to the ProcessorContext.
+     * object and associated to the StateStoreContext.
      *
      * <p> The type of the values is dependent on the {@link org.apache.kafka.common.config.ConfigDef.Type type} of the property
      * (e.g. the value of {@link org.apache.kafka.streams.StreamsConfig#DEFAULT_KEY_SERDE_CLASS_CONFIG DEFAULT_KEY_SERDE_CLASS_CONFIG}
@@ -102,7 +102,7 @@ public interface StateStoreContext {
      * stripping the prefix.
      *
      * <p> The config properties are defined in the {@link org.apache.kafka.streams.StreamsConfig}
-     * object and associated to the ProcessorContext.
+     * object and associated to the StateStoreContext.
      *
      * @param prefix the properties prefix
      * @return the key/values matching the given prefix from the StreamsConfig properties.
--- a/streams/src/main/java/org/apache/kafka/streams/processor/StateStoreContext.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/StateStoreContext.java
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.processor;
+
+import org.apache.kafka.common.serialization.Serde;
+import org.apache.kafka.streams.StreamsMetrics;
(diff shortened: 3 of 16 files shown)


streams/examples/src/test/java/org/apache/kafka/streams/examples/wordcount/WordCountProcessorTest.java
```diff
@@ -36,7 +36,7 @@
 public class WordCountProcessorTest {
     @Test
     public void test() {
-        final MockProcessorContext context = new MockProcessorContext();
+        final MockProcessorContext<String, String> context = new MockProcessorContext<>();
```
vvcephei, line 39: Switching over to the mock of the new API.

streams/examples/src/test/java/org/apache/kafka/streams/examples/wordcount/WordCountProcessorTest.java
```diff
@@ -61,7 +61,7 @@ public void test() {
         context.scheduledPunctuators().get(0).getPunctuator().punctuate(0L);
 
         // finally, we can verify the output.
-        final Iterator<MockProcessorContext.CapturedForward> capturedForwards = context.forwarded().iterator();
+        final Iterator<MockProcessorContext.CapturedForward<String, String>> capturedForwards = context.forwarded().iterator();
```
vvcephei, line 63: Since the new ProcessorContext bounds the forward calls, I've also added types to the captured forwards.

streams/src/main/java/org/apache/kafka/streams/processor/StateStore.java
```diff
@@ -45,6 +47,27 @@
...
+     * interface {@link BatchingStateRestoreCallback} which extends {@link StateRestoreCallback} to
+     * let users implement bulk-load restoration logic instead of restoring one record at a time.
+     * <p>
+     * This method is not called if {@link StateStore#init(ProcessorContext, org.apache.kafka.streams.processor.StateStore)}
+     * is implemented.
+     *
+     * @throws IllegalStateException If store gets registered after initialized is already finished
+     * @throws StreamsException if the store's change log does not contain the partition
+     */
+    void init(org.apache.kafka.streams.processor.ProcessorContext context, StateStore root);
```
vvcephei: Not sure why the diff happened this way, but this is the original `init` method. I've just qualified ProcessorContext and added this note to the javadoc: `This method is not called if {@link StateStore#init(ProcessorContext, StateStore)} is implemented.`

Later on, when we deprecate `processor.ProcessorContext`, this method will also become deprecated, pushing store implementers to move to the new API.

streams/src/main/java/org/apache/kafka/streams/processor/StateStore.java
```diff
@@ -61,7 +84,14 @@
...
      */
-    void init(ProcessorContext context, StateStore root);
+    default void init(final ProcessorContext<?, ?> context, final StateStore root) {
+        final org.apache.kafka.streams.processor.ProcessorContext adapted =
+            ProcessorContextReverseAdapter.adapt(
+                context,
+                new ProcessorContextReverseAdapter.UnsupportedDeprecatedForwarder()
+            );
+        init(adapted, root);
+    }
```
vvcephei: We have to add this so that we can pass in the new ProcessorContext. The default implementation delegates to the old `init` method so that existing store implementations will function with no changes.

If the only callers were internal, we could just adapt at the call site. Unfortunately, users can also call `StateStore#init`, and they would do it if they have their own store implementations or if they use `MockProcessorContext` to test a stateful processor.

streams/src/main/java/org/apache/kafka/streams/processor/StateStore.java
```diff
@@ -45,6 +47,27 @@
...
+     * interface {@link BatchingStateRestoreCallback} which extends {@link StateRestoreCallback} to
+     * let users implement bulk-load restoration logic instead of restoring one record at a time.
+     * <p>
+     * This method is not called if {@link StateStore#init(ProcessorContext, StateStore)}
+     * is implemented.
+     *
+     * @throws IllegalStateException If store gets registered after initialized is already finished
+     * @throws StreamsException if the store's change log does not contain the partition
+     */
+    void init(org.apache.kafka.streams.processor.ProcessorContext context, StateStore root);
```
vvcephei: Not sure why the diff happened this way, but this is the existing `init` method. I just qualified the reference to `processor.ProcessorContext` and added the note to the javadoc: `This method is not called if {@link StateStore#init(ProcessorContext, StateStore)} is implemented.`

Once the old `processor.ProcessorContext` is deprecated, this method will also become deprecated, pushing store implementers to use the other one instead.

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalProcessorContextReverseAdapter.java
```diff
@@ -0,0 +1,248 @@
...
+import org.apache.kafka.streams.processor.To;
+import org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl;
+import org.apache.kafka.streams.state.StoreBuilder;
+import org.apache.kafka.streams.state.internals.ThreadCache;
+
+import java.io.File;
+import java.time.Duration;
+import java.util.Map;
+
+public final class InternalProcessorContextReverseAdapter implements InternalProcessorContext {
```
vvcephei: I just renamed this from `ProcessorContextReverseAdapter` (which confusingly implemented `InternalProcessorContext`), so that I could add an adapter specifically for non-Internal `ProcessorContext`s.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorAdapter.java
```diff
@@ -47,7 +47,7 @@ private ProcessorAdapter(final org.apache.kafka.streams.processor.Processor<KIn,
     @SuppressWarnings("unchecked")
     @Override
     public void init(final ProcessorContext<KOut, VOut> context) {
-        delegate.init(ProcessorContextReverseAdapter.adapt((InternalApiProcessorContext<Object, Object>) context));
+        delegate.init(InternalProcessorContextReverseAdapter.adapt((InternalApiProcessorContext<Object, Object>) context));
```
vvcephei: renamed.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextReverseAdapter.java
```diff
@@ -18,35 +18,52 @@
...
-import org.apache.kafka.streams.state.StoreBuilder;
-import org.apache.kafka.streams.state.internals.ThreadCache;
 
 import java.io.File;
 import java.time.Duration;
 import java.util.Map;
 
-public final class ProcessorContextReverseAdapter implements InternalProcessorContext {
-    private final InternalApiProcessorContext<Object, Object> delegate;
+public final class ProcessorContextReverseAdapter implements org.apache.kafka.streams.processor.ProcessorContext {
```
vvcephei: Here's the new one, which directly adapts non-Internal `ProcessorContext`. This became necessary, since the `MockProcessorContext` is _not_ an `InternalProcessorContext`, but it still needs to be adapted internally.

Once the KIP is complete, most of these adapters will go away.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextReverseAdapter.java
```diff
@@ -18,35 +18,52 @@
...
+    private final ProcessorContext<Object, Object> delegate;
+    private final DeprecatedForwarder deprecatedForwarder;
 
-    static InternalProcessorContext adapt(final InternalApiProcessorContext<Object, Object> delegate) {
+    public interface DeprecatedForwarder {
+        <K, V> void forward(final K key, final V value, final int childIndex);
+    }
```
vvcephei: Awkwardly, we have no way to adapt the old (deprecated) `forward(final K key, final V value, final int childIndex)`, since it doesn't exist on the new API. If we want to, we could add it, but I don't think we want to.

Another option is to just take this moment to drop it from `processor.ProcessorContext`. It was deprecated in 2.0.0, quite a long time ago.

As it is, I took the most conservative approach I could. We will throw an UnsupportedOperationException only as a last resort.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextReverseAdapter.java
```diff
@@ -18,35 +18,52 @@
...
+    public static final class UnsupportedDeprecatedForwarder implements DeprecatedForwarder {
+        @Override
+        public <K, V> void forward(final K key, final V value, final int childIndex) {
+            throw new UnsupportedOperationException("Forwarding by index was deprecated in 2.0 and is not supported by this ProcessorContext.");
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    public static org.apache.kafka.streams.processor.ProcessorContext adapt(final ProcessorContext<?, ?> delegate,
+                                                                            final DeprecatedForwarder deprecatedForwarder) {
```
vvcephei: In an attempt to avoid throwing an exception unless there's no alternative, callers of `adapt` have to define what happens if someone calls `forward(final K key, final V value, final int childIndex)`

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextReverseAdapter.java
```diff
@@ -18,35 +18,52 @@
...
         if (delegate instanceof ProcessorContextAdapter) {
-            return ((ProcessorContextAdapter<Object, Object>) delegate).delegate();
+            return ((ProcessorContextAdapter<?, ?>) delegate).delegate();
```
vvcephei: If we're adapting an adapter, just unwrap it.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextReverseAdapter.java
```diff
@@ -18,35 +18,52 @@
...
+        } else if (delegate instanceof InternalApiProcessorContext) {
+            return InternalProcessorContextReverseAdapter.adapt((InternalApiProcessorContext<Object, Object>) delegate);
```
vvcephei: If what we're adapting is actually an InternalProcessorContext, good news! We'll just use the other adapter instead.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextReverseAdapter.java
```diff
@@ -18,35 +18,52 @@
...
         } else {
-            return new ProcessorContextReverseAdapter(delegate);
+            return new ProcessorContextReverseAdapter(delegate, deprecatedForwarder);
```
vvcephei: Last resort: what we're adapting really is a plain `ProcessorContext`.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextReverseAdapter.java
```diff
@@ -193,7 +130,7 @@ public Cancellable schedule(final Duration interval, final PunctuationType type,
     @Deprecated
     @Override
     public <K, V> void forward(final K key, final V value, final int childIndex) {
-        delegate.forward(key, value, To.child((currentNode().children()).get(childIndex).name()));
+        deprecatedForwarder.forward(key, value, childIndex);
```
vvcephei: Here's where we call the defined behavior instead of delegating to the nonexistent method on the delegate.

streams/src/test/java/org/apache/kafka/streams/processor/internals/GlobalProcessorContextImplTest.java
```diff
@@ -142,7 +143,7 @@ public void shouldNotAllowToSchedulePunctuations() {
     public void shouldNotAllowInitForKeyValueStore() {
         final StateStore store = globalContext.getStateStore(GLOBAL_KEY_VALUE_STORE_NAME);
         try {
-            store.init(null, null);
+            store.init((ProcessorContext) null, null);
```
vvcephei, line 146: casting `null` to resolve the right overload.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
```diff
@@ -0,0 +1,546 @@
...
+ * <p>
+ * The tests for this class (org.apache.kafka.streams.MockProcessorContextTest) include several behavioral
+ * tests that serve as example usage.
+ * <p>
+ * Note that this class does not take any automated actions (such as firing scheduled punctuators).
+ * It simply captures any data it witnesses.
+ * If you require more automated tests, we recommend wrapping your {@link Processor} in a minimal source-processor-sink
+ * {@link Topology} and using the {@link TopologyTestDriver}.
+ */
+public class MockProcessorContext<KForward, VForward> implements ProcessorContext<KForward, VForward>, RecordCollector.Supplier {
```
vvcephei, line 72: new MockProcessorContext implementing the new `api.ProcessorContext`. Note the generics.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
```diff
@@ -0,0 +1,546 @@
...
+    private String topic;
+    private Integer partition;
+    private Long offset;
+    private Headers headers;
+    private Long timestamp;
+
+    // mocks ================================================
+    private final Map<String, StateStore> stateStores = new HashMap<>();
+    private final List<CapturedPunctuator> punctuators = new LinkedList<>();
+    private final List<CapturedForward<KForward, VForward>> capturedForwards = new LinkedList<>();
```
vvcephei, line 89: Note, we're capturing typed values now.

streams/test-utils/src/test/java/org/apache/kafka/streams/MockApiProcessorContextTest.java
```diff
@@ -0,0 +1,405 @@
...
+import java.util.Iterator;
+import java.util.Properties;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class MockApiProcessorContextTest {
```
vvcephei, line 44: The test for the new `MockProcessorContext`.

vvcephei:
Hey @abbccdda , this is a Part 4 PR that I extracted out from Part 3 (#9221 ) when it became too large. If you have a chance, I'd appreciate your review. Especially wrt the parts that require KIP amendments.

Thanks as always for your reviews!

streams/examples/src/test/java/org/apache/kafka/streams/examples/wordcount/WordCountProcessorTest.java
```diff
@@ -36,19 +36,19 @@
...
+        final MockProcessorContext<String, String> context = new MockProcessorContext<>();
 
         // Create, initialize, and register the state store.
         final KeyValueStore<String, Integer> store =
             Stores.keyValueStoreBuilder(Stores.inMemoryKeyValueStore("Counts"), Serdes.String(), Serdes.Integer())
                 .withLoggingDisabled() // Changelog is not supported by MockProcessorContext.
                 // Caching is disabled by default, but FYI: caching is also not supported by MockProcessorContext.
                 .build();
-        store.init(context, store);
+        store.init(context.getStateStoreContext(), store);
```
vvcephei, line 47: Here's where we're switching contexts to the StateStoreContext to invoke the new API.

streams/src/main/java/org/apache/kafka/streams/processor/StateStore.java
```diff
@@ -61,7 +84,9 @@
      * @throws IllegalStateException If store gets registered after initialized is already finished
      * @throws StreamsException if the store's change log does not contain the partition
      */
-    void init(ProcessorContext context, StateStore root);
+    default void init(final StateStoreContext context, final StateStore root) {
+        init(StoreToProcessorContextAdapter.adapt(context), root);
+    }
```
vvcephei, line 89: Proposing to add this new version of `init` since the old `ProcessorContext` is slated to be deprecated once KIP-478 is fully implemented.

Rather than just accepting the whole new ProcessorContext, I'm proposing to add a new StateStoreContext to avoid leaking a bunch of methods that should really only be invoked in Processor implementations.

streams/src/main/java/org/apache/kafka/streams/processor/StateStoreContext.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.kafka.streams.StreamsMetrics;
+import org.apache.kafka.streams.errors.StreamsException;
+
+import java.io.File;
+import java.util.Map;
+
+/**
+ * Processor context interface.
+ */
+public interface StateStoreContext {
```
vvcephei, line 29: Here's the proposal for the new context. I've preserved all the "general context" members while dropping all the "record context" ones and anything else that seemed inappropriate for state stores to invoke.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StoreToProcessorContextAdapter.java
```diff
@@ -0,0 +1,163 @@
...
+public final class StoreToProcessorContextAdapter implements ProcessorContext {
+    private final StateStoreContext delegate;
+
+    public static ProcessorContext adapt(final StateStoreContext delegate) {
+        if (delegate instanceof ProcessorToStoreContextAdapter) {
+            return ((ProcessorToStoreContextAdapter) delegate).delegate();
+        } else {
+            return new StoreToProcessorContextAdapter(delegate);
+        }
+    }
```
vvcephei: To let the new method delegate to the old one, we adapt the provided StateStoreContext to ProcessorContext.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
```diff
@@ -0,0 +1,602 @@
...
+        );
+    }
+
+    /**
+     * Used to get a {@link StateStoreContext} for use with
+     * {@link StateStore#init(StateStoreContext, StateStore)}
+     * if you need to initialize a store for your tests.
+     * @return a {@link StateStoreContext} that delegates to this ProcessorContext.
+     */
+    public StateStoreContext getStateStoreContext() {
```
vvcephei, line 536: Added this so that users can still just have one context that they can use for initializing their processors and stores.

vvcephei (commented):
On second thought, it seems much cleaner to add a new kind of context for stores.

streams/src/main/java/org/apache/kafka/streams/processor/StateStoreContext.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Returns the application id.
+     *
+     * @return the application id
+     */
+    String applicationId();
```
guozhangwang, line 36: Are appId / taskId / metrics / appConfigs only needed for state store context? Is it possible that they would also be used within Processor?

vvcephei: Thanks, @guozhangwang . I think they'd be used in Processors also, and these methods would still be present on ProcessorContext, so they would still be available.

vvcephei:
This was superceded by other PRs.

PR #8955, opened 2020-06-30
KAFKA-10020: Create a new version of a scala Serdes without name clash (KIP-616)

LMnet:
Implementation of a solution for KIP-616.

Wildcard import of the old `org.apache.kafka.streams.scala.Serdes` leads
to a name clash because some of implicits has the same names as types
from the scala's std lib. New
`org.apache.kafka.streams.scala.serialization.Serdes` is the same as old
`Serdes`, but without name clashes. The old one was marked as deprecated.

Also, missing serdes for `UUID`, `ByteBuffer` and `Short` types are present in
the new `Serdes`.

PR #9223, opened 2020-08-26
KAFKA-10438 Lazy initialization of record header to reduce memory usa…

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-10438

There is no checks for header key so instantiating key (bytes to string) is unnecessary. The risk of this PR is the exception of converting byte[] to string can't be discovered quickly (the conversion error is rare so it should be fine).

**JMH RESULT**

1. ops: +12%
1. The optimization of memory usage is very small as the cost of creating extra ```ByteBuffer``` is almost same to byte array copy (used to construct ```String```). Using large key results in better improvement but I don't think large key is common case.

**BEFORE**
```
Benchmark                                                                     (bufferSupplierStr)  (bytes)  (compressionType)  (headerKeySize)  (maxBatchSize)  (maxHeaderSize)  (messageSize)  (messageVersion)   Mode  Cnt        Score      Error   Units
RecordBatchIterationBenchmark.measureValidation                                        NO_CACHING   RANDOM               NONE               10             200                5           1000                 2  thrpt   15  2035938.174 ± 1653.566   ops/s
RecordBatchIterationBenchmark.measureValidation:·gc.alloc.rate.norm                    NO_CACHING   RANDOM               NONE               10             200                5           1000                 2  thrpt   15     2040.000 ±    0.001    B/op
```

```
Benchmark                                                                     (bufferSupplierStr)  (bytes)  (compressionType)  (headerKeySize)  (maxBatchSize)  (maxHeaderSize)  (messageSize)  (messageVersion)   Mode  Cnt        Score      Error   Units
RecordBatchIterationBenchmark.measureValidation                                        NO_CACHING   RANDOM               NONE               30             200                5           1000                 2  thrpt   15  1979193.376 ± 1239.286   ops/s
RecordBatchIterationBenchmark.measureValidation:·gc.alloc.rate.norm                    NO_CACHING   RANDOM               NONE               30             200                5           1000                 2  thrpt   15     2120.000 ±    0.001    B/op
```

**AFTER**

```
Benchmark                                                                     (bufferSupplierStr)  (bytes)  (compressionType)  (headerKeySize)  (maxBatchSize)  (maxHeaderSize)  (messageSize)  (messageVersion)   Mode  Cnt        Score      Error   Units
RecordBatchIterationBenchmark.measureValidation                                        NO_CACHING   RANDOM               NONE               10             200                5           1000                 2  thrpt   15  2289115.973 ± 2661.856   ops/s
RecordBatchIterationBenchmark.measureValidation:·gc.alloc.rate.norm                    NO_CACHING   RANDOM               NONE               10             200                5           1000                 2  thrpt   15     2032.000 ±    0.001    B/op
```

```
Benchmark                                                                     (bufferSupplierStr)  (bytes)  (compressionType)  (headerKeySize)  (maxBatchSize)  (maxHeaderSize)  (messageSize)  (messageVersion)   Mode  Cnt        Score     Error   Units
RecordBatchIterationBenchmark.measureValidation                                        NO_CACHING   RANDOM               NONE               30             200                5           1000                 2  thrpt   15  2222625.706 ± 908.358   ops/s
RecordBatchIterationBenchmark.measureValidation:·gc.alloc.rate.norm                    NO_CACHING   RANDOM               NONE               30             200                5           1000                 2  thrpt   15     2040.000 ±   0.001    B/op

```

### DIFF
--- a/clients/src/test/java/org/apache/kafka/common/record/DefaultRecordTest.java
+++ b/clients/src/test/java/org/apache/kafka/common/record/DefaultRecordTest.java
@@ -270,7 +270,7 @@ public class DefaultRecordTest {
         DefaultRecord.readPartiallyFrom(inputStream, skipArray, 0L, 0L, RecordBatch.NO_SEQUENCE, null);
     }
 
-    @Test(expected = StringIndexOutOfBoundsException.class)
+    @Test(expected = InvalidRecordException.class)
     public void testInvalidHeaderKey() {
         byte attributes = 0;
         long timestampDelta = 2;
--- a/jmh-benchmarks/src/main/java/org/apache/kafka/jmh/record/RecordBatchBenchmarkSuite.java
+++ b/jmh-benchmarks/src/main/java/org/apache/kafka/jmh/record/RecordBatchBenchmarkSuite.java
@@ -1,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.jmh.record;
+
+import kafka.server.BrokerTopicStats;
+import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.record.AbstractRecords;
+import org.apache.kafka.common.record.BufferSupplier;
+import org.apache.kafka.common.record.CompressionType;
+import org.apache.kafka.common.record.MemoryRecords;
+import org.apache.kafka.common.record.MemoryRecordsBuilder;
+import org.apache.kafka.common.record.Record;
+import org.apache.kafka.common.record.RecordBatch;
+import org.apache.kafka.common.record.TimestampType;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.Random;
+import java.util.stream.IntStream;
+
+import static org.apache.kafka.common.record.RecordBatch.CURRENT_MAGIC_VALUE;
+
+@State(Scope.Benchmark)
+public abstract class RecordBatchBenchmarkSuite {
+    private static final int MAX_HEADER_SIZE = 5;
+    private static final int HEADER_KEY_SIZE = 30;
+
+    private final Random random = new Random(0);
+
+    final int batchCount = 100;
+
+    public enum Bytes {
+        RANDOM, ONES
+    }
+
+    @Param(value = {"1", "2", "10", "50", "200", "500"})
+    private int maxBatchSize = 200;
+
+    @Param(value = {"1", "2"})
+    byte messageVersion = CURRENT_MAGIC_VALUE;
+
+    @Param(value = {"100", "1000", "10000", "100000"})
+    private int messageSize = 1000;
+
+    @Param(value = {"RANDOM", "ONES"})
+    private Bytes bytes = Bytes.RANDOM;
+
+    @Param(value = {"NO_CACHING", "CREATE"})
+    private String bufferSupplierStr = "NO_CACHING";
+
+    // zero starting offset is much faster for v1 batches, but that will almost never happen
+    int startingOffset;
+
+    // Used by measureSingleMessage
+    ByteBuffer singleBatchBuffer;
+
+    // Used by measureVariableBatchSize
+    ByteBuffer[] batchBuffers;
+    BufferSupplier bufferSupplier;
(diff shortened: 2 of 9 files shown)


clients/src/test/java/org/apache/kafka/common/record/DefaultRecordTest.java
```diff
@@ -270,7 +270,7 @@ public void testInvalidNumHeadersPartial() throws IOException {
         DefaultRecord.readPartiallyFrom(inputStream, skipArray, 0L, 0L, RecordBatch.NO_SEQUENCE, null);
     }
 
-    @Test(expected = StringIndexOutOfBoundsException.class)
+    @Test(expected = InvalidRecordException.class)
```
ijuma, line 273: Since we parse the header lazily now, what is causing this exception?

ijuma:
Good change. Any thoughts on test for this?

clients/src/test/java/org/apache/kafka/common/record/DefaultRecordTest.java
chia7712: The exception is caused by ```headerKeyBuffer.limit(headerKeySize);```. The new limit is larger than capacity

chia7712:
> Any thoughts on test for this?

The JMH is attached.

chia7712:
the failed build (jdk8 + scala 2.12) is related to #8955

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/record/RecordBatchIterationBenchmark.java
```diff
@@ -142,15 +169,15 @@ private ByteBuffer createBatch(int batchSize) {
...
+            builder.append(0, null, value, headers);
         }
         return builder.build().buffer();
     }
 
     @Benchmark
     public void measureValidation(Blackhole bh) throws IOException {
         MemoryRecords records = MemoryRecords.readableRecords(singleBatchBuffer.duplicate());
-        LogValidator.validateMessagesAndAssignOffsetsCompressed(records, new TopicPartition("a", 0),
+        LogValidator.validateMessagesAndAssignOffsets(records, new TopicPartition("a", 0),
```
ijuma: Why did we change this?

chia7712: > Why did we change this?

run JMH with uncompressed data.

The alternative way is that we can add an new benchmark for ```assignOffsetsNonCompressed```. However, the test data of this class includes both uncompressed and compressed so it is weird to input compressed data to test ```assignOffsetsNonCompressed```.

ijuma: Right, but the goal of this benchmark was to test the compressed case where there is additional overhead due to compression buffer allocation and so on. As you say, it would be better to have another benchmark for the uncompressed case.

chia7712: > it would be better to have another benchmark for the uncompressed case.

I creates two new classes - ```CompressedRecordBatchValidationBenchmark``` and ```UncompressedRecordBatchValidationBenchmark```. They use different compression type and measure different validation function.

chia7712:
```
kafka.network.DynamicConnectionQuotaTest.testDynamicListenerConnectionCreationRateQuota
```
pass on my local.

chia7712:
```
 Build / JDK 8 / org.apache.kafka.connect.mirror.MirrorConnectorsIntegrationTest.testReplication	1 分 17 秒	1
 Build / JDK 15 / kafka.api.SaslSslAdminIntegrationTest.testCreateDeleteTopics	2 分 2 秒	1
 Build / JDK 15 / org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]	1 分 21 秒	1
 Build / JDK 15 / org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]	
```
I don't observe the relation between those failures and this patch.

chia7712:
```
Build / JDK 8 / kafka.api.TransactionsBounceTest.testWithGroupMetadata
```
this flaky is traced by https://issues.apache.org/jira/browse/KAFKA-10251

chia7712:
@ijuma please take a look :)

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/record/RecordBatchBenchmarkSuite.java
```diff
@@ -0,0 +1,149 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.Random;
+import java.util.stream.IntStream;
+
+import static org.apache.kafka.common.record.RecordBatch.CURRENT_MAGIC_VALUE;
+
+@State(Scope.Benchmark)
+public abstract class RecordBatchBenchmarkSuite {
```
ijuma: Maybe this should be `BaseRecordBatchBenchmark`?

chia7712: done

chia7712:
```
Build / JDK 8 / org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]
```

only jdk 8 gets failure. This flaky is unrelated to this patch. @ijuma Please take a look.

ijuma (approved):
LGTM, thanks.

ijuma:
Unrelated flaky failures:

>     Build / JDK 8 / org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]
>     Build / JDK 8 / org.apache.kafka.streams.integration.RestoreIntegrationTest.shouldRecycleStateFromStandbyTaskPromotedToActiveTaskAndNotRestore

PR #9224, opened 2020-08-26
KAFKA-10304: refactor MM2 integration tests

ning2008wisc:
The main proposals of this PR:
(1) extract the common functions into a base class `MirrorConnectorsIntegrationBaseTest`
(2) test in SSL-enabled cluster


connect/mirror/src/test/java/org/apache/kafka/connect/mirror/TestUtils.java
```diff
@@ -16,12 +16,51 @@
...
+
+import static org.apache.kafka.connect.mirror.TestUtils.NUM_WORKERS;
+import static org.apache.kafka.test.TestUtils.waitForCondition;
+import static org.apache.kafka.connect.util.clusters.EmbeddedConnectClusterAssertions.CONNECTOR_SETUP_DURATION_MS;
+import static org.junit.Assert.assertTrue;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class TestUtils {
```
ning2008wisc: propose `TestUtils` to be the central place to host common functions that will be used by integration tests

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/TestUtils.java
```diff
@@ -32,4 +71,141 @@
         }
         return props;
     }
+    
+    public static void waitUntilMirrorMakerIsRunning(EmbeddedConnectCluster connectCluster, List<Class> mirrorClasses,
```
ning2008wisc: This is the simple move from `MirrorConnectorsIntegrationTest` with generalization of connector class

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/TestUtils.java
```diff
@@ -32,4 +73,166 @@
         }
         return props;
     }
+    
+    /*
+     * launch the connectors on kafka connect cluster, then check if they are running
+     */
+    public static void waitUntilMirrorMakerIsRunning(EmbeddedConnectCluster connectCluster, List<Class> connectorClasses,
```
ning2008wisc: this is mostly copy-paste from `MirrorConnectorsIntegrationTest`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -207,13 +173,16 @@ public void testReplication() throws InterruptedException {
...
 
-        waitUntilMirrorMakerIsRunning(primary, mm2Config, "backup", "primary");
+        waitUntilMirrorMakerIsRunning(primary, CONNECTOR_LIST, mm2Config, "backup", "primary");   
 
         MirrorClient primaryClient = new MirrorClient(mm2Config.clientConfig("primary"));
         MirrorClient backupClient = new MirrorClient(mm2Config.clientConfig("backup"));
-
+        
+        assertEquals("topic config was not synced", TopicConfig.CLEANUP_POLICY_COMPACT, 
+                getTopicConfig(backup.kafka(), "primary.test-topic-1", TopicConfig.CLEANUP_POLICY_CONFIG));
```
ning2008wisc: Add a check for topic config sync, since the topic created on primary cluster has a "cleanup.policy" config

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -396,27 +330,67 @@ public void testOneWayReplicationWithAutoOffsetSync() throws InterruptedExceptio
...
+        
+        consumer.close();
+    }
+    
+    void createTopics() {
+        // to verify topic config will be sync-ed across clusters
+        Map<String, String> topicConfig = new HashMap<>();
+        topicConfig.put(TopicConfig.CLEANUP_POLICY_CONFIG, TopicConfig.CLEANUP_POLICY_COMPACT);
+        // create these topics before starting the connectors so we don't need to wait for discovery
+        primary.kafka().createTopic("test-topic-1", NUM_PARTITIONS, 1, topicConfig);
```
ning2008wisc: when creating `test-topic-1` topic on primary cluster, add a topic config. Later on, we will check if the config is synced from primary to backup cluster.

ning2008wisc:
very appreciated for any feedback on what to test additionally and how to get close to the real scenario. Some of the ideas come from https://github.com/apache/kafka/blob/trunk/connect/runtime/src/test/java/org/apache/kafka/connect/integration/ConnectWorkerIntegrationTest.java

ning2008wisc:
@ryannedolan @mimaison when possible, very appreciated for your attentions and feedback :) Thanks

ning2008wisc:
Screen Shot 2020-10-04 at 10 50 17 PM
./gradlew connect:mirror:test -> all tests pass

ning2008wisc:
@mimaison I think this pr is ready for initial review. Please take your time when available. Thanks

ning2008wisc:
Hello @mimaison @ryannedolan this PR does not contain any functional change, it should be straightforward to review. Thanks for your time.

ning2008wisc:
@edoardocomar @rhauch given this PR does not contain functional change and have been pending over 45 days, could you please direct me to the right reviewers to take a look? Thanks

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+import static org.apache.kafka.connect.mirror.TestUtils.expectedRecords;
+import static org.apache.kafka.connect.util.clusters.EmbeddedConnectClusterAssertions.CONNECTOR_SETUP_DURATION_MS;
+import static org.apache.kafka.test.TestUtils.waitForCondition;
+import kafka.server.KafkaConfig$;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+//import org.junit.After;
```
mimaison: We can remove this

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+import org.apache.kafka.common.config.types.Password;
+import org.apache.kafka.common.utils.Exit;
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.connect.errors.ConnectException;
+import org.apache.kafka.connect.mirror.MirrorMakerConfig;
+import org.apache.kafka.connect.mirror.SourceAndTarget;
+import org.apache.kafka.connect.util.clusters.EmbeddedConnectCluster;
+import org.apache.kafka.connect.util.clusters.EmbeddedKafkaCluster;
+import org.apache.kafka.test.IntegrationTest;
+import org.junit.experimental.categories.Category;
```
mimaison: This can go with the other `org.junit` imports below

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+     * retrieve the config value based on the input cluster, topic and config name
+     */
+    protected static String getTopicConfig(EmbeddedKafkaCluster cluster, String topic, String configName) {
+        Admin client = cluster.createAdminClient();
+        Collection<ConfigResource> cr =  Collections.singleton(
+                new ConfigResource(ConfigResource.Type.TOPIC, topic)); 
+        try {
+            DescribeConfigsResult configsResult = client.describeConfigs(cr);
+            Config allConfigs = (Config) configsResult.all().get().values().toArray()[0];
+            Iterator configIterator = allConfigs.entries().iterator();
```
mimaison: `Iterator`-> `Iterator<ConfigEntry>`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -0,0 +1,372 @@
...
+ *
+ * MM2 is configured with active/active replication between two Kafka clusters. Tests validate that
+ * records sent to either cluster arrive at the other cluster. Then, a consumer group is migrated from
+ * one cluster to the other and back. Tests validate that consumer offsets are translated and replicated
+ * between clusters during this failover and failback.
+ */
+@Category(IntegrationTest.class)
+public class MirrorConnectorsIntegrationTest extends MirrorConnectorsIntegrationBaseTest {
+
+    private static final Logger log = LoggerFactory.getLogger(MirrorConnectorsIntegrationTest.class);
```
mimaison: It looks like this is not used, can we remove it?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -0,0 +1,372 @@
...
+    
+    private static final List<Class> CONNECTOR_LIST =
```
mimaison: We can use `List<Class<? extends Connector>` to avoid the warning

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -0,0 +1,372 @@
...
+     * under broker restart / failure
+     */
+    @Test
+    public void testWithBrokerRestart() throws InterruptedException {
+        String consumerGroupName = "consumer-group-testWithBrokerRestart";
+        // test with a higher number of records
+        int numRecords = NUM_RECORDS_PRODUCED * 8;
+        // the sleep time between two produces used by background producer 
+        int sleepMs = 50;
+        int joinTimeoutMs = 1000;
```
mimaison: This is not used, can we remove it?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -0,0 +1,372 @@
...
+        waitUntilMirrorMakerIsRunning(backup, CONNECTOR_LIST, mm2Config, "primary", "backup");
+
+        // create a consumer at backup cluster with same consumer group Id to consume 1 topic
+        Consumer<byte[], byte[]> backupConsumer = backup.kafka().createConsumerAndSubscribeTo(
+            consumerProps, "primary.test-topic-1");
+
+        waitForConsumerGroupOffsetSync(backup, backupConsumer, Collections.singletonList("primary.test-topic-1"), 
+            consumerGroupName, NUM_RECORDS_PRODUCED);
+
+        ConsumerRecords records = backupConsumer.poll(Duration.ofMillis(500));
```
mimaison: `ConsumerRecords` -> `ConsumerRecords<byte[], byte[]>`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+    }
+ 
+    /*
+     * delete all topics of the input kafka cluster
+     */
+    protected static void deleteAllTopics(EmbeddedKafkaCluster cluster) {
+        Admin client = cluster.createAdminClient();
+        try {
+            client.deleteTopics(client.listTopics().names().get());
+        } catch (Throwable e) {
```
mimaison: What do we do if there's an exception? If it's expected, let's make it clear

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.After;
+import org.junit.Test;
+
+/**
+ * Tests MM2 replication with SSL enabled at backup kafka cluster
+ */
+@Category(IntegrationTest.class)
+public class MirrorConnectorsIntegrationSSLTest extends MirrorConnectorsIntegrationBaseTest {
+
+    private static final Logger log = LoggerFactory.getLogger(MirrorConnectorsIntegrationSSLTest.class);
```
mimaison: It looks like this is not used anywhere

build.gradle
```diff
@@ -2058,6 +2058,7 @@ project(':connect:mirror') {
     testCompile libs.junitJupiterApi
     testCompile libs.junitVintageEngine
     testCompile libs.mockitoCore
+    testCompile libs.bcpkix
```
mimaison: Is this needed at compile time? As far as I can tell, it's only needed when running the SSL test, so it could be `testRuntime libs.bcpkix`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+        // create these topics before starting the connectors so we don't need to wait for discovery
+        primary.kafka().createTopic("test-topic-1", NUM_PARTITIONS, 1, topicConfig);
+        primary.kafka().createTopic("backup.test-topic-1", 1);
+        primary.kafka().createTopic("heartbeats", 1);
+        backup.kafka().createTopic("test-topic-1", NUM_PARTITIONS);
+        backup.kafka().createTopic("primary.test-topic-1", 1);
+        backup.kafka().createTopic("heartbeats", 1);
+    }
+    
+    protected KafkaConsumer<String, String> createSslConsumer(Map<String, Object> consumerProps, String... topics) {
```
mimaison: I find having a method specific for SSL strange. Callers should not have to know, this should be retrieved automatically based on the cluster being targeted

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
```diff
@@ -0,0 +1,98 @@
...
+    
+    @Before
+    public void setup() throws InterruptedException {
+        try {
+            Map<String, Object> sslConfig = TestSslUtils.createSslConfig(false, true, Mode.SERVER, TestUtils.tempFile(), "testCert");
+            backupBrokerProps.put(KafkaConfig$.MODULE$.ListenersProp(), "SSL://localhost:0");
+            backupBrokerProps.put(KafkaConfig$.MODULE$.InterBrokerListenerNameProp(), "SSL");
+            backupBrokerProps.putAll(sslConfig);
+        } catch (final Exception e) {
+            throw new RuntimeException(e);
```
mimaison: Why are we rethrowing as `RuntimeException`?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        // one-way replication from primary -> backup
+        mm2Props.put("backup->primary.enabled", "false");
+        mm2Config = new MirrorMakerConfig(mm2Props);
+
+        waitUntilMirrorMakerIsRunning(backup, CONNECTOR_LIST, mm2Config, "primary", "backup");
+
+        // sleep few seconds to let MM2 replicate some records for "end" consumer to consume them
+        Thread.sleep(TimeUnit.SECONDS.toMillis(3));
```
mimaison: That looks pretty brittle? Is there a condition we can wait on?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+        consumer.commitSync();
+        consumer.close();
+    }
+   
+    /*
+     * MM2 config to use in integration tests
+     */
+    protected static Map<String, String> basicMM2Config() {
+        Map<String, String> mm2Props = new HashMap<>();
+        mm2Props.put("clusters", "primary, backup");
```
mimaison: Could the cluster aliases be constant as these are used all over the place

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+        primary.kafka().createTopic("heartbeats", 1);
+        backup.kafka().createTopic("test-topic-1", NUM_PARTITIONS);
+        backup.kafka().createTopic("primary.test-topic-1", 1);
+        backup.kafka().createTopic("heartbeats", 1);
+    }
+    
+    protected KafkaConsumer<String, String> createSslConsumer(Map<String, Object> consumerProps, String... topics) {
+        Map<String, Object> props = new HashMap<>(consumerProps);
+
+        if (props.get(ConsumerConfig.GROUP_ID_CONFIG) == null) {
```
mimaison: We can use `putIfAbstent()`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+
+        // ssl config
+        props.putAll(sslProps.entrySet().stream().collect(Collectors.toMap(
+            e -> String.valueOf(e.getKey()), e ->  String.valueOf(e.getValue()))));
+        
+        KafkaConsumer<String, String> consumer;
+        try {
+            consumer = new KafkaConsumer<>(props);
+        } catch (Throwable t) {
+            throw new ConnectException("Failed to create consumer", t);
```
mimaison: Why are we rethrowing as `ConnectException`?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -0,0 +1,372 @@
...
+        }, CHECKPOINT_DURATION_MS, "Offsets not translated upstream to primary cluster.");
+
+        Map<TopicPartition, OffsetAndMetadata> primaryOffsets = primaryClient.remoteConsumerOffsets(consumerGroupName, "backup",
+                Duration.ofMillis(CHECKPOINT_DURATION_MS));
+ 
+        // Failback consumer group to primary cluster
+        backupConsumer = primary.kafka().createConsumer(Collections.singletonMap("group.id", consumerGroupName));
+        backupConsumer.assign(primaryOffsets.keySet());
+        primaryOffsets.forEach(backupConsumer::seek);
+        backupConsumer.poll(Duration.ofMillis(500));
```
mimaison: We seem to use `poll(Duration.ofMillis(500))` in many places, can we have a constant?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+
+    
+    protected Map<String, String> mm2Props;
+    protected MirrorMakerConfig mm2Config; 
+    protected EmbeddedConnectCluster primary;
+    protected EmbeddedConnectCluster backup;
+    
+    private final AtomicBoolean exited = new AtomicBoolean(false);
+    private Properties primaryBrokerProps = new Properties();
+    protected Properties backupBrokerProps = new Properties();
```
mimaison, line 103: Why is this field `protected` while `primaryBrokerProps` is `private`?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+    private Map<String, String> primaryWorkerProps;
```
mimaison: Should we do `new HashMap<>();` like the line below? Why is it different?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+        mm2Props.put("status.storage.replication.factor", "1");
+        mm2Props.put("replication.factor", "1");
+        
+        return mm2Props;
+    }
+    
+    /*
+     * restart kafka broker and make sure it is successful
+     */
+    protected static void restartKafkaBroker(EmbeddedConnectCluster connect) throws InterruptedException {
```
mimaison: This restarts all brokers. I find it strange this takes a `EmbeddedConnectCluster`. Also I wonder why this is static.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+        Thread.sleep(TimeUnit.SECONDS.toMillis(10));
+
+        connect.assertions().assertExactlyNumWorkersAreUp(NUM_WORKERS,
+                "Group of workers did not remain the same within the designated time.");
+
+        // Allow for the workers to rebalance and reach a steady state
+        Thread.sleep(TimeUnit.SECONDS.toMillis(10));
+
+        // Expect that the broker has started again
+        assertTrue("Failed to start kafka broker within " + CONNECTOR_SETUP_DURATION_MS + "ms",
```
mimaison: Should we use this check as a condition to wait? Sleeping 10 secs feels pretty brittle

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
```diff
@@ -0,0 +1,98 @@
...
+            backupBrokerProps.put(KafkaConfig$.MODULE$.InterBrokerListenerNameProp(), "SSL");
+            backupBrokerProps.putAll(sslConfig);
+        } catch (final Exception e) {
+            throw new RuntimeException(e);
+        }
+        startClusters();
+    }
+    
+    @After
+    public void close() {
```
mimaison: Can we move this in the base class?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -0,0 +1,372 @@
...
+@Category(IntegrationTest.class)
+public class MirrorConnectorsIntegrationTest extends MirrorConnectorsIntegrationBaseTest {
+
+    private static final Logger log = LoggerFactory.getLogger(MirrorConnectorsIntegrationTest.class);
+    
+    private static final List<Class> CONNECTOR_LIST = 
+            Arrays.asList(MirrorSourceConnector.class, MirrorCheckpointConnector.class, MirrorHeartbeatConnector.class);
+    
+    @Before
+    public void setup() throws InterruptedException {
```
mimaison: Can we move this in the base class?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,423 @@
...
+            return numExpectedRecords == totalConsumedRecords.addAndGet(records.count());
+        }, RECORD_CONSUME_DURATION_MS, "Consumer cannot consume all records in time");
+        consumer.commitSync();
+        consumer.close();
+    }
+   
+    /*
+     * MM2 config to use in integration tests
+     */
+    protected static Map<String, String> basicMM2Config() {
```
mimaison: Can this be `private`?

mimaison (changes requested):
Thanks for the PR @ning2008wisc.

The MM2 integration test is pretty complicated and I was hoping the refactoring would simplify it/remove code. I understand this PR also contains new feature but I feel like overall the complexity stayed or increased.

I've left a bunch of comments. Some of them are nits/style but there're also questions about the overall shape.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
ning2008wisc: referred to the example at https://github.com/apache/kafka/blob/trunk/streams/src/test/java/org/apache/kafka/streams/integration/ResetIntegrationWithSslTest.java#L58

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
ning2008wisc: because `backupBrokerProps` is referred in `MirrorConnectorsIntegrationSSLTest`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
ning2008wisc: if move to base class, will close() still be called at demolition stage of test?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
ning2008wisc: in `MirrorConnectorsIntegrationSSLTest`, we put ssl config into `backupBrokerProps`, so the setup() of both `MirrorConnectorsIntegrationTest` and `MirrorConnectorsIntegrationSSLTest` will be slightly different

ning2008wisc:
Thanks @mimaison for your high-level advice and detailed review.

(1) I responded to your every comments. A "thumb-up" means I made the suggested change
(2) regarding the increased complexity, here is my analysis for each test case:
     (2a) `testReplication`: This is an existing test and no new stuff are added, except checking if topic config are also mirrored. If we believe the topic config validation is trivial, I am more than happy to remove it
     (2b) `testReplicationWithEmptyPartition`: This is an existing test and no new stuff are added
     (2c) `testOneWayReplicationWithAutoOffsetSync`: This is an existing test and no new stuff are added
     (2d) `testWithBrokerRestart`: This is a **new** test, which introduced background producer and consumer, called `ThreadedProducer` and `ThreadedConsumer`.  The purpose of background producer and consumer is to better test the failure case to avoid serialization (produce -> broker restart -> consumer) which can not reveal the insight that consumers consume the same records more than once when broker failure happens.
     (2e) `MirrorConnectorsIntegrationSSLTest`: This is a **new** test and it should be pretty lean and neat.

In addition, the **new** base test class `MirrorConnectorsIntegrationBaseTest` follows the existing practice of K-stream integration test structure https://github.com/apache/kafka/blob/trunk/streams/src/test/java/org/apache/kafka/streams/integration/ResetIntegrationWithSslTest.java#L37 so I assume it is likely the right way to structure tests with more than one scenarios (SSL v.s. non-SSL)

If we have to control the complexity, I would prefer to drop `testWithBrokerRestart` and keep `MirrorConnectorsIntegrationSSLTest`, as it makes sense to run simple validation in SSL setup.

Thanks and would like to hear more thoughts and feedback.

ning2008wisc:
A kind reminder for additional feedback and comments, based on my response above

mimaison:
@ning2008wisc Thanks for the quick update. It's still on my todo list but unfortunately I don't have time to do reviews this week. I hope to take another look next week.

mimaison:
@ning2008wisc I've not forgotten this PR, I just haven't had time to do reviews yet :(

mimaison:
> If we have to control the complexity, I would prefer to drop testWithBrokerRestart and keep MirrorConnectorsIntegrationSSLTest, as it makes sense to run simple validation in SSL setup.

I think that would be a good idea. We can try to introduce this test in a follow up PR.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
mimaison: Yes it will

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,407 @@
...
+    protected EmbeddedConnectCluster backup;
+    
+    private final AtomicBoolean exited = new AtomicBoolean(false);
+    private Properties primaryBrokerProps = new Properties();
+    protected Properties backupBrokerProps = new Properties();
+    private Map<String, String> primaryWorkerProps = new HashMap<>();
+    private Map<String, String> backupWorkerProps = new HashMap<>();
+    private Properties sslProps = new Properties();
+    
+    private void loadSslPropsFromBrokerConfig() {
```
mimaison: Why do we have SSL specific methods here? Could we move all the SSL bits into the SSL class?

We have fields for the configurations. So we could set them accordingly (without or without SSL) in each concrete class. Then in the base class, we just use the fields to create the clusters without having to know if it's SSL or not.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,407 @@
...
+    /*
+     * delete all topics of the input kafka cluster
+     */
+    protected static void deleteAllTopics(EmbeddedKafkaCluster cluster) {
+        Admin client = cluster.createAdminClient();
+        try {
+            client.deleteTopics(client.listTopics().names().get());
+        } catch (Throwable e) {
+            // should not run into exception normally. In case of Exception, 
+            // simply fail the test and investigate
```
mimaison: By catching `Throwable`, aren't we silencing the error and not immediately failing the test?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
```diff
@@ -0,0 +1,89 @@
...
+public class MirrorConnectorsIntegrationSSLTest extends MirrorConnectorsIntegrationBaseTest {
+    
+    private static final List<Class<? extends Connector>> CONNECTOR_LIST = 
+            Arrays.asList(MirrorSourceConnector.class, MirrorCheckpointConnector.class, MirrorHeartbeatConnector.class);
+    
+    @Before
+    public void setup() throws InterruptedException {
+        try {
+            Map<String, Object> sslConfig = TestSslUtils.createSslConfig(false, true, Mode.SERVER, TestUtils.tempFile(), "testCert");
+            backupBrokerProps.put(KafkaConfig$.MODULE$.ListenersProp(), "SSL://localhost:0");
```
mimaison: Is there a reason we are only enabling SSL in one of the 2 clusters?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.After;
+import org.junit.Test;
+
+/**
+ * Tests MM2 replication with SSL enabled at backup kafka cluster
+ */
+@Category(IntegrationTest.class)
+public class MirrorConnectorsIntegrationSSLTest extends MirrorConnectorsIntegrationBaseTest {
+    
+    private static final List<Class<? extends Connector>> CONNECTOR_LIST =
```
mimaison: We have this field in both concrete class. Can we move it to the base class instead?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,407 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import static org.junit.Assert.assertFalse;
+import org.junit.experimental.categories.Category;
+
+/**
+ * Common Test functions for MM2 integration tests
+ */
+@Category(IntegrationTest.class)
+public class MirrorConnectorsIntegrationBaseTest {
```
mimaison: Should this class be `abstract`?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,407 @@
...
+            Config allConfigs = (Config) configsResult.all().get().values().toArray()[0];
+            Iterator<ConfigEntry> configIterator = allConfigs.entries().iterator();
+            while (configIterator.hasNext()) {
+                ConfigEntry currentConfig = configIterator.next();     
+                if (currentConfig.name().equals(configName)) {
+                    return currentConfig.value();
+                }
+            }
+        } catch (Throwable e) {
+            // should not run into exception normally. In case of Exception,
```
mimaison: By catching Throwable, aren't we silencing the error and not immediately failing the test?

Also we should not be catching Throwable, if the JVM throws an Error we want it to fail the test and never want to handle it. There're a bunch of them in these files

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/TestUtils.java
```diff
@@ -32,4 +32,15 @@
         }
         return props;
     }
-}
+    
+    /*
+     * return records with different but predictable key and value 
+     */
+    public static Map<String, String> expectedRecords(int numRecords) {
```
mimaison: Can we find a better name?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/TestUtils.java
```diff
@@ -32,4 +32,15 @@
...
+        Map<String, String> expectedRecords = new HashMap<>();
+        for (int i = 0; i < numRecords; i++) {
+            expectedRecords.put("key-" + i, "message-" + i);
+        }
+        return expectedRecords;
+    }
+}
```
mimaison: nit: we tend to have a new line at the end of files

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,407 @@
...
+        }
+        KafkaConsumer<String, String> consumer = new KafkaConsumer<>(props);
+        consumer.subscribe(Arrays.asList(topics));
+        return consumer;
+    }
+    
+    /*
+     * Generate some consumer activity on both clusters to ensure the checkpoint connector always starts promptly
+     */
+    protected void dummyConsumption() throws InterruptedException {
```
mimaison: Can we find a better name?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,407 @@
...
+        
+        // wait for the kafka brokers to come back online
+        waitForCondition(() -> {
+            return kafkaCluster.runningBrokers().size() > 0;
+        }, BROKER_RESTART_TIMEOUT_MS, "Failed to start kafka broker within");
+    }
+    
+    protected void createTopics() {
+        // to verify topic config will be sync-ed across clusters
+        Map<String, String> topicConfig = new HashMap<>();
```
mimaison: We can use `Collections.singletonMap()`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,407 @@
...
+    }
+    
+    protected KafkaConsumer<String, String> createConsumer(EmbeddedConnectCluster connect, Map<String, Object> consumerProps, String... topics) {
+        Map<String, Object> props = new HashMap<>(consumerProps);
+
+        props.putIfAbsent(ConsumerConfig.GROUP_ID_CONFIG, UUID.randomUUID().toString());
+        props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, connect.kafka().bootstrapServers());
+        props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false");
+        props.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+        props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.StringDeserializer");
```
mimaison: We can use `StringDeserializer.class.getName()`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
```diff
@@ -0,0 +1,89 @@
...
+        startClusters();
+    }
+    
+    @After
+    public void close() {
+        shutdownClusters();
+    }
+    
+    @Test
+    public void testReplicationSSL() throws InterruptedException {
```
mimaison: This test is not specific to SSL, why do we have it ?
Should instead all tests in MirrorConnectorsIntegrationTest go to the base class so we run them both with and without SSL?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -0,0 +1,370 @@
...
+            consumerGroupName, NUM_RECORDS_PRODUCED);
+
+        records = backupConsumer.poll(Duration.ofMillis(CONSUMER_POLL_TIMEOUT_MS));
+        // similar reasoning as above, no more records to consume by the same consumer group at backup cluster
+        assertEquals("consumer record size is not zero", 0, records.count());
+        backupConsumer.close();
+    }
+    
+    /*
+     * This test is to validate MirrorSourceConnector follows "at most once" delivery guarantee
```
mimaison: Isn't MM2/Connect using at least once by default? ie, the producer in the runtime can cause duplicates.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -0,0 +1,370 @@
...
+     * under broker restart / failure
+     */
+    @Test
+    public void testWithBrokerRestart() throws InterruptedException {
```
mimaison: I don't understand what this test is doing.
Why do we need background clients instead of producing upfront and consuming the data mirrorred at the end of the test?
It looks like we are testing the primary->backup scenario but we are restarting the backup cluster. The source connector should not interact with the backup cluster.

mimaison (changes requested):
Thanks for the updates. I've made another pass.

I've left a bunch of minor comments but I'd like to highlight bigger issues on the following 2 points:

- SSL tests: Even though MM2 does not do anything special with SSL (nor SASL) I think it's fine adding tests for SSL. However this should not cause such a large change. In this PR, SSL has leaked in all classes. Instead we want an SSL class that just changes the configuration.

  For a good example, see https://github.com/apache/kafka/blob/trunk/core/src/test/scala/integration/kafka/api/SslConsumerTest.scala. If you can get it in a similar shape, then it makes sense to cover SSL

- testWithBrokerRestart: Testing semantics of connectors is tricky. I don't think the current test does it properly at the moment. Have a look at the comments I left on the test. Maybe we should focus on the other issues first and have another attempt in a different PR. WDYT?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
ning2008wisc: `warmUpConsumper` sounds better?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/TestUtils.java
ning2008wisc: rename to `generateRecords` ?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
ning2008wisc: Good point. I am thinking in future, if we have test for `MirrorSinkConnector` (which will also extend the `MirrorConnectorsIntegrationBaseTest` class), it may be more flexible to have each extended test class to define what is the list of connectors to boot up?

I am also fine to move `CONNECTOR_LIST ` into the base class now and move it out later on when we have `MirrorSinkConnector`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
ning2008wisc: Good question. The primary reasons are: (1) I have seen some MM2 use cases which mirror data from in-house Kafka cluster to public hosted Kafka. For example, https://issues.apache.org/jira/browse/KAFKA-10704 So this setup (non-SSL -> SSL) seems to represent a certain coverage of use cases. (2) There are 4 combinations between SSL and non-SSL, so adding all combinations are great, but also verbose. Proving MM2 can work with one SSL-enabled cluster should be convincing enough to claim "MM2 can mirror in SSL case".

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
ning2008wisc: since this is in Integration test, if we run into `catch`, the test should fail immediately, right? If you have suggestions on what to put here, I am happy to follow :)

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
ning2008wisc: catching Throwable has been there since the original author wrote MM2. It seems a tricky thing and I am happy to hear your suggestions on what to put in the "catch" block.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
ning2008wisc: sorry - it is a typo, should be "at least once". I am removing this test from this PR

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
ning2008wisc: > Why do we need background clients instead of producing upfront and consuming the data mirrorred at the end of the test?

I think I answered the above question as the comments  in the code. Please see below,

> // the purpose of background producer and consumer is to better test the failure case
> // to avoid serialization (produce -> broker restart -> consumer) by decoupling the
> // producer, embedded kafka and consumer. Since the consumer offsets are stored in
> // kafka topic at backup cluster and offset commit is periodic, when kafka broker (at backup)
> // restarts, duplicate records will be consumed, meaning "at least once" delivery guarantee

I believe this test has unique value and is more close to the realistic scenario, since the producer -> kafka -> consumer are always running on different machines, rather than **within one** process or thread.

Due to the complexity limitation, I am happy to remove this test for now

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
ning2008wisc: > Could we move all the SSL bits into the SSL class?
I believe yes

> We have fields for the configurations. So we could set them accordingly (without or without SSL) in each concrete class. Then in the base class, we just use the fields to create the clusters without having to know if it's SSL or not.
I agree

ning2008wisc:
Thanks @mimaison  for your suggestive and insightful feedback. 

Regarding to your 2 major concerns, I agree and I believe there exists a feasible solution, especially for a very lean SSL test. I will update this PR early next week by addressing all your major and minor comments.

ning2008wisc:
@mimaison Thanks again for your previous detailed review. I updated the PR to resolve the exact 2 concerns you raised.

Very appreciated for your another review!

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,617 @@
...
+    private static final Logger log = LoggerFactory.getLogger(MirrorConnectorsIntegrationBaseTest.class);
+    
+    private static final int NUM_RECORDS_PER_PARTITION = 10;
+    private static final int NUM_PARTITIONS = 10;
+    private static final int NUM_RECORDS_PRODUCED = NUM_PARTITIONS * NUM_RECORDS_PER_PARTITION;
+    private static final int RECORD_TRANSFER_DURATION_MS = 30_000;
+    private static final int CHECKPOINT_DURATION_MS = 20_000;
+    private static final int RECORD_CONSUME_DURATION_MS = 20_000;
+    private static final int OFFSET_SYNC_DURATION_MS = 30_000;
+    private static final int NUM_WORKERS = 3;
```
mimaison, line 86: This constant is unused. Should we use it instead of hardcoding `3` below when we create connect workers?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,617 @@
...
+    private static final int CONSUMER_POLL_TIMEOUT_MS = 500;
+    private static final int BROKER_RESTART_TIMEOUT_MS = 10_000;
+    private static final long DEFAULT_PRODUCE_SEND_DURATION_MS = TimeUnit.SECONDS.toMillis(120);
```
mimaison: This is unused too

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,617 @@
...
+    private MirrorMakerConfig mm2Config; 
+    private EmbeddedConnectCluster primary;
+    private EmbeddedConnectCluster backup;
+    
+    private final AtomicBoolean exited = new AtomicBoolean(false);
+    protected Properties primaryBrokerProps = new Properties();
+    protected Properties backupBrokerProps = new Properties();
+    private Map<String, String> primaryWorkerProps = new HashMap<>();
+    private Map<String, String> backupWorkerProps = new HashMap<>(); 
+    abstract Map<String, Object> getSslConfig();
```
mimaison: Ideally we want to get rid of this method as it makes no sense in tests that are not SSL.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,617 @@
...
+
+    protected void startClusters() throws InterruptedException {
+        primaryBrokerProps.put("auto.create.topics.enable", "false");
+        backupBrokerProps.put("auto.create.topics.enable", "false");
+        
+        mm2Props = basicMM2Config();
+        
+        final Map<String, Object> sslConfig = getSslConfig();
+        if (sslConfig != null) {
```
mimaison: This should be done in the SSL class. The base class should not be aware of SSL and just use configurations from the concrete classes

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+@Category(IntegrationTest.class)
+public class MirrorConnectorsIntegrationTest extends MirrorConnectorsIntegrationBaseTest {
+    
+    @Before
+    public void setup() throws InterruptedException {
+        startClusters();
+    }
+
+    @After
```
mimaison: Can we move this method to the base class?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationTest.java
mimaison: Well in `MirrorConnectorsIntegrationSSLTest`, we can override it, but we should put the default implementation in the base class

connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java
```diff
@@ -161,6 +161,11 @@ private void start(int[] brokerPorts, String[] logDirs) {
         producerProps.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers());
         producerProps.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.ByteArraySerializer");
         producerProps.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.ByteArraySerializer");
+        if (sslEnabled()) {
+        	producerProps.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, brokerConfig.get(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG));
+        	producerProps.put(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, ((Password) brokerConfig.get(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG)).value());
```
mimaison: Do we really need to cast to `Password` and call `value()`? Can't we just pass `brokerConfig.get(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG)`?

connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java
```diff
@@ -278,6 +283,14 @@ protected boolean hasState(KafkaServer server, Predicate<BrokerState> desiredSta
             return false;
         }
     }
+    
+    public boolean sslEnabled() {
+        final Object listeners = brokerConfig.get(KafkaConfig$.MODULE$.ListenersProp());
```
mimaison: `brokerConfig` is a `Properties`, so we can call `getProperty()` to get back a `String` directly

connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java
```diff
@@ -278,6 +283,14 @@ protected boolean hasState(KafkaServer server, Predicate<BrokerState> desiredSta
...
+        if (listeners != null && listeners.toString().contains("SSL")) {
```
mimaison: Instead of doing:
```
if (condition) {
  return true;
} else {
  return false;
}
```
You can do:
```
return condition;
```

connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java
```diff
@@ -444,7 +457,11 @@ public Admin createAdminClient() {
         putIfAbsent(props, AUTO_OFFSET_RESET_CONFIG, "earliest");
         putIfAbsent(props, KEY_DESERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.ByteArrayDeserializer");
         putIfAbsent(props, VALUE_DESERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.ByteArrayDeserializer");
-
+        if (sslEnabled()) {
+        	putIfAbsent(props,SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, brokerConfig.get(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG));
+        	putIfAbsent(props,SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, ((Password) brokerConfig.get(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG)).value());
```
mimaison: Same as above, I think we can just pass `brokerConfig.get(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG)`?

connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java
```diff
@@ -161,6 +161,11 @@ private void start(int[] brokerPorts, String[] logDirs) {
         producerProps.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers());
         producerProps.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.ByteArraySerializer");
         producerProps.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.ByteArraySerializer");
+        if (sslEnabled()) {
+        	producerProps.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, brokerConfig.get(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG));
```
mimaison: A few lines in this file (like this one) contain tabs, we use spaces in Kafka

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
mimaison: But why is this needed here? I don't know what the other test is doing but I don't understand why it's used here

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,617 @@
...
+     */
+    private static <T> void waitForConsumingAllRecords(Consumer<T, T> consumer, int numExpectedRecords) 
+            throws InterruptedException {
+        final AtomicInteger totalConsumedRecords = new AtomicInteger(0);
+        waitForCondition(() -> {
+            ConsumerRecords<T, T> records = consumer.poll(Duration.ofMillis(CONSUMER_POLL_TIMEOUT_MS));
+            return numExpectedRecords == totalConsumedRecords.addAndGet(records.count());
+        }, RECORD_CONSUME_DURATION_MS, "Consumer cannot consume all records in time");
+        consumer.commitSync();
+        consumer.close();
```
mimaison: I find it strange that this method closes the consumer it received.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,617 @@
...
+     * retrieve the config value based on the input cluster, topic and config name
+     */
+    private static String getTopicConfig(EmbeddedKafkaCluster cluster, String topic, String configName) {
+        Admin client = cluster.createAdminClient();
+        Collection<ConfigResource> cr =  Collections.singleton(
+                new ConfigResource(ConfigResource.Type.TOPIC, topic)); 
+        try {
+            DescribeConfigsResult configsResult = client.describeConfigs(cr);
+            Config allConfigs = (Config) configsResult.all().get().values().toArray()[0];
+            Iterator<ConfigEntry> configIterator = allConfigs.entries().iterator();
```
mimaison: Can we use `allConfigs.get()` to find the configs we want instead of searching for them?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
mimaison: Just let the Exception flow, that will automatically fail the test

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,617 @@
...
+    
+    private static final int NUM_RECORDS_PER_PARTITION = 10;
+    private static final int NUM_PARTITIONS = 10;
+    private static final int NUM_RECORDS_PRODUCED = NUM_PARTITIONS * NUM_RECORDS_PER_PARTITION;
+    private static final int RECORD_TRANSFER_DURATION_MS = 30_000;
+    private static final int CHECKPOINT_DURATION_MS = 20_000;
+    private static final int RECORD_CONSUME_DURATION_MS = 20_000;
+    private static final int OFFSET_SYNC_DURATION_MS = 30_000;
+    private static final int NUM_WORKERS = 3;
+    private static final int CONSUMER_POLL_TIMEOUT_MS = 500;
```
mimaison: This is always used with `Duration.ofMillis()`. Should we store the `Duration` object directly?

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,617 @@
...
+            } catch (Throwable e) {
+                return false;
+            }
+        }, CHECKPOINT_DURATION_MS, "Offsets not translated downstream to primary cluster.");
+
+        waitForCondition(() -> {
+            try {
+                return primaryClient.remoteConsumerOffsets(consumerGroupName, BACKUP_CLUSTER_ALIAS,
+                    Duration.ofMillis(CHECKPOINT_DURATION_MS)).containsKey(new TopicPartition("test-topic-1", 0));
+            } catch (Throwable e) {
```
mimaison: I understand this was probably already there but we should avoid catching `Throwable`. I'm not sure the try/catch block is necessarily helping in this cases

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,617 @@
...
+        primaryConsumer.poll(Duration.ofMillis(CONSUMER_POLL_TIMEOUT_MS));
+        primaryConsumer.commitAsync();
+
+        assertTrue("Consumer failedover to zero offset.", primaryConsumer.position(new TopicPartition("primary.test-topic-1", 0)) > 0);
+        assertTrue("Consumer failedover beyond expected offset.", primaryConsumer.position(
+            new TopicPartition("primary.test-topic-1", 0)) <= NUM_RECORDS_PRODUCED);
+        assertTrue("Checkpoints were not emitted upstream to primary cluster.", primary.kafka().consume(1,
+            CHECKPOINT_DURATION_MS, "backup.checkpoints.internal").count() > 0);
+
+        primaryConsumer.close();
```
mimaison: Closing a client in the middle of a test can be confusing. I wonder if we could wrap this whole block about `primaryConsumer` in a try with resource block. WDYT?

mimaison (changes requested):
Thanks @ning2008wisc for the updates.
It's looking much much better. There are still a few things that need to be improved to merge but it's almost there.

ning2008wisc:
https://github.com/apache/kafka/pull/9224#discussion_r535179359

I think the `try catch` block is needed for several Exception.

```
/kafka/connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java:41: error: unreported exception IOException; must be caught or declared to be thrown
        sslConfig = TestSslUtils.createSslConfig(false, true, Mode.SERVER, TestUtils.tempFile(), "testCert");

```

```
/kafka/connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java:42: error: unreported exception GeneralSecurityException; must be caught or declared to be thrown
            sslConfig = TestSslUtils.createSslConfig(false, true, Mode.SERVER, TestUtils.tempFile(), "testCert");
```

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
ning2008wisc: Closing the consumer will make the re-use of the same consumer instance in a clean state. I tried to remove `consumer.close();` and it caused test failures.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
ning2008wisc: `allConfigs` is a `java.util.Collection` object and seems does not directly support `get()` https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
ning2008wisc: yeah - I think we can do "try with resource block", updated

rhauch:
@ning2008wisc, @mimaison: Please see https://issues.apache.org/jira/browse/KAFKA-10811 and #9698 for a required fix to the `MirrorConnectorsIntegrationTest` to prevent the `testReplication()` method from prematurely terminating the builds. Please be sure this (or an equivalent) is incorporated into this refactoring.

ning2008wisc:
@rhauch if your pr is merged first, I will be happy to rebase on https://github.com/apache/kafka/pull/9698

connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java
```diff
@@ -278,6 +283,11 @@ protected boolean hasState(KafkaServer server, Predicate<BrokerState> desiredSta
             return false;
         }
     }
+    
+    public boolean sslEnabled() {
+        final String listeners = brokerConfig.getProperty(KafkaConfig$.MODULE$.ListenersProp());
+        return (listeners != null && listeners.contains("SSL")) ? true : false;
```
mimaison: We can return `listeners != null && listeners.contains("SSL")`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
mimaison: Yes but the code that created the consumer should close it. If I call `waitForConsumingAllRecords()`, I'd not expect it to close my consumer instance.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,577 @@
...
+     * retrieve the config value based on the input cluster, topic and config name
+     */
+    private static String getTopicConfig(EmbeddedKafkaCluster cluster, String topic, String configName) 
+        throws Exception {
+        Admin client = cluster.createAdminClient();
+        Collection<ConfigResource> cr =  Collections.singleton(
+                new ConfigResource(ConfigResource.Type.TOPIC, topic)); 
+
+        DescribeConfigsResult configsResult = client.describeConfigs(cr);
+        Config allConfigs = (Config) configsResult.all().get().values().toArray()[0];
```
mimaison, line 482: We can use `Config.get()` to directly access the configuration we want, see https://kafka.apache.org/26/javadoc/org/apache/kafka/clients/admin/Config.html#get-java.lang.String-

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,577 @@
...
+        backupConsumer.close();
+    }
+    
+    /*
+     * launch the connectors on kafka connect cluster and check if they are running
+     */
+    private static void waitUntilMirrorMakerIsRunning(EmbeddedConnectCluster connectCluster, 
+            List<Class<? extends Connector>> connectorClasses, MirrorMakerConfig mm2Config, 
+            String primary, String backup) throws InterruptedException {
+        for (int i = 0; i < connectorClasses.size(); i++) {
```
mimaison: We could use a "for each" loop here, something like:
```
for (Class<? extends Connector> connector : connectorClasses) {
    connectCluster.configureConnector(connector.getSimpleName(), mm2Config.connectorBaseConfig(
        new SourceAndTarget(primary, backup), connector));
}
```

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,577 @@
...
+            String connector = connectorClasses.get(i).getSimpleName();
+            connectCluster.configureConnector(connector, mm2Config.connectorBaseConfig(
+                new SourceAndTarget(primary, backup), connectorClasses.get(i)));
+        }
+        
+        // we wait for the connector and tasks to come up for each connector, so that when we do the
+        // actual testing, we are certain that the tasks are up and running; this will prevent
+        // flaky tests where the connector and tasks didn't start up in time for the tests to be run
+        List<String> connectorNames = connectorClasses.stream().map(x -> x.getSimpleName())
```
mimaison: This intermediate `List` is not really useful. We could just change the loop below to iterate over the connector classes and call `getSimpleName()` on each of them

mimaison: I'm assuming it's worth keeping this 2nd loop separate from the first one for performance reasons. At first glance it looks strange to iterate over the same collection twice in a row

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
mimaison: This still needs to be addressed

mimaison (changes requested):
Thanks for the update @ning2008wisc. You've now addressed the most major issues. I've left a few more comments suggesting small improvements. A
Can you also rebase on trunk to resolve the conflict?

ning2008wisc:
Hello @mimaison  I addressed all your of comments, please take the final review.

rhauch:
@ning2008wisc, https://issues.apache.org/jira/browse/KAFKA-10811 and #9698 have been merged (to multiple branches).

ning2008wisc:
@rhauch I incorporated https://github.com/apache/kafka/pull/9698 in my latest version here

mimaison:
@ning2008wisc There are a few checkstyle failures:
```
> Task :connect:runtime:checkstyleTest
[ant:checkstyle] [ERROR] /Users/mickael/github-ws/kafka/connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java:458:30: ',' is not followed by whitespace. [WhitespaceAfter]
[ant:checkstyle] [ERROR] /Users/mickael/github-ws/kafka/connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java:459:30: ',' is not followed by whitespace. [WhitespaceAfter]
[ant:checkstyle] [ERROR] /Users/mickael/github-ws/kafka/connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java:460:30: ',' is not followed by whitespace. [WhitespaceAfter]
```
You can run `./gradlew connect:runtime:test` to reproduce

ning2008wisc:
@mimaison thanks for pointing out. Fixed the checkstyle, sorry for just running the check for `connect/mirror`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationSSLTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.experimental.categories.Category;
+import org.junit.Before;
+
+/**
+ * Tests MM2 replication with SSL enabled at backup kafka cluster
+ */
+@Category(IntegrationTest.class)
+public class MirrorConnectorsIntegrationSSLTest extends MirrorConnectorsIntegrationBaseTest {
+    
+    private static Map<String, Object> sslConfig;
```
mimaison: We don't need this field, this could be a local in `startClusters()`

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/integration/MirrorConnectorsIntegrationBaseTest.java
```diff
@@ -0,0 +1,599 @@
...
+    private static final int NUM_RECORDS_PER_PARTITION = 10;
+    private static final int NUM_PARTITIONS = 10;
+    private static final int NUM_RECORDS_PRODUCED = NUM_PARTITIONS * NUM_RECORDS_PER_PARTITION;
+    private static final int RECORD_TRANSFER_DURATION_MS = 30_000;
+    private static final int CHECKPOINT_DURATION_MS = 20_000;
+    private static final int RECORD_CONSUME_DURATION_MS = 20_000;
+    private static final int OFFSET_SYNC_DURATION_MS = 30_000;
+    private static final int NUM_WORKERS = 3;
+    private static final Duration CONSUMER_POLL_TIMEOUT_MS = Duration.ofMillis(500);
+    private static final int BROKER_RESTART_TIMEOUT_MS = 10_000;
```
mimaison: This field is unused

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
```diff
@@ -1,595 +0,0 @@
...
-            } catch (Throwable e) {
-                return false;
-            }
-        }, CHECKPOINT_DURATION_MS, "Offsets not translated upstream to primary cluster.");
-
-        Map<TopicPartition, OffsetAndMetadata> primaryOffsets = primaryClient.remoteConsumerOffsets(consumerGroupName, "backup",
-                Duration.ofMillis(CHECKPOINT_DURATION_MS));
-
-        primaryClient.close();
-        backupClient.close();
```
mimaison, line 346: It looks like we lost these lines and maybe some others too. These came from https://github.com/apache/kafka/pull/9698. Can you make sure we don't miss any existing logic?

mimaison (changes requested):
Thanks again for the updates! It looks like we lost some logic that was added recently in `MirrorConnectorsIntegrationTest.java `.
Can you take a look and make sure we don't lose any logic?

ning2008wisc:
@mimaison thanks so much for your careful review. I checked https://github.com/apache/kafka/pull/9698 again and make sure we do not miss anything at this moment.

ning2008wisc:
@mimaison thanks so much for all your efforts on reviewing. Really appreciated if you may have time before end of this year to do 1-2 final reviews to merge this PR.

mimaison (approved):
Thanks @ning2008wisc for all the updates! LGTM

ning2008wisc:
Thanks @mimaison  so much for your multiple rounds of valuable feedback and comments which greatly improving the code quality of this PR

PR #9225, opened 2020-08-27
KAFKA-10431: Sequential selection of payload in ProducerPerformance

huxihx:
https://issues.apache.org/jira/browse/KAFKA-10431

Currently, ProducerPerformance randomly selects a payload from the payload file when sending message. This might cause the some payloads being selected more times than others. It's better to do sequential selection of payloads.

comment to ping reviewers.*

### DIFF
--- a/tools/src/main/java/org/apache/kafka/tools/ProducerPerformance.java
+++ b/tools/src/main/java/org/apache/kafka/tools/ProducerPerformance.java
@@ -126,6 +126,8 @@ public class ProducerPerformance {
 
             int currentTransactionSize = 0;
             long transactionStartTime = 0;
+            int currentIndex = 0;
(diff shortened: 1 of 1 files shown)


huxihx:
@ijuma Please review this patch. Thanks.

zaahirlaher:
@huxihx  Currently the --num-records is the actual number of records sent, which are randomly selected from the payload file. 
What if  --num-records < the entries in the payload file? Would you then start reading from the start of the payload file again? This would also result in duplicates.

huxihx:
@zaahirlaher Duplicates could not be avoided with the randomly-selecting strategy as well. Besides, it could happen that some events would not be chosen to be sent at all. Looping all the messages from the start of the file is a better message-selection method.

PR #9226, opened 2020-08-27
KAFKA-10444: Jenkinsfile testing

mumrah:
This PR adds a Jenkinsfile to the build to replace the existing jenkins.sh script.

The build makes use of the `parallel` directive to run the JDK 8, 11, and 14 builds in parallel. Each build will report separate statuses here in Github like:

image

Only users with write access to the repository (i.e., committers) will be able to modify Jenkinsfile through pull requests, otherwise the build will use the Jenkinsfile on the target branch.

Here's a build result showing our typical flaky tests https://ci-builds.apache.org/job/Kafka/job/kafka-pr/view/change-requests/job/PR-9226/57/#showFailuresLink

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,16 +34,12 @@ void setBuildStatus(String context, String message, String state) {
     ]);
 }
 
-void doValidation(String scalaVersion) {
-  echo "Scala Version: ${scalaVersion}"
-  environment {
-    SCALA_VERSION = scalaVersion
-  }
+def doValidation() {
   sh '''
     ./gradlew -PscalaVersion=$SCALA_VERSION clean compileJava compileScala compileTestJava compileTestScala \
-      spotlessScalaCheck checkstyleMain checkstyleTest spotbugsMain rat \
-      --profile --no-daemon --continue -PxmlSpotBugsReport=true \"$@\" \
-      || { echo 'Validation steps failed'; exit 1; }
+        spotlessScalaCheck checkstyleMain checkstyleTest spotbugsMain rat \
+        --profile --no-daemon --continue -PxmlSpotBugsReport=true \"$@\" \
+        || { echo 'Validation steps failed'; exit 1; }
   '''
 }
 
@@ -56,9 +52,12 @@ pipeline {
 	  tools {
 	    jdk 'JDK 1.8 (latest)'
 	  }
+          environment {
+            SCALA_VERSION=2.12
+          }
 	  steps {
 	    sh 'gradle -version'
-	    doValidation('2.12')
+	    doValidation()
 	  }
 	}
 
@@ -66,9 +65,12 @@ pipeline {
 	  tools {
 	    jdk 'JDK 11 (latest)'
 	  }
+          environment {
+            SCALA_VERSION=2.13
+          }
 	  steps {
 	    sh 'gradle -version'
-	    doValidation('2.13')
+	    doValidation()
 	    // setBuildStatus("continuous-integration/jenkins/test-check-1", "Check is running", "PENDING")
 	  }
 	}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -80,34 +80,32 @@ pipeline {
     stage('Build') {
       parallel {
 	stage('JDK 8') {
-          node {
-	    ools {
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,200 @@
...
+ *  limitations under the License.
+ *
+ */
+
+def setupGradle() {
+  // Delete gradle cache to workaround cache corruption bugs, see KAFKA-3167
+  dir('.gradle') {
+    deleteDir()
+  }
+  sh './gradlew -version'
```
ijuma, line 25: Is this informational so that we know the gradle version? Or is there more to it?

Jenkinsfile
```diff
@@ -0,0 +1,200 @@
...
+
+def doValidation() {
+  try {
+    sh '''
+      ./gradlew -PscalaVersion=$SCALA_VERSION clean compileJava compileScala compileTestJava compileTestScala \
+          spotlessScalaCheck checkstyleMain checkstyleTest spotbugsMain rat \
+          --profile --no-daemon --continue -PxmlSpotBugsReport=true \"$@\"
+    '''
+  } catch(err) {
+    error('Validation checks failed, aborting this build')
```
ijuma: Should we include `err` in the output somehow? Or is that not needed?

ijuma (commented):
Thanks for the PR. Nice work! A few comments below.

Jenkinsfile
mumrah: Yes informational to see the gradle version and jdk that's running

Jenkinsfile
mumrah: I think the err here will be like an ExectuionException -- i'm going to try something different than this try/catch actually

Jenkinsfile
```diff
@@ -0,0 +1,200 @@
...
+
+def doTest() {
+  try {
+    sh '''
+      ./gradlew -PscalaVersion=$SCALA_VERSION unitTest integrationTest \
+          --profile --no-daemon --continue -PtestLoggingEvents=started,passed,skipped,failed "$@"
+    '''
+  } catch(err) {
+    echo 'Some tests failed, marking this build UNSTABLE'
+    currentBuild.result = 'UNSTABLE'
```
lbradstreet: I think this may mark a build as UNSTABLE if one of gradle's executors exited. We're trying to achieve the property that the build won't fail completely if one of the tests fails, right? If so, I think the best way is to supply `-PignoreFailures=true` at https://github.com/apache/kafka/pull/9226/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R44, and add:

```
  task integrationTest(type: Test, dependsOn: compileJava) {                                                                                                                                                                         
    ignoreFailures = userIgnoreFailures                                                                                                                                                                                              
```

```
                                                                                                                                                                                                                                     
  task unitTest(type: Test, dependsOn: compileJava) {                                                                                                                                                                                
    ignoreFailures = userIgnoreFailures                                                                                                                                                                                              
```

```
ext {                                                                                                                                                                                                                                
  userIgnoreFailures = project.hasProperty('ignoreFailures') ? ignoreFailures : false                                                                                                                                                
```

to build.gradle.

mumrah: I see, so in this case we'll let the `junit` directive set the build to unstable?

lbradstreet: Correct, it'll be picked up when the test output files are picked up.

mumrah: So, this seems to have caused us to lose which build had which results in the summary. 

Before we had:
image

Now we have: 
image

I'll try moving the `junit` call inside the actual stage and see if that helps

mumrah:
Example failed build with a compile error: https://ci-builds.apache.org/job/Kafka/job/kafka-pr/view/change-requests/job/PR-9226/67

One with a rat failure: https://ci-builds.apache.org/job/Kafka/job/kafka-pr/view/change-requests/job/PR-9226/66/

Jenkinsfile
mumrah: Moving the junit into the stage seemed to work

mumrah:
Closing in favor of #9238 for a clean slate

PR #9227, opened 2020-08-27
KAFKA-10439: Connect's Values to parse BigInteger as Decimal with zero scale.

avocader:
The `org.apache.kafka.connect.data.Values#parse` method parses integers, which are larger than `Long.MAX_VALUE` as `double` with `Schema.FLOAT64_SCHEMA`.

That means we are losing precision for these larger integers.

For example:
```
SchemaAndValue schemaAndValue = Values.parseString("9223372036854775808");
```
returns:
```
SchemaAndValue{schema=Schema{FLOAT64}, value=9.223372036854776E18}
```
Also, this method parses values, that can be parsed as `FLOAT32` to `FLOAT64`.

This PR changes parsing logic, to use `FLOAT32`/`FLOAT64` for numbers that don't have and fraction part(`decimal.scale()!=0`) only, and use an arbitrary-precision `org.apache.kafka.connect.data.Decimal` otherwise.
Also, it updates the method to parse numbers, that can be represented as `float` to `FLOAT64`.

Added unit tests, that cover parsing `BigInteger`, `Byte`, `Short`, `Integer`, `Long`, `Float`, `Double` types.

### DIFF
--- a/connect/api/src/main/java/org/apache/kafka/connect/data/Values.java
+++ b/connect/api/src/main/java/org/apache/kafka/connect/data/Values.java
@@ -943,8 +943,14 @@ public class Values {
                 } catch (ArithmeticException e) {
                     // continue
                 }
+                float fValue = decimal.floatValue();
+                if (fValue != Float.NEGATIVE_INFINITY && fValue != Float.POSITIVE_INFINITY
+                    && decimal.scale() != 0) {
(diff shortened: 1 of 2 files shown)


kkonstantine:
ok to test

kkonstantine:
Closing temporarily to attempt to trigger a jenkins build.

PR #9232, opened 2020-08-31
KAFKA-9924: Add remaining property-based RocksDB metrics as described in KIP-607

cadonna:
This commit adds the remaining property-based RocksDB metrics as described in KIP-607, except for num-entries-active-mem-table, which was added in PR #9177.


cadonna:
Call for review: @guozhangwang @vvcephei

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/RocksDBMetricsRecorder.java
```diff
@@ -150,14 +176,55 @@ private void verifyStatistics(final String segmentName, final Statistics statist
...
+                                               final Statistics statistics) {
+        for (final DbAndCacheAndStatistics valueProviders : storeToValueProviders.values()) {
+            verifyIfSomeAreNull(segmentName, statistics, valueProviders.statistics, "statistics");
+            verifyIfSomeAreNull(segmentName, cache, valueProviders.cache, "cache");
+            if (db == valueProviders.db) {
+                throw new IllegalStateException("DB instance for store " + segmentName + " of task " + taskId +
+                    " was already added for another segment as a value provider. This is a bug in Kafka Streams. " +
+                    "Please open a bug report under https://issues.apache.org/jira/projects/KAFKA/issues");
+            }
+            if (storeToValueProviders.size() == 1 && cache != valueProviders.cache) {
```
guozhangwang, line 182: Hmm, why we need the second condition to determine `singleCache = false` here?

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/RocksDBMetricsRecorder.java
```diff
@@ -150,14 +176,55 @@ private void verifyStatistics(final String segmentName, final Statistics statist
...
+                singleCache = false;
+            } else if (singleCache && cache != valueProviders.cache || !singleCache && cache == valueProviders.cache) {
+                throw new IllegalStateException("Caches for store " + storeName + " of task " + taskId +
+                    " are either not all distinct or do not all refer to the same cache. This is a bug in Kafka Streams. " +
+                    "Please open a bug report under https://issues.apache.org/jira/projects/KAFKA/issues");
+            }
+        }
+    }
+
+    private void verifyIfSomeAreNull(final String segmentName,
```
guozhangwang: nit: `verifyConsistentSegmentValueProviders`?

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/RocksDBMetricsRecorder.java
```diff
@@ -174,22 +241,163 @@ private void initSensors(final StreamsMetricsImpl streamsMetrics, final RocksDBM
...
             BigInteger result = BigInteger.valueOf(0);
             for (final DbAndCacheAndStatistics valueProvider : storeToValueProviders.values()) {
                 try {
                     // values of RocksDB properties are of type unsigned long in C++, i.e., in Java we need to use
                     // BigInteger and construct the object from the byte representation of the value
                     result = result.add(new BigInteger(1, longToBytes(
-                        valueProvider.db.getAggregatedLongProperty(ROCKSDB_PROPERTIES_PREFIX + NUMBER_OF_ENTRIES_ACTIVE_MEMTABLE))));
+                        valueProvider.db.getAggregatedLongProperty(ROCKSDB_PROPERTIES_PREFIX + propertyName)
+                    )));
+                } catch (final RocksDBException e) {
```
guozhangwang, line 351: Is this a piggy-backed fix to wrap RocksDBException here?

guozhangwang (commented):
A couple of questions, otherwise LGTM

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/RocksDBMetricsRecorder.java
cadonna: When RocksDB uses the memory bounded configuration, all RocksDB instances use the same cache, i.e., the reference to the same cache. 
At this point in the code the value providers for the given segment with the cache `cache` has not been added yet to the value providers used in this recorder. To understand if different caches are used for different segments (i.e., `singleCache = false`), it is not enough to just check if only one single cache has been already added, we also need to check if the already added cache (i.e., `valueProviders.cache`) is different from the cache to add (`cache`).

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/RocksDBMetricsRecorder.java
cadonna: I will rename it to `verifyConsistencyOfValueProvidersAcrossSegments()`. WDYT?

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/RocksDBMetricsRecorder.java
cadonna: Actually, I just followed the pattern found in `RocksDBStore`.

PR #9235, opened 2020-08-31
KAFKA-10449: Add some important parameter desc in connect-distributed.properties

tinawenqiao:
…rties.

comment to ping reviewers.*

### DIFF
--- a/config/connect-distributed.properties
+++ b/config/connect-distributed.properties
@@ -66,14 +66,24 @@ status.storage.replication.factor=1
 # Flush much faster than normal, which is useful for testing/debugging
 offset.flush.interval.ms=10000
 
+# List of comma-separated URIs the REST API will listen on. The supported protocols are HTTP and HTTPS.
+# Specify hostname as 0.0.0.0 to bind to all interfaces.
+# Leave hostname empty to bind to default interface.
+# After this parameter is set, rest.host.name/port will not take effect.
+# Examples of legal listener lists: HTTP://myhost:8083,HTTPS://myhost:8084"
+#listeners=HTTP://:8083
+
 # These are provided to inform the user about the presence of the REST host and port configs 
 # Hostname & Port for the REST API to listen on. If this is set, it will bind to the interface used to listen to requests.
+# DEPRECATED As of 1.1.0: only used when listeners is not set. Use listeners instead.
 #rest.host.name=
 #rest.port=8083
 
 # The Hostname & Port that will be given out to other workers to connect to i.e. URLs that are routable from other servers.
+# If not set, it uses the value for "listeners" or the rest.host.name/port if configured.
 #rest.advertised.host.name=
 #rest.advertised.port=
+#rest.advertised.listener=
 
 # Set to a list of filesystem paths separated by commas (,) to enable class loading isolation for plugins
 # (connectors, converters, transformations). The list should consist of top level directories that include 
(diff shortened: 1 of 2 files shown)


tinawenqiao:
In WokerConfig.java we found that REST_HOST_NAME_CONFIG(rest.host.name) and REST_PORT_CONFIG(rest.port) were deprecated. And some new configuration parameters are introduced such as LISTENERS_CONFIG(listeners), REST_ADVERTISED_LISTENER_CONFIG(rest.advertised.listener),ADMIN_LISTENERS_CONFIG(admin.listeners) but not list in the sample conf file.

mimaison:
Thanks for the PR @tinawenqiao.

It would be nice to also mention `rest.port` and `rest.host.name` are deprecated in their descriptions. Could you add something like:
`DEPRECATED: only used when listeners is not set. Use listeners instead. ` to their descriptions in `WorkerConfigs`?

Thanks

tinawenqiao:
> Thanks for the PR @tinawenqiao.
> 
> It would be nice to also mention `rest.port` and `rest.host.name` are deprecated in their descriptions. Could you add something like:
> `DEPRECATED: only used when listeners is not set. Use listeners instead. ` to their descriptions in `WorkerConfigs`?
> 
> Thanks

Thanks for your notice. A new patch is ready.

mimaison:
@tinawenqiao It looks like you made a mistake when rebasing your branch, can you ensure the PR only contains your diff? Thanks

tinawenqiao:
> @tinawenqiao It looks like you made a mistake when rebasing your branch, can you ensure the PR only contains your diff? Thanks

Sorry for the mistake. A clean patch is submit. Is it ok now?

mimaison:
Thanks for the update, it looks much better.

I should have been clearer, I meant to update the documentation text of these settings, not just the comments in `WorkerConfigs`.

For example in https://github.com/apache/kafka/pull/9235/files#diff-5664ddd985065623bab1870e031ce4f3fcfee0bc4951e41b6b0e1584cd32fa2dR153, we could do something like:
```java
private static final String REST_HOST_NAME_DOC
    = "Hostname for the REST API. If this is set, it will only bind to this interface. Deprecated, only used when listeners is not set. Use listeners instead.";
```

tinawenqiao:
> Thanks for the update, it looks much better.
> 
> I should have been clearer, I meant to update the documentation text of these settings, not just the comments in `WorkerConfigs`.
> 
> For example in https://github.com/apache/kafka/pull/9235/files#diff-5664ddd985065623bab1870e031ce4f3fcfee0bc4951e41b6b0e1584cd32fa2dR153, we could do something like:
> 
> ```java
> private static final String REST_HOST_NAME_DOC
>     = "Hostname for the REST API. If this is set, it will only bind to this interface. Deprecated, only used when listeners is not set. Use listeners instead.";
> ```

Doc desc has been added.

tinawenqiao:
> > Thanks for the update, it looks much better.
> > I should have been clearer, I meant to update the documentation text of these settings, not just the comments in `WorkerConfigs`.
> > For example in https://github.com/apache/kafka/pull/9235/files#diff-5664ddd985065623bab1870e031ce4f3fcfee0bc4951e41b6b0e1584cd32fa2dR153, we could do something like:
> > ```java
> > private static final String REST_HOST_NAME_DOC
> >     = "Hostname for the REST API. If this is set, it will only bind to this interface. Deprecated, only used when listeners is not set. Use listeners instead.";
> > ```
> 
> Doc desc has been added.

@mimaison is it ok now?

mimaison (approved):

PR #9236, opened 2020-09-01
MINOR: Log warn message with details when there's kerberos login issue

cnZach:
…though we will still retry

Currently, we just capture the exception and retry later, we can't figure out what happened at runtime, it would be helpful to log a warn message with details.

Minor logging change, so no extra test is required. 

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/security/kerberos/KerberosLogin.java
+++ b/clients/src/main/java/org/apache/kafka/common/security/kerberos/KerberosLogin.java
@@ -238,7 +238,7 @@ public class KerberosLogin extends AbstractLogin {
                             break;
                         } catch (LoginException le) {
                             if (retry > 0) {
-                                log.warn("[Principal={}]: Error when trying to re-Login, but will retry ", principal, e);
+                                log.warn("[Principal={}]: Error when trying to re-Login, but will retry ", principal, le);
                                 --retry;
                                 // sleep for 10 seconds.
                                 try {
--- a/clients/src/main/java/org/apache/kafka/common/security/kerberos/KerberosLogin.java
+++ b/clients/src/main/java/org/apache/kafka/common/security/kerberos/KerberosLogin.java
@@ -238,7 +238,7 @@ public class KerberosLogin extends AbstractLogin {
                             break;
                         } catch (LoginException le) {
                             if (retry > 0) {
-                                log.warn("[Principal={}]: Error when trying to  re-Login, but will retry ", principal, e);
+                                log.warn("[Principal={}]: Error when trying to re-Login, but will retry ", principal, e);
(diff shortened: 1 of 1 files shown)


cnZach:
Hi @omkreddy or @rajinisivaram , what do you think about this? I just want to print out the error. It will be helpful for troubleshooting kerberos login issue.

clients/src/main/java/org/apache/kafka/common/security/kerberos/KerberosLogin.java
```diff
@@ -237,6 +238,7 @@ public LoginContext login() throws LoginException {
                             break;
                         } catch (LoginException le) {
                             if (retry > 0) {
+                                log.warn("[Principal={}]: Error when trying to re-Login, but will retry ", principal, e);
```
omkreddy: e => le ?

cnZach: oops, yes, corrected.

omkreddy:
ok to test

cnZach:
Not sure what happened... Can you trigger the test again?

omkreddy:
@cnZach  can you rebase this PR against trunk? That will trigger the build.

cnZach:
> @cnZach can you rebase this PR against trunk? That will trigger the build.

Hi @omkreddy , do I need to change anything for this PR?

clients/src/main/java/org/apache/kafka/common/security/kerberos/KerberosLogin.java
```diff
@@ -213,6 +213,7 @@ public LoginContext login() throws LoginException {
                             break;
                         } catch (Exception e) {
                             if (retry > 0) {
+                                log.warn("[Principal={}]: Error when trying to renew with TicketCache, but will retry ", principal, e);
```
rajinisivaram, line 216: @cnZach Looks like the intention was to print out the exception with stack trace? You would want to use:
```
public void warn(String msg, Throwable t);
```

cnZach: Hi @rajinisivaram , do you mean we should use  `log.warn(s"[Principal=$principal]: Error when trying to renew with TicketCache, but will retry ", e); ` ?

omkreddy (approved):
@cnZach Thanks for the PR. LGTM.

clients/src/main/java/org/apache/kafka/common/security/kerberos/KerberosLogin.java
omkreddy: slf4j logger methods logs exception if the last argument is throwable. so this is fine.

PR #9238, opened 2020-09-01
KAFKA-10444: Configure PR builds via Jenkinsfile

mumrah:
This PR adds a Jenkinsfile to the build to replace the existing jenkins.sh script.

The build makes use of the parallel directive to run the JDK 8, 11, and 15 builds in parallel. Each build will report status to PRs in Github including which tests failed.

Only users with write access to the repository (i.e., committers) will be able to modify Jenkinsfile through pull requests, otherwise the build will use the Jenkinsfile on the target branch.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,169 @@
+/*
+ *
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


ijuma (approved):
Great work. The build is well structured and easy to understand and it does what we want:
1. Testing for multiple scala/java versions.
2. Each scala/java combination is executed in parallel.
3. Separate status report for each scala/java version.
4. The validate (vs test) stage is done first and fails fast (although I think it only really does that if every scala/java version fails in the validate step).
5. Build failures or JVM crashes result in a `Failed` build while test failures result in an `Unstable` build.

Looks good to me if the tests pass. https://github.com/apache/kafka/pull/9226 has various tests that were performed and what the results look like for various cases.

Jenkinsfile
```diff
@@ -0,0 +1,169 @@
...
+            timestamps()
+          }
+          environment {
+            SCALA_VERSION=2.13
+          }
+          steps {
+            setupGradle()
+            doValidation()
+            doTest()
+            echo 'Skipping Kafka Streams archetype test for Java 14'
```
chia7712: for Java 14 -> for Java 15

build.gradle
```diff
@@ -336,6 +337,7 @@ subprojects {
 
   task integrationTest(type: Test, dependsOn: compileJava) {
     maxParallelForks = userMaxForks ?: Runtime.runtime.availableProcessors()
+    ignoreFailures = userIgnoreFailures
```
chia7712, line 341: Should this new option be supported by *test task*?

PR #9247, opened 2020-09-02
KAFKA-10362: When resuming Streams active task with EOS, the checkpoint file is deleted

DOJI45:
What was the issue?
https://issues.apache.org/jira/browse/KAFKA-10362 

Deleted the checkpoint file before the transition from SUSPENDED state to RESTORING state


DOJI45:
@guozhangwang request you to please review PR

guozhangwang:
test this please

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorStateManager.java
```diff
@@ -662,4 +662,10 @@ public TopicPartition registeredChangelogPartitionFor(final String storeName) {
     public String changelogFor(final String storeName) {
         return storeToChangelogTopic.get(storeName);
     }
+
+    public void deleteCheckPointFile() throws IOException {
```
guozhangwang: nit: I'd suggest we just inline this function inside `StreamTask` since 1) this is only triggered with EOS enabled, and its name `deleteCheckPointFile` maybe a bit misleading, and 2) it is a very simply function anyways.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -332,6 +332,15 @@ public void resume() {
             case SUSPENDED:
                 // just transit the state without any logical changes: suspended and restoring states
                 // are not actually any different for inner modules
+
+                // Deleting checkpoint file before transition to RESTORING state (KAFKA-10362)
+                try {
+                    stateMgr.deleteCheckPointFile();
+                    log.debug("Deleted check point file");
```
guozhangwang: nit: `log.debug("Deleted check point file upon resuming with EOS enabled");`

guozhangwang (commented):
@DOJI45 Thanks for the PR! Could you also add a unit test to complete the coverage? You can find some examples in StreamTaskTest or ProcessorStateManagerTest

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorStateManager.java
DOJI45: Hi @guozhangwang 

I agree with you that name `deleteCheckPointFile` is a bit misleading, i think we can have a better name ( please suggest a better name :) )

The reasons I had written a new method inside `ProcessorStateManager` is because
-  I felt that, logically; deleting the checkpoint file operation should be under `ProcessorStateManager` as deleting file comes under state management
-  If I have to write this inline, i will have to import `checkpointFile` and `eosEnabled` into `StreamTask` class; but these were already imported in `ProcessorStateManager`, so I created a new method in `ProcessorStateManager`

Please suggest how to go about it.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
DOJI45: I will change the log message

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorStateManager.java
guozhangwang: SGTM. We can keep it as is then.

guozhangwang:
> @DOJI45 Thanks for the PR! Could you also add a unit test to complete the coverage? You can find some examples in StreamTaskTest or ProcessorStateManagerTest

Seems this comment is not addressed yet?

DOJI45:
> > @DOJI45 Thanks for the PR! Could you also add a unit test to complete the coverage? You can find some examples in StreamTaskTest or ProcessorStateManagerTest
> 
> Seems this comment is not addressed yet?

Hi @guozhangwang 

I wrote unit test cases in `ProcessorStateManagerTest` for the new method that I have added

Not sure if I can write unit test case in `StreamTaskTest` class, because functionality wise i won't be able to validate anything, as deleting checkpoint file is tested in `ProcessorStateManagerTest`, and coverage is taken care by the existing unit test cases.

Please suggest how to proceed?

guozhangwang:
Merged to trunk, thanks @DOJI45

ableegoldman:
Cherrypicked to 2.6 for 2.6.2 release

PR #9249, opened 2020-09-03
KAFKA-10458: Add update quota for TokenBucket registered with Sensor

apovzner:
For Rate() metric with quota config, we update quota by updating config of KafkaMetric. However, it is not enough for TokenBucket, because it uses quota config on record() to properly calculate the number of tokens. Sensor passes config stored in the corresponding StatAndConfig, which currently never changes. This means that after updating quota via KafkaMetric.config, which is our current and only method, Sensor would record the value using old quota but then measure the value to check for quota violation using the new quota value. This PR adds update method to Sensor that properly updates quota for TokenBucket.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
@@ -53,10 +53,18 @@ public final class Sensor {
     private final Object metricLock;
 
     private static class StatAndConfig {
+        public final MetricName metricName;
         public final Stat stat;
-        public final MetricConfig config;
+        public MetricConfig config;
 
+        // used for a compound stat
         StatAndConfig(Stat stat, MetricConfig config) {
+            this(stat, config, null);
+        }
+
+        // used for measurable stat to enable updating metric config associated with the stat
+        StatAndConfig(Stat stat, MetricConfig config, MetricName metricName) {
+            this.metricName = metricName;
             this.stat = stat;
             this.config = config;
         }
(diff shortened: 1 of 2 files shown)


dajac:
@apovzner Thanks for the PR. I will review it asap.

clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
```diff
@@ -324,11 +332,35 @@ public synchronized boolean add(final MetricName metricName, final MeasurableSta
...
         }
     }
 
+    /**
+     * Update config of a measurable stat and metric registered with this sensor that corresponds to a given metric name
+     * @param metricName The name of the metric to update
+     * @param config     New configuration for this metric
+     * @return true if config was updated, false if sensor expired or metric is not registered with the sensor
+     */
+    public synchronized boolean update(final MetricName metricName, final MetricConfig config) {
```
gwenshap, line 346: nit: The method signature for StatAndConfig has MetricConfig parameter first and MetricName second. This one has the reverse order.

dajac (commented):
@apovzner Thanks for the PR and for spotting that gap! That is definitely a mistake of mine.

I am not fully convinced by the approach that you propose here. My main concern is that the `ClientQuotaManager` relies on the `KafkaMetric` to update the quota when required. I have got the impression that moving from using `KafkaMetric` to using `Sensor` may be doable but I am not 100% sure. Have you already looked at how we could use that new update method from that part of the code?

If we want to proceed with this, I would be great to cover the migration of `ClientQuotaManager` as well in order to have an holistic view of the solution. We should also remove the `config` method in the `KafkaMetric` class.

dajac:
@apovzner I have looked at this further and I have thought about another approach. I have opened a PR for it: https://github.com/apache/kafka/pull/9272. I hope that you don't mind. Could you take a look so we can discuss both approaches and decide what is best?

apovzner:
@dajac I like your approach better. I will close this PR.

PR #9250, opened 2020-09-04
KAFKA-10461 The config of closing heartbeat is invalid.

jiweiautohome:
https://issues.apache.org/jira/projects/KAFKA/issues/KAFKA-10461?filter=allissues

### DIFF
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorHeartbeatTask.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorHeartbeatTask.java
@@ -63,6 +63,9 @@ public class MirrorHeartbeatTask extends SourceTask {
 
     @Override
     public List<SourceRecord> poll() throws InterruptedException {
+        if (interval.toMillis < 0) {
+            return null;
+        }
         // pause to throttle, unless we've stopped
         if (stopped.await(interval.toMillis(), TimeUnit.MILLISECONDS)) {
             // SourceWorkerTask expects non-zero batches or null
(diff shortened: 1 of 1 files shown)


chia7712:
Could you add unit test?

chia7712:
https://github.com/apache/kafka/blob/trunk/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorHeartbeatConnector.java#L64

```
    @Override
    public List<Map<String, String>> taskConfigs(int maxTasks) {
        // if the heartbeats emission is disabled by setting `emit.heartbeats.enabled` to `false`,
        // the interval heartbeat emission will be negative and no `MirrorHeartbeatTask` will be created
        if (config.emitHeartbeatsInterval().isNegative()) {
            return Collections.emptyList();
        }
        // just need a single task
        return Collections.singletonList(config.originalsStrings());
    }
```

It seems that no task will be created if ```emit.heartbeats.enabled``` is set to false. Do you observe something weird?

jiweiautohome:
> Could you add unit test?

The version of kafka we using is 2.4. It's fixed in 2.5.
https://github.com/apache/kafka/blob/16ec1793d53700623c9cb43e711f585aafd44dd4/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorHeartbeatConnector.java#L64

jiweiautohome:
The version of kafka we using is 2.4. It's fixed in 2.5.
https://github.com/apache/kafka/blob/5fa66b3b8c43c4e7de591704581de88fd911a705/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorHeartbeatConnector.java#L54

PR #9252, opened 2020-09-04
KAFKA-10241: Add test for message compatibility

tombentley:
This test compares the current working tree version of protocol/message json
files with the same versions of those files as released in previous versions
of Kafka (and in git HEAD). As such it can detect when a message format is 
changed in an incompabile way, for example by adding a field to an existing 
API version. The verification implements all the rules mentioned in the 
protocol README.md plus a few others.

The test is factored into an abstract test class, which means it can be used in
other places where message JSON is used. I added a test for Kafka Streams,
for example.

Because the test works by looking at git tags it wouldn't be robust to certain
refactorings (e.g. changing the directory in which the message JSON files
reside). It also currently doesn't cope with refactoring field names.


tombentley:
@abbccdda please could you take a look at this, since you opened https://issues.apache.org/jira/browse/KAFKA-10241. The test currently fails because `IncrementalAlterConfigsResponse` changed the name of a field between Kafka 2.3.1 and 2.4.0. That would be easy to address if you felt that this test was along the right lines.

tombentley:
@abbccdda or perhaps @cmccabe or @dajac any chance one of you could take a look?

PR #9253, opened 2020-09-04
KAFKA-10366 & KAFKA-9649: Implement KIP-659 to allow TimeWindowedDeserializer and TimeWindowedSerde to handle window size

lct45:
See KIP details and discussions here: https://cwiki.apache.org/confluence/display/KAFKA/KIP-659%3A+Improve+TimeWindowedDeserializer+and+TimeWindowedSerde+to+handle+window+size

Deprecates methods that allow users to skip setting a window size when one is needed. Adds a window size streams config to allow the `timeWindowedDeserializer` to calculate window end time.


streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedDeserializer.java
```diff
@@ -61,6 +61,21 @@ public Long getWindowSize() {
     @SuppressWarnings("unchecked")
     @Override
     public void configure(final Map<String, ?> configs, final boolean isKey) {
+        //check if the config is set and the window size is already set from the constructor
+        final Long configWindowSize;
+        if (configs.get(StreamsConfig.WINDOW_SIZE_MS_CONFIG) instanceof String) {
```
lct45, line 66: The console consumer made this check necessary - if there's a simpler way to do this lmk

lct45:
call for review @ableegoldman @cadonna @wcarlson5

streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedDeserializer.java
```diff
@@ -33,22 +33,22 @@
...
     // Default constructor needed by Kafka
     public TimeWindowedDeserializer() {
-        this(null, Long.MAX_VALUE);
+        this(null, null);
     }
 
-    // TODO: fix this part as last bits of KAFKA-4468
+    @Deprecated
     public TimeWindowedDeserializer(final Deserializer<T> inner) {
         this(inner, Long.MAX_VALUE);
```
wcarlson5, line 48: Do you want to change this from `Long.MAX_VALUE` as well?

streams/src/test/java/org/apache/kafka/streams/integration/KStreamAggregationIntegrationTest.java
```diff
@@ -327,7 +329,7 @@ public void shouldAggregateWindowed() throws Exception {
         startStreams();
 
         final List<KeyValueTimestamp<Windowed<String>, Integer>> windowedMessages = receiveMessagesWithTimestamp(
-            new TimeWindowedDeserializer<>(),
```
wcarlson5, line 330: I don't think we should get rid of the generics unless we have to

streams/src/test/java/org/apache/kafka/streams/integration/KStreamAggregationIntegrationTest.java
```diff
@@ -235,25 +235,27 @@ public void shouldReduceWindowed() throws Exception {
                 .thenComparing(KeyValueTimestamp::value);
 
         windowedOutput.sort(comparator);
-        final long firstBatchWindow = firstBatchTimestamp / 500 * 500;
-        final long secondBatchWindow = secondBatchTimestamp / 500 * 500;
+        final long firstBatchWindowStart = firstBatchTimestamp / 500 * 500;
+        final long firstBatchWindowEnd = firstBatchWindowStart + 500;
```
wcarlson5, line 239: Why don't you use `timeDifference` like you did below?

streams/src/test/java/org/apache/kafka/streams/kstream/TimeWindowedDeserializerTest.java
```diff
@@ -56,4 +57,23 @@ public void testWindowedValueDeserializerNoArgConstructors() {
         assertNotNull("Inner deserializer should be not null", inner);
         assertTrue("Inner deserializer type should be ByteArrayDeserializer", inner instanceof ByteArrayDeserializer);
     }
+
+    @Test
+    public void setWindowSizeThroughConfigs() {
+        props.put(StreamsConfig.WINDOW_SIZE_MS_CONFIG, "500");
+        final TimeWindowedDeserializer<?> deserializer = new TimeWindowedDeserializer<>();
+        deserializer.configure(props, false);
```
wcarlson5, line 69: Is there some sort of check you can verify here?

wcarlson5 (commented):
@lct45 Thanks for the PR! I just had a few questions.

streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedDeserializer.java
lct45: I thought about it but figured since it's deprecated anyway and we want to keep backwards compatibility I would leave it. WDYT?

streams/src/test/java/org/apache/kafka/streams/integration/KStreamAggregationIntegrationTest.java
lct45: The window size has to be set either in the constructor or in the configs so if we use the generics we have to use the configs, but the general idea was to push users to do the constructor over the configs. Also in this scenario, I wanted to confirm that it would work fine without configs

streams/src/test/java/org/apache/kafka/streams/integration/KStreamAggregationIntegrationTest.java
lct45: These are time windows so it would be `windowSize`, but I didn't write this test, just updated it to fit with the updated deserializer. I can switch it out to `windowSize` if you think it would help with readability, WYDT?

streams/src/test/java/org/apache/kafka/streams/kstream/TimeWindowedDeserializerTest.java
lct45: Ah yeah, good catch

streams/src/test/java/org/apache/kafka/streams/integration/KStreamAggregationIntegrationTest.java
wcarlson5: It would probably be better practice but it's not necessary to fix. So if you feel like it

streams/src/test/java/org/apache/kafka/streams/integration/KStreamAggregationIntegrationTest.java
wcarlson5: That is fine. but I think It should be like `new TimeWindowedDeserializer<String>(windowedSerde.deserializer(), 500L)` if that complies. We want to keep the type checking

streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedDeserializer.java
wcarlson5: You are probably right since we are deprecating it. But I don't have a strong opinion on this either way

wcarlson5 (commented):
Overall LGTM. Nice job!

guozhangwang (approved):
LGTM!

guozhangwang:
Retriggered tests.

streams/streams-scala/src/test/scala/org/apache/kafka/streams/scala/SerdesUnitTest.scala
```diff
@@ -0,0 +1,35 @@
...
+
+class SerdesUnitTest {
+
+  @Test
+  def shouldLogMessageWhenTimeWindowedSerdeIsUsed(): Unit = {
+
+    Serdes.timeWindowedSerde(new TimeWindowedSerde[String]())
+    val appender = LogCaptureAppender.createAndRegister()
+    val warning = appender.getMessages()
+    assertFalse("There should be a warning about TimeWindowedDeserializer", warning.isEmpty)
```
lct45: Not sure if this is the most secure check - went through a few different iterations but got type issues with all of them except this one. If someone has a better idea, lmk

lct45: This test also appears to fail locally since it uses the constructor we deprecate in this PR and from a quick search it doesn't look like scala suppresses these warnings as easily as Java. Any thoughts on if this test is vital / a way to make this work? cc @guozhangwang 

If I suppress warnings, the test fails (: thoughts on viability?

streams/streams-scala/src/main/scala/org/apache/kafka/streams/scala/Serdes.scala
```diff
@@ -41,8 +43,19 @@ object Serdes {
   implicit def Integer: Serde[Int] = JSerdes.Integer().asInstanceOf[Serde[Int]]
   implicit def JavaInteger: Serde[java.lang.Integer] = JSerdes.Integer()
 
-  implicit def timeWindowedSerde[T](implicit tSerde: Serde[T]): WindowedSerdes.TimeWindowedSerde[T] =
-    new WindowedSerdes.TimeWindowedSerde[T](tSerde)
+  implicit def timeWindowedSerde[T](implicit inner: Serde[T]): Serde[Windowed[T]] =
+    new JSerdes.WrapperSerde[Windowed[T]](
+      new TimeWindowedSerializer[T](inner.serializer),
+      new TimeWindowedDeserializer[T](inner.deserializer) {
+        override def deserialize(topic: String, data: Array[Byte]): Windowed[T] = {
```
guozhangwang: Thinking about this once again: if users did used the `timeWindowedSerde`, then we would log a WARN on each record, which would soon flood the log files. So instead of log at the `deserialize` function, maybe it's better to just log once at the `timeWindowedSerde` itself?

streams/streams-scala/src/test/scala/org/apache/kafka/streams/scala/SerdesUnitTest.scala
guozhangwang: Personally I'm not a big fan to add tests coverage for deprecated functions to make sure the expected warning is generated :) Also, is this test correct? Since currently the `Serdes.timeWindowedSerde` itself would not log any warning, and we would only log when the `deserialize` is called right (see my other comment)? 

Anyways, I'm actually okay if you would like to just remove this test. But on the other hand we should try to avoid a WARN at a per-record granularity too.

streams/streams-scala/src/main/scala/org/apache/kafka/streams/scala/Serdes.scala
lct45: Yeah, I agree that we don't want to spam the logs. There is the `deprecated` annotation above as well - with that do we even need the log messaged for the `timeWindowedSerde`?

streams/streams-scala/src/test/scala/org/apache/kafka/streams/scala/SerdesUnitTest.scala
lct45: Cool, I'm going to go ahead and take the test out. Thanks!

streams/streams-scala/src/main/scala/org/apache/kafka/streams/scala/Serdes.scala
vvcephei: Sorry if I'm missing the point here, @lct45 , but why do we need any custom code at all in the scala wrapper? Shouldn't we just wrap the Java Serde, which you've already set up to warn upon `configure` if there's no constructor arg and no configured value?

lct45: Hmm, maybe we don't need custom code? I did this implementation back in the fall before [KIP-616](https://cwiki.apache.org/confluence/display/KAFKA/KIP-616%3A+Rename+implicit+Serdes+instances+in+kafka-streams-scala) was merged. You know the scala better than I do, but I think you're right that the deprecated warning in the `TimeWindowedDeserializer` will be enough now that KIP-616 added the `deprecated` annotation to this class

guozhangwang:
@lct45 I've merged to trunk. Thanks!

Please feel free to update the corresponding JIRA ticket and KIP status.

PR #9257, opened 2020-09-07
KAFKA-10463 the necessary utilities in Dockerfile should include git

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-10463

the default image of Dockerfile is ```openjdk:8``` and it pre-installed git so it is fine that necessary utilities does not include git. However, the later version of openjdk image does not include git by default and error message "git: command not found" ensues.

### DIFF
--- a/tests/docker/Dockerfile
+++ b/tests/docker/Dockerfile
@@ -32,7 +32,7 @@ ARG ducker_creator=default
 LABEL ducker.creator=$ducker_creator
 
 # Update Linux and install necessary utilities.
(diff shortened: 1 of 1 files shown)


chia7712:
@ijuma Could you take a look? I want to run system test on different jdk. I configure ```jdk_version``` to change the os of kafka dockerfile. Without this patch, it is impossible to switch to other popular jdk images (for example, openjdk:11 and azul/zulu-openjdk:11) quickly since the ```git``` is not installed.

ijuma:
Thanks for the PR. Can you please add a comment explaining that `git` is included in openjdk:8, but not openjdk:11?

chia7712:
> Can you please add a comment explaining that git is included in openjdk:8, but not openjdk:11?

done. Thanks for reviews!

ijuma (approved):
LGTM, thanks.

PR #9262, opened 2020-09-08
MINOR: Fix log message when tasks directory is cleaned manually

cadonna:
Currently when a task directory is cleaned manually the message
for the state dir cleaner is logged instead of the message for
the manual cleanup. This is because the code checks the elapsed
time since the last update before it checks whether the cleanup
is a manual call. This commit changes the order of the checks.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
@@ -282,7 +282,6 @@ public class StateDirectory {
         try {
             cleanRemovedTasksCalledByUser();
         } catch (final Exception e) {
-            // this is already logged within cleanRemovedTasks
             throw new StreamsException(e);
         }
         // remove global dir
@@ -327,16 +326,16 @@ public class StateDirectory {
                             Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
                         }
                     }
-                } catch (final OverlappingFileLockException | IOException e) {
+                } catch (final OverlappingFileLockException | IOException exception) {
                     log.warn("{} Swallowed the following exception during deletion of obsolete state directory {} for task {}: {}",
-                        logPrefix(), dirName, id, e);
+                        logPrefix(), dirName, id, exception);
                 } finally {
                     try {
                         unlock(id);
-                    } catch (final IOException e) {
+                    } catch (final IOException exception) {
                         log.warn("{} Swallowed the following exception during unlocking after " +
                                 "deletion of obsolete state directory for task {}: {}",
-                            logPrefix(), dirName, e);
+                            logPrefix(), dirName, exception);
                     }
                 }
             }
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
@@ -289,10 +289,13 @@ public class StateDirectory {
             if (stateDir.exists()) {
                 Utils.delete(globalStateDir().getAbsoluteFile());
             }
-        } catch (final IOException e) {
-            log.error("{} Failed to delete global state directory of {} due to an unexpected exception",
-                logPrefix(), appId, e);
-            throw new StreamsException(e);
+        } catch (final IOException exception) {
+            log.error(
+                String.format("%s Failed to delete global state directory of %s due to an unexpected exception",
+                    logPrefix(), appId),
+                exception
+            );
+            throw new StreamsException(exception);
         }
     }
 
@@ -327,15 +330,20 @@ public class StateDirectory {
                         }
                     }
                 } catch (final OverlappingFileLockException | IOException exception) {
-                    log.warn("{} Swallowed the following exception during deletion of obsolete state directory {} for task {}: {}",
-                        logPrefix(), dirName, id, exception);
+                    log.warn(
+                        String.format("%s Swallowed the following exception during deletion of obsolete state directory %s for task %s:",
+                            logPrefix(), dirName, id),
+                        exception
+                    );
                 } finally {
                     try {
                         unlock(id);
                     } catch (final IOException exception) {
-                        log.warn("{} Swallowed the following exception during unlocking after " +
-                                "deletion of obsolete state directory for task {}: {}",
-                            logPrefix(), dirName, exception);
+                        log.warn(
+                            String.format("%s Swallowed the following exception during unlocking after deletion of obsolete " +
+                                "state directory %s for task %s:", logPrefix(), dirName, id),
+                            exception
+                        );
                     }
                 }
             }
@@ -352,10 +360,16 @@ public class StateDirectory {
                         log.info("{} Deleting state directory {} for task {} as user calling cleanup.",
                             logPrefix(), dirName, id);
                         Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
+                    } else {
+                        log.warn("{} Could not get lock for state directory {} for task {} as user calling cleanup.",
+                            logPrefix(), dirName, id);
                     }
                 } catch (final OverlappingFileLockException | IOException exception) {
-                    log.error("{} Failed to delete state directory {} for task {} with exception: {}",
-                        logPrefix(), dirName, id, exception);
+                    log.error(
+                        String.format("%s Failed to delete state directory %s for task %s with exception:",
+                            logPrefix(), dirName, id),
+                        exception
+                    );
                     throw exception;
                 } finally {
                     try {
@@ -364,8 +378,11 @@ public class StateDirectory {
                         // the whole directory
                         Utils.delete(taskDir);
                     } catch (final IOException exception) {
-                        log.error("{} Failed to release lock on state directory {} for task {} with exception: {}",
-                            logPrefix(), dirName, id, exception);
+                        log.error(
+                            String.format("%s Failed to release lock on state directory %s for task %s with exception:",
+                                logPrefix(), dirName, id),
+                            exception
+                        );
                         throw exception;
                     }
                 }
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
@@ -292,7 +292,7 @@ public class StateDirectory {
             }
         } catch (final IOException e) {
             log.error("{} Failed to delete global state directory of {} due to an unexpected exception",
-                appId, logPrefix(), e);
+                logPrefix(), appId, e);
             throw new StreamsException(e);
         }
     }
@@ -348,30 +348,27 @@ public class StateDirectory {
             final String dirName = taskDir.getName();
             final TaskId id = TaskId.parse(dirName);
             if (!locks.containsKey(id)) {
-                Exception exception = null;
                 try {
                     if (lock(id)) {
                         log.info("{} Deleting state directory {} for task {} as user calling cleanup.",
-                                logPrefix(), dirName, id);
+                            logPrefix(), dirName, id);
                         Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
                     }
-                } catch (final OverlappingFileLockException | IOException e) {
-                    exception = e;
+                } catch (final OverlappingFileLockException | IOException exception) {
+                    log.error("{} Failed to delete state directory {} for task {} with exception: {}",
+                        logPrefix(), dirName, id, exception);
+                    throw exception;
                 } finally {
                     try {
                         unlock(id);
-
                         // for manual user call, stream threads are not running so it is safe to delete
                         // the whole directory
                         Utils.delete(taskDir);
-                    } catch (final IOException e) {
-                        exception = e;
+                    } catch (final IOException exception) {
+                        log.error("{} Failed to release lock on state directory {} for task {} with exception: {}",
+                            logPrefix(), dirName, id, exception);
+                        throw exception;
                     }
-                }
-
-                if (exception != null) {
-                    log.error("{} Failed to release the state directory lock.", logPrefix());
-                    throw exception;
                 }
             }
         }
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
@@ -328,14 +328,14 @@ public class StateDirectory {
                     if (lock(id)) {
                         final long now = time.milliseconds();
                         final long lastModifiedMs = taskDir.lastModified();
-                        if (now > lastModifiedMs + cleanupDelayMs) {
-                            log.info("{} Deleting obsolete state directory {} for task {} as {}ms has elapsed (cleanup delay is {}ms).",
-                                logPrefix(), dirName, id, now - lastModifiedMs, cleanupDelayMs);
-
-                            Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
-                        } else if (manualUserCall) {
+                        if (manualUserCall) {
                             log.info("{} Deleting state directory {} for task {} as user calling cleanup.",
                                 logPrefix(), dirName, id);
+
+                            Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
+                        } else if (now > lastModifiedMs + cleanupDelayMs) {
+                            log.info("{} Deleting obsolete state directory {} for task {} as {}ms has elapsed (cleanup delay is {}ms).",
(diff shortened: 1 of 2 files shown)


cadonna:
Call for review: @guozhangwang @ableegoldman @mjsax @vvcephei @wcarlson5

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -328,15 +328,15 @@ private synchronized void cleanRemovedTasks(final long cleanupDelayMs,
                     if (lock(id)) {
                         final long now = time.milliseconds();
                         final long lastModifiedMs = taskDir.lastModified();
-                        if (now > lastModifiedMs + cleanupDelayMs) {
-                            log.info("{} Deleting obsolete state directory {} for task {} as {}ms has elapsed (cleanup delay is {}ms).",
-                                logPrefix(), dirName, id, now - lastModifiedMs, cleanupDelayMs);
-
-                            Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
-                        } else if (manualUserCall) {
+                        if (manualUserCall) {
```
ableegoldman: Honestly it kind of seems like there is enough divergent logic to merit splitting this up into separate methods for the manual vs cleanup-delay cases.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -328,15 +328,15 @@ private synchronized void cleanRemovedTasks(final long cleanupDelayMs,
...
                             log.info("{} Deleting state directory {} for task {} as user calling cleanup.",
                                 logPrefix(), dirName, id);
 
+                            Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
```
wcarlson5: duplicate line of 340, can probably pull out of the if. Then there should be no need to make a separate method as the only divergence is a log msg.

wcarlson5 (commented):
seems like simple enough log change, can make a small change but not necessary

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
cadonna: That is not entirely true, because the exception handling differs between user call and cleaner thread call. The cleaner thread call swallows all exceptions and the user call throws all exceptions.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
cadonna: I think this is a good idea. I will split the method into two.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -306,40 +306,52 @@ public synchronized void clean() {
...
                             Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
                         }
                     }
+                } catch (final OverlappingFileLockException | IOException e) {
+                    // do nothing
+                } finally {
+                    try {
+                        unlock(id);
+                    } catch (final IOException e) {
+                        // do nothing
```
ableegoldman: Can we at least log a warning with the exception we're swallowing? Same for the `catch (final OverlappingFileLockException | IOException e) ` above

ableegoldman (approved):
Nice, one suggestion to log the swallowed exceptions but otherwise LGTM. Thanks for breaking up the method too, it definitely made things easier to follow

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -280,7 +280,7 @@ synchronized void unlock(final TaskId taskId) throws IOException {
     public synchronized void clean() {
         // remove task dirs
         try {
-            cleanRemovedTasks(0, true);
+            cleanRemovedTasksCalledByUser();
         } catch (final Exception e) {
             // this is already logged within cleanRemovedTasks
```
chia7712: cleanRemovedTasks -> cleanRemovedTasksCalledByUser

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -306,40 +306,55 @@ public synchronized void clean() {
...
+            final TaskId id = TaskId.parse(dirName);
+            if (!locks.containsKey(id)) {
+                Exception exception = null;
+                try {
+                    if (lock(id)) {
+                        log.info("{} Deleting state directory {} for task {} as user calling cleanup.",
+                                logPrefix(), dirName, id);
+                        Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
+                    }
                 } catch (final OverlappingFileLockException | IOException e) {
```
chia7712: the method ```clean``` catches ```Exception``` already. Could we get rid of those try-catch statements? the code ```log.error("{} Failed to release the state directory lock.", logPrefix());``` can be moved to ```clean```. For example:
```java
    public synchronized void clean() {
        // remove task dirs
        try {
            cleanRemovedTasksCalledByUser();
        } catch (final Exception e) {
            log.error("{} Failed to release the state directory lock.", logPrefix());
            throw new StreamsException(e);
        }
```
```java
    private void cleanRemovedTasksCalledByUser() throws Exception {
        for (final File taskDir : listAllTaskDirectories()) {
            final String dirName = taskDir.getName();
            final TaskId id = TaskId.parse(dirName);
            if (!locks.containsKey(id) && lock(id)) {
                try {
                    log.info("{} Deleting state directory {} for task {} as user calling cleanup.",
                            logPrefix(), dirName, id);
                    Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
                } finally {
                    unlock(id);
                    // for manual user call, stream threads are not running so it is safe to delete
                    // the whole directory
                    Utils.delete(taskDir);
                }
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
cadonna: I am afraid I do not understand this comment.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
cadonna: Good observation! I applied some of your suggestions but left the `try-catch` statements since they allow to log more specific information about the encountered error.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
chia7712: The previous comment is invalid as the method "cleanRemovedTasks" is removed. The replacement is "cleanRemovedTasksCalledByUser"

cadonna:
Retest this please

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -292,7 +291,7 @@ public synchronized void clean() {
             }
         } catch (final IOException e) {
             log.error("{} Failed to delete global state directory of {} due to an unexpected exception",
-                appId, logPrefix(), e);
+                logPrefix(), appId, e);
```
vvcephei: Just noticed something else wrong with this message. `e` will get ignored. We should instead use `log.error(String.format("...", logPrefix(), appId), e)`

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -306,59 +305,69 @@ public synchronized void clean() {
...
-                                logPrefix(), dirName, id);
-
                             Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
                         }
                     }
-                } catch (final OverlappingFileLockException | IOException e) {
-                    exception = e;
+                } catch (final OverlappingFileLockException | IOException exception) {
+                    log.warn("{} Swallowed the following exception during deletion of obsolete state directory {} for task {}: {}",
+                        logPrefix(), dirName, id, exception);
```
vvcephei: This log will be incomplete. We report the exception as the cause:

```suggestion
                    log.warn(String.format("%s Swallowed the following exception during deletion of obsolete state directory %s for task %s",
                        logPrefix(), dirName, id), exception);
```

This feedback applies to pretty much all the warn/err logs in this PR.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
mjsax: Nice catch!

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -306,59 +305,69 @@ public synchronized void clean() {
...
-    private synchronized void cleanRemovedTasks(final long cleanupDelayMs,
-                                                final boolean manualUserCall) throws Exception {
-        final File[] taskDirs = listAllTaskDirectories();
-        if (taskDirs == null || taskDirs.length == 0) {
-            return; // nothing to do
-        }
-
-        for (final File taskDir : taskDirs) {
+    private void cleanRemovedTasksCalledByCleanerThread(final long cleanupDelayMs) {
+        for (final File taskDir : listAllTaskDirectories()) {
```
mjsax, line 318: I this save? I _think_ if `listAllTaskDirectories()` returns `null` we would get an NPE?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -306,59 +305,69 @@ public synchronized void clean() {
...
                     }
                 }
+            }
+        }
+    }
 
-                if (exception != null && manualUserCall) {
-                    log.error("{} Failed to release the state directory lock.", logPrefix());
+    private void cleanRemovedTasksCalledByUser() throws Exception {
+        for (final File taskDir : listAllTaskDirectories()) {
```
mjsax, line 354: As above.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -306,59 +305,69 @@ public synchronized void clean() {
...
+            final String dirName = taskDir.getName();
+            final TaskId id = TaskId.parse(dirName);
+            if (!locks.containsKey(id)) {
+                try {
+                    if (lock(id)) {
+                        log.info("{} Deleting state directory {} for task {} as user calling cleanup.",
+                            logPrefix(), dirName, id);
+                        Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
+                    }
```
mjsax, line 366: Should we log a message that we could not get the lock? -- For a manual user call, this might be good to know?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -306,59 +305,69 @@ public synchronized void clean() {
...
+                } finally {
+                    try {
+                        unlock(id);
+                        // for manual user call, stream threads are not running so it is safe to delete
+                        // the whole directory
+                        Utils.delete(taskDir);
+                    } catch (final IOException exception) {
+                        log.error("{} Failed to release lock on state directory {} for task {} with exception: {}",
+                            logPrefix(), dirName, id, exception);
+                        throw exception;
```
mjsax, line 386: Might be better to use an `Exception` variable `firstException` and rethrow at the end if not `null` -- IIRC, behavior is undefined if we throw a second exception (ie, `finally` would executed after the first (outer) `catch` block.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -306,59 +305,69 @@ public synchronized void clean() {
...
                 try {
                     if (lock(id)) {
                         final long now = time.milliseconds();
                         final long lastModifiedMs = taskDir.lastModified();
                         if (now > lastModifiedMs + cleanupDelayMs) {
                             log.info("{} Deleting obsolete state directory {} for task {} as {}ms has elapsed (cleanup delay is {}ms).",
                                 logPrefix(), dirName, id, now - lastModifiedMs, cleanupDelayMs);
-
-                            Utils.delete(taskDir, Collections.singletonList(new File(taskDir, LOCK_FILE_NAME)));
-                        } else if (manualUserCall) {
```
mjsax, line 336: Could we not fix this by flipping the logic to:
```
if (manualUserCall) {
} else if (now > lastModifiedMs + cleanupDelayMs) {
}
```

To avoid code duplication?

mjsax (commented):
View minor comments. Overall LGTM; I am just wondering about the code duplication?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
ableegoldman: We recently "fixed" `listAllTaskDirectories` to guarantee that it never returns null. We just missed to update all the null checks when we did that

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
ableegoldman: That's what Bruno originally did, but the original `cleanRemovedTasks` method branched on the `manualUserCall` flag in several places and was pretty difficult to follow (imo). So (also imo) it's cleaner to split it up into two methods that make it clear what the expected behavior is in each case. Just my 2 cents

vvcephei (changes requested):
Thanks @cadonna , this logical change looks good to me. I had some feedback on the log messages, though.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
cadonna: I saw the behavior @ableegoldman mentions and then I changed the code. So it is safe.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
cadonna: I agree with @ableegoldman. Now the code is easier to follow.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
cadonna: According to the Java Language Specification, it is not undefined. See https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.2. There it says:

> If the catch block completes abruptly for reason R, then the finally block is executed. Then there is a choice:
> 
>    - If the finally block completes normally, then the try statement completes abruptly for reason R.
> 
>    - If the finally block completes abruptly for reason S, then the try statement completes abruptly for reason S (and reason R is discarded).
> 

That means, the last thrown exception is propagated. That was also the behavior of the refactored code. Since the goal of this PR is actually just to fix the log message without changing the behavior, I would leave this change to a future PR if we feel that the behavior should change.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
cadonna: Good catch!

vvcephei:
Test failure unrelated: `Build / JDK 11 / kafka.api.ConsumerBounceTest.testRollingBrokerRestartsWithSmallerMaxGroupSizeConfigDisruptsBigGroup`

chia7712:
@cadonna 

```
org.apache.kafka.streams.processor.internals.StateDirectoryTest.shouldLogStateDirCleanerMessage failed, log available in /home/chia7712/kafka/streams/build/reports/testOutput/org.apache.kafka.streams.processor.internals.StateDirectoryTest.shouldLogStateDirCleanerMessage.test.stdout

org.apache.kafka.streams.processor.internals.StateDirectoryTest > shouldLogStateDirCleanerMessage FAILED
    java.lang.AssertionError:
    Expected: a collection containing a string ending with "ms has elapsed (cleanup delay is 0ms)."
         but: was empty
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
        at org.apache.kafka.streams.processor.internals.StateDirectoryTest.shouldLogStateDirCleanerMessage(StateDirectoryTest.java:569)
```

the new test introduced by this PR is failed. The failed test does not update the current time of ```MockTime``` so the state directory can't be cleanup. I have filed a ticket for that failed test (https://issues.apache.org/jira/browse/KAFKA-10514).

vvcephei:
Cherry-picked to 2.6

PR #9266, opened 2020-09-08
KAFKA-10469: Resolve logger levels hierarchically

tombentley:
Previous to root logger level was used, ignoring intervening loggers with different levels.


tombentley:
@ijuma please could you take a look?

stanislavkozlovski (commented):
This is a good catch - thank you for the PR.

This looks pretty good to me. As a nit, could we add a test that replicas the issue reported? In other words, have default logger be INFO, kafka.controller be TRACE and assert some sub-package of the controller is correct

tombentley:
@stanislavkozlovski I'm not sure how what you describe differs from the test implemented, though it's not identically the same because the tests have their own `log4j.properties` which configures `kafka`, but not `kafka.controller`. I guess it might be clearer (and less fragile) if the test loggers were `foo` and `foo.bar`.

stanislavkozlovski:
Sorry, I mistook `kafka` for the root logger. I agree it tests the same path

stanislavkozlovski (approved):

core/src/test/scala/kafka/utils/LoggingTest.scala
```diff
@@ -58,4 +60,20 @@ class LoggingTest extends Logging {
 
     assertEquals(logging.getClass.getName, logging.log.underlying.getName)
   }
+
+  @Test
+  def testLoggerLevelIsResolved(): Unit = {
+    val controller = new Log4jController()
+    val previousLevel = controller.getLogLevel("kafka")
+    try {
+      controller.setLogLevel("kafka", "TRACE")
```
dongjinleekr, line 76: @tombentley A small propsal: would `org.apache.log4j.Level.TRACE.toString` be better?

It passes compilation checks.

dongjinleekr (approved):

core/src/test/scala/kafka/utils/LoggingTest.scala
tombentley: @dongjinleekr that would be fine for the `setLogLevel()` and `assertEquals` calls, but there would still be the level name in the `assertTrue(...contains("kafka=TRACE"))` calls. I think fixing it there would make the code harder to read, and given that the declaration of `TRACE` is in an external API on which we depend, and the name will not change there doesn't seem to be any benefit to using the declaration rather than the name as a literal.

dongjinleekr: @tombentley Great. Your stance is reasonable. I am convinced.

tombentley:
@ijuma @hachikuji @guozhangwang any chance one of you could review this?

tombentley:
@ijuma @hachikuji @guozhangwang @mjsax @mimaison please could one of you could take a look?

core/src/test/scala/kafka/utils/LoggingTest.scala
```diff
@@ -58,4 +60,20 @@ class LoggingTest extends Logging {
...
+      Logger(LoggerFactory.getLogger("kafka.utils.Log4jControllerTest")).trace("test")
```
ijuma: Can we add a comment explaining why we are invoking the logger here?

core/src/main/scala/kafka/utils/Log4jController.scala
```diff
@@ -29,6 +29,24 @@ import scala.jdk.CollectionConverters._
 object Log4jController {
   val ROOT_LOGGER = "root"
 
+  def resolveLevel(logger: Logger): String = {
```
ijuma: Should this be private?

core/src/test/scala/kafka/utils/LoggingTest.scala
```diff
@@ -58,4 +60,22 @@ class LoggingTest extends Logging {
...
+    val previousLevel = controller.getLogLevel("kafka")
+    try {
+      controller.setLogLevel("kafka", "TRACE")
+      // Do some logging so that the Logger is created within the hierarchy
+      // (until loggers are used only loggers in the config file exist)
+      Logger(LoggerFactory.getLogger("kafka.utils.Log4jControllerTest")).trace("test")
+      Assert.assertEquals("TRACE", controller.getLogLevel("kafka"))
+      Assert.assertEquals("TRACE", controller.getLogLevel("kafka.utils.Log4jControllerTest"))
+      Assert.assertTrue(controller.getLoggers.contains("kafka=TRACE"))
+      Assert.assertTrue(controller.getLoggers.contains("kafka.utils.Log4jControllerTest=TRACE"))
```
ijuma: Nit: we don't usually include the `Assert.` prefix in test assertions.

tombentley:
Many thanks @ijuma all fixed.

ijuma (approved):
LGTM, thanks!

core/src/test/scala/kafka/utils/LoggingTest.scala
dongjinleekr: I verified that it works without `com.typesafe.scalalogging.Logger#apply` here, i.e.,  `LoggerFactory.getLogger("kafka.utils.Log4jControllerTest").trace("test")`. Other test cases like `RestServerTest` don't invoke it. It would be better to remove it.

dongjinleekr (changes requested):
Please remove `com.typesafe.scalalogging.Logger` from `LoggingTest`.

tombentley:
@dongjinleekr done.

dongjinleekr (approved):
Great! Looks Perfect for Me! :smile:

ijuma:
Build timed out for some reason, started another one.

tombentley:
@ijuma I've fixed an infinite loop, could you trigger the build again please?

ijuma:
@tombentley `testDescribeConfigsForLog4jLogLevels` is failing.

tombentley:
@ijuma sorry about that. Now fixed.

tombentley:
@ijuma any chance you could trigger a CI build? Thanks!

core/src/test/scala/integration/kafka/api/PlaintextAdminIntegrationTest.scala
```diff
@@ -2061,11 +2062,12 @@ class PlaintextAdminIntegrationTest extends BaseAdminIntegrationTest {
...
   def testDescribeConfigsForLog4jLogLevels(): Unit = {
     client = Admin.create(createConfig)
-
+    LoggerFactory.getLogger("kafka.cluster.Replica").trace("Message to create the logger")
     val loggerConfig = describeBrokerLoggers()
-    val rootLogLevel = loggerConfig.get(Log4jController.ROOT_LOGGER).value()
+    val kafkaLogLevel = loggerConfig.get("kafka").value()
     val logCleanerLogLevelConfig = loggerConfig.get("kafka.cluster.Replica")
-    assertEquals(rootLogLevel, logCleanerLogLevelConfig.value()) // we expect an undefined log level to be the same as the root logger
+    // we expect an undefined log level to be the same as its first ancestor logger
```
ijuma: Wouldn't the first ancestor be `kafka.cluster`?

tombentley: It would, but that's not configured in the `log4j.properties` used for this test, so we inherit from the 2nd ancestor, which is ERROR, where as the root logger was OFF. Which is why I had to fix this test. Let me fix the comment.

ijuma (approved):
LGTM, thanks.

tombentley:
@ijuma it's just been brought to my attention that the root looger default behaviour is actually explictly specified in [KIP-412](https://cwiki.apache.org/confluence/display/KAFKA/KIP-412%3A+Extend+Admin+API+to+support+dynamic+application+log+levels#KIP412:ExtendAdminAPItosupportdynamicapplicationloglevels-DescribeConfigs) (search for "Returns the root logger's log level if this logger is not set explicitly yet"). I hadn't previously realised that this was specified in a KIP. So is it OK to do this without a KIP?

ijuma:
I thought this was a bug versus changing the spec. If we are changing the spec, then we do need a KIP.

tombentley:
@ijuma I initially saw it as a bug, because it's pretty surprising behaviour (because it differs from how logger inheritance works in log4j), but I hadn't realised until after you merged that it was specified in a KIP. KIP freeze for 2.7 is today, so there's no time to get a KIP approved for this in the 2.7 timeframe. I guess that means we'll have to revert it for now and I write a KIP to target 3.0. Sorry about this.

ijuma:
@tombentley Thanks for the explanation. I have reverted this change for now: https://github.com/apache/kafka/commit/75e53585255a0c2737dee84ddbad39094c5e8500

tombentley:
Thanks for the update @ijuma, I'll prepare a KIP for 3.0.

tombentley:
@ijuma [KIP-676](https://cwiki.apache.org/confluence/display/KAFKA/KIP-676%3A+Respect+logging+hierarchy?moved=true) has been approved, so I think we should be able to merge this change to trunk.

tombentley:
Maybe @vvcephei, @mimaison or @gwenshap could take a look at this since you voted on the KIP? Thanks!

tombentley:
We just need to revert 75e5358 (which reverted the original merged). Do you need an new PR for that?

ijuma:
I can reapply the change soon.

tombentley:
Great, thanks Ismael.

ijuma:
I have reapplied the change https://github.com/apache/kafka/commit/960e64072592bc1ad939518b58dbdfb17068680d

tombentley:
Thanks Ismael.

PR #9267, opened 2020-09-08
MINOR: Add debug logs for StreamThread

vvcephei:
Add debug logs to see when Streams calls poll, process, commit, etc. 

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -644,7 +644,7 @@ public class StreamThread extends Thread {
              *  6. Otherwise, increment N.
              */
             do {
-                log.debug("Invoking TaskManager#process with {} iterations.", numIterations);
+                log.debug("Processing tasks with {} iterations.", numIterations);
                 final int processed = taskManager.process(numIterations, time);
                 final long processLatency = advanceNowAndComputeLatency();
                 totalProcessLatency += processLatency;
@@ -662,7 +662,9 @@ public class StreamThread extends Thread {
                     totalProcessed += processed;
                 }
 
-                log.debug("TaskManager#process handled {} records; invoking TaskManager#punctuate", processed);
+                log.debug("Processed {} records with {} iterations; invoking punctuators if necessary",
+                          processed,
+                          numIterations);
 
                 final int punctuated = taskManager.punctuate();
                 final long punctuateLatency = advanceNowAndComputeLatency();
@@ -671,7 +673,7 @@ public class StreamThread extends Thread {
                     punctuateSensor.record(punctuateLatency / (double) punctuated, now);
                 }
 
-                log.debug("TaskManager#punctuate executed: {}", punctuated);
+                log.debug("{} punctuators ran.", punctuated);
 
                 final int committed = maybeCommit();
                 final long commitLatency = advanceNowAndComputeLatency();
@@ -713,39 +715,40 @@ public class StreamThread extends Thread {
     }
 
     private void initializeAndRestorePhase() {
-        {
-            // only try to initialize the assigned tasks
-            // if the state is still in PARTITION_ASSIGNED after the poll call
-            final State stateSnapshot = state;
-            if (stateSnapshot == State.PARTITIONS_ASSIGNED
-                || stateSnapshot == State.RUNNING && taskManager.needsInitializationOrRestoration()) {
+        // only try to initialize the assigned tasks
+        // if the state is still in PARTITION_ASSIGNED after the poll call
+        final State stateSnapshot = state;
+        if (stateSnapshot == State.PARTITIONS_ASSIGNED
+            || stateSnapshot == State.RUNNING && taskManager.needsInitializationOrRestoration()) {
 
-                log.debug("State is {}; initializing and restoring", stateSnapshot);
+            log.debug("State is {}; initializing tasks if necessary", stateSnapshot);
 
-                // transit to restore active is idempotent so we can call it multiple times
-                changelogReader.enforceRestoreActive();
+            // transit to restore active is idempotent so we can call it multiple times
+            changelogReader.enforceRestoreActive();
 
-                if (taskManager.tryToCompleteRestoration()) {
-                    changelogReader.transitToUpdateStandby();
+            if (taskManager.tryToCompleteRestoration()) {
+                changelogReader.transitToUpdateStandby();
 
-                    setState(State.RUNNING);
-                }
+                setState(State.RUNNING);
+            }
 
-                if (log.isDebugEnabled()) {
-                    log.debug("Initialization and restore call done. State is {}", state);
-                }
+            if (log.isDebugEnabled()) {
+                log.debug("Initialization call done. State is {}", state);
             }
         }
 
-        log.debug("Invoking ChangeLogReader#restore");
+        if (log.isDebugEnabled()) {
+            log.debug("Idempotently invoking restoration logic in state {}", state);
+        }
         // we can always let changelog reader try restoring in order to initialize the changelogs;
         // if there's no active restoring or standby updating it would not try to fetch any data
         changelogReader.restore();
+        log.debug("Idempotent restore call done. Thread state has not changed.");
     }
 
     private long pollPhase() {
         final ConsumerRecords<byte[], byte[]> records;
-        log.debug("Invoking Consumer#poll");
+        log.debug("Invoking poll on main Consumer");
 
         if (state == State.PARTITIONS_ASSIGNED) {
             // try to fetch some records with zero poll millis
@@ -772,7 +775,7 @@ public class StreamThread extends Thread {
         final long pollLatency = advanceNowAndComputeLatency();
 
         if (log.isDebugEnabled()) {
-            log.debug("Consumer#poll completed in {} ms and fetched {} records", pollLatency, records.count());
+            log.debug("Main Consumer poll completed in {} ms and fetched {} records", pollLatency, records.count());
         }
         pollSensor.record(pollLatency, now);
 
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -612,63 +612,18 @@ public class StreamThread extends Thread {
         final long startMs = time.milliseconds();
         now = startMs;
 
-        if (state == State.PARTITIONS_ASSIGNED) {
-            // try to fetch some records with zero poll millis
-            // to unblock the restoration as soon as possible
-            records = pollRequests(Duration.ZERO);
-        } else if (state == State.PARTITIONS_REVOKED) {
-            // try to fetch som records with zero poll millis to unblock
-            // other useful work while waiting for the join response
-            records = pollRequests(Duration.ZERO);
-        } else if (state == State.RUNNING || state == State.STARTING) {
-            // try to fetch some records with normal poll time
-            // in order to get long polling
-            records = pollRequests(pollTime);
-        } else if (state == State.PENDING_SHUTDOWN) {
-            // we are only here because there's rebalance in progress,
-            // just poll with zero to complete it
-            records = pollRequests(Duration.ZERO);
-        } else {
-            // any other state should not happen
-            log.error("Unexpected state {} during normal iteration", state);
-            throw new StreamsException(logPrefix + "Unexpected state " + state + " during normal iteration");
-        }
-
-        final long pollLatency = advanceNowAndComputeLatency();
-
-        pollSensor.record(pollLatency, now);
-        if (records != null && !records.isEmpty()) {
-            pollRecordsSensor.record(records.count(), now);
-            taskManager.addRecordsToTasks(records);
-        }
+        final long pollLatency = pollPhase();
 
         // Shutdown hook could potentially be triggered and transit the thread state to PENDING_SHUTDOWN during #pollRequests().
         // The task manager internal states could be uninitialized if the state transition happens during #onPartitionsAssigned().
         // Should only proceed when the thread is still running after #pollRequests(), because no external state mutation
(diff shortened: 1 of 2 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -612,63 +612,18 @@ void runOnce() {
...
-        }
-
-        final long pollLatency = advanceNowAndComputeLatency();
-
-        pollSensor.record(pollLatency, now);
-        if (records != null && !records.isEmpty()) {
-            pollRecordsSensor.record(records.count(), now);
-            taskManager.addRecordsToTasks(records);
-        }
+        final long pollLatency = pollPhase();
```
vvcephei, line 615: `runOnce` was too long, according to checkStyle, so I factored out some of the execution phases.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -612,63 +612,18 @@ void runOnce() {
...
 
         // Shutdown hook could potentially be triggered and transit the thread state to PENDING_SHUTDOWN during #pollRequests().
         // The task manager internal states could be uninitialized if the state transition happens during #onPartitionsAssigned().
         // Should only proceed when the thread is still running after #pollRequests(), because no external state mutation
         // could affect the task manager state beyond this point within #runOnce().
         if (!isRunning()) {
-            log.debug("State already transits to {}, skipping the run once call after poll request", state);
+            log.debug("Thread state is already {}, skipping the run once call after poll request", state);
```
vvcephei, line 622: Just a slight rewording I thought could be clearer.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -752,6 +712,77 @@ void runOnce() {
...
 
+    private void initializeAndRestorePhase() {
+        {
+            // only try to initialize the assigned tasks
+            // if the state is still in PARTITION_ASSIGNED after the poll call
+            final State stateSnapshot = state;
+            if (stateSnapshot == State.PARTITIONS_ASSIGNED
+                || stateSnapshot == State.RUNNING && taskManager.needsInitializationOrRestoration()) {
+
+                log.debug("State is {}; initializing and restoring", stateSnapshot);
```
vvcephei: I wanted to make all my new `debug` statements zero-cost if debug isn't enabled. Since `state` is volatile, resolving the arguments for this call would result in an uncached read of main memory regardless of the log level. Since we don't expect state to actually change in these few lines, I just read it once and saved it in `stateSnapshot`.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -752,6 +712,77 @@ void runOnce() {
...
+
+                if (taskManager.tryToCompleteRestoration()) {
+                    changelogReader.transitToUpdateStandby();
+
+                    setState(State.RUNNING);
+                }
+
+                if (log.isDebugEnabled()) {
+                    log.debug("Initialization and restore call done. State is {}", state);
+                }
```
vvcephei: This time, we really need to read the field, so I'm guarding the method call for the common case in which debug is off.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -752,6 +712,77 @@ void runOnce() {
...
+            // any other state should not happen
+            log.error("Unexpected state {} during normal iteration", state);
+            throw new StreamsException(logPrefix + "Unexpected state " + state + " during normal iteration");
+        }
+
+        final long pollLatency = advanceNowAndComputeLatency();
+
+        if (log.isDebugEnabled()) {
+            log.debug("Consumer#poll completed in {} ms and fetched {} records", pollLatency, records.count());
+        }
```
vvcephei: Guarded to avoid calling `records.count()` when debug is off.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -752,6 +712,77 @@ void runOnce() {
...
+        pollSensor.record(pollLatency, now);
+
+        if (!records.isEmpty()) {
```
vvcephei, line 782: Note, this was previously `records != null && !records.isEmpty()`:
https://github.com/apache/kafka/pull/9267/files#diff-045aeaddb4232a85a8560186b4901e69L640

However, `records` can never be null, except if `Consumer#poll` returns `null`, which it does not. It turned out the reason for checking this condition was that there was exactly one test that relied on a nice mock returning `null`. I fixed the test below.

Note, the only reason I messed with this was to simplify the debug log message on L775. Otherwise, I'd have needed to think of what to say if `records` were `null`.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -760,7 +791,7 @@ void runOnce() {
      * @throws TaskMigratedException if the task producer got fenced (EOS only)
      */
     private ConsumerRecords<byte[], byte[]> pollRequests(final Duration pollTime) {
-        ConsumerRecords<byte[], byte[]> records = null;
+        ConsumerRecords<byte[], byte[]> records = ConsumerRecords.empty();
```
vvcephei, line 797: Not necessary, but also not harmful, since it's a static final instance anyway. I thought it was nicer for self-documentation this way.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -845,8 +876,8 @@ private void addToResetList(final TopicPartition partition, final Set<TopicParti
     int maybeCommit() {
         final int committed;
         if (now - lastCommitMs > commitTimeMs) {
-            if (log.isTraceEnabled()) {
-                log.trace("Committing all active tasks {} and standby tasks {} since {}ms has elapsed (commit interval is {}ms)",
+            if (log.isDebugEnabled()) {
+                log.debug("Committing all active tasks {} and standby tasks {} since {}ms has elapsed (commit interval is {}ms)",
```
vvcephei, line 883: Switched these to debug now, since they seem to fit with the newly added logs.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamThreadTest.java
```diff
@@ -496,6 +498,7 @@ public void shouldEnforceRebalanceAfterNextScheduledProbingRebalanceTime() throw
         );
         
         final Consumer<byte[], byte[]> mockConsumer = EasyMock.createNiceMock(Consumer.class);
+        expect(mockConsumer.poll(anyObject())).andStubReturn(ConsumerRecords.empty());
```
vvcephei, line 501: This was a bad input for the test, since `poll` can never actually return `null`.

vvcephei (commented):
Hey @wcarlson5 @guozhangwang ,

Do you mind taking a look at this? I ran into the lack of these logs when I was literally trying to "debug" a Streams problem this weekend.

I considered whether to make this debug- or trace-level, and settled on debug. I tested how much actually gets logged with this logger on. The log frequency is somewhere in the neighborhood of 10s to 100s of millis. Subjectively, this is too high to leave on by default, but I think it's still within my expectations for "debug". I feel like "trace" is more like "print every record you handle", so it would be more like 1s of millis between messages, if not multiple messages per millisecond.

Sorry about the diff; I had to factor out some of the logic because checkstyle was yelling at me about the length of runOnce.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -752,6 +712,77 @@ void runOnce() {
...
+                    setState(State.RUNNING);
+                }
+
+                if (log.isDebugEnabled()) {
+                    log.debug("Initialization and restore call done. State is {}", state);
+                }
+            }
+        }
+
+        log.debug("Invoking ChangeLogReader#restore");
```
wcarlson5: Is this necessary with the logs inside restore()? 

maybe can include snapshotState so we can see if it's STARTING or RUNNING? because we don't see the state unless it enters the initialization. Not sure if this would be useful

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -706,13 +662,17 @@ void runOnce() {
...
+                log.debug("TaskManager#process handled {} records; invoking TaskManager#punctuate", processed);
+
                 final int punctuated = taskManager.punctuate();
                 final long punctuateLatency = advanceNowAndComputeLatency();
                 totalPunctuateLatency += punctuateLatency;
                 if (punctuated > 0) {
                     punctuateSensor.record(punctuateLatency / (double) punctuated, now);
                 }
 
+                log.debug("TaskManager#punctuate executed: {}", punctuated);
```
wcarlson5: trace?

wcarlson5 (approved):
Looking through the logging seems to be rather comprehensive now. Unless you want to record every record in trace. 

I am however unsure why you decided all these logs needed to be at the same level. While you are right they could all work in debug, it might be a good to drop a few of the more specific log down to trace. Not 100% sure on which those are though... 

But I don't see any real problems here

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -706,13 +662,17 @@ void runOnce() {
                     totalProcessed += processed;
                 }
 
+                log.debug("TaskManager#process handled {} records; invoking TaskManager#punctuate", processed);
```
guozhangwang: Nit: I'd suggest we do not expose internal class names in log entries, e.g. here we can say "Processed {} records with {} iterations, invoking punctuation now", ditto below.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -689,6 +644,7 @@ void runOnce() {
              *  6. Otherwise, increment N.
              */
             do {
+                log.debug("Invoking TaskManager#process with {} iterations.", numIterations);
```
guozhangwang: What's the rationale of recording both the starting and the ending of a procedure? If it is for trouble shooting purposes only maybe the starting log entry can be trace while ending entry is debug?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
guozhangwang: SG.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
guozhangwang: Not sure what's the purpose of this log entry?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
vvcephei: Yes, it's for debugging purposes, so I'd like to keep them both at debug level. When I was handling the incident that inspired this PR, what I really wanted to see in the logs is both how often we called poll and which operation in the loop was blocking. In theory, I'd only really need the "before" or the "after" logs for this, but only if I also have the source code pulled up to compare with the log messages. I'm proposing to (redundantly) log both "before" and "after" messages so that the logs will be context-free and people in the future would be able to tell what is happening just from the narrative of the logs themselves.

guozhangwang: SG

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
vvcephei: Thanks, all. The purpose is simply to make the narrative of StreamThread's debug logs unambiguous and complete. The purpose of logging it here instead of relying only on ChangeLogReader's logs is that you can enable _just_ StreamThread's debug logger and get a complete high-level view of what's happening. If you want to then drill down into the restore call itself, you could enable ChangeLogReader's debug log.

Good point, @wcarlson5 , about logging the state. I'll add it.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
vvcephei: Note, this isn't saying that we invoked a single punctuator, but that we invoked _all_ the punctuators that are runnable right now. It's also a top-level phase of executing StreamThread, so I don't think we can leave it out of the debug logs without telling an incomplete story of what's happening. If we log this (or any of the proposed logs) at trace level, the only consequence is that users who want to _debug_ StreamThread would have to use "trace" level instead of "debug" level. This seems to add unnecessary complexity to the logging.

IMO, it's better to reserve "trace" level for _very_ low-level logs, such as logging the progress of individual documents through the processors, and use "debug" level for higher-level summary logs like the ones in this PR.

vvcephei:
Thanks for the review, @wcarlson5 and @guozhangwang !

I've re-worded the logs to avoid internal class names and added the state before calling restore. Aside from those changes, I responded in the threads above.

WDYT?

guozhangwang (approved):
LGTM

vvcephei:
Thanks, @guozhangwang !

vvcephei:
Test failures were unrelated:
```
    Build / JDK 8 / kafka.network.ConnectionQuotasTest.testNoConnectionLimitsByDefault
    Build / JDK 8 / kafka.network.ConnectionQuotasTest.testNoConnectionLimitsByDefault
    Build / JDK 11 / kafka.network.DynamicConnectionQuotaTest.testDynamicListenerConnectionCreationRateQuota
```

PR #9270, opened 2020-09-09
KAFKA-10284: Group membership update due to static member rejoin should be persisted

feyman2016:


abbccdda:
retest this please

abbccdda:
retest this please

abbccdda:
test this

core/src/test/scala/unit/kafka/coordinator/group/GroupCoordinatorTest.scala
```diff
@@ -3882,6 +3942,21 @@ class GroupCoordinatorTest {
     Await.result(responseFuture, Duration(rebalanceTimeout + 100, TimeUnit.MILLISECONDS))
   }
 
+  private def staticJoinGroupWithPersistence(groupId: String,
+                                 memberId: String,
```
abbccdda: nit: alignment

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -1040,21 +1040,36 @@ class GroupCoordinator(val brokerId: Int,
...
-          error = Errors.NONE))
+        // check if group's selectedProtocol of next generation will change, if not, simply store group to persist the
+        // updated static member, if yes, rebalance should be triggered to let the group's assignment and selectProtocol consistent
+        val selectedProtocolOfNextGeneration = group.selectProtocol
+        if (group.protocolName.contains(selectedProtocolOfNextGeneration)) {
+          info(s"Static member which joins during Stable stage and doesn't affect selectProtocol will not trigger rebalance.")
+          val groupAssignment: Map[String, Array[Byte]] = group.allMemberMetadata.map(member => member.memberId -> member.assignment).toMap
+          groupManager.storeGroup(group, groupAssignment, error => {
+            group.inLock {
+              if (error != Errors.NONE) {
```
abbccdda: Should we reply join failure if the persistence of the group metadata failed?

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -1040,21 +1040,36 @@ class GroupCoordinator(val brokerId: Int,
...
+            // We want to avoid current leader performing trivial assignment while the group
+            // is in stable stage, because the new assignment in leader's next sync call
+            // won't be broadcast by a stable group. This could be guaranteed by
+            // always returning the old leader id so that the current leader won't assume itself
+            // as a leader based on the returned message, since the new member.id won't match
+            // returned leader id, therefore no assignment will be performed.
+            leaderId = currentLeader,
+            error = Errors.NONE))
+        } else {
+          maybePrepareRebalance(group, s"Group's selectedProtocol will change because static member ${member.memberId} with instance id $groupInstanceId joined with change of protocol")
```
abbccdda, line 1093: Could you elaborate why this case is possible? We do have checks for `!group.supportsProtocols(protocolType, MemberMetadata.plainProtocolSet(protocols)` in the caller, so if the group protocol is incompatible, won't we just reject the rejoin?

feyman2016: For example, if previously there are one leader + one follower(both static) in a group, the protocols are: `List(("range", metadata), ("roundrobin", metadata))` for both, and current selected protocol is `range`, if later the follower join again with protocols: `List(("roundrobin", metadata))`, and the selectedProtocol should be `roundrobin`, now, the selectedProtocol and the actual assignment is not consistent, here I let it rebalance to make sure that the selectedProtocol and actual assignment are consistent. On the other way around, if we don't do the rebalance, we cannot successfully persist the group since this line `val metadata = memberMetadata.metadata(protocol)` in `kafka.coordinator.group.GroupMetadataManager#groupMetadataValue` will fail. Test `staticMemberRejoinWithUnknownMemberIdAndChangeOfProtocolWithSelectedProtocolChanged ` is for this case.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
feyman2016: Indeed, here need to be revised, will update~

core/src/test/scala/unit/kafka/coordinator/group/GroupCoordinatorTest.scala
feyman2016: Fixed, thanks!

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
vvcephei: Hi @feyman2016 , I just noticed this PR. Thanks for picking this up!

I'm wondering something along similar lines; should we move the whole join callback inside this block? It seems like that's the way we can delay the join group response until after the update is actually persisted.

feyman2016: Thanks for the review, @vvcephei . You're right for the current commit, but as @abbccdda mentioned, the current commit does have issues, I'm revising it to `reset` the `static member` in the group and return the error if `storeGroup` encountered any errors. Code is ready, but met some local issues when writing unittests, will update once it's ready.

vvcephei:
Hi @feyman2016 and @abbccdda ,

I hope you've both been well. I just wanted to check in on the status of this PR, since we still have a flaky test failure related to it.

Thanks,
-John

feyman2016:
Hi, @vvcephei and @abbccdda ,sorry this PR takes a very long time,  just updated the PR, could you help to review? Thanks!

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
feyman2016: @vvcephei  @abbccdda Revised the persistence failure handling logic, now it will revert the member id update in the groupMetaData if any persistence error encountered and call the responseCallback with the returned error.

core/src/test/scala/unit/kafka/coordinator/group/GroupCoordinatorTest.scala
```diff
@@ -3789,6 +3844,41 @@ class GroupCoordinatorTest {
...
 
+    EasyMock.expect(replicaManager.getMagic(EasyMock.anyObject())).andReturn(Some(RecordBatch.MAGIC_VALUE_V1)).anyTimes()
+    EasyMock.replay(replicaManager)
+
+    groupCoordinator.handleJoinGroup(groupId, memberId, groupInstanceId,
+      requireKnownMemberId, "clientId", "clientHost", rebalanceTimeout, sessionTimeout, protocolType, protocols, responseCallback)
+    responseFuture
+  }
+
+  private def sendStaticJoinGroupWithPersistence(groupId: String,
```
abbccdda, line 3855: Seems the helper only gets called once?

feyman2016: Yes, exactly, should I make it inline ?

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -1037,24 +1037,52 @@ class GroupCoordinator(val brokerId: Int,
...
+            protocolType = group.protocolType,
+            protocolName = group.protocolName,
+            // We want to avoid current leader performing trivial assignment while the group
+            // is in stable stage, because the new assignment in leader's next sync call
+            // won't be broadcast by a stable group. This could be guaranteed by
+            // always returning the old leader id so that the current leader won't assume itself
+            // as a leader based on the returned message, since the new member.id won't match
+            // returned leader id, therefore no assignment will be performed.
+            leaderId = currentLeader,
+            error = Errors.NONE))
```
vvcephei: Hey @feyman2016 ,

I'm still wondering if this block should be inside the callback of `storeGroup`. Otherwise, we would already have sent the response to the client before the `storeGroup` completes, and the client will never see the error response on L1058.

Or did we specifically decide to make the `storeGroup` call best effort?

(cc @abbccdda )

vvcephei (commented):
Thanks @feyman2016 , just one question :)

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
feyman2016: @vvcephei I think it make sense, will put the call of `group.maybeInvokeJoinCallback` in the callback, thanks!

feyman2016: @vvcephei Updated as proposed, thanks!

feyman2016:
@abbccdda @vvcephei Updated the PR, thanks!

vvcephei (approved):
Thanks @feyman2016 ! This looks good to me. I'll leave it to @abbccdda to merge.

feyman2016:
Thanks! @vvcephei

vvcephei:
I've just merged trunk into this branch in an effort to get Jenkins to build.

I tried just re-running the Jenkins job, but it kept erroring out on some gradle/environment variables. I noticed that other recent PRs are building, so theorized that if we merge trunk in, we might get the build to work.

vvcephei:
FYI, I just got a successful build of this locally with master merged in.

vvcephei:
By the way, it was true that trunk was broken. Fixed by: https://github.com/apache/kafka/commit/2db67db8e1329cb2e047322cff81d97ff98b4328

feyman2016:
@vvcephei Thanks for the help, fyi, I also tried to build locally with the apache/trunk merged, it succeeded.

vvcephei:
The test failure was unrelated: Build / JDK 15 / org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]

vvcephei:

vvcephei:
Thanks for this fix, @feyman2016 !

vvcephei:
Cherry-picked to 2.6 (cc @mimaison )

vvcephei:
Cherry-picked to 2.5

feyman2016:
Thanks a lot for the review and merge @abbccdda @vvcephei!

PR #9272, opened 2020-09-09
KAFKA-10458; Updating controller quota does not work since Token Bucket

dajac:
This PR fixes two issues that have been introduced by https://github.com/apache/kafka/pull/9114.

* When I switched the metric from `Rate` to `TokenBucket` in the `ControllerMutationQuotaManager`, I have mixed up the metrics. That broke the quota update path.

* When a quota is updated, the `ClientQuotaManager` updates the `MetricConfig` of the `KafkaMetric`. That update was not reflected into the `Sensor` so the `Sensor` was still using the `MetricConfig` that it has been created with.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
@@ -63,11 +63,11 @@ public final class Sensor {
         }
 
         public Stat stat() {
-            return this.stat;
+            return stat;
         }
 
         public MetricConfig config() {
-            return this.configSupplier.get();
+            return configSupplier.get();
         }
     }
 
--- a/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
@@ -16,6 +16,7 @@
  */
 package org.apache.kafka.common.metrics;
 
+import java.util.function.Supplier;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.metrics.CompoundStat.NamedMeasurable;
 import org.apache.kafka.common.metrics.stats.TokenBucket;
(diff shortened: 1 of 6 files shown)


clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
```diff
@@ -53,12 +54,20 @@
...
 
-        StatAndConfig(Stat stat, MetricConfig config) {
+        StatAndConfig(Stat stat, Supplier<MetricConfig> configSupplier) {
             this.stat = stat;
-            this.config = config;
+            this.configSupplier = configSupplier;
+        }
+
+        public Stat stat() {
+            return this.stat;
```
apovzner: nit: `this` is not really needed, since it's not ambiguous here as in constructor,

apovzner (approved):
Thanks for finding a way to fix this without changing Sensor and/or Metrics API. Looks great!

clients/src/main/java/org/apache/kafka/common/metrics/Sensor.java
```diff
@@ -53,12 +54,20 @@
...
+        }
+
+        public MetricConfig config() {
+            return this.configSupplier.get();
```
rajinisivaram: I guess if you are removing `this.` above, you could remove it here as well for consistency.

rajinisivaram: I guess if you are removing `this.` above, you could remove it here as well for consistency.

rajinisivaram (approved):
@dajac Thanks for the PR, LGTM. I will merge once you get a chance to address @apovzner 's comment.

dajac:
@apovzner @rajinisivaram Thanks for your feedback. I have addressed your comments.

rajinisivaram:
@dajac Thanks for the update, test failures not related, merging to trunk.

PR #9273, opened 2020-09-09
KAFKA-9331: changes for Streams uncaught exception handler

wcarlson5:
Adding an option to shutdown an entire streams application for a Streams thread

@cadonna


streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -1436,6 +1447,16 @@ public void onAssignment(final Assignment assignment, final ConsumerGroupMetadat
...
                 break;
+            case 8:
+                validateActiveTaskEncoding(partitions, info, logPrefix);
+
+                activeTasks = getActiveTasks(partitions, info);
+                partitionsByHost = info.partitionsByHost();
+                standbyPartitionsByHost = info.standbyPartitionByHost();
+                topicToPartitionInfo = getTopicPartitionInfo(partitionsByHost);
+                encodedNextScheduledRebalanceMs = info.nextRebalanceMs();
+                //recive the shutdown then call the request close
```
wcarlson5: the error is read in above so just need to handle like version 7

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -1053,13 +1067,13 @@ ConsumerRebalanceListener rebalanceListener() {
 
     Consumer<byte[], byte[]> restoreConsumer() {
         return restoreConsumer;
-    };
```
wcarlson5: extra ; removed

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -1369,7 +1381,6 @@ public void onAssignment(final Assignment assignment, final ConsumerGroupMetadat
 
         final AssignmentInfo info = AssignmentInfo.decode(assignment.userData());
         if (info.errCode() != AssignorError.NONE.code()) {
-            // set flag to shutdown streams app
             assignmentErrorCode.set(info.errCode());
```
wcarlson5, line 1380: records the error to be handled in onPartitionsAssigned with the other possible error

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -1229,4 +1233,13 @@ boolean needsInitializationOrRestoration() {
     public void setPartitionResetter(final java.util.function.Consumer<Set<TopicPartition>> resetter) {
         this.resetter = resetter;
     }
+
+    public void flagForShutdownRequest(){
+        this.shutdownRequested.set(2);
```
wcarlson5: error code for INCOMPLETE_SOURCE_TOPIC_METADATA is 1, SHUTDOWN_REQUESTED is 2

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/AssignmentInfo.java
```diff
@@ -186,6 +186,15 @@ public ByteBuffer encode() {
...
                     out.writeLong(nextRebalanceMs);
                     break;
+                case 8:
+                    out.writeInt(usedVersion);
+                    out.writeInt(commonlySupportedVersion);
+                    encodeActiveAndStandbyTaskAssignment(out);
+                    encodeActiveAndStandbyHostPartitions(out);
+                    out.writeInt(errCode);
+                    out.writeLong(nextRebalanceMs);
+                    out.writeInt(0);
```
wcarlson5: no error in this case, we only want to trigger shutdown form subscriptionUserData

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/AssignmentInfo.java
```diff
@@ -360,6 +369,16 @@ public static AssignmentInfo decode(final ByteBuffer data) {
...
                     break;
+                case 8:
+                    commonlySupportedVersion = in.readInt();
+                    assignmentInfo = new AssignmentInfo(usedVersion, commonlySupportedVersion);
+                    decodeActiveTasks(assignmentInfo, in);
+                    decodeStandbyTasks(assignmentInfo, in);
+                    decodeActiveAndStandbyHostPartitions(assignmentInfo, in);
+                    assignmentInfo.errCode = in.readInt();
+                    assignmentInfo.nextRebalanceMs = in.readLong();
+                    in.readInt();
```
wcarlson5: I don't know if this is necessary but its best to get the error out of the buffer

streams/src/test/java/org/apache/kafka/streams/integration/AppShutdownIntegrationTest.java
```diff
@@ -0,0 +1,140 @@
...
+        );
+
+
+        try (final KafkaStreams kafkaStreams = new KafkaStreams(builder.build(), properties)) {
+            final CountDownLatch latch = new CountDownLatch(1);
+
+            kafkaStreams.start();
+            produceMessages(0L, inputTopic);
+
+            latch.await(10, TimeUnit.SECONDS);
```
wcarlson5: should replace with a shutdown hook latch with counter instead

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -566,10 +567,22 @@ void runLoop() {
                 }
             } catch (final TaskMigratedException e) {
                 handleTaskMigrated(e);
+            } catch (final ShutdownRequestedException e){
+                handleShutdownRequest(e);
```
wcarlson5: The interruption in this location caused ...


```
TaskManager
	MetadataState:
	Tasks:
 died (org.apache.zookeeper.server.NIOServerCnxnFactory:92)
java.lang.IllegalStateException: Illegal state SUSPENDED while completing restoration for active task 0_0
	at org.apache.kafka.streams.processor.internals.StreamTask.completeRestoration(StreamTask.java:245)
	at org.apache.kafka.streams.processor.internals.TaskManager.tryToCompleteRestoration(TaskManager.java:487)
	at org.apache.kafka.streams.processor.internals.StreamThread.runOnce(StreamThread.java:675)
	at org.apache.kafka.streams.processor.internals.StreamThread.runLoop(StreamThread.java:554)
	at org.apache.kafka.streams.processor.internals.StreamThread.run(StreamThread.java:513)
```

Not sure that is okay, the test still runs. If its is shutting down maybe it does matter and I'm not sure where else to put it

streams/src/test/java/org/apache/kafka/streams/integration/AppShutdownIntegrationTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Override
+    public void init(final ProcessorContext context) {
+//        throw new ShutdownRequestedException("integration test");
+    }
+
+    @Override
+    public void process(final Object key, final Object value) {
+        valueList.add(new KeyValue<>(key, value));
+//        throw new ShutdownRequestedException("integration test");
```
wcarlson5: uncomment for test to pass, still will not cause a shutdown until request close is added to the rebalance listener

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsRebalanceListener.java
```diff
@@ -54,6 +54,9 @@ public void onPartitionsAssigned(final Collection<TopicPartition> partitions) {
         if (assignmentErrorCode.get() == AssignorError.INCOMPLETE_SOURCE_TOPIC_METADATA.code()) {
             log.error("Received error code {}", assignmentErrorCode.get());
             throw new MissingSourceTopicException("One or more source topics were missing during rebalance");
+        }else if(assignmentErrorCode.get() == AssignorError.SHUTDOWN_REQUESTED.code()){
+            //throw new ShutdownRequestedException("onPartition assigned"); //TODO: receive request and call requestClose()
+            //requestClose();
```
wcarlson5: until the shut down application form the stream thread functionality is added this will not shutdown the app

wcarlson5 (commented):
This is the first way I found to send a shutdown signal. It seems the other error (INCOMPLETE_SOURCE_TOPIC_METADATA) is discovered by the leader in assign. Therefore it does not have to use the subscription Info metadata in a rebalance and also is not included in the schema. This is why I had to make a schema version 8.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -242,7 +242,8 @@ public ByteBuffer subscriptionUserData(final Set<String> topics) {
             LATEST_SUPPORTED_VERSION,
             taskManager.processId(),
             userEndPoint,
-            taskManager.getTaskOffsetSums())
+            taskManager.getTaskOffsetSums(),
+            taskManager.isShutdownRequested())
```
wcarlson5: tests are failing because the Mock of test Manager is setting the shutdownRequested flag to null. not sure If I should fix the Mock or add a redundant null check and fix

vvcephei: This works for the POC, but for the real change, I'd recommend using an independent flag, such as an AtomicBoolean.

In the past, we used to piggy-back all kinds of inappropriate flags through the TaskManager just because it happened to be shared between these components. But it ultimately made TaskManager pretty hard to maintain. Better to just keep orthogonal things independent.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
vvcephei: If it's identical, you can just do:
```
case 7:
case 8:
    validateActive....
    ...
    break;
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/AssignmentInfo.java
vvcephei: What is this for? It looks like the actual "shutdown" error code is just encoded in the `errCode` field.

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfo.java
```diff
@@ -82,11 +83,21 @@ public SubscriptionInfo(final int version,
...
                             final Map<TaskId, Long> taskOffsetSums) {
+        this(version, latestSupportedVersion, processId, userEndPoint, taskOffsetSums, new AtomicInteger(0));
+    }
+
+    public SubscriptionInfo(final int version,
+                            final int latestSupportedVersion,
+                            final UUID processId,
+                            final String userEndPoint,
+                            final Map<TaskId, Long> taskOffsetSums,
+                            final AtomicInteger shutdownRequested) {
```
vvcephei: It seems like we can look at this field two ways:
1. if we're really just flagging "shutdownRequested or not", then a `boolean` is fine.
2. if we're creating a general facility for members to send error codes to the assignor, then we should call this field "errorCode" instead, and an `int` is fine.

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfo.java
```diff
@@ -96,6 +107,14 @@ public SubscriptionInfo(final int version,
             data.setLatestSupportedVersion(latestSupportedVersion);
         }
 
+        if (version >= 8) {
+            if (shutdownRequested == null) { //check is required for testing because the mock sets it to null
```
vvcephei: We should instead mock realistic values. Having extra branches in the code just to handle the default behavior of EasyMock creates maintenance problems.

streams/src/main/resources/common/message/SubscriptionInfoData.json
```diff
@@ -48,6 +48,11 @@
       "versions": "2+",
       "type": "bytes"
     },
+    {
+      "name": "shutdownRequested",
+      "versions": "8+",
+      "type": "int32"
```
vvcephei: 32 bits is a lot of possible error codes. Note: these numbers are encoding enum values. According to the java spec, you can't have more than 2^16 members a class. It looks like some people actually experimented with how many enum values they could add before the compiler fails, and it tops out at 2746 values. (https://stackoverflow.com/questions/4468388/maximum-number-of-enum-elements-in-java)

Either way, an int16 (`short`) should be more than enough codes.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
vvcephei: Interesting. This looks like it's probably just a consequence of the way the test is set up. Some debugging should clear it up.

vvcephei (commented):
Hey @wcarlson5 , thanks for this. I know this is just a POC, but I left some comments in the hopes that it helps you out.

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfo.java
wcarlson5: Agree. I don't want to leave it there but until I fix easy mock I want it to pass

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
wcarlson5: that makes sense, when you mean independent flag would you recommend adding it to the assignorConfiguration like the assignmentErrorCode?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
wcarlson5: good call

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/AssignmentInfo.java
wcarlson5: yep not actually necessary, I got confused between AssignmentInfo and Subscription info because they use the same protocol version but use different schemas... i think

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfo.java
wcarlson5: I've changed it to a `bool` for now. I think it fits this use case better and if want to expand to use other codes it would be relatively low effort to make it an `int`. What fo you think would be better?

Edit: I was going to do that but checkstyle won't let me import AtomicBoolean, so Ive rolled the shutdown the assignorError all the way, which is probably better than the half way method I was using

wcarlson5 (commented):
I changed the shutdownRequested "flag" to use the AssignorError completely as it was already a value in the Enum.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
vvcephei: Yep, that's what I meant.

vvcephei: I'd fix the mock. You can use `expect().andStubReturn()` to just get it to return "no error" every time.

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/AssignmentInfo.java
vvcephei: Cool. Yes, they have different schemas. SubscriptionInfo has been migrated to use an explicit schema, while AssignmentInfo still uses hand-coded de/serialization code. Never got around to migrating AssignmentInfo.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -931,6 +932,22 @@ public synchronized boolean close(final long timeout, final TimeUnit timeUnit) {
         return close(timeoutMs);
     }
 
+    /**
+     * Attempts to shutdown an application.
+     * If there is an alive StreamThread it will succeed, if there is not it will fail
+     *
+     * @return Will return true if shutdown is initiated false if it is not possible.
+     */
+    public boolean shutdownApplication() {
```
wcarlson5: This is the idea for a shutdown that users can initiate without throwing an exception. It will still need at least one alive stream thread

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -570,6 +581,21 @@ void runLoop() {
         }
     }
 
+    public void initiateShutdown(){
+        shutdownRequested.set(true);
```
wcarlson5: can not directly call sendShutdownRequest because the consumer should only accessed from the stream thread

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+    StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse handle(final Exception exception);
+
+    /**
+     * Enumeration that describes the response from the exception handler.
+     */
+    enum StreamsUncaughtExceptionHandlerResponse {
+
+
+        SHUTDOWN_STREAM_THREAD(0, "SHUTDOWN_STREAM_THREAD"),
+        REPLACE_STREAM_THREAD(1, "REPLACE_STREAM_THREAD"),
```
wcarlson5: This may need to be removed until 663 get merged

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -570,6 +581,49 @@ void runLoop() {
...
+            log.error("Encountered the following exception during processing " +
+                    "and the thread is going to shut down: ", e);
+            return;
+        }
+        switch (action) {
+            case SHUTDOWN_STREAM_THREAD:
+                log.error("Encountered the following exception during processing " +
+                        "and the thread is going to shut down: ", e);
+                break;
+            case REPLACE_STREAM_THREAD:
```
wcarlson5: dependent on 663

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -570,6 +581,49 @@ void runLoop() {
         }
     }
 
+    public void setStreamsUncaughtExceptionHandler(final StreamsUncaughtExceptionHandler streamsUncaughtExceptionHandler, final KafkaStreams kafkaStreams) {
```
wcarlson5: I am looking into using a wrapper and call back so we do not have to introduce a KafkaStreams object here but until I figure that out this will be stop gap.

though this will not work in the rebalance listener as it needs the reference as well. maybe this is the only way idk yet. @cadonna WDYT?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -570,6 +581,49 @@ void runLoop() {
...
+                        "and the thread is going to shut down: ", e);
+                break;
+            case REPLACE_STREAM_THREAD:
+                log.error("Encountered the following exception during processing " +
+                        "and the the stream thread will be replaced: ", e);
+                break;
+            case SHUTDOWN_KAFKA_STREAMS_CLIENT:
+                log.error("Encountered the following exception during processing " +
+                        "and the client is going to shut down: ", e);
+                kafkaStreams.close(Duration.ZERO);
```
wcarlson5: currently leaves client state in not running instead of error.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -529,8 +541,7 @@ public void run() {
                 }
             }
 
-            log.error("Encountered the following exception during processing " +
-                "and the thread is going to shut down: ", e);
+            handleStreamsUncaughtException(e);
             throw e;
```
wcarlson5, line 553: this will trigger the old uncaught exception handler

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -529,8 +541,7 @@ public void run() {
                 }
             }
 
-            log.error("Encountered the following exception during processing " +
-                "and the thread is going to shut down: ", e);
+            handleStreamsUncaughtException(e);
```
wcarlson5: need to verify that new handler is set, if we make it so you can only have either the new handler or the old set

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsRebalanceListener.java
```diff
@@ -54,6 +55,9 @@ public void onPartitionsAssigned(final Collection<TopicPartition> partitions) {
         if (assignmentErrorCode.get() == AssignorError.INCOMPLETE_SOURCE_TOPIC_METADATA.code()) {
             log.error("Received error code {}", assignmentErrorCode.get());
             throw new MissingSourceTopicException("One or more source topics were missing during rebalance");
+        } else if (assignmentErrorCode.get() == AssignorError.SHUTDOWN_REQUESTED.code()) {
+            //throw new ShutdownRequestedException("onPartition assigned"); //TODO: receive request and call requestClose()
+            streamThread.kafkaStreams.close(Duration.ZERO);
```
wcarlson5, line 60: blocking, need to find a non blocking call

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -364,6 +369,31 @@ public void setUncaughtExceptionHandler(final Thread.UncaughtExceptionHandler eh
...
     }
 
+    /**
+     * Set the handler invoked when a {@link StreamsConfig#NUM_STREAM_THREADS_CONFIG internal thread} abruptly
+     * terminates due to an uncaught exception.
+     *
+     * @param eh the uncaught exception handler of type {@link StreamsUncaughtExceptionHandler} for all internal threads; {@code null} deletes the current handler
+     * @throws IllegalStateException if this {@code KafkaStreams} instance is not in state {@link State#CREATED CREATED}.
+     */
+    public void setUncaughtExceptionHandler(final StreamsUncaughtExceptionHandler eh) {
```
wcarlson5: when we set a new handler should we reset the old one so they can't both be triggered?
Same for the other way around

wcarlson5 (commented):
I still need to add tests for the different options. I might want to make a system test for the shutdown as well as the integration tests.

There are a few things I still need improve but can probably get some feedback now

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -436,6 +496,8 @@ private void maybeSetError() {
             }
 
             if (setState(State.ERROR)) {
+                metrics.close();
```
wcarlson5: If the client is in error we should close these threads, for the same reason we are closing them for this kip. So might as well close them now.

This is also ensures they are closed on all clients after a shutdown

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -364,6 +368,62 @@ public void setUncaughtExceptionHandler(final Thread.UncaughtExceptionHandler eh
...
+                                                                                                                   final StreamsUncaughtExceptionHandler streamsUncaughtExceptionHandler) {
+        final StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse action = streamsUncaughtExceptionHandler.handle(e);
+        switch (action) {
+            case SHUTDOWN_STREAM_THREAD:
+                log.error("Encountered the following exception during processing " +
+                        "and the thread is going to shut down: ", e);
+                break;
+            case REPLACE_STREAM_THREAD:
+                log.error("Encountered the following exception during processing " +
+                        "and the the stream thread will be replaced: ", e); //TODO: add then remove, wait until 663 is merged
```
wcarlson5: need to wait to @cadonna to merge 663 before this is added. Can do a partial implantation then add later though

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -550,6 +561,10 @@ void runLoop() {
         // until the rebalance is completed before we close and commit the tasks
         while (isRunning() || taskManager.isRebalanceInProgress()) {
             try {
+                if (shutdownRequested.get()) {
+                    sendShutdownRequest(shutdownTypeRequested);
+                    return;
```
wcarlson5: if we are shutting down may as well start now instead of waiting for the rebalance. this will stop every thread

wcarlson5 (commented):
Hopefully the implantation is done. I need to figure out what I can unit test and write a system test for the shutdown

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -346,6 +347,9 @@ public void setStateListener(final KafkaStreams.StateListener listener) {
      *
      * @param eh the uncaught exception handler for all internal threads; {@code null} deletes the current handler
      * @throws IllegalStateException if this {@code KafkaStreams} instance is not in state {@link State#CREATED CREATED}.
+     *
+     * @Deprecated user {@link KafkaStreams#setUncaughtExceptionHandler(StreamsUncaughtExceptionHandler)} instead.
```
vvcephei:
```suggestion
     * @Deprecated Since 2.7.0. Use {@link KafkaStreams#setUncaughtExceptionHandler(StreamsUncaughtExceptionHandler)} instead.
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -364,6 +368,62 @@ public void setUncaughtExceptionHandler(final Thread.UncaughtExceptionHandler eh
...
     }
 
+    /**
+     * Set the handler invoked when a {@link StreamsConfig#NUM_STREAM_THREADS_CONFIG internal thread} abruptly
+     * terminates due to an uncaught exception.
+     *
+     * @param eh the uncaught exception handler of type {@link StreamsUncaughtExceptionHandler} for all internal threads; {@code null} deletes the current handler
+     * @throws IllegalStateException if this {@code KafkaStreams} instance is not in state {@link State#CREATED CREATED}.
+     */
+    public void setUncaughtExceptionHandler(final StreamsUncaughtExceptionHandler eh) {
```
vvcephei:
```suggestion
    public void setUncaughtExceptionHandler(final StreamsUncaughtExceptionHandler uncaughtExceptionHandler) {
```

We prefer to resist the urge to abbreviate, especially in the public-facing APIs.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -364,6 +368,62 @@ public void setUncaughtExceptionHandler(final Thread.UncaughtExceptionHandler eh
         }
     }
 
+    /**
+     * Set the handler invoked when a {@link StreamsConfig#NUM_STREAM_THREADS_CONFIG internal thread} abruptly
+     * terminates due to an uncaught exception.
```
vvcephei:
```suggestion
     * Set the handler invoked when a {@link StreamsConfig#NUM_STREAM_THREADS_CONFIG internal thread}
     * throws an unexpected exception. These might be exceptions indicating rare bugs in Kafka Streams, or they
     * might be exceptions thrown by your code, for example a NullPointerException thrown from your processor
     * logic.
     * <p>
     * Note, this handler must be threadsafe, since it will be shared among all threads, and invoked from any
     * thread that encounters such an exception.
```

I think it's wrong to say that this is invoked when the thread abruptly terminates, because it's not. That's how the JVM handler works, but we're actually executing this handler while the thread is still running, and in fact that thread itself is what calls the handler.

It also seemed appropriate to elaborate a little more on the usage of this method.

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.errors;
+
+public interface StreamsUncaughtExceptionHandler {
+    /**
+     * Inspect a record and the exception received.
+     * @param exception the actual exception
+     */
+    StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse handle(final Exception exception);
```
vvcephei: Let's tweak this API to Throwable:
```suggestion
    StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse handle(final Throwable exception);
```

Here's a good explanation of why: https://stackoverflow.com/questions/2274102/difference-between-using-throwable-and-exception-in-a-try-catch

The benefit is that we could handle `Error`s as well as `Exception`s. However, this comes with the obligation that we should not continue to use the thread after an Error occurs. I think we can deal with this restriction reasonably.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
vvcephei: Let's just comment out this whole enum value until 663 is implemented, that way we won't have a TODO hanging around in case 663 doesn't make the release for some reason.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -364,6 +368,62 @@ public void setUncaughtExceptionHandler(final Thread.UncaughtExceptionHandler eh
...
+            case REPLACE_STREAM_THREAD:
+                log.error("Encountered the following exception during processing " +
+                        "and the the stream thread will be replaced: ", e); //TODO: add then remove, wait until 663 is merged
+                break;
+            case SHUTDOWN_KAFKA_STREAMS_CLIENT:
+                log.error("Encountered the following exception during processing " +
+                        "and the client is going to shut down: ", e);
+                for (final StreamThread streamThread: threads) {
+                    streamThread.shutdown();
+                }
```
vvcephei: It seems safer to just call the nonblocking close method:
```suggestion
                close(Duration.ZERO);
```
That way, it'll properly set the state, stop the cleaner thread, etc.

vvcephei: Oh, I forgot; the reason you're doing it this way is to transition to ERROR, not actually shut down, right?

In that case, it seems pretty odd to call this option "shut down", since it doesn't actually _shut down_, it only kills all the threads, leaving the final "shut down" as an exercise to the user.

If I recall correctly, the preference of the group was in favor of this behavior, in which case, I'd advocate for a different name. Maybe just `STOP_STREAM_THREAD`, `STOP_ALL_STREAM_THREADS`, and `STOP_ALL_STREAM_THREADS_IN_CLUSTER`.

I've been on the fence about whether I should leave this feedback or not, but decided to go ahead and pass it on to you because I just got confused by the names, despite having recently participating in that discussion. So it seems likely that users would also be confused and think that we're doing the wrong thing by not actually shutting down the client.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
vvcephei: Actually, now that I have a more in-depth picture of what is going on, I disagree. I think we should leave these threads running until users actually call `KafkaStreams.close`, since the could alternatively add more threads (via KIP-633) to transition back from ERROR state into RUNNING, at which point, we'll be sorry that we killed these threads, right?

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
vvcephei: +1 I think we should at least comment it out until that other PR is actually in the codebase.

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+        SHUTDOWN_STREAM_THREAD(0, "SHUTDOWN_STREAM_THREAD"),
+        REPLACE_STREAM_THREAD(1, "REPLACE_STREAM_THREAD"),
+        SHUTDOWN_KAFKA_STREAMS_CLIENT(2, "SHUTDOWN_KAFKA_STREAMS_CLIENT"),
+        SHUTDOWN_KAFKA_STREAMS_APPLICATION(3, "SHUTDOWN_KAFKA_STREAMS_APPLICATION");
+
+
+        /** an english description of the api--this is for debugging and can change */
+        public final String name;
+
+        /** the permanent and immutable id of an API--this can't change ever */
```
vvcephei: Thanks for clarifying this.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
vvcephei: I'm not sure this is right, actually. If we wanted to use the current thread to send the "poison pill" subscription, it needs to keep running and call poll again.

Maybe instead we should have a default implementation of the uncaughtExceptionHandler that invokes the legacy one and then returns `SHUTDOWN_STREAM_THREAD`, and then the implementation of that case in KafkaStreams would be to re-throw the exception.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
vvcephei: I can't tell; are we guaranteed to actually send the joinGroup request by now?

Maybe it's safer to just keep running this loop until we get back the "you should shut down" response assignment. We _could/should_ add a condition into `runOnce` so that we don't actually process anything once we have `shutdownRequested` set.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsRebalanceListener.java
```diff
@@ -54,6 +54,8 @@ public void onPartitionsAssigned(final Collection<TopicPartition> partitions) {
         if (assignmentErrorCode.get() == AssignorError.INCOMPLETE_SOURCE_TOPIC_METADATA.code()) {
             log.error("Received error code {}", assignmentErrorCode.get());
             throw new MissingSourceTopicException("One or more source topics were missing during rebalance");
+        } else if (assignmentErrorCode.get() == AssignorError.SHUTDOWN_REQUESTED.code()) {
+            streamThread.shutdown(); //TODO: 663 should set client to error if all streams are dead
```
vvcephei: didn't quite follow this TODO

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfo.java
vvcephei: Cool, so maybe this field should be called something else, like "subscription flag"?

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -616,7 +616,7 @@ public void shouldThrowExceptionSettingUncaughtExceptionHandlerNotInCreateState(
         final KafkaStreams streams = new KafkaStreams(getBuilderWithSource().build(), props, supplier, time);
         streams.start();
         try {
-            streams.setUncaughtExceptionHandler(null);
+            streams.setUncaughtExceptionHandler((Thread.UncaughtExceptionHandler) null);
```
vvcephei: Should we also test the behavior of setting the new handler to `null`? And actually, why is `null` allowed now, but it wasn't before?

streams/src/test/java/org/apache/kafka/streams/integration/AppShutdownIntegrationTest.java
```diff
@@ -0,0 +1,152 @@
...
+                        mkEntry(StreamsConfig.NUM_STANDBY_REPLICAS_CONFIG, "5"),
+                        mkEntry(StreamsConfig.ACCEPTABLE_RECOVERY_LAG_CONFIG, "6"),
+                        mkEntry(StreamsConfig.MAX_WARMUP_REPLICAS_CONFIG, "7"),
+                        mkEntry(StreamsConfig.PROBING_REBALANCE_INTERVAL_MS_CONFIG, "480000")
+                )
+        );
+
+        try (final KafkaStreams kafkaStreams = new KafkaStreams(builder.build(), properties)) {
+            final CountDownLatch latch = new CountDownLatch(1);
+            kafkaStreams.setUncaughtExceptionHandler(exception -> StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse.SHUTDOWN_KAFKA_STREAMS_APPLICATION);
```
vvcephei: Feel free to import `SHUTDOWN_KAFKA_STREAMS_APPLICATION` and save like 50 characters on this line ;)

vvcephei (commented):
Hey @wcarlson5 , thanks for this PR!

I made a pass on all but the incomplete test.

In addition to my comments so far, I have a somewhat fancy question for you:

What will happen if we are in the middle of a rolling upgrade to this new version of the code and there is an exception, and there is an exception handler registered that says to send the poison pill subscription? Because of the rolling upgrade in progress, we would be in a situation where we have negotiated a protocol version less than 8 (for example, 7). In that case, our subscription encoder would leave that field out and not send it to the leader (because the leader might really be running version 7 and not understand it). Then, the poison pill will be ignored.

I could see two approaches here:
1. Just leave the local flag set and continuously try rejoining the group with the poison pill. Hopefully, the group will eventually all move up to version 8, at which point, the poison pill will take effect. This might not work as expected, as the constant rejoins might actually prevent other instances from getting upgraded.
2. Just print an error message that says "Sorry, this option isn't available until the whole cluster is upgraded to version 2.7+. Falling back to stop all local threads instead."

WDYT?

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
wcarlson5: Hmmm that make sense mostly. However it seems to handle the application shutdown case we need to the keep the thread alive to trigger a rebalance. Could we somehow remove the option of shutdown for some cases?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsRebalanceListener.java
wcarlson5: that is legacy, removed

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
wcarlson5: The test isn't really related to null its check to see if the client is in the correct state to set the handler. 

The reason I added the cast is because the complier complains otherwise. I added the same test for the new handler below.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: you are right. It caused a problem with Clients with a single thread as well I moved it into run loop and this resolved the problems

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Maybe when we transition out of error we should restart them. I don't think the metrics thread really matters, but I closed the the state cleaner bc @cadonna thought it might cause state loss we would want to avoid

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: you are right again. See previous reponse

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: should I also change the old handler? it uses the same name

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfo.java
wcarlson5: sure

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
vvcephei: Oh, sure. Now I know why you picked this name :)

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
vvcephei: That's a good question. Maybe we could just document that that option is an unsuitable response to an Error, and also log an `ERROR` message if you select it in response to an Error. It's not _always_ bad to ignore an Error, but it usually is. We can leave it to users to decide what they want to do.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
vvcephei: It certainly might. I'm just wary of how far into the uncanny valley we're going here. Streams is going to be put into a state that's very similar to the one that `close()` produces, but not identical. What will then happen when they _do_ call close? What will happen when we realize that something else needs to be done as part of closing the instance (will we even remember that we should consider doing it here as well)?

OTOH, we could instead change direction on the "error-vs-shutdown" debase and just make all these methods call `close(ZERO)` instead. Then, the _real_ close method will be invoked, and Streams will go through a well-known transition through `PENDING_SHUTDOWN` to `NOT_RUNNING`.

It would then be a problem for a later date (after KIP-663) if someone wanted to request that instead the app should stop all running threads so they can manually call "addThread" later.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
vvcephei: Ah, gotcha. Thanks.

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
wcarlson5: That seems to be a fine solution. It will still attempt the shutdown but may fail. As long as we warn I guess it will work

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: We did have it this way in the kip. If we stick to this for now I think that we can clear this up easily when we decide what we want to do with the sates in general in when we take care of the discussion in KIP-663

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I am okay with renaming, but I will wait to for everything else to be cleared up to see if it is still necessary.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -384,9 +390,8 @@ public static StreamThread create(final InternalTopologyBuilder builder,
             builder,
             threadId,
             logContext,
-            assignmentErrorCode,
```
wcarlson5, line 387: this diff is a bit off. The assignmentErrorCode was added in this PR. I didn't change the order of the params for no reason...

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+        public final int id;
+
+        StreamsUncaughtExceptionHandlerResponse(final int id, final String name) {
```
ableegoldman, line 44: What is the `id`? And what is the `name`, is that the thread name? If so, can we clarify this?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -364,6 +370,73 @@ public void setUncaughtExceptionHandler(final Thread.UncaughtExceptionHandler eh
...
+     * thread that encounters such an exception.
+     *
+     * @param streamsUncaughtExceptionHandler the uncaught exception handler of type {@link StreamsUncaughtExceptionHandler} for all internal threads; {@code null} deletes the current handler
+     * @throws IllegalStateException if this {@code KafkaStreams} instance is not in state {@link State#CREATED CREATED}.
+     */
+    public void setUncaughtExceptionHandler(final StreamsUncaughtExceptionHandler streamsUncaughtExceptionHandler) {
+        final StreamsUncaughtExceptionHandler handler = throwable -> handleStreamsUncaughtException(throwable, streamsUncaughtExceptionHandler);
+        synchronized (stateLock) {
+            if (state == State.CREATED) {
+                for (final StreamThread thread : threads) {
```
ableegoldman, line 395: What's our plan for the global thread?  I didn't think of this during the KIP discussion, and sorry if it was brought up there and I just forgot about it. But it seems like we should still give users a non-deprecated way to set a handler for the global thread.

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
ableegoldman: Oh, sorry, I thought this was the `handle` method. Ignore me 😄

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.errors;
+
+public interface StreamsUncaughtExceptionHandler {
+    /**
+     * Inspect a record and the exception received.
+     * @param exception the actual exception
+     */
+    StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse handle(final Throwable exception);
```
ableegoldman: Should we consider also passing in the thread name?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
ableegoldman: If we want to use the same StreamsUncaughtExceptionHandler so users don't have to implement two different handlers, which sounds right to me, then maybe we should add a parameter to the `handle` method to indicate whether it's a global or stream thread. Or split into a separate `handleStreamThreadException` and `handleGlobalThreadException` methods, or something like that (and possibly rename `SHUTDOWN_STREAM_THREAD` to `SHUTDOWN_THREAD`)
WDYT?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
cadonna: I think there is a misunderstanding regarding `metrics.close()`. We should not close the metrics here. We should stop the `rocksDBMetricsRecordingService` with `rocksDBMetricsRecordingService.shutdownNow()`. That is what the KIP says.

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Enumeration that describes the response from the exception handler.
+     */
+    enum StreamsUncaughtExceptionHandlerResponse {
+
+
```
cadonna: nit: I would delete these two line

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+        SHUTDOWN_STREAM_THREAD(0, "SHUTDOWN_STREAM_THREAD"),
+//        REPLACE_STREAM_THREAD(1, "REPLACE_STREAM_THREAD"),
+        SHUTDOWN_KAFKA_STREAMS_CLIENT(2, "SHUTDOWN_KAFKA_STREAMS_CLIENT"),
+        SHUTDOWN_KAFKA_STREAMS_APPLICATION(3, "SHUTDOWN_KAFKA_STREAMS_APPLICATION");
+
```
cadonna, line 33: nit: I would delete this line

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -384,9 +390,8 @@ public static StreamThread create(final InternalTopologyBuilder builder,
...
-            nextScheduledRebalanceMs
-        );
+            nextScheduledRebalanceMs,
+            assignmentErrorCode);
```
cadonna: The closing parenthesis should be in a new line.
```suggestion
            assignmentErrorCode
        );
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -528,12 +536,10 @@ public void run() {
                     throw e;
                 }
             }
-
-            log.error("Encountered the following exception during processing " +
-                "and the thread is going to shut down: ", e);
             throw e;
         } finally {
             completeShutdown(cleanRun);
+
```
cadonna: Why did you add this line? I would remove it.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -566,10 +572,36 @@ void runLoop() {
                 }
             } catch (final TaskMigratedException e) {
                 handleTaskMigrated(e);
+            } catch (final Exception e) {
+                if (streamsUncaughtExceptionHandler != null) {
```
cadonna: I would avoid this `null` check by setting the `streamsUncaughtExceptionHandler` by default to the default uncaught exception handler.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -936,6 +968,8 @@ public void shutdown() {
         }
     }
 
+
+
```
cadonna: Please remove

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -342,6 +347,13 @@ public GroupAssignment assign(final Cluster metadata, final GroupSubscription gr
         // the maximum of the depending sub-topologies source topics' number of partitions
         final Map<Integer, TopicsInfo> topicGroups = taskManager.builder().topicGroups();
 
+        if (shutdownRequested) {
+            return new GroupAssignment(
+                    errorAssignment(clientMetadataMap,
+                            AssignorError.SHUTDOWN_REQUESTED.code())
```
cadonna: You can put all of it in one line.

```suggestion
            return new GroupAssignment(errorAssignment(clientMetadataMap, AssignorError.SHUTDOWN_REQUESTED.code()));
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -342,6 +347,13 @@ public GroupAssignment assign(final Cluster metadata, final GroupSubscription gr
         // the maximum of the depending sub-topologies source topics' number of partitions
         final Map<Integer, TopicsInfo> topicGroups = taskManager.builder().topicGroups();
 
+        if (shutdownRequested) {
```
cadonna: Is there a reason, that prevents you to check the `shutdownRequested` flag in the first step of `Step One`?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -364,6 +370,73 @@ public void setUncaughtExceptionHandler(final Thread.UncaughtExceptionHandler eh
...
+                    if (streamsUncaughtExceptionHandler != null)  {
```
cadonna: I think it is better to throw if the passed in exception handler is `null` and set the default uncaught exception handler in the `StreamThread` constructor.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -617,7 +619,19 @@ public void shouldThrowExceptionSettingUncaughtExceptionHandlerNotInCreateState(
...
+            fail("Should throw IllegalStateException");
+        } catch (final IllegalStateException e) {
+            // expected
+        }
+    }
```
cadonna, line 626: This test misses the verification whether the `StreamThread#setUncaughtExceptionHandler()` is called. This has been missing also before this PR.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -617,7 +619,19 @@ public void shouldThrowExceptionSettingUncaughtExceptionHandlerNotInCreateState(
...
+
+    @Test
+    public void shouldThrowExceptionSettingStreamsUncaughtExceptionHandlerNotInCreateState() {
```
cadonna, line 629: This test misses the verification whether the `StreamThread#setStreamsUncaughtExceptionHandler()` is called. Additionally, a test is missing that tests when `null` is passed into `KafkaStreams#setUncaughtExceptionHandler()`.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -364,6 +370,73 @@ public void setUncaughtExceptionHandler(final Thread.UncaughtExceptionHandler eh
...
+    }
+
+    private StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse handleStreamsUncaughtException(final Throwable e,
+                                                                                                                   final StreamsUncaughtExceptionHandler streamsUncaughtExceptionHandler) {
+        final StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse action = streamsUncaughtExceptionHandler.handle(e);
+        switch (action) {
+            case SHUTDOWN_STREAM_THREAD:
+                log.error("Encountered the following exception during processing " +
+                        "and the thread is going to shut down: ", e);
+                break;
```
cadonna: What about not considering this case here and logging and throwing the exception where the handler is called in `StreamThread`. So we would have all handling logic for this case in one place.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamThreadTest.java
```diff
@@ -226,8 +226,8 @@ private StreamThread createStreamThread(@SuppressWarnings("SameParameterValue")
             0,
             stateDirectory,
             new MockStateRestoreListener(),
-            threadIdx
-        );
+            threadIdx,
+            new AtomicInteger());
```
cadonna: The closing parenthesis should go to new line:
```suggestion
            new AtomicInteger()
        );
```
Please check all your changes in this file.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamThreadTest.java
```diff
@@ -932,9 +929,8 @@ public void shouldShutdownTaskManagerOnClose() {
             internalTopologyBuilder,
             CLIENT_ID,
             new LogContext(""),
-            new AtomicInteger(),
-            new AtomicLong(Long.MAX_VALUE)
-        ).updateThreadMetadata(getSharedAdminClientId(CLIENT_ID));
+            new AtomicLong(Long.MAX_VALUE),
+            new AtomicInteger()).updateThreadMetadata(getSharedAdminClientId(CLIENT_ID));
```
cadonna: Also here it should be:
```suggestion
            new AtomicInteger()
        ).updateThreadMetadata(getSharedAdminClientId(CLIENT_ID));
```

cadonna (commented):
@wcarlson5, Thank you for the PR!

I think, you could have opened two PRs instead of one. One for the introduction of the uncaught exception handler and another for the shutdown request. Smaller PRs are easier to review, IMO.

Here my other feedback.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I am not so sure. The null in the other handler is used to "reset" to the default behavior. I already set in the constructor

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I makes more sense to me to have all the cases in the same place. Since it is the default all the other cases will execute the rest of the logic as well so the rethrowing of the error not really in this case.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: That does make sense, I will change it

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
wcarlson5: Since the handler is now executing on the thread we don't need to pass that in

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
wcarlson5: okay

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
wcarlson5: okay

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: fixed

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: removed

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: this check is redundant. It should never be null here, I will remove it

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: removed

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
wcarlson5: i can

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
wcarlson5: no. Moved

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
wcarlson5: the UncaughtExceptionHandler which is set should not be called in this case. Do you mean we should verify it is not called?

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
wcarlson5: I am not sure what you mean by this can you clarify?

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamThreadTest.java
wcarlson5: turns out the auto refactoring does not work with new params at the end of methods. fixed.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamThreadTest.java
wcarlson5: foxed the 5 of these as well

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I am not sure, we never talked about the global stream thread. 

I don't think that it makes sense to use the same handler as its is quite different than a normal streamThread. I would lean towards making a separate method to just set it like it was

wcarlson5 (commented):
I think I got to all the comments. @cadonna @ableegoldman 

From what I can tell we have a couple of unanswered questions:
1) what state to leave the client in
2) what to do with the global thread

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
cadonna: I do not think we need to emulate the behavior of the java uncaught exception handler here. I do not see why a user should try to reset the uncaught exception handler. If we receive this requirement, we can still add it. I have the impression the "reset" makes this code unnecessarily complex.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
cadonna: Sorry, my bad! It doesn't matter whether it is called or not, since the `IllegalStateException` is thrown.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
cadonna: Actually, there are two tests missing. With the test you added, you test the `else` branch on line 400 in `KafkaStreams`.  But you do not test the `then` branch of the same `if` statement and within the `then` branch there are again a `then` branch and a `else` branch to test. 
If you do without the inner `else` branch as I proposed in another comment, you need to test the exception that is thrown when `null` is passed to the method.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
ableegoldman: I think users may find at least some of the same functionality to be useful, in particular the "shutdown client" and "shutdown application" enums. I also feel like users may want to be able to restart the global thread.  I think we can consider that out of scope for now, but I'd prefer to avoid introducing a new method that just accepts the old kind of uncaught exception handler if we're just going to deprecate that too.

But I do agree that it's pretty different, and we also don't want to commit to adding this functionality for the global thread right away so we should make sure it's clear what is and isn't implemented. So what do you think about mirroring the new handler for the StreamThread, eg adding a "GlobalUncaughtExceptionHandler", and just only including the `SHUTDOWN_THREAD` enum in the initial version? That way we're set up to easily extend the global handling feature without necessarily needing to implement it right away

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
ableegoldman: I agree with Bruno, I don't see why anyone would want to reset the exception handler (why set it in the first place then?). But if for some reason they really did set it at some point in their code and then later on want to revert to the default behavior, they can just pass in a handler that returns `SHUTDOWN_STREAMTHREAD_THREAD` themselves. If that's really what they want, then they should specify it.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: We ended up going with adding a new method to the handler

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Changed this to require not null

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -364,6 +370,98 @@ public void setUncaughtExceptionHandler(final Thread.UncaughtExceptionHandler eh
...
+                throw new IllegalStateException("Can only set UncaughtExceptionHandler in CREATED state. " +
+                        "Current state is: " + state);
+            }
+        }
+    }
+
+    private StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponseGlobalThread handleStreamsUncaughtExceptionGlobalThread(final Throwable e,
+                                                                                                                   final StreamsUncaughtExceptionHandler streamsUncaughtExceptionHandler) {
+        final StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponseGlobalThread action = streamsUncaughtExceptionHandler.handleExceptionInGlobalThread(e);
+        switch (action) {
```
wcarlson5: currently there is only one option but there may be more added with 406 so I will make as easy to extend as I can

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -931,6 +1033,62 @@ public synchronized boolean close(final long timeout, final TimeUnit timeUnit) {
         return close(timeoutMs);
     }
 
+    private void closeToError() {
```
wcarlson5, line 1003: This does that the same thing as close except state change. A user can still call close and it will change the error. There was talk about making Error terminal in that case this will not change.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -281,6 +285,14 @@ public boolean isRunning() {
     private final Admin adminClient;
     private final InternalTopologyBuilder builder;
 
+    private final AtomicInteger assignmentErrorCode;
+
+    private final ShutdownErrorHook shutdownErrorHook;
+
+    public interface ShutdownErrorHook {
```
wcarlson5, line 291: in order to let the rebalance listen shutdown with an error we just use this so we don't have to pass the KafkaStreams object and make a new public method

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
wcarlson5: That makes sense. Thank you for clarifying. I have added the null test

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamThreadTest.java
```diff
@@ -226,7 +226,9 @@ private StreamThread createStreamThread(@SuppressWarnings("SameParameterValue")
             0,
             stateDirectory,
             new MockStateRestoreListener(),
-            threadIdx
+            threadIdx,
+            new AtomicInteger(),
+            () -> { }
```
wcarlson5: This is an empty shutdown hook. I think it is better than null but I am not sure

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
```diff
@@ -0,0 +1,73 @@
...
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.errors;
+
+public interface StreamsUncaughtExceptionHandler {
+    /**
+     * Inspect the exception received in a stream thread and respond with an action.
+     * @param exception the actual exception
+     */
+    StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse handle(final Throwable exception);
```
wcarlson5: By making this not default we allow the handler to be over written by a lambda, this means we have to provide a default in the constructor but I think that is worth it.

If a user only wants to set the global handler they will have to set this as well but if they are doing that I assume they only have a global thread and in that case it does not matter what the handle function is as it will never be called.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -780,11 +874,13 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
                 delegatingStateRestoreListener,
-                i + 1);
+                i + 1,
+                new AtomicInteger(),
+                shutdownErrorHook
+            );
             threadState.put(threads[i].getId(), threads[i].state());
             storeProviders.add(new StreamThreadStateStoreProvider(threads[i], internalTopologyBuilder));
         }
-
```
wcarlson5, line 789: removed 3 empty lines to keep the constructor under the line limit

streams/src/test/java/org/apache/kafka/streams/integration/AppShutdownIntegrationTest.java
```diff
@@ -0,0 +1,279 @@
...
+
+            kafkaStreams.setUncaughtExceptionHandler(exception -> StreamsUncaughtExceptionHandler.StreamsUncaughtExceptionHandlerResponse.SHUTDOWN_STREAM_THREAD);
+
+            kafkaStreams.start();
+
+            produceMessages(0L, inputTopic, "A");
+            latch.await(15, TimeUnit.SECONDS);
+
+            assertThat(processorValueCollector.size(), equalTo(2));
+            assertThat(kafkaStreams.state(), equalTo(KafkaStreams.State.ERROR));
```
wcarlson5: This will be running when we make it so that no threads stays in running in KIP-663

wcarlson5 (commented):
Global handler now has its own call back
ShutdownApplication closes but leaves in state ERROR
ShutdownClient just calls close

responded to @abbccdda and @cadonna's comments as well

wcarlson5:
Moved to https://github.com/apache/kafka/pull/9487

PR #9274, opened 2020-09-09
KAFKA-10131: Remove use_zk_connection flag

rondagostino:
Eliminate "use_zk_connection" and similarly-named flags in public system test methods.  System tests should always use admin client when the broker supports it.  We do need to bootstrap ACLs prior to Kafka starting or prior to enabling the
authorizer, so the ability to force the use of a direct ZooKeeper connection rather than the admin client still exists for this case.

This PR also converts a few methods in kafka.py that were connecting directly to ZooKeeper for leader and replica/ISR lists
to use the admin client when the broker version supports it.  It also adds support for using the admin client when
performing "kafka-topics.sh ... -if-not-exists" which was added via KIP-604 in v2.6.

### DIFF
--- a/tests/kafkatest/services/kafka/kafka.py
+++ b/tests/kafkatest/services/kafka/kafka.py
@@ -254,7 +254,7 @@ class KafkaService(KafkaPathResolverMixin, JmxMixin, Service):
     def alive(self, node):
         return len(self.pids(node)) > 0
 
-    def start(self, add_principals="", use_zk_to_create_topic=True):
+    def start(self, add_principals=""):
         if self.zk_client_secure and not self.zk.zk_client_secure_port:
             raise Exception("Unable to start Kafka: TLS to Zookeeper requested but Zookeeper secure port not enabled")
         self.open_port(self.security_protocol)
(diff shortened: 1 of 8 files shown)


cmccabe:
Thanks, @rondagostino .  Is there any way to avoid moving the acls stuff out of `kafka_acls.py`?  It seems nice to have some separation rather than having everything in `kafka.py`

rondagostino:
@cmccabe Thanks for the review.  I put the logic back into `kafka_acls.py`.  Take a look and let me know if this is what you were thinking.

cmccabe:
Looks good.  LGTM pending a system test run

rondagostino:
@cmccabe There were several test failures at first, so I pushed commit https://github.com/apache/kafka/pull/9274/commits/fb417ade01613dad8fa2574c4ea3484f3fa4e65f to fix them and reran the full system test suite.  The results of this most recent run are at https://confluent-kafka-branch-builder-system-test-results.s3-us-west-2.amazonaws.com/2020-09-14--001.1600107441--rondagostino--KAFKA-10131--fb417ade0/report.txt:

```
====================================================================================================
SESSION REPORT (ALL TESTS)
ducktape version: 0.7.9
session_id:       2020-09-14--001
run time:         439 minutes 49.208 seconds
tests run:        684
passed:           513
failed:           1
ignored:          170
====================================================================================================
```

The one failure was due to an infrastructure issue:

```
test_id:    kafkatest.tests.connect.connect_distributed_test.ConnectDistributedTest.test_restart_failed_connector.connect_protocol=sessioned
status:     FAIL
run time:   8.720 seconds


    Unable to open channel.
```

cmccabe:
LGTM.  Thanks, @rondagostino

PR #9275, opened 2020-09-09
KAFKA-10435; Fetch protocol changes for KIP-595

hachikuji:
This patch bumps the `Fetch` protocol as specified by KIP-595: https://cwiki.apache.org/confluence/display/KAFKA/KIP-595%3A+A+Raft+Protocol+for+the+Metadata+Quorum. The main differences are the following:

- Truncation detection 
- Leader discovery through the response
- Flexible version support

The most notable change is truncation detection. This patch adds logic in the request handling path to detect truncation, but it does not change the replica fetchers to make use of this capability. We are planning to do this separately.


hachikuji:
Note this patch includes #9277. I will rebase once it is merged.

clients/src/main/java/org/apache/kafka/common/requests/FetchRequest.java
```diff
@@ -232,19 +263,25 @@ public FetchRequest build(short version) {
...
-                        .setFetchOffset(entry.getValue().fetchOffset)
-                        .setLogStartOffset(entry.getValue().logStartOffset)
-                        .setPartitionMaxBytes(entry.getValue().maxBytes));
+                FetchRequestData.FetchPartition fetchPartition = new FetchRequestData.FetchPartition()
+                    .setPartition(topicPartition.partition())
+                    .setCurrentLeaderEpoch(partitionData.currentLeaderEpoch.orElse(RecordBatch.NO_PARTITION_LEADER_EPOCH))
+                    .setFetchOffset(partitionData.fetchOffset)
+                    .setLogStartOffset(partitionData.logStartOffset)
+                    .setPartitionMaxBytes(partitionData.maxBytes);
+                partitionData.lastFetchedEpoch.ifPresent(fetchPartition::setLastFetchedEpoch);
```
rajinisivaram: `currentLeaderEpoch` and `lastFetchEpoch` are both set to -1 by default, but they are set in different ways above, is that deliberate?

core/src/test/scala/unit/kafka/server/FetchRequestTest.scala
```diff
@@ -212,6 +212,44 @@ class FetchRequestTest extends BaseRequestTest {
...
+
+    // Validate the expected truncation
+    val fetchResponse = sendFetchRequest(secondLeaderId, fetchRequest)
+    val partitionData = fetchResponse.responseData.get(topicPartition)
+    assertEquals(Errors.NONE, partitionData.error)
+    assertEquals(0L, partitionData.records.sizeInBytes())
+    assertTrue(partitionData.truncationOffset.isPresent)
+
+    // Should be exactly 100, but use a fuzzy truncation estimate in case there were produce retries
+    assertTrue(partitionData.truncationOffset.get < 150)
```
rajinisivaram: we could produce without retries, wait for completion and then check exact value?

core/src/main/scala/kafka/server/epoch/LeaderEpochFileCache.scala
```diff
@@ -91,8 +95,23 @@ class LeaderEpochFileCache(topicPartition: TopicPartition,
     }
   }
 
+  def removeEntries(predicate: EpochEntry => Boolean): Seq[EpochEntry] = {
+    val removedEpochs = mutable.ListBuffer.empty[EpochEntry]
+    val iterator = epochs.entrySet().iterator()
+
+    while (iterator.hasNext) {
```
rajinisivaram, line 128: I guess we would always be removing a range from the start or end. Are we going through the whole map and checking each one because there is no suitable API to remove multiple entries from start or end?

rajinisivaram (approved):
@hachikuji Thanks for the PR. Left a few minor comments/questions, apart from that LGTM.

rajinisivaram (approved):
@hachikuji Thanks for the PR. Left a few minor comments/questions, apart from that LGTM.

clients/src/main/java/org/apache/kafka/common/requests/FetchRequest.java
hachikuji: I guess I was trying to rely on the default from the schema, but I agree probably no reason to be inconsistent.

core/src/test/scala/unit/kafka/server/FetchRequestTest.scala
hachikuji: That's a good idea. Let me try that.

core/src/main/scala/kafka/server/epoch/LeaderEpochFileCache.scala
hachikuji: That's a good question. We probably could be smarter in `truncateAndAppend` if it was safe to assume epoch/start offset increases monotonically. Maybe we just need to add the logic to validate that invariant on startup. Let me take a look at that.

core/src/main/scala/kafka/server/epoch/LeaderEpochFileCache.scala
```diff
@@ -47,63 +47,81 @@ class LeaderEpochFileCache(topicPartition: TopicPartition,
...
+  private def assign(entry: EpochEntry): Boolean = {
+    if (entry.epoch < 0 || entry.startOffset < 0) {
+      throw new IllegalArgumentException(s"Received invalid partition leader epoch entry $entry")
+    }
+
+    // Check whether the append is needed before acquiring the write lock
+    // in order to avoid contention with readers in the common case
+    latestEntry.foreach { lastEntry =>
+      if (entry.epoch == lastEntry.epoch && entry.startOffset >= lastEntry.startOffset) {
+        return false
```
rajinisivaram: Do we need to check again inside the lock before update because this is outside the lock?

rajinisivaram (approved):
@hachikuji Thanks for the update, looks good. Left one question because I wasn't sure. Apart from that LGTM

core/src/main/scala/kafka/server/epoch/LeaderEpochFileCache.scala
hachikuji: Yeah, let me add it. I had debated it and convinced myself it was not needed, but that probably relies a bit too heavily on locking at higher layers. Probably better to err on the safe side.

rajinisivaram (approved):
@hachikuji Thanks for the update, LGTM

core/src/main/scala/kafka/api/ApiVersion.scala
```diff
@@ -100,7 +100,9 @@ object ApiVersion {
     // Introduced StopReplicaRequest V3 containing the leader epoch for each partition (KIP-570)
     KAFKA_2_6_IV0,
     // Introduced feature versioning support (KIP-584)
-    KAFKA_2_7_IV0
+    KAFKA_2_7_IV0,
+    // Bup Fetch protocol for Raft protocol (KIP-595)
```
ijuma, line 104: Maybe you can fix this typo in your next PR.

hachikuji: Oops! Will do.

PR #9276, opened 2020-09-09
KAFKA-10473: Add docs on partition size-on-disk, and other log-related metrics

wushujames:
Add docs on the following JMX metrics

kafka.log,type=Log,name=Size
kafka.log,type=Log,name=NumLogSegments
kafka.log,type=Log,name=LogStartOffset
kafka.log,type=Log,name=LogEndOffset

To test this, I opened the docs in a web browser


docs/ops.html
```diff
@@ -1129,6 +1128,26 @@ <h4><a id="remote_jmx" href="#remote_jmx">Security Considerations for Remote Mon
...
       </tr>
+      <tr>
+        <td>Size of a partition on disk (in bytes)</td>
+        <td>kafka.log:type=Log,name=Size,topic=([-.\w]+),partition=([0-9]+)</td>
+        <td>The size of a partition on disk, measured in bytes.</td>
+      </tr>
+      <tr>
+        <td>Number of log segments in a partition</td>
+        <td>kafka.log:type=Log,name=NumLogSegments,topic=([-.\w]+),partition=([0-9]+)</td>
+        <td>The number of log segments in a partition.</td>
```
wushujames, line 1050: I wasn't sure what information to add for this column of "normal value". 
* For date-based retention topics, it should roughly be sizeOfTopic/segmentSize
* But for compacted topics, that isn't the case
* If this is too high, it will mean more file handles.
* If it is too high, it might inefficient (is this true?)

Anyway, I wasn't sure what to put, so I left it simple.

docs/ops.html
```diff
@@ -1129,6 +1128,26 @@ <h4><a id="remote_jmx" href="#remote_jmx">Security Considerations for Remote Mon
...
+      </tr>
+      <tr>
+        <td>First offset in a partition</td>
+        <td>kafka.log:type=Log,name=LogStartOffset,topic=([-.\w]+),partition=([0-9]+)</td>
+        <td>The first offset in a partition.</td>
+      </tr>
+      <tr>
+        <td>Last offset in a partition</td>
+        <td>kafka.log:type=Log,name=LogEndOffset,topic=([-.\w]+),partition=([0-9]+)</td>
+        <td>The last offset in a partition.</td>
```
wushujames, line 1060: Is this the last committed offset? Or, is it uncommitted offsets? Meaning, if a producer appends a record there but the followers haven't yet acked it, will this value get incremented?

Does that level of detail matter, in the docs?

wushujames: Tag @hachikuji

wushujames:
@guozhangwang Can I get a review?

wushujames:
The pull request lists lots of failed checks. However, this pull request only changed an HTML file, and it seems unrelated to the test failures in those checks.

dongjinleekr:
@wushujames Could you make some updates? I think it would be good for this PR to be included in the 2.7.0 release.

wushujames:
@dongjinleekr : I've pulled in the latest changes from trunk. I'll check back in after the builds have completed.

Aside from pulling the latest changes from trunk, were there any other updates/changes you think are needed?

wushujames:
@dongjinleekr All builds failed, but none of the failures seem related to my change.

docs/ops.html
guozhangwang: This is the last uncomitted offset: after the local append it would be incremented, not waiting for other replicas to ack.

docs/ops.html
guozhangwang: I think this is fine :)

guozhangwang (approved):

PR #9091, opened 2020-07-28
MINOR; Make KafkaAdminClientTest.testDescribeLogDirsPartialFailure and KafkaAdminClientTest.testAlterReplicaLogDirsPartialFailure test more reliable

dajac:
I have seem them failing locally from times to times. It seems that the changes made in https://github.com/apache/kafka/pull/8864 have made them more fragile. I have updated them to be more reliable.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9277, opened 2020-09-09
MINOR: Fix JSON generation of nested structs with non-matching type/name

hachikuji:
The schema specification allows a struct type name to differ from the field name. This works with the generated `Message` classes, but not with the generated JSON converter. The problem seems to be that the type name gets replaced with the field name when the struct is registered.


cmccabe:
LGTM

hachikuji:
One flaky failure which is fixed by #9091. Will merge to trunk.

PR #9279, opened 2020-09-10
MINOR: Fix common struct `JsonConverter` and `Schema` generation

hachikuji:
This patch fixes a couple problems with the use of the `StructRegistry`. First, it fixes registration so that it is consistently based on the typename of the struct. Previously structs were registered under the field name which meant that fields which referred to common structs resulted in multiple entries. Second, the patch fixes `SchemaGenerator` so that common structs are considered first.

### DIFF
--- a/generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
+++ b/generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
@@ -53,32 +53,23 @@ public final class JsonConverterGenerator implements MessageClassGenerator {
         buffer.incrementIndent();
         generateConverters(message.dataClassName(), message.struct(),
             message.validVersions());
-
         for (Iterator<StructRegistry.StructInfo> iter = structRegistry.structs();
                 iter.hasNext(); ) {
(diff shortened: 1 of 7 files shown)


generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
```diff
@@ -53,17 +53,17 @@ public void generateAndWrite(MessageSpec message, BufferedWriter writer)
         buffer.incrementIndent();
         generateConverters(message.dataClassName(), message.struct(),
             message.validVersions());
+
+        for (Iterator<StructSpec> iter = structRegistry.commonStructs(); iter.hasNext(); ) {
+            generateStructClass(iter.next(), message.validVersions());
+        }
+
         for (Iterator<StructRegistry.StructInfo> iter = structRegistry.structs();
```
cmccabe: It's not clear whether we need to generate the common struct JSON converters first.  Can you try without (or explain why it needs to be with?)  Or is the iterator returned from `structs` not returning the correct data for common structs?

cmccabe:
LGTM

PR #9281, opened 2020-09-12
KAFKA-10478: Allow duplicated ports in advertised.listeners

asdaraujo:
Remove the requirement for unique port numbers for the advertised.listener parameters.
This restriction makes for the listeners parameter but there's not reason to apply the
same logic for advertised.listeners.

Being able to do this opens possibilities for some practical applications when using
Kerberos authentication. For example, when configuring Kafka using Kerberos authentication
and a Load Balancer we need to have two SASL_SSL listeners: (A) one running with the
kafka/hostname principal and (B) another using kafka/lb_name, which is necessary for
proper authentication when using the LB FQDN. After bootstrap, though, the client receives
the brokers' addresses with the actual host FQDNs advertised by the brokers. To connect
to the brokerd using the hostnames the client must connect to the listener A to be able to
authenticate successfully with Kerberos.

All unit/integration tests have passed.

- [X] Verify design and implementation 
- [X] Verify test coverage and CI build status
- [X] Verify documentation (including upgrade notes)

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaConfig.scala
+++ b/core/src/main/scala/kafka/server/KafkaConfig.scala
@@ -642,12 +642,13 @@ object KafkaConfig {
   "Use <code>listeners</code> instead. \n" +
   "hostname of broker. If this is set, it will only bind to this address. If this is not set, it will bind to all interfaces"
   val ListenersDoc = "Listener List - Comma-separated list of URIs we will listen on and the listener names." +
-  s" If the listener name is not a security protocol, $ListenerSecurityProtocolMapProp must also be set.\n" +
-  " Specify hostname as 0.0.0.0 to bind to all interfaces.\n" +
-  " Leave hostname empty to bind to default interface.\n" +
-  " Examples of legal listener lists:\n" +
-  " PLAINTEXT://myhost:9092,SSL://:9091\n" +
-  " CLIENT://0.0.0.0:9092,REPLICATION://localhost:9093\n"
+    s" If the listener name is not a security protocol, <code>$ListenerSecurityProtocolMapProp<code> must also be set.\n" +
+    " Listener names and port numbers must be unique.\n" +
+    " Specify hostname as 0.0.0.0 to bind to all interfaces.\n" +
+    " Leave hostname empty to bind to default interface.\n" +
+    " Examples of legal listener lists:\n" +
+    " PLAINTEXT://myhost:9092,SSL://:9091\n" +
+    " CLIENT://0.0.0.0:9092,REPLICATION://localhost:9093\n"
   val AdvertisedHostNameDoc = "DEPRECATED: only used when <code>advertised.listeners</code> or <code>listeners</code> are not set. " +
   "Use <code>advertised.listeners</code> instead. \n" +
   "Hostname to publish to ZooKeeper for clients to use. In IaaS environments, this may " +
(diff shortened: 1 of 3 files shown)


core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1686,9 +1686,9 @@ class KafkaConfig(val props: java.util.Map[_, _], doLog: Boolean, dynamicConfigO
   def advertisedListeners: Seq[EndPoint] = {
     val advertisedListenersProp = getString(KafkaConfig.AdvertisedListenersProp)
     if (advertisedListenersProp != null)
-      CoreUtils.listenerListToEndPoints(advertisedListenersProp, listenerSecurityProtocolMap)
+      CoreUtils.listenerListToEndPoints(advertisedListenersProp, listenerSecurityProtocolMap, false)
```
viktorsomogyi: I would suggest to use named parameter or an enum for the booleans to improve readability.

viktorsomogyi (commented):
Please also update the documentation, I think it's worth to note that in case of `listeners` it's not allowed but for `advertised.listeners` it is for the LB use-case.
Otherwise it looks good.

core/src/main/scala/kafka/server/KafkaConfig.scala
asdaraujo: Updated

asdaraujo:
> Please also update the documentation, I think it's worth to note that in case of `listeners` it's not allowed but for `advertised.listeners` it is for the LB use-case.
> Otherwise it looks good.

Thanks, @viktorsomogyi .  I've updated the documentation as well.

viktorsomogyi (approved):
@asdaraujo thanks for the PR, it looks good to me.
@mimaison do you know who is the right person to review/merge this or could you review it as well please?

tombentley (approved):
LGTM.

core/src/test/scala/unit/kafka/server/KafkaConfigTest.scala
```diff
@@ -218,16 +220,26 @@ class KafkaConfigTest {
     props.put(KafkaConfig.ZkConnectProp, "localhost:2181")
 
     // listeners with duplicate port
-    props.put(KafkaConfig.ListenersProp, "PLAINTEXT://localhost:9091,TRACE://localhost:9091")
-    assertFalse(isValidKafkaConfig(props))
+    props.put(KafkaConfig.ListenersProp, "PLAINTEXT://localhost:9091,SSL://localhost:9091")
+    var caught = intercept[IllegalArgumentException] { KafkaConfig.fromProps(props) }
+    assertThat(caught.getMessage(), containsString("Each listener must have a different port"))
```
mimaison: Can we use something like:
```
assertTrue(caught.getMessage.contains("Each listener must have a different port"))
```
That way we don't need to import hamcrest and only use junit.

This can be applied to all `assertThat()` added in this PR

mimaison (changes requested):
Thanks for the PR. The change overall looks good.
I've suggested a small cleanup

core/src/test/scala/unit/kafka/server/KafkaConfigTest.scala
asdaraujo: Nice. I'll change that.

mimaison (approved):
LGTM

mimaison:
Test failures are not related, merging

PR #9283, opened 2020-09-14
KAFKA-9697 hide the metrics of control.plane.listener.name if it is u…

chia7712:
Both of ```ControlPlaneNetworkProcessorAvgIdlePercent``` and ```ControlPlaneExpiredConnectionsKilledCount```are exposed with meaningless value when ```control.plane.listener.name``` is unset. They should be hidden as other metrics when ```control.plane.listener.name``` is unset.


dongjinleekr (commented):
Do we need to update the `docs/ops.html`?

For example of `NetworkProcessorAvgIdlePercent`,

From:
`between 0 and 1, ideally &gt 0.3`

To:
`between 0 and 1, ideally &gt 0.3. Available iff control.plane.listener.name is set.`

chia7712:
As it may break metrics compatibility, I'm going to close it.

PR #9284, opened 2020-09-14
KAFKA-10479 Throw exception if users try to update configs of existen…

chia7712:
```scala
def immutableListenerConfigs(kafkaConfig: KafkaConfig, prefix: String): Map[String, AnyRef] = {
      newConfig.originals.asScala.filter { case (key, _) =>
        key.startsWith(prefix) && !DynamicSecurityConfigs.contains(key)
      }
    }
```

We don't actually compare new configs to origin configs so the suitable exception is not thrown.

### DIFF
--- a/core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
+++ b/core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
@@ -905,14 +905,12 @@ class DynamicListenerConfig(server: KafkaServer) extends BrokerReconfigurable wi
       throw new ConfigException(s"Listeners '$newListeners' must be subset of listener map '${newConfig.listenerSecurityProtocolMap}'")
     newListeners.keySet.intersect(oldListeners.keySet).foreach { listenerName =>
       def immutableListenerConfigs(kafkaConfig: KafkaConfig, prefix: String): Map[String, AnyRef] = {
-        kafkaConfig.originals.asScala.filter { case (key, _) =>
-          key.startsWith(prefix) && !DynamicSecurityConfigs.contains(key)
+        kafkaConfig.originalsWithPrefix(prefix, true).asScala.filter { case (key, _) =>
+          // skip the reconfigurable configs
+          !DynamicSecurityConfigs.contains(key) && !SocketServer.ListenerReconfigurableConfigs.contains(key)
         }
       }
-      val prefix = listenerName.configPrefix
-      val newListenerProps = immutableListenerConfigs(newConfig, prefix)
-      val oldListenerProps = immutableListenerConfigs(oldConfig, prefix)
-      if (newListenerProps != oldListenerProps)
+      if (immutableListenerConfigs(newConfig, listenerName.configPrefix) != immutableListenerConfigs(oldConfig, listenerName.configPrefix))
         throw new ConfigException(s"Configs cannot be updated dynamically for existing listener $listenerName, " +
           "restart broker or create a new listener for update")
       if (oldConfig.listenerSecurityProtocolMap(listenerName) != newConfig.listenerSecurityProtocolMap(listenerName))
--- a/core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
+++ b/core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
@@ -895,13 +895,6 @@ class DynamicListenerConfig(server: KafkaServer) extends BrokerReconfigurable wi
   }
 
   def validateReconfiguration(newConfig: KafkaConfig): Unit = {
-
-    def immutableListenerConfigs(kafkaConfig: KafkaConfig, prefix: String): Map[String, AnyRef] = {
-      newConfig.originals.asScala.filter { case (key, _) =>
-        key.startsWith(prefix) && !DynamicSecurityConfigs.contains(key)
-      }
-    }
-
     val oldConfig = server.config
     val newListeners = listenersToMap(newConfig.listeners)
     val newAdvertisedListeners = listenersToMap(newConfig.advertisedListeners)
@@ -911,6 +904,11 @@ class DynamicListenerConfig(server: KafkaServer) extends BrokerReconfigurable wi
     if (!newListeners.keySet.subsetOf(newConfig.listenerSecurityProtocolMap.keySet))
       throw new ConfigException(s"Listeners '$newListeners' must be subset of listener map '${newConfig.listenerSecurityProtocolMap}'")
     newListeners.keySet.intersect(oldListeners.keySet).foreach { listenerName =>
+      def immutableListenerConfigs(kafkaConfig: KafkaConfig, prefix: String): Map[String, AnyRef] = {
+        kafkaConfig.originals.asScala.filter { case (key, _) =>
+          key.startsWith(prefix) && !DynamicSecurityConfigs.contains(key)
+        }
+      }
       val prefix = listenerName.configPrefix
       val newListenerProps = immutableListenerConfigs(newConfig, prefix)
       val oldListenerProps = immutableListenerConfigs(oldConfig, prefix)
--- a/core/src/test/scala/unit/kafka/server/DynamicBrokerConfigTest.scala
+++ b/core/src/test/scala/unit/kafka/server/DynamicBrokerConfigTest.scala
@@ -327,10 +327,12 @@ class DynamicBrokerConfigTest {
     EasyMock.replay(kafkaServer)
 
     props.put(KafkaConfig.ListenersProp, "PLAINTEXT://hostname:9092,SASL_PLAINTEXT://hostname:9093")
-    val newConfig = KafkaConfig(props)
+    new DynamicListenerConfig(kafkaServer).validateReconfiguration(KafkaConfig(props))
 
+    // it is illegal to update configs of existent listeners
+    props.put("listener.name.plaintext.you.should.not.pass", "failure")
     val dynamicListenerConfig = new DynamicListenerConfig(kafkaServer)
-    dynamicListenerConfig.validateReconfiguration(newConfig)
+    assertThrows(classOf[ConfigException], () => dynamicListenerConfig.validateReconfiguration(KafkaConfig(props)))
   }
 
   @Test
--- a/core/src/test/scala/unit/kafka/server/DynamicBrokerConfigTest.scala
+++ b/core/src/test/scala/unit/kafka/server/DynamicBrokerConfigTest.scala
@@ -329,7 +329,7 @@ class DynamicBrokerConfigTest {
     props.put(KafkaConfig.ListenersProp, "PLAINTEXT://hostname:9092,SASL_PLAINTEXT://hostname:9093")
     new DynamicListenerConfig(kafkaServer).validateReconfiguration(KafkaConfig(props))
 
-    // it is illegal to update configs of existent listeners
+    // it is illegal to update non-reconfiguable configs of existent listeners
     props.put("listener.name.plaintext.you.should.not.pass", "failure")
     val dynamicListenerConfig = new DynamicListenerConfig(kafkaServer)
     assertThrows(classOf[ConfigException], () => dynamicListenerConfig.validateReconfiguration(KafkaConfig(props)))
--- a/docs/upgrade.html
+++ b/docs/upgrade.html
@@ -28,11 +28,10 @@
         Note that parameter <code>retry.backoff.ms</code> is not impacted by this change.
     </li>
     <li>Altering non-reconfigurable configs of existent listeners causes <code>InvalidRequestException</code>.
-        By contrast, the previous behavior would have caused the updated configuration to be persisted, but it wouldn't
-        take effect until the broker was restarted. The old behavior is not worth keeping since that behavior is unintended.
-        see <a href="https://github.com/apache/kafka/pull/9284">KAFKA-10479</a> for more discussion.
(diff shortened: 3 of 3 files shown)


chia7712:
the failure tests are unrelated to this patch. rebase :)

@hachikuji @rajinisivaram Could you take a look?

This issue was introduced by https://github.com/apache/kafka/commit/4019b21d602c0912d9dc0286f10701e7e724f750

core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
```diff
@@ -910,6 +903,11 @@ class DynamicListenerConfig(server: KafkaServer) extends BrokerReconfigurable wi
     if (!newListeners.keySet.subsetOf(newConfig.listenerSecurityProtocolMap.keySet))
       throw new ConfigException(s"Listeners '$newListeners' must be subset of listener map '${newConfig.listenerSecurityProtocolMap}'")
     newListeners.keySet.intersect(oldListeners.keySet).foreach { listenerName =>
+      def immutableListenerConfigs(kafkaConfig: KafkaConfig, prefix: String): Map[String, AnyRef] = {
+        kafkaConfig.originals.asScala.filter { case (key, _) =>
+          key.startsWith(prefix) && !DynamicSecurityConfigs.contains(key)
+        }
+      }
       val prefix = listenerName.configPrefix
       val newListenerProps = immutableListenerConfigs(newConfig, prefix)
```
dongjinleekr: Right, this line does not work correctly as of present.

dongjinleekr (approved):

hachikuji (approved):
LGTM. Good catch!

hachikuji:
@chia7712 Can you take a look at DynamicConnectionsQuotaTest? Looks like this patch is causing it to fail.

chia7712:
> Can you take a look at DynamicConnectionsQuotaTest? Looks like this patch is causing it to fail.

The error message should be unrelate to this patch. Also, I loop it 5 times on my local. all pass.
```
java.lang.AssertionError: Found unexpected threads during @AfterClass, allThreads=HashSet(/127.0.0.1:52490 to /127.0.0.1:43059 workers Thread 2, Common-Cleaner, metrics-meter-tick-thread-2, main, metrics-meter-tick-thread-1, Signal Dispatcher, scala-execution-context-global-139, /127.0.0.1:52490 to /127.0.0.1:43059 workers Thread 3, Reference Handler, kafka-admin-client-thread | adminclient-4, scala-execution-context-global-373, scala-execution-context-global-140, Test worker, Finalizer), unexpected=HashSet(kafka-admin-client-thread | adminclient-4)
```

I will rebase this patch to trigger QA again.

hachikuji:
I see the failure consistently running this branch locally. Here is the error I get from `testDynamicListenerConnectionQuota`:

```
java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.InvalidRequestException: Invalid config value for resource ConfigResource(type=BROKER, name='0'): Invalid value org.apache.kafka.common.config.ConfigException: Configs cannot be updated dynamically for existing listener ListenerName(PLAINTEXT), restart broker or create a new listener for update for configuration Invalid dynamic configuration

	at org.apache.kafka.common.internals.KafkaFutureImpl.wrapAndThrow(KafkaFutureImpl.java:45)
	at org.apache.kafka.common.internals.KafkaFutureImpl.access$000(KafkaFutureImpl.java:32)
	at org.apache.kafka.common.internals.KafkaFutureImpl$SingleWaiter.await(KafkaFutureImpl.java:89)
	at org.apache.kafka.common.internals.KafkaFutureImpl.get(KafkaFutureImpl.java:260)
	at kafka.network.DynamicConnectionQuotaTest.reconfigureServers(DynamicConnectionQuotaTest.scala:225)
	at kafka.network.DynamicConnectionQuotaTest.testDynamicListenerConnectionQuota(DynamicConnectionQuotaTest.scala:126)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:103)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:63)
	at org.junit.runners.ParentRunner$4.run(ParentRunner.java:331)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:79)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:329)
	at org.junit.runners.ParentRunner.access$100(ParentRunner.java:66)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:293)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:413)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:230)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:58)
Caused by: org.apache.kafka.common.errors.InvalidRequestException: Invalid config value for resource ConfigResource(type=BROKER, name='0'): Invalid value org.apache.kafka.common.config.ConfigException: Configs cannot be updated dynamically for existing listener ListenerName(PLAINTEXT), restart broker or create a new listener for update for configuration Invalid dynamic configuration

```

It looks to me like the test case is attempting to update a config which is now triggering the fixed validation.

chia7712:
> I see the failure consistently running this branch locally

my bad. I looped tests with incorrect branch :(

I have fixed the failed test. @dongjinleekr @hachikuji please take a look

chia7712:
```
Build / JDK 11 / kafka.api.ConsumerBounceTest.testRollingBrokerRestartsWithSmallerMaxGroupSizeConfigDisruptsBigGroup
```

only jdk 11 get failed test. It is unrelated to this patch.

@hachikuji It should be ready to be merged. BTW, this patch does break the behavior of altering non-reconfigurable configs of existent listeners as we didn't throw exception before. Not sure whether we should do more for this change. Maybe It would be better to add extra comment to git comment when merging this patch. For example:

`
this patch introduces an incompatible behavior change. Altering non-reconfigurable configs of existent listeners causes exception instead of being a NOP 
`

hachikuji (commented):
@chia7712 Good question on compatibility impact. So the previous behavior would have caused the updated configuration to be persisted, but it wouldn't take effect until the broker was restarted. Is that right?

chia7712:
> So the previous behavior would have caused the updated configuration to be persisted, but it wouldn't take effect until the broker was restarted. Is that right?

You do give a better explanation :)

hachikuji:
@chia7712 Mentioning the fix in the upgrade notes seems the best we can do. The old (unintended) behavior does not seem worth keeping for the sake of compatibility. @rajinisivaram What do you think?

rajinisivaram (approved):
@chia7712 Thanks for the PR, good catch! @hachikuji Yes, agree that we should merge this and mention in upgrade notes.

chia7712:
```
Build / JDK 11 / kafka.api.ConsumerBounceTest.testRollingBrokerRestartsWithSmallerMaxGroupSizeConfigDisruptsBigGroup
```
it is flaky on trunk branch also so it is unrelated to this patch

hachikuji:
@chia7712 To clarify, were you going to mention this in `docs/upgrade.html`?

chia7712:
> To clarify, were you going to mention this in docs/upgrade.html?

will copy that

chia7712:
@hachikuji the ```docs/upgrade.html``` is updated :)

docs/upgrade.html
```diff
@@ -27,6 +27,14 @@ <h5><a id="upgrade_270_notable" href="#upgrade_270_notable">Notable changes in 2
...
         Note that parameter <code>retry.backoff.ms</code> is not impacted by this change.
     </li>
+    <li>Altering non-reconfigurable configs of existent listeners causes <code>InvalidRequestException</code>.
+        By contrast, the previous behavior would have caused the updated configuration to be persisted, but it wouldn't
+        take effect until the broker was restarted. This change breaks behavior compatibility but the old behavior is not
+        worth keeping since that behavior is unintended. see
+        <a href="https://github.com/apache/kafka/pull/9284">KAFKA-10479</a> for more discussion.
+        see <code>DynamicBrokerConfig.DynamicSecurityConfigs</code> and <code>SocketServer.ListenerReconfigurableConfigs</code>
+        for reconfigurable configs of existent listeners.
+    </li>
```
ijuma, line 35: Does it really break compatibility? It seems like the previous behavior wasn't working as intended anyway.

core/src/test/scala/unit/kafka/server/DynamicBrokerConfigTest.scala
```diff
@@ -327,10 +327,12 @@ class DynamicBrokerConfigTest {
     EasyMock.replay(kafkaServer)
 
     props.put(KafkaConfig.ListenersProp, "PLAINTEXT://hostname:9092,SASL_PLAINTEXT://hostname:9093")
-    val newConfig = KafkaConfig(props)
+    new DynamicListenerConfig(kafkaServer).validateReconfiguration(KafkaConfig(props))
 
+    // it is illegal to update configs of existent listeners
+    props.put("listener.name.plaintext.you.should.not.pass", "failure")
```
ijuma, line 333: It is possible to update configs of existing listeners if they are dynamic, right?

docs/upgrade.html
chia7712: > Does it really break compatibility? It seems like the previous behavior wasn't working as intended anyway.

It seems to me the APIs behavior is changed so the compatibility is broken.

For example, the code which is used to change non-reconfigurable configs gets exception now so users have to write something to handle "new" exception. Of course, that case should be very rare so mentioning this in docs/upgrade.html is enough to this incompatible change.

core/src/test/scala/unit/kafka/server/DynamicBrokerConfigTest.scala
chia7712: you are right. I will revise the comment.

docs/upgrade.html
ijuma: I know what you're saying, but this change aligns the implementation with the specified behavior. The previous behavior was a bug as it failed to report an error even though it did not change the config dynamically.

chia7712: > The previous behavior was a bug as it failed to report an error even though it did not change the config dynamically.

ok. it makes sense. I will remove this comment from ```docs/upgrade.html```

ijuma: I think we should keep the comment, I was suggesting we tweak the description.

chia7712: > I think we should keep the comment, I was suggesting we tweak the description.

got it.

docs/upgrade.html
```diff
@@ -27,6 +27,14 @@ <h5><a id="upgrade_270_notable" href="#upgrade_270_notable">Notable changes in 2
         <code>default.api.timeout.ms</code>, and Kafka Streams' new <code>task.timeout.ms</code> parameters instead.
         Note that parameter <code>retry.backoff.ms</code> is not impacted by this change.
     </li>
+    <li>Altering non-reconfigurable configs of existent listeners causes <code>InvalidRequestException</code>.
+        By contrast, the previous behavior would have caused the updated configuration to be persisted, but it wouldn't
+        take effect until the broker was restarted. This change breaks behavior compatibility but the old behavior is not
```
hachikuji: nit: can we take out the sentence about behavior compatibility. I think it is enough to say that the previous behavior was unintended.

chia7712: I had deleted the sentence about behavior compatibility :)

hachikuji (approved):

chia7712:
@hachikuji Thanks for reviews and update!

PR #9285, opened 2020-09-14
Minor: fix streams quickstart in Jenkinsfile

mumrah:
Running commands like "cd" or setting variables inside `sh` steps do not survive to following steps in Jenkinsfile. Instead we're supposed to use things like `dir("somedir") { ... }` and injecting variables with `env` directive or directly into the command string (like I've done here).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -50,10 +50,7 @@ def doStreamsArchetype() {
          || { echo 'Could not install kafka-streams.jar (and dependencies) locally`'; exit 1; }
   '''
 
-  sh '''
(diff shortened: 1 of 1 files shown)


vvcephei (approved):
Thanks, @mumrah !

cadonna (approved):
Thanks a lot @mumrah !

mumrah:
Here's the evidence that it's working now https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-9285/1/execution/node/134/log/ 

image

PR #9290, opened 2020-09-16
KAFKA-10487; Fetch response should return diverging epoch and end offset

hachikuji:
This patch changes the Fetch response schema to include both the diverging epoch and its end offset rather than just the offset. This allows for more accurate truncation on the follower. This is the schema that was originally specified in KIP-595, but we altered it during the discussion.


hachikuji:
An alternative name for "diverging epoch" is "previous epoch" which is used in the literature. I preferred the former because it is more descriptive and because the returned epoch could be equal to the last_fetched_epoch in the request.

rajinisivaram (approved):
@hachikuji Thanks for the PR, LGTM

PR #9291, opened 2020-09-16
KAFKA-10264 Flaky Test TransactionsTest.testBumpTransactionalEpoch

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-10264

The test case sends two records before killing broker. The failure is caused by that the both records are NOT in a single batch. The failure of first record can abort second batch and then produces KafkaException rather than TimeoutException.


hachikuji (approved):
LGTM

PR #9292, opened 2020-09-16
KAFKA-8238: adding number of messages/bytes read

vamossagar12:
comment to ping reviewers.*

### DIFF
--- a/core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala
+++ b/core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala
@@ -572,6 +572,9 @@ class GroupMetadataManager(brokerId: Int,
         val pendingOffsets = mutable.Map[Long, mutable.Map[GroupTopicPartition, CommitRecordMetadataAndOffset]]()
         val loadedGroups = mutable.Map[String, GroupMetadata]()
         val removedGroups = mutable.Set[String]()
+        var batchesRead = 0
+        var messagesRead = 0
+        var bytesRead = 0
 
         // buffer may not be needed if records are read from memory
         var buffer = ByteBuffer.allocate(0)
@@ -627,6 +630,8 @@ class GroupMetadataManager(brokerId: Int,
                     }
                 }
                 pendingOffsets.remove(batch.producerId)
+                messagesRead += 1
+                bytesRead += record.sizeInBytes()
               }
             } else {
               var batchBaseOffset: Option[Long] = None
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala
```diff
@@ -670,12 +675,16 @@ class GroupMetadataManager(brokerId: Int,
...
+                messagesRead += 1
+                bytesRead += record.sizeInBytes()
               }
             }
+            batchesRead += 1
             currOffset = batch.nextOffset
           }
         }
 
+        info(s"Number of messages/bytes/batches read: $messagesRead/$bytesRead/$batchesRead for topic $topicPartition")
```
chia7712: how about replacing "for topic $topicPartition" by "for partition $topicPartition"

vamossagar12: done..

chia7712 (approved):
@vamossagar12 Thanks for contribution. LGTM. nit: Could you add Jira number to title?

@stanislavkozlovski @ijuma @cmccabe Could you take a final look?

stanislavkozlovski (approved):
LGTM!

core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala
```diff
@@ -670,12 +675,16 @@ class GroupMetadataManager(brokerId: Int,
...
+                messagesRead += 1
+                bytesRead += record.sizeInBytes()
               }
             }
+            batchesRead += 1
             currOffset = batch.nextOffset
           }
         }
 
+        info(s"Number of messages/bytes/batches read: $messagesRead/$bytesRead/$batchesRead for partition $topicPartition")
```
ijuma, line 687: Will this be too noisy if there are lots of partitions? Do we want to have the log output be for all partitions instead?

vamossagar12: So, you are saying we add up all items like messages/bytes/batches read across all partitions and print them for the total number of partitions? 
And what you are saying is true, if we have a lot of partitions, then logs can be flooded with just these lines... 
The only thought there is, sometimes it might be useful debugging information.. So, instead of info, if we print it at a debug level?

dajac: We already log a message when the loading for a partition is completed: https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala#L552. We could perhaps add more information to that one instead of adding a new one.

vamossagar12: ok.. @dajac @ijuma , let me know what you guys think about this PR..

vamossagar12:
@dajac , @stanislavkozlovski, @ijuma  I was closing out on some old PRs of mine and found this one. Do you think this is still useful or should I just close it?

PR #9293, opened 2020-09-16
POC: Allow bypassing the cache layer in IQ

vvcephei:
Just a quick POC of how we can optionally bypass the caching layer in IQ. Some queries that are common in IQ but rare in the PAPI perform really poorly with the caching layer involved. We should have a comprehensive solution to this in the future, but for now it seems worth considering just giving a way to bypass it.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
@@ -46,41 +46,88 @@ import static org.apache.kafka.streams.processor.internals.metrics.StreamsMetric
  * @param <K>
  * @param <V>
  */
-public class MeteredKeyValueStore<K, V>
-    extends WrappedStateStore<KeyValueStore<Bytes, byte[]>, K, V>
+public class MeteredKeyValueStore<K, V, OwnType extends MeteredKeyValueStore<K, V, OwnType>>
+    extends MeteredStore<K, V, MeteredKeyValueStore<K, V, OwnType>, KeyValueStore<Bytes, byte[]>>
     implements KeyValueStore<K, V> {
 
     final Serde<K> keySerde;
     final Serde<V> valueSerde;
     StateSerdes<K, V> serdes;
 
-    private final String metricsScope;
+    protected final String metricsScope;
     protected final Time time;
     protected Sensor putSensor;
-    private Sensor putIfAbsentSensor;
+    protected Sensor putIfAbsentSensor;
     protected Sensor getSensor;
-    private Sensor deleteSensor;
-    private Sensor putAllSensor;
-    private Sensor allSensor;
-    private Sensor rangeSensor;
-    private Sensor flushSensor;
-    private Sensor e2eLatencySensor;
-    private ProcessorContext context;
-    private StreamsMetricsImpl streamsMetrics;
-    private final String threadId;
-    private String taskId;
+    protected Sensor deleteSensor;
+    protected Sensor putAllSensor;
+    protected Sensor allSensor;
+    protected Sensor rangeSensor;
+    protected Sensor flushSensor;
+    protected Sensor e2eLatencySensor;
+    protected ProcessorContext context;
+    protected StreamsMetricsImpl streamsMetrics;
+    protected final String threadId;
+    protected String taskId;
 
     MeteredKeyValueStore(final KeyValueStore<Bytes, byte[]> inner,
                          final String metricsScope,
                          final Time time,
                          final Serde<K> keySerde,
                          final Serde<V> valueSerde) {
+        this(
+            inner,
+            metricsScope,
+            Thread.currentThread().getName(),
+            time != null ? time : Time.SYSTEM,
+            keySerde,
+            valueSerde
+        );
+    }
+
+    protected MeteredKeyValueStore(final KeyValueStore<Bytes, byte[]> inner,
+                         final String metricsScope,
+                         final String threadId,
+                         final Time time,
+                         final Serde<K> keySerde,
+                         final Serde<V> valueSerde) {
         super(inner);
         this.metricsScope = metricsScope;
-        threadId = Thread.currentThread().getName();
-        this.time = time != null ? time : Time.SYSTEM;
+        this.threadId = threadId;
+        this.time = time;
         this.keySerde = keySerde;
         this.valueSerde = valueSerde;
+    }
+
+    public MeteredKeyValueStore<K, V, OwnType> reWrap(final KeyValueStore<Bytes, byte[]> inner) {
+        final MeteredKeyValueStore<K, V, OwnType> reWrapped = new MeteredKeyValueStore<>(
+            inner,
+            metricsScope,
+            threadId,
+            time,
+            keySerde,
+            valueSerde
+        );
+
+        copyInit(reWrapped);
+
+        return reWrapped;
+    }
+
+    protected void copyInit(final MeteredKeyValueStore<K, V, OwnType> reWrapped) {
+        reWrapped.context = context;
+        reWrapped.taskId = taskId;
+        reWrapped.serdes = serdes;
+        reWrapped.streamsMetrics = streamsMetrics;
+        reWrapped.putSensor = putSensor;
+        reWrapped.putIfAbsentSensor = putIfAbsentSensor;
+        reWrapped.putAllSensor = putAllSensor;
+        reWrapped.getSensor = getSensor;
+        reWrapped.allSensor = allSensor;
+        reWrapped.rangeSensor = rangeSensor;
+        reWrapped.flushSensor = flushSensor;
+        reWrapped.deleteSensor = deleteSensor;
+        reWrapped.e2eLatencySensor = e2eLatencySensor;
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -46,43 +46,90 @@
  * @param <K>
  * @param <V>
  */
-public class MeteredKeyValueStore<K, V>
-    extends WrappedStateStore<KeyValueStore<Bytes, byte[]>, K, V>
+public class MeteredKeyValueStore<K, V, OwnType extends MeteredKeyValueStore<K, V, OwnType>>
```
vvcephei, line 49: The self-type reference is pretty mind-bending, but basically what it's for is to let `reWrap` return the same type you call it on. So, `MeteredKeyValueStore#reWrap` returns a `MeteredKeyValueStore` with the same generic types, and `MeteredTimestampedKeyValueStore#reWrap` returns a `MeteredTimestampedKeyValueStore` with the same generic parameters.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -46,43 +46,90 @@
...
+    extends MeteredStore<K, V, MeteredKeyValueStore<K, V, OwnType>, KeyValueStore<Bytes, byte[]>>
     implements KeyValueStore<K, V> {
 
     final Serde<K> keySerde;
     final Serde<V> valueSerde;
     StateSerdes<K, V> serdes;
 
-    private final String metricsScope;
+    protected final String metricsScope;
```
vvcephei, line 57: need to be protected so that we can set them in `copyInit`

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -46,43 +46,90 @@
...
         this.metricsScope = metricsScope;
-        threadId = Thread.currentThread().getName();
-        this.time = time != null ? time : Time.SYSTEM;
+        this.threadId = threadId;
+        this.time = time;
         this.keySerde = keySerde;
         this.valueSerde = valueSerde;
     }
 
+    public MeteredKeyValueStore<K, V, OwnType> reWrap(final KeyValueStore<Bytes, byte[]> inner) {
```
vvcephei, line 102: This is the basic way that the cache bypass is implemented. You descend the stack of wrapped stores until you find the cache, then you pop it off, then you re-wrap the remaining store layers.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -46,43 +46,90 @@
...
+            keySerde,
+            valueSerde
+        );
+
+        copyInit(reWrapped);
+
+        return reWrapped;
+    }
+
+    protected void copyInit(final MeteredKeyValueStore<K, V, OwnType> reWrapped) {
```
vvcephei, line 117: We're basically cloning an already initialized Metered store, so we need to copy over everything that gets set in init.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredStore.java
```diff
@@ -0,0 +1,40 @@
...
+ * @param <V>
+ */
+public abstract class MeteredStore<K, V, OwnType extends MeteredStore<K, V, OwnType, InnerType>, InnerType extends StateStore>
+    extends WrappedStateStore<InnerType, K, V> {
+
+    public MeteredStore(final InnerType wrapped) {
+        super(wrapped);
+    }
+
+    abstract OwnType reWrap(final InnerType inner);
```
vvcephei, line 39: In retrospect, this actually belongs in `WrappedStateStore`. I'd fix it if we choose to go this way.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredTimestampedKeyValueStore.java
```diff
@@ -50,6 +50,31 @@
...
+                                            final String metricsScope,
+                                            final String threadId,
+                                            final Time time,
+                                            final Serde<K> keySerde,
+                                            final Serde<ValueAndTimestamp<V>> valueSerde) {
+        super(inner, metricsScope, threadId, time, keySerde, valueSerde);
+    }
+
+    @Override
+    public MeteredTimestampedKeyValueStore<K, V> reWrap(final KeyValueStore<Bytes, byte[]> inner) {
```
vvcephei, line 63: Overriding this here lets us preserve the correct type of the store when re-wrapping.

streams/src/main/java/org/apache/kafka/streams/state/internals/StreamThreadStateStoreProvider.java
```diff
@@ -84,26 +96,53 @@ public StreamThreadStateStoreProvider(final StreamThread streamThread,
...
-                return (T) store;
+                return (T) maybeUncached;
             }
         } else {
             return null;
         }
     }
 
+    @SuppressWarnings({"rawtypes", "unchecked"})
+    private static StateStore maybeBypassCache(final StateStore store, final boolean bypassCache) {
```
vvcephei, line 125: Here's where we actually bypass the cache. I've just realized that `reWrap` should be in `WrappedStateStore`, in which case, this logic should change a little, but the basic idea is there.

streams/src/main/java/org/apache/kafka/streams/state/internals/StreamThreadStateStoreProvider.java
```diff
@@ -84,26 +96,53 @@ public StreamThreadStateStoreProvider(final StreamThread streamThread,
...
+            final MeteredStore meteredStore = (MeteredStore) store;
+            final StateStore wrapped = meteredStore.wrapped();
+            final StateStore uncached = maybeBypassCache(wrapped, bypassCache);
+            if (uncached == wrapped) {
+                return store;
+            } else {
+                return meteredStore.reWrap(uncached);
+            }
+        } else if (WrappedStateStore.isCachingLayer(store)) {
+            return ((WrappedStateStore) store).wrapped();
```
vvcephei, line 138: Specifically, this is where we drop the caching layer from the returned store hierarchy.

streams/src/main/java/org/apache/kafka/streams/state/internals/StreamThreadStateStoreProvider.java
```diff
@@ -84,26 +96,53 @@ public StreamThreadStateStoreProvider(final StreamThread streamThread,
...
+        if (!bypassCache) {
+            return store;
+        } else if (store instanceof MeteredStore) {
+            final MeteredStore meteredStore = (MeteredStore) store;
+            final StateStore wrapped = meteredStore.wrapped();
+            final StateStore uncached = maybeBypassCache(wrapped, bypassCache);
+            if (uncached == wrapped) {
+                return store;
+            } else {
+                return meteredStore.reWrap(uncached);
```
vvcephei, line 135: And this is where, if we did actually strip out the caching layer, we restore the layers "above" it.

streams/src/test/java/org/apache/kafka/streams/integration/QueryableStateIntegrationTest.java
```diff
@@ -800,6 +811,76 @@ public void shouldBeAbleToQueryMapValuesAfterFilterState() throws Exception {
         }
     }
 
+    @Test
+    public void shouldQueryKeyValueWithCacheBypassed() throws Exception {
```
vvcephei, line 815: This is the test I used to verify the behavior. Note: this test does not verify that the cache is skipped, just that the query works with the cache skipped. I set breakpoints in the code to additionally verify that the cache is actually skipped.

If we go this route, I'd add unit tests so we can mechanically verify that we're skipping the cache.

vvcephei (commented):
@guozhangwang Here's the POC I promised you after our discussion of IQ + cache performance.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
guozhangwang: I had a slight (? well, maybe not so slight) different idea, which assumes that we only want to have the control at the whole app level, rather than a per-store level at IQ APIs only: suppose we just have, e.g. a config on Streams that enables bypass or not for IQ, then upon calling the get/range/etc APIs at the cached layer:

1) Whether the current thread is a stream thread, if yes, never bypass the cache.
2) If not, depend on the config to decide whether bypass the cache.

Doing so we then would not need the unwrap/rewrap and the template types as well. The question then goes back to: whether we would need bypassing at the per-store or global level? Personally I think it's not necessary for a finer-grained control, mainly because that in the long run with the new IQ API and the caller likely to have a totally different way to handle consistency / staleness, i.e. we would deprecate `staleStore` per store or global configs in the long run anyways, and the API calls itself is associated with a token to indicate whether allow staleness or bypass cache.

guozhangwang (commented):
Thanks @vvcephei ! I had a different thought about whether we should allow stores to be configured differently or just globally, if in the long run we just allow each IQ API call itself to indicate staleness / consistency tolerance with some token.

PR #9294, opened 2020-09-16
MINOR: Fix now that kafka.apache.org resolves to 3 IP addresses

jolshan:
ClusterConnectStatesTest and ClientUtilsTest were failing because they expected kafka.apache.org to resolve to 2 IP addresses. This updates the tests so they reflect that DNS resolves to 3 addresses.

### DIFF
--- a/clients/src/test/java/org/apache/kafka/clients/ClientUtilsTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/ClientUtilsTest.java
@@ -102,20 +102,20 @@ public class ClientUtilsTest {
 
     @Test
     public void testResolveDnsLookup() throws UnknownHostException {
-        // Note that kafka.apache.org resolves to 2 IP addresses
+        // Note that kafka.apache.org resolves to 3 IP addresses
         assertEquals(1, ClientUtils.resolve("kafka.apache.org", ClientDnsLookup.DEFAULT).size());
     }
 
     @Test
     public void testResolveDnsLookupAllIps() throws UnknownHostException {
-        // Note that kafka.apache.org resolves to 2 IP addresses
-        assertEquals(2, ClientUtils.resolve("kafka.apache.org", ClientDnsLookup.USE_ALL_DNS_IPS).size());
+        // Note that kafka.apache.org resolves to 3 IP addresses
+        assertEquals(3, ClientUtils.resolve("kafka.apache.org", ClientDnsLookup.USE_ALL_DNS_IPS).size());
     }
 
     @Test
     public void testResolveDnsLookupResolveCanonicalBootstrapServers() throws UnknownHostException {
         // Note that kafka.apache.org resolves to 2 IP addresses
-        assertEquals(2, ClientUtils.resolve("kafka.apache.org", ClientDnsLookup.RESOLVE_CANONICAL_BOOTSTRAP_SERVERS_ONLY).size());
+        assertEquals(3, ClientUtils.resolve("kafka.apache.org", ClientDnsLookup.RESOLVE_CANONICAL_BOOTSTRAP_SERVERS_ONLY).size());
     }
 
     private List<InetSocketAddress> checkWithoutLookup(String... url) {
--- a/clients/src/test/java/org/apache/kafka/clients/ClientUtilsTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/ClientUtilsTest.java
@@ -102,20 +102,20 @@ public class ClientUtilsTest {
 
     @Test
     public void testResolveDnsLookup() throws UnknownHostException {
-        // Note that kafka.apache.org resolves to 3 IP addresses
+        // Note that kafka.apache.org resolves to at least 2 IP addresses
         assertEquals(1, ClientUtils.resolve("kafka.apache.org", ClientDnsLookup.DEFAULT).size());
     }
 
     @Test
     public void testResolveDnsLookupAllIps() throws UnknownHostException {
-        // Note that kafka.apache.org resolves to 3 IP addresses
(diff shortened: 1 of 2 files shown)


abbccdda (commented):
These tests are really fragile, instead of verifying the number of addresses, could we also check the pattern of addresses or something similar to verify the correctness?

jolshan:
@abbccdda I agree that they are fragile. Could you give a little more detail about checking the patterns? I'm not sure if I follow.

dajac:
Netty used a in-memory DNS server for such tests: https://github.com/netty/netty/blob/master/resolver-dns/src/test/java/io/netty/resolver/dns/TestDnsServer.java. I suppose that we could do something similar in AK as well.

What about merging this one and filing a JIRA to improve this in the future? All these flaky tests due to this are quite annoying so we could already fix them. What do you think?

It would be great if we could backport it to previous versions as well.

mumrah:
Agreed with @dajac's suggestion. Maybe for this change we could have the assertion as > 1 rather than equalling a specific value

jolshan:
I also like @dajac's suggestion. I've updated the PR so that the tests assert more than 1 address as @mumrah said.

jolshan:
Also created the issue https://issues.apache.org/jira/browse/KAFKA-10496

clients/src/test/java/org/apache/kafka/clients/ClientUtilsTest.java
```diff
@@ -102,20 +102,20 @@ public void testResolveUnknownHostException() throws UnknownHostException {
 
     @Test
     public void testResolveDnsLookup() throws UnknownHostException {
-        // Note that kafka.apache.org resolves to 2 IP addresses
+        // Note that kafka.apache.org resolves to at least 2 IP addresses
```
mumrah, line 105: Comment here and below refer to 2 IP addresses still.

clients/src/test/java/org/apache/kafka/clients/ClusterConnectionStatesTest.java
```diff
@@ -273,22 +273,21 @@ public void testMultipleIPsWithDefault() throws UnknownHostException {
...
 
         connectionStates.connecting(nodeId1, time.milliseconds(), hostTwoIps, ClientDnsLookup.USE_ALL_DNS_IPS);
         InetAddress addr1 = connectionStates.currentAddress(nodeId1);
         connectionStates.connecting(nodeId1, time.milliseconds(), hostTwoIps, ClientDnsLookup.USE_ALL_DNS_IPS);
         InetAddress addr2 = connectionStates.currentAddress(nodeId1);
         assertNotSame(addr1, addr2);
-
         connectionStates.connecting(nodeId1, time.milliseconds(), hostTwoIps, ClientDnsLookup.USE_ALL_DNS_IPS);
         InetAddress addr3 = connectionStates.currentAddress(nodeId1);
-        assertSame(addr1, addr3);
```
mumrah, line 286: Hmm, so previously (and also with this change) we are assuming that we will get different resolved IPs for each connection to apache.kafka.org? This seems to rely on round-robbin DNS resolution that we can't really control. I think these assertions will always be prone to failure unless we can control the DNS server like @dajac suggested.

I guess we can commit this to unblock the tests, but we should definitely prioritize a better fix.

jolshan: Yeah. I was wondering if we should even keep this test in the current form, and I'm hoping that a DNS we can control would resolve this issue.

clients/src/test/java/org/apache/kafka/clients/ClientUtilsTest.java
jolshan: Are you saying to change it back? I thought it was fair to say "at least 2" to reflect we are asserting greater than 1.

mumrah (approved):
Tests looking much better, just the one or two flaky tests as usual rather than dozens of failures :)

mumrah:
Thanks for the patch @jolshan!

ijuma:
Have we cherry picked this to older branches? It seems like they will be broken otherwise.

clients/src/test/java/org/apache/kafka/clients/ClusterConnectionStatesTest.java
ijuma: The test is verifying the round robin that _our_ code does, not the DNS server. We should ensure we still have that coverage in the meantime.

mumrah: Thanks for clarifying, @ijuma. Seems like we could verify the round-robin behavior without relying on specific number of IPs.

mumrah:
2.6 backport applied cleanly, kicked off a build here https://ci-builds.apache.org/job/Kafka/job/kafka-2.6-jdk8/15/

~2.5 had a conflict, and one test fails for me locally. I'll look into it along with other branches this evening~
Edit: looks like the behavior of ClientUtils#resolve changed in 2.6 (84020bfc13) so the fix will be slightly different for older versions. How far back should we backport? I see jenkins jobs back to 2.0

clients/src/test/java/org/apache/kafka/clients/ClusterConnectionStatesTest.java
jolshan: I see. I changed it to show that the first three are different but I didn't include the check that it loops back to the first IP since that part seems the most fragile.

andrewegel:
> How far back should we backport?

Checking the AK -> CCS merge Job branches, the 2.3 through 2.5 branch unit tests are failing on the tests that this fixes.

ijuma:
@mumrah I think 2.0 and higher would be a good target. We don't need separate PRs as long as the cherry pick is clean.

PR #9297, opened 2020-09-17
KAFKA-10495:Fix spelling mistake

ouyangnengda:
The iff is wrong word.I fixed it to if.


dongjinleekr:
Hi @ouyangnengda,

I think it is not a problem. 'iff' means 'if and only if'. Please see [here](https://www.oxfordlearnersdictionaries.com/definition/english/iff).

ouyangnengda:
ok

PR #9299, opened 2020-09-17
MINOR: Use `Map.forKeyValue` to avoid tuple allocation in Scala 2.13

ijuma:
`foreachKv` invokes `foreachEntry` in Scala 2.13 and falls back to
`foreach` in Scala 2.12.

This change requires a newer version of scala-collection-compat, so
update it to the latest version (2.2.0).

Finally, included a minor clean-up in `GetOffsetShell` to use `toArray`
before `sortBy` since it's more efficient.

### DIFF
--- a/core/src/main/scala/kafka/admin/ZkSecurityMigrator.scala
+++ b/core/src/main/scala/kafka/admin/ZkSecurityMigrator.scala
@@ -20,6 +20,7 @@ package kafka.admin
 import joptsimple.{ArgumentAcceptingOptionSpec, OptionSet}
 import kafka.server.KafkaConfig
 import kafka.utils.{CommandDefaultOptions, CommandLineUtils, Exit, Logging}
+import kafka.utils.Implicits._
 import kafka.zk.{ControllerZNode, KafkaZkClient, ZkData, ZkSecurityMigratorUtils}
 import org.apache.kafka.common.security.JaasUtils
 import org.apache.kafka.common.utils.{Time, Utils}
@@ -128,7 +129,7 @@ object ZkSecurityMigrator extends Logging {
     // Now override any set system properties with explicitly-provided values from the config file
     // Emit INFO logs due to camel-case property names encouraging mistakes -- help people see mistakes they make
     info(s"Found ${zkTlsConfigFileProps.size()} ZooKeeper client configuration properties in file $filename")
-    zkTlsConfigFileProps.asScala.foreach { case (key, value) =>
+    zkTlsConfigFileProps.asScala.foreachKv { (key, value) =>
       info(s"Setting $key")
       KafkaConfig.setZooKeeperClientProperty(zkClientConfig, key, value)
     }
--- a/core/src/main/scala/kafka/admin/ZkSecurityMigrator.scala
+++ b/core/src/main/scala/kafka/admin/ZkSecurityMigrator.scala
@@ -129,7 +129,7 @@ object ZkSecurityMigrator extends Logging {
     // Now override any set system properties with explicitly-provided values from the config file
     // Emit INFO logs due to camel-case property names encouraging mistakes -- help people see mistakes they make
     info(s"Found ${zkTlsConfigFileProps.size()} ZooKeeper client configuration properties in file $filename")
-    zkTlsConfigFileProps.asScala.foreachKv { (key, value) =>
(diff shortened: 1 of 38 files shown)


core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -133,7 +133,7 @@ object GetOffsetShell {
         }
     }
 
-    partitionOffsets.toSeq.sortBy { case (tp, _) => tp.partition }.foreach { case (tp, offset) =>
+    partitionOffsets.toArray.sortBy { case (tp, _) => tp.partition }.foreach { case (tp, offset) =>
```
ijuma, line 136: Unrelated clean-up I noticed.

jsancio (approved):
LGTM! Thanks for the improvement.

core/src/main/scala/kafka/utils/Implicits.scala
```diff
@@ -46,4 +47,21 @@ object Implicits {
...
+   * (with the help of scala.collection.compat). `foreachEntry` avoids the tuple allocation and
+   * is more efficient.
+   *
+   * This was not named `foreachEntry` to avoid `unused import` warnings in Scala 2.13 (the implicit
+   * would not be triggered in Scala 2.13 since `Map.foreachEntry` would have precedence).
+   */
+  @nowarn("cat=unused-imports")
+  implicit class MapExtensionMethods[K, V](private val self: scala.collection.Map[K, V]) extends AnyVal {
+    import scala.collection.compat._
+    def foreachKv[U](f: (K, V) => U): Unit = {
```
hachikuji: The name reads a tad awkwardly. I wonder if `foreachKeyValue` would be too verbose. Or maybe `foreachMapEntry`?

hachikuji (approved):
LGTM. Added a couple alternative names for consideration, but will leave it up to you.

core/src/main/scala/kafka/utils/Implicits.scala
jsancio: Or `foreachPair`.

ijuma: Thanks for the feedback. I agree that the name could be improved. I was hoping for something concise and descriptive. `foreachPair` is not bad, but you'd expect to get a pair (tuple) and this is to avoid that. How about `forKeyValue`?

ijuma: Also, note that this will go away once we drop support for Scala 2.12.

jsancio: `forKeyValue` LGTM since `for` is the language's keyword for `foreach` or `map`.

ijuma: That was the motivation indeed. Great, I'll change it.

dajac:
'forKeyValue' sounds good to me as well.

core/src/main/scala/kafka/admin/ZkSecurityMigrator.scala
```diff
@@ -128,7 +129,7 @@ object ZkSecurityMigrator extends Logging {
     // Now override any set system properties with explicitly-provided values from the config file
     // Emit INFO logs due to camel-case property names encouraging mistakes -- help people see mistakes they make
     info(s"Found ${zkTlsConfigFileProps.size()} ZooKeeper client configuration properties in file $filename")
-    zkTlsConfigFileProps.asScala.foreach { case (key, value) =>
+    zkTlsConfigFileProps.asScala.forKeyValue { (key, value) =>
```
chia7712, line 132: Could we replace such ```asScala.forKeyValue``` by java ```Map.forEach``` ? We don't need to convert it to scala collection actually.

ijuma: See https://github.com/apache/kafka/commit/847ff8f55735568fe879d490af297e047b994f5e for an explanation why `asScala` still makes sense when dealing with `Properties`.

core/src/main/scala/kafka/utils/Implicits.scala
ijuma: Checked with @hachikuji and he's fine with this change.

PR #9300, opened 2020-09-17
KAFKA-10491: Check authorizations first in KafkaApis

rondagostino:
comment to ping reviewers.*

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaApis.scala
+++ b/core/src/main/scala/kafka/server/KafkaApis.scala
@@ -1770,45 +1770,57 @@ class KafkaApis(val requestChannel: RequestChannel,
       sendResponseMaybeThrottle(controllerMutationQuota, request, createResponse, onComplete = None)
     }
 
+    // be sure to check authorization first, before checking if this is the controller, to avoid leaking
+    // information about the system (i.e. who is the controller) to principals unauthorized for that information
+
     val createTopicsRequest = request.body[CreateTopicsRequest]
     val results = new CreatableTopicResultCollection(createTopicsRequest.data.topics.size)
-    if (!controller.isActive) {
-      createTopicsRequest.data.topics.forEach { topic =>
-        results.add(new CreatableTopicResult().setName(topic.name)
-          .setErrorCode(Errors.NOT_CONTROLLER.code))
-      }
-      sendResponseCallback(results)
-    } else {
-      createTopicsRequest.data.topics.forEach { topic =>
-        results.add(new CreatableTopicResult().setName(topic.name))
-      }
-      val hasClusterAuthorization = authorize(request.context, CREATE, CLUSTER, CLUSTER_NAME,
-        logIfDenied = false)
-      val topics = createTopicsRequest.data.topics.asScala.map(_.name)
-      val authorizedTopics =
-        if (hasClusterAuthorization) topics.toSet
-        else filterByAuthorized(request.context, CREATE, TOPIC, topics)(identity)
-      val authorizedForDescribeConfigs = filterByAuthorized(request.context, DESCRIBE_CONFIGS, TOPIC,
-        topics, logIfDenied = false)(identity).map(name => name -> results.find(name)).toMap
+    createTopicsRequest.data.topics.forEach { topic =>
+      results.add(new CreatableTopicResult().setName(topic.name))
+    }
+    val hasClusterAuthorization = authorize(request.context, CREATE, CLUSTER, CLUSTER_NAME,
+      logIfDenied = false)
+    val topics = createTopicsRequest.data.topics.asScala.map(_.name)
+    val authorizedTopics =
+      if (hasClusterAuthorization) topics.toSet
+      else filterByAuthorized(request.context, CREATE, TOPIC, topics)(identity)
+    val authorizedForDescribeConfigs = filterByAuthorized(request.context, DESCRIBE_CONFIGS, TOPIC,
+      topics, logIfDenied = false)(identity).map(name => name -> results.find(name)).toMap
 
-      results.forEach { topic =>
-        if (results.findAll(topic.name).size > 1) {
-          topic.setErrorCode(Errors.INVALID_REQUEST.code)
-          topic.setErrorMessage("Found multiple entries for this topic.")
-        } else if (!authorizedTopics.contains(topic.name)) {
-          topic.setErrorCode(Errors.TOPIC_AUTHORIZATION_FAILED.code)
-          topic.setErrorMessage("Authorization failed.")
-        }
-        if (!authorizedForDescribeConfigs.contains(topic.name)) {
-          topic.setTopicConfigErrorCode(Errors.TOPIC_AUTHORIZATION_FAILED.code)
-        }
+    results.forEach { topic =>
+      if (results.findAll(topic.name).size > 1) {
+        topic.setErrorCode(Errors.INVALID_REQUEST.code)
+        topic.setErrorMessage("Found multiple entries for this topic.")
+      } else if (!authorizedTopics.contains(topic.name)) {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1732,45 +1732,57 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+    createTopicsRequest.data.topics.forEach { topic =>
+      results.add(new CreatableTopicResult().setName(topic.name))
+    }
+    val hasClusterAuthorization = authorize(request.context, CREATE, CLUSTER, CLUSTER_NAME,
+      logIfDenied = false)
+    val topics = createTopicsRequest.data.topics.asScala.map(_.name)
+    val authorizedTopics =
+      if (hasClusterAuthorization) topics.toSet
+      else filterByAuthorized(request.context, CREATE, TOPIC, topics)(identity)
+    val authorizedForDescribeConfigs = filterByAuthorized(request.context, DESCRIBE_CONFIGS, TOPIC,
```
hachikuji, line 1787: Might not matter since `CreateTopics` requests are infrequent, but the two passes for authorization are a bit vexing. Feels like we are missing a good intermediate type between this handler and `AdminManager`. Maybe we can replace the 3 maps that we pass to `AdminManager.createTopic` with a single map which contains all the state we need for each topic.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1732,45 +1732,57 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+    createTopicsRequest.data.topics.forEach { topic =>
+      if (results.find(topic.name).errorCode == Errors.NONE.code) {
+        toCreate += topic.name -> topic
+      }
+    }
+    if (!controller.isActive) {
+      // don't provide the information that this node is not the controller unless they were authorized
+      // to perform at least one of their requests
+      sendResponseCallback(
+        if (toCreate.isEmpty) {
```
hachikuji: Hmm.. If a topic is unauthorized, I think it's more important to return the authorization failure. There's not any reason I can think of why `NOT_CONTROLLER` should take precedence. It would just cause the request to be unnecessarily retried against the new coordinator. Similarly for the other APIs.

core/src/main/scala/kafka/server/KafkaApis.scala
rondagostino: > Might not matter since CreateTopics requests are infrequent

I think it doesn't matter.  I agree with you this could be more elegant but would rather not deal with it in this PR if that's ok.

core/src/main/scala/kafka/server/KafkaApis.scala
rondagostino: All set, please take a look and let me know if what I did is reasonable.  In particular, take a look at the delete topics test because there are several possible combinations and I tested for all of them.  The specific case I am wondering about is when a topic is being deleted but it doesn't exist and the request is being sent to the non-controller; it responds with UNKNOWN_TOPIC_OR_PARTITION.  Technically since the broker is not the controller I thought perhaps it wouldn't have the most up-to-date information, so I'm not 100% certain about this choice.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1770,45 +1770,47 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-        }
-        if (!authorizedForDescribeConfigs.contains(topic.name)) {
-          topic.setTopicConfigErrorCode(Errors.TOPIC_AUTHORIZATION_FAILED.code)
-        }
-      }
-      val toCreate = mutable.Map[String, CreatableTopic]()
-      createTopicsRequest.data.topics.forEach { topic =>
-        if (results.find(topic.name).errorCode == Errors.NONE.code) {
-          toCreate += topic.name -> topic
+        if(topic.errorCode() == Errors.NONE.code()) {
```
hachikuji, line 1808: nit: convention is to add space after `if`. There are a few of these in the patch

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1770,45 +1770,47 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-        logIfDenied = false)
-      val topics = createTopicsRequest.data.topics.asScala.map(_.name)
-      val authorizedTopics =
-        if (hasClusterAuthorization) topics.toSet
-        else filterByAuthorized(request.context, CREATE, TOPIC, topics)(identity)
-      val authorizedForDescribeConfigs = filterByAuthorized(request.context, DESCRIBE_CONFIGS, TOPIC,
-        topics, logIfDenied = false)(identity).map(name => name -> results.find(name)).toMap
-
+    }
+    if (!controller.isActive) {
```
hachikuji, line 1802: I think we could probably roll this into the loop above. What I'm thinking is something like this:
```scala
    val toCreate = mutable.Map[String, CreatableTopic]()

    results.forEach { topic =>
      if (results.findAll(topic.name).size > 1) {
        topic.setErrorCode(Errors.INVALID_REQUEST.code)
        topic.setErrorMessage("Found multiple entries for this topic.")
      } else if (!authorizedTopics.contains(topic.name)) {
        topic.setErrorCode(Errors.TOPIC_AUTHORIZATION_FAILED.code)
        topic.setErrorMessage("Authorization failed.")
      } else if (!controller.isActive) {
        topic.setErrorCode(Errors.NOT_CONTROLLER.code)
      } else {
        toCreate += topic.name -> topic
      }
      if (!authorizedForDescribeConfigs.contains(topic.name)) {
        topic.setTopicConfigErrorCode(Errors.TOPIC_AUTHORIZATION_FAILED.code)
      }
   }

   if (toCreate.isEmpty) {
     sendResponseCallback(results)
   } else {
     ...
```

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1936,25 +1945,38 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-      // If no authorized topics return immediately
-      if (toDelete.isEmpty)
-        sendResponseCallback(results)
-      else {
-        def handleDeleteTopicsResults(errors: Map[String, Errors]): Unit = {
-          errors.foreach {
-            case (topicName, error) =>
-              results.find(topicName)
-                .setErrorCode(error.code)
+      if (!controller.isActive) {
```
hachikuji, line 1948: Same as above. I think it's a little more natural to roll the controller check into the loop above. If you do that, then the structure here will match what I suggested for create topic.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3078,15 +3100,22 @@ class KafkaApis(val requestChannel: RequestChannel,
...
 
-    if (!controller.isActive) {
-      sendResponseMaybeThrottle(request, requestThrottleMs =>
-        alterUserScramCredentialsRequest.getErrorResponse(requestThrottleMs, Errors.NOT_CONTROLLER.exception))
-    } else if (authorize(request.context, ALTER, CLUSTER, CLUSTER_NAME)) {
-      val result = adminManager.alterUserScramCredentials(
-        alterUserScramCredentialsRequest.data.upsertions().asScala, alterUserScramCredentialsRequest.data.deletions().asScala)
-      sendResponseMaybeThrottle(request, requestThrottleMs =>
-        new AlterUserScramCredentialsResponse(result.setThrottleTimeMs(requestThrottleMs)))
+    // be sure to check authorization first, before checking if this is the controller, to avoid leaking
```
hachikuji, line 3103: More of a nit, but the comments seem to overstate the importance of not leaking the controller. I think this is really more about establishing a consistent practice of always checking authorization first before doing anything else.

PR #9302, opened 2020-09-18
KAFKA-10149: Allow auto preferred leader election when partitions are reassigning

bob-barrett:
This patch removes the check for reassigning partitions when determining whether to trigger automatic leader election. This check can cause problems during long-running reassignments because a crashed broker can leave the partition leaderships in an unexpectedly unbalanced state. 

### DIFF
--- a/core/src/main/scala/kafka/controller/KafkaController.scala
+++ b/core/src/main/scala/kafka/controller/KafkaController.scala
@@ -1067,10 +1067,7 @@ class KafkaController(val config: KafkaConfig,
       // check ratio and if greater than desired ratio, trigger a rebalance for the topic partitions
       // that need to be on this broker
       if (imbalanceRatio > (config.leaderImbalancePerBrokerPercentage.toDouble / 100)) {
-        // do this check only if the broker is live and there are no partitions being reassigned currently
-        // and preferred replica election is not in progress
         val candidatePartitions = topicsNotInPreferredReplica.keys.filter(tp =>
-          controllerContext.partitionsBeingReassigned.isEmpty &&
           !topicDeletionManager.isTopicQueuedUpForDeletion(tp.topic) &&
           controllerContext.allTopics.contains(tp.topic) &&
           canPreferredReplicaBeLeader(tp)
--- a/core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
+++ b/core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
@@ -445,6 +445,29 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
   }
 
   @Test
+  def testAutoPreferredReplicaLeaderElectionWithReassigningPartitions(): Unit = {
+    servers = makeServers(2, autoLeaderRebalanceEnable = true)
+    val controllerId = TestUtils.waitUntilControllerElected(zkClient)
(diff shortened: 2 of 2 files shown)


core/src/main/scala/kafka/controller/KafkaController.scala
```diff
@@ -1067,10 +1067,7 @@ class KafkaController(val config: KafkaConfig,
       // check ratio and if greater than desired ratio, trigger a rebalance for the topic partitions
       // that need to be on this broker
       if (imbalanceRatio > (config.leaderImbalancePerBrokerPercentage.toDouble / 100)) {
-        // do this check only if the broker is live and there are no partitions being reassigned currently
-        // and preferred replica election is not in progress
         val candidatePartitions = topicsNotInPreferredReplica.keys.filter(tp =>
-          controllerContext.partitionsBeingReassigned.isEmpty &&
```
dajac, line 1073: @bob-barrett It totally makes sense to not block it the auto preferred leader election if there is any ongoing reassignments in the cluster. However, I wonder if we should continue to disallow it for the partitions being reassigned in order to not mess up with the their ongoing reassignment. I am not sure if this is a real concern or not though. Have you thought about this case?

hachikuji: It's a good question. During a reassignment, the adding replicas are always listed first which means the preferred leader is among the target replicas. My take is that we want to move the leadership onto the new preferred leader as soon as possible since the whole point of the reassignment is to take load off the removing replicas.

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
```diff
@@ -444,6 +444,29 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
...
+    val reassigningTp = new TopicPartition("reassigning", 0)
+    val reassigningAssignment = Map(reassigningTp.partition -> Seq(0))
+    TestUtils.createTopic(zkClient, reassigningTp.topic, partitionReplicaAssignment = reassigningAssignment, servers = servers)
+    servers(otherBrokerId).shutdown()
+    servers(otherBrokerId).awaitShutdown()
+    waitForPartitionState(tp, firstControllerEpoch, controllerId, LeaderAndIsr.initialLeaderEpoch + 1,
+      "failed to get expected partition state upon broker shutdown")
+
+    // Directly edit the controller context to simulate a reassigning partition
+    servers(controllerId).kafkaController.controllerContext.partitionsBeingReassigned.add(reassigningTp)
```
hachikuji, line 464: Feels a bit nasty to dip into the controller context directly. An alternative way of testing this might be something like this:

1. Start with two partitions
2. Reassign one partition to a new replica with a low throttle
3. Reassign the other partition so that just the preferred leader changes
4. Trigger preferred election and verify that the leader of the second partition has changed

core/src/main/scala/kafka/controller/KafkaController.scala
dajac: I totally agree. The sooner the better. I was wondering if there could be any conflicts with the reassignment logic. I have looked at it and I haven't seen anything. 👍

ijuma:
Is this still relevant?

songnon:
I think it's still relevant. Not sure if @bob-barrett still have plan to address it. Otherwise, I'll create a new PR.

songnon:
New PR https://github.com/apache/kafka/pull/12543

PR #9304, opened 2020-09-18
KAFKA-10502:TimestampRouter may occur  threadlocal leak

huangyiminghappy:
Threadlocal  may can not set null,because it may create a memory leak, you can see the link:
https://stackoverflow.com/questions/12424838/threadlocal-remove, so I think weather we can invoke thread local.remove instead it


huangyiminghappy:
@hachikuji  would  you help to review this pr.

guozhangwang:
Reading on the SO thread I think this fix makes sense. Merging to trunk now.

huangyiminghappy:
> Reading on the SO thread I think this fix makes sense. Merging to trunk now.

thanks

PR #9309, opened 2020-09-19
KAFKA-10503: MockProducer doesn't throw ClassCastException when no

gmunozfe:
partition for topic

Though MockProducer admits serializers in its constructors, it doesn't check during send method that those serializers are the proper ones to serialize key/value included into the ProducerRecord.

This check is only done if there is a partition assigned for that topic.

It would be an enhancement if these serialize methods were also invoked in simple scenarios, where no partition is assigned to a topic.

Added a unit test to check expected behaviour

### DIFF
--- a/clients/src/test/java/org/apache/kafka/clients/producer/MockProducerTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/producer/MockProducerTest.java
@@ -24,6 +24,7 @@ import org.apache.kafka.common.KafkaException;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.ProducerFencedException;
+import org.apache.kafka.common.serialization.IntegerSerializer;
 import org.apache.kafka.common.serialization.StringSerializer;
 import org.apache.kafka.test.MockSerializer;
 import org.junit.After;
@@ -43,6 +44,7 @@ import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -764,6 +766,14 @@ public class MockProducerTest {
             producer.flush();
             fail("Should have thrown as producer is already closed");
         } catch (IllegalStateException e) { }
+    }
+    
+    @Test
+    @SuppressWarnings("unchecked")
+    public void shouldThrowClassCastException() {
+        try (MockProducer<Integer, String> customProducer = new MockProducer<>(true, new IntegerSerializer(), new StringSerializer());) {
+            assertThrows(ClassCastException.class, () -> customProducer.send(new ProducerRecord(topic, "key1", "value1")));
+        }
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


gmunozfe:
@mjsax @guozhangwang @leonardge could you review? thanks!

clients/src/test/java/org/apache/kafka/clients/producer/MockProducerTest.java
```diff
@@ -765,6 +766,16 @@ public void shouldThrowOnFlushProducerIfProducerIsClosed() {
             fail("Should have thrown as producer is already closed");
         } catch (IllegalStateException e) { }
     }
+    
+    @Test
+    @SuppressWarnings("unchecked")
```
mjsax, line 772: We should avoid to suppress warning. Seems, you need to add generics to `ProducerRecord` in L774 to fix it?

clients/src/test/java/org/apache/kafka/clients/producer/MockProducerTest.java
```diff
@@ -765,6 +766,16 @@ public void shouldThrowOnFlushProducerIfProducerIsClosed() {
...
+    public void shouldThrowClassCastException() {
+        try (MockProducer<Integer, String> producer = new MockProducer<>(true, new IntegerSerializer(), new StringSerializer());) {
+            ProducerRecord record = new ProducerRecord(topic, "key1", "value1");
+            producer.send(record);
+            fail("Should have thrown ClassCastException because record cannot be casted with serializers");
```
mjsax: instead of `fail` we should use ```
assertThrows(
    ClassCastException.class,
    () -> producer.send(record);
);
```

For this case, we can remove the `catch` block below.

clients/src/test/java/org/apache/kafka/clients/producer/MockProducerTest.java
gmunozfe: Thanks a lot for your review @mjsax 
If I add generics, bounded type parameters will make compilation fail, as types are checked and incompatible.
In fact, the purpose of this enhancement is that MockProducer throws the exception when types are not restricted, wdyt?

clients/src/test/java/org/apache/kafka/clients/producer/MockProducerTest.java
gmunozfe: Done, I have followed the same approach that other tests in the same class, but assertThrows is also a good alternative

mjsax: Oh, I see. We actually want to pass in an incorrect key-type to get the `ClassCastException`. So you are right, for this case we cannot specify the generics. Keeping the code as-is an to suppress the warning is fine than. Totally missed this.

clients/src/test/java/org/apache/kafka/clients/producer/MockProducerTest.java
mjsax: Oh, I see. We actually want to pass in an incorrect key-type to get the ClassCastException. So you are right, for this case we cannot specify the generics. Keeping the code as-is and to suppress the warning is fine than. Totally missed this. Thanks for clarifying.

clients/src/test/java/org/apache/kafka/clients/producer/MockProducerTest.java
mjsax: Yes, we used an older version of the test libraries and thus existing code uses a different pattern. We just try to use the newer (and better) pattern in new code and migrate existing code lazily to the new pattern.

mjsax:
Thanks for the PR @gmunozfe. Merged to `trunk`.

PR #9311, opened 2020-09-19
KAFKA-9910: Implement new transaction timed out error

zhaohaidao:
The producer recovers by internally sending InitProducerId with the current epoch when received TRANSACTION_TIMED_OUT error.


clients/src/main/java/org/apache/kafka/common/errors/TransactionTimeoutException.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public TransactionTimeoutException() {
+        super();
+    }
+
+    public TransactionTimeoutException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public TransactionTimeoutException(String message) {
```
abbccdda, line 27: Do we need all other constructors?

clients/src/main/java/org/apache/kafka/common/errors/TransactionTimeoutException.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.common.errors;
+
+/**
+ * This exception indicates that the last ongoing transaction timed out on the coordinator.
+ * When encountering this exception, the producer should retry initialization with current epoch.
+ */
+public class TransactionTimeoutException extends ApiException {
```
abbccdda, line 23: `TransactionTimedOut`

clients/src/test/java/org/apache/kafka/clients/producer/internals/TransactionManagerTest.java
```diff
@@ -1577,6 +1578,59 @@ public void testInvalidProducerEpochConvertToProducerFencedInAddPartitionToTxn()
...
+
+        assertFalse(responseFuture.isDone());
+        prepareAddOffsetsToTxnResponse(Errors.TRANSACTION_TIMED_OUT, consumerGroupId, producerId, epoch);
+
+        verifyTxnTimeout(responseFuture);
+    }
+
+    @Test
+    public void testTxnTimeoutInEndTxn() throws InterruptedException {
+
```
abbccdda: nit: remove empty line

clients/src/test/java/org/apache/kafka/clients/producer/internals/TransactionManagerTest.java
```diff
@@ -1577,6 +1578,59 @@ public void testInvalidProducerEpochConvertToProducerFencedInAddPartitionToTxn()
...
+        Future<RecordMetadata> responseFuture = appendToAccumulator(tp0);
+
+        assertFalse(responseFuture.isDone());
+        prepareAddPartitionsToTxnResponse(Errors.NONE, tp0, epoch, producerId);
+        prepareProduceResponse(Errors.NONE, producerId, epoch);
+        prepareEndTxnResponse(Errors.TRANSACTION_TIMED_OUT, TransactionResult.COMMIT, producerId, epoch);
+
+        runUntil(commitResult::isCompleted);
+        runUntil(responseFuture::isDone);
+
```
abbccdda: nit: remove empty line

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -1072,7 +1076,11 @@ private void transitionTo(State target, RuntimeException error) {
             if (error == null)
                 throw new IllegalArgumentException("Cannot transition to " + target + " with a null exception");
             lastError = error;
+            abortableError = error;
         } else {
+            if (target != State.ABORTING_TRANSACTION) {
```
abbccdda, line 1094: Does this relate to the optimization https://issues.apache.org/jira/browse/KAFKA-10504?

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -1585,6 +1596,8 @@ public void handleResponse(AbstractResponse response) {
                 // We could still receive INVALID_PRODUCER_EPOCH from old versioned transaction coordinator,
                 // just treat it the same as PRODUCE_FENCED.
                 fatalError(Errors.PRODUCER_FENCED.exception());
+            } else if (error == Errors.TRANSACTION_TIMED_OUT) {
```
abbccdda: Could we merge with other cases on L1452?

clients/src/test/java/org/apache/kafka/clients/producer/internals/TransactionManagerTest.java
```diff
@@ -2959,8 +3035,7 @@ public void testBumpTransactionalEpochOnRecoverableAddPartitionRequestError() {
         assertTrue(transactionManager.isReady());  // make sure we are ready for a transaction now.
     }
 
-    @Test
-    public void testBumpTransactionalEpochOnRecoverableAddOffsetsRequestError() throws InterruptedException {
+    private void verifyBumpTransactionalEpochOnRecoverableAddOffsetsRequestError(Errors errors) throws InterruptedException {
```
abbccdda, line 3090: nit: this func could be put after the callers

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
```diff
@@ -381,24 +385,35 @@ class TransactionCoordinator(brokerId: Int,
...
+            } else if (txnMetadata.pendingTransitionInProgress
+              && !txnMetadata.pendingState.contains(PrepareEpochFence)
+              && !txnMetadata.pendingState.contains(PrepareEpochBumpThenAbort))
               Left(Errors.CONCURRENT_TRANSACTIONS)
             else txnMetadata.state match {
               case Ongoing =>
                 val nextState = if (txnMarkerResult == TransactionResult.COMMIT)
                   PrepareCommit
-                else
+                else {
```
abbccdda: nit: could remove the braces

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
```diff
@@ -381,24 +385,35 @@ class TransactionCoordinator(brokerId: Int,
...
                   PrepareAbort
-
-                if (nextState == PrepareAbort && txnMetadata.pendingState.contains(PrepareEpochFence)) {
+                }
+                if (nextState == PrepareAbort && (txnMetadata.pendingState.get == PrepareEpochFence
+                  || txnMetadata.pendingState.get == PrepareEpochBumpThenAbort)) {
                   // We should clear the pending state to make way for the transition to PrepareAbort and also bump
                   // the epoch in the transaction metadata we are about to append.
-                  isEpochFence = true
+                  isEpochFence = txnMetadata.pendingState.get == PrepareEpochFence
```
abbccdda: Correct me if wrong, but it seems that the reasoning for adding `PrepareEpochBumpThenAbort` is to avoid setting `lastProducerEpoch` to -1 in the time out state? If so, I was wondering whether we have a simpler way to set the lastProducerEpoch when we make the transition, just as in `prepareTransitionTo`

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -2045,9 +2045,11 @@ class KafkaApis(val requestChannel: RequestChannel,
     if (authorize(request.context, WRITE, TRANSACTIONAL_ID, transactionalId)) {
       def sendResponseCallback(error: Errors): Unit = {
         def createResponse(requestThrottleMs: Int): AbstractResponse = {
+          val needOld = endTxnRequest.version < 2 && (error == Errors.PRODUCER_FENCED || error == Errors
+            .TRANSACTION_TIMED_OUT)
           val finalError =
-            if (endTxnRequest.version < 2 && error == Errors.PRODUCER_FENCED) {
-              // For older clients, they could not understand the new PRODUCER_FENCED error code,
+            if (needOld) {
```
abbccdda, line 2059: We should add a helper to return old error code, such as `maybeReturnOldClientError(Errors error, short requestVersion)`

abbccdda (commented):
Thanks for the great PR, got a couple of comments

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
zhaohaidao: yes

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
zhaohaidao: According to my understanding, the semantics of PrepareEpochFence and  PrepareEpochBumpThenAbort 
 are different. If a new state is not introduced, what should the parameter `newState` fill in when transitionTo is called?

zhaohaidao: Thanks for your advice. I found a simpler way to set the lastProducerEpoch and PrepareEpochBumpThenAbort is removed

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -1198,18 +1209,23 @@ boolean canBumpEpoch() {
         return coordinatorSupportsBumpingEpoch;
     }
 
+    private void resetTransactions() {
+        newPartitionsInTransaction.clear();
+        pendingPartitionsInTransaction.clear();
+        partitionsInTransaction.clear();
+
```
abbccdda: nit: remove new line

clients/src/main/java/org/apache/kafka/common/requests/EndTxnResponse.java
```diff
@@ -34,6 +34,7 @@
  *   - {@link Errors#INVALID_PRODUCER_ID_MAPPING}
  *   - {@link Errors#INVALID_PRODUCER_EPOCH} // for version <=1
  *   - {@link Errors#PRODUCER_FENCED}
+ *   - {@link Errors#TRANSACTION_TIMED_OUT}
```
abbccdda, line 37: We should also add this to the `AddOffsetsToTxnResponse`

clients/src/main/java/org/apache/kafka/common/errors/TransactionTimeOutException.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.common.errors;
+
+/**
+ * This exception indicates that the last ongoing transaction timed out on the coordinator.
+ * When encountering this exception, the producer should retry initialization with current epoch.
+ */
+public class TransactionTimeOutException extends ApiException {
```
hachikuji: Can we change the name to `TransactionTimeoutException`. Seems more consistent to use lower-case 'o' (see e.g. `TimeoutException`).

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -369,7 +372,9 @@ private TransactionalRequestResult beginCompletingTransaction(TransactionResult
         // If the error is an INVALID_PRODUCER_ID_MAPPING error, the server will not accept an EndTxnRequest, so skip
         // directly to InitProducerId. Otherwise, we must first abort the transaction, because the producer will be
         // fenced if we directly call InitProducerId.
-        if (!(lastError instanceof InvalidPidMappingException)) {
+        boolean needEndTxn = !(abortableError instanceof InvalidPidMappingException)
```
abbccdda, line 390: We should also update the comment above. And I think we could extract `needEndTxn` as a helper function, if it is true universally to check abortableError for txn ending.

clients/src/main/java/org/apache/kafka/common/requests/AddPartitionsToTxnResponse.java
```diff
@@ -40,6 +40,7 @@
  *   - {@link Errors#INVALID_PRODUCER_ID_MAPPING}
  *   - {@link Errors#INVALID_PRODUCER_EPOCH} // for version <=1
  *   - {@link Errors#PRODUCER_FENCED}
+ *   - {@link Errors#TRANSACTION_TIMED_OUT}
```
abbccdda, line 43: We should also update the comment in corresponding JSON files to mention about the new error code, such as AddPartition/AddOffsets/EndTxn request/response.json

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -2336,6 +2335,11 @@ class KafkaApis(val requestChannel: RequestChannel,
     }
   }
 
+  def maybeReturnOldClientError(error: Errors, requestVersion: Short): Boolean = {
+    requestVersion < 2 && (error == Errors.PRODUCER_FENCED || error == Errors
```
hachikuji: We need to bump the protocol versions for the new error code. Otherwise older clients using version 2 will not know how to handle `TRANSACTION_TIMED_OUT` and will translate it to `UNKNOWN_SERVER_ERROR`, which is a regression.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -2083,8 +2083,8 @@ class KafkaApis(val requestChannel: RequestChannel,
       def sendResponseCallback(error: Errors): Unit = {
         def createResponse(requestThrottleMs: Int): AbstractResponse = {
           val finalError =
-            if (endTxnRequest.version < 2 && error == Errors.PRODUCER_FENCED) {
-              // For older clients, they could not understand the new PRODUCER_FENCED error code,
+            if (maybeReturnOldClientError(error, endTxnRequest.version)) {
```
hachikuji: nit: I am not sure this helper is worthwhile because it only works if the versions are aligned consistently. That is true as of now, but likely untrue in the future. Explicit version checks for each api would be safer.

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: @abbccdda clarified offline that the version has already been bumped for 2.7, so this is not needed.

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
```diff
@@ -381,9 +386,16 @@ class TransactionCoordinator(brokerId: Int,
             if (txnMetadata.producerId != producerId)
               Left(Errors.INVALID_PRODUCER_ID_MAPPING)
             // Strict equality is enforced on the client side requests, as they shouldn't bump the producer epoch.
-            else if ((isFromClient && producerEpoch != txnMetadata.producerEpoch) || producerEpoch < txnMetadata.producerEpoch)
+            else if (isFromClient && producerEpoch != txnMetadata.producerEpoch) {
```
abbccdda: I'm not sure the logic here is actually simplified since we still return `PRODUCER_FENCED` twice. If not, we could still just add the inner if-else to the existing condition ` ((isFromClient && producerEpoch != txnMetadata.producerEpoch) || producerEpoch < txnMetadata.producerEpoch)`

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
```diff
@@ -364,7 +368,8 @@ class TransactionCoordinator(brokerId: Int,
                              producerEpoch: Short,
                              txnMarkerResult: TransactionResult,
                              isFromClient: Boolean,
-                             responseCallback: EndTxnCallback): Unit = {
+                             responseCallback: EndTxnCallback,
+                             timeoutAbort: Boolean = false): Unit = {
```
hachikuji: nit: can we avoid optional fields? it is better for the caller to be explicit

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
```diff
@@ -381,9 +386,16 @@ class TransactionCoordinator(brokerId: Int,
             if (txnMetadata.producerId != producerId)
               Left(Errors.INVALID_PRODUCER_ID_MAPPING)
             // Strict equality is enforced on the client side requests, as they shouldn't bump the producer epoch.
-            else if ((isFromClient && producerEpoch != txnMetadata.producerEpoch) || producerEpoch < txnMetadata.producerEpoch)
+            else if (isFromClient && producerEpoch != txnMetadata.producerEpoch) {
+              if (producerEpoch == txnMetadata.lastProducerEpoch) {
```
hachikuji, line 463: The code assumes that if we get a receive an EndTxn request and `lastProducerEpoch` has been set, then it must be because the coordinator timed out the transaction. That is definitely true in the common case, but I'm wondering if it is worth adding some state to `TransactionMetadata` which explicitly indicates that the transaction was timed out. Not super important and could be done in a follow-up.

core/src/test/scala/integration/kafka/api/TransactionsTest.scala
```diff
@@ -555,11 +555,11 @@ class TransactionsTest extends KafkaServerTestHarness {
...
       // Now that the transaction has expired, the second send should fail with a ProducerFencedException.
       producer.send(TestUtils.producerRecordWithExpectedTransactionStatus(topic1, null, "2", "2", willBeCommitted = false)).get()
-      fail("should have raised a ProducerFencedException since the transaction has expired")
+      fail("should have raised a TransactionTimeOutException since the transaction has expired")
     } catch {
-      case _: ProducerFencedException =>
+      case _: TransactionTimeOutException =>
       case e: ExecutionException =>
-      assertTrue(e.getCause.isInstanceOf[ProducerFencedException])
+      assertTrue(e.getCause.isInstanceOf[TransactionTimeOutException])
```
hachikuji: The main thing we need to test here is that the producer remains usable after the timeout.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -1589,7 +1607,8 @@ public void handleResponse(AbstractResponse response) {
                 fatalError(error.exception());
             } else if (error == Errors.INVALID_TXN_STATE) {
                 fatalError(error.exception());
-            } else if (error == Errors.UNKNOWN_PRODUCER_ID || error == Errors.INVALID_PRODUCER_ID_MAPPING) {
+            } else if (error == Errors.UNKNOWN_PRODUCER_ID || error == Errors.INVALID_PRODUCER_ID_MAPPING
+                    || error == Errors.TRANSACTION_TIMED_OUT) {
```
hachikuji: The documentation suggests that after catching `TransactionTimeoutException`, the user can just begin a new transaction. Unless I'm missing something, however, it seems like we still require an explicit call to `abortTransaction`. That is actually what I prefer, but we should clarify the expectation in the documentation. It is super important for the integration test to cover the full expected flow.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
zhaohaidao: I am not sure this helper is worthwhile because it is only used in beginCompletingTransaction.

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
zhaohaidao: Thanks for your adivce. I have create an issue to track this follow-up: https://issues.apache.org/jira/projects/KAFKA/issues/KAFKA-10596?filter=allissues#

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
zhaohaidao: Thanks for your advice. I will update doc and the kip later

core/src/test/scala/integration/kafka/api/TransactionsTest.scala
zhaohaidao: Thanks for your reminding.
I added a case to verify the producer remains usable after the timeout
https://github.com/apache/kafka/pull/9311/files#diff-231b8aacf24c8458f322b4a9f872d4efR584

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: I see your pt, we could get rid of the helper.

core/src/test/scala/integration/kafka/api/TransactionsTest.scala
```diff
@@ -580,6 +580,52 @@ class TransactionsTest extends KafkaServerTestHarness {
...
+
+    val transactionalConsumer = transactionalConsumers.head
+    transactionalConsumer.subscribe(List(topic1).asJava)
+
+    val transactionalRecords = TestUtils.consumeRecordsFor(transactionalConsumer, 1000)
+    assertTrue(transactionalRecords.isEmpty)
+
+    producer.abortTransaction()
+    producer.beginTransaction()
+    // The second message and hence the second AddPartitions request should be successfully sent.
```
abbccdda: What does this comment suggest?

core/src/test/scala/integration/kafka/api/TransactionsTest.scala
```diff
@@ -580,6 +580,52 @@ class TransactionsTest extends KafkaServerTestHarness {
...
+    // Wait for the expiration cycle to kick in.
+    Thread.sleep(600)
+
+    try {
+      // Now that the transaction has expired, the second send should fail with a ProducerFencedException.
+      producer.commitTransaction()
+      fail("should have raised a TransactionTimeOutException since the transaction has expired")
+    } catch {
+      case _: TransactionTimeoutException =>
+      case e: ExecutionException =>
```
abbccdda: Why would we wrap the txn timeout as an execution exception?

core/src/test/scala/integration/kafka/api/TransactionsTest.scala
```diff
@@ -580,6 +580,52 @@ class TransactionsTest extends KafkaServerTestHarness {
     assertTrue(transactionalRecords.isEmpty)
   }
 
+  @Test
+  def testProducerUsableAfterTxnTimeOutAbortExcetion(): Unit = {
```
abbccdda, line 564: Could we use the same template and reproduce the scenario where user doesn't call abortTxn and immediately resume working?

core/src/test/scala/integration/kafka/api/TransactionsTest.scala
zhaohaidao: This wrap action refers to case testFencingOnAddPartitions, however it doesn’t seem that we need it.

core/src/test/scala/integration/kafka/api/TransactionsTest.scala
zhaohaidao: This comment refers to other cases, which is meaningless. I will remove it.

core/src/test/scala/integration/kafka/api/TransactionsTest.scala
zhaohaidao: ok. The reverse verification is added.

zhaohaidao:
@abbccdda @hachikuji Comments have been addressed, could you continue to review the pr.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
abbccdda: If this is the case, could we just remove this intermediate variable?

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -1198,18 +1209,22 @@ boolean canBumpEpoch() {
...
     private void completeTransaction() {
         if (epochBumpRequired) {
             transitionTo(State.INITIALIZING);
         } else {
             transitionTo(State.READY);
         }
         lastError = null;
+        abortableError = null;
         epochBumpRequired = false;
-        transactionStarted = false;
```
abbccdda, line 1217: Do we miss the update for this flag?

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -1438,7 +1455,8 @@ public void handleResponse(AbstractResponse response) {
                     log.debug("Did not attempt to add partition {} to transaction because other partitions in the " +
                             "batch had errors.", topicPartition);
                     hasPartitionErrors = true;
-                } else if (error == Errors.UNKNOWN_PRODUCER_ID || error == Errors.INVALID_PRODUCER_ID_MAPPING) {
+                } else if (error == Errors.UNKNOWN_PRODUCER_ID || error == Errors.INVALID_PRODUCER_ID_MAPPING
```
abbccdda: nit: we could get a helper for the combined error equality check

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
```diff
@@ -381,9 +388,14 @@ class TransactionCoordinator(brokerId: Int,
             if (txnMetadata.producerId != producerId)
               Left(Errors.INVALID_PRODUCER_ID_MAPPING)
             // Strict equality is enforced on the client side requests, as they shouldn't bump the producer epoch.
-            else if ((isFromClient && producerEpoch != txnMetadata.producerEpoch) || producerEpoch < txnMetadata.producerEpoch)
-              Left(Errors.PRODUCER_FENCED)
-            else if (txnMetadata.pendingTransitionInProgress && txnMetadata.pendingState.get != PrepareEpochFence)
+            else if (isFromClient && producerEpoch != txnMetadata.producerEpoch || producerEpoch < txnMetadata.producerEpoch) {
```
abbccdda: nit: I'm slightly prefer the previous condition format with () around &&.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -671,6 +782,57 @@ class KafkaApisTest {
     }
   }
 
+  @Test
+  def shouldReplaceTxnTimeoutWithInvalidProducerEpochInEndTxnWithOlderClient(): Unit = {
```
abbccdda, line 1496: Could we consolidate the tests with `shouldReplaceProducerFencedWithInvalidProducerEpochInEndTxnWithOlderClient`? Similar to other error code replacement tests.

abbccdda (commented):
Thanks for the updates, I think we are almost there.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
zhaohaidao: Thanks for reminding.

zhaohaidao: If this is the case, could we just remove this intermediate variable?
================================
It could have been omitted, but now this variable has a new use when call initTransactions.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
zhaohaidao: Correct me if I misunderstood your comment.  
Error code replacement tests with TxnTimeout have beed added which is followed.
shouldReplaceTxnTimeoutWithInvalidProducerEpochInAddOffsetToTxnWithOlderClient
shouldReplaceTxnTimeoutWithInvalidProducerEpochInAddPartitionToTxnWithOlderClient
shouldReplaceTxnTimeoutWithInvalidProducerEpochInEndTxnWithOlderClient

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
zhaohaidao: done

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
zhaohaidao: done

abbccdda:
retest this please

abbccdda (commented):
Thanks for the PR @zhaohaidao, it seems that this one has been a bit outdated and I sincerely apologize for the delay. The core change we need to make upfront is bumping the request versions to V4 in order to recognize the new error code. @hachikuji and I will have a plan for getting this merged next week.

arn-aud:
This feature would be very useful and seems almost ready. Any help wanted to complete the task?

PR #9313, opened 2020-09-21
[mm2] Fix consumer/producer properties override

scanterog:
Currently the producer/consumer properties override for the MirrorSourceTask and OffsetSyncStore do not work. This is due the props stored into MirrorConnectorConfig have a `target.cluster` or `source.cluster` prefix. For example:
```
source.cluster.producer.bootstrap.servers -> localhost:9092
target.cluster.consumer.max.poll.interval.ms -> 120000
target.cluster.consumer.auto.offset.reset -> latest
source.cluster.admin.bootstrap.servers -> localhost:9092
```

The [sourceConsumerConfig](https://github.com/apache/kafka/blob/aa0cd667bcd5c4025e84097192030b59165ac9d0/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java#L240) will strip this prefix and store all props as:

```
producer.bootstrap.servers -> localhost:9092
consumer.auto.offset.reset -> latest
consumer.max.poll.interval.ms -> 120000
alias -> A
bootstrap.servers -> localhost:9092
admin.bootstrap.servers -> localhost:9092
consumer.bootstrap.servers -> localhost:9092
```

The next line `props.keySet().retainAll(MirrorClientConfig.CLIENT_CONFIG_DEF.names());` will strip all the props not defined in this common CLIENT_CONFIG_DEF definition. Not relevant but can be confusing if you think the next line will filter on the current values in props.

Finally, the `props.putAll(originalsWithPrefix(CONSUMER_CLIENT_PREFIX));` is based on the on the `originals` variable that have the `target.cluster` or `source.cluster` prefix. There's no single property with the "consumer." prefix. This patterns repeats with the producer config.

This PR also allows to override the hardcoded `auto.offset.reset` value.

### DIFF
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
@@ -199,6 +199,8 @@ public class MirrorConnectorConfig extends AbstractConfig {
 
     protected static final String SOURCE_CLUSTER_PREFIX = MirrorMakerConfig.SOURCE_CLUSTER_PREFIX;
     protected static final String TARGET_CLUSTER_PREFIX = MirrorMakerConfig.TARGET_CLUSTER_PREFIX;
+    protected static final String SOURCE_PREFIX = MirrorMakerConfig.SOURCE_PREFIX;
+    protected static final String TARGET_PREFIX = MirrorMakerConfig.TARGET_PREFIX;
     protected static final String PRODUCER_CLIENT_PREFIX = "producer.";
     protected static final String CONSUMER_CLIENT_PREFIX = "consumer.";
     protected static final String ADMIN_CLIENT_PREFIX = "admin.";
@@ -234,7 +236,7 @@ public class MirrorConnectorConfig extends AbstractConfig {
         props.putAll(originalsWithPrefix(SOURCE_CLUSTER_PREFIX));
         props.keySet().retainAll(MirrorClientConfig.CLIENT_CONFIG_DEF.names());
         props.putAll(originalsWithPrefix(PRODUCER_CLIENT_PREFIX));
-        props.putAll(originalsWithPrefix(SOURCE_CLUSTER_PREFIX + PRODUCER_CLIENT_PREFIX));
+        props.putAll(originalsWithPrefix(SOURCE_PREFIX + PRODUCER_CLIENT_PREFIX));
         return props;
     }
 
@@ -243,7 +245,7 @@ public class MirrorConnectorConfig extends AbstractConfig {
         props.putAll(originalsWithPrefix(SOURCE_CLUSTER_PREFIX));
         props.keySet().retainAll(MirrorClientConfig.CLIENT_CONFIG_DEF.names());
         props.putAll(originalsWithPrefix(CONSUMER_CLIENT_PREFIX));
-        props.putAll(originalsWithPrefix(SOURCE_CLUSTER_PREFIX + CONSUMER_CLIENT_PREFIX));
+        props.putAll(originalsWithPrefix(SOURCE_PREFIX + CONSUMER_CLIENT_PREFIX));
         props.put(ENABLE_AUTO_COMMIT_CONFIG, "false");
         props.putIfAbsent(AUTO_OFFSET_RESET_CONFIG, "earliest");
         return props;
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
@@ -199,8 +199,8 @@ public class MirrorConnectorConfig extends AbstractConfig {
 
     protected static final String SOURCE_CLUSTER_PREFIX = MirrorMakerConfig.SOURCE_CLUSTER_PREFIX;
     protected static final String TARGET_CLUSTER_PREFIX = MirrorMakerConfig.TARGET_CLUSTER_PREFIX;
-    protected static final String PRODUCER_CLIENT_PREFIX = "producer.";
-    protected static final String CONSUMER_CLIENT_PREFIX = "consumer.";
+    protected static final String PRODUCER_CLIENT_PREFIX = SOURCE_CLUSTER_PREFIX + "producer.";
+    protected static final String CONSUMER_CLIENT_PREFIX = SOURCE_CLUSTER_PREFIX + "consumer.";
     protected static final String ADMIN_CLIENT_PREFIX = "admin.";
     protected static final String SOURCE_ADMIN_CLIENT_PREFIX = "source.admin.";
     protected static final String TARGET_ADMIN_CLIENT_PREFIX = "target.admin.";
@@ -243,7 +243,7 @@ public class MirrorConnectorConfig extends AbstractConfig {
         props.keySet().retainAll(MirrorClientConfig.CLIENT_CONFIG_DEF.names());
         props.putAll(originalsWithPrefix(CONSUMER_CLIENT_PREFIX));
         props.put(ENABLE_AUTO_COMMIT_CONFIG, "false");
-        props.put(AUTO_OFFSET_RESET_CONFIG, "earliest");
+        props.putIfAbsent(AUTO_OFFSET_RESET_CONFIG, "earliest");
         return props;
     }
 
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
@@ -204,8 +204,6 @@ public class MirrorConnectorConfig extends AbstractConfig {
     protected static final String PRODUCER_CLIENT_PREFIX = "producer.";
     protected static final String CONSUMER_CLIENT_PREFIX = "consumer.";
     protected static final String ADMIN_CLIENT_PREFIX = "admin.";
-    protected static final String SOURCE_ADMIN_CLIENT_PREFIX = "source.admin.";
-    protected static final String TARGET_ADMIN_CLIENT_PREFIX = "target.admin.";
 
     public MirrorConnectorConfig(Map<String, String> props) {
         this(CONNECTOR_CONFIG_DEF, props);
@@ -271,7 +269,7 @@ public class MirrorConnectorConfig extends AbstractConfig {
         props.putAll(originalsWithPrefix(TARGET_CLUSTER_PREFIX));
         props.keySet().retainAll(MirrorClientConfig.CLIENT_CONFIG_DEF.names());
         props.putAll(originalsWithPrefix(ADMIN_CLIENT_PREFIX));
-        props.putAll(originalsWithPrefix(TARGET_ADMIN_CLIENT_PREFIX));
+        props.putAll(originalsWithPrefix(TARGET_PREFIX + ADMIN_CLIENT_PREFIX));
         return props;
     }
 
@@ -280,7 +278,7 @@ public class MirrorConnectorConfig extends AbstractConfig {
         props.putAll(originalsWithPrefix(SOURCE_CLUSTER_PREFIX));
         props.keySet().retainAll(MirrorClientConfig.CLIENT_CONFIG_DEF.names());
         props.putAll(originalsWithPrefix(ADMIN_CLIENT_PREFIX));
-        props.putAll(originalsWithPrefix(SOURCE_ADMIN_CLIENT_PREFIX));
+        props.putAll(originalsWithPrefix(SOURCE_PREFIX + ADMIN_CLIENT_PREFIX));
         return props;
     }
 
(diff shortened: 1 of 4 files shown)


scanterog:
Hello @ryannedolan @mimaison. Your review will be appreciated. I think the test failure is unrelated. Thanks!

ryannedolan (approved):
lgtm. This bug has come up a few times, and I know some people have patched MM2 to get around it. I'd consider it a bug fix, so probably doesn't need a KIP or anything. Thanks!

scanterog:
Thanks @ryannedolan! what is the usual procedure to get this merged?

ryannedolan:
> Thanks @ryannedolan! what is the usual procedure to get this merged?

A committer must approve and merge.

scanterog:
@hachikuji is there any chance you can review this? Thanks!

mimaison:
@scanterog Thanks for the PR. Can you add a test for the changes?

scanterog:
@mimaison just added it. Let me know if there's anything else to fix. Thanks!

connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
```diff
@@ -199,8 +199,8 @@
 
     protected static final String SOURCE_CLUSTER_PREFIX = MirrorMakerConfig.SOURCE_CLUSTER_PREFIX;
     protected static final String TARGET_CLUSTER_PREFIX = MirrorMakerConfig.TARGET_CLUSTER_PREFIX;
-    protected static final String PRODUCER_CLIENT_PREFIX = "producer.";
-    protected static final String CONSUMER_CLIENT_PREFIX = "consumer.";
+    protected static final String PRODUCER_CLIENT_PREFIX = SOURCE_CLUSTER_PREFIX + "producer.";
+    protected static final String CONSUMER_CLIENT_PREFIX = SOURCE_CLUSTER_PREFIX + "consumer.";
```
mimaison: These changes are going to break existing users. For example, I have connectors with a few settings prefixed with `consumer.`. I wonder if we could keep the old behaviour (even if partially broken) while adding the proper prefixes

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorConfigTest.java
```diff
@@ -119,4 +120,41 @@ public void testNonMutationOfConfigDef() {
...
+        Map<String, String> connectorProps = makeProps(
+                MirrorConnectorConfig.PRODUCER_CLIENT_PREFIX + "acks", "1"
+        );
+        MirrorConnectorConfig config = new MirrorConnectorConfig(connectorProps);
+        Map<String, Object> connectorProducerProps = config.sourceProducerConfig();
+        Map<String, Object> expectedProducerProps = new HashMap<>();
+        expectedProducerProps.put("acks", "1");
+        assertEquals(expectedProducerProps, connectorProducerProps);
+    }
+
```
mimaison, line 175: While we're at it, could we also add tests for `targetAdminConfig()` and `sourceAdminConfig()`?

mimaison (changes requested):

connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
scanterog: I'm not sure if I follow it. ~This prefix "source_cluster" is not user specified. It is prefixed somewhere else by mm2 (I can get the lines if you want later on).~ The config specified [here](https://github.com/apache/kafka/tree/trunk/connect/mirror#producer--consumer--admin-config-used-by-mm2) will actually be honored. It does not imply changes on the MirrorMaker config side at least. Are you talking about the case of running MirrorMaker in a connect cluster rather than as a dedicated cluster? I didn't check that one TBH. Maybe @ryannedolan has more insight on this one?

mimaison: Yes I'm running MM2 in a Connect cluster.

scanterog: Gotcha. I just peek'd at it again and you're right. It seems the proper fix would be to fix the way mm2 populates the MirrorConnectorConfig to avoid this change. WDYT @ryannedolan ? Not sure how involved this change would be.

scanterog: I have put a new fix that should handle both cases. Details on why I think this is a reasonable fix:
* There's no change on what is defined [here](https://github.com/apache/kafka/tree/trunk/connect/mirror#producer--consumer--admin-config-used-by-mm2). The properties that starts with `producer.` or `consumer.` are still ignored in mm2 dedicated mode.
* MM2 needs to differentiate the properties for the source cluster and we can't just remove the `source.cluster` prefix.
* For running the mirror connectors on a Connect cluster, the only "drawback" is that now you can provide an override with `source.cluster.consumer` or `source.cluster.producer`.

mimaison: It's unfortunate for admin we use `source.admin` as the prefix ...

So we'd be left with a configuration like:
```
"source.cluster.bootstrap.servers": "localhost:9092",
"source.cluster.security.protocol": "SASL_SSL",
"source.cluster.producer.some-producer-setting": 123,
"source.cluster.consumer.some-consumer-setting": 123,
"source.admin.some-admin-setting": 123
```

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorConfigTest.java
```diff
@@ -119,4 +120,41 @@ public void testNonMutationOfConfigDef() {
             connectorConfigDef.names().contains(taskSpecificProperty)
         ));
     }
+
+    @Test
+    public void testSourceConsumerConfig() {
```
mimaison, line 125: We should test settings with the `source.cluster.` prefix too. Same in the test below

mimaison (changes requested):

connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
mimaison: Maybe not including `cluster` for all the client specific settings would be slightly better, WDYT?
```
"source.cluster.bootstrap.servers": "localhost:9092",
"source.cluster.security.protocol": "SASL_SSL",
"source.producer.some-producer-setting": 123,
"source.consumer.some-consumer-setting": 123,
"source.admin.some-admin-setting": 123
```

scanterog: For connect, you can still use (or the one you've shared).
```
"source.cluster.bootstrap.servers": "localhost:9092",
"source.cluster.security.protocol": "SASL_SSL",
"producer.some-producer-setting": 123
"consumer.some-consumer-setting": 123
"source.admin.some-admin-setting": 123
```

Yes, we could normalize that but I'm not touching admin config in this PR.

scanterog: > Maybe not including cluster for all the client specific settings would be slightly better, WDYT?

Somehow I missed this. I do not have a strong opinion on that but I do like the idea on normalizing it. But changing that will still break existing users though. How do you want to proceed?

scanterog: @mimaison just a friendly ping.

mimaison: I don't think the format mentioned in https://github.com/apache/kafka/pull/9313#discussion_r498298987 would break compatibility.

scanterog: For users running on a Connect cluster, that format is not supported right now:
```
source.producer.some-producer-setting: 123
```

The only supported is (unless I'm missing something):
```
"producer.some-producer-setting": 123
```

mimaison: I meant add support for that format. We obviously want to keep supporting the existing formats

scanterog: 🤦 got it. I misunderstood it. Let me give it a try.

scanterog: I think the new code does what we want. Also tested a bin from this on my infra and it is working fine. Let me know if this works for you.

mimaison:
@scanterog 
`testClusterConfigProperties()` seems to be failing now. Can you fix it before I take a look at the PR?

```
org.apache.kafka.connect.mirror.MirrorMakerConfigTest > testClusterConfigProperties FAILED
    java.lang.AssertionError: source.cluster.bootstrap.servers is set expected:<servers-one> but was:<null>
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:120)
        at org.apache.kafka.connect.mirror.MirrorMakerConfigTest.testClusterConfigProperties(MirrorMakerConfigTest.java:55)
```

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorMakerConfigTest.java
```diff
@@ -52,10 +52,10 @@ public void testClusterConfigProperties() {
             "replication.factor", "4"));
         Map<String, String> connectorProps = mirrorConfig.connectorBaseConfig(new SourceAndTarget("a", "b"),
             MirrorSourceConnector.class);
-        assertEquals("source.cluster.bootstrap.servers is set",
-            "servers-one", connectorProps.get("source.cluster.bootstrap.servers"));
-        assertEquals("target.cluster.bootstrap.servers is set",
-            "servers-two", connectorProps.get("target.cluster.bootstrap.servers"));
+        assertEquals("source.bootstrap.servers is set",
```
mimaison: I was not expecting to see changes in existing tests if we are not breaking compatibility, `source.cluster.bootstrap.servers` should continue working and still be tested

scanterog: That will require more changes. This does not break for the mirrorMaker driver main config. I guess it does for on Connect though

scanterog: I think I have achieved what we want. Explicitly setting "source." prefix for props starting with consumer|producer|admin and setting "source.cluster." otherwise. All tests passed. The code runs fine on my infra.

@mimaison please take a look whenever you get some time :)

connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
```diff
@@ -245,10 +254,17 @@ public MirrorClientConfig clientConfig(String cluster) {
         Map<String, String> strings = originalsStrings();
         strings.keySet().removeIf(x -> !x.startsWith(prefix));
         return strings;
-    } 
+    }
+
+    static Map<String, String> prefixForGeneralAttrs(String prefix, Map<String, String> props) {
```
mimaison: Can we find a better name for these 2 new methods? Something like `clusterConfigsWithPrefix()` and `clientConfigsWithPrefix()`?

connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorConnectorConfig.java
```diff
@@ -199,6 +199,8 @@
 
     protected static final String SOURCE_CLUSTER_PREFIX = MirrorMakerConfig.SOURCE_CLUSTER_PREFIX;
     protected static final String TARGET_CLUSTER_PREFIX = MirrorMakerConfig.TARGET_CLUSTER_PREFIX;
+    protected static final String SOURCE_PREFIX = MirrorMakerConfig.SOURCE_PREFIX;
+    protected static final String TARGET_PREFIX = MirrorMakerConfig.TARGET_PREFIX;
     protected static final String PRODUCER_CLIENT_PREFIX = "producer.";
     protected static final String CONSUMER_CLIENT_PREFIX = "consumer.";
     protected static final String ADMIN_CLIENT_PREFIX = "admin.";
```
mimaison, line 206: We could get rid of `SOURCE_ADMIN_CLIENT_PREFIX` and `TARGET_ADMIN_CLIENT_PREFIX` and instead use `SOURCE_PREFIX + ADMIN_CLIENT_PREFIX` or `TARGET_PREFIX + ADMIN_CLIENT_PREFIX` in the method getting Admin Client configs so they are the similar to the Producer and Consumer methods

mimaison (changes requested):
Thanks @scanterog, the changes look good now. I've left a couple of very minor comments that would be nice to address. I'll be happy to merge then

connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
scanterog: Indeed. Thanks!

scanterog:
> Thanks @scanterog, the changes look good now. I've left a couple of very minor comments that would be nice to address. I'll be happy to merge then

Thanks. There's one test failing for JDK15 apparently (need to check if it is legit). Works locally with JDK11. I wish I could rerun those tests. Many flakeys ones.

mimaison:
@scanterog I've rekicked the build

scanterog:
> @scanterog I've rekicked the build

I think this is ready to merge :P

mimaison (approved):
LGTM, thanks for the PR

mimaison:
Test failures are not related, merging to trunk

PR #9314, opened 2020-09-21
KAFKA-8360: Docs do not mention RequestQueueSize JMX metric

ankitk-me:
What? :: Mentioning "Request Queue Size" under the Monitoring tab. RequestQueueSize is an important metric to monitor the number of requests in the queue. As a crowded queue might face issue processing incoming or outgoing requests

@viktorsomogyi  can you please review this?

Thanks!!


viktorsomogyi (approved):

viktorsomogyi (changes requested):
Just noticed that the metric actually inserted between two others that are somewhat related to each other (message conversion time and rate). Would you please move this before or after them? And sorry for noticing it just now.

PR #9315, opened 2020-09-21
KAFKA-10496: Removed relying on external DNS servers in tests

piotrrzysko:
As ticket suggested I’ve tried to introduce an in-memory DNS server for testing purposes, but unfortunately capability to change default DNS provider has been removed in Java 9: https://bugs.openjdk.java.net/browse/JDK-8134577.  There is a request for restoring it: https://bugs.openjdk.java.net/browse/JDK-8192780, but so far has not been implemented.

Therefore this PR decouples DNS resolution from `InetAddress.getAllByName` implementation in order to mock its behavior in tests. 

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/ClientUtils.java
+++ b/clients/src/main/java/org/apache/kafka/clients/ClientUtils.java
@@ -59,7 +59,7 @@ public final class ClientUtils {
                         throw new ConfigException("Invalid url in " + CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG + ": " + url);
 
                     if (clientDnsLookup == ClientDnsLookup.RESOLVE_CANONICAL_BOOTSTRAP_SERVERS_ONLY) {
-                        InetAddress[] inetAddresses = InetAddress.getAllByName(host);
+                        InetAddress[] inetAddresses = DefaultDnsNameResolver.DEFAULT_DNS_NAME_RESOLVER.resolve(host);
                         for (InetAddress inetAddress : inetAddresses) {
                             String resolvedCanonicalName = inetAddress.getCanonicalHostName();
                             InetSocketAddress address = new InetSocketAddress(resolvedCanonicalName, port);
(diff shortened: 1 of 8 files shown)


piotrrzysko:
@jolshan @mumrah @dajac Could you please take look at this PR?

jolshan (commented):
Looks pretty good to me! Glad we got testMultipleIPsWithUseAll() to continue checking the round robin behavior of the connectionStates.currentAddress

It wouldn't let me comment on the line of this test, but do we want to update testParseAndValidateAddresses() and testParseAndValidateAddressesWithReverseLookup() in ClientUtilsTest? It seems that the addresses used are ones reserved for documentation purposes, so they are probably less fragile, but maybe we want all the tests to use the mock DNS.

mumrah (commented):
Thanks for the patch, @piotrrzysko! I like the decoupling here, makes testing a lot simpler and more clear. 

Should we let callers of `ClientUtils#parseAndValidateAddresses` provide the `DnsNameResolver` implementation as well? It might be nice to somehow combine the ClientDnsLookup with the resolver, but that might complicate things a bit.

piotrrzysko:
@jolshan @mumrah Thanks for the review! Of course, I agree with you that it would be nice to use the mock in all tests, but after rethinking the solution suggested in this PR I came to the conclusion that it is not so easy.

I assumed that `InetAddress#getAllByName` is the only place where DNS resolution happens - it’s not true. The constructor of `InetSocketAddress` that is used by `ClientUtils#parseAndValidateAddresses` in some circumstances (when a passed hostname is a machine name rather than a literal IP address) may also call an external DNS server. This behavior is much harder to mock. Perhaps it is not a problem for tests from `ClientUtilsTest`, because as I checked only `localhost` is resolved in this way. However, I wouldn’t like to introduce abstraction over DNS resolving mechanism that is unable to cover all cases.

Unfortunately, the only solution that comes to my mind is to try a hack based on reflection that will allow replacing the default DNS provider in tests. It was something that I wanted to avoid because it can cause problems with future versions of Java. Nevertheless, I can check if it is possible and how ugly it will be :). 

WDYT?

mumrah:
Hi @piotrrzysko 👋. Earlier this year, https://github.com/apache/kafka/commit/131d4753cfed65ed6dee0a8c754765c97c3d513f was committed which added an interface very similar to what you have here along with other host resolution fixes. 

Looking over this PR, it seems like your ClientUtilsTest fixes could remove some potential flaky tests. Could you rebase this PR with trunk?

ijuma:
Fyi, Java 18+ do have a SPI for this:
https://bugs.openjdk.org/browse/JDK-8263693

PR #9317, opened 2020-09-22
KAFKA-10509: Added throttle connection accept rate metric (KIP-612)

apovzner:
This PR adds metric to track throttle time of connection accept rate (when hitting connection attempt rate quota), which is a part of KIP-612:
> kafka.network:type=socket-server-metrics,name=connection-accept-throttle-time,listener={listenerName}
> Type: SampledStat.Avg
> Description: Average throttle time due to violating per-listener or broker-wide connection acceptance rate quota on a given listener.

This PR also does 2 fixes:
* Ensures that per-listener connection quotas are configured on broker startup (from static broker config, if one is set). 
* Reduces quota values used in testDynamicListenerConnectionCreationRateQuota test, and the duration of `verifyConnectionRate` (that creates connections and verifies rate). We observed once that this test exhausted ephemeral port count. I reduced quotas used in this test by half, since this does not change the correctness of the test, while it also reduces the number of connections made by this test.

### DIFF
--- a/core/src/main/scala/kafka/network/SocketServer.scala
+++ b/core/src/main/scala/kafka/network/SocketServer.scala
@@ -1263,7 +1263,7 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
         listenerCounts.remove(listenerName)
         // once listener is removed from maxConnectionsPerListener, no metrics will be recorded into listener's sensor
         // so it is safe to remove sensor here
-        listenerQuota.removeSensors()
+        listenerQuota.close()
         counts.notifyAll() // wake up any waiting acceptors to close cleanly
         config.removeReconfigurable(listenerQuota)
       }
@@ -1303,12 +1303,6 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
     counts.synchronized {
       val startThrottleTimeMs = time.milliseconds
       val throttleTimeMs = math.max(recordConnectionAndGetThrottleTimeMs(listenerName, startThrottleTimeMs), 0)
-      if (throttleTimeMs > 0) {
-        // record throttle time due to hitting connection rate limit
-        // connection could be throttled longer if the limit of the number of active connections is reached as well
-        maxConnectionsPerListener.get(listenerName)
-          .foreach(_.connectionRateThrottleSensor.record(throttleTimeMs.toDouble, startThrottleTimeMs))
-      }
 
       if (throttleTimeMs > 0 || !connectionSlotAvailable(listenerName)) {
         val startNs = time.nanoseconds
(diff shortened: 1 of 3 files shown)


core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1292,6 +1292,12 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
     counts.synchronized {
       val startThrottleTimeMs = time.milliseconds
       val throttleTimeMs = math.max(recordConnectionAndGetThrottleTimeMs(listenerName, startThrottleTimeMs), 0)
+      if (throttleTimeMs > 0) {
+        // record throttle time due to hitting connection rate limit
+        // connection could be throttled longer if the limit of the number of active connections is reached as well
+        maxConnectionsPerListener.get(listenerName)
+          .foreach(_.connectionRateThrottleSensor.record(throttleTimeMs.toDouble, startThrottleTimeMs))
```
dajac: It is a bit annoying that we have to lookup the `ListenerConnectionQuota` twice. Once in `recordConnectionAndGetThrottleTimeMs` and once here. I wonder if we could look it up once to record the rate and to record the throttle time.

I am thinking about the following to be more concrete. We could add two methods in `ListenerConnectionQuota`: 1) `record` and 2) `recordThrottleTime`. They would encapsulate the logic to respectively record the rate (for the listener and the broker like we do now) and the throttle time. So here we could look up the `ListenerConnectionQuota` record and get the throttle time, and record the throttle time if > 0.

That could improve the readability a bit. I am not sure that it would make a difference from a performance point of view. WDYT?

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1414,7 +1420,8 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
 
   class ListenerConnectionQuota(lock: Object, listener: ListenerName) extends ListenerReconfigurable {
     @volatile private var _maxConnections = Int.MaxValue
-    val connectionRateSensor = createConnectionRateQuotaSensor(Int.MaxValue, Some(listener.value))
+    val connectionRateSensor: Sensor = createConnectionRateQuotaSensor(Int.MaxValue, Some(listener.value))
+    val connectionRateThrottleSensor: Sensor = createConnectionRateThrottleSensor()
```
dajac: nit: Do we really need to provide the type here? It seems that we usually don't provide it in that file.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1447,13 +1454,33 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
       }
     }
 
+    def removeSensors(): Unit = {
```
dajac: nit: What about naming this `close`? It feels a bit more natural to call `close` when we don't need the object anymore and it would be more aligned with the `close` methods that we already have in this file. For instance, `ConnectionQuotas#close` which also cleanup the metrics.

BTW, not related to this PR but it seems that we don't close the `ListenerConnectionQuota` when the `ConnectionQuotas` is closed. I suppose that we leave metrics around. Is it the case?

dajac (commented):
@apovzner Thanks for the PR. I have left few comments and suggestions.

core/src/main/scala/kafka/network/SocketServer.scala
apovzner: I added type because style check suggested (and also checked style guide) to annotate public members. However, I will change this to package scope, since we only need an outer class to access them. So, it was useful that you asked.

core/src/main/scala/kafka/network/SocketServer.scala
apovzner: agreed, `close` will be consistent. 

About not closing ListenerConnectionQuota on ConnectionQuotas.close(): this is on broker shutdown path where KafkaServer calls metrics.close() on shutdown as well (and metrics are owned by KafkaServer), so I don't think we are leaking anything. But I think it's better to remove listener sensors on ConnectionQuotas.close() anyways, so I will make that change.

core/src/main/scala/kafka/network/SocketServer.scala
apovzner: That code seemed readable to me, but perhaps looking up `ListenerConnectionQuota` is better. I made a change to lookup once, please take a look.

apovzner:
@dajac Thanks for the review, I addressed your comments.

apovzner:
PR Build failed due to know flaky tests:
https://issues.apache.org/jira/browse/KAFKA-8266: kafka.api.ConsumerBounceTest.testRollingBrokerRestartsWithSmallerMaxGroupSizeConfigDisruptsBigGroup
https://issues.apache.org/jira/browse/KAFKA-8460: kafka.api.PlaintextConsumerTest.testLowMaxFetchSizeForRequestAndPartition

dajac (approved):
@apovzner Thanks for the update. LGTM.

rajinisivaram (approved):
@apovzner Thanks for the PR, LGTM. Merging to trunk.

PR #9319, opened 2020-09-22
KAFKA-10413: Allow for even distribution of lost/new tasks when multiple Connect workers join at the same time

ramesh-muthusamy:
Allow even distribution of lost/new tasks when more than one worker joins the group at the same time

Issue description:
Existing issue 1 description : When more than one worker joins the consumer group the incremental co operative assignor revokes and re assigns atmost average number of tasks per worker.

Issue: This results in the additional workers joining the group stay idle and would require more future rebalances to happen to have even distribution of tasks.

Fix: As part of task assignment calculation following a deployment, the reassignment of tasks are calculated by revoking all the tasks above ceil(average) number of tasks.

Existing issue 2 description: When more than one worker is lost and rejoins the group at most one worker will be re assigned with the lost tasks from all the workers that left the group.

Issue: In scenarios where more than one worker is lost and rejoins the group only one among them gets assigned all the partitions that were lost in the past. The additional workers that have joined would not get any task assigned to them until a rebalance that happens in future.

Fix: As part fo lost task re assignment all the new workers that have joined the group would be considered for task assignment and would be assigned in a round robin fashion with the new tasks.

Testing strategy : System testing in a Kube environment completed.
UT : updated to UT

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
@@ -571,7 +571,7 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
         log.debug("Previous rounded down (floor) average number of connectors per worker {}", totalActiveConnectorsNum / existingWorkersNum);
         int floorConnectors = totalActiveConnectorsNum / totalWorkersNum;
         int ceilConnectors = (int) Math.ceil((float) totalActiveConnectorsNum / totalWorkersNum);
-        log.debug("New rounded down (floor) average number of connectors per worker floor connectors {} ciel connectors ", floorConnectors,ceilConnectors);
+        log.debug("New rounded down (floor) average number of connectors per worker floor connectors {} ciel connectors ", floorConnectors, ceilConnectors);
 
 
         log.debug("Previous rounded down (floor) average number of tasks per worker {}", totalActiveTasksNum / existingWorkersNum);
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
@@ -471,7 +471,6 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
                     WorkerLoad worker = candidateWorkerIterator.next();
                     log.debug("Assigning task id {} to member {}", task, worker.worker());
                     worker.assign(task);
-                    log.debug("Assigned task id {} to member {}", task, worker.worker());
                 }
             } else {
                 log.debug("No single candidate worker was found to assign lost tasks. Treating lost tasks as new tasks");
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
@@ -577,9 +577,8 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
 
         log.debug("Previous rounded down (floor) average number of tasks per worker {}", totalActiveTasksNum / existingWorkersNum);
         int floorTasks = totalActiveTasksNum / totalWorkersNum;
-        log.debug("New rounded down (floor) average number of tasks per worker {}", floorTasks);
         int ceilTasks = (int) Math.ceil((float) totalActiveTasksNum / totalWorkersNum);
-        log.debug("New rounded down (ceil) average number of tasks per worker {}", ceilTasks);
+        log.debug("New average number of tasks per worker: floor={}, ceiling={}", floorTasks, ceilTasks);
 
         int numToRevoke = floorConnectors;
         for (WorkerLoad existing : existingWorkers) {
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
@@ -450,7 +450,9 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
             }
 
             if (!candidateWorkerLoad.isEmpty()) {
-                log.debug("A list of candidate workers has been found to assign lost tasks: {}", candidateWorkerLoad.stream().map(WorkerLoad::worker).collect(Collectors.joining(",")));
+                log.debug("Assigning lost tasks to {} candidate workers: {}", 
+                        candidateWorkerLoad.size(),
+                        candidateWorkerLoad.stream().map(WorkerLoad::worker).collect(Collectors.joining(",")));
                 Iterator<WorkerLoad> candidateWorkerIterator = candidateWorkerLoad.iterator();
                 for (String connector : lostAssignments.connectors()) {
                     // Loop over the the candidate workers as many times as it takes
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
@@ -259,7 +259,7 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
         // Do not revoke resources for re-assignment while a delayed rebalance is active
         // Also we do not revoke in two consecutive rebalances by the same leader
         canRevoke = delay == 0 && canRevoke;
-        log.debug("Connector and task to revoke assgn post lb calculation: {}", toRevoke);
+        log.debug("Connector and task to revoke assignment post load balancer calculation: {}", toRevoke);
         // Compute the connectors-and-tasks to be revoked for load balancing without taking into
         // account the deleted ones.
         log.debug("Can leader revoke tasks in this assignment? {} (delay: {})", canRevoke, delay);
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
@@ -597,7 +597,7 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
         for (WorkerLoad existing : existingWorkers) {
             Iterator<ConnectorTaskId> tasks = existing.tasks().iterator();
             numToRevoke = existing.tasksSize() - ceilTasks;
-            log.debug("revoke number of tasks per worker {}", numToRevoke);
+            log.debug("Tasks on worker {} is higher than ceiling, so revoking {} tasks", existing, numToRevoke);
             for (int i = existing.tasksSize(); i > floorTasks && numToRevoke > 0; --i, --numToRevoke) {
                 ConnectorsAndTasks resources = revoking.computeIfAbsent(
                     existing.worker(),
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
@@ -462,7 +462,6 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
                     WorkerLoad worker = candidateWorkerIterator.next();
                     log.debug("Assigning connector id {} to member {}", connector, worker.worker());
                     worker.assign(connector);
-                    log.debug("Assigned connector id {} to member {}", connector, worker.worker());
                 }
                 candidateWorkerIterator = candidateWorkerLoad.iterator();
                 for (ConnectorTaskId task : lostAssignments.tasks()) {
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
@@ -259,7 +259,6 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
         // Do not revoke resources for re-assignment while a delayed rebalance is active
         // Also we do not revoke in two consecutive rebalances by the same leader
         canRevoke = delay == 0 && canRevoke;
-        log.debug("Connector and task to revoke assignment post load balancer calculation: {}", toRevoke);
         // Compute the connectors-and-tasks to be revoked for load balancing without taking into
         // account the deleted ones.
         log.debug("Can leader revoke tasks in this assignment? {} (delay: {})", canRevoke, delay);
@@ -571,16 +570,19 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
         // We have at least one worker assignment (the leader itself) so totalWorkersNum can't be 0
         log.debug("Previous rounded down (floor) average number of connectors per worker {}", totalActiveConnectorsNum / existingWorkersNum);
         int floorConnectors = totalActiveConnectorsNum / totalWorkersNum;
-        log.debug("New rounded down (floor) average number of connectors per worker {}", floorConnectors);
+        int ceilConnectors = (int) Math.ceil((float) totalActiveConnectorsNum / totalWorkersNum);
+        log.debug("New rounded down (floor) average number of connectors per worker floor connectors {} ciel connectors ", floorConnectors,ceilConnectors);
+
 
         log.debug("Previous rounded down (floor) average number of tasks per worker {}", totalActiveTasksNum / existingWorkersNum);
         int floorTasks = totalActiveTasksNum / totalWorkersNum;
         int ceilTasks = (int) Math.ceil((float) totalActiveTasksNum / totalWorkersNum);
         log.debug("New average number of tasks per worker: floor={}, ceiling={}", floorTasks, ceilTasks);
+        int numToRevoke;
 
-        int numToRevoke = floorConnectors;
         for (WorkerLoad existing : existingWorkers) {
             Iterator<String> connectors = existing.connectors().iterator();
+            numToRevoke = existing.connectorsSize() - ceilConnectors;
             for (int i = existing.connectorsSize(); i > floorConnectors && numToRevoke > 0; --i, --numToRevoke) {
                 ConnectorsAndTasks resources = revoking.computeIfAbsent(
                     existing.worker(),
@@ -592,7 +594,6 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
             }
         }
 
-        numToRevoke = floorTasks;
         for (WorkerLoad existing : existingWorkers) {
             Iterator<ConnectorTaskId> tasks = existing.tasks().iterator();
             numToRevoke = existing.tasksSize() - ceilTasks;
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
@@ -260,7 +260,7 @@ public class IncrementalCooperativeAssignor implements ConnectAssignor {
         // Do not revoke resources for re-assignment while a delayed rebalance is active
         // Also we do not revoke in two consecutive rebalances by the same leader
         canRevoke = delay == 0 && canRevoke;
-
+        log.debug("Connector and task to revoke assgn post lb calculation: {}", toRevoke);
         // Compute the connectors-and-tasks to be revoked for load balancing without taking into
         // account the deleted ones.
         log.debug("Can leader revoke tasks in this assignment? {} (delay: {})", canRevoke, delay);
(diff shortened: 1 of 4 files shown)


ramesh-muthusamy:
@kkonstantine  can you help reviewing this PR

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -260,7 +259,7 @@ private Long ensureLeaderConfig(long maxOffset, WorkerCoordinator coordinator) {
         // Do not revoke resources for re-assignment while a delayed rebalance is active
         // Also we do not revoke in two consecutive rebalances by the same leader
         canRevoke = delay == 0 && canRevoke;
-
+        log.debug("Connector and task to revoke assgn post lb calculation: {}", toRevoke);
```
rhauch: Can you please make this log message more readable?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -445,16 +444,34 @@ protected void handleLostAssignments(ConnectorsAndTasks lostAssignments,
...
                 candidateWorkerLoad = pickCandidateWorkerForReassignment(completeWorkerAssignment);
             }
 
-            if (candidateWorkerLoad.isPresent()) {
-                WorkerLoad workerLoad = candidateWorkerLoad.get();
-                log.debug("A candidate worker has been found to assign lost tasks: {}", workerLoad.worker());
-                lostAssignments.connectors().forEach(workerLoad::assign);
-                lostAssignments.tasks().forEach(workerLoad::assign);
+            if (!candidateWorkerLoad.isEmpty()) {
+                log.debug("A list of candidate workers has been found to assign lost tasks: {}", candidateWorkerLoad.stream().map(WorkerLoad::worker).collect(Collectors.joining(",")));
```
rhauch:
```suggestion
                log.debug("Assigning lost tasks to {} candidate workers: {}", 
                        candidateWorkerLoad.size(),
                        candidateWorkerLoad.stream().map(WorkerLoad::worker).collect(Collectors.joining(",")));
```

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -445,16 +444,34 @@ protected void handleLostAssignments(ConnectorsAndTasks lostAssignments,
...
+                Iterator<WorkerLoad> candidateWorkerIterator = candidateWorkerLoad.iterator();
+                for (String connector : lostAssignments.connectors()) {
+                    // Loop over the the candidate workers as many times as it takes
+                    if (!candidateWorkerIterator.hasNext()) {
+                        candidateWorkerIterator = candidateWorkerLoad.iterator();
+                    }
+                    WorkerLoad worker = candidateWorkerIterator.next();
+                    log.debug("Assigning connector id {} to member {}", connector, worker.worker());
+                    worker.assign(connector);
+                    log.debug("Assigned connector id {} to member {}", connector, worker.worker());
```
rhauch:
```suggestion
```
Do we need both of these debug messages? After all, `worker.assign(...)` is just adding a string to a collection. How about keeping the first one since this is at this point an on-going process and we've not actually assigned anything to the actual worker node.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -445,16 +444,34 @@ protected void handleLostAssignments(ConnectorsAndTasks lostAssignments,
...
+                }
+                candidateWorkerIterator = candidateWorkerLoad.iterator();
+                for (ConnectorTaskId task : lostAssignments.tasks()) {
+                    if (!candidateWorkerIterator.hasNext()) {
+                        candidateWorkerIterator = candidateWorkerLoad.iterator();
+                    }
+                    WorkerLoad worker = candidateWorkerIterator.next();
+                    log.debug("Assigning task id {} to member {}", task, worker.worker());
+                    worker.assign(task);
+                    log.debug("Assigned task id {} to member {}", task, worker.worker());
```
rhauch:
```suggestion
```
Do we need both of these debug messages? After all, `worker.assign(...)` is just adding a `ConnectorTaskId` to a collection. How about keeping the first one since this is at this point an on-going process and we've not actually assigned anything to the actual worker node.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -559,6 +576,8 @@ private void resetDelay() {
         log.debug("Previous rounded down (floor) average number of tasks per worker {}", totalActiveTasksNum / existingWorkersNum);
         int floorTasks = totalActiveTasksNum / totalWorkersNum;
         log.debug("New rounded down (floor) average number of tasks per worker {}", floorTasks);
+        int ceilTasks = (int) Math.ceil((float) totalActiveTasksNum / totalWorkersNum);
+        log.debug("New rounded down (ceil) average number of tasks per worker {}", ceilTasks);
```
rhauch: What do you think about combining these log messages?
```suggestion
        int ceilTasks = (int) Math.ceil((float) totalActiveTasksNum / totalWorkersNum);
        log.debug("New average number of tasks per worker: floor={}, ceiling={}", floorTasks, ceilTasks);
```

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -577,15 +596,14 @@ private void resetDelay() {
         numToRevoke = floorTasks;
         for (WorkerLoad existing : existingWorkers) {
             Iterator<ConnectorTaskId> tasks = existing.tasks().iterator();
+            numToRevoke = existing.tasksSize() - ceilTasks;
```
rhauch, line 597: Isn't it possible that `numToRevoke` might be negative?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -577,15 +596,14 @@ private void resetDelay() {
...
+            log.debug("revoke number of tasks per worker {}", numToRevoke);
```
rhauch:
```suggestion
            log.debug("Tasks on worker {} is higher than ceiling, so revoking {} tasks", existing, numToRevoke);
```

rhauch (commented):
Thanks, @ramesh-muthusamy! A few comments/questions inline below.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
ramesh-muthusamy: yes it can be and it is captured under the next for loop condition.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
ramesh-muthusamy: I have updated the message.

ramesh-muthusamy:
@rhauch  I have incorporated the review comments, please help in reviewing the changes.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -260,7 +259,7 @@ private Long ensureLeaderConfig(long maxOffset, WorkerCoordinator coordinator) {
         // Do not revoke resources for re-assignment while a delayed rebalance is active
         // Also we do not revoke in two consecutive rebalances by the same leader
         canRevoke = delay == 0 && canRevoke;
-
+        log.debug("Connector and task to revoke assignment post load balancer calculation: {}", toRevoke);
```
kkonstantine: Why do we print the assignment again here. They don't seem to change since we log them[ in line 249](https://github.com/apache/kafka/pull/9319/files#diff-e24067b121eb960feebfa099bd9c30382e330eaf5db39302a9d7a50e29b3acb4L249)

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -445,16 +444,34 @@ protected void handleLostAssignments(ConnectorsAndTasks lostAssignments,
...
-                log.debug("A candidate worker has been found to assign lost tasks: {}", workerLoad.worker());
-                lostAssignments.connectors().forEach(workerLoad::assign);
-                lostAssignments.tasks().forEach(workerLoad::assign);
+            if (!candidateWorkerLoad.isEmpty()) {
+                log.debug("Assigning lost tasks to {} candidate workers: {}", 
+                        candidateWorkerLoad.size(),
+                        candidateWorkerLoad.stream().map(WorkerLoad::worker).collect(Collectors.joining(",")));
+                Iterator<WorkerLoad> candidateWorkerIterator = candidateWorkerLoad.iterator();
+                for (String connector : lostAssignments.connectors()) {
+                    // Loop over the the candidate workers as many times as it takes
```
kkonstantine, line 458: So we balance the lost tasks among the new workers now. That balances the tasks but only among the new workers. Have you checked how this works with task revocation called right after?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -577,15 +595,14 @@ private void resetDelay() {
         numToRevoke = floorTasks;
```
kkonstantine: This assignment is unused now.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
kkonstantine: Can you explain a bit what you aim to achieve with this change here?

kkonstantine: Also, we apply this logic in tasks only. But why not in the connectors too, if it helps?

kkonstantine (commented):
Thanks for the PR @ramesh-muthusamy! 
It'd be great to improve on the cases where we end up with imbalanced assignments. 

I added a commit with a test to replicate your use case and an assertion we used to have but we removed as flaky. If we are to test for imbalanced assignments we probably need to bring it back. But we'll also have to run the test several times to make sure it's not flaky. (I'll add some comments to explain the different waits in a follow up). Feel free to take a look.

kkonstantine:
Also, please try to rebase on top of the latest trunk changes to see if a build can be triggered in that case.

kkonstantine:
After writing another test, I confirmed my impression after first reading the code. So, this fix would help if all the lost workers joined after leaving the group at about the same time. But if a completely new worker joins in addition to a returning worker, then the fact that we distribute the tasks between the new workers, will lead to an imbalance compared to any existing workers. For example, 

With 2 workers with each having 8 tasks each, if 1 worker leaves and rejoins within the scheduled delay but also another new worker joins, then we'll end up with the worker that never left having 8 tasks and the other two workers having 4 tasks each.

ramesh-muthusamy:
@kkonstantine  this is a good scenario, let me test and see if 
it is already captured. Thanks for test case.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
ramesh-muthusamy: yes , I will remove the same

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
ramesh-muthusamy: It not only balances the tasks for new workers it also does revocation if there is any overloaded worker.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
ramesh-muthusamy: @kkonstantine  we are trying to revoke the additional tasks assigned [if any] to a worker so that in the next stage the revoked tasks can be re assigned to the new workers that get added to the group.  Yes we need to apply this to connectors as well, I will update the PR for the changes .

seankumar-tl:
@ramesh-muthusamy I can confirm seeing the behavior that @kkonstantine has reported. This is with Spot instances and running multiple connectors for topics with a large number of partitions. Task distribution does not take into consideration new workers that have joined and continues to result in an uneven distribution of tasks. 

Screen Shot 2020-11-03 at 12 35 59 PM

ramesh-muthusamy:
> After writing another test, I confirmed my impression after first reading the code. So, this fix would help if all the lost workers joined after leaving the group at about the same time. But if a completely new worker joins in addition to a returning worker, then the fact that we distribute the tasks between the new workers, will lead to an imbalance compared to any existing workers. For example,
> 
> With 2 workers with each having 8 tasks each, if 1 worker leaves and rejoins within the scheduled delay but also another new worker joins, then we'll end up with the worker that never left having 8 tasks and the other two workers having 4 tasks each.

@kkonstantine  I think we are already capturing this scenario, since we are taking all the existing workers while trying to do the revocation even the worker which had 8 tasks in this case and was running fine would participate in the revocation process. This would result in the existing worker with 3 tasks and the two new workers with 3 and 2 tasks respectively.

ramesh-muthusamy:
> @ramesh-muthusamy I can confirm seeing the behavior that @kkonstantine has reported. This is with Spot instances and running multiple connectors for topics with a large number of partitions. Task distribution does not take into consideration new workers that have joined and continues to result in an uneven distribution of tasks.
> 
> Screen Shot 2020-11-03 at 12 35 59 PM

@seankumar-tl  thanks for bringing this up, this PR intends to resolve the same issue. We are facing the same issue while running connector workers on K8's.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
ramesh-muthusamy: This is resolved.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
ramesh-muthusamy: this is resolved

ramesh-muthusamy:
@kkonstantine  thanks for the test case, I was able to validate that the tasks are evenly distributed among the new worker as well for both connectors and tasks. Please review. Attaching a screenshot of the test case task reassignment for reference. 
Screenshot 2020-11-16 at 11 28 23 AM

BDeus:
Just a quick question.
Can we confirmed that when a KafkaConnect Cluster process mutiple connectors, the task of each connector as equally balanced on the workers ?
Here an example:
With two connectors, one "HeavyLoad" (4 tasks) and another "LowLoad" (4 tasks) on 2 workers KafkaConnect.

I expect to have:
- Broker1:HeavyLoad1
- Broker1:HeavyLoad2
- Broker1:LowLoad1
- Broker1:LowLoad2
- Broker2:HeavyLoad3
- Broker2:HeavyLoad4
- Broker2:LowLoad3
- Broker2:LowLoad4

And not a random assignement that can lead to:
- Broker1:HeavyLoad1
- Broker1:HeavyLoad2
- Broker1:HeavyLoad3
- Broker1:HeavyLoad4
- Broker2:LowLoad1
- Broker2:LowLoad2
- Broker2:LowLoad3
- Broker2:LowLoad4

seankumar-tl:
@BDeus, I'm running @ramesh-muthusamy fix in production as a hotfix and can confirm this is now the behavior. Our cluster is processing more than a million messages a second and is stable with this fix.

BDeus:
> @BDeus, I'm running @ramesh-muthusamy fix in production as a hotfix and can confirm this is now the behavior. Our cluster is processing more than a million messages a second and is stable with this fix.

Awesome job.
Do you run multiple connectors with different load @seankumar-tl ? We had to rollback the connect.protocol to EAGER to bypass this issue (with the cons we know)

ramesh-muthusamy:
> > @BDeus, I'm running @ramesh-muthusamy fix in production as a hotfix and can confirm this is now the behavior. Our cluster is processing more than a million messages a second and is stable with this fix.
> 
> Awesome job.
> Do you run multiple connectors with different load @seankumar-tl ? We had to rollback the connect.protocol to EAGER to bypass this issue (with the cons we know)

@BDeus  it is currently not possible to differentiate the load parameters of a connector, it is  assumed that all connectors are equally while ingestion. We do have similar scenarios where we run multiple clusters  instead of loading a single cluster with variety of loads.

ramesh-muthusamy:
> @BDeus, I'm running @ramesh-muthusamy fix in production as a hotfix and can confirm this is now the behavior. Our cluster is processing more than a million messages a second and is stable with this fix.

Awesome , thanks @seankumar-tl

BDeus:
> @BDeus it is currently not possible to differentiate the load parameters of a connector, it is assumed that all connectors are equally while ingestion. We do have similar scenarios where we run multiple clusters instead of loading a single cluster with variety of loads.

I'm aware that the connector's load cannot be differentiate.
But currently, in eager mode, it's assure that each connector is equally balanced on each workers and that worker as the same amount of tasks.
I'm asking myself if it will be the same result or if it randomly move tasks from all connectors to workers to assure an equally balanced tasks (and can lead to have a worker with more task from "connector1" than another worker) ?
Thank you for your answers

ramesh-muthusamy:
We will get the same results because the task distribution considers all
the tasks independent of a connector.

ramesh-muthusamy:
@kkonstantine  gentle reminder to help reviewing the comments and updates.

ramesh-muthusamy:
@kkonstantine  requesting help in reviewing the changes

ramesh-muthusamy:
test this please

ramesh-muthusamy:
@kkonstantine  requesting your help in reviewing the PR

seankumar-tl:
@rhauch @kkonstantine - This is a major piece of functionality that is broken when running Kafka Connect at scale. Can we please get this reviewed and merged?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -554,15 +570,19 @@ private void resetDelay() {
         // We have at least one worker assignment (the leader itself) so totalWorkersNum can't be 0
         log.debug("Previous rounded down (floor) average number of connectors per worker {}", totalActiveConnectorsNum / existingWorkersNum);
         int floorConnectors = totalActiveConnectorsNum / totalWorkersNum;
-        log.debug("New rounded down (floor) average number of connectors per worker {}", floorConnectors);
+        int ceilConnectors = (int) Math.ceil((float) totalActiveConnectorsNum / totalWorkersNum);
+        log.debug("New rounded down (floor) average number of connectors per worker floor connectors {} ciel connectors ", floorConnectors, ceilConnectors);
```
kkonstantine:
```suggestion
        log.debug("New average number of connectors per worker rounded down (floor) {} and rounded up (ceil) {}", floorConnectors, ceilConnectors);
```

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -554,15 +570,19 @@ private void resetDelay() {
...
+
 
         log.debug("Previous rounded down (floor) average number of tasks per worker {}", totalActiveTasksNum / existingWorkersNum);
         int floorTasks = totalActiveTasksNum / totalWorkersNum;
-        log.debug("New rounded down (floor) average number of tasks per worker {}", floorTasks);
+        int ceilTasks = (int) Math.ceil((float) totalActiveTasksNum / totalWorkersNum);
+        log.debug("New average number of tasks per worker: floor= {}, ceiling= {}", floorTasks, ceilTasks);
```
kkonstantine:
```suggestion
        log.debug("New average number of tasks per worker rounded down (floor) {} and rounded up (ceil) {}", floorTasks, ceilTasks);
```

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -554,15 +570,19 @@ private void resetDelay() {
...
         int floorConnectors = totalActiveConnectorsNum / totalWorkersNum;
-        log.debug("New rounded down (floor) average number of connectors per worker {}", floorConnectors);
+        int ceilConnectors = (int) Math.ceil((float) totalActiveConnectorsNum / totalWorkersNum);
+        log.debug("New rounded down (floor) average number of connectors per worker floor connectors {} ciel connectors ", floorConnectors, ceilConnectors);
+
 
         log.debug("Previous rounded down (floor) average number of tasks per worker {}", totalActiveTasksNum / existingWorkersNum);
         int floorTasks = totalActiveTasksNum / totalWorkersNum;
-        log.debug("New rounded down (floor) average number of tasks per worker {}", floorTasks);
+        int ceilTasks = (int) Math.ceil((float) totalActiveTasksNum / totalWorkersNum);
```
kkonstantine: We can use the fact that these are non-negative integers. 

```suggestion
        int ceilTasks = floorTasks + ((totalActiveTasksNum % totalWorkersNum == 0) ? 0 : 1);
```

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -554,15 +570,19 @@ private void resetDelay() {
         // We have at least one worker assignment (the leader itself) so totalWorkersNum can't be 0
         log.debug("Previous rounded down (floor) average number of connectors per worker {}", totalActiveConnectorsNum / existingWorkersNum);
         int floorConnectors = totalActiveConnectorsNum / totalWorkersNum;
-        log.debug("New rounded down (floor) average number of connectors per worker {}", floorConnectors);
+        int ceilConnectors = (int) Math.ceil((float) totalActiveConnectorsNum / totalWorkersNum);
```
kkonstantine: We can use the fact that these are non-negative integers. 

```suggestion
        int ceilConnectors = floorConnectors + ((totalActiveConnectorsNum % totalWorkersNum == 0) ? 0 : 1);
```

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -260,7 +259,6 @@ private Long ensureLeaderConfig(long maxOffset, WorkerCoordinator coordinator) {
         // Do not revoke resources for re-assignment while a delayed rebalance is active
         // Also we do not revoke in two consecutive rebalances by the same leader
         canRevoke = delay == 0 && canRevoke;
-
```
kkonstantine: nit: unneeded change

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -554,15 +570,19 @@ private void resetDelay() {
...
         int floorTasks = totalActiveTasksNum / totalWorkersNum;
-        log.debug("New rounded down (floor) average number of tasks per worker {}", floorTasks);
+        int ceilTasks = (int) Math.ceil((float) totalActiveTasksNum / totalWorkersNum);
+        log.debug("New average number of tasks per worker: floor= {}, ceiling= {}", floorTasks, ceilTasks);
+        int numToRevoke;
 
-        int numToRevoke = floorConnectors;
         for (WorkerLoad existing : existingWorkers) {
             Iterator<String> connectors = existing.connectors().iterator();
+            numToRevoke = existing.connectorsSize() - ceilConnectors;
```
kkonstantine, line 586: we should remove the `if` branch on the `numToRevoke` here too, right?

kkonstantine (commented):
Thanks for the responses @ramesh-muthusamy
Left a few final comments. I think we are close.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
ramesh-muthusamy: removed the if block , thanks for pointing out.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
ramesh-muthusamy: thanks resolved it

ramesh-muthusamy:
@kkonstantine  thanks for taking time to re review the PR. I have incorporated the review comments. Please review.

kkonstantine (approved):
Thanks for this important fix @ramesh-muthusamy !

LGTM 🎉 

JDK15 green, two unrelated failures failed JDK8 and JDK11

RossierFl:
Thanks a lot everyone for this fix !

PR #9332, opened 2020-09-24
KAFKA-10511; Ensure monotonic start epoch/offset updates in `MockLog`

hachikuji:
There is a minor difference in behavior between the epoch caching logic in `MockLog` from the behavior in `LeaderEpochFileCache`. The latter ensures that every new epoch/start offset entry added to the cache increases monotonically over the previous entries. This patch brings the behavior of `MockLog` in line. 

It also simplifies the `assignEpochStartOffset` api in `ReplicatedLog`. We always intend to use the log end offset, so this patch removes the start offset parameter.

### DIFF
--- a/raft/src/test/java/org/apache/kafka/raft/MockLog.java
+++ b/raft/src/test/java/org/apache/kafka/raft/MockLog.java
@@ -310,10 +310,10 @@ public class MockLog implements ReplicatedLog {
     }
 
     @Override
-    public void assignEpochStartOffset(int epoch, long startOffset) {
-        if (startOffset != endOffset().offset)
-            throw new IllegalArgumentException(
-                "Can only assign epoch for the end offset " + endOffset().offset + ", but get offset " + startOffset);
+    public void initializeLeaderEpoch(int epoch) {
+        long startOffset = endOffset().offset;
+        epochStartOffsets.removeIf(epochStartOffset ->
+            epochStartOffset.startOffset >= startOffset || epochStartOffset.epoch >= epoch);
         epochStartOffsets.add(new EpochStartOffset(epoch, startOffset));
     }
 
--- a/raft/src/test/java/org/apache/kafka/raft/MockLogTest.java
+++ b/raft/src/test/java/org/apache/kafka/raft/MockLogTest.java
@@ -141,13 +141,8 @@ public class MockLogTest {
 
     @Test
     public void testAssignEpochStartOffset() {
-        log.assignEpochStartOffset(2, 0);
+        log.initializeLeaderEpoch(2);
         assertEquals(2, log.lastFetchedEpoch());
-    }
-
-    @Test
-    public void testAssignEpochStartOffsetNotEqualToEndOffset() {
-        assertThrows(IllegalArgumentException.class, () -> log.assignEpochStartOffset(2, 1));
     }
 
     @Test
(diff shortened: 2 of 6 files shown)


raft/src/test/java/org/apache/kafka/raft/MockLog.java
```diff
@@ -310,10 +310,10 @@ public LogFetchInfo read(long startOffset, Isolation isolation) {
     }
 
     @Override
-    public void assignEpochStartOffset(int epoch, long startOffset) {
-        if (startOffset != endOffset().offset)
-            throw new IllegalArgumentException(
-                "Can only assign epoch for the end offset " + endOffset().offset + ", but get offset " + startOffset);
+    public void initializeLeaderEpoch(int epoch) {
+        long startOffset = endOffset().offset;
+        epochStartOffsets.removeIf(epochStartOffset ->
```
abbccdda, line 315: So the idea is to wipe out the older epoch (as epoch - 1)'s end offset, and the search for (epoch - 1) would give (epoch - 2) end offset? In the case where we put leader change message in mock log, this should never happen right?

raft/src/test/java/org/apache/kafka/raft/MockLogTest.java
```diff
@@ -370,6 +365,23 @@ public void testReadOutOfRangeOffset() {
...
+    public void testMonotonicEpochStartOffset() {
+        appendBatch(5, 1);
+        assertEquals(5L, log.endOffset().offset);
+
+        log.initializeLeaderEpoch(2);
+        assertEquals(Optional.of(new OffsetAndEpoch(5L, 1)), log.endOffsetForEpoch(1));
+        assertEquals(Optional.of(new OffsetAndEpoch(5L, 2)), log.endOffsetForEpoch(2));
+
+        // Initialize a new epoch at the same end offset. The epoch cache ensures
+        // that the start offset of each retained epoch increases monotonically.
```
abbccdda, line 378: I thought this is a test for end offset?

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
```diff
@@ -712,11 +712,6 @@ void initializeElection(ElectionState election) {
 
             nodes.values().forEach(state -> {
                 state.store.writeElectionState(election);
-                if (election.hasLeader()) {
```
abbccdda, line 715: So the reason that this is safe is because we no longer try assigning offset other than the end offset?

raft/src/test/java/org/apache/kafka/raft/MockLog.java
hachikuji: There's no guarantee that the leader change message ever gets committed. It is even possible for the leadership to change multiple times before a leader change message can be committed. There is no correctness problem with the current implementation. I just wanted the behavior to be consistent with `LeaderEpochFileCache`.

raft/src/test/java/org/apache/kafka/raft/MockLogTest.java
hachikuji: The epoch cache tracks tuples of (epoch, start offset). The start offset of a new leader epoch is the current log end offset.

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
hachikuji: Oh, I think this was just unnecessary. The purpose of this logic is to initialize a certain election state prior to starting the simulation. But the code already initializes the epoch state on startup, so we didn't need this.

abbccdda (approved):
LGTM!

PR #9334, opened 2020-09-24
KAFKA-10516; Disable automatic retry of `THROTTLING_QUOTA_EXCEEDED` errors in the `kafka-topics` command (KIP-599)

dajac:
This PR does two things:
* As stated in KIP-599, we'd like to disable the automatic retry of `THROTTLING_QUOTA_EXCEEDED` errors in order to avoid letting the command hangs until `default.api.timeout.ms` is reached.
* Change the default value of `ErrorMessage` to `null` in the `DeleteTopicsResponse`. I have noted that it defaults to an empty string at the moment. The Admin client only uses the default message when the `ErrorMessage` is `null` so we end up with an empty message provided back to the user when `ErrorMessage` is not explicitly set by the broker.

### DIFF
--- a/core/src/main/scala/kafka/admin/TopicCommand.scala
+++ b/core/src/main/scala/kafka/admin/TopicCommand.scala
@@ -28,6 +28,9 @@ import kafka.utils.Implicits._
 import kafka.utils._
 import kafka.zk.{AdminZkClient, KafkaZkClient}
 import org.apache.kafka.clients.CommonClientConfigs
+import org.apache.kafka.clients.admin.CreatePartitionsOptions
+import org.apache.kafka.clients.admin.CreateTopicsOptions
+import org.apache.kafka.clients.admin.DeleteTopicsOptions
 import org.apache.kafka.clients.admin.{Admin, ConfigEntry, ListTopicsOptions, NewPartitions, NewTopic, PartitionReassignment, Config => JConfig}
 import org.apache.kafka.common.{Node, TopicPartition, TopicPartitionInfo}
 import org.apache.kafka.common.config.ConfigResource.Type
@@ -47,7 +50,6 @@ import scala.io.StdIn
 object TopicCommand extends Logging {
 
   def main(args: Array[String]): Unit = {
-
     val opts = new TopicCommandOptions(args)
     opts.checkArgs()
 
@@ -69,14 +71,24 @@ object TopicCommand extends Logging {
       else if (opts.hasDeleteOption)
         topicService.deleteTopic(opts)
     } catch {
+      case e: ExecutionException =>
+        if (e.getCause != null)
+          printException(e.getCause)
+        else
+          printException(e)
+        exitCode = 1
       case e: Throwable =>
-        println("Error while executing topic command : " + e.getMessage)
-        error(Utils.stackTrace(e))
+        printException(e)
         exitCode = 1
     } finally {
       topicService.close()
       Exit.exit(exitCode)
     }
+  }
+
+  private def printException(e: Throwable): Unit = {
+    println("Error while executing topic command : " + e.getMessage)
+    error(Utils.stackTrace(e))
   }
 
   class CommandTopicPartition(opts: TopicCommandOptions) {
@@ -244,7 +256,8 @@ object TopicCommand extends Logging {
           .toMap.asJava
 
         newTopic.configs(configsMap)
-        val createResult = adminClient.createTopics(Collections.singleton(newTopic))
+        val createResult = adminClient.createTopics(Collections.singleton(newTopic),
+          new CreateTopicsOptions().retryOnQuotaViolation(false))
         createResult.all().get()
         println(s"Created topic ${topic.name}.")
       } catch {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/kafka/admin/TopicCommand.scala
```diff
@@ -69,16 +71,26 @@ object TopicCommand extends Logging {
...
         exitCode = 1
     } finally {
       topicService.close()
       Exit.exit(exitCode)
     }
   }
 
+  private def printException(e: Throwable): Unit = {
+    println("Error while executing topic command : " + e.getMessage)
+    error(Utils.stackTrace(e))
```
rajinisivaram, line 91: Couldn't we just use `error(message, e)`?

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
```diff
@@ -844,4 +851,72 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
...
     rows(0).startsWith(s"Topic:$testTopicName\tPartitionCount:1")
   }
+
+  @Test
+  def testCreateTopicDoesNotRetryThrottlingQuotaExceededException(): Unit = {
+    val adminClient = Mockito.mock(classOf[Admin])
+    val topicService = AdminClientTopicService(adminClient)
+
+    val result = AdminClientTestUtils.createTopicsResult(testTopicName, Errors.THROTTLING_QUOTA_EXCEEDED.exception())
+    Mockito.when(adminClient.createTopics(ArgumentMatchers.any(), ArgumentMatchers.any())).thenReturn(result)
```
rajinisivaram: May be worth importing `org.mockito.Mockito._` and  `org.mockito.ArgumentMatchers._` to avoid repeating the class name everywhere.

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
```diff
@@ -844,4 +851,72 @@ class TopicCommandWithAdminClientTest extends KafkaServerTestHarness with Loggin
...
+
+    assertThrows(classOf[ThrottlingQuotaExceededException],
+      () => topicService.createTopic(new TopicCommandOptions(Array("--topic", testTopicName))))
+
+    val expectedNewTopic = new NewTopic(testTopicName, Option.empty[Integer].asJava, Option.empty[java.lang.Short].asJava)
```
rajinisivaram: We could just use Optional.empty instead of creating in Scala and converting?

rajinisivaram (commented):
@dajac Thanks for the PR, looks good. Left a few minor comments.

core/src/test/scala/unit/kafka/admin/TopicCommandWithAdminClientTest.scala
dajac: That makes sense. I had to introduce an alias for `ArgumentMatchers.eq` to not conflict with `eq`. I went with `eqThat` to remain inline with `argThat`.

core/src/main/scala/kafka/admin/TopicCommand.scala
dajac: Do you mean using `error(message, e)` to replace both `println` and `error`? I think that we are using 'println` here in order to print the message to stdout without any logger related stuff and regardless of how the logger is configured. Changing to using `error(message, e)` would break this and potentially break existing application due to introducing the logger related stuff for that message. I think that we should keep `println` here.

However, I wonder if using `error` is appropriate here as it basically prints a stacktrace for every errors. The UX does not look good. Would it make sense to use `debug` instead? The message of the exception is printed anyway and I don't think that the stacktrace provides much to regular users. WDYT?

dajac:
@rajinisivaram Thanks for your comments. I have updated the PR and answered your questions.

core/src/main/scala/kafka/admin/TopicCommand.scala
rajinisivaram: Sorry, I should have been more clear. I just meant, why do we call `Utils.stackTrace(e))` when `error(message, e)` gives you a stack trace for free. In terms of changing from `error` to `debug`, I think we include stack track in other commands. But each command seems to use something different. ConfigCommand, for example, doesn't print stacktrace for config exceptions. It logs both the error message and stacktrace to stderr. AclCommand prints for everything to stdout. Let's just keep `error` for now and maybe open a JIRA to improve and make tools consistent later?

dajac: That's a very good question. I guess that we did so in order to just print out the stacktrace without the message. Otherwise, we would have the error message printed out by the `println`, followed by the same message printed out by logger, followed by the stacktrace. Having the message twice is not necessary. I had a quick look at other commands and we do so everywhere.

I will open a JIRA to make tools more consistent. Good idea.

rajinisivaram (approved):
@dajac Thanks for the PR, merging to trunk.

PR #9335, opened 2020-09-24
Update quickstart.html

fredsh2k:
i've followed quickstart instructions here https://kafka.apache.org/documentation/#quickstart_send
and when i run:
`bin/kafka-console-producer.sh --topic quickstart-events --bootstrap-server localhost:9092`
i get:
`bootstrap-server is not a recognized option`
the solution was to use:
`bin/kafka-console-producer.sh --topic quickstart-events --broker-list localhost:9092`

### DIFF
--- a/docs/quickstart.html
+++ b/docs/quickstart.html
@@ -73,7 +73,7 @@ test
 Run the producer and then type a few messages into the console to send to the server.</p>
 
 <pre class="brush: bash;">
-&gt; bin/kafka-console-producer.sh --bootstrap-server localhost:9092 --topic test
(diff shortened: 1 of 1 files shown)


mimaison:
Which version of Kafka are you using?

Since 2.6, `--bootstrap-server` is the recommended flag. `--broker-list` is now deprecated and will be removed in the next major version.

The quickstart page is up to date and recommends the current preferred flag

fredsh2k:
kafka_2.12-2.4.0

ok, thanks for the help :)

PR #9344, opened 2020-09-28
MINOR; Preserve ThrottlingQuotaExceededException when request timeouts after being retried due to a quota violation (KIP-599)

dajac:
KIP-599 had proposed to keep returning the `ThrottlingQuotaExceededException` to the called even when the request times out due to reaching `default.api.timeout.ms`. The current implementation does not cover this yet.

From KIP-599:
> Once `default.api.timeout.ms` has been reached, the topics which were throttled will return the `ThrottlingQuotaExceededException` to the caller.

This PR adds the logic to preserve the `ThrottlingQuotaExceededException` when topics are retried. The `throttleTimeMs` is also adjusted accordingly as the request could remain pending or in-flight for quite a long time.

I have run various tests on clusters with enabled quotas and I, indeed, find it better to preserve the exception. Otherwise, the caller does not really understand what is going on. This allows the caller to take the appropriate measure and also to take the `throttleTimeMs` into consideration.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
+++ b/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
@@ -1445,6 +1445,19 @@ public class KafkaAdminClient extends AdminClient {
                 entry.getValue().completeExceptionally(new ApiException(messageFormatter.apply(entry.getKey()))));
     }
 
+    /**
+     * Fail futures in the given Map which were retried due to exceeding quota.
+     */
+    private static <K, V> void completeQuotaExceededException(
+            Map<K, KafkaFutureImpl<V>> futures,
+            Map<K, ThrottlingQuotaExceededException> quotaExceededExceptions,
+            int throttleTimeDelta) {
+        quotaExceededExceptions.forEach((key, value) -> futures.get(key).completeExceptionally(
+            new ThrottlingQuotaExceededException(
+                Math.max(0, value.throttleTimeMs() - throttleTimeDelta),
+                value.getMessage())));
+    }
+
     @Override
     public CreateTopicsResult createTopics(final Collection<NewTopic> newTopics,
                                            final CreateTopicsOptions options) {
@@ -1464,7 +1477,8 @@ public class KafkaAdminClient extends AdminClient {
         if (!topics.isEmpty()) {
             final long now = time.milliseconds();
             final long deadline = calcDeadlineMs(now, options.timeoutMs());
-            final Call call = getCreateTopicsCall(options, topicFutures, topics, deadline);
+            final Call call = getCreateTopicsCall(options, topicFutures, topics,
+                Collections.emptyMap(), now, deadline);
             runnable.call(call, now);
         }
         return new CreateTopicsResult(new HashMap<>(topicFutures));
@@ -1473,6 +1487,8 @@ public class KafkaAdminClient extends AdminClient {
     private Call getCreateTopicsCall(final CreateTopicsOptions options,
                                      final Map<String, KafkaFutureImpl<TopicMetadataAndConfig>> futures,
                                      final CreatableTopicCollection topics,
+                                     final Map<String, ThrottlingQuotaExceededException> quotaExceededExceptions,
+                                     final long now,
                                      final long deadline) {
         return new Call("createTopics", deadline, new ControllerNodeProvider()) {
             @Override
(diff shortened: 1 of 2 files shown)


clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1554,6 +1575,11 @@ private ConfigEntry configEntry(CreatableTopicConfigs config) {
 
             @Override
             void handleFailure(Throwable throwable) {
+                // If there were any topics retries due to a quota exceeded exception, we propagate
+                // the initial error back to the caller.
+                completeQuotaExceededException(futures, quotaExceededExceptions,
```
rajinisivaram: Do we want to return quota exceeded in all cases? Apart from timeouts, it seems like we should propagate failures rather an earlier quota exceeded exception?  That is, if we were throttled for 5 millis and then see a failure, the failure is more useful than the fact that we were throttled for 5 millis?

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
```diff
@@ -733,7 +733,9 @@ public void testCreateTopicsRetryThrottlingExceptionWhenEnabledUntilRequestTimeO
             time.sleep(defaultApiTimeout + 1);
 
             assertNull(result.values().get("topic1").get());
-            TestUtils.assertFutureThrows(result.values().get("topic2"), TimeoutException.class);
+            ThrottlingQuotaExceededException e = TestUtils.assertFutureThrows(result.values().get("topic2"),
+                ThrottlingQuotaExceededException.class);
+            assertEquals(0, e.throttleTimeMs());
```
rajinisivaram, line 738: I can see why we return the delta, but this looks odd when it says throttled with a time of zero.

rajinisivaram (commented):
@dajac Thanks for the PR, left a question, just to make sure I am understanding this correctly.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
dajac: Definitely. I meant to do it for `TimeoutException` only but I have forgotten it while implementing it :(. Let me correct this.

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
dajac: Yeah, I do agree but this could happen even if this should be rare. This test case is a bit stretched to verify that throttle time does not go below zero.

The reasoning of doing this is that a client could be throttled for longer than `default.api.timeout.ms`. When this happens, I believe that we should return an adjusted throttle time such that the client does not have to re-wait for the time that it has already waited for.

dajac:
@rajinisivaram Thanks for your comments. I have updated the PR.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1554,6 +1575,13 @@ private ConfigEntry configEntry(CreatableTopicConfigs config) {
 
             @Override
             void handleFailure(Throwable throwable) {
+                // If there were any topics retries due to a quota exceeded exception, we propagate
+                // the initial error back to the caller if the request timed out.
+                if (options.shouldRetryOnQuotaViolation() && throwable instanceof TimeoutException) {
+                    completeQuotaExceededException(futures, quotaExceededExceptions,
```
rajinisivaram: Perhaps we could make this `maybeCompleteQuotaExceededException` and pass in the throwable, so that we can do the check for TimeoutException in that helper method rather than in every `handleFailure`?

rajinisivaram (approved):
@dajac Thanks for the update, left one minor comment, apart from that LGTM

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
dajac: That's a very good suggestion, thanks.

dajac:
@rajinisivaram Thanks for your suggestion. I have updated the PR.

rajinisivaram (approved):
@dajac Thanks for the changes, LGTM. Will merge when builds complete.

rajinisivaram:
Builds are good, merging to trunk.

PR #8181, opened 2020-02-27
KAFKA-9584 Headers ConcurrentModificationException

MicahRam:
Prevent exceptions when modifying headers from within punctuate by creating a new Headers before punctuate is performed. 

Unit test added

This contribution is my original work and I license it to the project under the project's open source license.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9346, opened 2020-09-28
KAFKA-10535: Split ProcessorContext into Processor/StateStore/RecordContext

vvcephei:
Split up the ProcessorContext into separate containers more appropriate
for usage in `Processor#init` (`api.ProcessorContext`), 
`StateStore#init` (`StateStoreContext`), and `Processor#process`
(the new `Record` and `RecordMetadata` classes).


streams/src/main/java/org/apache/kafka/streams/processor/ProcessorContext.java
```diff
@@ -90,7 +90,7 @@ void register(final StateStore store,
      * @param name The store name
      * @return The state store instance
      */
-    StateStore getStateStore(final String name);
+    <S extends StateStore> S getStateStore(final String name);
```
vvcephei, line 93: This is a backward-compatible change. It's necessary if we want to capitalize on the opportunity to implement all the new interfaces with a common ProcessorContextImpl (which I think we do).

streams/src/main/java/org/apache/kafka/streams/processor/StateStore.java
```diff
@@ -61,7 +84,9 @@
      * @throws IllegalStateException If store gets registered after initialized is already finished
      * @throws StreamsException if the store's change log does not contain the partition
      */
-    void init(ProcessorContext context, StateStore root);
+    default void init(final StateStoreContext context, final StateStore root) {
+        init(StoreToProcessorContextAdapter.adapt(context), root);
+    }
```
vvcephei, line 89: Here's where the newly proposed StateStoreContext links in.

streams/src/main/java/org/apache/kafka/streams/processor/StateStoreContext.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.kafka.streams.StreamsMetrics;
+import org.apache.kafka.streams.errors.StreamsException;
+
+import java.io.File;
+import java.util.Map;
+
+/**
+ * State store context interface.
+ */
+public interface StateStoreContext {
```
vvcephei, line 29: The new StateStoreContext. It contains only the parts of the ProcessorContext that should be appropriate within a StateStore.

streams/src/main/java/org/apache/kafka/streams/processor/api/Processor.java
```diff
@@ -46,12 +47,12 @@
...
-     * Process the record with the given key and value.
+     * Process the record. Note that record metadata is undefined in cases such as a forward call from a punctuator.
      *
-     * @param key the key for the record
-     * @param value the value for the record
+     * @param record the record to process
+     * @param recordMetadata the metadata of the record, if it is defined
      */
-    void process(KIn key, VIn value);
+    void process(Record<KIn, VIn> record, Optional<RecordMetadata> recordMetadata);
```
vvcephei, line 55: Here's what the updated `Processor` API looks like. The "data" fields are in the `Record` (note that `ProcessorContext#forward` also accepts this class), and the "context" fields are in `RecordMetadata`, which is Optional, since it's undefined sometimes (like during a punctuation).

streams/src/main/java/org/apache/kafka/streams/processor/api/ProcessorContext.java
```diff
@@ -140,76 +125,25 @@ Cancellable schedule(final Duration interval,
...
+     * Forwards a record to the specified child processor.
      *
-     * @param key key
-     * @param value value
-     * @param to the options to use when forwarding
+     * @param record The record to forward
+     * @param childName The name of the child processor to receive the record
      */
-    <K extends KForward, V extends VForward> void forward(final K key, final V value, final To to);
+    <K extends KForward, V extends VForward> void forward(Record<K, V> record, final String childName);
```
vvcephei, line 140: Here are the symmetrical `forward` calls. Note, I dropped `To` from the API, since the timestamp will now flow through `Record`, and the only other piece of information was the child name vs. all. If desired, we could add a new "To-esque" object instead of just a String child name (if we think we may want to extend it in the future).

streams/src/main/java/org/apache/kafka/streams/processor/api/Record.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.apache.kafka.common.header.Headers;
+
+public class Record<K, V> {
+    private final K key;
+    private final V value;
+    private final long timestamp;
+    private final Headers headers;
+
+    public Record(final K key, final V value, final long timestamp, final Headers headers) {
```
vvcephei, line 28: The new data object.

streams/src/main/java/org/apache/kafka/streams/processor/api/Record.java
```diff
@@ -0,0 +1,72 @@
...
+        return timestamp;
+    }
+
+    public Headers headers() {
+        return headers;
+    }
+
+    public <NewK> Record<NewK, V> withKey(final NewK key) {
+        return new Record<>(key, value, timestamp, headers);
+    }
```
vvcephei, line 60: Note, each of these methods returns a shallow copy class, so this class itself is immutable (though the values may be mutable).

streams/src/main/java/org/apache/kafka/streams/processor/api/RecordMetadata.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.processor.api;
+
+public interface RecordMetadata {
```
vvcephei, line 19: Metadata is an interface rather than a data class, since it'll never be constructed by users. Note that this also enables us to just re-use the ProcessorContextImpl as the implementation for this.

streams/src/main/java/org/apache/kafka/streams/processor/api/StreamsHeaders.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.kafka.common.header.Headers;
+import org.apache.kafka.common.header.internals.RecordHeader;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.stream.Collectors;
+
+public class StreamsHeaders implements Headers {
```
vvcephei: This is an immutable Headers implementation. Whenever you add or remove a header, you actually get a new, independent object. This is the real, fundamental solution to the problem we patched in https://github.com/apache/kafka/pull/8181

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractProcessorContext.java
```diff
@@ -156,9 +157,10 @@ public long offset() {
     @Override
     public Headers headers() {
         if (recordContext == null) {
-            throw new IllegalStateException("This should not happen as headers() should only be called while a record is processed");
+            return StreamsHeaders.emptyHeaders();
```
vvcephei: I did some hacking around in this PR to make things work reasonably. I'm not sure I always made exactly the right call, but the main point was to prototype the proposed public API, which seems to work nicely.

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -147,7 +148,7 @@ public void setGlobalProcessorContext(final InternalProcessorContext globalProce
             globalStoreNames.add(stateStore.name());
             final String sourceTopic = storeToChangelogTopic.get(stateStore.name());
             changelogTopics.add(sourceTopic);
-            stateStore.init(globalProcessorContext, stateStore);
+            stateStore.init((StateStoreContext) globalProcessorContext, stateStore);
```
vvcephei, line 151: When you see this, it's because the concrete context type implements both StateStoreContext and ProcessorContext, and we have `StateStore#init` overloads that take both, so I had to select the method we want to invoke.

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalApiProcessorContext.java
```diff
@@ -1,119 +0,0 @@
...
-import org.apache.kafka.streams.state.StoreBuilder;
-import org.apache.kafka.streams.state.internals.ThreadCache;
-import org.apache.kafka.streams.state.internals.ThreadCache.DirtyEntryFlushListener;
-
-/**
- * For internal use so we can update the {@link RecordContext} and current
- * {@link ProcessorNode} when we are forwarding items that have been evicted or flushed from
- * {@link ThreadCache}
- */
-public interface InternalApiProcessorContext<KForward, VForward> extends ProcessorContext<KForward, VForward> {
```
vvcephei, line 34: Because the updated new API no longer clashes with the old one, we don't need a separate implementation anymore.

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorContextAdapter.java
```diff
@@ -1,235 +0,0 @@
...
-import org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl;
-import org.apache.kafka.streams.state.StoreBuilder;
-import org.apache.kafka.streams.state.internals.ThreadCache;
-
-import java.io.File;
-import java.time.Duration;
-import java.util.Map;
-
-public final class ProcessorContextAdapter<KForward, VForward>
-    implements ProcessorContext<KForward, VForward>, InternalApiProcessorContext<KForward, VForward> {
```
vvcephei, line 39: Because the updated new API no longer clashes with the old one, we also don't need these adapters anymore.

vvcephei (commented):
Hey @mjsax , what do you think about this KIP-478 alternative?

vvcephei:
Closing POC in favor of new PR: https://github.com/apache/kafka/pull/9361

PR #9347, opened 2020-09-28
KAFKA-10531: Check for negative values to Thread.sleep call

soondenana:
System.currentTimeMillis() is not monotonic, so using that to calculate
time to sleep can result in negative values. That will throw
IllegalArgumentException.

This change checks for that and sleeps for a second (to avoid tight
loop) if the value returned is negative.

This change need to be backported to older branches that have Connect.

comment to ping reviewers.*

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/utils/Utils.java
+++ b/clients/src/main/java/org/apache/kafka/common/utils/Utils.java
@@ -334,6 +334,11 @@ public final class Utils {
      */
     public static void sleep(long ms) {
         try {
+            if (ms <= 0) {
+                // No need to sleep
+                log.debug("Skipping sleep as asked to sleep for {} msec", ms);
+                return;
+            }
             Thread.sleep(ms);
         } catch (InterruptedException e) {
             // this is okay, we just wake up early
--- a/clients/src/main/java/org/apache/kafka/common/utils/Utils.java
+++ b/clients/src/main/java/org/apache/kafka/common/utils/Utils.java
@@ -334,11 +334,6 @@ public final class Utils {
      */
     public static void sleep(long ms) {
         try {
-            if (ms <= 0) {
-                // No need to sleep
-                log.debug("Skipping sleep as asked to sleep for {} msec", ms);
-                return;
-            }
             Thread.sleep(ms);
         } catch (InterruptedException e) {
             // this is okay, we just wake up early
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
@@ -136,7 +136,7 @@ public class KafkaBasedLog<K, V> {
         List<PartitionInfo> partitionInfos = consumer.partitionsFor(topic);
         long started = time.nanoseconds();
         long maxSleepMs = 1_000;
-        long sleepMs = 10;
+        long sleepMs = 100;
         while (partitionInfos == null && time.nanoseconds() - started < CREATE_TOPIC_TIMEOUT_NS) {
             time.sleep(sleepMs);
             sleepMs = Math.min(2 * sleepMs, maxSleepMs);
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
@@ -45,6 +45,7 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 
 
 /**
@@ -70,7 +71,7 @@ import java.util.concurrent.Future;
  */
 public class KafkaBasedLog<K, V> {
     private static final Logger log = LoggerFactory.getLogger(KafkaBasedLog.class);
-    private static final long CREATE_TOPIC_TIMEOUT_MS = 30000;
+    private static final long CREATE_TOPIC_TIMEOUT_NS = TimeUnit.NANOSECONDS.convert(30, TimeUnit.SECONDS);
 
     private Time time;
     private final String topic;
@@ -133,11 +134,14 @@ public class KafkaBasedLog<K, V> {
         List<TopicPartition> partitions = new ArrayList<>();
 
         // We expect that the topics will have been created either manually by the user or automatically by the herder
-        List<PartitionInfo> partitionInfos = null;
-        long started = time.milliseconds();
-        while (partitionInfos == null && time.milliseconds() - started < CREATE_TOPIC_TIMEOUT_MS) {
+        List<PartitionInfo> partitionInfos = consumer.partitionsFor(topic);
+        long started = time.nanoseconds();
+        long maxSleepMs = 1_000;
+        long sleepMs = 10;
+        while (partitionInfos == null && time.nanoseconds() - started < CREATE_TOPIC_TIMEOUT_NS) {
+            Utils.sleep(sleepMs);
+            sleepMs = Math.min(2 * sleepMs, maxSleepMs);
             partitionInfos = consumer.partitionsFor(topic);
-            waitForTopicCreate(started, time);
         }
         if (partitionInfos == null)
             throw new ConnectException("Could not look up partition metadata for offset backing store topic in" +
@@ -161,18 +165,6 @@ public class KafkaBasedLog<K, V> {
         log.info("Finished reading KafkaBasedLog for topic " + topic);
 
         log.info("Started KafkaBasedLog for topic " + topic);
-    }
-
-    /**
-     * Sleep for some time so that topic used for this KafkaBasedLog gets created. Note that
-     * {@code System.currentTimeMillis()} is not monotonic, so check for that condition.
-     */
-    // Visible for Testing
-    static void waitForTopicCreate(long started, Time time) {
-        long timeToWait = Math.max(time.milliseconds() - started, 1);
-        timeToWait = Math.min(timeToWait, 1000);
-        log.debug("Going to wait for {} msecs", timeToWait);
-        Utils.sleep(timeToWait);
     }
 
     public void stop() {
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
@@ -137,7 +137,7 @@ public class KafkaBasedLog<K, V> {
         long started = time.milliseconds();
         while (partitionInfos == null && time.milliseconds() - started < CREATE_TOPIC_TIMEOUT_MS) {
             partitionInfos = consumer.partitionsFor(topic);
-            Utils.sleep(Math.min(time.milliseconds() - started, 1000));
+            waitForTopicCreate(started, time);
         }
         if (partitionInfos == null)
             throw new ConnectException("Could not look up partition metadata for offset backing store topic in" +
@@ -161,6 +161,18 @@ public class KafkaBasedLog<K, V> {
         log.info("Finished reading KafkaBasedLog for topic " + topic);
 
         log.info("Started KafkaBasedLog for topic " + topic);
+    }
+
+    /**
+     * Sleep for some time so that topic used for this KafkaBasedLog gets created. Note that
+     * {@code System.currentTimeMillis()} is not monotonic, so check for that condition.
+     */
+    // Visible for Testing
+    static void waitForTopicCreate(long started, Time time) {
+        long timeToWait = Math.max(time.milliseconds() - started, 1);
+        timeToWait = Math.min(timeToWait, 1000);
+        log.debug("Going to wait for {} msecs", timeToWait);
+        Utils.sleep(timeToWait);
     }
 
     public void stop() {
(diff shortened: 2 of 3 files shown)


connect/runtime/src/test/java/org/apache/kafka/connect/util/KafkaBasedLogTest.java
```diff
@@ -536,6 +536,22 @@ public void onCompletion(RecordMetadata metadata, Exception exception) {
         PowerMock.verifyAll();
     }
 
+    /**
+     * Check if the waitForTopicCreate method doesn't throw if time moves backward, and works
+     * correctly if it increases.
+     */
+    @Test
+    public void testWatiForTopicCreate() {
```
splett2: nit: typo in `Wati`

connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
```diff
@@ -163,6 +163,18 @@ public void start() {
         log.info("Started KafkaBasedLog for topic " + topic);
     }
 
+    /**
+     * Sleep for some time so that topic used for this KafkaBasedLog gets created. Note that
+     * {@code System.currentTimeMillis()} is not monotonic, so check for that condition.
```
ijuma: Why don't we use a monotonic timer?

soondenana: Wanted to make least amount of change. I can update the code to use monotonic 'nanoTime` instead (nanoseconds in Time interface). We will also need to convert that to milli before passing to sleep (unless we want to add nano to those interfaces too, like Utils.sleep)

soondenana: In fact I don't like this loop altogether. Going to rewrite it so that it doesn't use these constructs.

ijuma: You can use `hiResClockMs` so that you get the value in `ms`.

connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
```diff
@@ -133,11 +134,14 @@ public void start() {
...
         // We expect that the topics will have been created either manually by the user or automatically by the herder
-        List<PartitionInfo> partitionInfos = null;
-        long started = time.milliseconds();
-        while (partitionInfos == null && time.milliseconds() - started < CREATE_TOPIC_TIMEOUT_MS) {
+        List<PartitionInfo> partitionInfos = consumer.partitionsFor(topic);
+        long started = time.nanoseconds();
+        long maxSleepMs = 1_000;
+        long sleepMs = 10;
+        while (partitionInfos == null && time.nanoseconds() - started < CREATE_TOPIC_TIMEOUT_NS) {
+            Utils.sleep(sleepMs);
```
ijuma: `Time` has a `sleep` method too btw.

connect/runtime/src/test/java/org/apache/kafka/connect/util/KafkaBasedLogTest.java
soondenana: Fixed (actually removed the test)

connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
soondenana: In fact we don't need value from clock to sleep, only need it to find elapsed time and timeout. I have decoupled these two, and also fixed a minor issue where it was sleeping first time even if topic was present.

Please take a look.

connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
soondenana: Updated.

stanislavkozlovski (approved):
LGTM! Thanks for the PR!

connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
```diff
@@ -133,11 +133,14 @@ public void start() {
...
-        long started = time.milliseconds();
-        while (partitionInfos == null && time.milliseconds() - started < CREATE_TOPIC_TIMEOUT_MS) {
+        List<PartitionInfo> partitionInfos = consumer.partitionsFor(topic);
+        long started = time.nanoseconds();
+        long maxSleepMs = 1_000;
+        long sleepMs = 10;
+        while (partitionInfos == null && time.nanoseconds() - started < CREATE_TOPIC_TIMEOUT_NS) {
+            time.sleep(sleepMs);
+            sleepMs = Math.min(2 * sleepMs, maxSleepMs);
             partitionInfos = consumer.partitionsFor(topic);
```
chia7712, line 143: Why not using ```partitionsFor(String topic, Duration timeout) ``` (https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java#L1944) to replace while loop?

soondenana: I think the semantics needed here is different. The timeout in `partitionsFor` is the max amount of time the api can block waiting for response before it fails with `TimeoutException`. However, the api can return within timeout with empty results as newly created topics data has not been propagated yet. We then have to retry again until `partitionsFor` returns the partition data (upto a max time).

chia7712: It seems to me the behavior of ```timeout``` is not consistent in consumer methods. The ```timeout``` used by other methods (for example:  ```position```, ```offsetsForTimes```, ```beginningOffsets``` and ```endOffsets```) is to await the result of specify partitions. It means consumer will send a request again if the timer is not expired and the specify partition has no metadata (i.e topics data has not been propagated yet). Maybe ```partitionsFor``` should be fixed for consistent behavior.

soondenana: Not sure about this, no result for "partitionInfos" is a valid result. There is no point in automatic retrying. While for other apis that retry automatically, they do it if they get an invalid result back. If `null` was a valid result for them, they shouldn't retry either.

chia7712: @soondenana Thanks for responses. This patch LGTM. What I want to discuss is unrelated to this patch.

> they do it if they get an invalid result back. If null was a valid result for them, they shouldn't retry either.

If the topic is not exist (or not been propagated yet), ```partitionFor``` can return null. By contrast, ```beginningOffsets``` (and other methods) throws ```TimeoutException```. In order to make consistent behavior, ```beginningOffsets``` (and other methods) should let the topic (or partition) be absent in the returned ```Map```. WDYT?

chia7712 (approved):

rhauch:
@soondenana if the goal is to minimize the changes, would it be sufficient to change the code to use `Time.sleep(...)` instead of `Thread.sleep(...)`, and then change the `SystemTime.sleep(...)` implementation to return immediately if the supplied number of milliseconds is <= 0?

Are there performance implications of using `System.nanoTime()` instead of `System.currentTimeMillis()`?

connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
```diff
@@ -70,7 +70,7 @@
  */
 public class KafkaBasedLog<K, V> {
     private static final Logger log = LoggerFactory.getLogger(KafkaBasedLog.class);
-    private static final long CREATE_TOPIC_TIMEOUT_MS = 30000;
+    private static final long CREATE_TOPIC_TIMEOUT_NS = TimeUnit.NANOSECONDS.convert(30, TimeUnit.SECONDS);
```
rhauch: We are often using this pattern in other places for time-related constants:
```suggestion
    private static final long CREATE_TOPIC_TIMEOUT_NS = TimeUnit.SECONDS.toNanos(30);
```

connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
```diff
@@ -133,11 +133,14 @@ public void start() {
         List<TopicPartition> partitions = new ArrayList<>();
 
         // We expect that the topics will have been created either manually by the user or automatically by the herder
-        List<PartitionInfo> partitionInfos = null;
-        long started = time.milliseconds();
-        while (partitionInfos == null && time.milliseconds() - started < CREATE_TOPIC_TIMEOUT_MS) {
+        List<PartitionInfo> partitionInfos = consumer.partitionsFor(topic);
+        long started = time.nanoseconds();
+        long maxSleepMs = 1_000;
```
rhauch: Let's define a constant above:
```
private static final long MAX_SLEEP_MS = TimeUnit.SECONDS.toMillis(1);
```
and then we can replace `maxSleepMs` with `MAX_SLEEP_MS`.

rhauch (commented):
I have a few comments/suggestions on the current proposal, though I still wonder about the performance implication of using `System.nanoTime()` instead of `System.currentTimeMillis()`.

soondenana:
> @soondenana if the goal is to minimize the changes, would it be sufficient to change the code to use `Time.sleep(...)` instead of `Thread.sleep(...)`, and then change the `SystemTime.sleep(...)` implementation to return immediately if the supplied number of milliseconds is <= 0?
> 
> Are there performance implications of using `System.nanoTime()` instead of `System.currentTimeMillis()`?

Thanks @rhauch for taking a look. The PR started with only fixing negative sleep vlaue, but as we started looking at code more, there were multiple issue so the "minimize change" idea was dropped. Here are 3 issues with original code:

1. Negative sleep value
2. Using elapsed time since loop started to decide on next sleep time.
3. Sleeping even if the `partitionsFor` call would be successful for first time

Considering that I decided to rewrite the loop to fix all three issues. 

Good question on performance. It seems like `System.nanoTime()` is slower than `System.currentTimeMillis` (one is read from h/w using i/o another one is reading from memory), but not sure if that has any implication here for this use case. We are waiting for topic to appear in metadata and sleeping for seconds. Few milliseconds difference should not matter.

connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
soondenana: Fixed.

connect/runtime/src/main/java/org/apache/kafka/connect/util/KafkaBasedLog.java
soondenana: Updated.

rhauch:
> Considering that I decided to rewrite the loop to fix all three issues.

That makes sense, and I noticed the same when reviewing.

> Good question on performance. It seems like System.nanoTime() is slower than System.currentTimeMillis (one is read from h/w using i/o another one is reading from memory), but not sure if that has any implication here for this use case. We are waiting for topic to appear in metadata and sleeping for seconds. Few milliseconds difference should not matter.

That's probably true. But if `Thread.sleep(negativeTime)` throws an `IllegalArgumentException`, should we also change `Utils.sleep(...)` (called by `SystemTime.sleep(...)` that is now used above) to check and return immediately for a negative number of milliseconds to sleep?

soondenana:
> should we also change Utils.sleep(...) (called by SystemTime.sleep(...) that is now used above) to check and return immediately for a negative number of milliseconds to sleep?

Yes, makes sense. Updated code to do so.

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
```diff
@@ -331,6 +331,11 @@ public static ByteBuffer wrapNullable(byte[] array) {
      */
     public static void sleep(long ms) {
         try {
+            if (ms <= 0) {
+                // No need to sleep
+                log.debug("Skipping sleep as asked to sleep for {} msec", ms);
```
ijuma: Won't this hide bugs?

soondenana: Yes. I have seen things done both ways i.e. behave reasonably in face of weird args vs throw exception. `Time.sleep` combines both approaches, throws for values `< 0` but skips if sleep `= 0` and goes to sleep for values `> 0`.

Its a judgement call, we can delegate to JDK and remove this check.

rhauch (approved):
LGTM. Thanks, @soondenana. And I'm fine removing the `Utils.sleep(...)` change, too, if that's the preferred approach.

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
ijuma: Overflow is not uncommon when using things like Long.MaxValue for timeouts, so assuming negative is not a bug seems dangerous.

soondenana: Ok, I see the point. Overflow can be caused by time arithmetic too. Reverted the change.

soondenana:
There was an error when building `streams.examples`:

```
[2020-10-05T08:40:05.722Z] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-archetype-plugin:3.2.0:generate (default-cli) on project standalone-pom: A Maven project already exists in the directory /home/jenkins/workspace/Kafka_kafka-pr_PR-9347/streams/quickstart/test-streams-archetype/streams.examples -> [Help 1]
```

The failure is not related to this PR.

PR #9348, opened 2020-09-29
KAFKA-10527; Voters should not reinitialize as leader in same epoch

hachikuji:
One of the invariants that the raft replication protocol ensures is that each record is uniquely identified by leader epoch and offset. This can be violated if a leader remains elected with the same epoch between restarts since unflushed data could be lost.


guozhangwang (approved):
LGTM!

PR #9349, opened 2020-09-29
MINOR: add proper checks to KafkaConsumer.groupMetadata

chia7712:
add following checks to ```KafkaConsumer.groupMetadata```

1. null check of coordinator (replace NPE by ```InvalidGroupIdException``` which is same to other methods)
1. concurrent check (```groupMetadata``` is not thread-safe so concurrent check is necessary)

### DIFF
--- a/clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
@@ -103,6 +103,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.ConcurrentModificationException;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
```diff
@@ -1994,6 +1995,24 @@ public void testGetGroupMetadata() {
...
+    @Test
+    public void testInvalidGroupMetadata() {
+        Properties props = new Properties();
+        props.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9999");
+        props.setProperty(ConsumerConfig.METRIC_REPORTER_CLASSES_CONFIG, MockMetricsReporter.class.getName());
+        KafkaConsumer<String, String> consumer = new KafkaConsumer<>(
+                props, new StringDeserializer(), new StringDeserializer());
+
+        // concurrent access is illegal
+        consumer.currentThread.set(123);
```
hachikuji: Another way to test this might be to use `MockClient.enableBlockingUntilWakeup`. That would let us block in `Consumer.poll`.

hachikuji (commented):
Thanks, the fix looks good to me. Had a minor suggestion you can consider for the test case.

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
chia7712: will Roger that!

hachikuji (approved):
LGTM. Thanks for the patch.

PR #9352, opened 2020-09-29
KAFKA-10533; KafkaRaftClient should flush log after appends

hachikuji:
This patch adds missing flush logic to `KafkaRaftClient`. The initial flushing behavior is simplistic. We guarantee that the leader will not replicate above the last flushed offset and we guarantee that the follower will not fetch data above its own flush point. More sophisticated flush behavior is proposed in KAFKA-10526.

We have also extended the simulation test so that it covers flush behavior. When a node is shutdown, all unflushed data is lost. We were able to confirm that the monotonic high watermark invariant fails without the added `flush` calls.

This patch also piggybacks a fix to the `TestRaftServer` implementation. The initial check-in contained a bug which caused `RequestChannel` to fail sending responses because the disabled APIs did not have metrics registered. As a result of this, it is impossible to elect leaders.

### DIFF
--- a/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
+++ b/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
@@ -985,13 +985,7 @@ public class KafkaRaftClient implements RaftClient {
             } else {
                 Records records = (Records) partitionResponse.recordSet();
                 if (records.sizeInBytes() > 0) {
-                    LogAppendInfo info = log.appendAsFollower(records);
-                    log.flush();
-
-                    OffsetAndEpoch endOffset = endOffset();
-                    kafkaRaftMetrics.updateFetchedRecords(info.lastOffset - info.firstOffset + 1);
-                    kafkaRaftMetrics.updateLogEnd(endOffset);
-                    logger.trace("Follower end offset updated to {} after append", endOffset);
+                    appendAsFollower(records);
                 }
                 OptionalLong highWatermark = partitionResponse.highWatermark() < 0 ?
                     OptionalLong.empty() : OptionalLong.of(partitionResponse.highWatermark());
@@ -1003,6 +997,18 @@ public class KafkaRaftClient implements RaftClient {
         } else {
             return handleUnexpectedError(error, responseMetadata);
         }
+    }
+
+    private void appendAsFollower(
+        Records records
+    ) {
+        LogAppendInfo info = log.appendAsFollower(records);
+        log.flush();
+
+        OffsetAndEpoch endOffset = endOffset();
+        kafkaRaftMetrics.updateFetchedRecords(info.lastOffset - info.firstOffset + 1);
+        kafkaRaftMetrics.updateLogEnd(endOffset);
+        logger.trace("Follower end offset updated to {} after append", endOffset);
     }
 
     private LogAppendInfo appendAsLeader(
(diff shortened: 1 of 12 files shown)


guozhangwang:
test this please

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -193,8 +193,9 @@ private void updateFollowerHighWatermark(
     ) {
         highWatermarkOpt.ifPresent(highWatermark -> {
             long newHighWatermark = Math.min(endOffset().offset, highWatermark);
-            state.updateHighWatermark(OptionalLong.of(newHighWatermark));
-            updateHighWatermark(state, currentTimeMs);
+            if (state.updateHighWatermark(OptionalLong.of(newHighWatermark))) {
```
guozhangwang, line 196: Just to clarify this is not a correctness bugfix, but just to optimize away unnecessary purgatory access right?

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
```diff
@@ -114,14 +113,6 @@ private boolean updateHighWatermark() {
         return false;
     }
 
-    private OptionalLong quorumMajorityFetchTimestamp() {
```
guozhangwang, line 117: Good catch.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
```diff
@@ -103,6 +103,16 @@
...
 
+    /**
+     * Flush the current log to disk.
+     */
+    void flush();
+
+    /**
+     * Get the last offset which has been flushed to disk.
+     */
+    long lastFlushedOffset();
```
guozhangwang, line 114: Is this function going to be used for non-testing code in the future?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -985,6 +986,8 @@ private boolean handleFetchResponse(
                 Records records = (Records) partitionResponse.recordSet();
                 if (records.sizeInBytes() > 0) {
                     LogAppendInfo info = log.appendAsFollower(records);
+                    log.flush();
```
guozhangwang: nit: maybe we can wrap the flushing and hwm updating logic in a `flushFollowerLog` as well.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Yeah, right. Also to avoid the log spam when the high watermark doesn't actually increment.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
hachikuji: I think we will need it to implement the leader deferred fsync or any other optimization which allows the followers to read unflushed data from the leader. I decided to add it to the interface since we have a trivial implementation for it in `KafkaMetadataLog` already. We can save it for later if you prefer.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Makes sense.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
guozhangwang: nah, I'm just curious to see what's in your future plan. Let's just keep it.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: I considered this, but it's a little more awkward since the high watermark is coming from the request. However, I pushed an update which factors out an `appendAsFollower` call which is nearly symmetric with the existing `appendAsLeader`.

guozhangwang (approved):

PR #9353, opened 2020-09-29
KAFKA-10521; Skip partition watch registration when `AlterIsr` is expected

hachikuji:
Before `AlterIsr` with KIP-497, the controller would register watches in Zookeeper for each reassigning partition so that it could be notified immediately when the ISR was expanded and the reassignment could be completed. This notification is not needed with the latest IBP when `AlterIsr` is enabled because the controller will execute all ISR changes itself.

There is one subtle detail. If we are in the middle of a roll in order to bump the IBP, then it is possible for the controller to be on the latest IBP while some of the brokers are still on the older one. In this case, the brokers on the older IBP will not send `AlterIsr`, but we can still rely on the delayed notification through the `isr_notifications` path to complete reassignments. This seems like a reasonable tradeoff since it should be a short window before the roll is completed.

### DIFF
--- a/core/src/main/scala/kafka/controller/KafkaController.scala
+++ b/core/src/main/scala/kafka/controller/KafkaController.scala
@@ -2080,6 +2080,17 @@ class KafkaController(val config: KafkaConfig,
       if (partitions.nonEmpty) {
         updateLeaderAndIsrCache(partitions)
         processUpdateNotifications(partitions)
+
+        // During a partial upgrade, the controller may be on an IBP which assumes
+        // ISR changes through the `AlterIsr` API while some brokers are on an older
+        // IBP which assumes notification through Zookeeper. In this case, since the
+        // controller will not have registered watches for reassigning partitions, we
+        // can still rely on the batch ISR change notification path in order to
+        // complete the reassignment.
+        partitions.filter(controllerContext.partitionsBeingReassigned.contains).foreach { partition =>
+          val assignment = controllerContext.partitionFullReplicaAssignment(partition)
+          onPartitionReassignment(partition, assignment)
+        }
       }
     } finally {
       // delete the notifications
--- a/core/src/main/scala/kafka/controller/KafkaController.scala
+++ b/core/src/main/scala/kafka/controller/KafkaController.scala
@@ -1837,13 +1837,17 @@ class KafkaController(val config: KafkaConfig,
     if (!isActive) return
 
     if (controllerContext.partitionsBeingReassigned.contains(topicPartition)) {
-      val reassignment = controllerContext.partitionFullReplicaAssignment(topicPartition)
-      if (isReassignmentComplete(topicPartition, reassignment)) {
-        // resume the partition reassignment process
-        info(s"Target replicas ${reassignment.targetReplicas} have all caught up with the leader for " +
-          s"reassigning partition $topicPartition")
-        onPartitionReassignment(topicPartition, reassignment)
-      }
+      maybeCompleteReassignment(topicPartition)
+    }
+  }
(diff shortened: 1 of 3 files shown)


dajac:
> There is one subtle detail. If we are in the middle of a roll in order to bump the IBP, then it is possible for the controller to be on the latest IBP while some of the brokers are still on the older one. In this case, the brokers on the older IBP will not send AlterIsr, but we can still rely on the delayed notification through the isr_notifications path to complete reassignments. This seems like a reasonable tradeoff since it should be a short window before the roll is completed.

I do agree that this is a reasonable tradeoff.

dajac (commented):
LGTM, thanks for the PR.

core/src/main/scala/kafka/controller/KafkaController.scala
```diff
@@ -1907,6 +1915,20 @@ class KafkaController(val config: KafkaConfig,
...
+
+    // After we have returned the result of the `AlterIsr` request, we should check whether
+    // there are any reassignments which can be completed by a successful ISR expansion.
+    response.left.foreach { alterIsrResponses =>
+      alterIsrResponses.forKeyValue { (topicPartition, partitionResponse) =>
+        if (controllerContext.partitionsBeingReassigned.contains(topicPartition)) {
+          val isSuccessfulUpdate = partitionResponse.isRight
+          if (isSuccessfulUpdate) {
+            val assignment = controllerContext.partitionFullReplicaAssignment(topicPartition)
+            onPartitionReassignment(topicPartition, assignment)
```
junrao: This old logic in processPartitionReassignmentIsrChange() does an info logging before calling onPartitionReassignment(). Should we do the same thing here?

junrao (commented):
@hachikuji : Thanks for the PR. A couple of comments.

"There is one subtle detail. If we are in the middle of a roll in order to bump the IBP, then it is possible for the controller to be on the latest IBP while some of the brokers are still on the older one. In this case, the brokers on the older IBP will not send AlterIsr, but we can still rely on the delayed notification through the isr_notifications path to complete reassignments. This seems like a reasonable tradeoff since it should be a short window before the roll is completed."

Currently, processIsrChangeNotification() doesn't seem to call onPartitionReassignment(). onPartitionReassignment() is called on broker restart. So, if all brokers continue the IBP bump process, the reassignement will complete. However, in the rare cases where the IBP bump process stops, the reassignment may not complete.

hachikuji:
@junrao Great catch. I had just assumed the check existed. I updated the code and added a partial upgrade integration test case.

core/src/test/scala/integration/kafka/admin/ReassignPartitionsIntegrationTest.scala
```diff
@@ -60,13 +57,53 @@ class ReassignPartitionsIntegrationTest extends ZooKeeperTestHarness {
...
+    val configOverrides = Map(KafkaConfig.InterBrokerProtocolVersionProp -> KAFKA_2_7_IV1.version)
+    cluster = new ReassignPartitionsTestCluster(zkConnect, configOverrides = configOverrides)
+    cluster.setup()
+    executeAndVerifyReassignment()
+  }
+
+  @Test
+  def testReassignmentCompletionDuringPartialUpgrade(): Unit = {
+    // Test reassignment during a partial upgrade when some brokers are relying on
+    // `AlterIsr` and some on rely on the old notification logic through Zookeeper.
```
junrao: typo "some on"

junrao (approved):
@hachikuji : Thanks for the updated PR. LGTM. Just a minor comment below.

PR #9354, opened 2020-09-29
KAFKA-10134 Follow-up: Set the re-join flag in heartbeat failure

guozhangwang:

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
@@ -948,7 +948,7 @@ public abstract class AbstractCoordinator implements Closeable {
     synchronized void resetGenerationOnResponseError(ApiKeys api, Errors error) {
         log.debug("Resetting generation after encountering {} from {} response and requesting re-join", error, api);
 
-        resetState();
+        resetStateAndRejoin();
     }
 
     synchronized void resetGenerationOnLeaveGroup() {
--- a/clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
@@ -1862,7 +1862,6 @@ public class KafkaConsumerTest {
         TestUtils.waitForCondition(() -> {
             consumer.poll(Duration.ofMillis(100L));
             return consumer.assignment().equals(Utils.mkSet(tp0, t2p0));
-
         }, "Does not complete rebalance in time");
 
         assertEquals(Utils.mkSet(topic, topic2), consumer.subscription());
@@ -1933,14 +1932,13 @@ public class KafkaConsumerTest {
         fetches1.put(tp0, new FetchInfo(3, 1));
         client.respondFrom(fetchResponse(fetches1), node);
 
-        // now complete teh rebalance
+        // now complete the rebalance
         client.respondFrom(syncGroupResponse(Arrays.asList(tp0, t3p0), Errors.NONE), coordinator);
 
         AtomicInteger count = new AtomicInteger(0);
         TestUtils.waitForCondition(() -> {
             ConsumerRecords<String, String> recs = consumer.poll(Duration.ofMillis(100L));
-            System.out.println("count " + count.addAndGet(recs.count()));
-            return consumer.assignment().equals(Utils.mkSet(tp0, t3p0)) && count.get() == 1;
+            return consumer.assignment().equals(Utils.mkSet(tp0, t3p0)) && count.addAndGet(recs.count()) == 1;
 
         }, "Does not complete rebalance in time");
 
@@ -1958,8 +1956,7 @@ public class KafkaConsumerTest {
         count.set(0);
         TestUtils.waitForCondition(() -> {
             ConsumerRecords<String, String> recs = consumer.poll(Duration.ofMillis(100L));
-            System.out.println("count2 " + count.addAndGet(recs.count()));
-            return count.get() == 101;
+            return count.addAndGet(recs.count()) == 101;
 
         }, "Does not complete rebalance in time");
 
--- a/clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
@@ -120,6 +120,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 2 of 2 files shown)


guozhangwang:
@ableegoldman @abbccdda to review

ableegoldman (approved):
Thanks for digging into this. Maybe we can set up alerts on the `last-rebalance-seconds-ago` metric to catch this sort of thing faster

One thing I don't yet understand is, why did this affect all three threads on a single instance (at about the same time)? Was it just because the instance in question didn't have caught-up state for any active tasks, thus it was only assigned stateless tasks across all three threads?

Also, if the thread is only assigned stateless tasks, then shouldn't it reach RUNNING and therefore start to call `commit` _earlier_ than a thread with some stateful tasks? But we observed that the threads on the problem-instance seem to never rejoin the group at all, right? Is that just another symptom of this bug?

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -948,7 +948,7 @@ private synchronized void resetStateAndRejoin() {
     synchronized void resetGenerationOnResponseError(ApiKeys api, Errors error) {
         log.debug("Resetting generation after encountering {} from {} response and requesting re-join", error, api);
 
-        resetState();
+        resetStateAndRejoin();
```
chia7712, line 951: @dajac I feel this patch can fix https://issues.apache.org/jira/browse/KAFKA-8266

The test case in https://issues.apache.org/jira/browse/KAFKA-8266 restarts all brokers to activate new configs. However, the data of group (__consumer_offsets) may get lost if all brokers are killed too quick (the log folder is changed when restarting broker so the restarting broker has to fetch data from other brokers). The heartbeat of running consumer will encounter ```UNKNOWN_MEMBER_ID``` in sending heartbeat request since the group data is gone. Without this patch, the heartbeat thread is disabled and the state is in ```UNJOINED``` but the ```rejoinNeeded``` is still false. In short, the consumer is not going to rejoin group so we can't see expected error.

dajac: Nice find! I analyzed the logs and I was actually trying to find out why the consumer was not rejoining. The logs coincides with what you described. It may be due to this as you pointed out.

guozhangwang:
> One thing I don't yet understand is, why did this affect all three threads on a single instance (at about the same time)? Was it just because the instance in question didn't have caught-up state for any active tasks, thus it was only assigned stateless tasks across all three threads?
> 
> Also, if the thread is only assigned stateless tasks, then shouldn't it reach RUNNING and therefore start to call `commit` _earlier_ than a thread with some stateful tasks? But we observed that the threads on the problem-instance seem to never rejoin the group at all, right? Is that just another symptom of this bug?

From the soak logs what I observed is that, the three threads from that clients stopped making any log entries at different times, roughly 10 mins in between, but the pattern are the same: once they received active tasks that are all stateless, and then the hb thread reported error right after the assignment, the tasks only completes initialization but never completes restoration (note that since they are stateless, they should normally transit to running right after the next iteration). Without lower-level logs I cannot tell for sure, but since the thread-process-rate of those threads indeed drops to zero, my suspicion is that the hb error did not set the consumer's to re-join, and the poll call maybe blocked on pollForHeartbeat since the state is set to `UNJOINED` which cause its poll timeout to MAX_VALUES.

That being said, I cannot comfortable say with 100 percent confidence that this is "the" root cause of what we observed in the soaking cluster, but at least it is "an" issue that I can discover from the logs.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
guozhangwang: @chia7712 @dajac I think this bug is only introduced recently in this PR: https://github.com/apache/kafka/pull/8834. If your tests is executed before it then maybe there are other issues yet to be discovered.

dajac: That test became flaky quite recently. Time wise, that could coincides. I will try to run the test without https://github.com/apache/kafka/pull/8834 to check tomorrow.

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
```diff
@@ -1934,12 +1933,20 @@ public void testReturnRecordsDuringRebalance() {
...
         client.respondFrom(fetchResponse(fetches1), node);
 
-        records = consumer.poll(Duration.ZERO);
+        // now complete teh rebalance
+        client.respondFrom(syncGroupResponse(Arrays.asList(tp0, t3p0), Errors.NONE), coordinator);
+
+        AtomicInteger count = new AtomicInteger(0);
+        TestUtils.waitForCondition(() -> {
+            ConsumerRecords<String, String> recs = consumer.poll(Duration.ofMillis(100L));
+            System.out.println("count " + count.addAndGet(recs.count()));
```
abbccdda: Could we remove this print statement?

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
```diff
@@ -1934,12 +1933,20 @@ public void testReturnRecordsDuringRebalance() {
         fetches1.put(tp0, new FetchInfo(3, 1));
         client.respondFrom(fetchResponse(fetches1), node);
 
-        records = consumer.poll(Duration.ZERO);
+        // now complete teh rebalance
```
abbccdda: s/teh/the

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
```diff
@@ -1948,10 +1955,14 @@ public void testReturnRecordsDuringRebalance() {
         fetches1.put(t3p0, new FetchInfo(0, 100));
         client.respondFrom(fetchResponse(fetches1), node);
 
-        records = consumer.poll(Duration.ZERO);
+        count.set(0);
+        TestUtils.waitForCondition(() -> {
+            ConsumerRecords<String, String> recs = consumer.poll(Duration.ofMillis(100L));
+            System.out.println("count2 " + count.addAndGet(recs.count()));
```
abbccdda: Same here

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
abbccdda: It seems that we would rejoin on FENCED_INSTANCE_ID as well, is that intentional? https://github.com/apache/kafka/blob/bd462df20321ff5b75a7e3eae70634268582d90b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java#L1106

guozhangwang: That is fine since in the outer caller `FENCED_INSTANCE_ID` would be treated as fatal, in line 1386.

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
guozhangwang: Ah, thanks!

PR #9359, opened 2020-10-01
kafka-10273 Connect Converters should produce actionable error messages

shadikajevand:
I have written actionable error messages for the JSON-converter 

@rhauch Hi, please review if this is acceptable, if anything needs to be changed and if I need to do the same for the rest of the converters.


rhauch (commented):
Thanks, @shadikajevand. I hope the following suggestions help:

First, anything that is general or converter-independent should be added in the wrapper messages thrown in `convertKey(...)` and `convertValue(...)` in `WorkerSinkTask` that provide much more context. Unfortunately, we don't have those in `WorkerSourceTask` and probably should, because the context provided by those methods help provide the 
we don't know that this converter is being used for keys or values, and will handle all converters. All of the suggested changes here are good, but they apply to all key or value converters. It may be better to enhance the messages for the `WorkerSinkTask` exceptions (and the to-be-added `WorkerSourceTask` exceptions) to mention checking the configurations.

Second, there are some cases that we can/may want to check in `JsonConverter`, and [KAFKA-10273](https://issues.apache.org/jira/browse/KAFKA-10273) gives some more concrete suggestions for the JSON converter, including:
1. what is the error that happens when `schemas.enable=true` but the record doesn't contain the `schema` and `payload` fields? The actionable message might suggest to change `value.converter.schemas.enable=false` (or `key.converter.schemas.enable=false` in the connector configuration or Connect worker configuration.
2. Is there an error when `schemas.enable=false` but the record contains the `schema` and `payload` fields? If so, then an actionable message might suggest to change `value.converter.schemas.enable=true` (or `key.converter.schemas.enable=true` in the connector configuration or Connect worker configuration.

There are others, too. These may require looking at the exceptions (type and message) and maybe even a bit of the parsed JSON document (if there is one) to determine the case and provide a case-specific actionable error message. Only converter-specific messages/suggestions should be in the converter; generic suggestions should go into the `convertKey(...)` and `convertValue(...)` in the `WorkerSinkTask` and `WorkerSourceTask` classes.

And, in the JsonConverter if we need to suggest changes to configs (e.g., "set `value.converter.schemas.enable=true` ...), then we should use the correct prefix based upon `config.type()`.

shadikajevand:
Thanks @rhauch I will look into it :)

shadikajevand:
Hi @rhauch I am working on convertKey(...) and convertValue(...) methods for WorkerSourceTask. I'm throwing a DataException in these two methods when an exception occures. In the RetryWithToleranceOperator class it states that:

If it fails, this class does one of these two things: (1) if the failure occurred due to a tolerable exception, then
 * set appropriate error reason in the {@link ProcessingContext} and return null, or (2) if the exception is not tolerated,
 * then it is wrapped into a ConnectException and rethrown to the caller.

My question is, will this exception be tolerated? I am assuming that it is not a tolarable exception. In this case what is the best way of handling this in the RetryWithToleranceOperator?

shadikajevand:
@rhauch Hi I have updated the PR. Please review when you have time. Thanks!

connect/api/src/main/java/org/apache/kafka/connect/storage/StringConverter.java
```diff
@@ -79,7 +79,7 @@ public void configure(Map<String, ?> configs, boolean isKey) {
         try {
             return serializer.serialize(topic, value == null ? null : value.toString());
         } catch (SerializationException e) {
-            throw new DataException("Failed to serialize to a string: ", e);
+            throw new DataException(this.getClass() + " Failed to serialize to a string: ", e);
```
rhauch: Using `Class.toString()` method here will output a message like:
```
class org.apache.kafka.connect.storage.StringConverter Failed to serialize to a string: ...
```
First, beginning this exception message with `class` it not capitalized even though it's the start of a sentence, and `Failed` should not be capitalized in the middle of a sentence. Plus, I'm unconvinced that including the package is helpful.

But more importantly, I don't think we should change these exception messages in any of the converters, because we'd have to change them in all converter implementations to remain consistent. While we could do that for AK-provided converters, there are lots of other converters that may have modeled their error messages after these, and would then become inconsistent. Plus, why change all of them when we're catching the exception in the WorkerSinkTask and WorkerSourceTask? Wouldn't it make more sense to change it there so that we don't have to change any converters?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java
```diff
@@ -531,19 +532,25 @@ private SchemaAndValue convertKey(ConsumerRecord<byte[], byte[]> msg) {
...
         } catch (Exception e) {
-            log.error("{} Error converting message key in topic '{}' partition {} at offset {} and timestamp {}: {}",
-                    this, msg.topic(), msg.partition(), msg.offset(), msg.timestamp(), e.getMessage(), e);
-            throw e;
+            String errorMessage = "Error while deserializing the key for record in topic %s, partition %s, timestamp %s, and at offset %s. " +
+                    "Check the key.converter and key.converter.* settings in the connector configuration, " +
+                    "and ensure that the converter matches the converter/serializer used by the application that produced this record. " +
+                    "Underlying converter error: %s";
+            log.error(String.format(errorMessage, msg.topic(), msg.partition(), msg.timestamp(), msg.offset(), e.getMessage()), e);
+            throw new DataException(String.format(errorMessage, msg.topic(), msg.partition(), msg.offset(), msg.timestamp(), e.getMessage()), e);
```
rhauch: Why not create a single formatted string and use it for both the log and exception messages? Also, would it be useful to have the short name of the converter class in the message as well? Maybe something like:
``````suggestion
            String errorMessage = String.format("Error while deserializing the key for record in topic '%s', partition %s (timestamp %s) at offset %s. " +
                    "Check the 'key.converter' and 'key.converter.*' settings in the connector configuration, " +
                    "and ensure that the converter matches the converter/serializer used by the application that produced this record. " +
                    "Underlying converter error: %s", keyConverter.getClass().getSimpleName(), msg.topic(), msg.partition(), msg.timestamp(), msg.offset(), e.getMessage());
            log.error(errorMessage, e);
            throw new DataException(errorMessage, e);
```
Note the single quotes and slight rewording.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java
```diff
@@ -531,19 +532,25 @@ private SchemaAndValue convertKey(ConsumerRecord<byte[], byte[]> msg) {
...
         } catch (Exception e) {
-            log.error("{} Error converting message value in topic '{}' partition {} at offset {} and timestamp {}: {}",
-                    this, msg.topic(), msg.partition(), msg.offset(), msg.timestamp(), e.getMessage(), e);
-            throw e;
+            String errorMessage = "Error while deserializing the value for record in topic %s, partition %s, timestamp %s, and at offset %s. " +
+                    "Check the value.converter and value.converter.* settings in the connector configuration, " +
+                    "and ensure that the converter matches the converter/serializer used by the application that produced this record. " +
+                    "Underlying converter error: %s";
+            log.error(String.format(errorMessage, msg.topic(), msg.partition(), msg.timestamp(), msg.offset(), e.getMessage()), e);
+            throw new DataException(String.format(errorMessage, msg.topic(), msg.partition(), msg.offset(), msg.timestamp(), e.getMessage()), e);
```
rhauch: Similar to above suggestion.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
```diff
@@ -602,6 +602,30 @@ private synchronized void finishSuccessfulFlush() {
...
 
+    private byte[] convertKey(SourceRecord record, RecordHeaders headers) {
+        try {
+            return keyConverter.fromConnectData(record.topic(), headers, record.keySchema(), record.key());
+        } catch (Exception e) {
+            String errorMessage = "%s Error while serializing the key for a source record to topic: %s. Check the key.converter and key.converter.* " +
+                    "settings in the connector configuration, or in the worker configuration if the connector is inheriting the connector configuration. " +
+                    "Underlying converter error: %s";
+            log.error(String.format(errorMessage, this, record.topic(), e.getMessage()), e);
+            throw new RetriableException(String.format(errorMessage, this, record.topic(), e.getMessage()), e);
```
rhauch: Similar to above suggestion.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
```diff
@@ -602,6 +602,30 @@ private synchronized void finishSuccessfulFlush() {
...
+
+    private byte[] convertValue(SourceRecord record, RecordHeaders headers) {
+        try {
+            return valueConverter.fromConnectData(record.topic(), headers, record.valueSchema(), record.value());
+        } catch (Exception e) {
+            String errorMessage = "%s Error while serializing the value for a source record to topic: %s. Check the value.converter and value.converter.* " +
+                    "settings in the connector configuration, or in the worker configuration if the connector is inheriting the connector configuration. " +
+                    "Underlying converter error: %s";
+            log.error(String.format(errorMessage, this, record.topic(), e.getMessage()), e);
+            throw new RetriableException(String.format(errorMessage, this, record.topic(), e.getMessage()), e);
```
rhauch: Similar to above suggestion.

rhauch (commented):
Thanks, @shadikajevand. This is going to be quite useful, but I think we can further improve the exception message a bit more, and focus the changes where it helps the most.

shadikajevand:
@rhauch Thanks, I really appreciate your help. I have updated the code now. Please review when you have time.

PR #9364, opened 2020-10-02
KAFKA-10471 Mark broker crash during log loading as unclean shutdown

RamanVerma:
LogManager writes a clean shutdown file when the broker shuts down. The
presence of this file indicates that the broker had a clean shutdown and
log recovery is not needed upon the next boot up.

Earlier, LogManager would check for this file at the start of log loading workflow,
and delete it after the log has been loaded. If the broker were to crash
while loading logs, the file would not be deleted and mislead LogManager when it
tries to load logs upon next boot up. Hence, a crash during log loading
will not be considered a hard reset of broker.

As part of this fix, we delete the clean shutdown file as soon as we
look it up, at the start of log loading workflow. Thereafter, we maintain a boolean
flag to indicate if broker underwent clean shutdown or not. So, if the
broker were to crash while logs are being loaded, LogManager will be
able to detect this as a hard reset.

comment to ping reviewers.*

### DIFF
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -244,7 +244,8 @@ class Log(@volatile private var _dir: File,
           val producerIdExpirationCheckIntervalMs: Int,
           val topicPartition: TopicPartition,
           val producerStateManager: ProducerStateManager,
-          logDirFailureChannel: LogDirFailureChannel) extends Logging with KafkaMetricsGroup {
+          logDirFailureChannel: LogDirFailureChannel,
+          val hadCleanShutdown: Boolean = true) extends Logging with KafkaMetricsGroup {
 
   import kafka.log.Log._
 
@@ -318,7 +319,7 @@ class Log(@volatile private var _dir: File,
     // during log recovery may have deleted some files without the Log.producerStateManager instance witnessing the
     // deletion.
     producerStateManager.removeStraySnapshots(segments.values().asScala.map(_.baseOffset).toSeq)
-    loadProducerState(logEndOffset, reloadFromCleanShutdown = hasCleanShutdownFile)
+    loadProducerState(logEndOffset, reloadFromCleanShutdown = hadCleanShutdown)
   }
 
   def dir: File = _dir
@@ -804,9 +805,9 @@ class Log(@volatile private var _dir: File,
    * logs are loaded.
    * @throws LogSegmentOffsetOverflowException if we encountered a legacy segment with offset overflow
    */
-  private def recoverLog(): Long = {
+  private[log] def recoverLog(): Long = {
     // if we have the clean shutdown marker, skip recovery
-    if (!hasCleanShutdownFile) {
+    if (!hadCleanShutdown) {
       // okay we need to actually recover this log
       val unflushed = logSegments(this.recoveryPoint, Long.MaxValue).iterator
       var truncated = false
@@ -974,11 +975,6 @@ class Log(@volatile private var _dir: File,
       producerId -> lastRecord
     }
   }
-
-  /**
-   * Check if we have the "clean shutdown" file
-   */
-  private def hasCleanShutdownFile: Boolean = new File(dir.getParentFile, CleanShutdownFile).exists()
 
   /**
    * The number of segments in the log.
@@ -2487,11 +2483,12 @@ object Log {
             time: Time = Time.SYSTEM,
             maxProducerIdExpirationMs: Int,
             producerIdExpirationCheckIntervalMs: Int,
-            logDirFailureChannel: LogDirFailureChannel): Log = {
+            logDirFailureChannel: LogDirFailureChannel,
+            lastShutdownClean: Boolean = true): Log = {
     val topicPartition = Log.parseTopicPartitionName(dir)
     val producerStateManager = new ProducerStateManager(topicPartition, dir, maxProducerIdExpirationMs)
     new Log(dir, config, logStartOffset, recoveryPoint, scheduler, brokerTopicStats, time, maxProducerIdExpirationMs,
-      producerIdExpirationCheckIntervalMs, topicPartition, producerStateManager, logDirFailureChannel)
+      producerIdExpirationCheckIntervalMs, topicPartition, producerStateManager, logDirFailureChannel, lastShutdownClean)
   }
 
   /**
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -231,6 +231,8 @@ case object SegmentDeletion extends LogStartOffsetIncrementReason {
  * @param time The time instance used for checking the clock
  * @param maxProducerIdExpirationMs The maximum amount of time to wait before a producer id is considered expired
  * @param producerIdExpirationCheckIntervalMs How often to check for producer ids which need to be expired
+ * @param hadCleanShutdown boolean flag to indicate if the Log had a clean/graceful shutdown last time. true means
+ *                         clean shutdown whereas false means a crash.
  */
 @threadsafe
 class Log(@volatile private var _dir: File,
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -247,7 +247,7 @@ class Log(@volatile private var _dir: File,
           val topicPartition: TopicPartition,
           val producerStateManager: ProducerStateManager,
           logDirFailureChannel: LogDirFailureChannel,
-          val hadCleanShutdown: Boolean = true) extends Logging with KafkaMetricsGroup {
+          private val hadCleanShutdown: Boolean = true) extends Logging with KafkaMetricsGroup {
 
   import kafka.log.Log._
 
--- a/core/src/main/scala/kafka/log/LogManager.scala
+++ b/core/src/main/scala/kafka/log/LogManager.scala
@@ -320,13 +320,7 @@ class LogManager(logDirs: Seq[File],
           info(s"Skipping recovery for all logs in $logDirAbsolutePath since clean shutdown file was found")
           // Cache the clean shutdown status and use that for rest of log loading workflow. Delete the CleanShutdownFile
           // so that if broker crashes while loading the log, it is considered hard shutdown during the next boot up. KAFKA-10471
-          try {
-            cleanShutdownFile.delete()
-          } catch {
-            case e: IOException =>
-              error(s"Error while deleting the clean shutdown file $cleanShutdownFile", e)
-              throw e
-          }
+          cleanShutdownFile.delete()
           hadCleanShutdown = true
         } else {
           // log recovery itself is being performed by `Log` class during initialization
--- a/core/src/main/scala/kafka/log/LogManager.scala
+++ b/core/src/main/scala/kafka/log/LogManager.scala
@@ -385,7 +385,7 @@ class LogManager(logDirs: Seq[File],
       }
 
       offlineDirs.foreach { case (dir, e) =>
-        logDirFailureChannel.maybeAddOfflineLogDir(dir, s"Error while deleting the clean shutdown file in dir $dir", e)
+        logDirFailureChannel.maybeAddOfflineLogDir(dir, s"Error while loading log dir $dir", e)
       }
     } catch {
       case e: ExecutionException =>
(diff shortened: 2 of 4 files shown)


core/src/main/scala/kafka/log/Log.scala
```diff
@@ -244,7 +244,8 @@ class Log(@volatile private var _dir: File,
           val producerIdExpirationCheckIntervalMs: Int,
           val topicPartition: TopicPartition,
           val producerStateManager: ProducerStateManager,
-          logDirFailureChannel: LogDirFailureChannel) extends Logging with KafkaMetricsGroup {
+          logDirFailureChannel: LogDirFailureChannel,
+          val hadCleanShutdown: Boolean = true) extends Logging with KafkaMetricsGroup {
```
junrao: Could we add the new param to the javadoc above?

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -4447,9 +4504,10 @@ class LogTest {
 
   private def recoverAndCheck(config: LogConfig,
                               expectedKeys: Iterable[Long],
-                              expectDeletedFiles: Boolean = true): Log = {
+                              expectDeletedFiles: Boolean = true,
```
junrao: It seems that all callers set expectDeletedFiles to false. So, do we need this param?

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -4429,9 +4485,10 @@ class LogTest {
                         scheduler: Scheduler = mockTime.scheduler,
                         time: Time = mockTime,
                         maxProducerIdExpirationMs: Int = 60 * 60 * 1000,
-                        producerIdExpirationCheckIntervalMs: Int = LogManager.ProducerIdExpirationCheckIntervalMs): Log = {
+                        producerIdExpirationCheckIntervalMs: Int = LogManager.ProducerIdExpirationCheckIntervalMs,
+                        lastShutdownClean: Boolean = true): Log = {
```
junrao, line 4617: The callers of createLog() in line 652 and 2205 seem to need lastShutdownClean to be false.

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -2882,11 +2953,8 @@ class LogTest {
     records.foreach(segment.append _)
     segment.close()
 
-    // Create clean shutdown file so that we do not split during the load
-    createCleanShutdownFile()
-
     val logConfig = LogTest.createLogConfig(indexIntervalBytes = 1, fileDeleteDelayMs = 1000)
-    val log = createLog(logDir, logConfig, recoveryPoint = Long.MaxValue)
+    val log = createLog(logDir, logConfig, recoveryPoint = Long.MaxValue, lastShutdownClean = true)
```
junrao: The change is unnecessary.

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -3073,9 +3139,8 @@ class LogTest {
     // check if recovery was attempted. Even if the recovery point is 0L, recovery should not be attempted as the
     // clean shutdown file exists.
     recoveryPoint = log.logEndOffset
-    log = createLog(logDir, logConfig)
+    log = createLog(logDir, logConfig, lastShutdownClean = true)
```
junrao: The change is unnecessary.

junrao (commented):
@RamanVerma : Thanks for the PR. A few comments below.

core/src/main/scala/kafka/log/Log.scala
RamanVerma: Done

core/src/test/scala/unit/kafka/log/LogTest.scala
RamanVerma: I think you meant the `hadCleanShutdown` parameter I added. I did not add the `expectDeletedFiles` parameter.
I will remove the new parameter and add a comment here to indicate that the method always assumes we had a hard reset.

core/src/test/scala/unit/kafka/log/LogTest.scala
RamanVerma: Removed

core/src/test/scala/unit/kafka/log/LogTest.scala
RamanVerma: Removed.

junrao:
@RamanVerma : Thanks for addressing the comments. Did you push the new commits?

core/src/test/scala/unit/kafka/log/LogTest.scala
RamanVerma: Done

core/src/test/scala/unit/kafka/log/LogTest.scala
RamanVerma: Done

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -298,26 +300,38 @@ class LogManager(logDirs: Seq[File],
...
 
         val cleanShutdownFile = new File(dir, Log.CleanShutdownFile)
         if (cleanShutdownFile.exists) {
           info(s"Skipping recovery for all logs in $logDirAbsolutePath since clean shutdown file was found")
+          // Cache the clean shutdown status and use that for rest of log loading workflow. Delete the CleanShutdownFile
+          // so that if broker crashes while loading the log, it is considered hard shutdown during the next boot up. KAFKA-10471
+          try {
+            cleanShutdownFile.delete()
+          } catch {
+            case e: IOException =>
```
junrao: This is an existing issue, but cleanShutdownFile.delete() doesn't seem to throw IOException.

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -1257,7 +1328,7 @@ class LogTest {
     log.close()
 
     // After reloading log, producer state should not be regenerated
-    val reloadedLog = createLog(logDir, logConfig, logStartOffset = 1L)
+    val reloadedLog = createLog(logDir, logConfig, logStartOffset = 1L, lastShutdownClean = false)
```
junrao, line 1460: Hmm, it seems that this tests expects a clean shutdown.

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -2131,12 +2202,12 @@ class LogTest {
       assertEquals("Should have same number of time index entries as before.", numTimeIndexEntries, log.activeSegment.timeIndex.entries)
     }
 
-    log = createLog(logDir, logConfig, recoveryPoint = lastOffset)
+    log = createLog(logDir, logConfig, recoveryPoint = lastOffset, lastShutdownClean = false)
```
junrao, line 2331: It seems that this expects a clean shutdown.

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -3623,7 +3690,7 @@ class LogTest {
     log.close()
 
     // reopen the log and recover from the beginning
-    val recoveredLog = createLog(logDir, LogConfig())
+    val recoveredLog = createLog(logDir, LogConfig(), lastShutdownClean = false)
```
junrao, line 3819: It seem the createLog() call on line 3976 inside testRecoverOnlyLastSegment() needs to have lastShutdownClean = false.

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -4445,11 +4504,9 @@ class LogTest {
     (log, segmentWithOverflow)
   }
 
-  private def recoverAndCheck(config: LogConfig,
-                              expectedKeys: Iterable[Long],
-                              expectDeletedFiles: Boolean = true): Log = {
-    LogTest.recoverAndCheck(logDir, config, expectedKeys, brokerTopicStats, mockTime, mockTime.scheduler,
-      expectDeletedFiles)
+  private def recoverAndCheck(config: LogConfig, expectedKeys: Iterable[Long], expectDeletedFiles: Boolean = true) = {
```
junrao: This is an existing problem. Since no callers are explicitly setting expectDeletedFiles, could we just remove this param?

junrao (commented):
@RamanVerma : Thanks for the updated PR. A few more comments. Also, could you rebase?

core/src/main/scala/kafka/log/LogManager.scala
RamanVerma: `java.nio.file.Files` API throws `IOException` but not the one used here `java.io.File`. Will remove this exception

core/src/test/scala/unit/kafka/log/LogTest.scala
RamanVerma: Jun, this test was not creating a clean shut down file before opening the log again. So, it would have gone through the recovery path. Hence, I have set `lastShutdownClean` parameter to `false`. Similarly, for line 2210.

core/src/test/scala/unit/kafka/log/LogTest.scala
RamanVerma: Jun, this test was not creating a clean shut down file before opening the log again. So, it would have been going through log recovery code path earlier as well.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -298,26 +300,32 @@ class LogManager(logDirs: Seq[File],
...
         if (cleanShutdownFile.exists) {
           info(s"Skipping recovery for all logs in $logDirAbsolutePath since clean shutdown file was found")
+          // Cache the clean shutdown status and use that for rest of log loading workflow. Delete the CleanShutdownFile
+          // so that if broker crashes while loading the log, it is considered hard shutdown during the next boot up. KAFKA-10471
+          cleanShutdownFile.delete()
+          hadCleanShutdown = true
         } else {
           // log recovery itself is being performed by `Log` class during initialization
           info(s"Attempting recovery for all logs in $logDirAbsolutePath since no clean shutdown file was found")
+          hadCleanShutdown = false
```
junrao: This line seems unnecessary since hadCleanShutdown is initialized to false.

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -978,10 +1051,10 @@ class LogTest {
       producerIdExpirationCheckIntervalMs = 30000,
       topicPartition = Log.parseTopicPartitionName(logDir),
       producerStateManager = stateManager,
-      logDirFailureChannel = null)
+      logDirFailureChannel = null,
+      hadCleanShutdown = true)
```
junrao: It seems we don't need to set hadCleanShutdown to true.

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -1021,10 +1092,10 @@ class LogTest {
...
```
junrao: It seems we don't need to set hadCleanShutdown to true.

junrao (commented):
@RamanVerma : Thanks for the updated PR. Just a few minor comments below.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -244,7 +246,8 @@ class Log(@volatile private var _dir: File,
...
```
junrao: It seems that hadCleanShutdown doesn't need to be exposed as a public val?

junrao (commented):
@RamanVerma : Thanks for the updated PR. Just one minor comment.

junrao (approved):
@RamanVerma : Thanks for the latest update. LGTM

junrao:
@RamanVerma  Was the test failure kafka.admin.ReassignPartitionsIntegrationTest.testLogDirReassignment in JDK 15 related to this PR?

RamanVerma:
@junrao I don't think the JDK 15 failure is related to my change.
- We see that tests pass till JDK 11
- I have run the test in my IDE with AdoptOpenJDK15 and it passed
- The test fails because the reassignment has not completed within timeout window. As part of this change, I have not changed the way cleanShutdownFile is created, or anything related to the reassignment of replicas. We just changed when this file is to be deleted.

PR #9367, opened 2020-10-02
KAFKA-10570 Rename JMXReporter configs for KIP-629

xvrl:
* rename whitelist/blacklist to include/exclude
* add utility methods to translate deprecated configs

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/utils/ConfigUtils.java
+++ b/clients/src/main/java/org/apache/kafka/common/utils/ConfigUtils.java
@@ -20,6 +20,7 @@ package org.apache.kafka.common.utils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


xvrl:
jdk8 test failures appear unrelated

clients/src/main/java/org/apache/kafka/common/utils/ConfigUtils.java
```diff
@@ -0,0 +1,92 @@
...
+        Map<String, T> newConfigs = configs.entrySet().stream()
+            .filter(e -> !aliasSet.contains(e.getKey()))
+            // filter out null values
+            .filter(e -> Objects.nonNull(e.getValue()))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+        Stream.of(aliasGroups).forEachOrdered(aliasGroup -> {
+            String target = aliasGroup[0];
+
+            List<String> deprecated = Stream.of(aliasGroup)
```
gwenshap: We may be able to save all this hassle by defining alias group as a map of `target` to a list of `deprecated` configs? We defined this as a 2-dim array but we always convert it to lists...

clients/src/main/java/org/apache/kafka/common/utils/ConfigUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+            String aliasString = String.join(", ", deprecated);
+
+            if (configs.containsKey(target)) {
+                // Ignore the deprecated key(s) because the actual key was set.
+                log.error(target + " was configured, as well as the deprecated alias(es) " +
+                          aliasString + ".  Using the value of " + target);
+                newConfigs.put(target, configs.get(target));
+            } else if (deprecated.size() > 1) {
+                log.error("The configuration keys " + aliasString + " are deprecated and may be " +
```
gwenshap, line 102: I like the extra-detailed error messages, thank you.

gwenshap (changes requested):
One minor comment regarding the deprecated data struct. 
Other than that, its great. Tests are awesome.

clients/src/main/java/org/apache/kafka/common/utils/ConfigUtils.java
xvrl: you have to thank @cmccabe for that ;)

clients/src/main/java/org/apache/kafka/common/utils/ConfigUtils.java
xvrl: I update the code to take a Map, and left a convenience method to take arrays, since it makes writing the translations easier.

gwenshap (approved):
LGTM. Thank you Xavier.

PR #9368, opened 2020-10-03
KAFKA-9274: Add timeout handling for state restore and StandbyTasks

mjsax:
- part of KIP-572

If a `TimeoutException` happens during restore of active tasks, or updating standby tasks, we need to trigger `task.timeout.ms` timeout.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
@@ -39,7 +39,10 @@ public abstract class AbstractTask implements Task {
 
     private Task.State state = CREATED;
     private long deadlineMs = NO_DEADLINE;
+
     protected Set<TopicPartition> inputPartitions;
+    protected Logger log;
+    protected String logPrefix;
 
     /**
      * If the checkpoint has not been loaded from the file yet (null), then we should not overwrite the checkpoint;
@@ -147,12 +150,9 @@ public abstract class AbstractTask implements Task {
         topology.updateSourceTopics(nodeToSourceTopics);
     }
 
-    /**
-     * @throws TimeoutException if {@code currentWallClockMs > task-timeout-deadline}
-     */
-    void maybeInitTaskTimeoutOrThrow(final long currentWallClockMs,
-                                     final Exception cause,
-                                     final Logger log) {
+    @Override
+    public void maybeInitTaskTimeoutOrThrow(final long currentWallClockMs,
+                                            final Exception cause) {
         if (deadlineMs == NO_DEADLINE) {
             deadlineMs = currentWallClockMs + taskTimeoutMs;
         } else if (currentWallClockMs > deadlineMs) {
@@ -187,7 +187,8 @@ public abstract class AbstractTask implements Task {
 
     }
 
-    void clearTaskTimeout(final Logger log) {
+    @Override
+    public void clearTaskTimeout() {
         if (deadlineMs != NO_DEADLINE) {
             log.debug("Clearing task timeout.");
             deadlineMs = NO_DEADLINE;
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
@@ -18,6 +18,7 @@ package org.apache.kafka.streams.processor.internals;
 
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.TimeoutException;
+import org.apache.kafka.common.utils.LogContext;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.errors.StreamsException;
 import org.apache.kafka.streams.errors.TaskMigratedException;
@@ -41,8 +42,9 @@ public abstract class AbstractTask implements Task {
     private long deadlineMs = NO_DEADLINE;
 
     protected Set<TopicPartition> inputPartitions;
-    protected Logger log;
-    protected String logPrefix;
+    protected final Logger log;
+    protected final LogContext logContext;
+    protected final String logPrefix;
 
     /**
      * If the checkpoint has not been loaded from the file yet (null), then we should not overwrite the checkpoint;
@@ -63,13 +65,20 @@ public abstract class AbstractTask implements Task {
                  final StateDirectory stateDirectory,
                  final ProcessorStateManager stateMgr,
                  final Set<TopicPartition> inputPartitions,
-                 final long taskTimeoutMs) {
+                 final long taskTimeoutMs,
+                 final String logPrefix,
+                 final Class<? extends AbstractTask> clazz) {
         this.id = id;
         this.stateMgr = stateMgr;
         this.topology = topology;
         this.inputPartitions = inputPartitions;
         this.stateDirectory = stateDirectory;
         this.taskTimeoutMs = taskTimeoutMs;
+
+        final String threadIdPrefix = String.format("stream-thread [%s] ", Thread.currentThread().getName());
+        this.logPrefix = threadIdPrefix + String.format("%s [%s] ", logPrefix, id);
+        this.logContext = new LogContext(logPrefix);
+        this.log = logContext.logger(clazz);
     }
 
     /**
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
@@ -147,9 +147,12 @@ public abstract class AbstractTask implements Task {
         topology.updateSourceTopics(nodeToSourceTopics);
     }
 
+    /**
+     * @throws TimeoutException if {@code currentWallClockMs > task-timeout-deadline}
+     */
     void maybeInitTaskTimeoutOrThrow(final long currentWallClockMs,
-                                     final TimeoutException timeoutException,
-                                     final Logger log) throws StreamsException {
+                                     final Exception cause,
+                                     final Logger log) {
         if (deadlineMs == NO_DEADLINE) {
             deadlineMs = currentWallClockMs + taskTimeoutMs;
         } else if (currentWallClockMs > deadlineMs) {
@@ -160,18 +163,20 @@ public abstract class AbstractTask implements Task {
                 StreamsConfig.TASK_TIMEOUT_MS_CONFIG
             );
 
-            if (timeoutException != null) {
-                throw new TimeoutException(errorMessage, timeoutException);
+            if (cause != null) {
+                throw new TimeoutException(errorMessage, cause);
             } else {
                 throw new TimeoutException(errorMessage);
             }
         }
 
-        if (timeoutException != null) {
+        if (cause != null) {
             log.debug(
-                "Timeout exception. Remaining time to deadline {}; retrying.",
-                deadlineMs - currentWallClockMs,
-                timeoutException
+                String.format(
+                    "Task did not make progress. Remaining time to deadline %d; retrying.",
+                    deadlineMs - currentWallClockMs
+                ),
+                cause
             );
         } else {
             log.debug(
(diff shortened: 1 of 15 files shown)


streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
```diff
@@ -694,7 +694,7 @@
                     CommonClientConfigs.SECURITY_PROTOCOL_DOC)
             .define(TASK_TIMEOUT_MS_CONFIG,
                     Type.LONG,
-                    Duration.ofSeconds(5L).toMillis(),
+                    Duration.ofMinutes(5L).toMillis(),
```
mjsax: We got this wrong in the original PR...

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
```diff
@@ -137,4 +146,48 @@ public void update(final Set<TopicPartition> topicPartitions, final Map<String,
         this.inputPartitions = topicPartitions;
         topology.updateSourceTopics(nodeToSourceTopics);
     }
+
+    @Override
+    public void maybeInitTaskTimeoutOrThrow(final long currentWallClockMs,
+                                            final TimeoutException timeoutException,
+                                            final Logger log) throws StreamsException {
```
mjsax: We want to use the logger of `StreamTask` or `StandbyTask` to get the proper log prefix; thus, we just pass it into this method.

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
```diff
@@ -137,4 +146,48 @@ public void update(final Set<TopicPartition> topicPartitions, final Map<String,
...
+            deadlineMs = currentWallClockMs + taskTimeoutMs;
+        } else if (currentWallClockMs > deadlineMs) {
+            final String errorMessage = String.format(
+                "Task %s did not make progress within %d ms. Adjust `%s` if needed.",
+                id,
+                currentWallClockMs - deadlineMs + taskTimeoutMs,
+                StreamsConfig.TASK_TIMEOUT_MS_CONFIG
+            );
+
+            if (timeoutException != null) {
```
mjsax: I added this, as for `poll()` we would not get a `TimeoutException` atm (even if I hold off to tackle poll() already)

We could also change this and assume that `timeoutException` is never `null`.

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
```diff
@@ -137,4 +146,48 @@ public void update(final Set<TopicPartition> topicPartitions, final Map<String,
...
+                "Task did not make progress. Remaining time to deadline {}; retrying.",
+                deadlineMs - currentWallClockMs
+            );
+        }
+
+    }
+
+    @Override
+    public void clearTaskTimeout(final Logger log) {
+        if (deadlineMs != NO_DEADLINE) {
```
mjsax, line 201: As we call this reset logic "blindly" we put this guard to avoid spamming the logs.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StoreChangelogReader.java
```diff
@@ -431,6 +432,9 @@ public void restore() {
                 // in order to make sure we call the main consumer#poll in time.
                 // TODO: once we move ChangelogReader to a separate thread this may no longer be a concern
                 polledRecords = restoreConsumer.poll(state == ChangelogReaderState.STANDBY_UPDATING ? Duration.ZERO : pollTime);
+
+                // TODO (?) If we cannot fetch records during restore, should we trigger `task.timeout.ms` ?
+                // TODO (?) If we cannot fetch records for standby task, should we trigger `task.timeout.ms` ?
```
mjsax, line 437: For the global-thread, we also consider the case that `poll()` returns nothing. For the global-thread it's a little simper though, as we restore on a per-partition basis.

For the `StoreChangelogReader` it's more complicated:
 - for StandbyTasks, there might be nothing to be restored and thus getting no records might be fine (however, the "metadata" is not easily available as it's encapsulated in other classes...
 - for active restoring tasks, it might be simpler...

Was holding off, as I am not sure how important this case is, with regard to robustness. We would not crash for this case. I would still like to get this done, but not necessarily for `2.7` release though.

streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
vvcephei: Oops! Good catch.

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
vvcephei: Sounds good. An alternative is to add an `abstract Logger log()` to AbstractTask's interface, which would make it clearer that the logger is still going to have the appropriate class name.

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
```diff
@@ -137,4 +146,48 @@ public void update(final Set<TopicPartition> topicPartitions, final Map<String,
...
+                throw new TimeoutException(errorMessage);
+            }
+        }
+
+        if (timeoutException != null) {
+            log.debug(
+                "Timeout exception. Remaining time to deadline {}; retrying.",
+                deadlineMs - currentWallClockMs,
+                timeoutException
+            );
```
vvcephei: This is the wrong format for this log message. The exception won't be logged. You have to format the string first:
```suggestion
            log.debug(
                String.format("Timeout exception. Remaining time to deadline %d; retrying.",
                deadlineMs - currentWallClockMs),
                timeoutException
            );
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/StoreChangelogReader.java
vvcephei: Thanks. I agree we can afford to leave this for future improvements.

It does seem like we should have some kind of improvement in the future, though. Having a restore or standby-update fail indefinitely would be just as damaging to an application's robustness as having the main consumer fail indefinitely.

Perhaps we can make some improvements to the Consumer first, though, so that we don't have to do so much guesswork to distinguish between "no records" and "no fetch".

streams/src/main/java/org/apache/kafka/streams/processor/internals/Task.java
```diff
@@ -205,4 +207,9 @@ default boolean maybePunctuateSystemTime() {
         return false;
     }
 
+    void maybeInitTaskTimeoutOrThrow(final long currentWallClockMs,
+                                     final TimeoutException timeoutException,
+                                     final Logger log) throws StreamsException;
```
vvcephei: StreamsException is unchecked, right?

It's better to document unchecked exceptions in the `@throws` javadoc tag. The `throws` keyword is for telling the compiler that you want callers instead of yourself to handle a _checked_ exception. I honestly have no idea why the java team chose to say "this is poor style" instead of just making it a compiler error, but that's the rationale. https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#throwstag

vvcephei (commented):
Hey @mjsax , thanks for the PR! I made a partial pass, and had a couple of minor thoughts.

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
```diff
@@ -148,7 +148,7 @@ public void update(final Set<TopicPartition> topicPartitions, final Map<String,
     }
 
     void maybeInitTaskTimeoutOrThrow(final long currentWallClockMs,
-                                     final TimeoutException timeoutException,
+                                     final Exception cause,
```
mjsax: While the KIP itself only talks about `TimeoutException` there could be other causes for not making progress and thus we should be a little bit more general here.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StoreChangelogReader.java
```diff
@@ -625,11 +670,16 @@ private void restoreChangelog(final ChangelogMetadata changelogMetadata) {
...
+
+            final Map<TopicPartition,  ListOffsetsResult.ListOffsetsResultInfo> resultPerPartition = result.all().get();
+            clearTaskTimeout(getTasksFromPartitions(tasks, partitions));
+
+            return resultPerPartition.entrySet().stream().collect(
+                    Collectors.toMap(Map.Entry::getKey, entry -> entry.getValue().offset())
+            );
+        } catch (final TimeoutException | InterruptedException | ExecutionException retriableException) {
             log.debug("Could not fetch all end offsets for {}, will retry in the next run loop", partitions);
+            maybeInitTaskTimeoutOrThrow(getTasksFromPartitions(tasks, partitions), retriableException);
```
mjsax, line 682: Here, we could also get different exceptions in addition to `TimeoutException` that we should just handle in the same way?

Not sure about `InterruptedException`? Should this ever be thrown? Would it indicate a bug and we should just die?

mjsax:
Updates this PR to add the missing test updates. Also rebased and squashed.

Call for review @vvcephei

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
```diff
@@ -147,9 +147,12 @@ public void update(final Set<TopicPartition> topicPartitions, final Map<String,
...
     }
 
+    /**
+     * @throws TimeoutException if {@code currentWallClockMs > task-timeout-deadline}
+     */
     void maybeInitTaskTimeoutOrThrow(final long currentWallClockMs,
-                                     final TimeoutException timeoutException,
-                                     final Logger log) throws StreamsException {
+                                     final Exception cause,
+                                     final Logger log) {
```
vvcephei: It seems like we ought to just define `log` at the AbstractTask level and avoid having two almost identical `maybeInitTaskTimeoutOrThrow` method definitions.

streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
vvcephei: Sounds good!

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -482,8 +486,10 @@ boolean tryToCompleteRestoration() {
                 if (restored.containsAll(task.changelogPartitions())) {
                     try {
                         task.completeRestoration();
-                    } catch (final TimeoutException e) {
-                        log.debug("Could not complete restoration for {} due to {}; will retry", task.id(), e);
+                        task.clearTaskTimeout();
+                    } catch (final TimeoutException timeoutException) {
+                        task.maybeInitTaskTimeoutOrThrow(now, timeoutException);
+                        log.debug("Could not complete restoration for {} due to {}; will retry", task.id(), timeoutException);
```
vvcephei:
```suggestion
                        log.debug(String.format("Could not complete restoration for %s; will retry", task.id()), timeoutException);
```

It might be a good idea to add tests for the log messages so we can tell if they're actually properly formatted or not. Hopefully, the log4j upgrade makes it easier to detect these logging bugs.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -454,18 +454,22 @@ private void addNewTask(final Task task) {
...
             try {
                 task.initializeIfNeeded();
-            } catch (final LockException | TimeoutException e) {
+                task.clearTaskTimeout();
+            } catch (final LockException retriableException) {
                 // it is possible that if there are multiple threads within the instance that one thread
                 // trying to grab the task from the other, while the other has not released the lock since
                 // it did not participate in the rebalance. In this case we can just retry in the next iteration
-                log.debug("Could not initialize {} due to the following exception; will retry", task.id(), e);
+                log.debug("Could not initialize {} due to the following exception; will retry", task.id(), retriableException);
```
vvcephei:
```suggestion
                log.debug(String.format("Could not initialize %s due to the following exception; will retry", task.id()), retriableException);
```

vvcephei (commented):
Hey, thanks for the update, @mjsax ! Just a few more questions...

mjsax:
Updated.

vvcephei (approved):
Thanks, @mjsax !

PR #9369, opened 2020-10-03
KAFKA-4715: Ignore case of CompressionType and OffsetResetStrategy

Fleshgrinder:
This allows the configuration values for CompressionType in the ProducerConfig and the OffsetResetStrategy in the ConsumerConfig to be specified regardless of their case.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/OffsetResetStrategy.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/OffsetResetStrategy.java
@@ -17,5 +17,15 @@
 package org.apache.kafka.clients.consumer;
 
 public enum OffsetResetStrategy {
-    LATEST, EARLIEST, NONE
+    LATEST, EARLIEST, NONE;
+
+    public static OffsetResetStrategy forName(final String name) {
+        for (final OffsetResetStrategy value : values()) {
+            if (value.name().equalsIgnoreCase(name)) {
+                return value;
+            }
+        }
+
+        throw new IllegalArgumentException("Unknown offset reset strategy: " + name);
+    }
 }
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/OffsetResetStrategy.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/OffsetResetStrategy.java
@@ -16,10 +16,29 @@
  */
 package org.apache.kafka.clients.consumer;
 
+import java.util.Locale;
+import java.util.Objects;
+
+/**
+ * @see ConsumerConfig#AUTO_OFFSET_RESET_CONFIG
+ * @see ConsumerConfig#AUTO_OFFSET_RESET_DOC
+ */
 public enum OffsetResetStrategy {
     LATEST, EARLIEST, NONE;
 
+    // Enums are singletons, this means that this conversion happens once the
+    // first time a variant is actually used, and never again.
+    private final String id = name().toLowerCase(Locale.ROOT);
+
+    /**
+     * Get the {@link OffsetResetStrategy} for the given {@code name}.
+     *
+     * @throws IllegalArgumentException if the given {@code name} does not match any {@link OffsetResetStrategy}.
+     * @throws NullPointerException     if the given {@code name} is null.
+     */
     public static OffsetResetStrategy forName(final String name) {
+        Objects.requireNonNull(name, "name must not be null");
+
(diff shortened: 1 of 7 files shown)


Fleshgrinder:
Failed tests seem unrelated to me.

vvcephei:
Thanks for the PR @Fleshgrinder ,

The code change looks good to me. Should we have some tests for this?

Thanks!
-John

Fleshgrinder:
> The code change looks good to me. Should we have some tests for this?

Always I'd say. 😊

Fleshgrinder:
@vvcephei done

clients/src/main/java/org/apache/kafka/clients/consumer/OffsetResetStrategy.java
```diff
@@ -16,6 +16,40 @@
...
+ * @see ConsumerConfig#AUTO_OFFSET_RESET_CONFIG
+ * @see ConsumerConfig#AUTO_OFFSET_RESET_DOC
+ */
 public enum OffsetResetStrategy {
-    LATEST, EARLIEST, NONE
+    LATEST, EARLIEST, NONE;
+
+    // Enums are singletons, this means that this conversion happens once the
+    // first time a variant is actually used, and never again.
+    private final String id = name().toLowerCase(Locale.ROOT);
```
chia7712, line 31: Sorry that I can't catch your point here. Could you give more explanation to me? Is it different to ```Enum#toString```?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -361,7 +363,7 @@ public static StreamThread create(final InternalTopologyBuilder builder,
         final AtomicLong nextScheduledRebalanceMs = new AtomicLong(Long.MAX_VALUE);
         consumerConfigs.put(StreamsConfig.InternalConfig.NEXT_SCHEDULED_REBALANCE_MS, nextScheduledRebalanceMs);
 
-        final String originalReset = (String) consumerConfigs.get(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG);
+        final String originalReset = OffsetResetStrategy.forName((String) consumerConfigs.get(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG)).toString();
```
chia7712, line 366: Not sure whether this change is necessary. The value should be validated already.

chia7712:
@Fleshgrinder I like this idea of ignoring case for enum type. However, it would be better to change all enum-related configs at once (and it may need KIP) since users may be confused by this inconsistent checks.

Fleshgrinder:
I can try to find the others.

Why inconsistent? This is the classical robustness principle: be lean in what you accept and strict in what you produce.

chia7712:
> Why inconsistent? This is the classical robustness principle: be lean in what you accept and strict in what you produce.

sorry for unclear description. My point was that all enum-related config should be case-insensitive so users don't need to check the source code to avoid encountering case issue.

clients/src/main/java/org/apache/kafka/clients/consumer/OffsetResetStrategy.java
Fleshgrinder: We can move the code to `toString` but than the `toLowerCase` conversion is executed every time `toString` is invoked. This way the conversion happens once: the first time the enum is used. From there on out the string stays and never needs to be recomputed. Meaning:

```java
System.out.println(OffsetResetStrategy.LATEST); // first access, `id` member is populated with `toLowerCase` result
System.out.println(OffsetResetStrategy.LATEST); // just uses `id`
System.out.println(OffsetResetStrategy.LATEST); // just uses `id`
System.out.println(OffsetResetStrategy.LATEST); // just uses `id`
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
Fleshgrinder: The value could be in any format: `LaTeSt`, `latest`, `LATEST`, … but we are interested only in its canonical form and the canonical form is provided through the Enum, hence, we have to parse it again. The alternative would be to call `toLowerCase(Locale.ROOT)` on it but we would effectively be duplicating the logic and that's very bad.

Ideal would actually be if our configuration is not giving us a string but rather the Enum member that was configured but I think we all know the our configuration is generally not ideal, so that's an issue for another time.

PR #9370, opened 2020-10-04
KAFKA-9234: Added Nullability Annotations to Admin

Fleshgrinder:
This introduces a new **compile-only** dependency on [JetBrains Java Annotations](https://github.com/JetBrains/java-annotations). I decided to go for this particular dependency because it does not introduce any visible change to our users, nor does it introduce any additional overhead to the build process. If we get an annotation wrong, nothing bad happens because it is totally up to the code to actually validate things. (This also means that no KIP is required.)

One downside (some might say upside) is that absolutely everything must be annotated. I personally think this is a good thing because it forces people to actually think about `null`, same is the case in Kotlin where I also have to define whether `null` is permitted or not absolutely everywhere. Kotlin is also the main reason why I am interested in this, and also why I am interested in having this everywhere.

I was not able to test all possible NPE situations because some require the call to actually be executed. For those where I was able to prove that `null` is invalid we could think about adding `Objects.requireNonNull` guards right at the beginning. For all others it is simple, if we prove it we can add the guards too. This might be wasteful for collections (especially if those collections contain collections). A possible way out here is to use `assert` and live with the fact that some NPEs are thrown much later and will make it hard to find the source. This, sadly, is not Kotlin, so there is no easy way for this.

## Bugfix
~This PR also contains a bug fix. `Admin.describeUserScramCredentials()` (without any arguments) is passing `null` to `Admin.describeUserScramCredentials(List<String>)` which is permitted but leads to an NPE. Maybe there is another ticket for this, or nobody ever called this method (which would raise the question why it exists). In any event, I changed `KafkaAdminClient` to use an empty list if `null` is passed.~ See https://github.com/apache/kafka/pull/9374, I will update the description and this branch once the other PR is in trunk.

----

**PS:** I am willing to continue this endeavor if it is decided to accept this PR (also if we decide to opt for another library that provides the ability to add null checks, e.g. Checker Framework, Lombok), simply because the experience in Kotlin with Kafka and the missing nullability information is horrid.

**PPS:** The JetBrains annotations contain additional annotations for [`@Unmodifiable`](https://javadoc.io/doc/org.jetbrains/annotations/latest/org/jetbrains/annotations/Unmodifiable.html), [`@ApiStatus.Internal`](https://javadoc.io/doc/org.jetbrains/annotations/latest/org/jetbrains/annotations/ApiStatus.Internal.html), [`@RegExp`](https://javadoc.io/doc/org.jetbrains/annotations/latest/org/intellij/lang/annotations/RegExp.html), and [many more](https://javadoc.io/doc/org.jetbrains/annotations/latest/index.html) that ease development work tremendously. I think it would be more than worthwhile to start using them as well if we decide to go with these annotations.

— [KAFKA-9234](https://issues.apache.org/jira/browse/KAFKA-9234)


rondagostino:
@Fleshgrinder Thanks for the PR.  I'll let others comment on the main gist of it (`null` has probably wasted more than a billion dollars, so I'm generally all for mitigating the damage to the extent possible).

Regarding the bug you found, I believe that since KIP-554 stated the behavior publicly the behavior has to stay that way unless there is another KIP to change it.  One approach is to add a test that exposes the issue and commit that test and the corresponding fix.  The issue can be exposed by eliminating the `usersRequestedList` parameter at https://github.com/apache/kafka/blob/trunk/clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java#L4528.  I would be glad to open a PR if that is ultimately how we decide we want to go, or feel free to do what you want/need here based on the discussion -- either way is fine with me.

Fleshgrinder:
Hey @rondagostino 😊 we can equally well fix it in the other direction and allow `null`. Let me update the PR and revert the doc changes.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4160,15 +4231,18 @@ void handleFailure(Throwable throwable) {
...
+    ) {
         final KafkaFutureImpl<DescribeUserScramCredentialsResponseData> dataFuture = new KafkaFutureImpl<>();
         final long now = time.milliseconds();
         Call call = new Call("describeUserScramCredentials", calcDeadlineMs(now, options.timeoutMs()),
                 new LeastLoadedNodeProvider()) {
             @Override
             public DescribeUserScramCredentialsRequest.Builder createRequest(int timeoutMs) {
                 return new DescribeUserScramCredentialsRequest.Builder(
-                        new DescribeUserScramCredentialsRequestData().setUsers(users.stream().map(user ->
+                        new DescribeUserScramCredentialsRequestData().setUsers(users == null ? Collections.emptyList() : users.stream().map(user ->
```
rondagostino, line 4245: Not 100% sure the parens are needed, but either they are or it increases clarity/decreases confusion to add them.
```suggestion
                        new DescribeUserScramCredentialsRequestData().setUsers((users == null ? Collections.emptyList() : users).stream().map(user ->
```

Fleshgrinder: This way we would also run the empty map through _stream_, which is not required. I tried to make the change with as little impact as possible but if I were to write it then I would do it as follows:

```java
        Call call = new Call("describeUserScramCredentials", calcDeadlineMs(now, options.timeoutMs()), new LeastLoadedNodeProvider()) {
            @Override
            public DescribeUserScramCredentialsRequest.Builder createRequest(final int timeoutMs) {
                final DescribeUserScramCredentialsRequestData requestData = new DescribeUserScramCredentialsRequestData();

                if (users != null && !users.isEmpty()) {
                    final List<UserName> userNames = new ArrayList<>(users.size());
                    for (final String user : users) {
                        userNames.add(new UserName().setName(user));
                    }
                    requestData.setUsers(userNames);
                }

                return new DescribeUserScramCredentialsRequest.Builder(requestData);
            }
```

This is more code, yes, and it is not using _stream_ anymore but it is both clearer, faster, and allocates less.

rondagostino: I’m thinking it would be best to fix this bug in a separate “MINOR: fix potential NPE...” PR since this PR may or may not get merged.  Do you agree? If so, either I can do it, or feel free to do it — I’ll go with whichever you wish.

Fleshgrinder: Happy to help, PR incoming. 😊

PR #9371, opened 2020-10-04
KAFKA-10510: Validate replication factor consistency on reassignment

piotrrzysko:
This patch introduces server-side validation that checks if applying requested reassignment won't lead to a situation where different partitions of the same topic have different replication factors. 

If a given partition has a pending reassignment, then the target replication factor is used by the validator.

### DIFF
--- a/core/src/main/scala/kafka/controller/KafkaController.scala
+++ b/core/src/main/scala/kafka/controller/KafkaController.scala
@@ -1575,6 +1575,11 @@ class KafkaController(val config: KafkaConfig,
         }
       }
 
+      validateReplicationFactors(partitionsToReassign).forKeyValue { (tp, apiError) =>
+        partitionsToReassign.remove(tp)
+        reassignmentResults.put(tp, apiError)
+      }
+
       // The latest reassignment (whether by API or through zk) always takes precedence,
       // so remove from active zk reassignment (if one exists)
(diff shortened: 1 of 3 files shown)


piotrrzysko:
Hi @stanislavkozlovski, would you mind taking a look at this PR?

stanislavkozlovski:
Thanks for the PR @piotrrzysko. The changes look good although I'm not sure if we're not regressing the ability to change the replication factor of a topic if we're to outright block it.

Perhaps we could have a workaround in the form of allowing the reassignment batch to run if it's changing all the partitions for the topic. Another solution I like more is adding a flag to the AlterPartitionReassignments API that allows it to change the replication factor and keeping this validation.

In the long term, I think we may want to consider an API that allows you to alter the replication factor of a topic.

cc @ijuma @gwenshap @cmccabe @hachikuji @soondenana do you happen to have any thoughts on the matter?

piotrrzysko:
Thanks for the review, @stanislavkozlovski. If I understand your point clearly, you are wondering whether this validation will block changing the replication factor for all partitions (replication factor of a topic). Right? It will not if all requested reassignments in the batch have the same number of target replicas for a given topic. Please take look at the test `ReassignPartitionsIntegrationTest::testReassignmentFailOnInconsistentReplicationFactorBetweenPartitions` - topic `bar`. It is a simple case because the topic has only one partition, but it shows that changing the replication factor from 3 to 4 is possible.

PR #9374, opened 2020-10-05
MINOR: Fix NPE in KafkaAdminClient.describeUserScramCredentials

Fleshgrinder:
This bug was detected as part of #9370. @rondagostino and I decided that this should be fixed right away and not until the other PR eventually gets merged (or not). I am not only properly handling `null` everywhere but also rewrote the function to be more readable (and more efficient, but I think this is not important here). I extended the existing test to go through all possible permutations of the users argument to make sure that we never get an NPE.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
+++ b/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
@@ -116,6 +116,7 @@ import org.apache.kafka.common.message.DescribeLogDirsRequestData;
 import org.apache.kafka.common.message.DescribeLogDirsRequestData.DescribableLogDirTopic;
 import org.apache.kafka.common.message.DescribeLogDirsResponseData;
 import org.apache.kafka.common.message.DescribeUserScramCredentialsRequestData;
+import org.apache.kafka.common.message.DescribeUserScramCredentialsRequestData.UserName;
 import org.apache.kafka.common.message.DescribeUserScramCredentialsResponseData;
 import org.apache.kafka.common.message.ExpireDelegationTokenRequestData;
 import org.apache.kafka.common.message.FindCoordinatorRequestData;
@@ -4166,10 +4167,22 @@ public class KafkaAdminClient extends AdminClient {
         Call call = new Call("describeUserScramCredentials", calcDeadlineMs(now, options.timeoutMs()),
                 new LeastLoadedNodeProvider()) {
             @Override
-            public DescribeUserScramCredentialsRequest.Builder createRequest(int timeoutMs) {
-                return new DescribeUserScramCredentialsRequest.Builder(
-                        new DescribeUserScramCredentialsRequestData().setUsers(users.stream().map(user ->
-                                new DescribeUserScramCredentialsRequestData.UserName().setName(user)).collect(Collectors.toList())));
+            public DescribeUserScramCredentialsRequest.Builder createRequest(final int timeoutMs) {
+                final DescribeUserScramCredentialsRequestData requestData = new DescribeUserScramCredentialsRequestData();
+
+                if (users != null && !users.isEmpty()) {
+                    final List<UserName> userNames = new ArrayList<>(users.size());
+
+                    for (final String user : users) {
+                        if (user != null) {
+                            userNames.add(new UserName().setName(user));
+                        }
+                    }
+
+                    requestData.setUsers(userNames);
+                }
+
+                return new DescribeUserScramCredentialsRequest.Builder(requestData);
             }
 
             @Override
--- a/clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
@@ -4492,10 +4492,10 @@ public class KafkaAdminClientTest {
             env.kafkaClient().setNodeApiVersions(NodeApiVersions.create());
 
             final String user0Name = "user0";
-            ScramMechanism user0ScramMechanism0 = ScramMechanism.SCRAM_SHA_256;
-            int user0Iterations0 = 4096;
-            ScramMechanism user0ScramMechanism1 = ScramMechanism.SCRAM_SHA_512;
-            int user0Iterations1 = 8192;
+            final ScramMechanism user0ScramMechanism0 = ScramMechanism.SCRAM_SHA_256;
+            final int user0Iterations0 = 4096;
+            final ScramMechanism user0ScramMechanism1 = ScramMechanism.SCRAM_SHA_512;
+            final int user0Iterations1 = 8192;
 
             final CredentialInfo user0CredentialInfo0 = new CredentialInfo();
             user0CredentialInfo0.setMechanism(user0ScramMechanism0.type());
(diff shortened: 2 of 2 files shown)


clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4166,10 +4167,22 @@ public DescribeUserScramCredentialsResult describeUserScramCredentials(List<Stri
         Call call = new Call("describeUserScramCredentials", calcDeadlineMs(now, options.timeoutMs()),
                 new LeastLoadedNodeProvider()) {
             @Override
-            public DescribeUserScramCredentialsRequest.Builder createRequest(int timeoutMs) {
-                return new DescribeUserScramCredentialsRequest.Builder(
-                        new DescribeUserScramCredentialsRequestData().setUsers(users.stream().map(user ->
```
Fleshgrinder, line 4171: `users.stream()` is the NPE source

rondagostino (commented):
Thanks for finding the bug and submitting this PR, @Fleshgrinder.  LGTM, glad you checked all 3 possibilities (null, empty list, full list of all users).

chia7712:
Nice finding!

Is there also a potential NPE in server-side? According to the protocol schema (https://github.com/apache/kafka/blob/trunk/clients/src/main/resources/common/message/DescribeUserScramCredentialsRequest.json#L23), the field ```users``` is nullable. Does server have to handle null as well?(https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/KafkaApis.scala#L3068).

For another, if server is able to handle null, it should be fine to set null to ```users```. Hence, we can simplify client code as following style.

```java
            @Override
            public DescribeUserScramCredentialsRequest.Builder createRequest(int timeoutMs) {
                return new DescribeUserScramCredentialsRequest.Builder(
                        new DescribeUserScramCredentialsRequestData().setUsers(users == null ? null : users.stream().map(user ->
                                new DescribeUserScramCredentialsRequestData.UserName().setName(user)).collect(Collectors.toList())));
            }
```

please let me know what you think :)

Fleshgrinder:
@chia7712 the server gets null right now, and the code you propose is more or less what I had originally. However, at least @rondagostino was directly confused by the ternary and precedence that's why we rewrote it to be easier to understand and easier to debug (now it's possible to set individual breakpoints). Modern languages like Kotlin and Rust have no ternary for a good reason. 😉 Imho it's better as is, despite being longer, especially because length doesn't translate directly to complex.

We'd also still be collecting `null` usernames. The field is nullable in the `UserName` class but collecting them is still a bad idea.

I need to check the server side.

chia7712:
this PR is to fix a critical bug. Maybe we should file a jira for it (and so it can be in 2.7.0 release plan)

rondagostino:
This feature is just being released in 2.7 and will itself be mentioned in the release notes, so a separate ticket for this issue seems unnecessary to me.

chia7712:
> This feature is just being released in 2.7 and will itself be mentioned in the release notes, so a separate ticket for this issue seems unnecessary to me.

I just worry this "minor" patch is not included by 2.7.0.

rondagostino:
> I just worry this "minor" patch is not included by 2.7.0.

Yeah, we'll need to make sure that happens.  @cmccabe, as the reviewer for the original code, could you take a look at this small PR to fix a potential NPE that I accidentally introduced?  It will need to go to the 2.7 branch as well as trunk if/when it is merged.

dajac:
@rondagostino Is this issue a blocker for 2.7? cc @bbejeck

rondagostino:
This isn’t a regression, so I don’t think it is a blocker.  NPE would only occur if someone were to code against the new API.  The kafka-configs.sh CLI works fine because it doesn’t exercise the API in a way that generates the NPE.

Would I like to get this fix in?  Yes.  But is it a blocker?  I don’t think so.

chia7712:
We can merge this patch so as to avoid making a release with (known) unavailable Public APIs. Honoring the protocol (null handle) on server-side can be a follow-up. Just my two cents.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4166,10 +4167,22 @@ public DescribeUserScramCredentialsResult describeUserScramCredentials(List<Stri
...
+                if (users != null && !users.isEmpty()) {
+                    final List<UserName> userNames = new ArrayList<>(users.size());
+
+                    for (final String user : users) {
+                        if (user != null) {
+                            userNames.add(new UserName().setName(user));
+                        }
+                    }
+
+                    requestData.setUsers(userNames);
```
dajac, line 4182: nit: I suggest to continue using the stream api here. It keeps the code smaller.

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
```diff
@@ -4505,50 +4505,57 @@ public void testDescribeUserScramCredentials() throws Exception {
...
-            assertEquals(user1Iterations, userScramCredentialsDescription1.credentialInfos().get(0).iterations());
-            assertEquals(userScramCredentialsDescription1, user1DescriptionFuture.get());
+                            .setCredentialInfos(singletonList(user1CredentialInfo))));
+            final DescribeUserScramCredentialsResponse response = new DescribeUserScramCredentialsResponse(responseData);
+
+            final Set<String> usersRequestedSet = new HashSet<>();
+            usersRequestedSet.add(user0Name);
+            usersRequestedSet.add(user1Name);
+
+            for (final List<String> users : asList(null, new ArrayList<String>(), asList(user0Name, null, user1Name))) {
```
dajac, line 4529: Would it make sense to extract these into separate unit tests? `testDescribeUserScramCredentials` could receive `users` as a argument such that we could reuse the code logic.

dajac (commented):
@Fleshgrinder Thanks for the PR. I have left few suggestions.

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
Fleshgrinder: We could but I do not see where we could reuse it, YAGNI?

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
Fleshgrinder: Not really, if we keep all checks in place it will only be a little shorter but definitely less efficient:

```java
final DescribeUserScramCredentialsRequestData requestData = new DescribeUserScramCredentialsRequestData();

if (users != null && !users.isEmpty()) {
    requestData.setUsers(users.stream()
        .filter(Objects::nonNull)
        .map(it -> new UserName().setName(it))
        .collect(Collectors.toList())
    );
}

return new DescribeUserScramCredentialsRequest.Builder(requestData);
```

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
dajac: Sorry, I was not clear. It would be great if we could define separate unit tests for the different cases that we want to test. In case of failure, we would know directly which one of the cases has failed. This is why I suggested to pass an argument `users` as an argument to `testDescribeUserScramCredentials`.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
dajac: We tend to use the steam api for such small transformations but I don't feel strong about this.

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
Fleshgrinder: We can rewrite it to a test factory, this way we can keep it in one function and reuse the initialization logic but will have _n_ different tests being executed so that it is clear which one failed.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
Fleshgrinder: I have no feelings regarding this either, for me Java is always just clunky no matter which API you use. A normal loop is just faster, that's all I can say, but I doubt that it matters at all.

dajac:
FYI - I just opened a PR to fix the issue on the broker side: https://github.com/apache/kafka/pull/9504.

dajac (approved):
@Fleshgrinder Thanks for the PR. Let's keep it as-is to get in 2.7. LGTM.

dajac:
Test failures not related. Merging to trunk and 2.7.

PR #9375, opened 2020-10-05
KAFKA-10574: Fix infinite loop in Values::parseString

C0urante:
[Jira](https://issues.apache.org/jira/browse/KAFKA-10574)

The special byte sequence `0xEF, 0xBF, 0xBF`, when parsed as a UTF-8 string, causes the `StringCharacterIterator` to return `CharacterIterator.DONE` from its `next()`, `current()`, and other similar methods. This caused an infinite loop in the `Values` class whenever that byte sequence was encountered.

The fix is pretty simple. To see if we're at the end of a string, we compare `StandardCharacterIterator::getIndex` to `StandardCharacterIterator::getEndIndex`, instead of comparing the last-read character to `CharacterIterator.DONE` (since that character may occur in strings that we're parsing).

I've added a unit test that replicates this bug (when the fix in the `Values` class is not present). I gave it a timeout of five seconds in case someone accidentally re-creates the infinite loop in order to save some time and potential confusion.

All existing unit tests for the `Values` class pass with this change.

### DIFF
--- a/connect/api/src/test/java/org/apache/kafka/connect/data/ValuesTest.java
+++ b/connect/api/src/test/java/org/apache/kafka/connect/data/ValuesTest.java
@@ -24,9 +24,7 @@ import org.junit.Test;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
-import java.text.CharacterIterator;
 import java.text.SimpleDateFormat;
-import java.text.StringCharacterIterator;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -80,7 +78,10 @@ public class ValuesTest {
 
     @Test(timeout = 5000)
     public void shouldNotEncounterInfiniteLoop() {
-        byte[] bytes = new byte[] { -17, -65,  -65 };
+        // This byte sequence gets parsed as CharacterIterator.DONE and can cause issues if
+        // comparisons to that character are done to check if the end of a string has been reached.
+        // For more information, see https://issues.apache.org/jira/browse/KAFKA-10574
+        byte[] bytes = new byte[] {-17, -65,  -65};
         String str = new String(bytes, StandardCharsets.UTF_8);
         SchemaAndValue schemaAndValue = Values.parseString(str);
         assertEquals(Type.STRING, schemaAndValue.schema().type());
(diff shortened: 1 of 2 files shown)


gharris1727 (approved):
LGTM!

C0urante:
@rhauch @kkonstantine can one of you take a look?

connect/api/src/test/java/org/apache/kafka/connect/data/ValuesTest.java
```diff
@@ -73,6 +76,15 @@
         INT_LIST.add(-987654321);
     }
 
+    @Test(timeout = 5000)
+    public void shouldNotEncounterInfiniteLoop() {
+        byte[] bytes = new byte[] { -17, -65,  -65 };
```
kkonstantine: we need a comment here to explain things.

kkonstantine (approved):
LGTM. The fixes makes sense to me but I'm not intimately familiar with this class. 
@rhauch, any concerns with the new check?


@C0urante I'd like to ask for a comment to the test so it's not as cryptic as assuming what these bytes mean for the given encoding.

connect/api/src/test/java/org/apache/kafka/connect/data/ValuesTest.java
C0urante: 👍   good call, will add

kkonstantine (commented):
fyi: jobs fail with checkstyle errors

kkonstantine (changes requested):
Need to fix checkstyle before merging.

C0urante:
Thanks @kkonstantine, I've added a comment and addressed the Checkstyle issues.

rhauch (approved):
LGTM, pending a green build. Thanks, @C0urante.

kkonstantine (approved):
LGTM after the latest fixes

C0urante:
Hi @rhauch @kkonstantine, this PR has been approved, all merge conflicts have been resolved, and every build is green. Would it be possible to merge sometime soon so that it can be backported to bug fix releases and avoid accruing merge conflicts in case the files touched here are modified in other PRs? Thanks!

sfc-gh-japatel:
Hello, I am trying to repro this issue in a connect environment, any guidance on what could be the message to hit this in 2.3.1 and eventually an infinite loop?

PR #9380, opened 2020-10-06
KAFKA-7334: Suggest changing config for state.dir in case of FileNotF…

voffcheg109:
When state.dir is left at default configuration, there is a chance that certain files under the state directory are cleaned by OS since the default dir starts with /tmp/kafka-streams.

I've added suggestion to change state.dir config in the exception message.
(https://issues.apache.org/jira/browse/KAFKA-7334)

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
@@ -446,7 +446,9 @@ public class GlobalStateManagerImpl implements GlobalStateManager {
         try {
             checkpointFile.write(filteredOffsets);
         } catch (final IOException e) {
-            log.warn("Failed to write offset checkpoint file to {} for global stores: {}", checkpointFile, e);
+            log.warn("Failed to write offset checkpoint file to {} for global stores: {}." +
+                " This may occur if OS cleaned the state.dir in case when it located in /tmp directory." +
+                " You can change location for state.dir to resolve problem", checkpointFile, e);
         }
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
@@ -447,8 +447,8 @@ public class GlobalStateManagerImpl implements GlobalStateManager {
             checkpointFile.write(filteredOffsets);
         } catch (final IOException e) {
             log.warn("Failed to write offset checkpoint file to {} for global stores: {}." +
-                " This may occur if OS cleaned the state.dir in case when it located in /tmp directory." +
-                " You can change location for state.dir to resolve problem", checkpointFile, e);
+                " This may occur if OS cleaned the state.dir in case when it is located in the (default) /tmp/kafka-streams directory." +
+                " Changing the location of state.dir may resolve the problem", checkpointFile, e);
         }
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorStateManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorStateManager.java
@@ -605,7 +605,9 @@ public class ProcessorStateManager implements StateManager {
         } catch (final IOException e) {
             log.warn("Failed to write offset checkpoint file to [{}]." +
                 " This may occur if OS cleaned the state.dir in case when it located in /tmp directory." +
-                " You can change location for state.dir to resolve problem", checkpointFile, e);
+                " You can change location for state.dir to resolve problem." +
+                " This can also occur due to running multiple instances on the same machine using the same state dir.",
+                checkpointFile, e);
         }
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorStateManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorStateManager.java
@@ -605,8 +605,8 @@ public class ProcessorStateManager implements StateManager {
         } catch (final IOException e) {
             log.warn("Failed to write offset checkpoint file to [{}]." +
                 " This may occur if OS cleaned the state.dir in case when it located in /tmp directory." +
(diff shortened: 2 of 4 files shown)


ableegoldman (commented):
Thanks for picking this up! This looks good but I have two additional suggestions:

1) Add a note to these warnings to clarify that this can also occur due to running multiple instances on the same machine using the same state dir
2) Check if the configured `state.dir` is `/tmp` and log an additional warning if so (I believe we parse this config in the constructor of StateDirectory)

voffcheg109:
@ableegoldman, I've made changes following you suggestions, but I got some strange errors, which don't reproduce locally. I think would be enough to retrigger builds. Is there any way to do it?

ableegoldman:
Don't worry about the build, looks like only some unrelated known flaky tests in `kafka.api` caused it to fail.

ableegoldman (approved):
LGTM, thanks for the improvement!

ableegoldman:
Ping @mjsax for merging responsibility. Can you cherrypick it back to 2.6 as well?

voffcheg109:
@mjsax, can you merge this pr? 

@ableegoldman, I've also created pr for 2.6 - https://github.com/apache/kafka/pull/9389

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -446,7 +446,9 @@ public void checkpoint() {
         try {
             checkpointFile.write(filteredOffsets);
         } catch (final IOException e) {
-            log.warn("Failed to write offset checkpoint file to {} for global stores: {}", checkpointFile, e);
+            log.warn("Failed to write offset checkpoint file to {} for global stores: {}." +
+                " This may occur if OS cleaned the state.dir in case when it located in /tmp directory." +
```
mjsax: `it located` -> `it is located`

`in /tmp directory` -> `in the (default) /tmp/kafka-streams directory` ?

streams/src/main/java/org/apache/kafka/streams/processor/internals/GlobalStateManagerImpl.java
```diff
@@ -446,7 +446,9 @@ public void checkpoint() {
...
+                " You can change location for state.dir to resolve problem", checkpointFile, e);
```
mjsax: Do we _know_ that it will resolve the problem? Maybe better:
```
Changing the location of state.dir may resolve the problem
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -98,6 +98,10 @@ public StateDirectory(final StreamsConfig config, final Time time, final boolean
             throw new ProcessorStateException(
                 String.format("state directory [%s] doesn't exist and couldn't be created", stateDir.getPath()));
         }
+        if (stateDirName.startsWith("/tmp")) {
```
mjsax: Wondering if this should go into `StreamsConfig` ?

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorStateManager.java
```diff
@@ -603,7 +603,11 @@ public void checkpoint() {
         try {
             checkpointFile.write(checkpointingOffsets);
         } catch (final IOException e) {
-            log.warn("Failed to write offset checkpoint file to [{}]", checkpointFile, e);
+            log.warn("Failed to write offset checkpoint file to [{}]." +
+                " This may occur if OS cleaned the state.dir in case when it located in /tmp directory." +
+                " You can change location for state.dir to resolve problem." +
+                " This can also occur due to running multiple instances on the same machine using the same state dir.",
```
mjsax: Hmmm... Not sure if I can follow? Why?

Also, if you run multiple instances on the same machine, it seems to be the same as running multiple threads in the same `KafkaStreams` client?

Also, if you configure different state directories, local state cannot be shared what might be wasteful?

Can you elaborate on this sentence?

ableegoldman: @mjsax you can't run multiple instances on the same machine with the same state.dir. For one thing, the locking mechanism is per-process*. If you run two different instances then you can get an active task on one instance and the corresponding standby on another. They would each think they owned the lock for that task directory, and concurrently access it (leading to the FileNotFoundException if one of them deletes the checkpoint, for example)

*on some systems. On others it isn't, but then you hit the opposite problem where a task is deadlocked because the other process grabbed the lock for its directory first

ableegoldman: Currently, running multiple threads within an instance is _not_ the same as running multiple instances on the same machine in the same state dir. I suppose technically if you could configure the two instances to have the same client UUID then the task assignor would make sure not to assign the same active/standby task to anyone on that machine, but I think there might be some unexpected side effects to running two separates instances with the same UUID.

Anyways, if you're aware of the problem enough to configure the client Ids directly, you should be aware enough to just configure to use separate state.dirs. I don't think I buy the argument about it being wasteful just because there's really no reason to ever do this in a production app. But people seem to do it in testing all the time, and run into this issue, hence the warning here

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
ableegoldman: Actually maybe we should wrap it in an `if hasPersistentStores` so that users won't get this warning if they don't have any persistent state

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorStateManager.java
mjsax: Thanks for clarification @ableegoldman!

voffcheg109:
@mjsax, I've fixed by your comments.

ableegoldman:
Tests passed, should be good to merge @mjsax . Btw the 2.7 branch was just cut so this should be cherrypicked back to 2.7.

mjsax:
Thanks for the PR @voffcheg109!

PR #9382, opened 2020-10-06
KAFKA-10554; Perform follower truncation based on diverging epochs in Fetch response

rajinisivaram:
For IBP 2.7 onwards, fetch responses include diverging epoch and offset in fetch responses if `lastFetchedEpoch` is provided in the fetch request. This PR uses that information for truncation and avoids the additional OffsetForLeaderEpoch requests in followers when `lastFetchEpoch` is known.

### DIFF
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -1385,7 +1385,7 @@ class Log(@volatile private var _dir: File,
     var validBytesCount = 0
     var firstOffset: Option[Long] = None
     var lastOffset = -1L
-    var lastLeaderEpoch: Option[Int] = None
+    var lastLeaderEpoch = RecordBatch.NO_PARTITION_LEADER_EPOCH
     var sourceCodec: CompressionCodec = NoCompressionCodec
     var monotonic = true
     var maxTimestamp = RecordBatch.NO_TIMESTAMP
@@ -1418,10 +1418,7 @@ class Log(@volatile private var _dir: File,
 
       // update the last offset seen
       lastOffset = batch.lastOffset
-      lastLeaderEpoch = if (batch.partitionLeaderEpoch != RecordBatch.NO_PARTITION_LEADER_EPOCH)
-        Some(batch.partitionLeaderEpoch)
-      else
-        None
+      lastLeaderEpoch = batch.partitionLeaderEpoch
 
       // Check if the message sizes are valid.
       val batchSize = batch.sizeInBytes
@@ -1453,7 +1450,11 @@ class Log(@volatile private var _dir: File,
 
     // Apply broker-side compression if any
     val targetCodec = BrokerCompressionCodec.getTargetCompressionCodec(config.compressionType, sourceCodec)
-    LogAppendInfo(firstOffset, lastOffset, lastLeaderEpoch, maxTimestamp, offsetOfMaxTimestamp, RecordBatch.NO_TIMESTAMP, logStartOffset,
+    val lastLeaderEpochOpt: Option[Int] = if (lastLeaderEpoch != RecordBatch.NO_PARTITION_LEADER_EPOCH)
+      Some(lastLeaderEpoch)
+    else
+      None
+    LogAppendInfo(firstOffset, lastOffset, lastLeaderEpochOpt, maxTimestamp, offsetOfMaxTimestamp, RecordBatch.NO_TIMESTAMP, logStartOffset,
       RecordConversionStats.EMPTY, sourceCodec, targetCodec, shallowMessageCount, validBytesCount, monotonic, lastOffsetOfFirstBatch)
   }
 
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -50,11 +50,11 @@ import scala.collection.mutable.{ArrayBuffer, ListBuffer}
 import scala.collection.{Seq, Set, mutable}
 
 object LogAppendInfo {
-  val UnknownLogAppendInfo = LogAppendInfo(None, -1, RecordBatch.NO_TIMESTAMP, -1L, RecordBatch.NO_TIMESTAMP, -1L,
+  val UnknownLogAppendInfo = LogAppendInfo(None, -1, None, RecordBatch.NO_TIMESTAMP, -1L, RecordBatch.NO_TIMESTAMP, -1L,
     RecordConversionStats.EMPTY, NoCompressionCodec, NoCompressionCodec, -1, -1, offsetsMonotonic = false, -1L)
 
   def unknownLogAppendInfoWithLogStartOffset(logStartOffset: Long): LogAppendInfo =
-    LogAppendInfo(None, -1, RecordBatch.NO_TIMESTAMP, -1L, RecordBatch.NO_TIMESTAMP, logStartOffset,
+    LogAppendInfo(None, -1, None, RecordBatch.NO_TIMESTAMP, -1L, RecordBatch.NO_TIMESTAMP, logStartOffset,
       RecordConversionStats.EMPTY, NoCompressionCodec, NoCompressionCodec, -1, -1,
       offsetsMonotonic = false, -1L)
 
@@ -64,7 +64,7 @@ object LogAppendInfo {
    * in unknownLogAppendInfoWithLogStartOffset, but with additiona fields recordErrors and errorMessage
    */
   def unknownLogAppendInfoWithAdditionalInfo(logStartOffset: Long, recordErrors: Seq[RecordError], errorMessage: String): LogAppendInfo =
-    LogAppendInfo(None, -1, RecordBatch.NO_TIMESTAMP, -1L, RecordBatch.NO_TIMESTAMP, logStartOffset,
+    LogAppendInfo(None, -1, None, RecordBatch.NO_TIMESTAMP, -1L, RecordBatch.NO_TIMESTAMP, logStartOffset,
       RecordConversionStats.EMPTY, NoCompressionCodec, NoCompressionCodec, -1, -1,
       offsetsMonotonic = false, -1L, recordErrors, errorMessage)
 }
@@ -82,6 +82,7 @@ object LeaderHwChange {
  * @param firstOffset The first offset in the message set unless the message format is less than V2 and we are appending
  *                    to the follower.
  * @param lastOffset The last offset in the message set
+ * @param lastLeaderEpoch The partition leader epoch corresponding to the last offset, if available.
  * @param maxTimestamp The maximum timestamp of the message set.
  * @param offsetOfMaxTimestamp The offset of the message with the maximum timestamp.
  * @param logAppendTime The log append time (if used) of the message set, otherwise Message.NoTimestamp
@@ -99,6 +100,7 @@ object LeaderHwChange {
  */
 case class LogAppendInfo(var firstOffset: Option[Long],
                          var lastOffset: Long,
+                         var lastLeaderEpoch: Option[Int],
                          var maxTimestamp: Long,
                          var offsetOfMaxTimestamp: Long,
                          var logAppendTime: Long,
@@ -1383,6 +1385,7 @@ class Log(@volatile private var _dir: File,
     var validBytesCount = 0
     var firstOffset: Option[Long] = None
     var lastOffset = -1L
+    var lastLeaderEpoch: Option[Int] = None
     var sourceCodec: CompressionCodec = NoCompressionCodec
     var monotonic = true
     var maxTimestamp = RecordBatch.NO_TIMESTAMP
@@ -1415,6 +1418,10 @@ class Log(@volatile private var _dir: File,
 
       // update the last offset seen
       lastOffset = batch.lastOffset
+      lastLeaderEpoch = if (batch.partitionLeaderEpoch != RecordBatch.NO_PARTITION_LEADER_EPOCH)
+        Some(batch.partitionLeaderEpoch)
+      else
+        None
 
       // Check if the message sizes are valid.
       val batchSize = batch.sizeInBytes
@@ -1446,7 +1453,7 @@ class Log(@volatile private var _dir: File,
 
     // Apply broker-side compression if any
     val targetCodec = BrokerCompressionCodec.getTargetCompressionCodec(config.compressionType, sourceCodec)
-    LogAppendInfo(firstOffset, lastOffset, maxTimestamp, offsetOfMaxTimestamp, RecordBatch.NO_TIMESTAMP, logStartOffset,
+    LogAppendInfo(firstOffset, lastOffset, lastLeaderEpoch, maxTimestamp, offsetOfMaxTimestamp, RecordBatch.NO_TIMESTAMP, logStartOffset,
       RecordConversionStats.EMPTY, sourceCodec, targetCodec, shallowMessageCount, validBytesCount, monotonic, lastOffsetOfFirstBatch)
   }
 
--- a/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
+++ b/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
@@ -143,7 +143,7 @@ abstract class AbstractFetcherManager[T <: AbstractFetcherThread](val name: Stri
         }
 
         val initialOffsetAndEpochs = initialFetchOffsets.map { case (tp, brokerAndInitOffset) =>
-          tp -> OffsetAndEpoch(brokerAndInitOffset.initOffset, brokerAndInitOffset.currentLeaderEpoch)
+          tp -> OffsetAndEpoch(brokerAndInitOffset.initOffset, brokerAndInitOffset.currentLeaderEpoch, brokerAndInitOffset.lastFetchedEpoch)
         }
 
         addPartitionsToFetcherThread(fetcherThread, initialOffsetAndEpochs)
@@ -227,6 +227,6 @@ class FailedPartitions {
 
 case class BrokerAndFetcherId(broker: BrokerEndPoint, fetcherId: Int)
 
-case class InitialFetchState(leader: BrokerEndPoint, currentLeaderEpoch: Int, initOffset: Long)
+case class InitialFetchState(leader: BrokerEndPoint, currentLeaderEpoch: Int, initOffset: Long, lastFetchedEpoch: Option[Int])
 
 case class BrokerIdAndFetcherId(brokerId: Int, fetcherId: Int)
--- a/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
+++ b/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
@@ -64,8 +64,8 @@ abstract class AbstractFetcherManager[T <: AbstractFetcherThread](val name: Stri
   def resizeThreadPool(newSize: Int): Unit = {
     def migratePartitions(newSize: Int): Unit = {
       fetcherThreadMap.forKeyValue { (id, thread) =>
-        val removedPartitions = thread.partitionsAndOffsets
-        removeFetcherForPartitions(removedPartitions.keySet)
+        val removedPartitions = thread.removeAllPartitions()
+        removeFetcherForPartitions(removedPartitions.keySet) // clear state for removed partitions
         if (id.fetcherId >= newSize)
           thread.shutdown()
         addFetcherForPartitions(removedPartitions)
@@ -223,6 +223,6 @@ class FailedPartitions {
 
 case class BrokerAndFetcherId(broker: BrokerEndPoint, fetcherId: Int)
 
-case class InitialFetchState(leader: BrokerEndPoint, currentLeaderEpoch: Int, initOffset: Long, lastFetchedEpoch: Option[Int])
+case class InitialFetchState(leader: BrokerEndPoint, currentLeaderEpoch: Int, initOffset: Long)
 
 case class BrokerIdAndFetcherId(brokerId: Int, fetcherId: Int)
--- a/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
+++ b/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
@@ -17,16 +17,14 @@
 
 package kafka.server
 
-import kafka.utils.Implicits._
-import kafka.utils.Logging
 import kafka.cluster.BrokerEndPoint
 import kafka.metrics.KafkaMetricsGroup
+import kafka.utils.Implicits._
+import kafka.utils.Logging
 import org.apache.kafka.common.TopicPartition
 import org.apache.kafka.common.utils.Utils
 
-import scala.collection.mutable
-import scala.collection.{Map, Set}
-import scala.jdk.CollectionConverters._
+import scala.collection.{Map, Set, mutable}
 
 abstract class AbstractFetcherManager[T <: AbstractFetcherThread](val name: String, clientId: String, numFetchers: Int)
   extends Logging with KafkaMetricsGroup {
@@ -65,7 +63,7 @@ abstract class AbstractFetcherManager[T <: AbstractFetcherThread](val name: Stri
   def resizeThreadPool(newSize: Int): Unit = {
     def migratePartitions(newSize: Int): Unit = {
       fetcherThreadMap.forKeyValue { (id, thread) =>
-        val partitionStates = removeFetcherForPartitions(thread.partitions.asScala)
+        val partitionStates = removeFetcherForPartitions(thread.partitions)
         if (id.fetcherId >= newSize)
           thread.shutdown()
         val fetchStates = partitionStates.map { case (topicPartition, currentFetchState) =>
--- a/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
+++ b/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
@@ -26,6 +26,7 @@ import org.apache.kafka.common.utils.Utils
 
 import scala.collection.mutable
 import scala.collection.{Map, Set}
+import scala.jdk.CollectionConverters._
 
 abstract class AbstractFetcherManager[T <: AbstractFetcherThread](val name: String, clientId: String, numFetchers: Int)
   extends Logging with KafkaMetricsGroup {
@@ -64,11 +65,16 @@ abstract class AbstractFetcherManager[T <: AbstractFetcherThread](val name: Stri
   def resizeThreadPool(newSize: Int): Unit = {
     def migratePartitions(newSize: Int): Unit = {
       fetcherThreadMap.forKeyValue { (id, thread) =>
-        val removedPartitions = thread.removeAllPartitions()
-        removeFetcherForPartitions(removedPartitions.keySet) // clear state for removed partitions
+        val partitionStates = removeFetcherForPartitions(thread.partitions.asScala)
         if (id.fetcherId >= newSize)
           thread.shutdown()
-        addFetcherForPartitions(removedPartitions)
+        val fetchStates = partitionStates.map { case (topicPartition, currentFetchState) =>
+          val initialFetchState = InitialFetchState(thread.sourceBroker,
+            currentLeaderEpoch = currentFetchState.currentLeaderEpoch,
+            initOffset = currentFetchState.fetchOffset)
+          topicPartition -> initialFetchState
+        }
+        addFetcherForPartitions(fetchStates)
       }
     }
     lock synchronized {
@@ -153,14 +159,16 @@ abstract class AbstractFetcherManager[T <: AbstractFetcherThread](val name: Stri
     info(s"Added fetcher to broker ${fetcherThread.sourceBroker.id} for partitions $initialOffsetAndEpochs")
   }
 
-  def removeFetcherForPartitions(partitions: Set[TopicPartition]): Unit = {
+  def removeFetcherForPartitions(partitions: Set[TopicPartition]): Map[TopicPartition, PartitionFetchState] = {
+    val fetchStates = mutable.Map.empty[TopicPartition, PartitionFetchState]
     lock synchronized {
       for (fetcher <- fetcherThreadMap.values)
-        fetcher.removePartitions(partitions)
+        fetchStates ++= fetcher.removePartitions(partitions)
       failedPartitions.removeAll(partitions)
     }
     if (partitions.nonEmpty)
       info(s"Removed fetcher for partitions $partitions")
+    fetchStates
   }
 
   def shutdownIdleFetcherThreads(): Unit = {
--- a/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
+++ b/core/src/main/scala/kafka/server/AbstractFetcherManager.scala
@@ -142,17 +142,13 @@ abstract class AbstractFetcherManager[T <: AbstractFetcherThread](val name: Stri
             addAndStartFetcherThread(brokerAndFetcherId, brokerIdAndFetcherId)
         }
 
-        val initialOffsetAndEpochs = initialFetchOffsets.map { case (tp, brokerAndInitOffset) =>
-          tp -> OffsetAndEpoch(brokerAndInitOffset.initOffset, brokerAndInitOffset.currentLeaderEpoch, brokerAndInitOffset.lastFetchedEpoch)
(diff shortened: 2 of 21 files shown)


core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -432,14 +455,22 @@ abstract class AbstractFetcherThread(name: String,
       failedPartitions.removeAll(initialFetchStates.keySet)
 
       initialFetchStates.forKeyValue { (tp, initialFetchState) =>
-        // We can skip the truncation step iff the leader epoch matches the existing epoch
+        // For IBP 2.7 onwards, we can rely on truncation based on diverging data returned in fetch responses.
+        // For older versions, we can skip the truncation step iff the leader epoch matches the existing epoch
         val currentState = partitionStates.stateValue(tp)
-        val updatedState = if (currentState != null && currentState.currentLeaderEpoch == initialFetchState.leaderEpoch) {
+        val updatedState = if (initialFetchState.offset >= 0 && isTruncationOnFetchSupported && initialFetchState.lastFetchedEpoch.nonEmpty) {
```
rite2nikhil: NIT: Should this be the first check in the if () statement ?

rajinisivaram: @rite2nikhil Thanks for the review. Did you mean changing this to:
```
if (isTruncationOnFetchSupported && initialFetchState.lastFetchedEpoch.nonEmpty && initialFetchState.offset >= 0)
```

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -813,8 +852,9 @@ case class OffsetTruncationState(offset: Long, truncationCompleted: Boolean) {
   override def toString: String = "offset:%d-truncationCompleted:%b".format(offset, truncationCompleted)
 }
 
-case class OffsetAndEpoch(offset: Long, leaderEpoch: Int) {
+case class OffsetAndEpoch(offset: Long, leaderEpoch: Int, lastFetchedEpoch: Option[Int] = None) {
```
hachikuji: Wondering if it might be better not to change this type since it is used in contexts where `lastFetchedEpoch` is not relevant. Following the types through here, we first have use `InitialFetchState` in `AbstractFetcherManager`:
```scala
def addFetcherForPartitions(partitionAndOffsets: Map[TopicPartition, InitialFetchState])
```
We then convert to `OffsetAndEpoch` which gets passed down to `AbstractFetcherThread`:
```scala
def addPartitions(initialFetchStates: Map[TopicPartition, OffsetAndEpoch]): Set[TopicPartition]
```
Then this gets converted to `PartitionFetchState`. I wonder if it's possible to skip the conversion through `OffsetAndEpoch` and use `InitialFetchState` consistently? Perhaps the only reason the current code doesn't do that is that `InitialFetchState` includes the broker end point which is not really relevant to the fetcher thread. Maybe that's not such a big deal?

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -432,14 +455,22 @@ abstract class AbstractFetcherThread(name: String,
...
+        // For older versions, we can skip the truncation step iff the leader epoch matches the existing epoch
         val currentState = partitionStates.stateValue(tp)
-        val updatedState = if (currentState != null && currentState.currentLeaderEpoch == initialFetchState.leaderEpoch) {
+        val updatedState = if (initialFetchState.offset >= 0 && isTruncationOnFetchSupported && initialFetchState.lastFetchedEpoch.nonEmpty) {
+          if (currentState != null)
+            currentState
+          else
+            PartitionFetchState(initialFetchState.offset, None, initialFetchState.leaderEpoch,
+              state = Fetching, initialFetchState.lastFetchedEpoch)
+        } else if (currentState != null && (currentState.currentLeaderEpoch == initialFetchState.leaderEpoch)) {
```
hachikuji: nit: unnecessary parenthesis

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -225,6 +227,20 @@ abstract class AbstractFetcherThread(name: String,
...
   }
 
+  private def truncateOnFetchResponse(responseData: Map[TopicPartition, FetchData]): Unit = {
+    val epochEndOffsets = responseData
+      .filter { case (tp, fetchData) => fetchData.error == Errors.NONE && fetchData.divergingEpoch.isPresent }
+      .map { case (tp, fetchData) =>
+        val divergingEpoch = fetchData.divergingEpoch.get
+        tp -> new EpochEndOffset(Errors.NONE, divergingEpoch.epoch, divergingEpoch.endOffset)
+      }.toMap
+    inLock(partitionMapLock) {
```
hachikuji, line 231: Borderline overkill perhaps, but we could check if `epochEndOffsets` is non-empty before acquiring the lock

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -225,6 +227,20 @@ abstract class AbstractFetcherThread(name: String,
     }
   }
 
+  private def truncateOnFetchResponse(responseData: Map[TopicPartition, FetchData]): Unit = {
+    val epochEndOffsets = responseData
```
hachikuji: Rather than doing an additional pass over the response partitions, would it be reasonable to build `epochEndOffsets` inline with the other error handling in `processFetchRequest`?

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -408,9 +428,12 @@ abstract class AbstractFetcherThread(name: String,
   def markPartitionsForTruncation(topicPartition: TopicPartition, truncationOffset: Long): Unit = {
     partitionMapLock.lockInterruptibly()
     try {
+      // It is safe to reset `lastFetchedEpoch` here since we don't expect diverging offsets
```
hachikuji: Can be done separately, but it would be nice to figure out how to move this logic into `ReplicaAlterLogDirManager` since this comment seems to only make sense if we assume this is the log dir fetcher and reconciliation with the leader has already completed. 

In fact, I wonder if it is possible to get rid of this code entirely. If the log dir fetcher is also tracking `lastFetchedEpoch`, then we could rely on detecting truncation dynamically through `ReplicaManager.fetchMessages` instead of the current somewhat clumsy coordination with the replica fetcher.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -432,14 +455,22 @@ abstract class AbstractFetcherThread(name: String,
...
 
       initialFetchStates.forKeyValue { (tp, initialFetchState) =>
-        // We can skip the truncation step iff the leader epoch matches the existing epoch
+        // For IBP 2.7 onwards, we can rely on truncation based on diverging data returned in fetch responses.
+        // For older versions, we can skip the truncation step iff the leader epoch matches the existing epoch
         val currentState = partitionStates.stateValue(tp)
-        val updatedState = if (currentState != null && currentState.currentLeaderEpoch == initialFetchState.leaderEpoch) {
+        val updatedState = if (initialFetchState.offset >= 0 && isTruncationOnFetchSupported && initialFetchState.lastFetchedEpoch.nonEmpty) {
+          if (currentState != null)
+            currentState
```
hachikuji: Below we only use `currentState` if the current epoch matches the initial epoch. Why is it safe to skip that check here?

core/src/main/scala/kafka/server/DelayedFetch.scala
```diff
@@ -77,6 +78,7 @@ class DelayedFetch(delayMs: Long,
    * Case E: This broker is the leader, but the requested epoch is now fenced
    * Case F: The fetch offset locates not on the last segment of the log
    * Case G: The accumulated bytes from all the fetching partitions exceeds the minimum bytes
+   * Case H: A diverging epoch was found, return response to trigger truncation
```
hachikuji: Good catch here and in `FetchSession`. Do you think we should consider doing these fixes separately so that we can get them into 2.7? Otherwise it might be difficult to tie this behavior to the 2.7 IBP.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: I had initially added another class because I didn't want to change `OffsetAndEpoch`, but I removed that because it looked like too many similar classes. Your suggestion to use InitialFetchState sounds much better, updated.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: Updated. The method is still there for older versions, but it is now disabled with IBP 2.7.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: I refactored this code a bit and added a check for Fetching state. Not sure if I have missed something though. I think we can continue to fetch without truncating if currentState is Fetching when `lastFetchedEpoch` is known. If we need to truncate, we will do that later when we get told about diverging epochs. Does that make sense?

core/src/main/scala/kafka/server/DelayedFetch.scala
rajinisivaram: Makes sense, will submit another PR with just those changes.

rajinisivaram:
@hachikuji Thanks for the review, have addressed the comments so far.

core/src/main/scala/kafka/server/DelayedFetch.scala
rajinisivaram: PR with the changes in DelayedFetch and FetchSession: https://github.com/apache/kafka/pull/9434

rajinisivaram: PR with the changes in DelayedFetch and FetchSession: https://github.com/apache/kafka/pull/9434

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rite2nikhil: yes

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -426,21 +454,42 @@ abstract class AbstractFetcherThread(name: String,
...
+    if (isTruncationOnFetchSupported && initialFetchState.initOffset >= 0 && initialFetchState.lastFetchedEpoch.nonEmpty) {
+      if (currentState == null) {
+        return PartitionFetchState(initialFetchState.initOffset, None, initialFetchState.currentLeaderEpoch,
+          state = Fetching, initialFetchState.lastFetchedEpoch)
+      }
+      // If we are in `Fetching` state can continue to fetch regardless of current leader epoch and truncate
+      // if necessary based on diverging epochs returned by the leader. If we are currently in Truncating state,
+      // fall through and handle based on current epoch.
+      if (currentState.state == Fetching) {
+        return currentState
```
hachikuji: Is it not possible that the `InitialFetchState` has a bump to the current leader epoch? We will still need the latest epoch in order to continue fetching.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -461,8 +510,9 @@ abstract class AbstractFetcherThread(name: String,
         val maybeTruncationComplete = fetchOffsets.get(topicPartition) match {
           case Some(offsetTruncationState) =>
             val state = if (offsetTruncationState.truncationCompleted) Fetching else Truncating
```
hachikuji: Do we need to adjust this? I think we want to remain in the `Fetching` state if truncation detection is through `Fetch`.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -461,8 +510,9 @@ abstract class AbstractFetcherThread(name: String,
...
+            // Resetting `lastFetchedEpoch` since we are truncating and don't expect diverging epoch in the next fetch
```
hachikuji: This is a little unclear to me. I guess it is safe to reset `lastFetchedEpoch` as long as we reinitialize it after the next leader change. On the other hand, it seems safer to always retain the value.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -629,7 +680,9 @@ abstract class AbstractFetcherThread(name: String,
 
       val initialLag = leaderEndOffset - offsetToFetch
       fetcherLagStats.getAndMaybePut(topicPartition).lag = initialLag
-      PartitionFetchState(offsetToFetch, Some(initialLag), currentLeaderEpoch, state = Fetching)
+      // We don't expect diverging epochs from the next fetch request, so resetting `lastFetchedEpoch`
```
hachikuji: Again it seems safe to keep `lastFetchedEpoch` in sync with the local log. If we have done a full truncation above, then `lastFetchedEpoch` will be `None`, but otherwise it seems like we should set it.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -770,7 +770,7 @@ class ReplicaManager(val config: KafkaConfig,
             logManager.abortAndPauseCleaning(topicPartition)
 
             val initialFetchState = InitialFetchState(BrokerEndPoint(config.brokerId, "localhost", -1),
-              partition.getLeaderEpoch, futureLog.highWatermark)
+              partition.getLeaderEpoch, futureLog.highWatermark, lastFetchedEpoch = None)
```
hachikuji: Do we need to initialize `lastFetchedEpoch`? It seems like the log may not be empty at this point.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -341,11 +352,18 @@ abstract class AbstractFetcherThread(name: String,
                       // ReplicaDirAlterThread may have removed topicPartition from the partitionStates after processing the partition data
                       if (validBytes > 0 && partitionStates.contains(topicPartition)) {
                         // Update partitionStates only if there is no exception during processPartitionData
-                        val newFetchState = PartitionFetchState(nextOffset, Some(lag), currentFetchState.currentLeaderEpoch, state = Fetching)
+                        val newFetchState = PartitionFetchState(nextOffset, Some(lag),
+                          currentFetchState.currentLeaderEpoch, state = Fetching,
+                          Some(currentFetchState.currentLeaderEpoch))
```
hachikuji: This doesn't seem right. The last fetched epoch is supposed to represent the epoch of the last fetched batch. The fetcher could be fetching the data from an older epoch here.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: Updated.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: Updated.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: Updated

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: Oops, fixed.

core/src/main/scala/kafka/server/ReplicaManager.scala
rajinisivaram: I had reset this because `ReassignPartitionsIntegrationTest.testAlterLogDirReassignmentThrottle` was failing consistently when the offset was test. Having spent a whole day looking at what I had broken, I think it is an existing issue. The test itself looks new and I can recreate it on trunk. Looks like https://issues.apache.org/jira/browse/KAFKA-9087. Will look into that separately. For now, I have updated this path.

rajinisivaram:
@hachikuji Thanks for the review, have addressed the comments.

core/src/main/scala/kafka/server/ReplicaManager.scala
rajinisivaram: Looking at this again, I think a bit more work is required to set the offsets and epoch correctly for AlterLogDirsThread in order to use `lastFetchedEpoch`. So I have reverted the changes for ReplicaAlterLogDirsThread. Will do that in a follow-on PR instead. In this PR, we will use the old truncation path in this case.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1388,6 +1390,7 @@ class Log(@volatile private var _dir: File,
     var validBytesCount = 0
     var firstOffset: Option[Long] = None
     var lastOffset = -1L
+    var lastLeaderEpoch: Option[Int] = None
```
hachikuji: nit: not sure how much it matters, but maybe we can avoid the extra garbage and just use an integer until we're ready to build the result?

core/src/main/scala/kafka/server/ReplicaManager.scala
hachikuji: Sounds fair.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1667,8 +1667,9 @@ class ReplicaManager(val config: KafkaConfig,
         val partitionsToMakeFollowerWithLeaderAndOffset = partitionsToMakeFollower.map { partition =>
           val leader = metadataCache.getAliveBrokers.find(_.id == partition.leaderReplicaIdOpt.get).get
             .brokerEndPoint(config.interBrokerListenerName)
-          val fetchOffset = partition.localLogOrException.highWatermark
-          partition.topicPartition -> InitialFetchState(leader, partition.getLeaderEpoch, fetchOffset)
+          val log = partition.localLogOrException
+          val (fetchOffset, lastFetchedEpoch) = initialFetchOffsetAndEpoch(log)
+          partition.topicPartition -> InitialFetchState(leader, partition.getLeaderEpoch, fetchOffset, lastFetchedEpoch)
```
hachikuji: nit: line below is misaligned

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -426,21 +451,34 @@ abstract class AbstractFetcherThread(name: String,
...
+   * From IBP 2.7 onwards, we can rely on truncation based on diverging data returned in fetch responses.
+   * For older versions, we can skip the truncation step iff the leader epoch matches the existing epoch.
+   */
+  private def partitionFetchState(tp: TopicPartition, initialFetchState: InitialFetchState, currentState: PartitionFetchState): PartitionFetchState = {
+    if (currentState != null && currentState.currentLeaderEpoch == initialFetchState.currentLeaderEpoch) {
+      currentState
+    } else if (isTruncationOnFetchSupported && initialFetchState.initOffset >= 0 && initialFetchState.lastFetchedEpoch.nonEmpty &&
+              (currentState == null || currentState.state == Fetching)) {
+      PartitionFetchState(initialFetchState.initOffset, None, initialFetchState.currentLeaderEpoch,
+          state = Fetching, initialFetchState.lastFetchedEpoch)
```
hachikuji: I am wondering in what situation we would find `currentState` non-null here. The current logic in `ReplicaManager.makeFollowers` always calls `removeFetcherForPartitions` before adding the partition back. The reason I ask is that I wasn't sure we should be taking the last fetched epoch from the initial state or if we should keep the current one. It seems like the latter might be more current?

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -221,7 +223,15 @@ abstract class AbstractFetcherThread(name: String,
...
-      updateFetchOffsetAndMaybeMarkTruncationComplete(fetchOffsets)
+      updateFetchOffsetAndMaybeMarkTruncationComplete(fetchOffsets, isTruncationOnFetchSupported)
+    }
+  }
+
+  private def truncateOnFetchResponse(epochEndOffsets: Map[TopicPartition, EpochEndOffset]): Unit = {
+    inLock(partitionMapLock) {
+      val ResultWithPartitions(fetchOffsets, partitionsWithError) = maybeTruncateToEpochEndOffsets(epochEndOffsets, Map.empty)
+      handlePartitionsWithErrors(partitionsWithError, "truncateOnFetchResponse")
+      updateFetchOffsetAndMaybeMarkTruncationComplete(fetchOffsets, maySkipTruncation = false)
```
hachikuji: It's not clear to me why we set `maySkipTruncation` to false here. If the truncation is not complete, wouldn't that put us in the `Truncating` state?

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -454,15 +492,23 @@ abstract class AbstractFetcherThread(name: String,
...
+                                                              maySkipTruncation: Boolean): Unit = {
     val newStates: Map[TopicPartition, PartitionFetchState] = partitionStates.partitionStateMap.asScala
       .map { case (topicPartition, currentFetchState) =>
         val maybeTruncationComplete = fetchOffsets.get(topicPartition) match {
           case Some(offsetTruncationState) =>
-            val state = if (offsetTruncationState.truncationCompleted) Fetching else Truncating
+            val (state, lastFetchedEpoch) = if (offsetTruncationState.truncationCompleted)
+              (Fetching, latestEpoch(topicPartition))
+            else if (maySkipTruncation && currentFetchState.lastFetchedEpoch.nonEmpty)
+              (Fetching, currentFetchState.lastFetchedEpoch)
```
hachikuji: I'm a little uncertain about this case. If we have truncated to an earlier offset, wouldn't we also need to reset last fetched epoch? I am thinking we should remove this check and modify the first one:
```scala
val (state, lastFetchedEpoch) = if (maySkipTruncation || offsetTruncationState.truncationCompleted)
  (Fetching, latestEpoch(topicPartition))
```
I might be missing something though.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: I couldn't come up with a case where `currentState` is non-null, so I removed that check. Haven't seen any test failures as a result, so hopefully that is ok.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: I was trying to to optimize for the case where lastFetchEpoch didn't need to be reset, but you are right, it would need resetting after truncation. Updated to use `latestEpoch(topicPartition))`.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: Fixed, removed that flag.

rajinisivaram:
@hachikuji Thanks for the reviews, addressed the comments.

rajinisivaram:
Ran system tests on the latest version, there were 7 failures which look like flaky tests that also fail on trunk (one TransactionTest and 6 variants of ConnectDistributedTest).

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
hachikuji: This check is a still a little hard to follow. I think we expect that if `initOffset` is negative, then `lastFetchedEpoch` will be empty and we will hit the `fetchOffsetAndTruncate` case below. Is that right? On the other hand, if `lastFetchedEpoch` is empty, then `initOffset` could still be non-negative if we have an old message format, which means we need to enter `Truncating` so that we can truncate to the high watermark. 

 One case that is not so clear is when `currentState` is non-null. Then we will enter the `Truncating` state below regardless whether `isTruncationOnFetchSupported` is set or not. Is that what we want?

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -669,11 +714,18 @@ abstract class AbstractFetcherThread(name: String,
     Option(partitionStates.stateValue(topicPartition))
   }
 
+  /**
+   * Returns current fetch state for each partition assigned to this thread. This is used to reassign
+   * partitions when thread pool is resized. We return `lastFetchedEpoch=None` to ensure we go through
```
hachikuji: This is probably ok. I guess an alternative would be to not take the initial last fetched epoch from `InitialFetchState`, but instead use `latestEpoch`.

core/src/main/scala/kafka/server/ReplicaFetcherThread.scala
```diff
@@ -102,6 +103,7 @@ class ReplicaFetcherThread(name: String,
   private val maxBytes = brokerConfig.replicaFetchResponseMaxBytes
   private val fetchSize = brokerConfig.replicaFetchMaxBytes
   private val brokerSupportsLeaderEpochRequest = brokerConfig.interBrokerProtocolVersion >= KAFKA_0_11_0_IV2
+  private val brokerSupportsTruncationOnFetch = ApiVersion.isTruncationOnFetchSupported(brokerConfig.interBrokerProtocolVersion)
```
hachikuji: nit: I don't think we need this. We can override `isTruncationOnFetchSupported` with a `val`

hachikuji (commented):
@rajinisivaram Thanks, I think we're getting close. I left a few small comments.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: @hachikuji Thanks for the review! You mentioned in another comment below that we could use latestEpoch instead of using the epoch from InitialFetchOffset. I have removed lastFetchEpoch from InitialFetchOffset and updated this logic to use latestEpoch. It does look neater now, hope I haven't missed any cases. I had to update some of the unit tests which rely on the initial fetch epoch request to use IBP 2.6, but all tests have passed in my local run. I will kick off a system test run as well.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: Updated.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -426,21 +451,35 @@ abstract class AbstractFetcherThread(name: String,
...
+   * For older versions, we can skip the truncation step iff the leader epoch matches the existing epoch.
+   */
+  private def partitionFetchState(tp: TopicPartition, initialFetchState: InitialFetchState, currentState: PartitionFetchState): PartitionFetchState = {
+    if (currentState != null && currentState.currentLeaderEpoch == initialFetchState.currentLeaderEpoch) {
+      currentState
+    } else if (initialFetchState.initOffset < 0) {
+      fetchOffsetAndTruncate(tp, initialFetchState.currentLeaderEpoch)
+    } else if (isTruncationOnFetchSupported) {
+      val lastFetchedEpoch = latestEpoch(tp)
+      val state = if (lastFetchedEpoch.exists(_ != EpochEndOffset.UNDEFINED_EPOCH)) Fetching else Truncating
```
hachikuji: Hmm.. Do we actually return `Some(EpochEndOffset.UNDEFINED_EPOCH)` from `latestEpoch`? That seems surprising.

Might be worth a comment here that we still go through the `Truncating` state here when the message format is old.

core/src/main/scala/kafka/server/AbstractFetcherManager.scala
```diff
@@ -64,8 +64,8 @@ abstract class AbstractFetcherManager[T <: AbstractFetcherThread](val name: Stri
   def resizeThreadPool(newSize: Int): Unit = {
     def migratePartitions(newSize: Int): Unit = {
       fetcherThreadMap.forKeyValue { (id, thread) =>
-        val removedPartitions = thread.partitionsAndOffsets
-        removeFetcherForPartitions(removedPartitions.keySet)
+        val removedPartitions = thread.removeAllPartitions()
+        removeFetcherForPartitions(removedPartitions.keySet) // clear state for removed partitions
```
hachikuji: This reads a bit odd following `removeAllPartitions`. I guess what we get from `removeFetcherForPartitions` is the clearing of `failedPartitions` and de-registration from `fetcherLagStats`. Not super important, but wonder if it's worth trying to consolidate a little. Maybe `removeFetcherForPartitions` could return the initial fetch offsets or something.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1691,6 +1692,18 @@ class ReplicaManager(val config: KafkaConfig,
     partitionsToMakeFollower
   }
 
+  /**
+   * From IBP 2.7 onwards, we send latest fetch epoch in the request and truncate if a
+   * diverging epoch is returned in the response, avoiding the need for a separate
+   * OffsetForLeaderEpoch request.
+   */
+  private def initialFetchOffset(log: Log): Long = {
```
hachikuji, line 1700: I think this could be saved for a follow-up, but I wonder if we should consider similarly letting the initial offset be determined by the fetcher thread on initialization rather than being passed in. I find it confusing that we expect this to be the high watermark in some cases. It seems a little slippery the way we rely on it in `AbstractFetcherThread.truncateToHighWatermark`.

core/src/test/scala/unit/kafka/server/ReplicaFetcherThreadTest.scala
```diff
@@ -453,6 +466,107 @@ class ReplicaFetcherThreadTest {
...
+    // Create the fetcher thread
+    val mockNetwork = new ReplicaFetcherMockBlockingSend(Collections.emptyMap(), brokerEndPoint, new SystemTime())
+    val thread = new ReplicaFetcherThread("bob", 0, brokerEndPoint, config, failedPartitions, replicaManager, new Metrics(), new SystemTime(), quota, Some(mockNetwork)) {
+      override def processPartitionData(topicPartition: TopicPartition, fetchOffset: Long, partitionData: FetchData): Option[LogAppendInfo] = None
+    }
+    thread.addPartitions(Map(t1p0 -> initialFetchState(initialLEO), t1p1 -> initialFetchState(initialLEO)))
+    val partitions = Set(t1p0, t1p1)
+
+    // Loop 1 -- both topic partitions skip epoch fetch and send fetch request since
+    // lastFetchedEpoch is set in initial fetch state.
```
hachikuji: nit: needs update?

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
rajinisivaram: I was being lazy with that check because we were using `Some(EpochEndOffset.UNDEFINED_EPOCH)` in AbstractFetcherThreadTest. I have updated the test and fixed the check. Added comment as well.

core/src/main/scala/kafka/server/AbstractFetcherManager.scala
rajinisivaram: Updated.

core/src/main/scala/kafka/server/ReplicaManager.scala
rajinisivaram: Yes, makes sense, will do that in a follow-up PR.

rajinisivaram:
@hachikuji Thanks for the review, I have addressed the comments.

rajinisivaram:
The system test run from this morning on this PR branch completed with 6 failures, which are known failures in ConnectDistributedTest that also failed in the last trunk test run.

hachikuji (approved):
@rajinisivaram LGTM. Thanks for the PR! Note I pushed a trivial tweak. I will merge once the build completes.

rajinisivaram:
@hachikuji Thanks for all your help and patience in reviewing this PR! Merging to trunk.

PR #9383, opened 2020-10-06
KAFKA-10455: Ensure that probing rebalances always occur

lct45:
Add data to subscriptionUserData to make sure that it's different each time a consumer rejoins

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
@@ -236,7 +236,7 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
         // Adds the following information to subscription
         // 1. Client UUID (a unique id assigned to an instance of KafkaStreams)
         // 2. Map from task id to its overall lag
-        // 3. Unique Field to ensure a rebalance when a thread rejoins
+        // 3. Unique Field to ensure a rebalance when a thread rejoins by forcing the user data to be different
 
         handleRebalanceStart(topics);
         if (usedSubscriptionMetadataVersion >= 8) {
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
@@ -28,6 +28,7 @@ import org.apache.kafka.common.Node;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.TimeoutException;
+import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.common.utils.LogContext;
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.common.utils.Utils;
@@ -183,6 +184,8 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
 
     private Supplier<TaskAssignor> taskAssignorSupplier;
 
+    private byte[] uniqueField;
+
     /**
      * We need to have the PartitionAssignor and its StreamThread to be mutually accessible since the former needs
      * later's cached metadata while sending subscriptions, and the latter needs former's returned assignment when
@@ -212,6 +215,7 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
         rebalanceProtocol = assignorConfiguration.rebalanceProtocol();
         taskAssignorSupplier = assignorConfiguration::taskAssignor;
         assignmentListener = assignorConfiguration.assignmentListener();
+        uniqueField = Bytes.EMPTY;
     }
 
     @Override
@@ -234,15 +238,19 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
         // Adds the following information to subscription
         // 1. Client UUID (a unique id assigned to an instance of KafkaStreams)
         // 2. Map from task id to its overall lag
+        // 3. Unique Field to ensure a rebalance when a thread rejoins
 
         handleRebalanceStart(topics);
+
+        uniqueField[0]++;
 
         return new SubscriptionInfo(
             usedSubscriptionMetadataVersion,
             LATEST_SUPPORTED_VERSION,
             taskManager.processId(),
             userEndPoint,
-            taskManager.getTaskOffsetSums())
+            taskManager.getTaskOffsetSums(),
+            uniqueField)
                 .encode();
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
@@ -182,7 +182,6 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
     private AssignmentListener assignmentListener;
 
     private Supplier<TaskAssignor> taskAssignorSupplier;
-
     private byte[] uniqueField;
 
     /**
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
@@ -215,7 +215,6 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
         taskAssignorSupplier = assignorConfiguration::taskAssignor;
         assignmentListener = assignorConfiguration.assignmentListener();
         uniqueField = usedSubscriptionMetadataVersion >= 8 ? new byte[1] : new byte[0];
-
     }
 
     @Override
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
@@ -1590,6 +1590,10 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
         return taskManager;
     }
 
+    protected byte[] uniqueField() {
+        return uniqueField;
+    }
+
     protected void handleRebalanceStart(final Set<String> topics) {
         taskManager.handleRebalanceStart(topics);
     }
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
@@ -213,7 +213,7 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
         rebalanceProtocol = assignorConfiguration.rebalanceProtocol();
         taskAssignorSupplier = assignorConfiguration::taskAssignor;
         assignmentListener = assignorConfiguration.assignmentListener();
-        uniqueField = usedSubscriptionMetadataVersion >= 8 ? new byte[1] : new byte[0];
+        uniqueField = new byte[1];
     }
 
     @Override
@@ -239,9 +239,8 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
         // 3. Unique Field to ensure a rebalance when a thread rejoins by forcing the user data to be different
 
         handleRebalanceStart(topics);
-        if (usedSubscriptionMetadataVersion >= 8) {
-            uniqueField[0]++;
-        }
+        uniqueField[0]++;
+
 
         return new SubscriptionInfo(
             usedSubscriptionMetadataVersion,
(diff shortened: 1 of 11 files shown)


lct45:
@ableegoldman ready for initial review

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -212,6 +213,7 @@ public void configure(final Map<String, ?> configs) {
         rebalanceProtocol = assignorConfiguration.rebalanceProtocol();
         taskAssignorSupplier = assignorConfiguration::taskAssignor;
         assignmentListener = assignorConfiguration.assignmentListener();
+        uniqueField = usedSubscriptionMetadataVersion >= 8 ? new byte[1] : new byte[0];
```
lct45: This could be moved out to the `assignorConfiguration` class to match the other fields but the code is fairly straightforward so I left it here for now

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -234,15 +236,20 @@ public ByteBuffer subscriptionUserData(final Set<String> topics) {
         // Adds the following information to subscription
         // 1. Client UUID (a unique id assigned to an instance of KafkaStreams)
         // 2. Map from task id to its overall lag
+        // 3. Unique Field to ensure a rebalance when a thread rejoins by forcing the user data to be different
 
         handleRebalanceStart(topics);
+        if (usedSubscriptionMetadataVersion >= 8) {
```
ableegoldman: Instead of hardcoded `8` all over, let's just define a constant for this similar to the `MIN_VERSION_OFFSET_SUM_SUBSCRIPTION ` in SubscriptionInfo

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
ableegoldman: Sounds good. I think it's also fine to just initialize it to `new byte[1]` regardless of whether the version is high enough to actually need it or not. It's just a single byte, and simpler is better

ableegoldman: Also, now that I think about it, the `usedSubscriptionMetadataVersion` will only ever be >= 8 at this point. It might be set to something lower at some point later on, but it has to be at least 8 when the assignor is just being created/configured

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
ableegoldman: If you want, I think it's also fine to just always flip the byte and not even check against the used subscription version.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignorTest.java
```diff
@@ -483,7 +485,7 @@ public void testEagerSubscription() {
         Collections.sort(subscription.topics());
         assertEquals(asList("topic1", "topic2"), subscription.topics());
 
-        final SubscriptionInfo info = getInfo(UUID_1, prevTasks, standbyTasks);
```
ableegoldman, line 486: Can you add a test that verifies that it goes back and forth between the two expected values when you call `partitionAssignor.subscriptionUserData` multiples times (let's say 3) -- also maybe add a verification that the `uniqueField` has a length of just 1

streams/src/test/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfoTest.java
```diff
@@ -286,23 +296,23 @@ public void shouldEncodeAndDecodeSmallerLatestSupportedVersion() {
...
-            new SubscriptionInfo(usedVersion, latestSupportedVersion, UUID_1, "localhost:80", TASK_OFFSET_SUMS);
+            new SubscriptionInfo(usedVersion, latestSupportedVersion, UUID_1, "localhost:80", TASK_OFFSET_SUMS, IGNORED_UNIQUE_FIELD);
         final SubscriptionInfo expectedInfo =
-            new SubscriptionInfo(usedVersion, latestSupportedVersion, UUID_1, "localhost:80", TASK_OFFSET_SUMS);
+            new SubscriptionInfo(usedVersion, latestSupportedVersion, UUID_1, "localhost:80", TASK_OFFSET_SUMS, IGNORED_UNIQUE_FIELD);
         assertEquals(expectedInfo, SubscriptionInfo.decode(info.encode()));
     }
 
     @Test
     public void shouldEncodeAndDecodeVersion7() {
```
ableegoldman, line 305: Can you add a test like this for the new version 8?

streams/src/test/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfoTest.java
```diff
@@ -313,7 +323,8 @@ public void shouldReturnTaskOffsetSumsMapForDecodedSubscription() {
             new SubscriptionInfo(MIN_VERSION_OFFSET_SUM_SUBSCRIPTION,
                                  LATEST_SUPPORTED_VERSION, UUID_1,
                                  "localhost:80",
-                                 TASK_OFFSET_SUMS)
+                                 TASK_OFFSET_SUMS,
+                                 IGNORED_UNIQUE_FIELD)
                 .encode());
         assertThat(info.taskOffsetSums(), is(TASK_OFFSET_SUMS));
     }
```
ableegoldman, line 329: I think we should also add a test to make sure that if you pass in a `uniqueField` to the SubscriptionInfo but the `usedVersion` is less than 8, that it does not actually encode this field.

ableegoldman (commented):
Can you also kick off the system tests?  We always break the version probing test in particular by accident

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
lct45: The autocode generator throws an error if `array.length != 0` and the version is < 8. It didn't seem to be a huge problem but there were some tests that used version 1 so initializing to `new byte[1]` threw errors

ableegoldman: That seems like an indication that the subscription is being encoded incorrectly. If the version is below 8, it shouldn't be trying to encode the `uniqueField` in the first place -- you shouldn't have to mimic that yourself by forcing it to be empty (and even if it is I worry there might be some additional bytes for metadata that would get stored for that field, when that field should not exist at all for versions < 8)

lct45:
System tests: https://jenkins.confluent.io/job/system-test-kafka-branch-builder/4204/

Looks like AVE_MEMFD_CREATE failed

streams/src/test/java/org/apache/kafka/streams/integration/HighAvailabilityTaskAssignorIntegrationTest.java
```diff
@@ -286,7 +286,8 @@ private static Properties streamsProperties(final String appId,
                 mkEntry(StreamsConfig.PROBING_REBALANCE_INTERVAL_MS_CONFIG, "60000"),
                 mkEntry(StreamsConfig.InternalConfig.ASSIGNMENT_LISTENER, configuredAssignmentListener),
                 mkEntry(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, "100"),
-                mkEntry(StreamsConfig.InternalConfig.INTERNAL_TASK_ASSIGNOR_CLASS, HighAvailabilityTaskAssignor.class.getName())
+                mkEntry(StreamsConfig.InternalConfig.INTERNAL_TASK_ASSIGNOR_CLASS, HighAvailabilityTaskAssignor.class.getName()),
+                mkEntry(StreamsConfig.NUM_STREAM_THREADS_CONFIG, 40)
```
ableegoldman, line 291: Can we add a comment here explaining why we set the thread count so high? I feel like we'll forget and be really confused when we stumble across this in the future.

streams/src/test/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfoTest.java
```diff
@@ -59,6 +60,7 @@
     );
 
     private final static String IGNORED_USER_ENDPOINT = "ignoredUserEndpoint:80";
+    private static final byte[] IGNORED_UNIQUE_FIELD = Bytes.EMPTY;
```
ableegoldman: nit: let's use `new byte[1]` for this to make sure it's actually being ignored when it should be (since apparently it won't notice if you just pass in empty bytes for this field on a version < 8)

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignorTest.java
lct45: > goes back and forth between the two expected values 

What do you mean by expected values? Toggling between the default of `0` and `1` after `partitionAssignor.subscriptionUserData` is called the first time?

ableegoldman: Yeah, it should be `0` the first time you call it, then `1` the second time, and then back to `0` again on the third call

lct45: I don't think it would go from `0` to `1` to `0` since the partition assigner always increases, so if you're using the same partition assignor it'll go from `0` to `1` to `2` etc

lct45:
One test failing: testRollingBrokerRestartsWithSmallerMaxGroupSizeConfigDisruptsBigGroup

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignorTest.java
ableegoldman: Oh right duh I was thinking it was just a single bit but it's a byte. In that case we should have a test that verifies it goes from `0` to `1` to `2`, etc -- might be good to verify the behavior on overflow as well, if you call `subscriptionUserData` the max_value number of times

streams/src/test/java/org/apache/kafka/streams/tests/StreamsUpgradeTest.java
```diff
@@ -158,7 +159,8 @@ public ByteBuffer subscriptionUserData(final Set<String> topics) {
                     LATEST_SUPPORTED_VERSION + 1,
                     taskManager.processId(),
                     userEndPoint(),
-                    taskManager.getTaskOffsetSums()
+                    taskManager.getTaskOffsetSums(),
+                    Bytes.EMPTY
```
ableegoldman: We should do the same thing that we do in StreamsPartitionAssignor here, ie a changing byte. This is supposed to be like the same exact thing but for a "future version" of the partition assignor, to test version probing and general compatibility. 
It's a bit annoying that you have to just copy the same exact stuff but that's just how it is for now

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignorTest.java
ableegoldman: Oh, saw that you already added the test for `0` --> `1` --> `2` behavior. In that case can you just add two things to that test: verify the behavior on overflow, and verify that the length is always exactly one byte

lct45:
https://jenkins.confluent.io/job/system-test-kafka-branch-builder/4228/ system tests part 2

ableegoldman (approved):
Thanks for the fix!

vvcephei (commented):
Thanks for this PR, @lct45 !

It looks good to me, but I have one question... Why do we use a size-1 byte array instead of a single byte for this field?

Just for clarity, the reason I asked is that the encoding of a byte array is to store an int representing the length of the array, followed by the actual bytes. So a size-one byte array is stored as 5 bytes, while a single byte is stored as 1 byte.

lct45:
> Why do we use a size-1 byte array instead of a single byte for this field?

@vvcephei There's a `RecordsFieldType` already created for `byte[]` but not one for `byte` so to get the `subscriptionInfo.json` to work I would've had to create a `RecordsFieldType` for `byte` and decided that leveraging the existing field type would be easier. If you're strongly opposed to this solution I can look into creating a new field type

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignorTest.java
```diff
@@ -2023,13 +2022,13 @@ public void testUniqueField() {
         configureDefaultPartitionAssignor();
         final Set<String> topics = mkSet("input");
 
-        assertEquals(1, partitionAssignor.uniqueField().length);
-        assertEquals(0, partitionAssignor.uniqueField()[0]);
+        //assertEquals(1, partitionAssignor.uniqueField().length);
```
vvcephei: we probably ought to delete the commented-out assertions.

vvcephei (approved):
Hey @lct45 , 

Thanks for the update.

The tests passed, so I'm ready to merge, but can you just push one small commit to delete the commented-out assertions before I merge?

Thanks!

vvcephei:
Thanks for the PR, @lct45 !

vvcephei:

vvcephei:
Cherry-picked to 2.6

PR #9384, opened 2020-10-06
MINOR: remove explicit passing of AdminClient into StreamsPartitionAssignor

mjsax:
Currently, we pass multiple object reference (`AdminClient`,`TaskManager`, and `Time`) into `StreamsPartitionAssignor`. Furthermore, we (miss)use `TaskManager#mainConsumer()` to get access to the main consumer (we need to do this, to avoid a cyclic dependency).

This PR unifies how object references are passed into a single `ReferenceContainer` class to
 - not "miss use" the TaskManager as reference container
 - unify how object references are passes

Note: we need to use a reference container to avoid cyclic dependencies, instead of using a config for each passed reference individually.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
+++ b/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
@@ -902,7 +902,6 @@ public class StreamsConfig extends AbstractConfig {
         // state into the assignor.
         public static final String TASK_MANAGER_FOR_PARTITION_ASSIGNOR = "__task.manager.instance__";
         public static final String STREAMS_METADATA_STATE_FOR_PARTITION_ASSIGNOR = "__streams.metadata.state.instance__";
-        public static final String STREAMS_ADMIN_CLIENT = "__streams.admin.client.instance__";
         public static final String ASSIGNMENT_ERROR_CODE = "__assignment.error.code__";
         public static final String NEXT_SCHEDULED_REBALANCE_MS = "__next.probing.rebalance.ms__";
         public static final String TIME = "__time__";
--- a/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
+++ b/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
@@ -901,8 +901,6 @@ public class StreamsConfig extends AbstractConfig {
         // These are not settable in the main Streams config; they are set by the StreamThread to pass internal
         // state into the assignor.
         public static final String REFERENCE_CONTAINER_PARTITION_ASSIGNOR = "__reference.container.instance__";
-        public static final String ASSIGNMENT_ERROR_CODE = "__assignment.error.code__";
-        public static final String NEXT_SCHEDULED_REBALANCE_MS = "__next.probing.rebalance.ms__";
 
         // This is settable in the main Streams config, but it's a private API for testing
         public static final String ASSIGNMENT_LISTENER = "__assignment.listener__";
--- a/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
+++ b/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
@@ -903,7 +903,6 @@ public class StreamsConfig extends AbstractConfig {
         public static final String REFERENCE_CONTAINER_PARTITION_ASSIGNOR = "__reference.container.instance__";
         public static final String ASSIGNMENT_ERROR_CODE = "__assignment.error.code__";
         public static final String NEXT_SCHEDULED_REBALANCE_MS = "__next.probing.rebalance.ms__";
-        public static final String TIME = "__time__";
 
         // This is settable in the main Streams config, but it's a private API for testing
         public static final String ASSIGNMENT_LISTENER = "__assignment.listener__";
--- a/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
+++ b/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
@@ -900,8 +900,7 @@ public class StreamsConfig extends AbstractConfig {
 
         // These are not settable in the main Streams config; they are set by the StreamThread to pass internal
         // state into the assignor.
-        public static final String TASK_MANAGER_FOR_PARTITION_ASSIGNOR = "__task.manager.instance__";
-        public static final String STREAMS_METADATA_STATE_FOR_PARTITION_ASSIGNOR = "__streams.metadata.state.instance__";
+        public static final String REFERENCE_CONTAINER_PARTITION_ASSIGNOR = "__reference.container.instance__";
         public static final String ASSIGNMENT_ERROR_CODE = "__assignment.error.code__";
         public static final String NEXT_SCHEDULED_REBALANCE_MS = "__next.probing.rebalance.ms__";
         public static final String TIME = "__time__";
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -40,6 +40,7 @@ import org.apache.kafka.streams.processor.StateRestoreListener;
 import org.apache.kafka.streams.processor.TaskId;
 import org.apache.kafka.streams.processor.TaskMetadata;
 import org.apache.kafka.streams.processor.ThreadMetadata;
+import org.apache.kafka.streams.processor.internals.assignment.ReferenceContainer;
 import org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl;
 import org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl.Version;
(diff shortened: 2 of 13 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/AssignorConfiguration.java
```diff
@@ -291,11 +271,7 @@ public String userEndPoint() {
         }
     }
 
-    public Admin adminClient() {
-        return adminClient;
-    }
-
-    public InternalTopicManager internalTopicManager() {
+    public InternalTopicManager internalTopicManager(final Admin adminClient) {
```
mjsax: Instead of passing the `Admin` as parameter, we could call `taskManager#adminClient()` in the next line, however, this requires to make the method `public` (it's package-private atm) thus I opted to for parameter-passing instead.

streams/src/test/java/org/apache/kafka/streams/processor/internals/HighAvailabilityStreamsPartitionAssignorTest.java
```diff
@@ -131,12 +130,13 @@
...
     // Make sure to complete setting up any mocks (such as TaskManager or AdminClient) before configuring the assignor
     private void configurePartitionAssignorWith(final Map<String, Object> props) {
+        EasyMock.replay(taskManager, adminClient);
+
         final Map<String, Object> configMap = configProps();
         configMap.putAll(props);
 
         streamsConfig = new StreamsConfig(configMap);
         partitionAssignor.configure(configMap);
-        EasyMock.replay(taskManager, adminClient);
```
mjsax: We need to setup the mocks before calling `partitionAssignor.configure()` now, as we call `taskManager#adminClient()` in this method (similar elsewhere)

streams/src/test/java/org/apache/kafka/streams/processor/internals/HighAvailabilityStreamsPartitionAssignorTest.java
```diff
@@ -203,8 +204,8 @@ public void shouldReturnAllActiveTasksToPreviousOwnerRegardlessOfBalanceAndTrigg
         builder.addStateStore(new MockKeyValueStoreBuilder("store1", false), "processor1");
         final Set<TaskId> allTasks = mkSet(TASK_0_0, TASK_0_1, TASK_0_2);
 
-        createMockTaskManager(allTasks);
         adminClient = EasyMock.createMock(AdminClient.class);
+        createMockTaskManager(allTasks);
```
mjsax: We need to setup the admin mock before the TM mock now (similar elsewhere)

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -1091,8 +1096,4 @@ ConsumerRebalanceListener rebalanceListener() {
     Admin adminClient() {
         return adminClient;
     }
-
-    InternalTopologyBuilder internalTopologyBuilder() {
```
mjsax, line 1095: Remove unused method

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -192,12 +191,11 @@ public String toString() {
      */
     @Override
     public void configure(final Map<String, ?> configs) {
-        final AssignorConfiguration assignorConfiguration = new AssignorConfiguration(configs);
+        assignorConfiguration = new AssignorConfiguration(configs);
```
mjsax: we need to keep a reference as we cannot get the `mainConsumer` at this point, as it will be set after this method finished

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -206,7 +204,6 @@ public void configure(final Map<String, ?> configs) {
         assignmentConfigs = assignorConfiguration.assignmentConfigs();
         partitionGrouper = assignorConfiguration.partitionGrouper();
         userEndPoint = assignorConfiguration.userEndPoint();
-        adminClient = assignorConfiguration.adminClient();
```
mjsax, line 209: Only used at a single place, just removing the variable.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -120,10 +120,6 @@ void setMainConsumer(final Consumer<byte[], byte[]> mainConsumer) {
         this.mainConsumer = mainConsumer;
     }
 
-    Consumer<byte[], byte[]> mainConsumer() {
```
mjsax, line 123: Decouple TM from StreamPartitionAssignor now and not "miss-use" it to get the consumer any longer.

streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
```diff
@@ -900,9 +900,7 @@
 
         // These are not settable in the main Streams config; they are set by the StreamThread to pass internal
         // state into the assignor.
-        public static final String TASK_MANAGER_FOR_PARTITION_ASSIGNOR = "__task.manager.instance__";
-        public static final String STREAMS_METADATA_STATE_FOR_PARTITION_ASSIGNOR = "__streams.metadata.state.instance__";
-        public static final String STREAMS_ADMIN_CLIENT = "__streams.admin.client.instance__";
+        public static final String REFERENCE_CONTAINER_PARTITION_ASSIGNOR = "__reference.container.instance__";
         public static final String ASSIGNMENT_ERROR_CODE = "__assignment.error.code__";
         public static final String NEXT_SCHEDULED_REBALANCE_MS = "__next.probing.rebalance.ms__";
         public static final String TIME = "__time__";
```
vvcephei: Should these be in the reference container as well?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
vvcephei: Thanks!

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
vvcephei: Thanks!

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
vvcephei: Thanks. Should we instead make `referenceContainer` a field and access its members to get the references? The purpose of `AssignorConfiguration` is to parse the configs, not to be a general container for the configured values. Otherwise, we wouldn't need to assign any of the other fields here.

vvcephei (commented):
Thanks for this cleanup, @mjsax ! Most of my comments are wondering if we can push this cleanup a little farther.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
mjsax: Sure.

streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
mjsax: Ack

mjsax:
@vvcephei Updated this PR (including PR description).

streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
vvcephei: Sorry, @mjsax , I was referring to all three of "assignment error code", "next scheduled rebalance ms", and "time".

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
vvcephei: Thanks for the update! Are you still planning to drop the `assignorConfiguration` field in favor of a `referenceContainer` field?

mjsax: I never intended to drop it. Maybe I miss understand your comment?

We could replace the field `StreamsParttionAssignor#taskManager` etc with `StreamsPartitionAssigner#referenceContainer` but it just make the code lines longer each time we need to access the TM (etc). Thus, it seems to make the code more readable if we just "extract" those field from the reference container once?

streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
mjsax: Ah. Ack.

mjsax:
Address comments. @vvcephei

vvcephei (approved):
Thanks, @mjsax !

PR #9385, opened 2020-10-06
KAFKA-9274: fix incorrect default value for `task.timeout.ms` config

mjsax:
Part of KIP-572.

Also add handler method to trigger/reset the timeout on a task.

Call for review @vvcephei


vvcephei (approved):
Thanks, @mjsax !

vvcephei (approved):
Thanks for this, too @mjsax

PR #9386, opened 2020-10-06
KAFKA-10024: Add dynamic configuration and enforce quota for per-IP connection rate limits (KIP-612, part 2)

splett2:
This PR implements the part of KIP-612 for adding IP throttling enforcement, and a ZK entity for configuring dynamic IP throttles.

I will add `kafka-configs` support as well as `KafkaApi` reconfiguration support in a follow-up PR.

### DIFF
--- a/core/src/main/scala/kafka/network/SocketServer.scala
+++ b/core/src/main/scala/kafka/network/SocketServer.scala
@@ -1362,8 +1362,9 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
   private val listenerCounts = mutable.Map[ListenerName, Int]()
   private[network] val maxConnectionsPerListener = mutable.Map[ListenerName, ListenerConnectionQuota]()
   @volatile private var totalCount = 0
+  // updates to defaultConnectionRatePerIp or connectionRatePerIp must be synchronized on `counts`
   @volatile private var defaultConnectionRatePerIp = DynamicConfig.Ip.DefaultConnectionCreationRate
-  private val connectionRatePerIp = mutable.Map[InetAddress, Int]()
+  private val connectionRatePerIp = new ConcurrentHashMap[InetAddress, Int]()
   // sensor that tracks broker-wide connection creation rate and limit (quota)
   private val brokerConnectionRateSensor = getOrCreateConnectionRateQuotaSensor(config.maxConnectionCreationRate, BrokerQuotaEntity)
   private val maxThrottleTimeMs = TimeUnit.SECONDS.toMillis(config.quotaWindowSizeSeconds.toLong)
@@ -1414,7 +1415,7 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
    * @param ip ip to update or default if None
    * @param maxConnectionRate new connection rate, or resets entity to default if None
    */
-  def updateIpConnectionRateQuota(ip: Option[InetAddress], maxConnectionRate: Option[Int]): Unit = {
+  def updateIpConnectionRateQuota(ip: Option[InetAddress], maxConnectionRate: Option[Int]): Unit = synchronized {
     def isIpConnectionRateMetric(metricName: MetricName) = {
       metricName.name == ConnectionRateMetricName &&
       metricName.group == MetricsGroup &&
@@ -1424,9 +1425,9 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
     def shouldUpdateQuota(metric: KafkaMetric, quotaLimit: Int) = {
       quotaLimit != metric.config.quota.bound
     }
-    counts.synchronized {
-      ip match {
-        case Some(address) =>
+    ip match {
+      case Some(address) =>
+        counts.synchronized {
           maxConnectionRate match {
             case Some(rate) =>
               info(s"Updating max connection rate override for $address to $rate")
@@ -1435,26 +1436,28 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
               info(s"Removing max connection rate override for $address")
               connectionRatePerIp.remove(address)
           }
-          updateConnectionRateQuota(connectionRateForIp(address), IpQuotaEntity(address))
-        case None =>
+        }
+        updateConnectionRateQuota(connectionRateForIp(address), IpQuotaEntity(address))
+      case None =>
+        counts.synchronized {
           defaultConnectionRatePerIp = maxConnectionRate.getOrElse(DynamicConfig.Ip.DefaultConnectionCreationRate)
-          info(s"Updated default max IP connection rate to $defaultConnectionRatePerIp")
-          metrics.metrics.forEach { (metricName, metric) =>
-            if (isIpConnectionRateMetric(metricName)) {
-              val quota = connectionRateForIp(InetAddress.getByName(metricName.tags.get(IpMetricTag)))
-              if (shouldUpdateQuota(metric, quota)) {
-                info(s"Updating existing connection rate quota config for ${metricName.tags} to $quota")
-                metric.config(rateQuotaMetricConfig(quota))
-              }
+        }
+        info(s"Updated default max IP connection rate to $defaultConnectionRatePerIp")
+        metrics.metrics.forEach { (metricName, metric) =>
+          if (isIpConnectionRateMetric(metricName)) {
+            val quota = connectionRateForIp(InetAddress.getByName(metricName.tags.get(IpMetricTag)))
+            if (shouldUpdateQuota(metric, quota)) {
+              debug(s"Updating existing connection rate quota config for ${metricName.tags} to $quota")
+              metric.config(rateQuotaMetricConfig(quota))
             }
           }
-      }
+        }
     }
   }
 
   // Visible for testing
   def connectionRateForIp(ip: InetAddress): Int = {
-    connectionRatePerIp.getOrElse(ip, defaultConnectionRatePerIp)
+    connectionRatePerIp.getOrDefault(ip, defaultConnectionRatePerIp)
   }
 
   private[network] def addListener(config: KafkaConfig, listenerName: ListenerName): Unit = {
--- a/core/src/main/scala/kafka/network/SocketServer.scala
+++ b/core/src/main/scala/kafka/network/SocketServer.scala
@@ -1425,8 +1425,10 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
     def shouldUpdateQuota(metric: KafkaMetric, quotaLimit: Int) = {
       quotaLimit != metric.config.quota.bound
     }
+
     ip match {
       case Some(address) =>
+        // synchronize on counts to ensure reading an IP connection rate quota and creating a quota config is atomic
         counts.synchronized {
           maxConnectionRate match {
             case Some(rate) =>
@@ -1439,6 +1441,7 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
         }
         updateConnectionRateQuota(connectionRateForIp(address), IpQuotaEntity(address))
       case None =>
+        // synchronize on counts to ensure reading an IP connection rate quota and creating a quota config is atomic
         counts.synchronized {
           defaultConnectionRatePerIp = maxConnectionRate.getOrElse(DynamicConfig.Ip.DefaultConnectionCreationRate)
         }
@@ -1605,7 +1608,9 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
   /**
    * Calculates the delay needed to bring the observed connection creation rate to the IP limit.
    * If the connection would cause an IP quota violation, un-record the connection for both IP,
-   * listener, and broker connection rate and throw a ConnectionThrottledException.
+   * listener, and broker connection rate and throw a ConnectionThrottledException. Calls to
+   * this function must be performed with the counts lock to ensure that reading the IP
+   * connection rate quota and creating the sensor's metric config is atomic.
    *
    * @param listenerName listener to unrecord connection if throttled
    * @param address ip address to record connection
--- a/core/src/main/scala/kafka/network/SocketServer.scala
+++ b/core/src/main/scala/kafka/network/SocketServer.scala
@@ -25,6 +25,7 @@ import java.util
 import java.util.Optional
 import java.util.concurrent._
 import java.util.concurrent.atomic._
+import java.util.concurrent.locks.ReentrantReadWriteLock
 
 import kafka.cluster.{BrokerEndPoint, EndPoint}
 import kafka.metrics.KafkaMetricsGroup
@@ -32,7 +33,7 @@ import kafka.network.Processor._
 import kafka.network.RequestChannel.{CloseConnectionResponse, EndThrottlingResponse, NoOpResponse, SendResponse, StartThrottlingResponse}
 import kafka.network.SocketServer._
 import kafka.security.CredentialProvider
-import kafka.server.{BrokerReconfigurable, KafkaConfig}
+import kafka.server.{BrokerReconfigurable, DynamicConfig, KafkaConfig, SensorAccess}
 import kafka.utils.Implicits._
 import kafka.utils._
 import org.apache.kafka.common.config.ConfigException
@@ -101,7 +102,7 @@ class SocketServer(val config: KafkaConfig,
     new RequestChannel(20, ControlPlaneMetricPrefix, time, allowDisabledApis))
 
   private var nextProcessorId = 0
-  private var connectionQuotas: ConnectionQuotas = _
+  val connectionQuotas = new ConnectionQuotas(config, time, metrics)
   private var startedProcessingRequests = false
   private var stoppedProcessingRequests = false
 
@@ -119,7 +120,6 @@ class SocketServer(val config: KafkaConfig,
    */
   def startup(startProcessingRequests: Boolean = true): Unit = {
     this.synchronized {
-      connectionQuotas = new ConnectionQuotas(config, time, metrics)
       createControlPlaneAcceptorAndProcessor(config.controlPlaneListener)
       createDataPlaneAcceptorsAndProcessors(config.numNetworkThreads, config.dataPlaneListeners)
       if (startProcessingRequests) {
@@ -137,7 +137,7 @@ class SocketServer(val config: KafkaConfig,
     })
     newGauge(s"${ControlPlaneMetricPrefix}NetworkProcessorAvgIdlePercent", () => SocketServer.this.synchronized {
       val ioWaitRatioMetricName = controlPlaneProcessorOpt.map { p =>
-        metrics.metricName("io-wait-ratio", "socket-server-metrics", p.metricTags)
+        metrics.metricName("io-wait-ratio", MetricsGroup, p.metricTags)
       }
       ioWaitRatioMetricName.map { metricName =>
         Option(metrics.metric(metricName)).fold(0.0)(m => Math.min(m.metricValue.asInstanceOf[Double], 1.0))
@@ -147,7 +147,7 @@ class SocketServer(val config: KafkaConfig,
     newGauge("MemoryPoolUsed", () => memoryPool.size() - memoryPool.availableMemory)
     newGauge(s"${DataPlaneMetricPrefix}ExpiredConnectionsKilledCount", () => SocketServer.this.synchronized {
       val expiredConnectionsKilledCountMetricNames = dataPlaneProcessors.values.asScala.iterator.map { p =>
-        metrics.metricName("expired-connections-killed-count", "socket-server-metrics", p.metricTags)
+        metrics.metricName("expired-connections-killed-count", MetricsGroup, p.metricTags)
       }
       expiredConnectionsKilledCountMetricNames.map { metricName =>
         Option(metrics.metric(metricName)).fold(0.0)(m => m.metricValue.asInstanceOf[Double])
@@ -155,7 +155,7 @@ class SocketServer(val config: KafkaConfig,
     })
     newGauge(s"${ControlPlaneMetricPrefix}ExpiredConnectionsKilledCount", () => SocketServer.this.synchronized {
       val expiredConnectionsKilledCountMetricNames = controlPlaneProcessorOpt.map { p =>
-        metrics.metricName("expired-connections-killed-count", "socket-server-metrics", p.metricTags)
+        metrics.metricName("expired-connections-killed-count", MetricsGroup, p.metricTags)
       }
       expiredConnectionsKilledCountMetricNames.map { metricName =>
         Option(metrics.metric(metricName)).fold(0.0)(m => m.metricValue.asInstanceOf[Double])
@@ -277,7 +277,7 @@ class SocketServer(val config: KafkaConfig,
     val sendBufferSize = config.socketSendBufferBytes
     val recvBufferSize = config.socketReceiveBufferBytes
     val brokerId = config.brokerId
-    new Acceptor(endPoint, sendBufferSize, recvBufferSize, brokerId, connectionQuotas, metricPrefix)
+    new Acceptor(endPoint, sendBufferSize, recvBufferSize, brokerId, connectionQuotas, metricPrefix, time)
   }
 
   private def addDataPlaneProcessors(acceptor: Acceptor, endpoint: EndPoint, newProcessorsPerListener: Int): Unit = {
@@ -524,9 +524,13 @@ private[kafka] abstract class AbstractServerThread(connectionQuotas: ConnectionQ
     if (channel != null) {
       debug(s"Closing connection from ${channel.socket.getRemoteSocketAddress()}")
       connectionQuotas.dec(listenerName, channel.socket.getInetAddress)
-      CoreUtils.swallow(channel.socket().close(), this, Level.ERROR)
-      CoreUtils.swallow(channel.close(), this, Level.ERROR)
+      closeSocket(channel)
     }
+  }
+
+  protected def closeSocket(channel: SocketChannel): Unit = {
+    CoreUtils.swallow(channel.socket().close(), this, Level.ERROR)
+    CoreUtils.swallow(channel.close(), this, Level.ERROR)
   }
 }
 
@@ -538,7 +542,8 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
                               val recvBufferSize: Int,
                               brokerId: Int,
                               connectionQuotas: ConnectionQuotas,
-                              metricPrefix: String) extends AbstractServerThread(connectionQuotas) with KafkaMetricsGroup {
+                              metricPrefix: String,
+                              time: Time) extends AbstractServerThread(connectionQuotas) with KafkaMetricsGroup {
 
   private val nioSelector = NSelector.open()
   val serverChannel = openServerSocket(endPoint.host, endPoint.port)
@@ -546,6 +551,9 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
   private val processorsStarted = new AtomicBoolean
   private val blockedPercentMeter = newMeter(s"${metricPrefix}AcceptorBlockedPercent",
     "blocked time", TimeUnit.NANOSECONDS, Map(ListenerMetricTag -> endPoint.listenerName.value))
+  private var currentProcessorIndex = 0
+  private case class DelayedCloseSocket(socket: SocketChannel, endThrottleTimeMs: Long)
+  private val throttledSockets = new util.HashMap[InetAddress, mutable.Queue[DelayedCloseSocket]]()
 
   private[network] def addProcessors(newProcessors: Buffer[Processor], processorThreadPrefix: String): Unit = synchronized {
     processors ++= newProcessors
@@ -600,43 +608,10 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
     serverChannel.register(nioSelector, SelectionKey.OP_ACCEPT)
     startupComplete()
     try {
-      var currentProcessorIndex = 0
       while (isRunning) {
         try {
-          val ready = nioSelector.select(500)
-          if (ready > 0) {
-            val keys = nioSelector.selectedKeys()
-            val iter = keys.iterator()
-            while (iter.hasNext && isRunning) {
-              try {
-                val key = iter.next
-                iter.remove()
-
-                if (key.isAcceptable) {
-                  accept(key).foreach { socketChannel =>
-                    // Assign the channel to the next processor (using round-robin) to which the
-                    // channel can be added without blocking. If newConnections queue is full on
-                    // all processors, block until the last one is able to accept a connection.
-                    var retriesLeft = synchronized(processors.length)
-                    var processor: Processor = null
-                    do {
-                      retriesLeft -= 1
-                      processor = synchronized {
-                        // adjust the index (if necessary) and retrieve the processor atomically for
-                        // correct behaviour in case the number of processors is reduced dynamically
-                        currentProcessorIndex = currentProcessorIndex % processors.length
-                        processors(currentProcessorIndex)
-                      }
-                      currentProcessorIndex += 1
-                    } while (!assignNewConnection(socketChannel, processor, retriesLeft == 0))
-                  }
-                } else
-                  throw new IllegalStateException("Unrecognized key state for acceptor thread.")
-              } catch {
-                case e: Throwable => error("Error while accepting connection", e)
-              }
-            }
-          }
+          acceptNewConnections()
+          closeThrottledConnections()
         }
         catch {
           // We catch all the throwables to prevent the acceptor thread from exiting on exceptions due
@@ -647,9 +622,12 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
         }
       }
     } finally {
-      debug("Closing server socket and selector.")
+      debug("Closing server socket, selector, and any throttled sockets.")
       CoreUtils.swallow(serverChannel.close(), this, Level.ERROR)
       CoreUtils.swallow(nioSelector.close(), this, Level.ERROR)
+      throttledSockets.asScala.values.flatten.foreach { throttledSocket =>
+        closeSocket(throttledSocket.socket)
+      }
       shutdownComplete()
     }
   }
@@ -679,6 +657,46 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
   }
 
   /**
+   * Listen for new connections and assign accepted connections to processors using round-robin
+   */
+  private def acceptNewConnections(): Unit = {
+    val ready = nioSelector.select(500)
+    if (ready > 0) {
+      val keys = nioSelector.selectedKeys()
+      val iter = keys.iterator()
+      while (iter.hasNext && isRunning) {
+        try {
+          val key = iter.next
+          iter.remove()
+
+          if (key.isAcceptable) {
+            accept(key).foreach { socketChannel =>
+              // Assign the channel to the next processor (using round-robin) to which the
+              // channel can be added without blocking. If newConnections queue is full on
+              // all processors, block until the last one is able to accept a connection.
+              var retriesLeft = synchronized(processors.length)
+              var processor: Processor = null
+              do {
+                retriesLeft -= 1
+                processor = synchronized {
+                  // adjust the index (if necessary) and retrieve the processor atomically for
+                  // correct behaviour in case the number of processors is reduced dynamically
+                  currentProcessorIndex = currentProcessorIndex % processors.length
+                  processors(currentProcessorIndex)
+                }
+                currentProcessorIndex += 1
+              } while (!assignNewConnection(socketChannel, processor, retriesLeft == 0))
+            }
+          } else
+            throw new IllegalStateException("Unrecognized key state for acceptor thread.")
+        } catch {
+          case e: Throwable => error("Error while accepting connection", e)
+        }
+      }
+    }
+  }
+
+  /**
    * Accept a new connection
    */
   private def accept(key: SelectionKey): Option[SocketChannel] = {
@@ -697,6 +715,32 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
         info(s"Rejected connection from ${e.ip}, address already has the configured maximum of ${e.count} connections.")
         close(endPoint.listenerName, socketChannel)
         None
+      case e: ConnectionThrottledException =>
+        val ip = socketChannel.socket.getInetAddress
+        debug(s"Delaying closing of connection from $ip for ${e.throttleTimeMs} ms")
+        val delayQueue = throttledSockets.computeIfAbsent(ip, _ => new mutable.Queue[DelayedCloseSocket])
+        val endThrottleTimeMs = e.startThrottleTimeMs + e.throttleTimeMs
+        delayQueue += DelayedCloseSocket(socketChannel, endThrottleTimeMs)
+        None
+    }
+  }
+
+  /**
+   * Close sockets for any connections that have been throttled
+   */
+  private def closeThrottledConnections(): Unit = {
+    val timeMs = time.milliseconds
+    val iter = throttledSockets.values.iterator
+    while (iter.hasNext) {
+      val throttledSockets = iter.next
+      throttledSockets.dequeueWhile { socket =>
+        socket.endThrottleTimeMs < timeMs
+      }.foreach { closingSocket =>
+        debug(s"Closing socket from ip ${closingSocket.socket.getRemoteAddress}")
+        closeSocket(closingSocket.socket)
+      }
+      if (throttledSockets.isEmpty)
+        iter.remove()
     }
   }
 
@@ -789,7 +833,7 @@ private[kafka] class Processor(val id: Int,
   )
 
   val expiredConnectionsKilledCount = new CumulativeSum()
-  private val expiredConnectionsKilledCountMetricName = metrics.metricName("expired-connections-killed-count", "socket-server-metrics", metricTags)
+  private val expiredConnectionsKilledCountMetricName = metrics.metricName("expired-connections-killed-count", MetricsGroup, metricTags)
   metrics.addMetric(expiredConnectionsKilledCountMetricName, expiredConnectionsKilledCount)
 
   private val selector = createSelector(
@@ -1273,6 +1317,22 @@ private[kafka] class Processor(val id: Int,
   }
 }
 
+sealed trait ConnectionQuotaEntity {
+  def entityName: String
+}
+
+private case class ListenerQuotaEntity(listenerName: String) extends ConnectionQuotaEntity {
+  override def entityName: String = listenerName
+}
+
+private case object BrokerQuotaEntity extends ConnectionQuotaEntity {
+  override def entityName: String = "broker-wide"
+}
+
+private case class IpQuotaEntity(ip: InetAddress) extends ConnectionQuotaEntity {
+  override def entityName: String = ip.toString
+}
+
 class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extends Logging with AutoCloseable {
 
   @volatile private var defaultMaxConnectionsPerIp: Int = config.maxConnectionsPerIp
@@ -1285,14 +1345,27 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
   private val listenerCounts = mutable.Map[ListenerName, Int]()
   private[network] val maxConnectionsPerListener = mutable.Map[ListenerName, ListenerConnectionQuota]()
   @volatile private var totalCount = 0
-
+  @volatile private var defaultConnectionRatePerIp = DynamicConfig.Ip.DefaultConnectionCreationRate
+  private val inactiveSensorExpirationTimeSeconds = TimeUnit.HOURS.toSeconds(1);
+  private val connectionRatePerIp = new ConcurrentHashMap[InetAddress, Int]()
+  private val lock = new ReentrantReadWriteLock()
+  private val sensorAccessor = new SensorAccess(lock, metrics)
   // sensor that tracks broker-wide connection creation rate and limit (quota)
-  private val brokerConnectionRateSensor = createConnectionRateQuotaSensor(config.maxConnectionCreationRate)
+  private val brokerConnectionRateSensor = getOrCreateConnectionRateQuotaSensor(config.maxConnectionCreationRate, BrokerQuotaEntity)
   private val maxThrottleTimeMs = TimeUnit.SECONDS.toMillis(config.quotaWindowSizeSeconds.toLong)
+
 
   def inc(listenerName: ListenerName, address: InetAddress, acceptorBlockedPercentMeter: com.yammer.metrics.core.Meter): Unit = {
     counts.synchronized {
-      waitForConnectionSlot(listenerName, acceptorBlockedPercentMeter)
+      val startThrottleTimeMs = time.milliseconds
+
+      val ipThrottleTimeMs = recordIpConnectionMaybeThrottle(address, startThrottleTimeMs)
+      if (ipThrottleTimeMs > 0) {
+        trace(s"Throttling $address for $ipThrottleTimeMs ms")
+        throw new ConnectionThrottledException(address, startThrottleTimeMs, ipThrottleTimeMs)
+      }
+
+      waitForConnectionSlot(listenerName, startThrottleTimeMs, acceptorBlockedPercentMeter)
 
       val count = counts.getOrElseUpdate(address, 0)
       counts.put(address, count + 1)
@@ -1324,7 +1397,56 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
   private[network] def updateBrokerMaxConnectionRate(maxConnectionRate: Int): Unit = {
     // if there is a connection waiting on the rate throttle delay, we will let it wait the original delay even if
     // the rate limit increases, because it is just one connection per listener and the code is simpler that way
-    updateConnectionRateQuota(maxConnectionRate)
+    updateConnectionRateQuota(maxConnectionRate, BrokerQuotaEntity)
+  }
+
+  /**
+   * Update the connection rate quota for a given IP and updates quota configs for updated IPs.
+   * If an IP is given, metric config will be updated only for the given IP, otherwise
+   * all metric configs will be checked and updated if required
+   *
+   * @param ip ip to update or default if None
+   * @param maxConnectionRate new connection rate, or resets entity to default if None
+   */
+  def updateIpConnectionRate(ip: Option[String], maxConnectionRate: Option[Int]): Unit = {
+    def isIpConnectionRateMetric(metricName: MetricName) = {
+      metricName.name == "connection-accept-rate" &&
+      metricName.group == MetricsGroup &&
+      metricName.tags.containsKey("ip")
+    }
+
+    def shouldUpdateQuota(metric: KafkaMetric, quotaLimit: Int) = {
+      quotaLimit != metric.config.quota.bound
+    }
+
+    ip match {
+      case Some(addr) =>
+        val address = InetAddress.getByName(addr)
+        if (maxConnectionRate.isDefined) {
+          info(s"Updating max connection rate override for $address to ${maxConnectionRate.get}")
+          connectionRatePerIp.put(address, maxConnectionRate.get)
+        } else {
+          info(s"Removing max connection rate override for $address")
+          connectionRatePerIp.remove(address)
+        }
+        updateConnectionRateQuota(connectionRateForIp(address), IpQuotaEntity(address))
+      case None =>
+        val newQuota = maxConnectionRate.getOrElse(Int.MaxValue)
+        info(s"Updating default max IP connection rate to $newQuota")
+        defaultConnectionRatePerIp = newQuota
+        val allMetrics = metrics.metrics
+        allMetrics.forEach { (metricName, metric) =>
+          if (isIpConnectionRateMetric(metricName) && shouldUpdateQuota(metric, newQuota)) {
+            info(s"Updating existing connection rate sensor for ${metricName.tags} to $newQuota")
+            metric.config(rateQuotaMetricConfig(newQuota))
+          }
+        }
+    }
+  }
+
+  // Visible for testing
+  def connectionRateForIp(ip: InetAddress): Int = {
+    connectionRatePerIp.getOrDefault(ip, defaultConnectionRatePerIp)
   }
 
   private[network] def addListener(config: KafkaConfig, listenerName: ListenerName): Unit = {
@@ -1382,9 +1504,9 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
   }
 
   private def waitForConnectionSlot(listenerName: ListenerName,
+                                    startThrottleTimeMs: Long,
                                     acceptorBlockedPercentMeter: com.yammer.metrics.core.Meter): Unit = {
     counts.synchronized {
-      val startThrottleTimeMs = time.milliseconds
       val throttleTimeMs = math.max(recordConnectionAndGetThrottleTimeMs(listenerName, startThrottleTimeMs), 0)
 
       if (throttleTimeMs > 0 || !connectionSlotAvailable(listenerName)) {
@@ -1454,6 +1576,29 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
   }
 
   /**
+   * Calculates the delay needed to bring the observed connection creation rate to the IP limit.
+   * If the connection would cause an IP quota violation, un-record the connection
+   *
+   * @param address
+   * @param timeMs
+   * @return
+   */
+  private def recordIpConnectionMaybeThrottle(address: InetAddress, timeMs: Long): Long = {
+    val connectionRateQuota = connectionRateForIp(address)
+    val quotaEnabled = connectionRateQuota != DynamicConfig.Ip.UnlimitedConnectionCreationRate
+    if (!quotaEnabled) {
+      return 0
+    }
+    val sensor = getOrCreateConnectionRateQuotaSensor(connectionRateForIp(address), IpQuotaEntity(address))
+    val throttleMs = recordAndGetThrottleTimeMs(sensor, timeMs)
+    if (throttleMs > 0) {
+      // unrecord the connection since we won't accept the connection
+      sensor.record(-1.0, timeMs, false)
+    }
+    throttleMs
+  }
+
+  /**
    * Records a new connection into a given connection acceptance rate sensor 'sensor' and returns throttle time
    * in milliseconds if quota got violated
    * @param sensor sensor to record connection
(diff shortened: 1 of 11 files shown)


core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1203,14 +1262,27 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+  private val brokerConnectionRateSensor = getOrCreateConnectionRateQuotaSensor(config.maxConnectionCreationRate, BrokerQuotaEntity)
   private val maxThrottleTimeMs = TimeUnit.SECONDS.toMillis(config.quotaWindowSizeSeconds.toLong)
 
+
   def inc(listenerName: ListenerName, address: InetAddress, acceptorBlockedPercentMeter: com.yammer.metrics.core.Meter): Unit = {
     counts.synchronized {
-      waitForConnectionSlot(listenerName, acceptorBlockedPercentMeter)
+      val startThrottleTimeMs = time.milliseconds
+
+      val ipThrottleTimeMs = recordIpConnectionMaybeThrottle(address, startThrottleTimeMs)
```
apovzner: It would be more efficient if we throttled IPs **after** we know that we can accept a connection based on broker-wide and per-listener limits, since reaching broker/listener limits block the acceptor thread while throttling IPs needs more processing. Otherwise, if you reach both broker and per IP limit, the broker will continue accepting and delaying connections where it is justified to block an acceptor thread based on reaching a broker rate limit. Basically, call `waitForConnectionSlot` first. Similar how we check per IP limit on number of connections after we know that we can accept a new connection based on broker/listener limits.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1242,7 +1314,56 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+        if (maxConnectionRate.isDefined) {
+          info(s"Updating max connection rate override for $address to ${maxConnectionRate.get}")
+          connectionRatePerIp.put(address, maxConnectionRate.get)
+        } else {
+          info(s"Removing max connection rate override for $address")
+          connectionRatePerIp.remove(address)
+        }
+        updateConnectionRateQuota(connectionRateForIp(address), IpQuotaEntity(address))
+      case None =>
+        val newQuota = maxConnectionRate.getOrElse(Int.MaxValue)
```
apovzner: You can use you new constant `DynamicConfig.Ip.UnlimitedConnectionCreationRate` instead of `Int.MaxValue` here.

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
```diff
@@ -409,6 +409,67 @@ class ConnectionQuotasTest {
     verifyConnectionCountOnEveryListener(connectionQuotas, connectionsPerListener)
   }
 
+  @Test
```
apovzner, line 424: It would be useful to add a test where we have both per-listener and per IP limit, and verify that it throttles based on which limit is reached first. Something like: 2 IPs, each per IP limit < per-listener limit, but sum of per IP limits > listener limit. So, if you reach limit on one IP, the broker would not throttle the second IP until it reaches per listener limit. Does not have to be exactly this, just need to verify how per IP throttling interacts with per listener throttling.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: @apovzner 
my reasoning for this is the following:
consider the case where we accept a connection at the broker/listener level, but reject it on IP level.
we would have already recorded the broker connection, so we'd be allocating rate to a rejected connection.

I suppose this we can work around this in a similar manner to `recordIpConnectionMaybeThrottle` by unrecording the listener/broker connection if the IP gets rejected.

apovzner: I see. Yes, I think unrecording is more efficient than keeping more delayed connections than needed. Basically, when you unrecord from per-IP metric, you can also unrecord from broker and listener metric as well.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: actually in this case, it should reset to `DynamicConfig.Ip.DefaultConnectionCreationRate` right?

splett2:
@apovzner 
Updated PR to address comments.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1203,14 +1261,28 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
   private val listenerCounts = mutable.Map[ListenerName, Int]()
   private[network] val maxConnectionsPerListener = mutable.Map[ListenerName, ListenerConnectionQuota]()
   @volatile private var totalCount = 0
-
+  @volatile private var defaultConnectionRatePerIp = DynamicConfig.Ip.DefaultConnectionCreationRate
+  private val inactiveSensorExpirationTimeSeconds = TimeUnit.HOURS.toSeconds(1);
```
dajac: Could we define this constant in a companion object?

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1394,28 +1554,38 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
-    val sensor = metrics.sensor(sensorName, rateQuotaMetricConfig(quotaLimit))
-    sensor.add(connectionRateMetricName(listenerOpt), new Rate, null)
-    info(s"Created $sensorName sensor, quotaLimit=$quotaLimit")
-    sensor
+  private def getOrCreateConnectionRateQuotaSensor(quotaLimit: Int, connectionQuotaEntity: ConnectionQuotaEntity): Sensor = {
+    val (sensorName, sensorExpiration) = connectionQuotaEntity match {
+      case BrokerQuotaEntity => ("ConnectionAcceptRate", Long.MaxValue)
+      case listenerEntity: ListenerQuotaEntity => (s"ConnectionAcceptRate-${listenerEntity.entityName}", Long.MaxValue)
+      case ipEntity: IpQuotaEntity => (s"ConnectionAcceptRate-${ipEntity.entityName}", inactiveSensorExpirationTimeSeconds)
+    }
```
dajac: Couldn't we rely on the `ConnectionQuotaEntity` trait to get these informations? That would avoid having to pattern match on the entity to get them.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1394,28 +1554,38 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
 
-  private def connectionRateMetricName(listenerOpt: Option[String]): MetricName = {
-    val tags = listenerOpt.map(listener => Map("listener" -> listener)).getOrElse(Map())
-    val namePrefix = listenerOpt.map(_ => "").getOrElse("broker-")
+  private def connectionRateMetricName(connectionQuotaEntity: ConnectionQuotaEntity): MetricName = {
+    val (namePrefix, tags) = connectionQuotaEntity match {
+      case BrokerQuotaEntity => ("broker-", Map.empty[String, String])
+      case listener: ListenerQuotaEntity => ("", Map("listener" -> listener.entityName))
+      case ip: IpQuotaEntity => ("", Map("ip" -> ip.entityName))
+    }
```
dajac: Same comment as before. Couldn't we rely on the `ConnectionQuotaEntity` trait here as well?

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1371,6 +1492,45 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+   * @param timeMs
+   * @return
+   */
+  private def recordIpConnectionMaybeThrottle(address: InetAddress, timeMs: Long): Long = {
+    val connectionRateQuota = connectionRateForIp(address)
+    val quotaEnabled = connectionRateQuota != DynamicConfig.Ip.UnlimitedConnectionCreationRate
+    if (!quotaEnabled) {
+      return 0
+    }
+    val sensor = getOrCreateConnectionRateQuotaSensor(connectionRateForIp(address), IpQuotaEntity(address))
```
dajac: nit: Can't we reuse `connectionRateQuota` directly?

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1371,6 +1492,45 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+    val throttleMs = recordAndGetThrottleTimeMs(sensor, timeMs)
+    if (throttleMs > 0) {
+      // unrecord the connection since we won't accept the connection
+      sensor.record(-1.0, timeMs, false)
+    }
```
dajac: Instead of recording, checking the quota, and unrecording if the quota is violated, we may be able to check the quota and record only if the quota is not violated yet. Your implementation suggests that we want to enforce a strict quota here. We did so for the controller mutations with the following implementation:

```
try {
      quotaSensor synchronized {
        quotaSensor.checkQuotas(timeMs)
        quotaSensor.record(permits, timeMs, false)
      }
    } catch {
      case e: QuotaViolationException =>
         ...
    }
```

It seems that this should work here as well. Let me know what you think about this.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -697,6 +714,31 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
         info(s"Rejected connection from ${e.ip}, address already has the configured maximum of ${e.count} connections.")
         close(endPoint.listenerName, socketChannel)
         None
+      case e: ConnectionThrottledException =>
+        val ip = socketChannel.socket.getInetAddress
+        debug(s"Delaying closing of connection from $ip for ${e.throttleTimeMs} ms")
+        val delayQueue = throttledSockets.computeIfAbsent(ip, _ => new mutable.Queue[DelayedCloseSocket])
```
dajac: - Do we really need to store delayed sockets per IP? It seems that we just close all the sockets in `throttledSockets` later on so it is not really necessary. Or have I missed something?
- Similarly, we use a `DelayQueue` for the other quotas. Is there any reason why we don't use the same data structure here?

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -600,43 +607,10 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
...
-                  }
-                } else
-                  throw new IllegalStateException("Unrecognized key state for acceptor thread.")
-              } catch {
-                case e: Throwable => error("Error while accepting connection", e)
-              }
-            }
-          }
+          acceptNewConnections()
+          closeThrottledConnections()
```
dajac, line 617: Out of curiosity, have we considered using a "reaper" running in a separate thread to drain the queue and close throttled connections?

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -697,6 +714,31 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
...
+    }
+  }
+
+  /**
+   * Close sockets for any connections that have been throttled
+   */
+  private def closeThrottledConnections(): Unit = {
+    val timeMs = time.milliseconds
+    val iter = throttledSockets.values.iterator
+    while (iter.hasNext) {
```
dajac: If we were using a shared `DelayQueue` for all IPs, iterating over all the IPs wouldn't be necessary when there are no throttled connections ready to be closed yet.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1242,7 +1314,56 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+   * Update the connection rate quota for a given IP and updates quota configs for updated IPs.
+   * If an IP is given, metric config will be updated only for the given IP, otherwise
+   * all metric configs will be checked and updated if required
+   *
+   * @param ip ip to update or default if None
+   * @param maxConnectionRate new connection rate, or resets entity to default if None
+   */
+  def updateIpConnectionRate(ip: Option[String], maxConnectionRate: Option[Int]): Unit = {
+    def isIpConnectionRateMetric(metricName: MetricName) = {
+      metricName.name == "connection-accept-rate" &&
```
dajac: Could we use a constant for the name and use it everywhere? Same for the tag.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1242,7 +1314,56 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+    }
+
+    def shouldUpdateQuota(metric: KafkaMetric, quotaLimit: Int) = {
+      quotaLimit != metric.config.quota.bound
+    }
+
+    ip match {
+      case Some(addr) =>
+        val address = InetAddress.getByName(addr)
+        if (maxConnectionRate.isDefined) {
```
dajac: nit: We could use pattern matching here to treat both cases.

core/src/test/scala/integration/kafka/network/DynamicConnectionQuotaTest.scala
```diff
@@ -240,6 +256,16 @@ class DynamicConnectionQuotaTest extends BaseRequestTest {
       s"Admin client connection not closed (initial = $initialConnectionCount, current = $connectionCount)")
   }
 
+  private def updateIpConnectionRate(ip: Option[String], updatedRate: Int): Unit = {
+    adminZkClient.changeIpConfig(ip.getOrElse(ConfigEntityName.Default),
```
dajac, line 260: nit: Could we use directly the `Admin` client here? We don't use the ZK admin client in the suite at all so it would be better if we don't use it. I am not sure if this depends on the pending implementation that you mentioned.

dajac (commented):
@splett2 Thanks for the PR. It looks good overall. I have left some comments/questions.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: I didn't do this because I saw that `checkQuotas` checks whether we are at the upper bound, not whether recording a new connection would throw an exception, which I suppose is why you are calling `record` with `checkQuotas = false`.

I think it makes sense to do here. The actual rate of permitted connections ends up being 1 higher than otherwise, but I think that's fine.

core/src/test/scala/integration/kafka/network/DynamicConnectionQuotaTest.scala
splett2: yeah, that depends on pending implementation. I planned on using `adminZkClient` in the interim and then replacing it with `Admin` in the PR adding `KafkaApis` support.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: I went with this data structure because throttle end time should be monotonic for a specific IP but not necessarily for all IPs. By maintaining a queue per IP, we can get constant time adds/removes for a throttled IP.

The downside of this is that in the case where there are no connections to unthrottle, `closeThrottledConnections` runs in `O(n)` for number of throttled IPs.

On the other hand `DelayQueue` uses a heap internally and has `O(log(n))` for each add/remove, and constant time for checking whether there are IPs to unthrottle. In scenarios where many connections will get unthrottled at once, the `DelayQueue` will perform significantly worse, since each `DelayQueue.dequeue` call will be `O(log(n))`.

I admit this could be a premature optimization, and a delay queue could work better in practice. Let me know what you think.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: I did consider this. I chose to do the work of closing connections in the acceptor thread to mirror the `Processor` thread which similarly has a `openSomeConnections()`, `closeSomeConnections()` loop.
Ditto for `Selector.poll()`

Additionally, as a future extension, we may want to re-try accepting a connection after its throttle time expires, and adding a new reaper thread would add additional complexity for that logic.

splett2 (commented):
Thanks for the reviews @dajac 
I responded to some of the questions, and I will try to address feedback.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: actually, i thought a bit more about it and in the case where an IP is supposed to be completely throttled (rate limit == 0), we would be accepting connections at a rate of 1 which would be very weird/unacceptable behavior. I left it at is. Let me know if that makes sense or if I have some holes in my logic.

splett2:
@dajac 
I updated the PR to address some of your comments. Let me know what you think, thanks!

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -526,10 +527,14 @@ private[kafka] abstract class AbstractServerThread(connectionQuotas: ConnectionQ
     if (channel != null) {
       debug(s"Closing connection from ${channel.socket.getRemoteSocketAddress()}")
       connectionQuotas.dec(listenerName, channel.socket.getInetAddress)
-      CoreUtils.swallow(channel.socket().close(), this, Level.ERROR)
-      CoreUtils.swallow(channel.close(), this, Level.ERROR)
+      closeSocket(channel)
     }
   }
+
+  protected def closeSocket(channel: SocketChannel): Unit = {
```
dajac, line 531: nit: Could we make this one private?

core/src/main/scala/kafka/network/SocketServer.scala
dajac: Thanks for the explanation.

I do agree that in the worst case scenario where we would have all the connections ready to be unthrottled, the heap based solution would perform worse. On the other hand, if there are no connections ready to be unthrottled, checking the heap is `O(1)` whereas we would be `O(N)` where `N` is the number of throttled connections.

On average, I do believe that the number of connections ready to be unthrottled at every run of the acceptor loop will be smaller (or even zero) than the total number of throttled connections. Cleaning `M` throttling connections with the heap would be `O(M log(N))` which is better than `O(N)` if `M < N / log(N)`.

I lean towards using a delay queue at the moment.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1246,7 +1337,57 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+      case None =>
+        val newQuota = maxConnectionRate.getOrElse(DynamicConfig.Ip.DefaultConnectionCreationRate)
+        info(s"Updating default max IP connection rate to $newQuota")
+        defaultConnectionRatePerIp = newQuota
+        val allMetrics = metrics.metrics
+        allMetrics.forEach { (metricName, metric) =>
+          if (isIpConnectionRateMetric(metricName) && shouldUpdateQuota(metric, newQuota)) {
+            info(s"Updating existing connection rate sensor for ${metricName.tags} to $newQuota")
+            metric.config(rateQuotaMetricConfig(newQuota))
+          }
```
dajac: I wonder if using `newQuota` is correct here. My understanding is that `ip == None` means that we update the default quota which is used if there is not per ip quota defined. So, we should also check if there is a per ip quota defined before overriding it with the new default, isn't it?

I would be great if we could add more unit tests to cover this logic with multiple IPs and/or default.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1375,6 +1516,45 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+      brokerConnectionRateSensor.record(-1.0, timeMs, false)
+    }
+    maxConnectionsPerListener
+      .get(listenerName)
+      .foreach(_.connectionRateSensor.record(-1.0, timeMs, false))
+  }
+
+  /**
+   * Calculates the delay needed to bring the observed connection creation rate to the IP limit.
+   * If the connection would cause an IP quota violation, un-record the connection
```
dajac: nit: Add `.` at the end of the sentence.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1375,6 +1516,45 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+   *
+   * @param address
+   * @param timeMs
+   * @return
```
dajac: nit: To be completed.

core/src/main/scala/kafka/zk/AdminZkClient.scala
```diff
@@ -384,6 +386,28 @@ class AdminZkClient(zkClient: KafkaZkClient) extends Logging {
     changeEntityConfig(ConfigType.User, sanitizedEntityName, configs)
   }
 
+  /**
+   * validates the IP configs
+   * @param ip
+   * @param configs
```
dajac: nit: To be completed.

core/src/main/scala/kafka/zk/AdminZkClient.scala
```diff
@@ -384,6 +386,28 @@ class AdminZkClient(zkClient: KafkaZkClient) extends Logging {
...
+    if (ip != ConfigEntityName.Default && !Utils.validHostPattern(ip))
+      throw new AdminOperationException(s"IP $ip is not a valid address.")
+    DynamicConfig.Ip.validate(configs)
+  }
+
+  /**
+   * Update the config for an IP. These overrides will be persisted between sessions, and will override any default
+   * IP properties.
+   * @param ip
+   * @param configs
```
dajac: nit: To be completed.

core/src/test/scala/integration/kafka/network/DynamicConnectionQuotaTest.scala
```diff
@@ -240,6 +256,16 @@ class DynamicConnectionQuotaTest extends BaseRequestTest {
...
+      CoreUtils.propsWith(DynamicConfig.Ip.IpConnectionRateOverrideProp, updatedRate.toString))
+    // use a random throwaway address if ip isn't specified to get the default value
+    TestUtils.waitUntilTrue(() => servers.head.socketServer.connectionQuotas.
+      connectionRateForIp(InetAddress.getByName(ip.getOrElse("255.255.3.4"))) == updatedRate,
```
dajac, line 264: What is `255.255.3.4`?

core/src/test/scala/integration/kafka/network/DynamicConnectionQuotaTest.scala
dajac: Ack, thanks.

core/src/test/scala/integration/kafka/network/DynamicConnectionQuotaTest.scala
```diff
@@ -222,14 +222,30 @@ class DynamicConnectionQuotaTest extends BaseRequestTest {
...
     val externalFuture = executor.submit((() =>
-      verifyConnectionRate(3, listenerConnRateLimit, "EXTERNAL")): Runnable)
+      verifyConnectionRate(3, listenerConnRateLimit, "EXTERNAL", ignoreIOExceptions = false)): Runnable)
+
     plaintextFuture.get(40, TimeUnit.SECONDS)
     externalFuture.get(40, TimeUnit.SECONDS)
     waitForConnectionCount(initialConnectionCount)
   }
 
+  @Test
```
dajac, line 234: It would be great if we could add few more tests to check the dynamic quotas. For instance, we could check the in-memory quota as well as the quota in the metrics for a given IP when there is nothing, when there is a default set, when there is a quota for the IP.

We could also add a test to verify that updating the default quota does not change the per-ip quota if it is already set.

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
```diff
@@ -409,6 +410,96 @@ class ConnectionQuotasTest {
...
+        0, true)): Callable[Boolean])
+    )
+
+    val ipsThrottledResults = futures.map(_.get(3, TimeUnit.SECONDS))
+    val throttledIps = ipsThrottledResults.filter(identity)
+    // at most one IP should get IP throttled before the acceptor blocks on listener quota
+    assertTrue("Expected BlockedPercentMeter metric for EXTERNAL listener to be recorded", blockedPercentMeters("EXTERNAL").count() > 0)
+    assertTrue("Expect at most one IP to get throttled", throttledIps.size < 2)
+  }
+
```
dajac, line 511: It would be good to add few basic unit tests to verify the basics. Stuff like verifying that the quota is updated in-memory, verifying that the metrics are updated correctly (got caught by this recently), etc.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1246,7 +1337,57 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
     // if there is a connection waiting on the rate throttle delay, we will let it wait the original delay even if
     // the rate limit increases, because it is just one connection per listener and the code is simpler that way
-    updateConnectionRateQuota(maxConnectionRate)
+    updateConnectionRateQuota(maxConnectionRate, BrokerQuotaEntity)
+  }
+
+  /**
+   * Update the connection rate quota for a given IP and updates quota configs for updated IPs.
+   * If an IP is given, metric config will be updated only for the given IP, otherwise
+   * all metric configs will be checked and updated if required
```
dajac: nit: "." at the end.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -680,6 +659,46 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
     serverChannel
   }
 
+  /**
+   * Listen for new connections and assign accepted connections to processors using round-robin
```
dajac: nit: `.` at the end of the sentence.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -699,6 +718,31 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
...
+        debug(s"Delaying closing of connection from $ip for ${e.throttleTimeMs} ms")
+        val delayQueue = throttledSockets.computeIfAbsent(ip, _ => new mutable.Queue[DelayedCloseSocket])
+        val endThrottleTimeMs = e.startThrottleTimeMs + e.throttleTimeMs
+        delayQueue += DelayedCloseSocket(socketChannel, endThrottleTimeMs)
+        None
+    }
+  }
+
+  /**
+   * Close sockets for any connections that have been throttled
```
dajac: nit: `.` at the end.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1196,6 +1240,41 @@ private[kafka] class Processor(val id: Int,
   }
 }
 
+sealed trait ConnectionQuotaEntity {
```
dajac, line 1318: nit: Could we add some doc to explain this trait?

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1398,33 +1578,33 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
-   * Updates quota configuration for a given listener or broker-wide (if 'listenerOpt' is None)
+   * Updates quota configuration for a given connection quota entity
    */
-  private def updateConnectionRateQuota(quotaLimit: Int, listenerOpt: Option[String] = None): Unit = {
-    val metric = metrics.metric(connectionRateMetricName(listenerOpt))
-    metric.config(rateQuotaMetricConfig(quotaLimit))
-    info(s"Updated ${listenerOpt.getOrElse("broker-wide")} max connection creation rate to $quotaLimit")
+  private def updateConnectionRateQuota(quotaLimit: Int, connectionQuotaEntity: ConnectionQuotaEntity): Unit = {
+    val metricOpt = Option(metrics.metric(connectionRateMetricName(connectionQuotaEntity)))
+    metricOpt.foreach { metric =>
```
dajac: nit: `metricOpt` is not really necessary. We could directly put `foreach` on the line above.

core/src/test/scala/unit/kafka/network/SocketServerTest.scala
```diff
@@ -880,6 +872,72 @@ class SocketServerTest {
     }
   }
 
+  @Test
+  def testConnectionRatePerIp(): Unit = {
+    val overrideProps = TestUtils.createBrokerConfig(0, TestUtils.MockZkConnect, port = 0)
+    overrideProps.remove("max.connections.per.ip")
```
dajac: nit: Could we use `MaxConnectionsPerIpProp` instead of hardcoding the string?

core/src/main/scala/kafka/network/SocketServer.scala
dajac: That makes sense.

dajac (commented):
@splett2 Thanks for the update. I have made another pass over it and I have left some more comments. Overall, the core logic looks pretty good besides few small things. I will need to do another pass on the tests.

While reviewing, I have realized that we are not consistent between how max connection quota and connection rate are configured. The former is driven by `max.connections.per.ip` and `max.connections.per.ip.overrides` and the latter via a dedicated `IP` entity in ZK. It is a bit annoying and perhaps confusing that we don't use the same mechanism to configure these two related quotas. I like the new IP entity. We may want to update the other one in the future in order to remain consistent. @apovzner What do you think about this?

core/src/main/scala/kafka/network/SocketServer.scala
apovzner: I agree, that it is better to optimize for checking connections vs. overhead of adding/removing to the queue, because `closeThrottledConnections` runs pretty often (on a loop), which I think also means that finding that there are no connections yet to unthrottle would also be common. Or very few connections to unthrottle. So, after reading all your evaluations above, I am also leaning towards using a delay queue here.

Not sure if the question # 1 about why we need to delay closing a connection got answered, so answering just in case. Since we want to throttle accepting connections from an IP, closing a connection due to reaching IP quota right away would not help with throttling accepting connections since that IP is going to reconnect.

core/src/main/scala/kafka/network/SocketServer.scala
apovzner: this method is also called from the derived class.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1207,14 +1286,26 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
   private val maxThrottleTimeMs = TimeUnit.SECONDS.toMillis(config.quotaWindowSizeSeconds.toLong)
 
   def inc(listenerName: ListenerName, address: InetAddress, acceptorBlockedPercentMeter: com.yammer.metrics.core.Meter): Unit = {
     counts.synchronized {
-      waitForConnectionSlot(listenerName, acceptorBlockedPercentMeter)
+      val startThrottleTimeMs = time.milliseconds
+
+      waitForConnectionSlot(listenerName, startThrottleTimeMs, acceptorBlockedPercentMeter)
+
+      val ipThrottleTimeMs = recordIpConnectionMaybeThrottle(address, startThrottleTimeMs)
```
apovzner: There are some corner cases here where `startThrottleTimeMs` could be in the past if `waitForConnectionSlot()` waited for an active connection slot to become available (if we exceeded the limit for the number of active connections), or waited for broker-wide or listener-wide rate to get back to quota. In some cases, ipThrottleTimeMs would be zero if we checked against the current time here.

Since getting System.currentTimeMillis() is not that expensive (as it used to be), I think it would be better to revert to `waitForConnectionSlot` getting its own time (as before this PR), and then `recordIpConnectionMaybeThrottle` getting current time and also calling the code block below and throwing ConnectionThrottledException. And then adding a comment here that `recordIpConnectionMaybeThrottle` would throw an exception if per-IP quota is exceeded. What do you think?

core/src/main/scala/kafka/network/SocketServer.scala
apovzner: @splett2 Maybe I was not looking at the right place in this PR, but does this PR handles the case if someone sets non-unlimited per IP default quota? Basically, /config/ips/<default> znode. 

Because if default is set and it is not unlimited, removing quota for an IP should fall back to configured IP default, and if IP default is not set, then fall back to `DynamicConfig.Ip.DefaultConnectionCreationRate`. Which I think means that we need to have "<default>" ip in the `connectionRatePerIp` if default is not unlimited, and use it when creating per-IP sensor with the right quota.

core/src/test/scala/integration/kafka/network/DynamicConnectionQuotaTest.scala
splett2: This is admittedly a little weird.
If `None` is given as the IP, we want to update the default connection rate. 
To verify that the default connection rate was updated, we need to call `connectionRateForIp` with some arbitrary IP address that hasn't been given a specific override. In this case, I used an arbitrary IP, `255.255.3.4` as mentioned in the comment.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: @dajac 
good catch, we shouldn't be using newQuota here. I agree, this should have been covered by testing, I'll work on adding some more unit tests for the metric config updating.

@apovzner 
It should handle that case, yeah.
If default is set to some value, when we remove quota for an ip, e.g. `updateConnectionRate(Some(ip), None)`, we remove the connection rate entry from the map and then call `getOrDefault(ip, defaultConnectionRatePerIp)`which should be whatever the non-unlimited per-IP default quota is.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: seems reasonable

core/src/main/scala/kafka/network/SocketServer.scala
splett2: Okay, I replaced the map/queue with a priority queue.
I did not use the java `DelayQueue` similar to the implementation in `ClientQuotaManager` because this throttling implementation does not use timeout-based polling or require a synchronized data structure, and there's significantly more boilerplate needed for a `DelayQueue`

splett2:
thanks for the comments, I updated the PR to incorporate feedback. Let me know if there's anything I missed

apovzner (approved):
LGTM. Thanks for the PR!

dajac:
@splett2 Could you rebase the PR? I'll review it on Monday.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1659,19 +1648,22 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
-    )
+    Option(metrics.getSensor(connectionQuotaEntity.sensorName)).getOrElse {
+      val sensor = metrics.sensor(
+        connectionQuotaEntity.sensorName,
+        rateQuotaMetricConfig(quotaLimit),
+        connectionQuotaEntity.sensorExpiration
+      )
+      sensor.add(connectionRateMetricName(connectionQuotaEntity), new Rate, null)
+      sensor
+    }
```
splett2: I removed use of `sensorAccess`, since all calls to `sensorAccessor.getOrCreate` were performed with the `counts` lock, so the read-write lock is redundant.

dajac:
@splett2 Code does not compile:
```
[Error] /home/jenkins/jenkins-agent/workspace/Kafka_kafka-pr_PR-9386/core/src/main/scala/kafka/network/SocketServer.scala:1456: value getOrDefault is not a member of scala.collection.mutable.Map[java.net.InetAddress,Int]
```

splett2:
@dajac 
fixed compilation, please take a look when you get the chance.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -538,14 +542,20 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
...
+                              time: Time) extends AbstractServerThread(connectionQuotas) with KafkaMetricsGroup {
 
   private val nioSelector = NSelector.open()
   val serverChannel = openServerSocket(endPoint.host, endPoint.port)
   private val processors = new ArrayBuffer[Processor]()
   private val processorsStarted = new AtomicBoolean
   private val blockedPercentMeter = newMeter(s"${metricPrefix}AcceptorBlockedPercent",
     "blocked time", TimeUnit.NANOSECONDS, Map(ListenerMetricTag -> endPoint.listenerName.value))
+  private var currentProcessorIndex = 0
+  private[network] case class DelayedCloseSocket(socket: SocketChannel, endThrottleTimeMs: Long) extends Ordered[DelayedCloseSocket] {
```
dajac, line 557: nit: It is a bit weird to have this case class define in the middle of the block of variables. Would it make sense to move it to another place? May be after the block of variables?

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -538,14 +542,20 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
...
+    override def compare(that: DelayedCloseSocket): Int = endThrottleTimeMs compare that.endThrottleTimeMs
+  }
+  private[network] val throttledSockets = new mutable.PriorityQueue[DelayedCloseSocket]()
```
dajac: Out of curiosity, have we considered using a `DelayQueue` instead of `PriorityQueue`? I was wondering if that could reduce the logic on our side. I am not sure if it is worth doing it though.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -647,9 +624,10 @@ private[kafka] class Acceptor(val endPoint: EndPoint,
         }
       }
     } finally {
-      debug("Closing server socket and selector.")
+      debug("Closing server socket, selector, and any throttled sockets.")
       CoreUtils.swallow(serverChannel.close(), this, Level.ERROR)
       CoreUtils.swallow(nioSelector.close(), this, Level.ERROR)
+      throttledSockets.foreach(throttledSocket => closeSocket(throttledSocket.socket))
```
dajac, line 631: nit: We may want to clear `throttledSockets` for completeness.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1453,6 +1582,47 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+    }
+    maxConnectionsPerListener
+      .get(listenerName)
+      .foreach(_.connectionRateSensor.record(-1.0, timeMs, false))
+  }
+
+  /**
+   * Calculates the delay needed to bring the observed connection creation rate to the IP limit.
+   * If the connection would cause an IP quota violation, un-record the connection for both IP,
+   * and throw ConnectionThrottledException
```
dajac: nit: `un-record the connection for both IP, and throw ConnectionThrottledException`. It seems that something is missing. Perhaps, you meant `for both IP and the listener, and throw ...`? We can also add a dot at the end of the sentence.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1324,7 +1401,59 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+      metricName.tags.containsKey(IpMetricTag)
+    }
+
+    def shouldUpdateQuota(metric: KafkaMetric, quotaLimit: Int) = {
+      quotaLimit != metric.config.quota.bound
+    }
+    counts.synchronized {
+      ip match {
+        case Some(addr) =>
+          val address = InetAddress.getByName(addr)
```
dajac: What would happen if `addr` can't be resolved by name? For instance, say that we have `10.0.0.0.1` in ZK or a string which does not represent an IP. It may be worth verifying the IP in the `IpConfigHandler`. If the ip is not valid, we could handle it properly.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1324,7 +1401,59 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+    def isIpConnectionRateMetric(metricName: MetricName) = {
+      metricName.name == ConnectionRateMetricName &&
+      metricName.group == MetricsGroup &&
+      metricName.tags.containsKey(IpMetricTag)
+    }
+
+    def shouldUpdateQuota(metric: KafkaMetric, quotaLimit: Int) = {
+      quotaLimit != metric.config.quota.bound
+    }
+    counts.synchronized {
```
dajac: I do wonder if synchronizing on `counts` is a good idea here. When a single IP is updated, that is not an issue. However, we all the metrics are updated, that could take some time depending on the number of metrics. Holding the lock in this case means that we can't accept any new connections until the update is completed. Am I getting this right?

core/src/main/scala/kafka/server/ConfigHandler.scala
```diff
@@ -185,6 +186,19 @@ class UserConfigHandler(private val quotaManagers: QuotaManagers, val credential
   }
 }
 
+class IpConfigHandler(private val connectionQuotas: ConnectionQuotas) extends ConfigHandler with Logging {
+
+  def processConfigChanges(ip: String, config: Properties): Unit = {
+    val ipConnectionRateQuota = Option(config.getProperty(DynamicConfig.Ip.IpConnectionRateOverrideProp)).map(_.toInt)
+    val updatedIp =
+      if (ip != ConfigEntityName.Default)
+        Some(ip)
```
dajac: As mentioned in a previous comment, it may be worth validating the IP here.

core/src/test/scala/unit/kafka/server/DynamicConfigChangeTest.scala
```diff
@@ -193,6 +194,64 @@ class DynamicConfigChangeTest extends KafkaServerTestHarness {
...
+    ipOverrideProps.put(DynamicConfig.Ip.IpConnectionRateOverrideProp, "10")
+    val ipDefaultProps = new Properties()
+    ipDefaultProps.put(DynamicConfig.Ip.IpConnectionRateOverrideProp, "20")
+    server.shutdown()
+
+    adminZkClient.changeIpConfig(ConfigEntityName.Default, ipDefaultProps)
+    adminZkClient.changeIpConfig("1.2.3.4", ipOverrideProps)
+
+    // Remove config change znodes to force quota initialization only through loading of ip quotas
+    zkClient.getChildren(ConfigEntityChangeNotificationZNode.path).foreach { p => zkClient.deletePath(ConfigEntityChangeNotificationZNode.path + "/" + p) }
```
dajac: nit: Could we break this long line?

core/src/test/scala/unit/kafka/server/DynamicConfigChangeTest.scala
```diff
@@ -193,6 +194,64 @@ class DynamicConfigChangeTest extends KafkaServerTestHarness {
...
+    ipDefaultProps.put(DynamicConfig.Ip.IpConnectionRateOverrideProp, "20")
+
+    val overriddenIp = InetAddress.getByName("1.2.3.4")
+    val defaultQuotaIp = InetAddress.getByName("2.3.4.5")
+    adminZkClient.changeIpConfig(ConfigEntityName.Default, ipDefaultProps)
+    adminZkClient.changeIpConfig(overriddenIp.getHostAddress, ipOverrideProps)
+
+    val connectionQuotas = servers.head.socketServer.connectionQuotas
+
+    TestUtils.retry(10000) {
```
dajac: This block is somewhat repeated multiple times in the test. How about defining a `verifyConnectionQuota` helper which verifies the quota of a given IP?

core/src/test/scala/unit/kafka/network/SocketServerTest.scala
```diff
@@ -880,6 +872,72 @@ class SocketServerTest {
...
+    overrideProps.put(KafkaConfig.NumQuotaSamplesProp, String.valueOf(2))
+    val connectionRate = 5
+    val overrideServer = new SocketServer(KafkaConfig.fromProps(overrideProps), new Metrics(), Time.SYSTEM, credentialProvider)
+    overrideServer.connectionQuotas.updateIpConnectionRateQuota(None, Some(connectionRate))
+    try {
+      overrideServer.startup()
+      // make the maximum allowable number of connections
+      (0 until connectionRate).map(_ => connect(overrideServer))
+      // now try one more (should get throttled)
+      var conn = connect(overrideServer)
```
dajac, line 889: I wonder if this could be a source of flakiness. If I understand this correctly, it works because all the connections are within the same sample interval (1s). Am I understanding it right? That's probably true most of the time but could be false sometimes, especially on overloaded Jenkins machines. Could we mock the time to avoid this?

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
```diff
@@ -477,6 +568,99 @@ class ConnectionQuotasTest {
...
+    connectionQuotas.updateIpConnectionRateQuota(None, Some(defaultOverrideRate))
+
+    // external listener IP should not have its quota updated to the new default
+    verifyIpConnectionQuota(externalListener.defaultIp, overrideIpRate)
+    // admin listener IP should have its quota updated with to the new default
+    verifyIpConnectionQuota(adminListener.defaultIp, defaultOverrideRate)
+
+    // remove default connection rate quota
+    connectionQuotas.updateIpConnectionRateQuota(None, None)
+    verifyIpConnectionQuota(adminListener.defaultIp, DynamicConfig.Ip.DefaultConnectionCreationRate)
```
dajac, line 643: Could we verify here that the `externalListener` still has the correct quota?

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
```diff
@@ -477,6 +568,99 @@ class ConnectionQuotasTest {
     assertTrue("Expected BlockedPercentMeter metric for EXTERNAL listener to be recorded", blockedPercentMeters("EXTERNAL").count() > 0)
   }
 
+  @Test
+  def testIpConnectionRateUpdate(): Unit = {
+    val config = KafkaConfig.fromProps(brokerPropsWithDefaultConnectionLimits)
+    connectionQuotas = new ConnectionQuotas(config, Time.SYSTEM, metrics)
+    connectionQuotas.addListener(config, listeners("EXTERNAL").listenerName)
+    connectionQuotas.addListener(config, listeners("ADMIN").listenerName)
+    connectionQuotas.addListener(config, listeners("REPLICATION").listenerName)
```
dajac: It seems that `REPLICATION` is never used in the test. We may remove it.

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
```diff
@@ -477,6 +568,99 @@ class ConnectionQuotasTest {
...
+    connectionQuotas.updateIpConnectionRateQuota(None, None)
+    verifyIpConnectionQuota(adminListener.defaultIp, DynamicConfig.Ip.DefaultConnectionCreationRate)
+
+    // remove override for external listener IP
+    connectionQuotas.updateIpConnectionRateQuota(Some(externalListener.defaultIp.getHostAddress), None)
+    verifyIpConnectionQuota(externalListener.defaultIp, DynamicConfig.Ip.DefaultConnectionCreationRate)
+  }
+
+  @Test
+  def testIpConnectionRateQuotaUpdate(): Unit = {
```
dajac: We may find better names for this test and/or the previous one. What is the difference between `testIpConnectionRateUpdate` and `testIpConnectionRateQuotaUpdate`?

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
```diff
@@ -477,6 +568,99 @@ class ConnectionQuotasTest {
...
+
+    // set a default IP quota, verify that quota gets propagated
+    connectionQuotas.updateIpConnectionRateQuota(None, Some(ipConnectionRateLimit))
+    assertThrows[ConnectionThrottledException] {
+      acceptConnections(connectionQuotas, adminListener, connectionRate)
+    }
+    assertEquals(s"Number of connections on $adminListener:",
+      connectionRate + ipConnectionRateLimit, connectionQuotas.get(adminListener.defaultIp))
+
+
```
dajac: nit: An empty line could be removed.

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
```diff
@@ -633,18 +837,27 @@ class ConnectionQuotasTest {
...
       // this method may block if broker-wide or listener limit on the number of connections is reached
-      connectionQuotas.inc(listenerName, address, blockedPercentMeters(listenerName.value))
-
+      try {
+        connectionQuotas.inc(listenerName, address, blockedPercentMeters(listenerName.value))
+      } catch {
+        case e: ConnectionThrottledException =>
+          if (!expectIpThrottle)
+            throw e
+          ipThrottled = true
```
dajac, line 897: When we hit this line, should we just stopped trying more connections? `ipThrottled` will remain true regardless.

core/src/main/scala/kafka/server/DynamicConfig.scala
```diff
@@ -127,6 +127,22 @@ object DynamicConfig {
     def validate(props: Properties) = DynamicConfig.validate(userConfigs, props, customPropsAllowed = false)
   }
 
+  object Ip {
+    val IpConnectionRateOverrideProp = "connection_creation_rate"
+    val UnlimitedConnectionCreationRate = Int.MaxValue
+    val DefaultConnectionCreationRate = UnlimitedConnectionCreationRate
+    val IpOverrideDoc = "A long representing the upper bound of connections accepted for the specified IP."
```
dajac: Here we say `long` but it is actually an `INT` that is defined below.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1453,6 +1582,47 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+    val quotaEnabled = connectionRateQuota != DynamicConfig.Ip.UnlimitedConnectionCreationRate
+    if (quotaEnabled) {
+      val sensor = getOrCreateConnectionRateQuotaSensor(connectionRateQuota, IpQuotaEntity(address))
+      val timeMs = time.milliseconds
+      val throttleMs = recordAndGetThrottleTimeMs(sensor, timeMs)
+      if (throttleMs > 0) {
+        trace(s"Throttling $address for $throttleMs ms")
+        // unrecord the connection since we won't accept the connection
+        sensor.record(-1.0, timeMs, false)
+        unrecordListenerConnection(listenerName, timeMs)
```
dajac, line 1629: It seems that we don't have any unit tests which exercise this un-recording logic. I wonder if we could add a unit test that verifies that the listener rate does not change when a connection is rejected due to violating the per-ip quota or something along these lines.

dajac (commented):
@splett2 Thanks for the updated PR. We are pretty close to merging it. I have left some more comments.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: yeah, it's undesirable behavior. However, this is in line with `ClientQuotaManager.updateQuota`, which write locks when updating quota, and also has to iterate over all metrics for any default quota updates.

I can't think of a good alternative, but I'm open to suggestions.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: I explained elsewhere, but my comment on may have gotten lost in the weeds.

for reference:

I did not use the java DelayQueue similar to the implementation in ClientQuotaManager because this throttling implementation does not use timeout-based polling or require a synchronized data structure, and there's a bit more boilerplate needed for using a DelayQueue.

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
splett2: I renamed to `testIpConnectionRateMetricUpdate` and `testEnforcedIpConnectionRateQuotaUpdate` to clarify.

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
splett2: a lot of the tests in `ConnectionQuotasTest` are using system time and are looking to reach a connection rate of N/s. so it may be the case that `ipThrottled` will remain true, but we want to verify that we can hit a target connection rate, even if connections in the middle of the interval get throttled.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: Moved IP resolution to `IpConfigHandler`.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: After thinking on this a bit more, I think that only locking on updates to `defaultConnectionRatePerIp` should be sufficient for correctness.

ZK dynamic config changes are processed within one thread, so we will only have one thread executing in `updateIpConnectionRateQuota`. If we want to be really careful about this, we can have `updateIpConnectionRateQuota` synchronized on `ConnectionQuotas`.

The case we want to avoid by synchronizing on `counts` is that thread 1 reads `defaultConnectionRatePerIp` as connection rate limit `A` while calling `inc()`, then thread 2 updates connection rate and quota metric config to `B`, then thread 1 resumes execution and creates a sensor/metric with quota limit `A` => inconsistency.

If we synchronize on `counts` for only updates to `connectionRateForIp/defaultConnectionRate`, we know that thread 1 that has read a connection rate quota as `A` will finish creating quota metrics with quota metric config `A` before thread 2 acquires the `counts` lock and updates `connectionRateForIp/defaultConnectionRate` to `B`. 

After thread 2 releases the `counts` lock, subsequent threads calling `inc()` will read the quota as `B` and create a metric as `B`. Thread 2 can then be able to update any quota metrics from `A` to `B`, without holding the `counts` lock knowing that there are no operations that could have read the default connection rate limit as `A` without already having finished created the sensor with quota as `A`, and that all subsequent quotas will be read and created as `B`.

The only issue remaining is that we can get concurrent reads of `connectionRatePerIp` while updating quota metrics, but we can just replace `mutable.Map` with `ConcurrentHashMap` which is preferable to coarsely locking on `counts`.

Let me know if I'm missing something here with respect to thread safety.

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
splett2: I took a closer look at some of the IP throttling tests, and I think they are a little too coarse-grained with respect to testing the quota rate being enforced. I updated the IP throttling rate above/below tests that don't expect to block to use `acceptConnectionsAndVerifyRate` with `MockTime` to verify that the expected quota rate holds while getting throttled.

It's a little messy, because a lot of the other tests for broker/listener quotas have to use `Time.SYSTEM` due to monitor waiting. I think it should also be possible to rewrite the broker/listener connection rate quota tests to use mock time by extending the `Time` interface, but I would prefer to do that in a follow-up PR.

core/src/main/scala/kafka/network/SocketServer.scala
dajac: Thanks for the clarification. I have missed that previous comment.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1324,7 +1404,60 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+  def updateIpConnectionRateQuota(ip: Option[InetAddress], maxConnectionRate: Option[Int]): Unit = synchronized {
+    def isIpConnectionRateMetric(metricName: MetricName) = {
+      metricName.name == ConnectionRateMetricName &&
+      metricName.group == MetricsGroup &&
+      metricName.tags.containsKey(IpMetricTag)
+    }
+
+    def shouldUpdateQuota(metric: KafkaMetric, quotaLimit: Int) = {
+      quotaLimit != metric.config.quota.bound
+    }
```
dajac, line 1427: nit: Could we add an empty line after this one?

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1324,7 +1404,60 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+    ip match {
+      case Some(address) =>
+        counts.synchronized {
```
dajac, line 1432: It would be great to add a comment here explaining why we do this. As `connectionRatePerIp` is a ConcurrentMap, it is not obvious.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1324,7 +1404,60 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+              info(s"Updating max connection rate override for $address to $rate")
+              connectionRatePerIp.put(address, rate)
+            case None =>
+              info(s"Removing max connection rate override for $address")
+              connectionRatePerIp.remove(address)
+          }
+        }
+        updateConnectionRateQuota(connectionRateForIp(address), IpQuotaEntity(address))
+      case None =>
+        counts.synchronized {
```
dajac, line 1445: We could also add a similar comment here.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1324,7 +1404,60 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+          defaultConnectionRatePerIp = maxConnectionRate.getOrElse(DynamicConfig.Ip.DefaultConnectionCreationRate)
+        }
+        info(s"Updated default max IP connection rate to $defaultConnectionRatePerIp")
+        metrics.metrics.forEach { (metricName, metric) =>
+          if (isIpConnectionRateMetric(metricName)) {
+            val quota = connectionRateForIp(InetAddress.getByName(metricName.tags.get(IpMetricTag)))
```
dajac, line 1451: It may be better to keep IP as string in `connectionRateForIp` in order to avoid having to resolve them when we update the metrics. I don't see any reason to `InetAddress` internally. Is there any? `InetAddress.getByName` seems sufficient when we validate the configuration in the config handler.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1476,33 +1650,36 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
...
+   * @param connectionQuotaEntity entity to create the sensor for
    */
-  private def createConnectionRateQuotaSensor(quotaLimit: Int, listenerOpt: Option[String] = None): Sensor = {
-    val sensorName = listenerOpt.map(listener => s"ConnectionAcceptRate-$listener").getOrElse("ConnectionAcceptRate")
-    val sensor = metrics.sensor(sensorName, rateQuotaMetricConfig(quotaLimit))
-    sensor.add(connectionRateMetricName(listenerOpt), new Rate, null)
-    info(s"Created $sensorName sensor, quotaLimit=$quotaLimit")
-    sensor
+  private def getOrCreateConnectionRateQuotaSensor(quotaLimit: Int, connectionQuotaEntity: ConnectionQuotaEntity): Sensor = {
+    Option(metrics.getSensor(connectionQuotaEntity.sensorName)).getOrElse {
```
dajac, line 1661: Could we also add a comment here which states that we don't explicitly protect this section because we hold the `counts` lock.

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
dajac: Sounds good. Let's try to clean this up in the follow-up PR.

core/src/main/scala/kafka/network/SocketServer.scala
dajac: Thanks for the explanation, David. I agree with your approach. In the end, we want to ensure that we don't create a Sensor/Metrics while `connectionRateForIp/defaultConnectionRate` are updated. It think that it is fine to rely `counts` lock to achieve this but we should add some comments to make this clear. I was considering if we should use a different and more explicit lock to achieve this but I am fine with using `counts` as long as we add good comments.

Having `updateIpConnectionRateQuota` synchronized on `ConnectionQuotas` sounds good to me. I think that we have to ensure that metrics are updated by one thread only. Otherwise, we may have inconsistencies.

dajac (commented):
@splett2 Thanks for the updated PR. It looks good overall. I have left few minor comments.

core/src/main/scala/kafka/network/SocketServer.scala
apovzner: I reviewed the commit with the locking change, and it looks correct to me. Thanks for writing up a thorough evaluation of the new logic. I think using `counts` as a lock is reasonable here, because we need to synchronize a lot of the same data that is used on the `inc()` path (protected by `counts`).  I agree about adding clear comments.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: I will add this detail to `recordIpConnectionMaybeThrottle`. I think that is a more fitting place to put the comment, since that's the section of code that calls `connectionRateForIp` and `getOrCreateConnectionRateQuotaSensor` which are the components we need to be atomic.

core/src/main/scala/kafka/network/SocketServer.scala
splett2: reasons for using `InetAddress` would be that it's more consistent with the other IP-related data structures in `ConnectionQuotas`, `maxConnectionsPerIpOverrides` and `counts`.

additionally, `inc()` is called with an `InetAddress`, so it's convenient to keep our data structure using `InetAddress` as a key so that we don't need to convert `InetAddress` => `String` in cases where we would not create a sensor (e.g., when IP quotas are disabled).

core/src/test/scala/unit/kafka/network/ConnectionQuotasTest.scala
splett2: Thanks. I filed https://issues.apache.org/jira/browse/KAFKA-10744 for the clean up/conversion.

core/src/main/scala/kafka/network/SocketServer.scala
dajac: Make sense, thanks.

dajac (approved):
@splett2 LGTM! Great work, thanks!

dajac:
Builds are green. Merging to trunk.

PR #9395, opened 2020-10-08
KAFKA-9726: Add LegacyReplicationPolicy for MM2

ivanyu:
This commit adds a new replication policy for MirrorMaker 2, `LegacyReplicationPolicy`. This policy imitates MirrorMaker 1 behavior of not renaming replicated topics. The exception is made for `heartbeats` topic, that is replicated according to `DefaultReplicationPolicy`.

Avoiding renaming topics brings a number of limitations, among which the most important one is the impossibility of detecting replication cycles. This makes cross-replication using `LegacyReplicationPolicy` effectively impossible. See `LegacyReplicationPolicy` Javadoc for details.

A new method `canTrackSource` is added to `ReplicationPolicy`. Its result indicates if the replication policy can track back to the source topic of a topic. It is needed to allow detecting target topics work when `LegacyReplicationPolicy` is used.

On the testing side, the tests have the same strategy as for `DefaultReplicationPolicy` with nicessary adjustments (e.g. no active/active replication is tested).

### DIFF
--- a/connect/mirror-client/src/main/java/org/apache/kafka/connect/mirror/LegacyReplicationPolicy.java
+++ b/connect/mirror-client/src/main/java/org/apache/kafka/connect/mirror/LegacyReplicationPolicy.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.connect.mirror;
+
+import org.apache.kafka.common.Configurable;
+
+import java.util.Map;
+
+import static org.apache.kafka.connect.mirror.MirrorClientConfig.HEARTBEATS_TOPIC;
+
+/**
+ * The replication policy that imitates the behavior of MirrorMaker 1.
+ *
+ * <p>The policy doesn't rename topics: {@code topic1} remains {@code topic1} after replication.
+ * There is one exception to this: for {@code heartbeats}, it behaves identical to {@link DefaultReplicationPolicy}.
+ *
+ * <p>The policy has some notable limitations. The most important one is that the policy is unable to detect
+ * cycles for any topic apart from {@code heartbeats}. This makes cross-replication effectively impossible.
+ *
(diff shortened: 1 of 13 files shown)


connect/mirror-client/src/main/java/org/apache/kafka/connect/mirror/ReplicationPolicy.java
```diff
@@ -57,4 +57,9 @@ default boolean isInternalTopic(String topic) {
         return topic.endsWith(".internal") || topic.endsWith("-internal") || topic.startsWith("__")
             || topic.startsWith(".");
     }
+
+    /** Checks if the policy can track back to the source of the topic. */
```
ryannedolan, line 61: I'm not sure what you mean by "track back to the source of the topic". The word "track" might mean a few things here, and it's not obvious what you mean. Can you clarify?

connect/mirror-client/src/main/java/org/apache/kafka/connect/mirror/ReplicationPolicy.java
```diff
@@ -57,4 +57,9 @@ default boolean isInternalTopic(String topic) {
...
+    default boolean canTrackSource(String topic) {
```
ryannedolan, line 62: If a public API change like this is required, you will need to propose a small KIP. I'm unclear why it's required tho, and ideally we would not alter the existing API if possible.

If a new method is required, I think "track" is too ambiguous and should not be used here.

connect/mirror-client/src/main/java/org/apache/kafka/connect/mirror/LegacyReplicationPolicy.java
```diff
@@ -0,0 +1,96 @@
...
+            return topic;
+        }
+    }
+
+    @Override
+    public String topicSource(final String topic) {
+        if (isOriginalTopicHeartbeats(topic)) {
+            return heartbeatTopicReplicationPolicy.topicSource(topic);
+        } else {
+            return null;
```
ryannedolan, line 66: I've seen alternative solutions floating around that use a configurable source here. Basically, the configuration passed to configure() is consulted to find the "source cluster", rather than looking at the topic name. That approach lets you return an actual source here, which obviates the new canTrackSource() method etc.

ivanyu: I've explored this possibility, too. The main problem with it is that the replication policy should answer differently for source and target clusters. It's essential for methods like `MirrorSourceConnector.isCycle` and `MirrorClient.remoteTopics`. For a source, `topicSource` should return `null`; for a target, a predefined value.

It leaves two possibility. In one, we set up two different replication policy instances with different configurations, e.g.:
```
replication.policy.source.class=org.apache.kafka.connect.mirror.LegacyReplicationPolicy
replication.policy.source.source=
replication.policy.target.class=org.apache.kafka.connect.mirror.LegacyReplicationPolicy
replication.policy.target.source=primary-cluster
```

Of course, we can make that the current configurations work as before.

Another possibility is to modify the `ReplicationPolicy` interface to allow it to pass additional information out (like `canTrackSource` or similar) or in (like `topicSource(String topic, boolean isSourceCluster)`).

What do you think would be the best approach?

mdedetrich:
@ivanyu Can you close this since a new PR has been created that is up to date with the latest `trunk`?

ivanyu:
Yep, I'm closing this one

PR #9396, opened 2020-10-08
KAFKA-10437: Implement new PAPI support for test-utils

vvcephei:
Implements KIP-478 for the test-utils module:
* adds mocks of the new ProcessorContext and StateStoreContext
* adds tests that all stores and store builders are usable with the new mock
* adds tests that the new Processor api is usable with the new mock
* updates the demonstration Processor to the new api

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -194,13 +194,13 @@
               files=".*[/\\]streams[/\\].*test[/\\].*.java"/>
 
     <suppress checks="CyclomaticComplexity"
-              files="(EosBetaUpgradeIntegrationTest|KStreamKStreamJoinTest|KTableKTableForeignKeyJoinIntegrationTest|RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapterTest|RelationalSmokeTest).java"/>
+              files="(EosBetaUpgradeIntegrationTest|KStreamKStreamJoinTest|KTableKTableForeignKeyJoinIntegrationTest|RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapterTest|RelationalSmokeTest|MockProcessorContextStateStoreTest).java"/>
 
     <suppress checks="JavaNCSS"
               files="(EosBetaUpgradeIntegrationTest|KStreamKStreamJoinTest|TaskManagerTest).java"/>
 
     <suppress checks="NPathComplexity"
-              files="(EosBetaUpgradeIntegrationTest|EosTestDriver|KStreamKStreamJoinTest|KTableKTableForeignKeyJoinIntegrationTest|RelationalSmokeTest).java"/>
+              files="(EosBetaUpgradeIntegrationTest|EosTestDriver|KStreamKStreamJoinTest|KTableKTableForeignKeyJoinIntegrationTest|RelationalSmokeTest|MockProcessorContextStateStoreTest).java"/>
 
     <suppress checks="(FinalLocalVariable|WhitespaceAround|LocalVariableName|ImportControl|AvoidStarImport)"
               files="Murmur3Test.java"/>
--- a/streams/src/main/java/org/apache/kafka/streams/processor/api/Record.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/api/Record.java
@@ -20,6 +20,8 @@ import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeaders;
 import org.apache.kafka.streams.errors.StreamsException;
 
+import java.util.Objects;
+
 /**
  * A data class representing an incoming record for processing in a {@link Processor}
  * or a record to forward to downstream processors via {@link ProcessorContext}.
@@ -161,5 +163,31 @@ public class Record<K, V> {
      */
     public Record<K, V> withHeaders(final Headers headers) {
         return new Record<>(key, value, timestamp, headers);
+    }
+
+    @Override
+    public String toString() {
+        return "Record{" +
+            "key=" + key +
+            ", value=" + value +
+            ", timestamp=" + timestamp +
+            ", headers=" + headers +
+            '}';
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        final Record<?, ?> record = (Record<?, ?>) o;
+        return timestamp == record.timestamp &&
+            Objects.equals(key, record.key) &&
+            Objects.equals(value, record.value) &&
+            Objects.equals(headers, record.headers);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(key, value, timestamp, headers);
     }
 }
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/InMemorySessionStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/InMemorySessionStore.java
@@ -75,18 +75,25 @@ public class InMemorySessionStore implements SessionStore<Bytes, byte[]> {
     @Deprecated
     @Override
     public void init(final ProcessorContext context, final StateStore root) {
-        this.context = (InternalProcessorContext) context;
-
-        final StreamsMetricsImpl metrics = this.context.metrics();
         final String threadId = Thread.currentThread().getName();
         final String taskName = context.taskId().toString();
-        expiredRecordSensor = TaskMetrics.droppedRecordsSensorOrExpiredWindowRecordDropSensor(
-            threadId,
-            taskName,
-            metricScope,
-            name,
-            metrics
-        );
+
+        // The provided context is not required to implement InternalProcessorContext,
+        // If it doesn't, we can't record this metric.
+        if (context instanceof InternalProcessorContext) {
+            this.context = (InternalProcessorContext) context;
+            final StreamsMetricsImpl metrics = this.context.metrics();
+            expiredRecordSensor = TaskMetrics.droppedRecordsSensorOrExpiredWindowRecordDropSensor(
+                threadId,
+                taskName,
+                metricScope,
+                name,
+                metrics
+            );
+        } else {
+            this.context = null;
+            expiredRecordSensor = null;
+        }
 
         if (root != null) {
             context.register(root, (key, value) -> put(SessionKeySchema.from(Bytes.wrap(key)), value));
@@ -102,7 +109,11 @@ public class InMemorySessionStore implements SessionStore<Bytes, byte[]> {
         observedStreamTime = Math.max(observedStreamTime, windowEndTimestamp);
 
         if (windowEndTimestamp <= observedStreamTime - retentionPeriod) {
-            expiredRecordSensor.record(1.0d, context.currentSystemTimeMs());
+            // The provided context is not required to implement InternalProcessorContext,
+            // If it doesn't, we can't record this metric (in fact, we wouldn't have even initialized it).
+            if (expiredRecordSensor != null && context != null) {
+                expiredRecordSensor.record(1.0d, context.currentSystemTimeMs());
+            }
             LOG.warn("Skipping record for expired segment.");
         } else {
             if (aggregate != null) {
--- a/streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
+++ b/streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
@@ -28,7 +28,6 @@ import org.apache.kafka.streams.TopologyTestDriver;
 import org.apache.kafka.streams.kstream.Transformer;
 import org.apache.kafka.streams.kstream.ValueTransformer;
 import org.apache.kafka.streams.processor.Cancellable;
-import org.apache.kafka.streams.processor.Processor;
 import org.apache.kafka.streams.processor.PunctuationType;
 import org.apache.kafka.streams.processor.Punctuator;
 import org.apache.kafka.streams.processor.StateRestoreCallback;
--- a/streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
+++ b/streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
@@ -1,0 +1,495 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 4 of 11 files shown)


checkstyle/suppressions.xml
```diff
@@ -194,13 +194,13 @@
               files=".*[/\\]streams[/\\].*test[/\\].*.java"/>
 
     <suppress checks="CyclomaticComplexity"
-              files="(EosBetaUpgradeIntegrationTest|KStreamKStreamJoinTest|KTableKTableForeignKeyJoinIntegrationTest|RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapterTest|RelationalSmokeTest).java"/>
+              files="(EosBetaUpgradeIntegrationTest|KStreamKStreamJoinTest|KTableKTableForeignKeyJoinIntegrationTest|RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapterTest|RelationalSmokeTest|MockProcessorContextStateStoreTest).java"/>
```
vvcephei, line 197: This test violates both measures of complexity by virtue of the way it works: the test specifically includes a lot of loops so that it can generate every combination of store and store builder configuration to verify that everything works as expected.

streams/src/main/java/org/apache/kafka/streams/processor/api/Record.java
```diff
@@ -162,4 +164,30 @@ public Headers headers() {
     public Record<K, V> withHeaders(final Headers headers) {
         return new Record<>(key, value, timestamp, headers);
     }
+
+    @Override
+    public String toString() {
```
vvcephei, line 169: added for quality-of-life debugging

streams/src/main/java/org/apache/kafka/streams/processor/api/Record.java
```diff
@@ -162,4 +164,30 @@ public Headers headers() {
...
+        return "Record{" +
+            "key=" + key +
+            ", value=" + value +
+            ", timestamp=" + timestamp +
+            ", headers=" + headers +
+            '}';
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
vvcephei, line 179: added for easier assertions in unit tests

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemorySessionStore.java
```diff
@@ -75,18 +75,25 @@ public String name() {
...
-            threadId,
-            taskName,
-            metricScope,
-            name,
-            metrics
-        );
+
+        // The provided context is not required to implement InternalProcessorContext,
+        // If it doesn't, we can't record this metric.
+        if (context instanceof InternalProcessorContext) {
```
vvcephei, line 83: As a testament to `MockProcessorContextStateStoreTest`, it actually found this bug. I had overlooked this usage while making the other root stores context-implementation agnostic in the last PR.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
```diff
@@ -0,0 +1,494 @@
...
+ * <p>
+ * The tests for this class (org.apache.kafka.streams.MockProcessorContextTest) include several behavioral
+ * tests that serve as example usage.
+ * <p>
+ * Note that this class does not take any automated actions (such as firing scheduled punctuators).
+ * It simply captures any data it witnesses.
+ * If you require more automated tests, we recommend wrapping your {@link Processor} in a minimal source-processor-sink
+ * {@link Topology} and using the {@link TopologyTestDriver}.
+ */
+public class MockProcessorContext<KForward, VForward> implements ProcessorContext<KForward, VForward>, RecordCollector.Supplier {
```
vvcephei, line 69: This is a clone of the `processor.MockProcessorContext`, but implementing the new `api.ProcessorContext` interface.

The two context interfaces don't technically collide, but creating a standalone implementation allows us to provide type safety in the CapturedForward inner class without any backward compatibility concerns.

It also (IMO) makes it easier for users to make a clean switch to the new API by removing any ambiguity about whether a test is making use of the old API or the new API, or some mixture of both.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
```diff
@@ -0,0 +1,494 @@
...
+
+    @Override
+    public File stateDir() {
+        return Objects.requireNonNull(
+            stateDir,
+            "The stateDir constructor argument was needed (probably for a state store) but not supplied. " +
+                "You can either reconfigure your test so that it doesn't need access to the disk " +
+                "(such as using an in-memory store), or use the full MockProcessorContext constructor to supply " +
+                "a non-null stateDir argument."
+        );
```
vvcephei, line 300: `stateDir` isn't a required argument, but having a null value for it and then trying to use a rocksdb store leads to a _very_ cryptic error like, "you have to open all column families". Placing a null-check right here allows us to fail fast and prints an explanatory message any time you use the context with a component that actually needs the stateDir.

This was also a problem with the prior implementation, but I didn't notice it until I implemented `MockProcessorContextStateStoreTest`.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    /**
+     * Used to get a {@link StateStoreContext} for use with
+     * {@link StateStore#init(StateStoreContext, StateStore)}
+     * if you need to initialize a store for your tests.
+     * @return a {@link StateStoreContext} that delegates to this ProcessorContext.
+     */
+    public StateStoreContext getStateStoreContext() {
```
vvcephei, line 446: Rather than have the MockProcessorContext implement StateStoreContext as well, I think it's a good idea to explicitly get a StateStoreContext view of the same object. This will help users to have a clearer understanding of which context is which.

streams/test-utils/src/test/java/org/apache/kafka/streams/MockProcessorContextTest.java
```diff
@@ -43,6 +43,7 @@
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+@SuppressWarnings("deprecation") // this is a test of a deprecated API
```
vvcephei, line 46: This is the pre-existing test of the old-API ProcessorContext.

streams/test-utils/src/test/java/org/apache/kafka/streams/MockProcessorContextTest.java
```diff
@@ -231,14 +230,13 @@ public void process(final String key, final Long value) {
 
         assertFalse(context.committed());
     }
-    @SuppressWarnings({"deprecation", "unchecked"}) // TODO deprecation will be fixed in KAFKA-10437
```
vvcephei, line 234: Don't need to suppress deprecation anymore, since it's moved to the class level, and don't need unchecked anymore thanks to the return-type change of `getStateStore`.

streams/test-utils/src/test/java/org/apache/kafka/streams/internals/KeyValueStoreFacadeTest.java
```diff
@@ -53,9 +54,9 @@ public void setup() {
         keyValueStoreFacade = new KeyValueStoreFacade<>(mockedKeyValueTimestampStore);
     }
 
-    @SuppressWarnings("deprecation") // TODO will be fixed in KAFKA-10437
+    @SuppressWarnings("deprecation") // test of deprecated method
```
vvcephei, line 57: My prior comment was wrong. We do want to test the deprecated init method as well as the non-deprecated one.

streams/test-utils/src/test/java/org/apache/kafka/streams/test/MockProcessorContextAPITest.java
```diff
@@ -0,0 +1,353 @@
...
+import static java.util.Arrays.asList;
+import static java.util.Collections.singletonList;
+import static org.apache.kafka.common.utils.Utils.mkEntry;
+import static org.apache.kafka.common.utils.Utils.mkMap;
+import static org.apache.kafka.common.utils.Utils.mkProperties;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.is;
+
+public class MockProcessorContextAPITest {
```
vvcephei, line 50: This is a copy of the unit test for the old MockProcessorContext.

streams/test-utils/src/test/java/org/apache/kafka/streams/test/MockProcessorContextStateStoreTest.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.List;
+
+import static java.util.Arrays.asList;
+import static org.apache.kafka.common.utils.Utils.mkEntry;
+import static org.apache.kafka.common.utils.Utils.mkMap;
+import static org.apache.kafka.common.utils.Utils.mkProperties;
+import static org.junit.Assert.assertThrows;
+
+@RunWith(value = Parameterized.class)
+public class MockProcessorContextStateStoreTest {
```
vvcephei, line 53: This is a new test. I wanted to make sure that no matter what kind of stores users select, they'll get well-defined results with the new MockProcessorContext.

Some store features are well-defined not to be supported (they throw an IllegalArgumentException when you try to enable caching or logging), but all store types should otherwise work.

Here, we're defining "work" as "`init` doesn't throw an exception", since this is the main involvement that the context itself has with the store.

streams/test-utils/src/test/java/org/apache/kafka/streams/test/wordcount/WindowedWordCountProcessorSupplier.java
```diff
@@ -18,43 +18,41 @@
...
+import org.apache.kafka.streams.processor.api.ProcessorSupplier;
+import org.apache.kafka.streams.processor.api.Record;
 import org.apache.kafka.streams.state.KeyValueIterator;
 import org.apache.kafka.streams.state.WindowStore;
 
 import java.time.Duration;
 import java.util.Locale;
 
-public final class WindowedWordCountProcessorSupplier implements ProcessorSupplier<String, String> {
+public final class WindowedWordCountProcessorSupplier implements ProcessorSupplier<String, String, String, String> {
```
vvcephei, line 32: Converting this processor to the new API.

streams/test-utils/src/test/java/org/apache/kafka/streams/test/wordcount/WindowedWordCountProcessorTest.java
```diff
@@ -31,19 +32,17 @@
...
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertThrows;
 
 public class WindowedWordCountProcessorTest {
-    @SuppressWarnings("deprecation") // TODO will be fixed in KAFKA-10437
     @Test
     public void shouldWorkWithInMemoryStore() {
-        final MockProcessorContext context = new MockProcessorContext();
+        final MockProcessorContext<String, String> context = new MockProcessorContext<>();
```
vvcephei, line 45: Converting the test for the processor to the new API as well.

vvcephei (commented):
Hi @guozhangwang , Any chance I could trouble you for one more review to get KIP-478 over the line?

This is the last Public API change that needs to go into 2.7.0 . 

Thanks,
-John

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
guozhangwang: Since we are adding a new class, could we have it extend InternalProcessorContext then, so that when we remove the old ones we can also cleanup the non-testing functions that branch on checking `instanceof InternalProcessorContext`: I'm assuming that `InternalProcessorContext` would stay in the end state, it would just extend `api.ProcessorContext<Object, Object>` and not `StateStoreContext` in the future.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
guozhangwang: I agree. I think in the long run MockProcessorContext could extend InternalProcessorContext, which only extends api.ProcessorContext. When that's the case we should be able to remove the non-testing code's safeguards.

streams/test-utils/src/test/java/org/apache/kafka/streams/test/wordcount/WindowedWordCountProcessorTest.java
```diff
@@ -136,54 +133,18 @@ public void shouldWorkWithPersistentStore() throws IOException {
...
+
+            assertThat(capturedForwards, is(expected));
         } finally {
             Utils.delete(stateDir);
         }
     }
-
-    @SuppressWarnings("deprecation") // TODO will be fixed in KAFKA-10437
-    @Test
-    public void shouldFailWithLogging() {
```
guozhangwang, line 153: Can we remove those two tests?

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
vvcephei: Unfortunately, that would actually expose `InternalProcessorContext` itself as a public interface (any supertype of a public type is a public type).

streams/test-utils/src/test/java/org/apache/kafka/streams/test/wordcount/WindowedWordCountProcessorTest.java
vvcephei: Yes, they are superceded by the new `MockProcessorContextStateStoreTest`, which verifies that any state store configured with logging or caching is rejected.

streams/test-utils/src/main/java/org/apache/kafka/streams/processor/api/MockProcessorContext.java
guozhangwang: `InternalProcessorContext` is already public interface but it's in `internals` package, so I figured it is okay?

Anyways, this is not much blocking this PR, so feel free to merge it anyways and we can keep discussing here while you merge.

vvcephei: Thanks. I'll go ahead with the merge, then.

It seems like `InternalProcessorContext` is only "public" in the sense that Java 8 is insufficiently expressive to make it private. I guess I should say "internal" and "external". But it is certainly an _internal_ type, therefore, we should not ever return it to the users when they call _external_ APIs (such as the constructor of this class).

Once we upgrade to java 9, we should add module definitions so that we stop exporting internal types in the public API completely. As I understand things, it would actually become a compiler error at that point for us to provide an external class that inherits from an internal interface.

I guess I don't see this as a big concern anyway. The way I implemented the state stores in the last PR makes them capable of accepting any implementation of `ProcessorContext` or `StateStoreContext`, although they'll disable some features if the context isn't an instance of `InternalProcessorContext`. To me, this seems like a fine situation. We could also note that there is https://cwiki.apache.org/confluence/display/KAFKA/KIP-622%3A+Add+currentSystemTimeMs+and+currentStreamTimeMs+to+ProcessorContext, which would remove the need for casting to InternalProcessorContext to begin with.

vvcephei:
The test failure was unrelated:
`Build / JDK 11 / kafka.api.PlaintextAdminIntegrationTest.testAlterReplicaLogDirs`

vvcephei:

PR #9397, opened 2020-10-08
KAFKA-10583: Add documentation on the thread-safety of KafkaAdminClient.

efeg:
Other than a Stack Overflow comment (see https://stackoverflow.com/a/61738065) by Colin Patrick McCabe (@cmccabe ) and a proposed design note on KIP-117 wiki, there is no source that verifies the thread-safety of `KafkaAdminClient`.

This patch updates JavaDoc of `KafkaAdminClient` to clarify its thread-safety.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
+++ b/clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
@@ -51,6 +51,9 @@ import org.apache.kafka.common.requests.LeaveGroupResponse;
  * This client was introduced in 0.11.0.0 and the API is still evolving. We will try to evolve the API in a compatible
  * manner, but we reserve the right to make breaking changes in minor releases, if necessary. We will update the
  * {@code InterfaceStability} annotation and this notice once the API is considered stable.
+ * <p>
+ * Instances returned from the {@code create} methods of this interface are guaranteed to be thread safe.
+ * </p>
  */
 @InterfaceStability.Evolving
 public interface Admin extends AutoCloseable {
--- a/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
+++ b/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
@@ -287,6 +287,9 @@ import static org.apache.kafka.common.utils.Utils.closeQuietly;
  * The default implementation of {@link Admin}. An instance of this class is created by invoking one of the
  * {@code create()} methods in {@code AdminClient}. Users should not refer to this class directly.
  *
+ * <p>
+ * This class is thread-safe.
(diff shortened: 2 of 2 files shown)


efeg:
@cmccabe Do you think you might be able to take a look at this PR?

tombentley:
@efeg the javadoc for `KafkaAdminClient` says "Users should not refer to this class directly", so just adding the thread-safety note to `KafkaAdminClient` probably isn't so helpful, we should add it to `Admin` too. Though since that's an interface probably some wording like "The instances returned from the {@code create} methods are guaranteed to be thread safe" would be better.

This is also an opportunity to improve the rather scant Javadoc for `Admin`. Compare it to the extensive Javadoc for `KafkaProducer` and `KafkaConsumer`.

efeg:
@tombentley thanks for the feedback!

> @efeg the javadoc for `KafkaAdminClient` says "Users should not refer to this class directly", so just adding the thread-safety note to `KafkaAdminClient` probably isn't so helpful, we should add it to `Admin` too. Though since that's an interface probably some wording like "The instances returned from the {@code create} methods are guaranteed to be thread safe" would be better.

Makes sense -- updated the JavaDoc of `Admin`.

> This is also an opportunity to improve the rather scant Javadoc for `Admin`. Compare it to the extensive Javadoc for `KafkaProducer` and `KafkaConsumer`.

I agree that `Admin` JavaDoc would benefit from a proper revision to make it on par with the other client docs. That being said, to avoid scope creep, I would prefer to keep this PR about the thread-safety of Admin clients.

tombentley (approved):

tombentley:
@cmccabe @ijuma @hachikuji any chance we could merge this trivial javadoc change?

chia7712 (approved):
LGTM

chia7712:
@tombentley I'm going to merge this PR but I failed to find out your email (used by reviewers list). The email in your PRs (for example https://github.com/apache/kafka/commit/775a08876a636cc9f4b6cfd89e305b8becf44670.patch) is ```tombentley@users.noreply.github.com``` and it is absolutely not your true email :)

tombentley:
@chia7712 `Tom Bentley <tbentley@redhat.com>` is what's used in my commits.

chia7712:
@efeg Thanks for the patch! merge it to trunk

tombentley:
@efeg do you want to have a go at writing a more thorough class Javadoc for `Admin`? If not I'll try to write something.

efeg:
> @efeg do you want to have a go at writing a more thorough class Javadoc for Admin? If not I'll try to write something.

@tombentley Thanks for your help with this PR!
Please feel free to create the PR for a more thorough `Admin` Javadoc when you have a chance.

PR #9398, opened 2020-10-08
MINOR update comments and docs to be gender-neutral

xvrl:
While this is not technically part of KIP-629, I believe this makes our codebase more inclusive as well.

cc @gwenshap


gwenshap (approved):
LGTM. Thank you.

PR #9401, opened 2020-10-09
KAFKA-9628 Replace Produce request/response with automated protocol

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-9628

### Benchmark

1. loop **30** times
1. calculate average

#### kafkatest.benchmarks.core.benchmark_test.Benchmark.test_producer_throughput

> @cluster(num_nodes=5)
> @parametrize(acks=-1, topic=TOPIC_REP_THREE)

- +0.3144915325 %
- 28.08766667 ->  28.1715625 (mb_per_sec)

> @cluster(num_nodes=5)
> @matrix(acks=[1], topic=[TOPIC_REP_THREE], message_size=[100000],compression_type=["none"], security_protocol=['PLAINTEXT'])

- +4.220730323 %
- 157.145 -> 163.7776667 (mb_per_sec)

> @cluster(num_nodes=7)
> @parametrize(acks=1, topic=TOPIC_REP_THREE, num_producers=3)

- +5.996241145%
- 57.64166667 -> 61.098 (mb_per_sec)

> @cluster(num_nodes=5)
> @parametrize(acks=1, topic=TOPIC_REP_THREE)

- +0.3979572536%
- 44.05833333 -> 44.23366667 (mb_per_sec)

> @cluster(num_nodes=5)
> @parametrize(acks=1, topic= TOPIC_REP_ONE)

- +2.228235226%
- 69.23266667 -> 70.77533333 (mb_per_sec)

### JMH results

In short, most ops performance are regression since we have to convert data to protocol data. The cost is inevitable (like other request/response) before we use protocol data directly.

### JMH for ProduceRequest

1. construction regression:
    - 281.474 -> 454.935 ns/op
    - 296.000 -> 1888.000 B/op
1. toErrorResponse regression:
    - 41.942 -> 107.528 ns/op
    - 1216.000 -> 1616.000 B/op
1. toStruct improvement:
    - 255.185 -> 90.728 ns/op
    - 864.000 -> 304.000 B/op

**BEFORE**
```
Benchmark                                                                        Mode  Cnt     Score    Error   Units
ProducerRequestBenchmark.constructorErrorResponse                                avgt   15    41.942 ±  0.036   ns/op
ProducerRequestBenchmark.constructorErrorResponse:·gc.alloc.rate                 avgt   15  6409.263 ±  5.478  MB/sec
ProducerRequestBenchmark.constructorErrorResponse:·gc.alloc.rate.norm            avgt   15   296.000 ±  0.001    B/op
ProducerRequestBenchmark.constructorErrorResponse:·gc.churn.G1_Eden_Space        avgt   15  6416.420 ± 76.071  MB/sec
ProducerRequestBenchmark.constructorErrorResponse:·gc.churn.G1_Eden_Space.norm   avgt   15   296.331 ±  3.539    B/op
ProducerRequestBenchmark.constructorErrorResponse:·gc.churn.G1_Old_Gen           avgt   15     0.002 ±  0.002  MB/sec
ProducerRequestBenchmark.constructorErrorResponse:·gc.churn.G1_Old_Gen.norm      avgt   15    ≈ 10⁻⁴             B/op
ProducerRequestBenchmark.constructorErrorResponse:·gc.count                      avgt   15   698.000           counts
ProducerRequestBenchmark.constructorErrorResponse:·gc.time                       avgt   15   378.000               ms
ProducerRequestBenchmark.constructorProduceRequest                               avgt   15   281.474 ±  3.286   ns/op
ProducerRequestBenchmark.constructorProduceRequest:·gc.alloc.rate                avgt   15  3923.868 ± 46.303  MB/sec
ProducerRequestBenchmark.constructorProduceRequest:·gc.alloc.rate.norm           avgt   15  1216.000 ±  0.001    B/op
ProducerRequestBenchmark.constructorProduceRequest:·gc.churn.G1_Eden_Space       avgt   15  3923.375 ± 59.568  MB/sec
ProducerRequestBenchmark.constructorProduceRequest:·gc.churn.G1_Eden_Space.norm  avgt   15  1215.844 ± 11.184    B/op
ProducerRequestBenchmark.constructorProduceRequest:·gc.churn.G1_Old_Gen          avgt   15     0.004 ±  0.001  MB/sec
ProducerRequestBenchmark.constructorProduceRequest:·gc.churn.G1_Old_Gen.norm     avgt   15     0.001 ±  0.001    B/op
ProducerRequestBenchmark.constructorProduceRequest:·gc.count                     avgt   15   515.000           counts
ProducerRequestBenchmark.constructorProduceRequest:·gc.time                      avgt   15   279.000               ms
ProducerRequestBenchmark.constructorStruct                                       avgt   15   255.185 ±  0.069   ns/op
ProducerRequestBenchmark.constructorStruct:·gc.alloc.rate                        avgt   15  3074.889 ±  0.823  MB/sec
ProducerRequestBenchmark.constructorStruct:·gc.alloc.rate.norm                   avgt   15   864.000 ±  0.001    B/op
ProducerRequestBenchmark.constructorStruct:·gc.churn.G1_Eden_Space               avgt   15  3077.737 ± 31.537  MB/sec
ProducerRequestBenchmark.constructorStruct:·gc.churn.G1_Eden_Space.norm          avgt   15   864.800 ±  8.823    B/op
ProducerRequestBenchmark.constructorStruct:·gc.churn.G1_Old_Gen                  avgt   15     0.003 ±  0.001  MB/sec
ProducerRequestBenchmark.constructorStruct:·gc.churn.G1_Old_Gen.norm             avgt   15     0.001 ±  0.001    B/op
ProducerRequestBenchmark.constructorStruct:·gc.count                             avgt   15   404.000           counts
ProducerRequestBenchmark.constructorStruct:·gc.time                              avgt   15   214.000               ms
```

**AFTER**
```
Benchmark                                                                        Mode  Cnt     Score    Error   Units
ProducerRequestBenchmark.constructorErrorResponse                                avgt   15   107.528 ±  0.270   ns/op
ProducerRequestBenchmark.constructorErrorResponse:·gc.alloc.rate                 avgt   15  4864.899 ± 12.132  MB/sec
ProducerRequestBenchmark.constructorErrorResponse:·gc.alloc.rate.norm            avgt   15   576.000 ±  0.001    B/op
ProducerRequestBenchmark.constructorErrorResponse:·gc.churn.G1_Eden_Space        avgt   15  4868.023 ± 61.943  MB/sec
ProducerRequestBenchmark.constructorErrorResponse:·gc.churn.G1_Eden_Space.norm   avgt   15   576.371 ±  7.331    B/op
ProducerRequestBenchmark.constructorErrorResponse:·gc.churn.G1_Old_Gen           avgt   15     0.005 ±  0.001  MB/sec
ProducerRequestBenchmark.constructorErrorResponse:·gc.churn.G1_Old_Gen.norm      avgt   15     0.001 ±  0.001    B/op
ProducerRequestBenchmark.constructorErrorResponse:·gc.count                      avgt   15   639.000           counts
ProducerRequestBenchmark.constructorErrorResponse:·gc.time                       avgt   15   339.000               ms
ProducerRequestBenchmark.constructorProduceRequest                               avgt   15   454.935 ±  0.332   ns/op
ProducerRequestBenchmark.constructorProduceRequest:·gc.alloc.rate                avgt   15  3769.014 ±  2.767  MB/sec
ProducerRequestBenchmark.constructorProduceRequest:·gc.alloc.rate.norm           avgt   15  1888.000 ±  0.001    B/op
ProducerRequestBenchmark.constructorProduceRequest:·gc.churn.G1_Eden_Space       avgt   15  3763.407 ± 31.530  MB/sec
ProducerRequestBenchmark.constructorProduceRequest:·gc.churn.G1_Eden_Space.norm  avgt   15  1885.190 ± 15.594    B/op
ProducerRequestBenchmark.constructorProduceRequest:·gc.churn.G1_Old_Gen          avgt   15     0.004 ±  0.001  MB/sec
ProducerRequestBenchmark.constructorProduceRequest:·gc.churn.G1_Old_Gen.norm     avgt   15     0.002 ±  0.001    B/op
ProducerRequestBenchmark.constructorProduceRequest:·gc.count                     avgt   15   494.000           counts
ProducerRequestBenchmark.constructorProduceRequest:·gc.time                      avgt   15   264.000               ms
ProducerRequestBenchmark.constructorStruct                                       avgt   15    90.728 ±  0.695   ns/op
ProducerRequestBenchmark.constructorStruct:·gc.alloc.rate                        avgt   15  3043.140 ± 23.246  MB/sec
ProducerRequestBenchmark.constructorStruct:·gc.alloc.rate.norm                   avgt   15   304.000 ±  0.001    B/op
ProducerRequestBenchmark.constructorStruct:·gc.churn.G1_Eden_Space               avgt   15  3047.251 ± 59.638  MB/sec
ProducerRequestBenchmark.constructorStruct:·gc.churn.G1_Eden_Space.norm          avgt   15   304.404 ±  5.034    B/op
ProducerRequestBenchmark.constructorStruct:·gc.churn.G1_Old_Gen                  avgt   15     0.003 ±  0.001  MB/sec
ProducerRequestBenchmark.constructorStruct:·gc.churn.G1_Old_Gen.norm             avgt   15    ≈ 10⁻⁴             B/op
ProducerRequestBenchmark.constructorStruct:·gc.count                             avgt   15   400.000           counts
ProducerRequestBenchmark.constructorStruct:·gc.time                              avgt   15   205.000               ms
```
### JMH for ProduceResponse

1. construction regression:
    - 3.293 -> 303.226 ns/op
    - 24.000 -> 1848.000 B/op
1. toStruct improvement:
    - 825.889 -> 311.725 ns/op
    - 2208.000 -> 896.000 B/op

**BEFORE**

```
Benchmark                                                                          Mode  Cnt     Score    Error   Units
ProducerResponseBenchmark.constructorProduceResponse                               avgt   15     3.293 ±  0.004   ns/op
ProducerResponseBenchmark.constructorProduceResponse:·gc.alloc.rate                avgt   15  6619.731 ±  9.075  MB/sec
ProducerResponseBenchmark.constructorProduceResponse:·gc.alloc.rate.norm           avgt   15    24.000 ±  0.001    B/op
ProducerResponseBenchmark.constructorProduceResponse:·gc.churn.G1_Eden_Space       avgt   15  6618.648 ±  0.153  MB/sec
ProducerResponseBenchmark.constructorProduceResponse:·gc.churn.G1_Eden_Space.norm  avgt   15    23.996 ±  0.033    B/op
ProducerResponseBenchmark.constructorProduceResponse:·gc.churn.G1_Old_Gen          avgt   15     0.003 ±  0.002  MB/sec
ProducerResponseBenchmark.constructorProduceResponse:·gc.churn.G1_Old_Gen.norm     avgt   15    ≈ 10⁻⁵             B/op
ProducerResponseBenchmark.constructorProduceResponse:·gc.count                     avgt   15   720.000           counts
ProducerResponseBenchmark.constructorProduceResponse:·gc.time                      avgt   15   383.000               ms
ProducerResponseBenchmark.constructorStruct                                        avgt   15   825.889 ±  0.638   ns/op
ProducerResponseBenchmark.constructorStruct:·gc.alloc.rate                         avgt   15  2428.000 ±  1.899  MB/sec
ProducerResponseBenchmark.constructorStruct:·gc.alloc.rate.norm                    avgt   15  2208.000 ±  0.001    B/op
ProducerResponseBenchmark.constructorStruct:·gc.churn.G1_Eden_Space                avgt   15  2430.196 ± 55.894  MB/sec
ProducerResponseBenchmark.constructorStruct:·gc.churn.G1_Eden_Space.norm           avgt   15  2210.001 ± 51.009    B/op
ProducerResponseBenchmark.constructorStruct:·gc.churn.G1_Old_Gen                   avgt   15     0.003 ±  0.001  MB/sec
ProducerResponseBenchmark.constructorStruct:·gc.churn.G1_Old_Gen.norm              avgt   15     0.002 ±  0.001    B/op
ProducerResponseBenchmark.constructorStruct:·gc.count                              avgt   15   319.000           counts
ProducerResponseBenchmark.constructorStruct:·gc.time                               avgt   15   166.000               ms
```

**AFTER**

```
Benchmark                                                                          Mode  Cnt     Score    Error   Units
ProducerResponseBenchmark.constructorProduceResponse                               avgt   15   303.226 ±  0.517   ns/op
ProducerResponseBenchmark.constructorProduceResponse:·gc.alloc.rate                avgt   15  5534.940 ±  9.439  MB/sec
ProducerResponseBenchmark.constructorProduceResponse:·gc.alloc.rate.norm           avgt   15  1848.000 ±  0.001    B/op
ProducerResponseBenchmark.constructorProduceResponse:·gc.churn.G1_Eden_Space       avgt   15  5534.046 ± 51.849  MB/sec
ProducerResponseBenchmark.constructorProduceResponse:·gc.churn.G1_Eden_Space.norm  avgt   15  1847.710 ± 18.105    B/op
ProducerResponseBenchmark.constructorProduceResponse:·gc.churn.G1_Old_Gen          avgt   15     0.007 ±  0.001  MB/sec
ProducerResponseBenchmark.constructorProduceResponse:·gc.churn.G1_Old_Gen.norm     avgt   15     0.002 ±  0.001    B/op
ProducerResponseBenchmark.constructorProduceResponse:·gc.count                     avgt   15   602.000           counts
ProducerResponseBenchmark.constructorProduceResponse:·gc.time                      avgt   15   318.000               ms
ProducerResponseBenchmark.constructorStruct                                        avgt   15   311.725 ±  3.132   ns/op
ProducerResponseBenchmark.constructorStruct:·gc.alloc.rate                         avgt   15  2610.602 ± 25.964  MB/sec
ProducerResponseBenchmark.constructorStruct:·gc.alloc.rate.norm                    avgt   15   896.000 ±  0.001    B/op
ProducerResponseBenchmark.constructorStruct:·gc.churn.G1_Eden_Space                avgt   15  2613.021 ± 42.965  MB/sec
ProducerResponseBenchmark.constructorStruct:·gc.churn.G1_Eden_Space.norm           avgt   15   896.824 ± 11.331    B/op
ProducerResponseBenchmark.constructorStruct:·gc.churn.G1_Old_Gen                   avgt   15     0.003 ±  0.001  MB/sec
ProducerResponseBenchmark.constructorStruct:·gc.churn.G1_Old_Gen.norm              avgt   15     0.001 ±  0.001    B/op
ProducerResponseBenchmark.constructorStruct:·gc.count                              avgt   15   343.000           counts
ProducerResponseBenchmark.constructorStruct:·gc.time                               avgt   15   194.000               ms
```


chia7712:
@hachikuji Could you please take a look?

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
```diff
@@ -210,65 +142,42 @@ public String toString() {
...
-        MutableRecordBatch entry = iterator.next();
-        hasIdempotentRecords = hasIdempotentRecords || entry.hasProducerId();
-        hasTransactionalRecords = hasTransactionalRecords || entry.isTransactional();
+        this.data = produceRequestData;
+        this.data.topicData().forEach(topicProduceData -> topicProduceData.partitions()
+            .forEach(partitionProduceData -> {
+                MemoryRecords records = MemoryRecords.readableRecords(partitionProduceData.records());
+                Iterator<MutableRecordBatch> iterator = records.batches().iterator();
+                MutableRecordBatch entry = iterator.next();
+                hasIdempotentRecords = hasIdempotentRecords || entry.hasProducerId();
```
hachikuji: Would it make sense to move this to the builder where we are already doing a pass over the partitions?

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
```diff
@@ -17,179 +17,48 @@
...
-    private static final String RESPONSES_KEY_NAME = "responses";
-
-    // topic level field names
-    private static final String PARTITION_RESPONSES_KEY_NAME = "partition_responses";
-
     public static final long INVALID_OFFSET = -1L;
+    private final ProduceResponseData data;
+    private final Map<TopicPartition, PartitionResponse> responses;
 
-    /**
```
hachikuji, line 54: Kind of a pity to lose this. Can we move it to the class documentation?

clients/src/main/resources/common/message/ProduceRequest.json
```diff
@@ -33,21 +33,21 @@
   "validVersions": "0-8",
   "flexibleVersions": "none",
   "fields": [
-    { "name": "TransactionalId", "type": "string", "versions": "3+", "nullableVersions": "0+", "entityType": "transactionalId",
+    { "name": "TransactionalId", "type": "string", "versions": "3+", "nullableVersions": "3+", "ignorable": true, "entityType": "transactionalId",
```
hachikuji: Hmm, not sure about making this ignorable. For transactional data, I think the broker would just fail if it cannot authorize the transactionalId.

Also, should we set the default to null?

clients/src/main/resources/common/message/ProduceRequest.json
```diff
@@ -33,21 +33,21 @@
...
       "about": "The transactional ID, or null if the producer is not transactional." },
     { "name": "Acks", "type": "int16", "versions": "0+",
       "about": "The number of acknowledgments the producer requires the leader to have received before considering a request complete. Allowed values: 0 for no acknowledgments, 1 for only the leader and -1 for the full ISR." },
-    { "name": "TimeoutMs", "type": "int32", "versions": "0+",
+    { "name": "Timeout", "type": "int32", "versions": "0+",
```
hachikuji: nit: I kind of liked the original name to make the unit clear. We're probably not consistent on this convention though.

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
```diff
@@ -218,6 +218,7 @@ private void checkSimpleRequestResponse(NetworkClient networkClient) {
                 request.apiKey().responseHeaderVersion(PRODUCE.latestVersion()));
         Struct resp = new Struct(PRODUCE.responseSchema(PRODUCE.latestVersion()));
         resp.set("responses", new Object[0]);
+        resp.set(CommonFields.THROTTLE_TIME_MS, 100);
```
hachikuji: Wondering if we may as well rewrite this using `ProduceResponseData`.

hachikuji (commented):
Thanks for the patch. Left a few small comments.

hachikuji:
@chia7712 One thing that would be useful is running the producer-performance test, just to make sure the the performance is inline. Might be worth checking flame graphs as well.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
hachikuji: Nevermind, I guess we have to do it here because the server needs to validate the request received from the client.

hachikuji: On the other hand, we might want to move this logic into a helper in `KafkaApis` so that these objects are dedicated only to serialization logic. Eventually we'll want to get rid of `ProduceRequest` and just use `ProduceRequestData`.

chia7712: clients module has some tests which depends on it so I moves the helper to ```RequestUtils```.

clients/src/main/resources/common/message/ProduceRequest.json
chia7712: you are right. will Roger that.

clients/src/main/resources/common/message/ProduceRequest.json
chia7712: This change is for compatibility. The name used by origin serialization is "timeout" rather than "timeoutMs" (see https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java#L54).

chia7712:
> One thing that would be useful is running the producer-performance test, just to make sure the the performance is inline. Might be worth checking flame graphs as well.

The cost of conversion is tiny in whole IO path. It seems to me JMH is more suitable for this patch. There are two new JMH for request and response. It benchmark the construction, toStruct and hot method. The result is attached to "description". @hachikuji Please take a look, thanks!

lbradstreet:
@chia7712 @hachikuji for the ProduceResponse handling, is this the overall broker side regression since you need both the construction and toStruct?
```
construction regression: 3.293 -> 580.099 ns/op
toStruct improvement: 825.889 -> 318.530 ns/op

overall response: 3.293+825.889 (old) = 829.182 vs 898.629 (new)
```

Could you please also provide an analysis of the garbage generation using `gc.alloc.rate.norm`?

chia7712:
@lbradstreet Thanks for your response.

>  is this the overall broker side regression since you need both the construction and toStruct?

you are right and it seems to me the solution to fix regression is that server should use automatic protocol response rather than wrapped response. However, it may make a big patch so it would be better to address in another PR. (BTW, fetch protocol has similar issue https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java#L281)

> Could you please also provide an analysis of the garbage generation using gc.alloc.rate.norm?

construction regression:
   - 3.293 -> 580.099 ns/op
   - 24.000 -> 2776.000 B/op

toStruct improvement:
  - 825.889 -> 318.530 ns/op
  - 2208.000 -> 896.000 B/op

We can reduce the regression (in construction) by replacing steam APIs by for-loop. However, I prefer stream Apis since it is more readable and the true solution is to use auto-generated protocols on server-side.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
```diff
@@ -560,13 +561,23 @@ private void handleProduceResponse(ClientResponse response, Map<TopicPartition,
             log.trace("Received produce response from node {} with correlation id {}", response.destination(), correlationId);
             // if we have a response, parse it
             if (response.hasResponse()) {
+                // TODO: Sender should exercise PartitionProduceResponse rather than ProduceResponse.PartitionResponse
```
hachikuji: Is the plan to save this for a follow-up? It looks like it will be a bit of effort to trace down all the uses, but seems doable.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
```diff
@@ -194,7 +107,27 @@ private ProduceRequest build(short version, boolean validate) {
                     ProduceRequest.validateRecords(version, records);
                 }
             }
-            return new ProduceRequest(version, acks, timeout, partitionRecords, transactionalId);
+
+            List<ProduceRequestData.TopicProduceData> tpd = partitionRecords
```
hachikuji: I wonder if we would get any benefit computing `partitionSizes` during this pass.

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
```diff
@@ -204,118 +75,78 @@ public ProduceResponse(Map<TopicPartition, PartitionResponse> responses) {
...
     public ProduceResponse(Map<TopicPartition, PartitionResponse> responses, int throttleTimeMs) {
-        this.responses = responses;
-        this.throttleTimeMs = throttleTimeMs;
+        this(new ProduceResponseData()
+            .setResponses(responses.entrySet()
+                .stream()
+                .collect(Collectors.groupingBy(e -> e.getKey().topic()))
+                .entrySet()
+                .stream()
+                .map(topicData -> new ProduceResponseData.TopicProduceResponse()
```
hachikuji: Not required, but this would be easier to follow up if we had some helpers.

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
```diff
@@ -204,118 +75,78 @@ public ProduceResponse(Map<TopicPartition, PartitionResponse> responses) {
...
+    }
 
-        return struct;
+    public ProduceResponseData data() {
+        return this.data;
     }
 
+    /**
+     * this method is used by testing only.
+     * TODO: refactor the tests which are using this method and then remove this method from production code.
```
hachikuji: Sounds good to refactor. Perhaps we can turn this TODO into a jira?

clients/src/main/resources/common/message/ProduceRequest.json
hachikuji: The names do not get serialized, so I think we can make them whatever we want.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -517,19 +517,23 @@ class KafkaApis(val requestChannel: RequestChannel,
...
 
-    val produceRecords = produceRequest.partitionRecordsOrFail.asScala
     val unauthorizedTopicResponses = mutable.Map[TopicPartition, PartitionResponse]()
     val nonExistingTopicResponses = mutable.Map[TopicPartition, PartitionResponse]()
     val invalidRequestResponses = mutable.Map[TopicPartition, PartitionResponse]()
     val authorizedRequestInfo = mutable.Map[TopicPartition, MemoryRecords]()
-    val authorizedTopics = filterByAuthorized(request.context, WRITE, TOPIC, produceRecords)(_._1.topic)
-
-    for ((topicPartition, memoryRecords) <- produceRecords) {
+    val authorizedTopics = filterByAuthorized(request.context, WRITE, TOPIC, produceRequest.dataOrException().topicData().asScala)(_.topic())
```
hachikuji: This logic surprised me a little bit until I realized that we were trying to avoid redundant authorization calls. Might be worth adding a comment since I was almost ready to suggest moving this logic into the loop.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -517,19 +517,23 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+
+    produceRequest.dataOrException().topicData().forEach(topic => topic.data().forEach { partition =>
+      val topicPartition = new TopicPartition(topic.topic(), partition.partition())
+      // This caller assumes the type is MemoryRecords and that is true on current serialization
+      // We cast the type to avoid causing big change to code base.
+      // TODO: maybe we need to refactor code to avoid this casting
```
hachikuji: That's a good question. I can't think of any great options. We ended up making `FetchResponse` generic to address a similar issue. I think the cast is reasonable for now. Can we move the TODO to a jira?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -517,19 +517,23 @@ class KafkaApis(val requestChannel: RequestChannel,
...
     val nonExistingTopicResponses = mutable.Map[TopicPartition, PartitionResponse]()
     val invalidRequestResponses = mutable.Map[TopicPartition, PartitionResponse]()
     val authorizedRequestInfo = mutable.Map[TopicPartition, MemoryRecords]()
-    val authorizedTopics = filterByAuthorized(request.context, WRITE, TOPIC, produceRecords)(_._1.topic)
-
-    for ((topicPartition, memoryRecords) <- produceRecords) {
+    val authorizedTopics = filterByAuthorized(request.context, WRITE, TOPIC, produceRequest.dataOrException().topicData().asScala)(_.topic())
+
+    produceRequest.dataOrException().topicData().forEach(topic => topic.data().forEach { partition =>
+      val topicPartition = new TopicPartition(topic.topic(), partition.partition())
```
hachikuji: nit: unnecessary parenthesis (a few of these around here)

hachikuji (commented):
This is looking pretty good. Left a few more comments.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
hachikuji: We could probably also compute `partitionSizes` lazily. I think the broker is the only one that uses it.

clients/src/main/resources/common/message/ProduceRequest.json
chia7712: This is a major question when I am processing those protocol migration PRs.

The Struct deserialization depends on the "name" (https://github.com/apache/kafka/blob/8e211eb72f9a45897cc37fed394a38096aa47feb/clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java#L250). Struct.get(xxx) searches the index according to "name". Hence, changing the name breaks compatibility since the Struct serialization can not find the correct field. (That is to say, the previous protocol fails to deserialize the data from new auto-generated protocol)

Please correct me if I misunderstood protocol mechanism.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
chia7712: The "data" may be cleared to null (for GC) so it is a risk if we lazily initialize the partition size.

I considered to remove it as it is used by log only (toString).

hachikuji: Right. I thought we might consider computing it just before the data gets cleared.

clients/src/main/resources/common/message/ProduceRequest.json
chia7712: oh, I trace the code again. You are right. The name is not serialized.

I will revise the field names as you suggested.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
chia7712: https://issues.apache.org/jira/browse/KAFKA-10696

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
chia7712: https://issues.apache.org/jira/browse/KAFKA-10697

core/src/main/scala/kafka/server/KafkaApis.scala
chia7712: https://issues.apache.org/jira/browse/KAFKA-10698

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
chia7712: Pardon me. why it is not required?

ijuma:
Can we summarize the regression here for a real world workload?

chia7712:
> Can we summarize the regression here for a real world workload?

@ijuma I have attached benchmark result to description. I will loop more benchmark later.

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
hachikuji: Oh, I was just emphasizing that it is a matter of taste. It's up to you if you agree or not.

hachikuji:
For what it's worth, I think we'll get back whatever we lose here by taking `Struct` out of the serialization path.

chia7712:
@hachikuji @ijuma @lbradstreet Could you take a look? There are some follow-ups which can get back the performance we lose here and I'd like to work on them as soon as possible :)

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
```diff
@@ -194,7 +106,27 @@ private ProduceRequest build(short version, boolean validate) {
...
```
dajac: I wonder if we could avoid all of this by requesting the `Sender` to create `TopicProduceData` directly. It seems that the `Sender` creates `partitionRecords` right before calling the builder: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java#L734. So we may be able to directly construct the expect data structure there.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
```diff
@@ -323,27 +222,30 @@ public String toString(boolean verbose) {
...
+                        .setRecordErrors(Collections.emptyList())
+                        .setBaseOffset(INVALID_OFFSET)
+                        .setLogAppendTimeMs(RecordBatch.NO_TIMESTAMP)
+                        .setLogStartOffset(INVALID_OFFSET)
+                        .setErrorMessage(e.getMessage())
+                        .setErrorCode(error.code()))
+                        .collect(Collectors.toList()))
+                    .setName(entry.getKey()))
+                .collect(Collectors.toList()))
+            .setThrottleTimeMs(throttleTimeMs));
```
dajac: It seems that we could create `ProduceResponseData` based on `data`. This avoids the cost of the group-by operation and the cost of constructing `partitionSizes`. That should bring the benchmark inline with what we had before. Would this work?

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
```diff
@@ -204,118 +75,79 @@ public ProduceResponse(Map<TopicPartition, PartitionResponse> responses) {
      * @param throttleTimeMs Time in milliseconds the response was throttled
      */
     public ProduceResponse(Map<TopicPartition, PartitionResponse> responses, int throttleTimeMs) {
-        this.responses = responses;
-        this.throttleTimeMs = throttleTimeMs;
+        this(new ProduceResponseData()
+            .setResponses(responses.entrySet()
```
dajac: As we care of performances here, I wonder if we should try not using the stream api here.

Another trick would be to  turn `TopicProduceResponse` in the `ProduceResponse` schema into a map by setting `"mapKey": true` for the topic name. This would allow to iterate over `responses`, get or create `TopicProduceResponse` for the topic, and add the `PartitionProduceResponse` into it.

It may be worth trying different implementation to compare their performances.

dajac (commented):
@chia7712 Thanks for the PR. I have left few suggestions.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
chia7712: nice suggestion.

Could I address this in follow-up? I had filed jira (KAFKA-10696 ~ KAFKA-10698) to have ```Sender``` use auto-generated protocol directly.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
chia7712: I used  ```data``` to generate ProduceResponseData. However, the ```data``` may be null when create ProduceResponseData. That is to say, it require if-else to handle null ```data``` in ```getErrorResponse```. It seems to me that is a bit ugly so not sure whether it is worth doing that.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
chia7712: oh. The jira I created does not cover this issue. open a new ticket (https://issues.apache.org/jira/browse/KAFKA-10709)

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
chia7712: > It may be worth trying different implementation to compare their performances.

As we all care for performance, I'm ok to say goodbye to stream api :)

chia7712: I have addressed your suggestion and it does improve the performance.

chia7712:
The last commit borrows some improvement from #9563.

chia7712:
@hachikuji @ijuma @lbradstreet @dajac I have updated the perf result. The regression is reduced by last commit. Please take a look.

hachikuji:
Posting allocation flame graphs from the producer before and after this patch:

Screen Shot 2020-11-16 at 5 26 56 PM
Screen Shot 2020-11-16 at 5 27 25 PM

So we succeeded in getting rid of the extra allocations in the network layer!

I generated these graphs using the producer performance test writing to a topic with 10 partitions on a cluster with a single broker.
```
> bin/kafka-producer-perf-test.sh --topic foo --num-records 250000000 --throughput -1  --record-size 256 --producer-props bootstrap.servers=localhost:9092
```

ijuma:
Nice! So what's the reason for the small regression in the PR description?

hachikuji:
Yeah, there is something strange going on, especially in regard to latency. Running the same producer performance test, I saw the following:
```
Patch:
250000000 records sent, 1347222.297068 records/sec (328.91 MB/sec), 91.98 ms avg latency, 1490.00 ms max latency, 71 ms 50th, 242 ms 95th, 320 ms 99th, 728 ms 99.9th.                                             

Trunk:
250000000 records sent, 1426264.954388 records/sec (348.21 MB/sec), 15.11 ms avg latency, 348.00 ms max latency, 3 ms 50th, 94 ms 95th, 179 ms 99th, 265 ms 99.9th.
```
I was able to reproduce similar results several times. Take this with a grain of salt, but from the flame graphs, I see the following differences:

`RequestContext.parseRequest`: 1% -> 0.45%
`RequestUtils.hasTransactionalRecords`: 0% -> 0.59%
`RequestUtils.hasIdempotentRecords`: 0% -> 0.14%
`KafkaApis.sendResponseCallback`: 3.20% -> 2.33%
`KafkaApis.clearPartitionRecords`: 0% -> 0.16%

I think `hasTransactionalRecords` and `hasIdempotentRecords` are the most obvious optimization targets (they also show up in allocations), but I do not think they explain the increase in latency. Just to be sure, I commented out these lines and I got similar results.

ijuma:
That is really weird. The difference seems significant enough that we need to understand it better before we can merge IMO.

hachikuji:
It would be helpful if someone can reproduce the tests I did to make sure it is not something funky in my environment.

hachikuji:
I think the large difference in latency in my test is due to the producer's buffer pool getting exhausted. I was looking at the "bufferpool-wait-ratio" metric exposed in the producer. With this patch, it was hovering around 0.6 while on trunk it remained around 0.01. I'll need to lower the throughput a little bit in order to get a better estimate of the regression.

chia7712:
@hachikuji @ijuma Thanks for all feedback. I'm going to do more tests :)

hachikuji:
@chia7712 It might be worth checking the fancy new `toSend` implementation. I did a quick test and found that gc overhead actually increased with this change even though the new implementation seemed much better for cpu.

chia7712:
@hachikuji @ijuma It seems to me the root cause of regression is caused by the conversion between old data struct and auto-generated protocol. As we all care for performance, I address more improvements (they were viewed as follow-up before) in this PR. 

There are 5 possible improvements.

1. - [x]  git rid of struct allocation (thanks to @hachikuji)
1. - [x]  rewrite RequestUtils.hasTransactionalRecords & RequestUtils.hasIdempotentRecords to avoid duplicate loop and record instantiation (thanks to @hachikuji)
1. - [x] Sender#sendProduceRequest should use auto-generated protocol directly (thanks to @dajac)
1. - [ ] [client side] Replace ProduceResponse.PartitionResponse by auto-generated PartitionProduceResponse (https://issues.apache.org/jira/browse/KAFKA-10709)
1. - [ ] [server side] KafkaApis#handleProduceRequest should use auto-generated protocol (https://issues.apache.org/jira/browse/KAFKA-10730)

The recent commits include only 1, 2, and 3 since they touch fewer code. I also re-run all perf tests:

1. currently all tests cases show better throughput with current PR even though 4. and 5. are not included.
1. it seems to me they can get better improvement by remaining tasks (4 and 5)

More perf tests are running and I will update the PR description later.

ijuma:
Thanks @chia7712. Can you check that the @hachikuji's perf test doesn't result in significantly increased latency?

chia7712:
> Can you check that the @hachikuji's perf test doesn't result in significantly increased latency?

I don't observer significantly increased latency on my local. The env is shown below.

- INTEL 10900K 10C20T 3.70 GHz ~ 5.30 GHz
- DDR4-2933 128 GB
- SSD
- Ubuntu 20.04
- OpenJDK 11.0.9.1

**trunk** (53026b799c609373696382c685b9080630128af2)
> 250000000 records sent, 1143578.577571 records/sec (279.19 MB/sec), 109.01 ms avg latency, 422.00 ms max latency, 107 ms 50th, 132 ms 95th, 146 ms 99th, 263 ms 99.9th.

**patch**
> 250000000 records sent, 1202958.315088 records/sec (293.69 MB/sec), 103.60 ms avg latency, 432.00 ms max latency, 101 ms 50th, 122 ms 95th, 146 ms 99th, 278 ms 99.9th.

hachikuji:
@chia7712 Thanks for the updates, looks promising. I'm more inclined to trust your results on linux. I have been testing on macos.    Let me see if I can reproduce in a better environment and then we're probably good to go.

hachikuji:
Here are a couple additional test runs. This was on Ubuntu 20 (ami-00831fc7c1e3ddc60). The machine type was m5a.xlarge with 200GB gp2 EBS storage. One instance was running the broker and one instance was running the producer perf test.

Commands:
```
bin/kafka-topics.sh --create --topic foo --replication-factor 1 --partitions 10 --bootstrap-server $BROKER
bin/kafka-producer-perf-test.sh --topic foo --num-records 250000000 --throughput -1  --record-size 256 --producer-props bootstrap.servers=$BROKER
```

Here are the results:
```
Patch:
250000000 records sent, 826003.925171 records/sec (201.66 MB/sec), 149.39 ms avg latency, 1623.00 ms max latency, 131 ms 50th, 380 ms 95th, 464 ms 99th, 650 ms 99.9th.
250000000 records sent, 825684.740355 records/sec (201.58 MB/sec), 149.22 ms avg latency, 1276.00 ms max latency, 124 ms 50th, 364 ms 95th, 451 ms 99th, 775 ms 99.9th.                                            

Trunk:
250000000 records sent, 833144.487250 records/sec (203.40 MB/sec), 148.20 ms avg latency, 1361.00 ms max latency, 111 ms 50th, 437 ms 95th, 551 ms 99th, 807 ms 99.9th.
250000000 records sent, 810927.409022 records/sec (197.98 MB/sec), 152.59 ms avg latency, 1430.00 ms max latency, 127 ms 50th, 382 ms 95th, 467 ms 99th, 809 ms 99.9th.                                            
```

Given variance in these tests, I think we're probably inline with trunk. I looked at the flame graph as well and did not observe any substantial difference in performance. Here are a few interesting highlights from one run. This patch is listed first with trunk second.

`Sender.sendProduceRequests`:

Screen Shot 2020-11-17 at 2 36 34 PM
Screen Shot 2020-11-17 at 2 36 49 PM

`KafkaChannel.write`:

Screen Shot 2020-11-17 at 2 32 56 PM
Screen Shot 2020-11-17 at 2 33 17 PM

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
```diff
@@ -560,13 +562,24 @@ private void handleProduceResponse(ClientResponse response, Map<TopicPartition,
...
+                // https://issues.apache.org/jira/browse/KAFKA-10696
```
hachikuji, line 566: nit: since we have the jira for tracking, can we remove the TODO? A few more of these in the PR.

clients/src/test/java/org/apache/kafka/common/requests/ProduceResponseTest.java
```diff
@@ -100,16 +113,92 @@ public void produceResponseRecordErrorsTest() {
...
+                assertNull(deserialized.errorMessage);
             }
         }
     }
+
+    /**
+     * the schema in this test is from previous code and the automatic protocol should be compatible to previous schema.
+     */
+    @Test
+    public void testCompatibility() {
```
hachikuji: I think this test might be overkill. We haven't done anything like this for the other converted APIs. It's a little similar to `MessageTest.testRequestSchemas`, which was useful verifying the generated schemas when the message generator was being written. Soon `testRequestSchemas` will be redundant, so I guess we have to decide if we just trust the generator and our compatibility system tests or if we want some other canonical representation. Thoughts?

hachikuji (approved):
LGTM. Left a couple comments, but I'm happy with the current PR. Would be helpful if @dajac could take one more look before we merge.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
chia7712: copy that

clients/src/test/java/org/apache/kafka/common/requests/ProduceResponseTest.java
chia7712: That makes sense to me. Will remove redundant test.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
```diff
@@ -560,13 +562,24 @@ private void handleProduceResponse(ClientResponse response, Map<TopicPartition,
...
+                    TopicPartition tp = new TopicPartition(r.name(), p.index());
+                    ProduceResponse.PartitionResponse partResp = new ProduceResponse.PartitionResponse(
+                            Errors.forCode(p.errorCode()),
+                            p.baseOffset(),
+                            p.logAppendTimeMs(),
+                            p.logStartOffset(),
+                            p.recordErrors()
+                                .stream()
+                                .map(e -> new ProduceResponse.RecordError(e.batchIndex(), e.batchIndexErrorMessage()))
+                                .collect(Collectors.toList()),
```
dajac, line 578: nit: As we got rid of the streaming api in this section, would it make sense to also remove this one?

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
```diff
@@ -203,119 +77,88 @@ public ProduceResponse(Map<TopicPartition, PartitionResponse> responses) {
...
+                    .setLogStartOffset(response.logStartOffset)
+                    .setLogAppendTimeMs(response.logAppendTime)
+                    .setErrorMessage(response.errorMessage)
+                    .setErrorCode(response.error.code())
+                    .setRecordErrors(response.recordErrors
+                        .stream()
+                        .map(e -> new ProduceResponseData.BatchIndexAndErrorMessage()
+                            .setBatchIndex(e.batchIndex)
+                            .setBatchIndexErrorMessage(e.message))
+                        .collect(Collectors.toList())));
```
dajac, line 111: nit: Should we remove this usage of the stream api here as well?

clients/src/main/java/org/apache/kafka/common/requests/RequestUtils.java
```diff
@@ -48,4 +53,41 @@ public static ByteBuffer serialize(Struct headerStruct, Struct bodyStruct) {
...
+                    }
+                }
+                // return early
+                if (hasIdempotentRecords && hasTransactionalRecords)
+                    return new AbstractMap.SimpleEntry<>(true, true);
+            }
+        }
+        return new AbstractMap.SimpleEntry<>(hasIdempotentRecords, hasTransactionalRecords);
+    }
+}
```
dajac: nit: Add a new line.

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
```diff
@@ -154,8 +155,11 @@ public void testClose() {
         client.poll(1, time.milliseconds());
         assertTrue("The client should be ready", client.isReady(node, time.milliseconds()));
 
-        ProduceRequest.Builder builder = ProduceRequest.Builder.forCurrentMagic((short) 1, 1000,
-                Collections.emptyMap());
+        ProduceRequest.Builder builder = ProduceRequest.forCurrentMagic(new ProduceRequestData()
+                .setTopicData(new ProduceRequestData.TopicProduceDataCollection())
+                .setAcks((short) 1)
+                .setTimeoutMs(1000)
+                .setTransactionalId(null));
```
dajac: nit: It seem that `TransactionalId` is `null` by default so we don't have to set it explicitly all the time. There are few cases in the this file and in others. I am not sure if this was intentional so I am also fine if you want to keep them.

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
```diff
@@ -198,8 +202,9 @@ private void checkSimpleRequestResponse(NetworkClient networkClient) {
         ResponseHeader respHeader =
             new ResponseHeader(request.correlationId(),
                 request.apiKey().responseHeaderVersion(PRODUCE.latestVersion()));
-        Struct resp = new Struct(PRODUCE.responseSchema(PRODUCE.latestVersion()));
-        resp.set("responses", new Object[0]);
+        Struct resp = new ProduceResponseData()
+                .setThrottleTimeMs(100)
+                .toStruct(ProduceResponseData.HIGHEST_SUPPORTED_VERSION);
```
dajac: nit: It may be better to use `PRODUCE.latestVersion()` to stay inline with L204. There are few cases like this in the file.

clients/src/test/java/org/apache/kafka/common/requests/ProduceRequestTest.java
```diff
@@ -192,16 +258,21 @@ public void testMixedTransactionalData() {
...
-                recordsByPartition, transactionalId);
+        ProduceRequest.Builder builder = ProduceRequest.forMagic(RecordBatch.CURRENT_MAGIC_VALUE,
+                new ProduceRequestData()
+                        .setTopicData(new ProduceRequestData.TopicProduceDataCollection(Arrays.asList(
+                                new ProduceRequestData.TopicProduceData().setName("foo").setPartitionData(Collections.singletonList(
+                                        new ProduceRequestData.PartitionProduceData().setIndex(0).setRecords(txnRecords))),
+                                new ProduceRequestData.TopicProduceData().setName("foo").setPartitionData(Collections.singletonList(
+                                        new ProduceRequestData.PartitionProduceData().setIndex(1).setRecords(nonTxnRecords))))
+                                .iterator()))
+                        .setAcks((short) 1)
```
dajac: nit: I am not sure if this is important or not but we were using `-1` previously.

clients/src/test/java/org/apache/kafka/common/requests/ProduceRequestTest.java
```diff
@@ -192,16 +258,21 @@ public void testMixedTransactionalData() {
...
+                        .setTimeoutMs(5000)
+                        .setTransactionalId(null));
```
dajac: We were setting `transactionalId` previously.

clients/src/main/resources/common/message/ProduceRequest.json
```diff
@@ -33,21 +33,21 @@
   "validVersions": "0-8",
   "flexibleVersions": "none",
   "fields": [
-    { "name": "TransactionalId", "type": "string", "versions": "3+", "nullableVersions": "0+", "entityType": "transactionalId",
+    { "name": "TransactionalId", "type": "string", "versions": "3+", "nullableVersions": "3+", "default": "null", "entityType": "transactionalId",
```
dajac, line 36: I wonder if this one should be `ignorable`. It seems that we were ignoring it before when it was not present in the target version:
```
struct.setIfExists(NULLABLE_TRANSACTIONAL_ID, transactionalId)
```

dajac (commented):
The PR looks great! I have left few minor comments.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
chia7712: The reason we got rid of streaming APIs is because it produces extra collection (groupBy). However, in this case we have to create a new collection to carry non-auto-generated data (and https://issues.apache.org/jira/browse/KAFKA-10696 will eliminate such conversion) even if we get rid of stream APIs. Hence, it should be fine to keep stream APIs here.

clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java
chia7712: ditto

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
chia7712: make sense to me. will remove this redundant setter

clients/src/main/resources/common/message/ProduceRequest.json
chia7712: you are right.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
dajac: Yeah, I do agree. It won't change much from a performance point of view. I was more thinking about this from a code consistency point of view. I don't feel strong about this at all. It is just that I usually prefer not to mix paradigms. I recognise that this is a personal taste :).

dajac (approved):
LGTM, pending Jenkins. Thanks for the PR!

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -1730,7 +1768,6 @@ private DeleteTopicsResponse createDeleteTopicsResponse() {
 
     private InitProducerIdRequest createInitPidRequest() {
         InitProducerIdRequestData requestData = new InitProducerIdRequestData()
-                .setTransactionalId(null)
```
dajac: We should keep this one.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -580,7 +580,6 @@ synchronized void bumpIdempotentEpochAndResetIdIfNeeded() {
             if (currentState != State.INITIALIZING && !hasProducerId()) {
                 transitionTo(State.INITIALIZING);
                 InitProducerIdRequestData requestData = new InitProducerIdRequestData()
-                        .setTransactionalId(null)
```
dajac: We should keep this one as well.

chia7712: Sorry that I did not test them on my local before pushing :(

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
chia7712: thanks!!!

clients/src/main/resources/common/message/ProduceRequest.json
hachikuji: Hmm I don't think it should be ignorable. The request would just fail if we drop it.

clients/src/main/resources/common/message/ProduceRequest.json
```diff
@@ -33,21 +33,21 @@
   "validVersions": "0-8",
   "flexibleVersions": "none",
   "fields": [
-    { "name": "TransactionalId", "type": "string", "versions": "3+", "nullableVersions": "0+", "entityType": "transactionalId",
+    { "name": "TransactionalId", "type": "string", "versions": "3+", "nullableVersions": "3+", "default": "null", "ignorable": true, "entityType": "transactionalId",
```
hachikuji: I don't think it should be ignorable. Transactional requests require this in order to authorize.

chia7712: It seems to me ignorable should be true in order to keep behavior consistency. With "ignore=false", setting value to ```TransactionalId``` can cause ```UnsupportedVersionException``` if the version is small than 3. The previous code (https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java#L286) does not cause such exception.

hachikuji: The previous code probably relied on the range checking of the message format to imply support here. My point is that the request is doomed to fail if it holds transactional data and we drop the transactionalId. So we may as well fail fast.

chia7712: > So we may as well fail fast.

That make sense. will revert this change.

hachikuji:
Great work @chia7712 ! With this and #9547, we have converted all of the protocols, which was a huge community effort!

clients/src/main/resources/common/message/ProduceRequest.json
dajac: Indeed, that makes sense. My bad!

PR #9406, opened 2020-10-09
KAFKA-10520; Ensure transactional producers poll if leastLoadedNode not available with max.in.flight=1

rajinisivaram:
We currently stop polling in `Sender` in a transactional producer if there is only one broker in the bootstrap server list and `max.in.flight.requests.per.connection=1` and Metadata response is pending when InitProducerId request is ready to be sent. In this scenario, we attempt to send FindCoordinator to `leastLoadedNode`, but since that is blocked due to `max.in.flight=1` as a result of the pending metadata response, we never unblock unless we poll. This PR ensures we poll in this case.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
@@ -444,9 +444,19 @@ public class Sender implements Runnable {
         AbstractRequest.Builder<?> requestBuilder = nextRequestHandler.requestBuilder();
         Node targetNode = null;
         try {
-            targetNode = awaitNodeReady(nextRequestHandler.coordinatorType());
-            if (targetNode == null) {
+            FindCoordinatorRequest.CoordinatorType coordinatorType = nextRequestHandler.coordinatorType();
+            targetNode = coordinatorType != null ?
+                    transactionManager.coordinator(coordinatorType) :
+                    client.leastLoadedNode(time.milliseconds());
+            if (targetNode != null) {
+                awaitNodeReady(targetNode, coordinatorType);
+            } else if (coordinatorType != null) {
+                log.trace("Coordinator not known for {}, will retry {} after finding coordinator.", coordinatorType, requestBuilder.apiKey());
                 maybeFindCoordinatorAndRetry(nextRequestHandler);
+                return true;
+            } else {
+                log.trace("No nodes available to send requests, polling until a node is ready.");
+                client.poll(retryBackoffMs, time.milliseconds());
                 return true;
             }
 
@@ -512,11 +522,7 @@ public class Sender implements Runnable {
         return running;
     }
 
-    private Node awaitNodeReady(FindCoordinatorRequest.CoordinatorType coordinatorType) throws IOException {
-        Node node = coordinatorType != null ?
-                transactionManager.coordinator(coordinatorType) :
-                client.leastLoadedNode(time.milliseconds());
-
+    private void awaitNodeReady(Node node, FindCoordinatorRequest.CoordinatorType coordinatorType) throws IOException {
         if (node != null && NetworkClientUtils.awaitReady(client, node, time, requestTimeoutMs)) {
             if (coordinatorType == FindCoordinatorRequest.CoordinatorType.TRANSACTION) {
                 // Indicate to the transaction manager that the coordinator is ready, allowing it to check ApiVersions
@@ -524,9 +530,7 @@ public class Sender implements Runnable {
                 // the time when the abortable error is handled
                 transactionManager.handleCoordinatorReady();
             }
-            return node;
         }
-        return null;
     }
 
     /**
--- a/clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
@@ -449,7 +449,11 @@ public class Sender implements Runnable {
                     transactionManager.coordinator(coordinatorType) :
                     client.leastLoadedNode(time.milliseconds());
             if (targetNode != null) {
-                awaitNodeReady(targetNode, coordinatorType);
+                if (!awaitNodeReady(targetNode, coordinatorType)) {
+                    log.trace("Target node {} not ready within request timeout, will retry when node is ready.", targetNode);
+                    maybeFindCoordinatorAndRetry(nextRequestHandler);
+                    return true;
+                }
             } else if (coordinatorType != null) {
                 log.trace("Coordinator not known for {}, will retry {} after finding coordinator.", coordinatorType, requestBuilder.apiKey());
                 maybeFindCoordinatorAndRetry(nextRequestHandler);
@@ -523,15 +527,17 @@ public class Sender implements Runnable {
         return running;
     }
 
-    private void awaitNodeReady(Node node, FindCoordinatorRequest.CoordinatorType coordinatorType) throws IOException {
-        if (node != null && NetworkClientUtils.awaitReady(client, node, time, requestTimeoutMs)) {
+    private boolean awaitNodeReady(Node node, FindCoordinatorRequest.CoordinatorType coordinatorType) throws IOException {
+        if (NetworkClientUtils.awaitReady(client, node, time, requestTimeoutMs)) {
             if (coordinatorType == FindCoordinatorRequest.CoordinatorType.TRANSACTION) {
                 // Indicate to the transaction manager that the coordinator is ready, allowing it to check ApiVersions
                 // This allows us to bump transactional epochs even if the coordinator is temporarily unavailable at
                 // the time when the abortable error is handled
                 transactionManager.handleCoordinatorReady();
             }
+            return true;
         }
+        return false;
     }
 
     /**
--- a/clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
@@ -455,7 +455,8 @@ public class Sender implements Runnable {
                 maybeFindCoordinatorAndRetry(nextRequestHandler);
                 return true;
             } else {
-                log.trace("No nodes available to send requests, polling until a node is ready.");
+                log.trace("No nodes available to send requests, will poll and retry when until a node is ready.");
+                transactionManager.retry(nextRequestHandler);
                 client.poll(retryBackoffMs, time.milliseconds());
                 return true;
             }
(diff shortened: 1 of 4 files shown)


clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
```diff
@@ -444,10 +444,20 @@ private boolean maybeSendAndPollTransactionalRequest() {
...
+            if (targetNode != null) {
+                awaitNodeReady(targetNode, coordinatorType);
+            } else if (coordinatorType != null) {
+                log.trace("Coordinator not known for {}, will retry {} after finding coordinator.", coordinatorType, requestBuilder.apiKey());
                 maybeFindCoordinatorAndRetry(nextRequestHandler);
                 return true;
+            } else {
+                log.trace("No nodes available to send requests, polling until a node is ready.");
+                client.poll(retryBackoffMs, time.milliseconds());
+                return true;
```
dajac: Don't we loose the current `nextRequestHandler` when we end up in this branch?

For instance, if `nextRequestHandler` is a `FindCoordinatorHandler` and there a no nodes available, `targetNode` is `null` and `coordinatorType` is `null` as well so we end up here and poll. We don't do anything with `nextRequestHandler` and return so it is gone. I suppose that it is not an issue for `FindCoordinatorHandler` as a new one will be enqueued automatically when another `TxnRequestHandler` handler is processed and the coordinator is unknown.

We may want to push back `nextRequestHandler` to the queue with `transactionManager.retry(nextRequestHandler)` in oder to handle all the cases.

I am not sure if that could really happen with any other `TxnRequestHandler` type though. What do you think?

rajinisivaram: @dajac Thanks for the review. As far as I can tell, `nextRequestHandler` returned by `transactionManager.nextRequest` would have retained the request in this case since it does a peek and removes the element only under certain conditions. I was trying to limit the amount of change to the minimal necessary to handle max.in.flight=1, do you think we should do more to look into the other cases?

dajac: @rajinisivaram It seems that a non-null `nextRequestHandler` would have been removed from the queue: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java#L872.

I've got the impression that `InitProducerIdHandler` may hit that code path as well as it could have a null `coordinatorType` when the transaction id is not set (idempotent producer). I think that we could unit test this actually by reusing your `testInitProducerIdWithMaxInFlightOne` but without setting the transactional id. Should we add this test to be on the safe side?

rajinisivaram: @dajac Sorry, that was my mistake. Updated code and added test.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
```diff
@@ -512,21 +523,15 @@ public boolean isRunning() {
         return running;
     }
 
-    private Node awaitNodeReady(FindCoordinatorRequest.CoordinatorType coordinatorType) throws IOException {
-        Node node = coordinatorType != null ?
-                transactionManager.coordinator(coordinatorType) :
-                client.leastLoadedNode(time.milliseconds());
-
+    private void awaitNodeReady(Node node, FindCoordinatorRequest.CoordinatorType coordinatorType) throws IOException {
         if (node != null && NetworkClientUtils.awaitReady(client, node, time, requestTimeoutMs)) {
```
chia7712: ```node != null``` Is this check necessary?

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
```diff
@@ -444,10 +444,21 @@ private boolean maybeSendAndPollTransactionalRequest() {
...
         Node targetNode = null;
         try {
-            targetNode = awaitNodeReady(nextRequestHandler.coordinatorType());
-            if (targetNode == null) {
+            FindCoordinatorRequest.CoordinatorType coordinatorType = nextRequestHandler.coordinatorType();
+            targetNode = coordinatorType != null ?
+                    transactionManager.coordinator(coordinatorType) :
+                    client.leastLoadedNode(time.milliseconds());
+            if (targetNode != null) {
+                awaitNodeReady(targetNode, coordinatorType);
```
chia7712: How to deal with the non-ready targetNode after calling ```awaitNodeReady``` ?

core/src/test/scala/integration/kafka/api/TransactionsWithMaxInFlightOneTest.scala
```diff
@@ -0,0 +1,131 @@
...
+
+import scala.collection.Seq
+import scala.collection.mutable.Buffer
+import scala.jdk.CollectionConverters._
+
+/**
+ * This is used to test transactions with one broker and `max.in.flight.requests.per.connection=1`.
+ * A single broker is used to verify edge cases where different requests are queued on the same connection.
+ */
+class TransactionsWithMaxInFlightOneTest extends KafkaServerTestHarness {
```
chia7712, line 39: this new test passes on my local even though I revert the change of ```Sender```.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
rajinisivaram: No, we don't need it anymore, removed.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
rajinisivaram: @chia7712 Thanks for the review. Good point, added check to retry in that case, also added unit test.

core/src/test/scala/integration/kafka/api/TransactionsWithMaxInFlightOneTest.scala
rajinisivaram: Yes, the issue this PR fixes is a timing issue that the new unit tests catch, but it is unlikely that we hit it in integration tests. I added this integration test anyway to make sure the fix hasn't broken anything. I was in two minds about whether to check it in, but included anyway since I thought it may help catch regressions with max.in.flight=1 in future. Also, Jenkins builds on slower machines may occasionally hit timing windows we don't usually see otherwise. Don't mind removing if we think it is not worth the maintenance cost.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
dajac: Thanks. I will make another pass now.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
```diff
@@ -444,10 +444,25 @@ private boolean maybeSendAndPollTransactionalRequest() {
...
+                }
+            } else if (coordinatorType != null) {
+                log.trace("Coordinator not known for {}, will retry {} after finding coordinator.", coordinatorType, requestBuilder.apiKey());
                 maybeFindCoordinatorAndRetry(nextRequestHandler);
                 return true;
+            } else {
+                log.trace("No nodes available to send requests, will poll and retry when until a node is ready.");
+                transactionManager.retry(nextRequestHandler);
+                client.poll(retryBackoffMs, time.milliseconds());
+                return true;
```
dajac, line 465: Looking at this branch again, sorry :). I was comparing it with the previous behavior and I have noticed that we would request a refresh of the metadata when the same conditions were met. That happened here:
```
    private void maybeFindCoordinatorAndRetry(TransactionManager.TxnRequestHandler nextRequestHandler) {
        if (nextRequestHandler.needsCoordinator()) {
            transactionManager.lookupCoordinator(nextRequestHandler);
        } else {
            // For non-coordinator requests, sleep here to prevent a tight loop when no node is available
            time.sleep(retryBackoffMs);
            metadata.requestUpdate();
        }

        transactionManager.retry(nextRequestHandler);
    }
```
When no node is available and `coordinatorType != null`, we ended up in the else branch here. I wonder if not doing `metadata.requestUpdate()` in our new implementation could be problematic and I also wonder if we could just swap that `time.sleep(retryBackoffMs)` by `client.poll(retryBackoffMs, time.milliseconds())` to achieve the same goal. The difference is that another metadata request would be sent in our particular case.

Have you noticed this small difference?

rajinisivaram: @dajac Yes, I was thinking about this earlier when adding the retry. We get to this `else` path when leastLoadedNode is null. For the other cases, we would have taken one of the other paths, which retain existing behaviour (unless I missed something, again!) When least loaded node is null, the old behaviour didn't quite work because you have to poll to change that state. And you don't have anywhere to send metadata requests to. So just polling seemed to be sufficient for this case. What do you think?

hachikuji: I agree polling seems sufficient. We will still have an opportunity to refresh metadata if the current connection fails for some reason.

clients/src/test/java/org/apache/kafka/clients/producer/internals/SenderTest.java
```diff
@@ -2667,4 +2760,43 @@ private void assertFutureFailure(Future<?> future, Class<? extends Exception> ex
         }
     }
 
+    private void createMockClientWithMaxFlightOneMetadataPending() {
+        client = new MockClient(time, metadata) {
```
hachikuji, line 2764: Wonder if we should consider adding max inflight behavior directly to `MockClient`. Seems like a notable difference from `NetworkClient`.

hachikuji (approved):
Thanks for the fix. LGTM.

clients/src/main/java/org/apache/kafka/clients/producer/internals/Sender.java
dajac: @rajinisivaram Yeah, I do agree. Polling seems sufficient in this case. Thanks for the clarification.

dajac (approved):
LGTM. Thanks for the fix.

hachikuji:
Merging to trunk and 2.7.

clients/src/test/java/org/apache/kafka/clients/producer/internals/SenderTest.java
rajinisivaram: @hachikuji Thanks for reviewing and merging! I have opened https://issues.apache.org/jira/browse/KAFKA-10626 to add max.in.flight to MockClient.

PR #9407, opened 2020-10-09
KAFKA-10611: Merge log error to avoid double error

bmaggi:
When using an error tracking system, 2 errors means 2 different alerts.
It's best to group the logs and have one error with all information.

For example when using with [Sentry](https://sentry.io/welcome/), this double line of log.error will create 2 different Issues.
One can merge the issues but it will be simpler to have a single error log line

I feel this is a trivial change, so following [Contribution rules](https://cwiki.apache.org/confluence/display/KAFKA/Contributing+Code+Changes#ContributingCodeChanges-PullRequest)
I didn't create a Jira ticket but simply open a pull request

If I'm wrong, please feel free to indicate me the correct process.

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerTask.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerTask.java
@@ -184,7 +184,7 @@ abstract class WorkerTask implements Runnable {
 
             execute();
         } catch (Throwable t) {
-            log.error("{} Task threw an uncaught and unrecoverable exception, task is being killed and will not recover until manually restarted", this, t);
+            log.error("{} Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted", this, t);
             throw t;
         } finally {
             doClose();
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerTask.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerTask.java
@@ -184,8 +184,7 @@ abstract class WorkerTask implements Runnable {
 
             execute();
         } catch (Throwable t) {
-            log.error("{} Task threw an uncaught and unrecoverable exception", this, t);
-            log.error("{} Task is being killed and will not recover until manually restarted", this);
+            log.error("{} Task threw an uncaught and unrecoverable exception, task is being killed and will not recover until manually restarted", this, t);
             throw t;
         } finally {
             doClose();


bmaggi:
🤔 not sure how failing tests are related to the patch 🤔

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerTask.java
```diff
@@ -184,8 +184,7 @@ private void doRun() throws InterruptedException {
 
             execute();
         } catch (Throwable t) {
-            log.error("{} Task threw an uncaught and unrecoverable exception", this, t);
-            log.error("{} Task is being killed and will not recover until manually restarted", this);
+            log.error("{} Task threw an uncaught and unrecoverable exception, task is being killed and will not recover until manually restarted", this, t);
```
kkonstantine: I'd suggest keeping the format slightly closer to what we had before. Specifically:

`Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted`

kkonstantine (commented):
Thanks @bmaggi

I agree with you. 
Yet this is probably the most characteristic error log in Connect. And it's been there for quite a while. To a certain degree, changing logs are user facing changes. 

I'm a bit on the fence as to whether this needs a KIP. I'm inclined to say no for now. 
However, I'd definitely recommend creating a jira issue and would consider merging only on the next minor release (or major release if that's the next one). Not in a patch release. 

What do you think about this change @rhauch @ewencp ?

ewencp (commented):
I'll defer to @kkonstantine on the details of format, but yeah, seems better to consolidate. Not sure why we had it as two separate log lines -- probably just an oversight in review and it seems to have come in with the status tracking, but doesn't seem to have been a reason to keep them separate.

The only thing I would request is that ensure at least the beginning of the message remains the same for filtering purposes. Logging is semi-public interface, so we have to take care when modifying it.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerTask.java
rhauch: I also like @kkonstantine's suggested format.

bmaggi:
Thanks for the comments.

Following suggestions: 
* I created a JIRA ticket [KAFKA-10611](https://issues.apache.org/jira/browse/KAFKA-10611) on the subject 
* I also added a new commit using the suggested text (closer to the old version)

kkonstantine (approved):
Thanks for addressing our comments @bmaggi 

LGTM

PR #9409, opened 2020-10-11
KAFKA-10599: Implement basic CLI tool for feature versioning system

kowshik:
This PR implements a basic CLI tool for feature versioning system. The [KIP-584 write up](https://cwiki.apache.org/confluence/display/KAFKA/KIP-584%3A+Versioning+scheme+for+features#KIP584:Versioningschemeforfeatures-Toolingsupport) has been updated to suit this PR. The following is implemented in this PR:
 - `--describe`:
    - Describe supported and finalized features.
    - Usage: `$> ./bin/kafka-features.sh --bootstrap-server host1:port1, host2:port2 --describe [--from-controller] [--command-config <path_to_java_properties_file>]`
    - Optionally, use the `--from-controller` option to get features from the controller.
 - `--upgrade-all`:
     - Upgrades all features known to the tool to their highest max version levels.
     - Usage: `$> ./bin/kafka-features.sh --bootstrap-server host1:port1, host2:port2 --upgrade-all [--dry-run] [--command-config <path_to_java_properties_file>]`
     - Optionally, use the `--dry-run` CLI option to preview the feature updates without actually applying them.
 - `--downgrade-all`:
    -  Downgrades existing finalized features to the highest max version levels known to this tool.
    - Usage: `$> ./bin/kafka-features.sh --bootstrap-server host1:port1, host2:port2  --downgrade-all [--dry-run] [--command-config <path_to_java_properties_file>]`.
    - Optionally, use the `--dry-run` CLI option to preview the feature updates without actually applying them.

**Tests:**
Added a new `FeatureCommand` integration test suite to test the CLI tool.

### DIFF
--- a/core/src/main/scala/kafka/admin/FeatureCommand.scala
+++ b/core/src/main/scala/kafka/admin/FeatureCommand.scala
@@ -73,14 +73,9 @@ class UpdateFeaturesException(message: String) extends RuntimeException(message)
  *
  * @param opts the CLI options
  */
-class FeatureApis(var opts: FeatureCommandOptions) {
+class FeatureApis(private var opts: FeatureCommandOptions) {
   private var supportedFeatures = BrokerFeatures.createDefault().supportedFeatures
-  private val adminClient = {
-    val props = new Properties()
-    props.putAll(opts.commandConfig)
-    props.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, opts.bootstrapServers)
-    Admin.create(props)
-  }
+  private var adminClient = createAdminClient(opts.commandConfig)
 
   private def pad(op: String): String = {
     f"$op%11s"
@@ -98,7 +93,16 @@ class FeatureApis(var opts: FeatureCommandOptions) {
 
   // For testing only.
   private[admin] def setOptions(newOpts: FeatureCommandOptions): Unit = {
+    adminClient.close()
+    adminClient = createAdminClient(newOpts.commandConfig)
     opts = newOpts
+  }
+
+  private def createAdminClient(props: Properties): Admin = {
+    val newProps = new Properties()
+    newProps.putAll(props)
+    newProps.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, opts.bootstrapServers)
+    Admin.create(newProps)
   }
 
   private def describeFeatures(sendRequestToController: Boolean): FeatureMetadata = {
--- a/core/src/main/scala/kafka/admin/FeatureCommand.scala
+++ b/core/src/main/scala/kafka/admin/FeatureCommand.scala
@@ -148,7 +148,7 @@ class FeatureApis(var opts: FeatureCommandOptions) {
   def upgradeAllFeatures(): Unit = {
     val metadata = describeFeatures(true)
     val existingFinalizedFeatures = metadata.finalizedFeatures
-    val updates = supportedFeatures.features().asScala.map {
+    val updates = supportedFeatures.features.asScala.map {
       case (feature, targetVersionRange) =>
         val existingVersionRange = existingFinalizedFeatures.get(feature)
         if (existingVersionRange == null) {
@@ -164,8 +164,8 @@ class FeatureApis(var opts: FeatureCommandOptions) {
               upgradeOp +
               s"\tFeature: $feature" +
               s"\tExistingFinalizedMaxVersion: ${existingVersionRange.maxVersionLevel}" +
-              s"\tNewFinalizedMaxVersion: ${targetVersionRange.max()}"
-            (feature, (updateStr, new FeatureUpdate(targetVersionRange.max(), false)))
+              s"\tNewFinalizedMaxVersion: ${targetVersionRange.max}"
+            (feature, (updateStr, new FeatureUpdate(targetVersionRange.max, false)))
           } else {
             (feature, null)
           }
@@ -191,7 +191,7 @@ class FeatureApis(var opts: FeatureCommandOptions) {
   def downgradeAllFeatures(): Unit = {
     val metadata = describeFeatures(true)
     val existingFinalizedFeatures = metadata.finalizedFeatures
-    val supportedFeaturesMap = supportedFeatures.features()
+    val supportedFeaturesMap = supportedFeatures.features
     val updates = existingFinalizedFeatures.asScala.map {
       case (feature, existingVersionRange) =>
         val targetVersionRange = supportedFeaturesMap.get(feature)
@@ -199,7 +199,7 @@ class FeatureApis(var opts: FeatureCommandOptions) {
           val updateStr =
             deleteOp +
             s"\tFeature: $feature" +
-            s"\tExistingFinalizedMaxVersion: ${existingVersionRange.maxVersionLevel()}" +
+            s"\tExistingFinalizedMaxVersion: ${existingVersionRange.maxVersionLevel}" +
             s"\tNewFinalizedMaxVersion: -"
           (feature, (updateStr, new FeatureUpdate(0, true)))
         } else {
@@ -208,8 +208,8 @@ class FeatureApis(var opts: FeatureCommandOptions) {
               downgradeOp +
               s"\tFeature: $feature" +
               s"\tExistingFinalizedMaxVersion: ${existingVersionRange.maxVersionLevel}" +
-              s"\tNewFinalizedMaxVersion: ${targetVersionRange.max()}"
-            (feature, (updateStr, new FeatureUpdate(targetVersionRange.max(), true)))
+              s"\tNewFinalizedMaxVersion: ${targetVersionRange.max}"
+            (feature, (updateStr, new FeatureUpdate(targetVersionRange.max, true)))
           } else {
             (feature, null)
           }
@@ -239,7 +239,7 @@ class FeatureApis(var opts: FeatureCommandOptions) {
       val result = adminClient.updateFeatures(
         updates.map { case(feature, (_, update)) => (feature, update)}.asJava,
         new UpdateFeaturesOptions())
-      val failures = ListMap(result.values().asScala.toSeq.sortBy(_._1):_*).map {
+      val failures = ListMap(result.values.asScala.toSeq.sortBy(_._1):_*).map {
         case (feature, updateFuture) =>
           val (updateStr, _) = updates(feature)
           try {
--- a/core/src/main/scala/kafka/admin/FeatureCommand.scala
+++ b/core/src/main/scala/kafka/admin/FeatureCommand.scala
@@ -75,7 +75,7 @@ class UpdateFeaturesException(message: String) extends RuntimeException(message)
  */
 class FeatureApis(private var opts: FeatureCommandOptions) {
   private var supportedFeatures = BrokerFeatures.createDefault().supportedFeatures
-  private var adminClient = createAdminClient(opts.commandConfig)
+  private var adminClient = FeatureApis.createAdminClient(opts)
 
   private def pad(op: String): String = {
     f"$op%11s"
@@ -94,15 +94,8 @@ class FeatureApis(private var opts: FeatureCommandOptions) {
   // For testing only.
   private[admin] def setOptions(newOpts: FeatureCommandOptions): Unit = {
     adminClient.close()
-    adminClient = createAdminClient(newOpts.commandConfig)
+    adminClient = FeatureApis.createAdminClient(newOpts)
     opts = newOpts
-  }
-
-  private def createAdminClient(props: Properties): Admin = {
-    val newProps = new Properties()
-    newProps.putAll(props)
-    newProps.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, opts.bootstrapServers)
-    Admin.create(newProps)
   }
 
   private def describeFeatures(sendRequestToController: Boolean): FeatureMetadata = {
@@ -402,5 +395,14 @@ class FeatureCommandOptions(args: Array[String]) extends CommandDefaultOptions(a
         parser,
         "Command can contain --from-controller option only when --describe action is provided.")
     }
+  }
+}
+
+object FeatureApis {
+  private def createAdminClient(opts: FeatureCommandOptions): Admin = {
+    val props = new Properties()
+    props.putAll(opts.commandConfig)
+    props.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, opts.bootstrapServers)
+    Admin.create(props)
   }
 }
--- a/core/src/main/scala/kafka/admin/FeatureCommand.scala
+++ b/core/src/main/scala/kafka/admin/FeatureCommand.scala
@@ -63,7 +63,8 @@ object FeatureCommand {
 class UpdateFeaturesException(message: String) extends RuntimeException(message)
 
 /**
- * A class that provides necessary APIs to bridge the Admin client feature APIs with the CLI tool.
+ * A class that provides necessary APIs to bridge feature APIs provided by the the Admin client with
+ * the requirements of the CLI tool.
  *
  * @param opts the CLI options
  */
--- a/core/src/main/scala/kafka/admin/FeatureCommand.scala
+++ b/core/src/main/scala/kafka/admin/FeatureCommand.scala
@@ -69,7 +69,12 @@ class UpdateFeaturesException(message: String) extends RuntimeException(message)
  */
 class FeatureApis(var opts: FeatureCommandOptions) {
   private var supportedFeatures = BrokerFeatures.createDefault().supportedFeatures
-  private val adminClient = createAdminClient(opts.commandConfig)
+  private val adminClient = {
+    val props = new Properties()
+    props.putAll(opts.commandConfig)
+    props.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, opts.bootstrapServers)
+    Admin.create(props)
+  }
 
   private def pad(op: String): String = {
     f"$op%11s"
@@ -251,7 +256,7 @@ class FeatureApis(var opts: FeatureCommandOptions) {
       println("Expected feature updates:" + ListMap(
         updates
           .toSeq
-          .sortBy{ case(feature, _) => feature} :_*)
+          .sortBy { case(feature, _) => feature} :_*)
           .map { case(_, (updateStr, _)) => updateStr}
           .mkString("\n"))
     } else {
@@ -299,13 +304,6 @@ class FeatureApis(var opts: FeatureCommandOptions) {
 
   def close(): Unit = {
     adminClient.close()
-  }
-
-  private def createAdminClient(commandConfig: Properties): Admin = {
-    val props = new Properties()
-    props.putAll(commandConfig)
-    props.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, opts.bootstrapServers)
-    Admin.create(props)
   }
 }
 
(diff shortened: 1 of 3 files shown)


kowshik:
@abbccdda @junrao : This PR is ready for review.

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,350 @@
...
+        printException(e)
+        exitCode = 1
+    } finally {
+      featureApis.close()
+      Exit.exit(exitCode)
+    }
+  }
+
+  private def printException(exception: Throwable): Unit = {
+    System.err.println("\nError encountered when executing command: " + Utils.stackTrace(exception))
```
abbccdda, line 64: Is the `\n` necessary?

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,350 @@
...
+    "dry-run",
+    "Performs a dry-run of upgrade/downgrade mutations to finalized feature without applying them.")
+
+  options = parser.parse(args : _*)
+
+  checkArgs()
+
+  def has(builder: OptionSpec[_]): Boolean = options.has(builder)
+
+  def hasDescribeOption: Boolean = has(describeOpt)
```
abbccdda, line 360: I feel we don't need the `has` prefix for all the options, which seem not matching with the conventions by looking at examples as `CofnigCommand.scala`

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,350 @@
...
+
+  def hasFromControllerOption: Boolean = has(fromControllerOpt)
+
+  def hasDryRunOption: Boolean = has(dryRunOpt)
+
+  def hasUpgradeAllOption: Boolean = has(upgradeAllOpt)
```
abbccdda, line 366: A random thought I have is that since describe/upgrade/downgrade are mutually-exclusive, we could define one flag like `--action` and pass in the different options as strings to translate the an enum matching scenario.

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,350 @@
...
+        parser,
+        "Command must include exactly one action: --describe, --upgrade-all, --downgrade-all.")
+    }
+    CommandLineUtils.checkRequiredArgs(parser, options, bootstrapServerOpt)
+    if (hasDryRunOption && !hasUpgradeAllOption && !hasDowngradeAllOption) {
+      CommandLineUtils.printUsageAndDie(
+        parser,
+        "Command can contain --dry-run option only when either --upgrade-all or --downgrade-all actions are provided.")
+    }
+    if (hasFromControllerOption && !hasDescribeOption) {
```
abbccdda, line 393: Could we just log a warning in this case instead of failing?

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,350 @@
...
+            updateFuture.get
+            println(updateStr + "\tResult: OK")
+            0
+          } catch {
+            case e: Exception =>
+              println(updateStr + "\tResult: FAILED due to " + e.getMessage)
+              1
+          }
+      }.sum
+      if (failures > 0) {
```
abbccdda, line 294: I'm not sure the exception thrown here is necessary since we already printed out the result, is this just for testing purpose?

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,350 @@
...
+  /**
+   * Describes the supported and finalized features. If the --from-controller CLI option
+   * is provided, then the request is issued only to the controller, otherwise the request is issued
+   * to any of the provided bootstrap servers.
+   */
+  def describeFeatures(): Unit = {
+    val result = describeFeatures(opts.hasFromControllerOption)
+    val features = result.supportedFeatures.asScala.keys.toSet ++ result.finalizedFeatures.asScala.keys.toSet
+    features.toList.sorted.foreach {
+      case feature =>
```
abbccdda: nit: case not necessary

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,350 @@
...
+
+  /**
+   * Describes the supported and finalized features. If the --from-controller CLI option
+   * is provided, then the request is issued only to the controller, otherwise the request is issued
+   * to any of the provided bootstrap servers.
+   */
+  def describeFeatures(): Unit = {
+    val result = describeFeatures(opts.hasFromControllerOption)
+    val features = result.supportedFeatures.asScala.keys.toSet ++ result.finalizedFeatures.asScala.keys.toSet
+    features.toList.sorted.foreach {
```
abbccdda, line 115: Why do we need to sort the features?

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,350 @@
...
+          if (supportedVersionRange == null) {
+            ("-", "-")
+          } else {
+            (supportedVersionRange.minVersion, supportedVersionRange.maxVersion)
+          }
+        }
+        print(s"Feature: $feature")
+        print(s"\tSupportedMinVersion: $supportedMinVersion")
+        print(s"\tSupportedMaxVersion: $supportedMaxVersion")
+        print(s"\tFinalizedMinVersionLevel: $finalizedMinVersionLevel")
```
abbccdda: nit: we could do the comment immediately after the version get extracted, just like moving L132 to be after L113

core/src/test/scala/unit/kafka/admin/FeatureCommandTest.scala
```diff
@@ -0,0 +1,245 @@
...
+    } finally {
+      featureApis.close()
+    }
+  }
+
+  /**
+   * Tests if the FeatureApis#upgradeAllFeatures API works as expected during a partial failure case.
+   */
+  @Test
+  def testUpgradeFeaturesFailure(): Unit = {
```
abbccdda, line 200: Do we need a downgrade failure test as well?

abbccdda (commented):
Thanks for the PR. Left a couple of comments, and one thing to point out is that we need to look at the KIP-584 definition for the basic tool usage and make sure the API names are aligned.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: @abbccdda How do you envision the proposed `--action` flag for the advanced CLI scenarios where we would want to allow for `--upgrade`, `--downgrade` and `--delete` options as part of the same CLI command? Please [see this section](https://cwiki.apache.org/confluence/display/KAFKA/KIP-584%3A+Versioning+scheme+for+features#KIP584:Versioningschemeforfeatures-AdvancedCLItoolusage) of the KIP. As such, the existing technique allows us to support the advanced CLI facilities in the future.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Yeah, it provides a separation between the regular o/p and the error.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: This convention is used in `TopicCommand.scala`. See: https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/admin/TopicCommand.scala#L698-L702

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: What do we lose by making a stricter check? This is not the expected usage of the command, and therefore it should be disallowed.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done. Good point. This is fixed now, as we are no longer printing exception from within `FeatureCommand.main`. Yes, the exception as such is useful for testing too.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: It is easy to compare command outputs when they are sorted based on feature names. Also, it is easier to write unit/integration tests when the output is uniform.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done.

core/src/test/scala/unit/kafka/admin/FeatureCommandTest.scala
kowshik: As such the downgrade can not fail currently on the server. The reason is the following: the `FeatureCommand` downgrades `maxVersionLevel` of existing finalized features to the max versions that it knows internally (since it links with the `BrokerFeatures` library). As a result, the only situation when the `FeatureCommand` can fail is if it tries to downgrade below the supported `minVersion`. But this can never happen since we don't (yet) have the facility on the broker to have a supported `minVersion` > 1 (this can only happen during feature deprecation, the support for this is future work).

On a side note, we could maybe write a test using `AdminClientUnitTestEnv` to create a mock setup specifically for the downgrade test and inject a mock admin client into the `FeatureApis` class. But that looks overkill to me...

core/src/main/scala/kafka/admin/FeatureCommand.scala
abbccdda: I just don't see it hinders the command to succeed, which means it's an ignorable property. Anyway, I'm not insisting either.

kowshik: Sounds good. I'll mark this resolved since you don't feel strongly about t.

kowshik:
> one thing to point out is that we need to look at the KIP-584 definition for the basic tool usage and make sure the API names are aligned.

This is done now. Please refer to the PR description.

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,359 @@
...
+    f"$op%11s"
+  }
+
+  private val addOp = pad("[Add]")
+  private val upgradeOp = pad("[Upgrade]")
+  private val deleteOp = pad("[Delete]")
+  private val downgradeOp = pad("[Downgrade]")
+
+  // For testing only.
+  def setSupportedFeatures(newFeatures: Features[SupportedVersionRange]): Unit = {
```
abbccdda: nit: we could make these functions package private

abbccdda (commented):
Overall LGTM

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done

kowshik:
@abbccdda Thanks for the review! I've addressed the most recent comment.

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,359 @@
...
+  private def createAdminClient(): Admin = {
+    val props = new Properties()
+    props.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, opts.bootstrapServers)
+    Admin.create(props)
+  }
+}
+
+class FeatureCommandOptions(args: Array[String]) extends CommandDefaultOptions(args) {
+  private val bootstrapServerOpt =
+    parser.accepts("bootstrap-server", "REQUIRED: The Kafka server(s) to connect to.")
```
junrao: Perhaps we can describe the format is host:port list.

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,359 @@
...
+        } else {
+          if (targetVersionRange.max > existingVersionRange.maxVersionLevel) {
+            val updateStr =
+              upgradeOp +
+              s"\tFeature: $feature" +
+              s"\tExistingFinalizedMaxVersion: ${existingVersionRange.maxVersionLevel}" +
+              s"\tNewFinalizedMaxVersion: ${targetVersionRange.max}"
+            (feature, (updateStr, new FeatureUpdate(targetVersionRange.max, false)))
+          } else {
+            (feature, null)
```
junrao: Perhaps we could use Option to avoid null?

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,359 @@
...
+   * only prints the expected feature updates to STDOUT without applying them.
+   *
+   * @param updates the feature updates to be applied via the admin client
+   *
+   * @throws UpdateFeaturesException if at least one of the feature updates failed
+   */
+  private def maybeApplyFeatureUpdates(updates: Map[String, (String, FeatureUpdate)]): Unit = {
+    if (opts.hasDryRunOption) {
+      println("Expected feature updates:")
+      println(ListMap(updates.toSeq.sortBy(_._1):_*)
```
junrao: Could we use case to avoid unnamed reference _._1?

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,359 @@
...
+                .map { case(_, (updateStr, _)) => updateStr}
+                .mkString("\n"))
+    } else {
+      val result = adminClient.updateFeatures(
+        updates.map { case(feature, (_, update)) => (feature, update)}.asJava,
+        new UpdateFeaturesOptions())
+      val failures = ListMap(result.values.asScala.toSeq.sortBy(_._1):_*).map {
```
junrao: Could we use case to avoid unnamed reference _._1?

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,359 @@
...
+        case (feature, updateFuture) =>
+          val (updateStr, _) = updates(feature)
+          try {
+            updateFuture.get
+            println(updateStr + "\tResult: OK")
+            0
+          } catch {
+            case e: Exception =>
+              println(updateStr + "\tResult: FAILED due to " + e.getMessage)
```
junrao: Do we need to get the cause from ExecutionException thrown from Future?

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,359 @@
...
+  }
+
+  def close(): Unit = {
+    adminClient.close()
+  }
+
+  private def createAdminClient(): Admin = {
+    val props = new Properties()
+    props.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, opts.bootstrapServers)
+    Admin.create(props)
```
junrao: The bootstrap port may be secured (e.g., SSL, SASL). So we need to be able to pass in the security configs from the command line. See commandConfigOpt in TopicCommand.

core/src/test/scala/unit/kafka/admin/FeatureCommandTest.scala
```diff
@@ -0,0 +1,245 @@
...
+import org.scalatest.Assertions.intercept
+
+class FeatureCommandTest extends BaseRequestTest {
+  override def brokerCount: Int = 3
+
+  override def brokerPropertyOverrides(props: Properties): Unit = {
+    props.put(KafkaConfig.InterBrokerProtocolVersionProp, KAFKA_2_7_IV0.toString)
+  }
+
+  private def defaultSupportedFeatures(): Features[SupportedVersionRange] = {
```
junrao: Could this be a val?

junrao (commented):
@kowshik : Thanks for the PR. A few comments below.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done. I'm printing the exception `toString` output now, instead of `getMessage`.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done. Great point! I've provided an additional arg: `--command-config <path_to_properties>` file that can be used to inject the additional properties into admin client.

core/src/test/scala/unit/kafka/admin/FeatureCommandTest.scala
kowshik: Done

kowshik:
@junrao Thank you for the review! I've address the comments and the PR is ready for review again.

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,393 @@
...
+      val failures = resultSortedByFeature.map {
+        case (feature, updateFuture) =>
+          val (updateStr, _) = updates(feature)
+          try {
+            updateFuture.get
+            println(updateStr + "\tResult: OK")
+            0
+          } catch {
+            case e: Exception =>
+              println(updateStr + "\tResult: FAILED due to " + e)
```
junrao, line 290: If we hit ExecutionException and the exception has a cause, the wrapper ExecutionException is not really relevant and we should just print the cause. See TopicCommand as a reference.

core/src/main/scala/kafka/admin/FeatureCommand.scala
```diff
@@ -0,0 +1,393 @@
...
+    try {
+      featureApis.execute()
+    } catch {
+      case e: IllegalArgumentException =>
+        printException(e)
+        opts.parser.printHelpOn(System.err)
+        exitCode = 1
+      case _: UpdateFeaturesException =>
+        exitCode = 1
+      case e: Throwable =>
```
junrao, line 54: It seems that we can hit ExecutionException through describeFeatures. We want to handle it properly too.

junrao (commented):
@kowshik : Thanks for the updated PR. Just a couple of more comments.

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done. Great point!

core/src/main/scala/kafka/admin/FeatureCommand.scala
kowshik: Done.

junrao (approved):
@kowshik : Thanks for the latest PR. LGTM.

@abbccdda : Do you want to take another look?

abbccdda (approved):
Nope, LGTM!

PR #9412, opened 2020-10-11
MINOR: Ensure broker advertises features only when versioning system is enabled

kowshik:
This is a fix that ensures the broker advertises features to ZK only when the feature versioning system is enabled.

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaServer.scala
+++ b/core/src/main/scala/kafka/server/KafkaServer.scala
@@ -35,7 +35,8 @@ import kafka.network.SocketServer
 import kafka.security.CredentialProvider
 import kafka.utils._
 import kafka.zk.{BrokerInfo, KafkaZkClient}
-import org.apache.kafka.clients.{ApiVersions, ClientDnsLookup, ManualMetadataUpdater, NetworkClient, NetworkClientUtils, CommonClientConfigs}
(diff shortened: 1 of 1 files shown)


kowshik:
@abbccdda @junrao : Please could you have a look at this fix. We have to merge it into trunk as well as the 2.7 release branch.

core/src/main/scala/kafka/server/KafkaServer.scala
```diff
@@ -494,8 +495,14 @@ class KafkaServer(val config: KafkaConfig, time: Time = Time.SYSTEM, threadNameP
     )
 
     val jmxPort = System.getProperty("com.sun.management.jmxremote.port", "-1").toInt
+
+    val supportedFeatures =
```
abbccdda, line 499: Could we add a test coverage for this logic?

kowshik:
@abbccdda Sorry for the confusion. Upon second look, I feel we can abandon this PR containing the fix.
I was looking at the code again, and it seems the requirement is already take care of in the earlier PRs. If IBP >= `KAFKA_2_7_IV0` we set Broker version to 5, which eventually causes features to be advertised in `BrokerIdZNode.encode`. As a result, the logic is already restrictive like the way we would have liked. Here are some pointers to the code:

https://github.com/apache/kafka/blob/fb4f297207ef62f71e4a6d2d0dac75752933043d/core/src/main/scala/kafka/zk/ZkData.scala#L99-L100

https://github.com/apache/kafka/blob/fb4f297207ef62f71e4a6d2d0dac75752933043d/core/src/main/scala/kafka/zk/ZkData.scala#L151-L153

PR #9417, opened 2020-10-12
MINOR: Fix flaky ControllerMutationQuotaTest.testQuotaMetric

splett2:
### What
`ClientQuotaManager.updateQuota` updates `quotaCallback` before updating metric configs.
`ClientQuotaManager.quota()` performs an unlocked call to `quotaLimit()`, so we can run into a case where `quota` returns the updated quota, but the metric config hasn't been updated.

The fix is to acquire the read lock before attempting to read the quota. An alternative would be to have `verifyQuotaMetric` run in a `waitUntilTrue`. 

### DIFF
--- a/core/src/test/scala/unit/kafka/server/ControllerMutationQuotaTest.scala
+++ b/core/src/test/scala/unit/kafka/server/ControllerMutationQuotaTest.scala
@@ -140,14 +140,14 @@ class ControllerMutationQuotaTest extends BaseRequestTest {
       assertEquals(Set(Errors.NONE), errors.values.toSet)
 
       // Metric must be there with the correct config
-      verifyQuotaMetric(principal.getName, ControllerMutationRate)
+      waitQuotaMetric(principal.getName, ControllerMutationRate)
 
       // Update quota
       defineUserQuota(ThrottledPrincipal.getName, Some(ControllerMutationRate * 2))
       waitUserQuota(ThrottledPrincipal.getName, ControllerMutationRate * 2)
 
       // Metric must be there with the updated config
-      verifyQuotaMetric(principal.getName, ControllerMutationRate * 2)
+      waitQuotaMetric(principal.getName, ControllerMutationRate * 2)
     }
   }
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/server/ClientQuotaManager.scala
```diff
@@ -360,8 +360,14 @@ class ClientQuotaManager(private val config: ClientQuotaManagerConfig,
...
-    val metricTags = quotaCallback.quotaMetricTags(clientQuotaType, userPrincipal, clientId)
-    Quota.upperBound(quotaLimit(metricTags))
+    // acquire read lock to ensure that both quota limit and metric config are updated atomically
+    lock.readLock().lock()
+    try {
+      val metricTags = quotaCallback.quotaMetricTags(clientQuotaType, userPrincipal, clientId)
+      Quota.upperBound(quotaLimit(metricTags))
+    } finally {
+      lock.readLock().unlock()
+    }
```
dajac: While this works, I wonder if this is the right way to overcome the issue. My concern is that there are other read paths which remain unprotected so we are not consistent. I would rather prefer to update the test as you suggested.

dajac (commented):
@splett2 Nice find. Thanks for the PR. I have left a general comment.

core/src/test/scala/unit/kafka/server/ControllerMutationQuotaTest.scala
```diff
@@ -381,16 +381,18 @@ class ControllerMutationQuotaTest extends BaseRequestTest {
...
-      case Some(metric) =>
-        val config = metric.config()
-        assertEquals(expectedQuota, config.quota().bound(), 0.1)
-        assertEquals(ControllerQuotaSamples, config.samples())
-        assertEquals(ControllerQuotaWindowSizeSeconds * 1000, config.timeWindowMs())
-
-      case None =>
-        fail(s"Quota metric of $user is not defined")
+  private def waitQuotaMetric(user: String, expectedQuota: Double): Unit = {
+    TestUtils.retry(200) {
```
dajac: nit: Could we use `DEFAULT_MAX_WAIT_MS` (defined in `o.a.k.t.TestUtils`) here?

dajac (commented):
@splett2 Thanks for the update. I just left one minor comment.

dajac:
@splett2 Thanks for the update. Could you rebase the PR to include https://github.com/apache/kafka/commit/2db67db8e1329cb2e047322cff81d97ff98b4328? Tests won't run otherwise.

dajac (approved):
@splett2 Thanks for the PR, LGTM.

PR #9418, opened 2020-10-12
KAFKA-10601; Add support for append linger to Raft implementation

hachikuji:
The main purpose of this patch is to add `quorum.linger.ms` behavior to the raft implementation. This gives users a powerful knob to tune the impact of fsync.  When an append is accepted from the state machine, it is held in an accumulator (similar to the producer) until the configured linger time is exceeded. This allows the implementation to amortize fsync overhead at the expense of some write latency.

The patch also improves our methodology for testing performance. Up to now, we have relied on the producer performance test, but it is difficult to simulate expected controller loads because producer performance is limited by other factors such as the number of producer clients and head-of-line blocking. Instead, this patch adds a workload generator which runs on the leader after election.

Finally, this patch brings us nearer to the write semantics expected by the KIP-500 controller. It makes the following changes:

- Introduce `RecordSerde<T>` interface which abstracts the underlying log implementation from `RaftClient`. The generic type is carried over to `RaftClient<T>` and is exposed through the read/write APIs.
- `RaftClient.append` is changed to `RaftClient.scheduleAppend` and returns the last offset of the expected log append.
- `RaftClient.scheduleAppend` accepts a list of records and ensures that the full set are included in a single batch.
- Introduce `RaftClient.Listener` with a single `handleCommit` API which will eventually replace `RaftClient.read` in order to surface committed data to the controller state machine. Currently `handleCommit` is only used for records appended by the leader.

(Note that this patch piggybacks on top of #9352. I will rebase as soon as this PR has been merged.)


raft/src/main/java/org/apache/kafka/raft/RaftClient.java
```diff
@@ -19,33 +19,51 @@
...
+         *
+         * Note that there is not a one-to-one correspondence between writes through
+         * {@link #scheduleAppend(int, List)} and this callback. The Raft implementation
+         * is free to batch together the records from multiple append calls provided
+         * that batch boundaries are respected. This means that each batch specified
+         * through {@link #scheduleAppend(int, List)} is guaranteed to be a subset of
+         * a batch passed to {@link #handleCommit(int, long, List)}.
+         *
+         * @param epoch the epoch in which the write was accepted
+         * @param lastOffset the last offset of the recor
```
hachikuji: fix typo

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
```diff
@@ -19,33 +19,51 @@
...
-     * @param timeoutMs Maximum time to wait for the append to complete
-     * @return A future containing the last offset and epoch of the appended records (if successful)
+     * @param epoch the current leader epoch
+     * @param records the list of records to append
+     * @return the offset within the current epoch that the log entries will be appended,
+     *         or null if the leader was unable to accept the write (e.g. due to memory
+     *         being reached).
      */
-    CompletableFuture<OffsetAndEpoch> append(Records records, AckMode ackMode, long timeoutMs);
+    Long scheduleAppend(int epoch, List<T> records);
```
jsancio, line 72: How about either 
```
OptionalLong scheduleAppend(int epoch, List<T> records);
```
or
```
void scheduleAppend(int epoch, List<T> records) throws BusyException;
```

I am okay with either solution but I am wondering why did you decide to return a `null` for this case instead of throwing some exception?

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -0,0 +1,296 @@
...
+import org.apache.kafka.raft.RecordSerde;
+
+import java.io.Closeable;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * TODO: Also flush after minimum size limit is reached?
```
jsancio: I vote yes for this.

I think if we use this for writing snapshot from the state machine, then minimum size is a more interesting metrics for flushing to disk vs lingerMs.

If we implement this so that either one has to be true then the client can set the `lingerMs` or `minSize` to MAX_VALUE if it wants to ignore those values.

jsancio (commented):
Thanks for the PR. Left some comments.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
hachikuji: Yeah, this was more of a workaround until we thought of something better. The current idea we are thinking about is letting the Raft layer return a backoff time, so the type would semantically be `Either[Offset, BackoffMs]`.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
hachikuji: Yeah, I think we should do it. I will create a JIRA and leave this for a follow-up.

clients/src/main/java/org/apache/kafka/common/protocol/DataOutputWritable.java
```diff
@@ -0,0 +1,129 @@
...
+package org.apache.kafka.common.protocol;
+
+import org.apache.kafka.common.utils.ByteUtils;
+import org.apache.kafka.common.utils.Utils;
+
+import java.io.DataOutput;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+public class DataOutputWritable implements Writable {
```
guozhangwang: nit: not clear why we want an intermediate class instead of just having a `DataOutputStreamWritable` class and its constructor parameter declared as `DataOutputStream`?

core/src/main/scala/kafka/common/RecordValidationException.scala
```diff
@@ -23,5 +23,6 @@ import org.apache.kafka.common.requests.ProduceResponse.RecordError
 import scala.collection.Seq
 
 class RecordValidationException(val invalidException: ApiException,
-                                val recordErrors: Seq[RecordError]) extends RuntimeException {
+                                val recordErrors: Seq[RecordError])
+  extends RuntimeException(invalidException) {
```
guozhangwang, line 27: I vaguely remember @rajinisivaram has another PR which may be overlapping with this fix, will let her to check and confirm if it is fine.

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
```diff
@@ -216,11 +216,9 @@ class KafkaNetworkChannel(time: Time,
     endpoints.put(id, node)
   }
 
-  def postInboundRequest(header: RequestHeader,
-                         request: AbstractRequest,
-                         onResponseReceived: ResponseHandler): Unit = {
+  def postInboundRequest(request: AbstractRequest, onResponseReceived: ResponseHandler): Unit = {
```
guozhangwang, line 227: nit: maybe add a comment indicating this function is for testing only (and hence we do not care about the correlation id).

core/src/main/scala/kafka/tools/TestRaftRequestHandler.scala
```diff
@@ -56,73 +47,8 @@ class TestRaftRequestHandler(
              | ApiKeys.END_QUORUM_EPOCH
              | ApiKeys.FETCH =>
           val requestBody = request.body[AbstractRequest]
-          networkChannel.postInboundRequest(
-            request.header,
-            requestBody,
-            response => sendResponse(request, Some(response)))
-
-        case ApiKeys.API_VERSIONS =>
```
guozhangwang, line 64: Could you elaborate a bit why we would not expect these api keys in the test handler any more?

core/src/main/scala/kafka/tools/TestRaftServer.scala
```diff
@@ -272,7 +291,79 @@ class TestRaftServer(val config: KafkaConfig) extends Logging {
...
+
+        val startTimeMs = time.milliseconds()
+        val sendTimeMs = if (throttler.maybeThrottle(recordCount, startTimeMs)) {
+          time.milliseconds()
+        } else {
+          startTimeMs
+        }
+
+        val offset = client.scheduleAppend(latestLeaderAndEpoch.epoch, Collections.singletonList(payload))
+        if (offset == null || offset == Long.MaxValue) {
```
guozhangwang: If `offset == Long.MaxValue` would sleeping save us anything? Should we terminate the testing server then?

core/src/main/scala/kafka/tools/TestRaftServer.scala
```diff
@@ -272,7 +291,79 @@ class TestRaftServer(val config: KafkaConfig) extends Logging {
...
+    }
+
+    override def handleCommit(epoch: Int, lastOffset: Long, records: util.List[Array[Byte]]): Unit = {
+      var offset = lastOffset - records.size() + 1
+      val currentTimeMs = time.milliseconds()
+
+      for (record <- records.asScala) {
+        val pendingAppend = pendingAppends.poll()
+        if (pendingAppend.epoch != epoch || pendingAppend.offset!= offset) {
+          warn(s"Expected next commit at offset ${pendingAppend.offset}, " +
```
guozhangwang: Should this ever happen? If not, then we could shutdown the server after logging the error.

core/src/main/scala/kafka/tools/TestRaftServer.scala
```diff
@@ -19,37 +19,44 @@ package kafka.tools
...
 
-import joptsimple.OptionParser
+import com.yammer.metrics.core.MetricName
+import joptsimple.OptionException
 import kafka.log.{Log, LogConfig, LogManager}
 import kafka.network.SocketServer
 import kafka.raft.{KafkaFuturePurgatory, KafkaMetadataLog, KafkaNetworkChannel}
 import kafka.security.CredentialProvider
 import kafka.server.{BrokerTopicStats, KafkaConfig, KafkaRequestHandlerPool, KafkaServer, LogDirFailureChannel}
+import kafka.tools.TestRaftServer.{ByteArraySerde, PendingAppend, ThroughputThrottler, WriteStats}
```
guozhangwang: Why we need to import the inner classes?

core/src/main/scala/kafka/tools/TestRaftServer.scala
```diff
@@ -272,7 +291,79 @@ class TestRaftServer(val config: KafkaConfig) extends Logging {
...
+    private var isLeader = false
+    private var throttler: ThroughputThrottler = _
+    private var recordCount = 0
+
+    override def doWork(): Unit = {
+      if (latestLeaderAndEpoch != client.currentLeaderAndEpoch()) {
+        latestLeaderAndEpoch = client.currentLeaderAndEpoch()
+        isLeader = latestLeaderAndEpoch.leaderId.orElse(-1) == brokerId
+        if (isLeader) {
+          pendingAppends.clear()
```
guozhangwang, line 316: Even if the leader has changed back, previously pending appends may still get committed right? I think it is sufficient to just poll-and-drop them in `handleCommit` when leader changes and not clearing the queue here.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -148,36 +162,44 @@ public KafkaRaftClient(RaftConfig raftConfig,
...
+        Time time,
+        Metrics metrics,
+        FuturePurgatory<LogOffset> fetchPurgatory,
+        FuturePurgatory<LogOffset> appendPurgatory,
+        Map<Integer, InetSocketAddress> voterAddresses,
+        int electionBackoffMaxMs,
+        int retryBackoffMs,
+        int requestTimeoutMs,
+        int fetchMaxWaitMs,
+        int lingerMs,
```
guozhangwang: nit: `appendLingerMs`?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1443,15 +1485,79 @@ private void pollShutdown(GracefulShutdown shutdown) throws IOException {
...
+        if (timeUnitFlush <= 0) {
+            List<BatchAccumulator.CompletedBatch<T>> batches = accumulator.flush();
+            Iterator<BatchAccumulator.CompletedBatch<T>> iterator = batches.iterator();
+
+            try {
+                while (iterator.hasNext()) {
+                    BatchAccumulator.CompletedBatch<T> batch = iterator.next();
+                    appendBatch(state, batch, currentTimeMs);
+                }
+                flushLeaderLog(state, currentTimeMs);
```
guozhangwang, line 1535: For now since we only have a single thread processing all incoming req/resp, this is okay; but when we multi-thread processing requests this would no longer be safe, since it is possible that some batches gets replicated and committed while not being flushed locally yet.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1605,100 +1708,18 @@ public void poll() throws IOException {
...
-        if (quorum.isObserver())
-            throw new IllegalStateException("Illegal attempt to write to an observer");
-
-        CompletableFuture<OffsetAndEpoch> future = new CompletableFuture<>();
-        UnwrittenAppend unwrittenAppend = new UnwrittenAppend(
-            records, time.milliseconds(), timeoutMs, ackMode, future);
+    public Long scheduleAppend(int epoch, List<T> records) {
+        BatchAccumulator<T> accumulator = this.accumulator;
+        if (accumulator == null) {
+            return Long.MAX_VALUE;
```
guozhangwang, line 1715: Why return MAX_VALUE instead of null here? If we want to use `null` to indicate `memory full` and use `MAX_VALUE` to indicate `not leader`, the javadoc should reflecting this.

Anyways, I think returning sth like a `combo(Offset, ErrorCode, backoffMs)` would be preferred in the end state.

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
```diff
@@ -104,29 +62,17 @@ public synchronized boolean isWritable() {
...
-        });
-    }
-
-    private SimpleRecord serialize(int value) {
-        ByteBuffer buffer = ByteBuffer.allocate(4);
-        Type.INT32.write(buffer, value);
-        buffer.flip();
-        return new SimpleRecord(buffer);
+        uncommitted += 1;
+        long offset = client.scheduleAppend(currentLeaderAndEpoch.epoch, Collections.singletonList(uncommitted));
```
guozhangwang: Should we handle `null / max-value` here?

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+    }
+
+    /**
+     * Get the number of batches including the one that is currently being
+     * written to (if it exists).
+     *
+     * @return
+     */
+    public int count() {
```
guozhangwang, line 278: Is this going to be used for non-testing code in the future? If it is only going to be for metrics purposes maybe we can allow it to be non thread-safe just to not blocking on other critical paths.

raft/src/main/java/org/apache/kafka/raft/internals/BatchBuilder.java
```diff
@@ -0,0 +1,231 @@
...
+import org.apache.kafka.common.utils.ByteBufferOutputStream;
+import org.apache.kafka.common.utils.ByteUtils;
+import org.apache.kafka.raft.RecordSerde;
+
+import java.io.DataOutputStream;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.List;
+
+public class BatchBuilder<T> {
```
guozhangwang, line 46: nit: can we consolidate this class with the existing one (of course, it needs to be generalized on T record) than creating a new class? Ditto for BatchAccumulator and BatchMemoryPool.

guozhangwang (commented):
Made a pass on non-testing code.

clients/src/main/java/org/apache/kafka/common/protocol/DataOutputWritable.java
hachikuji: Fair enough. We have no current need for `DataOutputWritable`, so I will remove this.

core/src/main/scala/kafka/tools/TestRaftRequestHandler.scala
hachikuji: I felt it was too difficult to approximate a controller workload using one or more producers because of the fact that we can only handle one request at a time. So I created a separate workload generator which executes on the leader and I removed all of this somewhat hacky handling logic which allowed us to use a producer.

core/src/main/scala/kafka/tools/TestRaftServer.scala
hachikuji: I will try to document this better, but `Long.MaxValue` is how we decided to handle the case where the epoch in `scheduleAppend` does not match the current epoch. This can happen because the raft epoch is updated asynchronously and there is no way to ensure the state machine has seen the latest value. The expectation is that the state machine will update its uncommitted state with an offset which is impossible to become committed. After it observes the epoch change, this uncommitted state will be discarded.

Note that although I added the explicit check here, it is not technically necessary. Let me consider removing it.

core/src/main/scala/kafka/tools/TestRaftServer.scala
hachikuji: That's true. In a follow-up, the `handleCommit` API will be expanded a bit to cover appends through replication as well, but for now, I think we can raise an error.

core/src/main/scala/kafka/tools/TestRaftServer.scala
hachikuji: Moved the import inside `TestRaftServer`.

core/src/main/scala/kafka/tools/TestRaftServer.scala
hachikuji: You are right that the appends may still be committed, but in this patch, the `handleCommit` API is only invoked for appends within the current epoch. I thought it seemed simpler for now to just reset state at the start of the epoch. We can be more clever in the future once `handleCommit` is extended to account for replication.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Yes, I agree with you. Of course it is ok if unflushed data gets replicated. The main thing we need to protect is incrementing the high watermark.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Yeah, see my comment above about the handling of `Long.MAX_VALUE`. This is an attempt to reduce the error handling in the state machine. The model that we are working toward here is the following:

1) the state machine gets notified that the node has become leader in some epoch
2) the state machine can schedule appends with this epoch and it will get back the expected append offset
3) the state machine treats scheduled appends as uncommitted until the call to `handleCommit`
4) if the node resigns its leadership, the state machine will get notified and it will be expected to drop uncommitted data

By using a sort of impossible offset sentinel, the state machine just needs to wait for the notification that the leader has resigned.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
hachikuji: I am not sure. We don't have a metric yet, so I thought we might as well start with a thread-safe implementation. If we add a metric in the future, we can probably use an `AtomicInteger` or something and eliminate the locking.

raft/src/main/java/org/apache/kafka/raft/internals/BatchBuilder.java
hachikuji: I am not sure I follow. Are you suggesting making `BatchBuilder` a nested class? I think what it's doing is complex enough that I wanted a separate class that could be tested in isolation.

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
hachikuji: This was actually a bug. The implementation was treating `correlationId` as unique across all connections, which of course was wrong. My fix was just to overwrite the `correlationId` from the header with one that could be unique, but obviously this loses traceability through the Raft layer. If it's ok with you, I'd like to address this problem more generally in a follow-up.

core/src/main/scala/kafka/common/RecordValidationException.scala
hachikuji: Yeah, doesn't need to be done here. I just noticed the trace was missing when debugging a failure.

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
jsancio: Currently, `postInboundRequest` in only used by `TestRaftRequestHandler` and `KafkaNetworkChannelTest`. Are you thinking that we will use the same or similar method when integrating with the broker code?

jsancio: Unrelated with this change but when reading the code. I was confused by the name `pollInboundResponses`. That function returns both pending request and responses.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -0,0 +1,294 @@
...
+     * @param records the list of records to include in a batch
+     * @return the offset of the last message or {@link Long#MAX_VALUE} if the epoch
+     *         does not match
+     */
+    public Long append(int epoch, List<T> records) {
+        if (epoch != this.epoch) {
+            // If the epoch does not match, then the state machine probably
+            // has not gotten the notification about the latest epoch change.
+            // In this case, ignore the append and return a large offset value
+            // which will never be committed.
```
jsancio, line 99: What are you trying to protect with this check? For example, the signature could be `public long append(List<T> records)` with the accumulator writing the correct epoch.

In https://github.com/apache/kafka/pull/9482 you implemented `handleClaim` in to only fire when the `Listener`'s "acknowledged" offset + 1 is >= to the leader's epoch start offset.

Thinking through the code's behaviour, I see this check catching the case the the raft replica lost leadership and won leadership before the `Listener` was able to asynchronously process `handleResign` and `handleClaim`.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -0,0 +1,294 @@
...
+
+    /**
+     * Append a list of records into an atomic batch. We guarantee all records
+     * are included in the same underlying record batch so that either all of
+     * the records become committed or none of them do.
+     *
+     * @param epoch the expected leader epoch
+     * @param records the list of records to include in a batch
+     * @return the offset of the last message or {@link Long#MAX_VALUE} if the epoch
+     *         does not match
```
jsancio: We should document this returns `null` and what it means when `null` is returned.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -0,0 +1,294 @@
...
+                time.milliseconds(),
+                false,
+                RecordBatch.NO_PARTITION_LEADER_EPOCH,
+                maxBatchSize
+            );
+        }
+    }
+
+    /**
+     * Check whether there are any batches which need flushing now.
```
jsancio: Throughout this type we use flush to mean "collect the current batch" and start a new one. When I started reading this code I was assuming that it meant flush to disk/IO. Should we use a different verb for this? I think Java tends to use "collect" for this type of operation. What do you think?

raft/src/main/java/org/apache/kafka/raft/internals/BatchBuilder.java
jsancio, line 46: I think it would be good to have a description of what this type is doing and how to use it. Reading the code and seeing how the buffer is shared, I am under the impression that this type should not be reused after the `build` method is called. Is that correct?

If this is correct, I think all of the public methods should check that they are not being called after `build` is called.

raft/src/main/java/org/apache/kafka/raft/internals/BatchMemoryPool.java
```diff
@@ -0,0 +1,105 @@
...
+                buffer.limit(sizeBytes);
+            }
+            return buffer;
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    @Override
+    public void release(ByteBuffer previouslyAllocated) {
```
jsancio, line 65: Let's check that `previouslyAllocated`'s capacity is `batchSize` else `buffer.limit(sizeBytes)` is going to throw with a less useful stacktrace.

jsancio (commented):
Thanks for the PR @hachikuji.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
hachikuji: Right. It is important to ensure that the state machine has observed the latest leader epoch. Otherwise there may be committed data inflight which the state machine has yet to see.

core/src/main/scala/kafka/common/RecordValidationException.scala
rajinisivaram: I had only changed client-side code, so changes here wouldn't have conflicted.

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
hachikuji: > Unrelated with this change but when reading the code. I was confused by the name pollInboundResponses. That function returns both pending request and responses.

I don't think that's right. It only delivers responses for outbound requests (in other words, inbound responses).

In regard to the confusion, I took an unconventional turn in the beginning by consolidating inbound and outbound requests through a single API. This was useful because it allowed inputs and outputs to be serialized, which made it easier to have deterministic simulation tests. The main defect from the perspective of Raft is that fetch requests cannot be handled concurrently. I think we are going to have to address this problem in the long term, but it is outside the scope of this PR.

jsancio: > I don't think that's right. It only delivers responses for outbound requests (in other words, inbound responses).

Hmm. I misspoke a bit. I should have said that `pollInboundResponses` returns both inbound requests and responses. These are my observations:

`postInboundRequest` puts requests in `undelivered`:
https://github.com/apache/kafka/pull/9418/files#diff-c9b2a05e69be1d2391d91a33f408eab4fde1245d4ea9cf9f52305c8cb89208ceR224

`pollInboundResponses` returns everything in `undelivered`:
https://github.com/apache/kafka/pull/9418/files#diff-c9b2a05e69be1d2391d91a33f408eab4fde1245d4ea9cf9f52305c8cb89208ceR201-R202

The raft client gets all of the inbound messages (requests and reponses) by calling `receive` which in term calls `pollInboundResponses`:
https://github.com/apache/kafka/pull/9418/files#diff-1da15c51e641ea46ea5c86201ab8f21cfee9e7c575102a39c7bae0d5ffd7de39L1596

hachikuji: My mistake. This _was_ being used for both inbound requests and responses. Let me change the name and add some docs.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
hachikuji: Perhaps `drain` is a better verb in this context.

raft/src/main/java/org/apache/kafka/raft/internals/BatchBuilder.java
hachikuji: Sure, we can do that. I was being a bit looser with API validation since this is bound for internal use only.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
guozhangwang: Sounds good.

raft/src/main/java/org/apache/kafka/raft/internals/BatchBuilder.java
guozhangwang: I think the pairs of those classes:

1) BatchAccumulator and RecordAccumulator
2) BatchMemoryPool and BufferPool
3) BatchBuilder and MemoryRecordsBuilder

are similar can maybe can be consolidated to a single class each.

hachikuji: Yes, that is fair. Initially I had planned to reuse as much as possible, but the existing classes had too much baggage. I wanted implementations which were simple and efficient for the needs of the raft layer.

guozhangwang:
LGTM.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -276,6 +299,17 @@ private void onBecomeLeader(long currentTimeMs) {
         resetConnections();
 
         kafkaRaftMetrics.maybeUpdateElectionLatency(currentTimeMs);
+
+        accumulator = new BatchAccumulator<>(
```
jsancio, line 303: @hachikuji Should we move this state to `LeaderState`? We can delegate `maybeCloseAccumulator` to the `transitionTo...` functions.

I have a similar requirements for snapshot. I am currently adding a `SnapshotWriter` to `FollowerState` to track this state. I haven't implemented it yet but I am thinking of making `FollowerState` `Closeable` which will handle the cleanup. The `transitionTo...` functions have the additional responsibility of calling `close` if necessary.

What do you think?

PR #9420, opened 2020-10-13
KAFKA-10604: The StreamsConfig.STATE_DIR_CONFIG's default value does not reflect the JVM parameter or OS-specific settings

dongjinleekr:
Make the default state store directory location to follow OS-specific temporary directory settings or `java.io.tmpdir` JVM parameter, with `Utils#getTempDir`.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
+++ b/streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
@@ -45,6 +45,7 @@ import org.apache.kafka.streams.processor.internals.StreamsPartitionAssignor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.HashMap;
@@ -604,7 +605,7 @@ public class StreamsConfig extends AbstractConfig {
                     REPLICATION_FACTOR_DOC)
             .define(STATE_DIR_CONFIG,
                     Type.STRING,
-                    "/tmp/kafka-streams",
+                    System.getProperty("java.io.tmpdir") + File.separator + "kafka-streams",
                     Importance.HIGH,
                     STATE_DIR_DOC)
 
--- a/core/src/main/scala/kafka/tools/StreamsResetter.java
+++ b/core/src/main/scala/kafka/tools/StreamsResetter.java
@@ -119,7 +119,7 @@ public class StreamsResetter {
             + "* This tool will not clean up the local state on the stream application instances (the persisted "
             + "stores used to cache aggregation results).\n"
             + "You need to call KafkaStreams#cleanUp() in your application or manually delete them from the "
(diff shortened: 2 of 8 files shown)


mjsax (approved):
Sorry for dropping this on the floor. I guess it looks good. However, the PR needs a rebase.

streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
```diff
@@ -602,7 +603,7 @@
                     REPLICATION_FACTOR_DOC)
             .define(STATE_DIR_CONFIG,
                     Type.STRING,
-                    "/tmp/kafka-streams",
+                    Utils.getTempDir().resolve("kafka-streams").toAbsolutePath().toString(),
```
vvcephei: Thanks for this @dongjinleekr !

It's kind of a bummer that we actually create this file while defining the configurations, since it results in leaking temp directories, even if users set this config to a different value.

Do you think we can do this in a lazy fashion?

mjsax: Good point John. Maybe we should just use `System.getProperty("java.io.tmpdir")` here?

We have already code in place in `StateDirectory` constructor to create the `./kafka-streams` sub-directory. Might be good enough without any additional changes?

dongjinleekr: @vvcephei @mjsax Totally agree. Since I am now finalizing KAFKA-8403 now, I will fix this with a integration test after that. 🙏

dongjinleekr:
@vvcephei @mjsax Fixed, with rebasing onto the latest trunk and validataion. (please see `StreamsConfigTest.shouldStateDirStartsWithJavaIOTmpDir`.)

mjsax:
@dongjinleekr -- the PR shows merge conflicts. Can you rebase once more. Sorry about that.

vvcephei:
Hey @dongjinleekr ,

Sorry for the force-push, but I had to rebase this and resolve a conflict before merging. Note that the conflict was from 462c89e0b436abd56864bea8bbcaf1ab70b7f66e, which re-organized the boolean conditions in the StateDirectory constructor, specifically where we warn if the state dir is a temp dir.

After resolving the conflict, I noticed there's no test for that warning, so I added one to be sure it works. It also looked like the temp dir check could actually be a bit simpler, so I just tweaked it rather than leaving a new comment for you to address.

I hope this is all ok. I'll let the tests run and merge in the morning, unless you have any objections.

Thanks!
-John

vvcephei:
Ah, a checkstyle violation. How embarrassing. Just fixed it.

vvcephei:
Only one flaky test failure:
```
 Build / JDK 11 / org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationSSLTest.testOneWayReplicationWithAutoOffsetSync()
```

Proceeding to merge...

vvcephei:
Merged!

Many thanks, @dongjinleekr !

PR #9422, opened 2020-10-13
KAFKA-10602: Make RetryWithToleranceOperator thread safe

tombentley:
ErrantRecordReporter uses a RetryWithToleranceOperator instance, which is necessarily
stateful, having a ProcessingContext of which there's supposed to be one per
task. That ProcessingContext is used by both
RetryWithToleranceOperator.executeFailed() and execute(), so it's not enough
to just synchronize executeFailed().

So make all public methods of RetryWithToleranceOperator synchronized so that
RetryWithToleranceOperator is now threadsafe.

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/errors/RetryWithToleranceOperator.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/errors/RetryWithToleranceOperator.java
@@ -51,6 +51,9 @@ import java.util.concurrent.ThreadLocalRandom;
  * set appropriate error reason in the {@link ProcessingContext} and return null, or (2) if the exception is not tolerated,
  * then it is wrapped into a ConnectException and rethrown to the caller.
  * <p>
+ *
+ * Instances of this class are thread safe.
+ * <p>
  */
 public class RetryWithToleranceOperator implements AutoCloseable {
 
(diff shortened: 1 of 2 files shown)


tombentley:
@aakashnshah @rhauch @kkonstantine please could one of you review?

kkonstantine (commented):
Thanks for the quick fix @tombentley 

I wonder if a test with two threads calling different methods could reproduce the issue and verify the fix. It would have to fail consistently of course, but I could see this happening for several concurrent iterations. Haven't explored this more, just a thought for now.

tombentley:
@kkonstantine yeah, I was hoping someone would recommend a testing approach. I'll see if I'm able to reproduce it in a test.

tombentley:
@kkonstantine I've written a small test which interleaves calls to `RetryWithToleranceOperator.execute` and `RetryWithToleranceOperator.executeFailed` and detects conflicts which can arise when those methods are not `synchronized`. It fails reliable for me when they're not `synchronized` and passes reliably when they are.

kkonstantine (commented):
Thanks @tombentley 
I like the test. Let's fix the checkstyle errors in the test (alignment of lambdas is a bit tricky) and I'll take a final look.

tombentley:
@kkonstantine thanks for taking a look, now sorted.

kkonstantine (approved):
That's a good test. I confirmed it fails in a straightforward way if the methods are not synchronized properly. 

Thanks for the fix @tombentley 
LGTM

kkonstantine:
The single failure in JDK 8 is unrelated.

PR #9425, opened 2020-10-13
KAFKA-10600: Connect should not add error to connector validation values for properties not in connector’s ConfigDef

rhauch:
By default the `Connector.validate(...)` will only include configuration values that are defined in the connector's `ConfigDef`, and validation is handled via `ConfigDef.Validator` objects. However, a connector can override the `Connector.validate(...)`, including in the output any configuration values whether or not they are not defined in the connector's `ConfigDef`.

Before this change, Connect always added an error to any configuration value that did not have a corresponding key defined in the connector's `ConfigDef`. Also, any errors on configuration values that did not have a corresponding key defined in the connector's `ConfigDef` was not counted in the total # of errors in the validation result.

This change removes the addition of those errors, so that only the errors returned by the connector's validation are included in the output, and all errors returned by the connector's validation output are counted in the total # of errors.

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/AbstractHerder.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/AbstractHerder.java
@@ -502,8 +502,8 @@ public abstract class AbstractHerder implements Herder, TaskStatus.Listener, Con
             String configName = configValue.name();
             configValueMap.put(configName, configValue);
             if (!configKeys.containsKey(configName)) {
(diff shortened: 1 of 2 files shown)


kkonstantine (commented):
The change makes sense to me @rhauch 
We seem to have a few tests for `generateResult`. Would it make sense to add one to check for this case too?

rhauch:
@kkonstantine thanks for the suggestion. I rebased and added a commit with more unit tests for `AbstractHerder.generateResult(...)`.

kkonstantine (approved):
LGTM
Thanks @rhauch

PR #9428, opened 2020-10-13
MINOR: fix a bug in removing elements from an ImplicitLinkedHashColle…

cmccabe:
Fix a bug that was introduced by change 86013dc9f8cf02813418f41c3c49c745f9ee6ea5 that resulted in incorrect behavior when deleting through an iterator.

The bug is that the hash table relies on a denseness invariant... if you remove something, you might have to move some other things.  Calling removeElementAtSlot will do this.  Calling removeFromList is not enough.

### DIFF
--- a/clients/src/test/java/org/apache/kafka/common/utils/ImplicitLinkedHashCollectionTest.java
+++ b/clients/src/test/java/org/apache/kafka/common/utils/ImplicitLinkedHashCollectionTest.java
@@ -32,7 +32,6 @@ import java.util.Set;
 
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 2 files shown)


clients/src/test/java/org/apache/kafka/common/utils/ImplicitLinkedHashCollectionTest.java
```diff
@@ -596,4 +598,25 @@ public void testMoveToEnd() {
...
+        }
+        assertEquals(100, coll.size());
+        Iterator<TestElement> iter = coll.iterator();
+        for (int i = 0; i < 50; i++) {
+            iter.next();
+            iter.remove();
+        }
+        assertEquals(50, coll.size());
+        for (int i = 50; i < 100; i++) {
+            assertNotNull(coll.find(elements.get(i)));
```
hachikuji: Could make this assertion a little stronger: 
```
            assertEquals(new TestElement(i, i), coll.find(elements.get(i)));
```

hachikuji (approved):
Minor suggestion. No need for re-review.

cmccabe:
Backported to 2.7 as well.

PR #9430, opened 2020-10-14
KAFKA-5235: GetOffsetShell: support for multiple topics and consumer configuration override

urbandan:
Implements KIP-635

Changes:

- Added kafka-get-offsets.sh script
- Removed deprecated max-wait-ms and offsets arguments
- Updated tool to query all topic-partitions by default
- Updated topic argument to support patterns
- Added topic-partitions argument to support a list of topic-partition patterns
- Added exclude-internal-topics to support filtering internal topics

Testing done: added new ducktape tests for the tool.

### DIFF
--- a/core/src/main/scala/kafka/tools/GetOffsetShell.scala
+++ b/core/src/main/scala/kafka/tools/GetOffsetShell.scala
@@ -43,27 +43,25 @@ object GetOffsetShell {
                            .withRequiredArg
                            .describedAs("HOST1:PORT1,...,HOST3:PORT3")
                            .ofType(classOf[String])
-    val topicPartitionOpt = parser.accepts("topic-partitions", "Comma separated list of topic-partition specifications to get the offsets for, with the format of topic:partition. The 'topic' part can be a regex or may be omitted to only specify the partitions, and query all authorized topics." +
+    val topicPartitionsOpt = parser.accepts("topic-partitions", "Comma separated list of topic-partition specifications to get the offsets for, with the format of topic:partition. The 'topic' part can be a regex or may be omitted to only specify the partitions, and query all authorized topics." +
                                             " The 'partition' part can be: a number, a range in the format of 'NUMBER-NUMBER' (lower inclusive, upper exclusive), an inclusive lower bound in the format of 'NUMBER-', an exclusive upper bound in the format of '-NUMBER' or may be omitted to accept all partitions of the specified topic.")
                            .withRequiredArg
                            .describedAs("topic:partition,...,topic:partition")
                            .ofType(classOf[String])
-                           .defaultsTo("")
-    val topicOpt = parser.accepts("topic", s"The topic to get the offsets for. It also accepts a regular expression. If not present, all authorized topics are queried. Ignored if $topicPartitionOpt is present.")
+    val topicOpt = parser.accepts("topic", s"The topic to get the offsets for. It also accepts a regular expression. If not present, all authorized topics are queried. Cannot be used if --topic-partitions is present.")
                            .withRequiredArg
                            .describedAs("topic")
                            .ofType(classOf[String])
-    val partitionOpt = parser.accepts("partitions", s"Comma separated list of partition ids to get the offsets for. If not present, all partitions of the authorized topics are queried. Ignored if $topicPartitionOpt is present.")
+    val partitionsOpt = parser.accepts("partitions", s"Comma separated list of partition ids to get the offsets for. If not present, all partitions of the authorized topics are queried. Cannot be used if --topic-partitions is present.")
                            .withRequiredArg
                            .describedAs("partition ids")
                            .ofType(classOf[String])
-                           .defaultsTo("")
-    val timeOpt = parser.accepts("time", "timestamp of the offsets before that. [Note: No offset is returned, if the timestamp greater than recently commited record timestamp is given.]")
+    val timeOpt = parser.accepts("time", "timestamp of the offsets before that. [Note: No offset is returned, if the timestamp greater than recently committed record timestamp is given.]")
                            .withRequiredArg
                            .describedAs("timestamp/-1(latest)/-2(earliest)")
                            .ofType(classOf[java.lang.Long])
                            .defaultsTo(-1L)
-    val commandConfigOpt = parser.accepts("command-config", s"Consumer config properties file.")
+    val commandConfigOpt = parser.accepts("command-config", s"Property file containing configs to be passed to Consumer Client.")
                            .withRequiredArg
                            .describedAs("config file")
                            .ofType(classOf[String])
@@ -87,9 +85,14 @@ object GetOffsetShell {
     ToolsUtils.validatePortOrDie(parser, brokerList)
     val excludeInternalTopics = options.has(excludeInternalTopicsOpt)
 
+    if (options.has(topicPartitionsOpt) && (options.has(topicOpt) || options.has(partitionsOpt))) {
+      System.err.println(s"--topic-partitions cannot be used with --topic or --partitions")
+      Exit.exit(1)
+    }
+
     val partitionIdsRequested: Set[Int] = {
-      val partitionsString = options.valueOf(partitionOpt)
-      if (partitionsString.isEmpty)
+      val partitionsString = options.valueOf(partitionsOpt)
+      if (partitionsString == null || partitionsString.isEmpty)
         Set.empty
       else
         partitionsString.split(",").map { partitionString =>
@@ -103,8 +106,8 @@ object GetOffsetShell {
     }
     val listOffsetsTimestamp = options.valueOf(timeOpt).longValue
 
-    val topicPartitionFilter = if (options.has(topicPartitionOpt)) {
-      Some(createTopicPartitionFilterWithPatternList(options.valueOf(topicPartitionOpt), excludeInternalTopics))
+    val topicPartitionFilter = if (options.has(topicPartitionsOpt)) {
+      createTopicPartitionFilterWithPatternList(options.valueOf(topicPartitionsOpt), excludeInternalTopics)
     } else {
       createTopicPartitionFilterWithTopicAndPartitionPattern(
         if (options.has(topicOpt)) Some(options.valueOf(topicOpt)) else None,
@@ -128,13 +131,7 @@ object GetOffsetShell {
       Exit.exit(1)
     }
 
-    val topicPartitions = partitionInfos.sortWith((tp1, tp2) => {
-      val topicComp = tp1.topic.compareTo(tp2.topic)
-      if(topicComp == 0)
-        tp1.partition < tp2.partition
-      else
-        topicComp < 0
-    }).flatMap { p =>
+    val topicPartitions = partitionInfos.flatMap { p =>
       if (p.leader == null) {
         System.err.println(s"Error: topic-partition ${p.topic}:${p.partition} does not have a leader. Skip getting offsets")
         None
@@ -162,7 +159,6 @@ object GetOffsetShell {
     }).foreach { case (tp, offset) =>
       println(s"${tp.topic}:${tp.partition}:${Option(offset).getOrElse("")}")
     }
-
   }
 
   /**
@@ -173,26 +169,28 @@ object GetOffsetShell {
    * TopicPattern: REGEX
    * PartitionPattern: NUMBER | NUMBER-(NUMBER)? | -NUMBER
    */
-  private def createTopicPartitionFilterWithPatternList(topicPartitions: String, excludeInternalTopics: Boolean): PartitionInfo => Boolean = {
+  def createTopicPartitionFilterWithPatternList(topicPartitions: String, excludeInternalTopics: Boolean): PartitionInfo => Boolean = {
     val ruleSpecs = topicPartitions.split(",")
-    val rules = ruleSpecs.map(ruleSpec => {
+    val rules = ruleSpecs.map { ruleSpec =>
       val parts = ruleSpec.split(":")
       if (parts.length == 1) {
-        val whitelist = IncludeList(parts(0))
-        tp: PartitionInfo => whitelist.isTopicAllowed(tp.topic, excludeInternalTopics)
+        val includeList = IncludeList(parts(0))
+        tp: PartitionInfo => includeList.isTopicAllowed(tp.topic, excludeInternalTopics)
       } else if (parts.length == 2) {
-        val partitionFilter = createPartitionFilter(parts(1))
+        try {
+          val partitionFilter = createPartitionFilter(parts(1))
+          val includeList = if (parts(0).trim().isEmpty) IncludeList(".*") else IncludeList(parts(0))
 
-        if (parts(0).trim().isEmpty) {
-          tp: PartitionInfo => partitionFilter.apply(tp.partition)
-        } else {
-          val whitelist = IncludeList(parts(0))
-          tp: PartitionInfo => whitelist.isTopicAllowed(tp.topic, excludeInternalTopics) && partitionFilter.apply(tp.partition)
+          tp: PartitionInfo => includeList.isTopicAllowed(tp.topic, excludeInternalTopics) && partitionFilter.apply(tp.partition)
+        } catch {
+          case e: IllegalArgumentException =>
+            throw new IllegalArgumentException(s"Invalid rule '$ruleSpec'; Issue: ${e.getMessage}")
         }
+
       } else {
         throw new IllegalArgumentException(s"Invalid topic-partition rule: $ruleSpec")
       }
-    })
+    }
 
     tp => rules.exists(rule => rule.apply(tp))
   }
@@ -202,70 +200,54 @@ object GetOffsetShell {
    * Expected format:
    * PartitionPattern: NUMBER | NUMBER-(NUMBER)? | -NUMBER
    */
-  private def createPartitionFilter(spec: String): Int => Boolean = {
-    if (spec.indexOf('-') != -1) {
-      val rangeParts = spec.split("-", -1)
-      if(rangeParts.length != 2 || rangeParts(0).isEmpty && rangeParts(1).isEmpty) {
-        throw new IllegalArgumentException(s"Invalid range specification: $spec")
-      }
-
-      if(rangeParts(0).isEmpty) {
-        val max = rangeParts(1).toInt
-        partition: Int => partition < max
-      } else if(rangeParts(1).isEmpty) {
-        val min = rangeParts(0).toInt
-        partition: Int => partition >= min
-      } else {
-        val min = rangeParts(0).toInt
-        val max = rangeParts(1).toInt
-
-        if (min > max) {
-          throw new IllegalArgumentException(s"Range lower bound cannot be greater than upper bound: $spec")
+  def createPartitionFilter(spec: String): Int => Boolean = {
+    try {
+      if (spec.indexOf('-') != -1) {
+        val rangeParts = spec.split("-", -1)
+        if (rangeParts.length != 2 || rangeParts(0).isEmpty && rangeParts(1).isEmpty) {
+          throw new IllegalArgumentException(s"Invalid range specification: $spec")
         }
 
-        partition: Int => partition >= min && partition < max
+        if (rangeParts(0).isEmpty) {
+          val max = rangeParts(1).toInt
+          partition: Int => partition < max
+        } else if (rangeParts(1).isEmpty) {
+          val min = rangeParts(0).toInt
+          partition: Int => partition >= min
+        } else {
+          val min = rangeParts(0).toInt
+          val max = rangeParts(1).toInt
+
+          if (min > max) {
+            throw new IllegalArgumentException(s"Range lower bound cannot be greater than upper bound: $spec")
+          }
+
+          partition: Int => partition >= min && partition < max
+        }
+      } else {
+        val number = spec.toInt
+        partition: Int => partition == number
       }
-    } else {
-      val number = spec.toInt
-      partition: Int => partition == number
+    } catch {
+      case _: NumberFormatException =>
+        throw new IllegalArgumentException(s"Expected a number in partition specification: $spec")
     }
   }
 
   /**
    * Creates a topic-partition filter based on a topic pattern and a set of partition ids.
    */
-  private def createTopicPartitionFilterWithTopicAndPartitionPattern(topicOpt: Option[String], excludeInternalTopics: Boolean, partitionIds: Set[Int]): Option[PartitionInfo => Boolean] = {
-    topicOpt match {
-      case Some(topic) =>
-        val topicsFilter = IncludeList(topic)
-        if(partitionIds.isEmpty)
-          Some(t => topicsFilter.isTopicAllowed(t.topic, excludeInternalTopics))
-        else
-          Some(t => topicsFilter.isTopicAllowed(t.topic, excludeInternalTopics) && partitionIds.contains(t.partition))
-      case None =>
-        if(excludeInternalTopics) {
-          if(partitionIds.isEmpty)
-            Some(t => !Topic.isInternal(t.topic))
-          else
-            Some(t => !Topic.isInternal(t.topic) && partitionIds.contains(t.partition))
-        } else {
-          if(partitionIds.isEmpty)
-            None
-          else
-            Some(t => partitionIds.contains(t.partition))
-        }
-    }
+  def createTopicPartitionFilterWithTopicAndPartitionPattern(topicOpt: Option[String], excludeInternalTopics: Boolean, partitionIds: Set[Int]): PartitionInfo => Boolean = {
+    val topicsFilter = IncludeList(topicOpt.getOrElse(".*"))
+    t => topicsFilter.isTopicAllowed(t.topic, excludeInternalTopics) && (partitionIds.isEmpty || partitionIds.contains(t.partition))
   }
 
   /**
-   * Return the partition infos. Filter them with topicPartitionFilter if specified.
+   * Return the partition infos. Filter them with topicPartitionFilter.
    */
-  private def listPartitionInfos(consumer: KafkaConsumer[_, _], topicPartitionFilter: Option[PartitionInfo => Boolean]): Seq[PartitionInfo] = {
-    val topicListUnfiltered = consumer.listTopics.asScala.values.flatMap { tp => tp.asScala }
-    val topicList = topicPartitionFilter match {
-      case Some(filter) => topicListUnfiltered.filter { tp => filter.apply(tp) }
-      case _ => topicListUnfiltered
-    }
-    topicList.toBuffer
+  private def listPartitionInfos(consumer: KafkaConsumer[_, _], topicPartitionFilter: PartitionInfo => Boolean): Seq[PartitionInfo] = {
+    consumer.listTopics.asScala.values.flatMap { partitions =>
+      partitions.asScala.filter { tp => topicPartitionFilter.apply(tp) }
+    }.toBuffer
   }
 }
--- a/core/src/main/scala/kafka/tools/GetOffsetShell.scala
+++ b/core/src/main/scala/kafka/tools/GetOffsetShell.scala
@@ -20,19 +20,30 @@ package kafka.tools
 
 import java.util.Properties
 import joptsimple._
-import kafka.utils.{CommandLineUtils, Exit, ToolsUtils, IncludeList}
+import kafka.utils.{CommandLineUtils, Exit, IncludeList, ToolsUtils}
 import org.apache.kafka.clients.consumer.{ConsumerConfig, KafkaConsumer}
 import org.apache.kafka.common.requests.ListOffsetsRequest
 import org.apache.kafka.common.{PartitionInfo, TopicPartition}
 import org.apache.kafka.common.serialization.ByteArrayDeserializer
 import org.apache.kafka.common.utils.Utils
 
+import java.util.regex.Pattern
 import scala.jdk.CollectionConverters._
 import scala.collection.Seq
+import scala.math.Ordering.Implicits.infixOrderingOps
 
 object GetOffsetShell {
+  private val topicPartitionPattern = Pattern.compile( "([^:,]*)(?::([0-9]*)(?:-([0-9]*))?)?")
 
   def main(args: Array[String]): Unit = {
+    try {
+      fetchOffsets(args)
+    } catch {
+      case e: Exception => Exit.exit(1, Some(e.getMessage))
+    }
+  }
+
+  private def fetchOffsets(args: Array[String]): Unit = {
     val parser = new OptionParser(false)
     val brokerListOpt = parser.accepts("broker-list", "DEPRECATED, use --bootstrap-server instead; ignored if --bootstrap-server is specified. The server(s) to connect to in the form HOST1:PORT1,HOST2:PORT2.")
                            .withRequiredArg
@@ -43,10 +54,11 @@ object GetOffsetShell {
                            .withRequiredArg
                            .describedAs("HOST1:PORT1,...,HOST3:PORT3")
                            .ofType(classOf[String])
-    val topicPartitionsOpt = parser.accepts("topic-partitions", "Comma separated list of topic-partition specifications to get the offsets for, with the format of topic:partition. The 'topic' part can be a regex or may be omitted to only specify the partitions, and query all authorized topics." +
-                                            " The 'partition' part can be: a number, a range in the format of 'NUMBER-NUMBER' (lower inclusive, upper exclusive), an inclusive lower bound in the format of 'NUMBER-', an exclusive upper bound in the format of '-NUMBER' or may be omitted to accept all partitions of the specified topic.")
+    val topicPartitionsOpt = parser.accepts("topic-partitions", s"Comma separated list of topic-partition patterns to get the offsets for, with the format of '$topicPartitionPattern'." +
+                                            " The first group is an optional regex for the topic name, if omitted, it matches any topic name." +
+                                            " The section after ':' describes a 'partition' pattern, which can be: a number, a range in the format of 'NUMBER-NUMBER' (lower inclusive, upper exclusive), an inclusive lower bound in the format of 'NUMBER-', an exclusive upper bound in the format of '-NUMBER' or may be omitted to accept all partitions.")
                            .withRequiredArg
-                           .describedAs("topic:partition,...,topic:partition")
+                           .describedAs("topic1:1,topic2:0-3,topic3,topic4:5-,topic5:-3")
                            .ofType(classOf[String])
     val topicOpt = parser.accepts("topic", s"The topic to get the offsets for. It also accepts a regular expression. If not present, all authorized topics are queried. Cannot be used if --topic-partitions is present.")
                            .withRequiredArg
@@ -86,29 +98,29 @@ object GetOffsetShell {
     val excludeInternalTopics = options.has(excludeInternalTopicsOpt)
 
     if (options.has(topicPartitionsOpt) && (options.has(topicOpt) || options.has(partitionsOpt))) {
-      System.err.println(s"--topic-partitions cannot be used with --topic or --partitions")
-      Exit.exit(1)
+      throw new IllegalArgumentException("--topic-partitions cannot be used with --topic or --partitions")
     }
 
-    val partitionIdsRequested: Set[Int] = {
-      val partitionsString = options.valueOf(partitionsOpt)
-      if (partitionsString == null || partitionsString.isEmpty)
-        Set.empty
-      else
-        partitionsString.split(",").map { partitionString =>
-          try partitionString.toInt
-          catch {
-            case _: NumberFormatException =>
-              System.err.println(s"--partitions expects a comma separated list of numeric partition ids, but received: $partitionsString")
-              Exit.exit(1)
-          }
-        }.toSet
-    }
     val listOffsetsTimestamp = options.valueOf(timeOpt).longValue
 
     val topicPartitionFilter = if (options.has(topicPartitionsOpt)) {
       createTopicPartitionFilterWithPatternList(options.valueOf(topicPartitionsOpt), excludeInternalTopics)
     } else {
+      val partitionIdsRequested: Set[Int] = {
+        val partitionsString = options.valueOf(partitionsOpt)
+        if (partitionsString == null || partitionsString.isEmpty)
+          Set.empty
+        else
+          partitionsString.split(",").map { partitionString =>
+            try partitionString.toInt
+            catch {
+              case _: NumberFormatException =>
+                throw new IllegalArgumentException(s"--partitions expects a comma separated list of numeric " +
+                  s"partition ids, but received: $partitionsString")
+            }
+          }.toSet
+      }
+
       createTopicPartitionFilterWithTopicAndPartitionPattern(
         if (options.has(topicOpt)) Some(options.valueOf(topicOpt)) else None,
         excludeInternalTopics,
@@ -127,8 +139,7 @@ object GetOffsetShell {
     val partitionInfos = listPartitionInfos(consumer, topicPartitionFilter)
 
     if (partitionInfos.isEmpty) {
-      System.err.println(s"Could not match any topic-partitions with the specified filters")
-      Exit.exit(1)
+      throw new IllegalArgumentException("Could not match any topic-partitions with the specified filters")
     }
 
     val topicPartitions = partitionInfos.flatMap { p =>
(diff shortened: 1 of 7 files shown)


dajac:
@urbandan Thanks for the PR. I haven't looked at it yet but I have already noticed that there are no unit/integration tests. We need some. You can look at how other commands are tested. Having ducktape tests is very good but not enough to catch bugs and regressions.

urbandan:
@dajac Thanks for the reminder, pushed a bunch of unit tests covering the GetOffsetShell tool.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -33,15 +35,26 @@ object GetOffsetShell {
...
+                           .requiredUnless("broker-list")
+                           .withRequiredArg
+                           .describedAs("HOST1:PORT1,...,HOST3:PORT3")
+                           .ofType(classOf[String])
+    val topicPartitionOpt = parser.accepts("topic-partitions", "Comma separated list of topic-partition specifications to get the offsets for, with the format of topic:partition. The 'topic' part can be a regex or may be omitted to only specify the partitions, and query all authorized topics." +
+                                            " The 'partition' part can be: a number, a range in the format of 'NUMBER-NUMBER' (lower inclusive, upper exclusive), an inclusive lower bound in the format of 'NUMBER-', an exclusive upper bound in the format of '-NUMBER' or may be omitted to accept all partitions of the specified topic.")
+                           .withRequiredArg
+                           .describedAs("topic:partition,...,topic:partition")
+                           .ofType(classOf[String])
+                           .defaultsTo("")
```
dajac: Is having empty string as default necessary? It shows as `(default: )` in the description which looks weird.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -33,15 +35,26 @@ object GetOffsetShell {
...
+    val topicOpt = parser.accepts("topic", s"The topic to get the offsets for. It also accepts a regular expression. If not present, all authorized topics are queried. Ignored if $topicPartitionOpt is present.")
```
dajac: We should replace `$topicPartitionOpt` by `--topic-partitions` in order to remain consistent with the other descriptions. `$topicPartitionOpt` generates `[topic-partitions]` in the description.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -33,15 +35,26 @@ object GetOffsetShell {
...
                            .withRequiredArg
                            .describedAs("topic")
                            .ofType(classOf[String])
-    val partitionOpt = parser.accepts("partitions", "comma separated list of partition ids. If not specified, it will find offsets for all partitions")
+    val partitionOpt = parser.accepts("partitions", s"Comma separated list of partition ids to get the offsets for. If not present, all partitions of the authorized topics are queried. Ignored if $topicPartitionOpt is present.")
```
dajac: ditto

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -51,28 +64,30 @@ object GetOffsetShell {
                            .describedAs("timestamp/-1(latest)/-2(earliest)")
                            .ofType(classOf[java.lang.Long])
                            .defaultsTo(-1L)
-    parser.accepts("offsets", "DEPRECATED AND IGNORED: number of offsets returned")
-                           .withRequiredArg
-                           .describedAs("count")
-                           .ofType(classOf[java.lang.Integer])
-                           .defaultsTo(1)
-    parser.accepts("max-wait-ms", "DEPRECATED AND IGNORED: The max amount of time each fetch request waits.")
+    val commandConfigOpt = parser.accepts("command-config", s"Consumer config properties file.")
```
dajac: How about `Property file containing configs to be passed to Consumer Client.` to remain inline with description of that field used by other command line tools?

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -51,28 +64,30 @@ object GetOffsetShell {
...
-                           .describedAs("ms")
-                           .ofType(classOf[java.lang.Integer])
-                           .defaultsTo(1000)
+                           .describedAs("config file")
+                           .ofType(classOf[String])
+    val excludeInternalTopicsOpt = parser.accepts("exclude-internal-topics", s"By default, internal topics are included. If specified, internal topics are excluded.")
 
-   if (args.length == 0)
-      CommandLineUtils.printUsageAndDie(parser, "An interactive shell for getting topic offsets.")
+    if (args.length == 0)
```
dajac, line 84: Should we also verify combination of arguments that are not valid? I am thinking about `--topic-partitions` vs `--topic` and `--partitions`. That would be better than silently ignoring.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +154,119 @@ object GetOffsetShell {
...
-    val partitionInfos = consumer.listTopics.asScala.filter { case (k, _) => k == topic }.values.flatMap(_.asScala).toBuffer
-    if (partitionInfos.isEmpty)
-      None
-    else if (partitionIds.isEmpty)
-      Some(partitionInfos)
-    else
-      Some(partitionInfos.filter(p => partitionIds.contains(p.partition)))
+  private def createTopicPartitionFilterWithPatternList(topicPartitions: String, excludeInternalTopics: Boolean): PartitionInfo => Boolean = {
+    val ruleSpecs = topicPartitions.split(",")
+    val rules = ruleSpecs.map(ruleSpec => {
```
dajac: nit: We usually structure blocks as follow:
```
ruleSpecs.map { ruleSpec => 
  ...
}
```
We can avoid the parenthesis.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +154,119 @@ object GetOffsetShell {
...
+      val parts = ruleSpec.split(":")
+      if (parts.length == 1) {
+        val whitelist = IncludeList(parts(0))
```
dajac: Let's try to remain inline with https://cwiki.apache.org/confluence/display/KAFKA/KIP-629%3A+Use+racially+neutral+terms+in+our+codebase. We should name the variable `includeList`.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +154,119 @@ object GetOffsetShell {
...
+        val max = rangeParts(1).toInt
+
+        if (min > max) {
+          throw new IllegalArgumentException(s"Range lower bound cannot be greater than upper bound: $spec")
+        }
+
+        partition: Int => partition >= min && partition < max
+      }
+    } else {
+      val number = spec.toInt
```
dajac: Should we handle the case where it is not a number and provide a meaning full error to the user?

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +154,119 @@ object GetOffsetShell {
...
+
+  /**
+   * Creates a partition filter based on a single id or a range.
+   * Expected format:
+   * PartitionPattern: NUMBER | NUMBER-(NUMBER)? | -NUMBER
+   */
+  private def createPartitionFilter(spec: String): Int => Boolean = {
+    if (spec.indexOf('-') != -1) {
+      val rangeParts = spec.split("-", -1)
+      if(rangeParts.length != 2 || rangeParts(0).isEmpty && rangeParts(1).isEmpty) {
```
dajac: nit: A space is missing after `if`. There are couple of similar cases below.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +154,119 @@ object GetOffsetShell {
...
+            Some(t => !Topic.isInternal(t.topic))
+          else
+            Some(t => !Topic.isInternal(t.topic) && partitionIds.contains(t.partition))
+        } else {
+          if(partitionIds.isEmpty)
+            None
+          else
+            Some(t => partitionIds.contains(t.partition))
+        }
+    }
```
dajac: Could we simplify this block to something like the following?
```
val topicFilter = IncludeList(topicOpt.getOrElse(".*"))
t => topicsFilter.isTopicAllowed(t.topic, excludeInternalTopics)) && (partitionIds.isEmpty || partitionIds.contains(t.partition))
```

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +154,119 @@ object GetOffsetShell {
...
     }
 
-    partitionOffsets.toArray.sortBy { case (tp, _) => tp.partition }.foreach { case (tp, offset) =>
-      println(s"$topic:${tp.partition}:${Option(offset).getOrElse("")}")
+    partitionOffsets.toSeq.sortWith((tp1, tp2) => {
+      val topicComp = tp1._1.topic.compareTo(tp2._1.topic)
+      if (topicComp == 0)
+        tp1._1.partition < tp2._1.partition
+      else
+        topicComp < 0
```
dajac: This block of code is identical to the one use at L132. Could we extract the predicate in a function and use it in both cases?

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -89,33 +104,40 @@ object GetOffsetShell {
...
-      (partitionIdsRequested -- partitionInfos.map(_.partition)).foreach { partitionId =>
-        System.err.println(s"Error: partition $partitionId does not exist")
-      }
+    if (partitionInfos.isEmpty) {
+      System.err.println(s"Could not match any topic-partitions with the specified filters")
+      Exit.exit(1)
     }
 
-    val topicPartitions = partitionInfos.sortBy(_.partition).flatMap { p =>
+    val topicPartitions = partitionInfos.sortWith((tp1, tp2) => {
```
dajac: Do we really need to sort the partitions here? What are the benefits?

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +154,119 @@ object GetOffsetShell {
...
+            Some(t => partitionIds.contains(t.partition))
+        }
+    }
+  }
+
+  /**
+   * Return the partition infos. Filter them with topicPartitionFilter if specified.
+   */
+  private def listPartitionInfos(consumer: KafkaConsumer[_, _], topicPartitionFilter: Option[PartitionInfo => Boolean]): Seq[PartitionInfo] = {
+    val topicListUnfiltered = consumer.listTopics.asScala.values.flatMap { tp => tp.asScala }
```
dajac: It might be worth doing the filtering in the `flatMap` directly. It would avoid having the construct the full list and to filter it afterwards. What do you think?

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
```diff
@@ -0,0 +1,194 @@
...
+import org.junit.Assert.{assertEquals, assertTrue}
+import org.junit.{Before, Test}
+
+class GetOffsetShellTest extends KafkaServerTestHarness with Logging {
+  private val topicCount = 4
+  private val offsetTopicPartitionCount = 4
+  private val topicPattern = Pattern.compile("test.*")
+
+  override def generateConfigs: collection.Seq[KafkaConfig] = TestUtils.createBrokerConfigs(1, zkConnect)
+    .map(p => {
```
dajac: nit: `map(tp => {` => `map { tp =>`

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
```diff
@@ -0,0 +1,194 @@
...
+    consumer.subscribe(topicPattern)
+    consumer.poll(Duration.ofMillis(1000))
+    consumer.commitSync()
+    consumer.close()
+  }
+
+  @Test
+  def testNoFilterOptions(): Unit = {
+    val offsets = executeAndParse(Array())
+    assertTrue(expectedOffsets() sameElements offsets.filter(r => !isConsumerOffsetTopicPartition(r)))
```
dajac: It would be better to use `assertEquals` as well as it outputs the values when the test fails. The same applies to all the test it seems.

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
```diff
@@ -0,0 +1,194 @@
...
+    assertEquals(offsetTopicPartitionCount, offsets.count(isConsumerOffsetTopicPartition))
+  }
+
+  @Test
+  def testInternalExcluded(): Unit = {
+    val offsets = executeAndParse(Array("--exclude-internal-topics"))
+    assertTrue(expectedOffsets() sameElements offsets.filter(r => !isConsumerOffsetTopicPartition(r)))
```
dajac: I am not sure to understand why you filter the `offsets` with `!isConsumerOffsetTopicPartition`. In this case, we expect `--exclude-internal-topics` to already do this, no?

In general, if `--exclude-internal-topics` is not set, we should verify that internal topics are there as well. I wouldn't filter them out from the result set.

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
```diff
@@ -0,0 +1,194 @@
...
+  @Test
+  def testTopicPatternArg(): Unit = {
+    val offsets = executeAndParse(Array("--topic", "topic.*"))
+    assertTrue(expectedOffsets() sameElements offsets)
+  }
+
+  @Test
+  def testPartitionsArg(): Unit = {
+    val offsets = executeAndParse(Array("--partitions", "0,1"))
+    assertTrue(expectedOffsets().filter(r => r._2 <= 1) sameElements offsets.filter(r => !isConsumerOffsetTopicPartition(r)))
```
dajac: nit: `r => r._2 <= 1` is quite hard to read. It might be better to deconstruct the record so we can name the variable `partition`.

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
```diff
@@ -0,0 +1,194 @@
...
+import kafka.server.KafkaConfig
+import kafka.utils.{Exit, Logging, TestUtils}
+import org.apache.kafka.clients.CommonClientConfigs
+import org.apache.kafka.clients.consumer.{ConsumerConfig, KafkaConsumer}
+import org.apache.kafka.clients.producer.{KafkaProducer, ProducerConfig, ProducerRecord}
+import org.apache.kafka.common.serialization.{StringDeserializer, StringSerializer}
+import org.junit.Assert.{assertEquals, assertTrue}
+import org.junit.{Before, Test}
+
+class GetOffsetShellTest extends KafkaServerTestHarness with Logging {
```
dajac, line 30: It would be great if we could add few tests to verify the mutually exclusive arguments. We should also add unit tests to verify the parsing of the specs. We could make the relevant package private so we can access them from here.

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
```diff
@@ -0,0 +1,194 @@
...
+  private val topicCount = 4
+  private val offsetTopicPartitionCount = 4
+  private val topicPattern = Pattern.compile("test.*")
```
dajac: Is `test.*` correct? It seems that topics are named `topic...`.

dajac (commented):
@urbandan Thanks for the PR and sorry for the delay. I have made a first pass and have left some comments.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Not at all, thank you for catching this

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Thanks, fixed it

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: I agree, good catch

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Thanks for the catch

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: You are right, should be handled, fixed it

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: I was mostly focusing on avoiding unnecessary filtering, but you are right, it shouldn't be an issue, your code is much cleaner

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Removed unnecessary sorting, no duplication

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: No point, good catch

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: I agree, fixed it

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
urbandan: Completely agree, fixed it

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
urbandan: I wasn't sure if I should define any expectations on the end offsets of the internal topic, that's why I was testing it separately.
But I think you are right, so I refactored the code to handle all topics uniformly

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
urbandan: Was used by the dummy consumer, removed

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
urbandan: Agree, added coverage for the arguments and the filter parsing as well

urbandan:
@dajac Thank you for your review. Managed to work on this, and addressed your points.

dajac:
@urbandan Thanks for the update. I will make another pass on it this week. It seems that there are compilation error. Could you take a look please?

urbandan:
@dajac Fixed the compilation issues - there is a test failure, but I don't think those are related to the change

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -33,49 +34,65 @@ object GetOffsetShell {
...
     ToolsUtils.validatePortOrDie(parser, brokerList)
-    val topic = options.valueOf(topicOpt)
+    val excludeInternalTopics = options.has(excludeInternalTopicsOpt)
+
+    if (options.has(topicPartitionsOpt) && (options.has(topicOpt) || options.has(partitionsOpt))) {
+      System.err.println(s"--topic-partitions cannot be used with --topic or --partitions")
+      Exit.exit(1)
+    }
+
     val partitionIdsRequested: Set[Int] = {
```
dajac: How about moving the computation of `partitionIdsRequested` to L112? It does not make sense to do all of this if `options.has(topicPartitionsOpt)`.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -89,33 +106,34 @@ object GetOffsetShell {
...
-    }
+    val partitionInfos = listPartitionInfos(consumer, topicPartitionFilter)
 
-    if (partitionIdsRequested.nonEmpty) {
-      (partitionIdsRequested -- partitionInfos.map(_.partition)).foreach { partitionId =>
-        System.err.println(s"Error: partition $partitionId does not exist")
-      }
+    if (partitionInfos.isEmpty) {
+      System.err.println(s"Could not match any topic-partitions with the specified filters")
+      Exit.exit(1)
```
dajac: I am not fan of using `Exit.exit` all over the places. I know that it was already like this before but I do wonder if we could improve this by throwing exceptions instead and by catching them in the `main`. I think that this would require to strip down the main function a bit.

Another reason for this is that the new code throws `IllegalArgumentException` in few places but I don't think that we catch them anywhere, do we? What do you think?

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +150,104 @@ object GetOffsetShell {
         }
     }
 
-    partitionOffsets.toArray.sortBy { case (tp, _) => tp.partition }.foreach { case (tp, offset) =>
-      println(s"$topic:${tp.partition}:${Option(offset).getOrElse("")}")
+    partitionOffsets.toSeq.sortWith((tp1, tp2) => {
```
dajac: I just learnt that we can actually compare TopicPartition with the following comparator in Scala:

```
  def compareTopicPartitions(a: TopicPartition, b: TopicPartition): Boolean = {
    (a.topic(), a.partition()) < (b.topic(), b.partition())
  }
```

and do `partitionOffsets.toSeq.sortWith(compareTopicPartitions).foreach { ...`. This makes the code a bit more readable.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +150,104 @@ object GetOffsetShell {
...
+   * Creates a topic-partition filter based on a list of patterns.
+   * Expected format:
+   * List: TopicPartitionPattern(, TopicPartitionPattern)*
+   * TopicPartitionPattern: TopicPattern(:PartitionPattern)? | :PartitionPattern
+   * TopicPattern: REGEX
+   * PartitionPattern: NUMBER | NUMBER-(NUMBER)? | -NUMBER
+   */
+  def createTopicPartitionFilterWithPatternList(topicPartitions: String, excludeInternalTopics: Boolean): PartitionInfo => Boolean = {
+    val ruleSpecs = topicPartitions.split(",")
+    val rules = ruleSpecs.map { ruleSpec =>
```
dajac: Out of curiosity, did you consider using a regex to parse the rule spec? It seems to me that something like this  `([^:]*):([0-9]*)(?:-([0-9]*))?` could work and could directly extract all the required parts. I haven't tested it. That could reduce the code while making the grammar of the rule spec more explicit.

core/src/test/scala/kafka/tools/GetOffsetShellTest.scala
```diff
@@ -0,0 +1,207 @@
...
+
+    assertEquals(Some(1), exitStatus)
+  }
+
+  private def executeAndParse(args: Array[String]): List[(String, Int, Option[Long])] = {
+    val output = executeAndGrabOutput(args)
+    output.split(System.lineSeparator())
+      .map(_.split(":"))
+      .filter(_.length >= 2)
+      .map(line => {
```
dajac: nit: `.map(line => {` -> `.map { line =>`

dajac (commented):
@urbandan Thanks for the updates. I have left few more comments.

dajac:
Could you rebase the PR as well?

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Good catch, thank you

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Overall, I agree. Removed the exit calls, but left other util calls possibly invoking exit.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Thanks, refactored it

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: I personally prefer not using regex, but I see your point - switched to regex usage.

urbandan:
@dajac Thank you for your comments, finally managed to address them. Also rebased the branch.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -20,102 +20,131 @@ package kafka.tools
...
+import scala.math.Ordering.Implicits.infixOrderingOps
 
 object GetOffsetShell {
+  private val topicPartitionPattern = Pattern.compile( "([^:,]*)(?::([0-9]*)(?:-([0-9]*))?)?")
 
   def main(args: Array[String]): Unit = {
+    try {
+      fetchOffsets(args)
+    } catch {
+      case e: Exception => Exit.exit(1, Some(e.getMessage))
```
dajac: It seems that we need to print out something here for the user. The default exit procedure doesn't do anything with the provided message. Take a look at the other commands to see how we handle this.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +161,85 @@ object GetOffsetShell {
         }
     }
 
-    partitionOffsets.toArray.sortBy { case (tp, _) => tp.partition }.foreach { case (tp, offset) =>
-      println(s"$topic:${tp.partition}:${Option(offset).getOrElse("")}")
+    partitionOffsets.toSeq.sortWith((tp1, tp2) => compareTopicPartitions(tp1._1, tp2._1)).foreach {
```
dajac: nit: `sortWith((tp1, tp2) => compareTopicPartitions(tp1._1, tp2._1))` -> `sortWith(compareTopicPartitions)`

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +161,85 @@ object GetOffsetShell {
...
+  def parseRuleSpec(ruleSpec: String, excludeInternalTopics: Boolean): PartitionInfo => Boolean = {
+    def wrapNullOrEmpty(s: String): Option[String] = {
+      if (s == null || s.isEmpty)
+        None
+      else
+        Some(s)
+    }
+
+    val matcher = topicPartitionPattern.matcher(ruleSpec)
+    if (!matcher.matches() || matcher.groupCount() == 0)
```
dajac: I don't think that `matcher.groupCount() == 0` is necessary. My understanding is that it will never be `0` as we have groups defined in the regex.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +161,85 @@ object GetOffsetShell {
...
+      throw new IllegalArgumentException(s"Invalid rule specification: $ruleSpec")
+
+    val topicPattern = wrapNullOrEmpty(matcher.group(1))
+    val lowerRange = if (matcher.groupCount() >= 2) wrapNullOrEmpty(matcher.group(2)) else None
```
dajac: It seems that checking `matcher.groupCount() >= 2` is not necessary. Groups will be `null` anyway if they are not filled.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +161,85 @@ object GetOffsetShell {
...
+        p => p >= lowerBound
+      }
+      else {
+        p => p == lowerBound
+      }
+    } else {
+      val upperBound = upperRange.get.toInt
+      val lowerBound = lowerRange.get.toInt
+      p => p >= lowerBound && p < upperBound
+    }
```
dajac: I think that we could further simplify this by differentiating the range case from the single partition case in the regex directly. We could for instance use the following regex: `"([^:,]*)(?::(?:([0-9]*)|(?:([0-9]*)-([0-9]*))))?"`.

It allows to do something like this for the processing:
```
    val matcher = topicPartitionPattern.matcher(ruleSpec)
    if (!matcher.matches())
      throw new IllegalArgumentException(s"Invalid rule specification: $ruleSpec")

    def group(group: Int): Option[String] = {
      Option(matcher.group(group)).filter(s => s != null && s.nonEmpty)
    }

    val topicFilter = IncludeList(group(1).getOrElse(".*"))
    val partitionFilter = group(2).map(_.toInt) match {
      case Some(partition) =>
        (p: Int) => p == partition
      case None =>
        val lowerRange = group(3).map(_.toInt).getOrElse(0)
        val upperRange = group(4).map(_.toInt).getOrElse(Int.MaxValue)
        (p: Int) => p >= lowerRange && p < upperRange
    }
```

Basically, if no single partition is provided, we use a range with `0` and `Int.MaxValue` as defaults.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -20,102 +20,131 @@ package kafka.tools
...
+      )
+    }
+
+    val config = if (options.has(commandConfigOpt))
+      Utils.loadProps(options.valueOf(commandConfigOpt))
+    else
+      new Properties
     config.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, brokerList)
     config.setProperty(ConsumerConfig.CLIENT_ID_CONFIG, clientId)
     val consumer = new KafkaConsumer(config, new ByteArrayDeserializer, new ByteArrayDeserializer)
```
dajac, line 126: Not related to your changes but it seems that we never close the consumer. It would be great if we could do it.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +161,85 @@ object GetOffsetShell {
...
-    val partitionInfos = consumer.listTopics.asScala.filter { case (k, _) => k == topic }.values.flatMap(_.asScala).toBuffer
-    if (partitionInfos.isEmpty)
-      None
-    else if (partitionIds.isEmpty)
-      Some(partitionInfos)
-    else
-      Some(partitionInfos.filter(p => partitionIds.contains(p.partition)))
+  def createTopicPartitionFilterWithPatternList(topicPartitions: String, excludeInternalTopics: Boolean): PartitionInfo => Boolean = {
+    val ruleSpecs = topicPartitions.split(",")
+    val rules = ruleSpecs.map { ruleSpec => parseRuleSpec(ruleSpec, excludeInternalTopics) }
```
dajac: nit: We tend to use parenthesis instead of curly braces when the lambda is on a single line.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +161,85 @@ object GetOffsetShell {
...
+        Some(s)
+    }
+
+    val matcher = topicPartitionPattern.matcher(ruleSpec)
+    if (!matcher.matches() || matcher.groupCount() == 0)
+      throw new IllegalArgumentException(s"Invalid rule specification: $ruleSpec")
+
+    val topicPattern = wrapNullOrEmpty(matcher.group(1))
+    val lowerRange = if (matcher.groupCount() >= 2) wrapNullOrEmpty(matcher.group(2)) else None
+    val upperRangeSection = if (matcher.groupCount() >= 3) matcher.group(3) else null
```
dajac: The `null` instead of the `None` is quite subtile. I did not see it at first. See my comment which suggest to push more to the regex.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +161,85 @@ object GetOffsetShell {
...
+      else {
+        p => p == lowerBound
+      }
+    } else {
+      val upperBound = upperRange.get.toInt
+      val lowerBound = lowerRange.get.toInt
+      p => p >= lowerBound && p < upperBound
+    }
+
+    tp => includeList.isTopicAllowed(tp.topic(), excludeInternalTopics) && partitionFilter.apply(tp.partition())
```
dajac: nit: The parenthesis after `topic` and `partition` could be omitted. It seems that we could call `partitionFilter` directly without the `apply`.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +161,85 @@ object GetOffsetShell {
...
+    val topicsFilter = IncludeList(topicOpt.getOrElse(".*"))
+    t => topicsFilter.isTopicAllowed(t.topic, excludeInternalTopics) && (partitionIds.isEmpty || partitionIds.contains(t.partition))
+  }
+
+  /**
+   * Return the partition infos. Filter them with topicPartitionFilter.
+   */
+  private def listPartitionInfos(consumer: KafkaConsumer[_, _], topicPartitionFilter: PartitionInfo => Boolean): Seq[PartitionInfo] = {
+    consumer.listTopics.asScala.values.flatMap { partitions =>
+      partitions.asScala.filter { tp => topicPartitionFilter.apply(tp) }
```
dajac: nit: We should use parenthesis here instead of curly braces. `filter(topicPartitionFilter)` might even work directly as well.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -20,102 +20,131 @@ package kafka.tools
...
 import org.apache.kafka.common.serialization.ByteArrayDeserializer
+import org.apache.kafka.common.utils.Utils
 
+import java.util.regex.Pattern
 import scala.jdk.CollectionConverters._
 import scala.collection.Seq
+import scala.math.Ordering.Implicits.infixOrderingOps
 
 object GetOffsetShell {
+  private val topicPartitionPattern = Pattern.compile( "([^:,]*)(?::([0-9]*)(?:-([0-9]*))?)?")
```
dajac: nit: We usually capitalize the first letter of constant in Scala.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -20,102 +20,131 @@ package kafka.tools
...
+          Set.empty
+        else
+          partitionsString.split(",").map { partitionString =>
+            try partitionString.toInt
+            catch {
+              case _: NumberFormatException =>
+                throw new IllegalArgumentException(s"--partitions expects a comma separated list of numeric " +
+                  s"partition ids, but received: $partitionsString")
+            }
+          }.toSet
```
dajac: nit: Should we consider moving this block in an method?

dajac (commented):
@urbandan Thanks for the updates. I left few more comments.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: The current compareTopicPartitions method expects TopicPartitions, but we have tuples in the collection.
Should I change compareTopicPartitions to expect tuples?

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Thanks, good catch

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: True, good catch

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Thank you, changed implementation to this

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: I agree, more readable

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -20,102 +20,120 @@ package kafka.tools
...
 import org.apache.kafka.common.serialization.ByteArrayDeserializer
+import org.apache.kafka.common.utils.Utils
 
+import java.util.regex.Pattern
 import scala.jdk.CollectionConverters._
 import scala.collection.Seq
+import scala.math.Ordering.Implicits.infixOrderingOps
 
 object GetOffsetShell {
+  private val TopicPartitionPattern = Pattern.compile( "([^:,]*)(?::(?:([0-9]*)|(?:([0-9]*)-([0-9]*))))?")
```
dajac: nit: The space before `"` could be removed.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -20,102 +20,120 @@ package kafka.tools
...
 
   def main(args: Array[String]): Unit = {
+    try {
+      fetchOffsets(args)
+    } catch {
+      case e: Exception =>
+        System.err.println(s"Error occurred: $e.getMessage")
```
dajac: We could use `println` here to be consistent with the other command line tools. Moreover, you need to use curly braces around `e.getMessage`: s"Error: ${e.getMessage}".

core/src/main/scala/kafka/tools/GetOffsetShell.scala
dajac: No, that's fine as it is then. I missed this.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +150,79 @@ object GetOffsetShell {
...
    */
-  private def listPartitionInfos(consumer: KafkaConsumer[_, _], topic: String, partitionIds: Set[Int]): Option[Seq[PartitionInfo]] = {
-    val partitionInfos = consumer.listTopics.asScala.filter { case (k, _) => k == topic }.values.flatMap(_.asScala).toBuffer
-    if (partitionInfos.isEmpty)
-      None
-    else if (partitionIds.isEmpty)
-      Some(partitionInfos)
+  def createTopicPartitionFilterWithPatternList(topicPartitions: String, excludeInternalTopics: Boolean): PartitionInfo => Boolean = {
+    val ruleSpecs = topicPartitions.split(",")
+    val rules = ruleSpecs.map { ruleSpec => parseRuleSpec(ruleSpec, excludeInternalTopics) }
```
dajac: `.map { ruleSpec => parseRuleSpec(ruleSpec, excludeInternalTopics) }` => `.map(ruleSpec => parseRuleSpec(ruleSpec, excludeInternalTopics))` - We usually don't use curly braces when the lambda is on one line.

core/src/main/scala/kafka/tools/GetOffsetShell.scala
```diff
@@ -132,23 +150,79 @@ object GetOffsetShell {
...
+    val partitionFilter = group(2).map(_.toInt) match {
+      case Some(partition) =>
+        (p: Int) => p == partition
+      case None =>
+        val lowerRange = group(3).map(_.toInt).getOrElse(0)
+        val upperRange = group(4).map(_.toInt).getOrElse(Int.MaxValue)
+        (p: Int) => p >= lowerRange && p < upperRange
+    }
+
+    tp => topicFilter.isTopicAllowed(tp.topic, excludeInternalTopics) && partitionFilter(tp.partition())
```
dajac: `tp.partition()` => `tp.partition`

core/src/main/scala/kafka/tools/GetOffsetShell.scala
dajac: Could we address this one as well? We could add a try/finally.

dajac (commented):
@urbandan Thanks for the updates. We are almost there! I just left few minor/cosmetic comments.

dajac:
Launched system tests: https://jenkins.confluent.io/job/system-test-kafka-branch-builder/4362/

core/src/main/scala/kafka/tools/GetOffsetShell.scala
urbandan: Sorry, missed this one, fixed it

urbandan:
@dajac Really appreciate your review, fixed the last couple of comments.

dajac:
@urbandan It seems that most of the get offset system tests failed: http://confluent-kafka-branch-builder-system-test-results.s3-us-west-2.amazonaws.com/2021-02-09--001.1612926625--urbandan--KIP-635_GetOffsetShell--8e69e1dc5/report.html. Could you take a look to see what went wrong?

urbandan:
@dajac I think I found the issue, pushed the fix. There was a typo in the command arg names. Could not get to execute the tests locally yet.

dajac:
Triggered another system test: https://jenkins.confluent.io/job/system-test-kafka-branch-builder/4364/

dajac:
All the `kafkatest.tests.core.get_offset_shell_test` system tests passed.

dajac (approved):
LGTM! @urbandan Thanks for your contribution!

dajac:
Failed tests are not related. Merging to trunk.

PR #9431, opened 2020-10-14
KAFKA-10426: Deadlock on session key update.

xakassi:
DistributedHerder goes to updateConfigsWithIncrementalCooperative() synchronized method and called configBackingStore.snapshot() which take a lock on internal object in KafkaConfigBackingStore class.

Meanwhile KafkaConfigBackingStore in ConsumeCallback inside synchronized block on internal object gets SESSION_KEY record and calls updateListener.onSessionKeyUpdate() which take a lock on DistributedHerder.

So, we have a Deadlock.

To avoid this, updateListener with new session key should be called outside synchronized block as it's done, for example, for updateListener.onTaskConfigUpdate(updatedTasks).

This PR is a copy of: https://github.com/apache/kafka/pull/9211

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/storage/KafkaConfigBackingStore.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/storage/KafkaConfigBackingStore.java
@@ -699,46 +699,42 @@ public class KafkaConfigBackingStore implements ConfigBackingStore {
                 if (started)
                     updateListener.onTaskConfigUpdate(updatedTasks);
             } else if (record.key().equals(SESSION_KEY_KEY)) {
-                SessionKey newSessionKey;
-                synchronized (lock) {
-                    if (value.value() == null) {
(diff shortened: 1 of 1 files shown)


xakassi:
Hi, @kkonstantine !
Seems one test failed, but I can't see why. It's not likely that my changes make it fail. Maybe we should rerun the tests? But I suppose I have no permissions for it.

kkonstantine (approved):
The failures were irrelevant to Connect and these changes. 
Thanks for the fix @xakassi !

LGTM

xakassi:
Great! Thank you, @kkonstantine !
One more question. I'm going to resolve my ticket https://issues.apache.org/jira/browse/KAFKA-10426
So which fix version should I set?
2.6.0 is the latest version now, right? So can I set fix version 2.6.1?

kkonstantine:
Thanks for checking @xakassi 
Merged all the way to 2.4. I updated the fix versions and closed the jira.

PR #9432, opened 2020-10-14
KAFKA-10559: Not letting TimeoutException shutdown the app during internal topic validation

vamossagar12:
comment to ping reviewers.*


ableegoldman:
LGTM

Looks like the Java 8 build may have stalled? The Java 11 & 15 builds both passed a while ago

mjsax:
Thanks for the PR @vamossagar12!

No ideal to just remove the test instead of adjusting them, but we can do this in a follow up PR. @ableegoldman told me she will take care of it.

mjsax:
Follow up PR: https://github.com/apache/kafka/pull/9446

vamossagar12:
Thanks @mjsax . I thought these were time out related tests so ended up removing them. Thanks for the feedback :)

mjsax:
Well, that are timeout related: they checked if we send the shut-down signal. After the change, they should check if the `TimeoutException` is thrown and only kills the local thread. :)

PR #9433, opened 2020-10-14
KAFKA-10607: Consistent behaviour for response errorCounts()

tombentley:


tombentley:
@abbccdda please could you take a look since you opened the JIRA? Thanks.

tombentley:
@abbccdda and chance you could take a look at this? Or maybe @dajac?

dongjinleekr (commented):
Good. All tests pass cleanly.
pr-9433-success
+1. Are there any other omitted classes? You know, we have lots of children classes in `AbstractResponse`.

tombentley:
@dongjinleekr I did a review of all the subclasses.

clients/src/main/java/org/apache/kafka/common/requests/AbstractResponse.java
```diff
@@ -55,6 +55,11 @@ public ByteBuffer serialize(ApiKeys apiKey, short version, int correlationId) {
         return RequestUtils.serialize(header.toStruct(), toStruct(version));
     }
 
+    /**
+     * The number of each type of error in the response, including {@link Errors#NONE} and top-level errors as well as
+     * more specificly scoped errors (such as topic or partition-level errors).
```
chia7712: typo: specificly -> specifically

tombentley:
@chia7712 fixed, thanks.

tombentley:
@chia7712 is there anything more you needed on this PR?

chia7712:
Could you offer test to make sure ```None``` is included.

Personally, the implementations of ```errorCounts``` are almost same. Maybe it should be implemented
 by auto-generated protocol so the consistency (code style and behavior) can be protected. @hachikuji WDYT?

tombentley:
> Personally, the implementations of `errorCounts` are almost same. Maybe it should be implemented
> by auto-generated protocol so the consistency (code style and behavior) can be protected. @hachikuji WDYT?

That's a good idea in theory, but in practice the code generator doesn't know which fields are error codes, apart from the convention that they're called `ErrorCode` and are an `int16`. If we ever needed to call an error code field something else `errorCounts()` would break. I think there is some value in having a richer way of expressing types (that is, separately from the "serialization type") in the RPC JSON (see https://issues.apache.org/jira/browse/KAFKA-7787 for example), but any effort to do that would require a KIP and potentially impose a burden on other client implementer who are consuming that JSON.

chia7712:
@tombentley Thanks for your sharing. It inspires me to think about the possible solution. The following comment is not related to this PR. just discussion 😃

> That's a good idea in theory, but in practice the code generator doesn't know which fields are error codes, apart from the convention that they're called ErrorCode and are an int16. 

How about using ```EntityType``` (called ```ErrorCode```). for example:
```json

      { "name": "ErrorCode", "type": "int16", "versions": "0+", "entityType": "errorCode",
        "about": "The error code, or 0 if there was no error." },
```

> If we ever needed to call an error code field something else errorCounts() would break. I think there is some value in having a richer way of expressing types (that is, separately from the "serialization type") in the RPC JSON (see https://issues.apache.org/jira/browse/KAFKA-7787 for example), but any effort to do that would require a KIP and potentially impose a burden on other client implementer who are consuming that JSON.

following above new entityType ```errorCode```
1. We can add a new method ```Map<Errors, Integer> errorCounts()``` to ```Message``` interface.
1. By default, it returns empty map (or None -> 1)
1. code generator gives custom implementation only if the message has either message field or error code field (i.e entity type is ErrorCode).
1. the composite message class should merge all error count from all fields

tombentley:
@chia7712 For this PR I've added a test. It revealed a few places where responses had error fields which weren't included in the count. This will affect the metrics for those RPCs, since they'll now be counting more error codes and potentially observing error code values which weren't observed before. I don't think this is a problem, but since it affects Fetch and Metadata requests, for example, I think it's worth calling out.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -1921,7 +1929,7 @@ private CreateAclsResponse createCreateAclsResponse() {
         return new CreateAclsResponse(new CreateAclsResponseData().setResults(asList(
             new CreateAclsResponseData.AclCreationResult(),
             new CreateAclsResponseData.AclCreationResult()
-                .setErrorCode(Errors.INVALID_REQUEST.code())
+                .setErrorCode(Errors.NONE.code())
```
chia7712, line 1997: pardon me, why to make this change?

chia7712 (approved):
@tombentley Thanks for updating code. overall LGTM. please take a look at a minor question.

@abbccdda Could you take a look to make sure this PR has addressed the issue you raised?

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
tombentley: I wanted to ensure that for error code fields in all RPCs we were including NONE in the count. I would have factored out a method taking the error code to use if any of the rest of the tests depended on the actual error code used, but they don't. We'd need to revisit this if [KIP-636](https://cwiki.apache.org/confluence/display/KAFKA/KIP-636%3A+Make+RPC+error+codes+and+messages+tagged+fields) were ever accepted (since it would interfere with coverage of the serialization test), but it seems harmless for the time being.

tombentley:
@abbccdda I think we're just waiting for your review here. Thanks.

chia7712:
@tombentley Could you trigger QA again?

tombentley:
@chia7712 I thought only committers can do that.

chia7712:
IIRC, the retest command does not work currently. Could you rebase code to trigger QA?

tombentley:
@chia7712 done

chia7712:
@tombentley thanks! I will merge this PR tomorrow if no objection :)

chia7712:
@tombentley Thanks for your patch!

tombentley:
Thanks @chia7712 for the review!

clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
```diff
@@ -109,8 +109,10 @@ public int throttleTimeMs() {
     @Override
     public Map<Errors, Integer> errorCounts() {
         Map<Errors, Integer> errorCounts = new HashMap<>();
-        data.topics().forEach(metadata ->
-            updateErrorCounts(errorCounts, Errors.forCode(metadata.errorCode())));
+        data.topics().forEach(metadata -> {
+            metadata.partitions().forEach(p -> updateErrorCounts(errorCounts, Errors.forCode(p.errorCode())));
```
ijuma, line 113: Why did we change this?

chia7712: It includes the partition-level error. It seems to me it follows the docs.
`
The number of each type of error in the response, including {@link Errors#NONE} and top-level errors as well as more specifically scoped errors (such as topic or partition-level errors).
`

ijuma: A metadata request has topics as the "scoped error" (one cannot request a metadata request for a given partition). So, it looks to me that we introduced a regression here. We should check if there are similar cases for other requests.

chia7712: > A metadata request has topics as the "scoped error" (one cannot request a metadata request for a given partition).

I check the server-side code. You are right. This change should be reverted. I will file a PR to fix it ASAP. This behavior can be broken inadvertently (server-side/client-side add/remove specifically scoped error). not sure whether we can prevent it effectively (maybe add more tests)

chia7712: https://github.com/apache/kafka/pull/11127

PR #9434, opened 2020-10-14
MINOR: Handle lastFetchedEpoch/divergingEpoch in FetchSession and DelayedFetch

rajinisivaram:
In 2.7, we added lastFetchedEpoch to fetch requests and divergingEpoch to fetch responses. We are not using these for truncation yet, but in order to use these for truncation with IBP 2.7 onwards in the next release, we should make sure that we handle these in all the supporting classes even in 2.7.

### DIFF
--- a/core/src/main/scala/kafka/cluster/Partition.scala
+++ b/core/src/main/scala/kafka/cluster/Partition.scala
@@ -1162,6 +1162,13 @@ class Partition(val topicPartition: TopicPartition,
     localLog.fetchOffsetSnapshot
   }
 
+  def hasDivergingEpoch(currentLeaderEpoch: Optional[Integer],
+                        lastFetchedEpoch: Int,
+                        fetchOffset: Long): Boolean = {
+    val epochEndOffset = lastOffsetForLeaderEpoch(currentLeaderEpoch, lastFetchedEpoch, fetchOnlyFromLeader = false)
+    epochEndOffset.leaderEpoch < lastFetchedEpoch || epochEndOffset.endOffset < fetchOffset
+  }
+
   def legacyFetchOffsetsForTimestamp(timestamp: Long,
                                      maxNumOffsets: Int,
                                      isFromConsumer: Boolean,
--- a/core/src/main/scala/kafka/cluster/Partition.scala
+++ b/core/src/main/scala/kafka/cluster/Partition.scala
@@ -1162,13 +1162,6 @@ class Partition(val topicPartition: TopicPartition,
     localLog.fetchOffsetSnapshot
   }
 
-  def hasDivergingEpoch(currentLeaderEpoch: Optional[Integer],
-                        lastFetchedEpoch: Int,
-                        fetchOffset: Long): Boolean = {
-    val epochEndOffset = lastOffsetForLeaderEpoch(currentLeaderEpoch, lastFetchedEpoch, fetchOnlyFromLeader = false)
-    epochEndOffset.leaderEpoch < lastFetchedEpoch || epochEndOffset.endOffset < fetchOffset
-  }
-
   def legacyFetchOffsetsForTimestamp(timestamp: Long,
                                      maxNumOffsets: Int,
                                      isFromConsumer: Boolean,
--- a/core/src/main/scala/kafka/server/DelayedFetch.scala
+++ b/core/src/main/scala/kafka/server/DelayedFetch.scala
@@ -27,11 +27,12 @@ import org.apache.kafka.common.requests.FetchRequest.PartitionData
 
 import scala.collection._
 
-case class FetchPartitionStatus(startOffsetMetadata: LogOffsetMetadata, fetchInfo: PartitionData) {
+case class FetchPartitionStatus(startOffsetMetadata: LogOffsetMetadata, fetchInfo: PartitionData, hasDivergingEpoch: Boolean) {
 
   override def toString: String = {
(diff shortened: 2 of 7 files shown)


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1084,7 +1084,7 @@ class ReplicaManager(val config: KafkaConfig,
       fetchInfos.foreach { case (topicPartition, partitionData) =>
         logReadResultMap.get(topicPartition).foreach(logReadResult => {
           val logOffsetMetadata = logReadResult.info.fetchOffsetMetadata
-          fetchPartitionStatus += (topicPartition -> FetchPartitionStatus(logOffsetMetadata, partitionData))
+          fetchPartitionStatus += (topicPartition -> FetchPartitionStatus(logOffsetMetadata, partitionData, logReadResult.divergingEpoch.nonEmpty))
```
hachikuji: Hmm.. If the `LogReadResult` has a diverging epoch, wouldn't we want to respond immediately?

core/src/main/scala/kafka/server/DelayedFetch.scala
```diff
@@ -88,6 +90,13 @@ class DelayedFetch(delayMs: Long,
         try {
           if (fetchOffset != LogOffsetMetadata.UnknownOffsetMetadata) {
             val partition = replicaManager.getPartitionOrException(topicPartition)
+
+            // Case H: Return diverging epoch in response to trigger truncation
+            if (fetchStatus.hasDivergingEpoch) {
```
hachikuji: Here we are using the status from the original fetch. I am wondering if we need to recheck below since it is possible to get a truncation while a fetch is in purgatory.

hachikuji (commented):
Thanks @rajinisivaram . Left a couple comments.

core/src/main/scala/kafka/server/ReplicaManager.scala
rajinisivaram: ah, yes, so we don't need to check the original result in DelayedFetch, we return immediately here. Updated.

core/src/main/scala/kafka/server/DelayedFetch.scala
rajinisivaram: @hachikuji Thanks for the review. Makes sense, I have added a new check at the end instead of this one, not sure if there is a better way to check.

core/src/main/scala/kafka/server/DelayedFetch.scala
```diff
@@ -96,6 +98,7 @@ class DelayedFetch(delayMs: Long,
               case FetchTxnCommitted => offsetSnapshot.lastStableOffset
             }
 
+
```
hachikuji: nit: unneeded newline

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -1162,6 +1162,13 @@ class Partition(val topicPartition: TopicPartition,
     localLog.fetchOffsetSnapshot
   }
 
+  def hasDivergingEpoch(currentLeaderEpoch: Optional[Integer],
+                        lastFetchedEpoch: Int,
+                        fetchOffset: Long): Boolean = {
+    val epochEndOffset = lastOffsetForLeaderEpoch(currentLeaderEpoch, lastFetchedEpoch, fetchOnlyFromLeader = false)
```
hachikuji: Do we need to check the error? Or are you relying on the check below failing if `UNDEFINED_EPOCH` is returned?

core/src/main/scala/kafka/server/FetchSession.scala
```diff
@@ -77,7 +77,8 @@ class CachedPartition(val topic: String,
                       var highWatermark: Long,
                       var leaderEpoch: Optional[Integer],
                       var fetcherLogStartOffset: Long,
-                      var localLogStartOffset: Long)
+                      var localLogStartOffset: Long,
+                      var lastFetchedEpoch: Optional[Integer] = Optional.empty[Integer])
```
hachikuji: Do we need to provide a default here?

core/src/main/scala/kafka/cluster/Partition.scala
rajinisivaram: @hachikuji Thanks for the review. I took the check from `Partition.readRecords`, but we throw exceptions there to return appropriate errors. I was thinking we would return true here for undefined epochs because of the check below and that would go through the other code path to return the appropriate errors or diverging epoch. Do you think we should do the same error handling hchecks ere as in readRecords to make the flow more obvious?

rajinisivaram: As discussed offline, moved the check to DelayedFetch.

core/src/main/scala/kafka/server/FetchSession.scala
rajinisivaram: removed

rajinisivaram:
@hachikuji Thanks for the review, have addressed the comments

hachikuji (approved):
LGTM. Thanks for the patch!

rajinisivaram:
@hachikuji Thanks for the reviews. Test failure not related (`kafka.network.ConnectionQuotasTest.testNoConnectionLimitsByDefault`). Merging to trunk and 2.7.

PR #9435, opened 2020-10-14
KAFKA-10606: Disable auto topic creation for fetch-all-topic-metadata request

Lincong:
There is a bug that causes fetch-all-topic-metadata requests triggering
auto topic creation. Details are described in KAFKA-10606. This is the
simplest way to fix this bug on the broker side.

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaApis.scala
+++ b/core/src/main/scala/kafka/server/KafkaApis.scala
@@ -1242,6 +1242,7 @@ class KafkaApis(val requestChannel: RequestChannel,
                                errorUnavailableListeners: Boolean): Seq[MetadataResponseTopic] = {
     val topicResponses = metadataCache.getTopicMetadata(topics, listenerName,
         errorUnavailableEndpoints, errorUnavailableListeners)
+
     if (topics.isEmpty || topicResponses.size == topics.size) {
       topicResponses
     } else {
@@ -1256,12 +1257,8 @@ class KafkaApis(val requestChannel: RequestChannel,
               topicMetadata
           )
         } else if (isFetchAllMetadata) {
-          // KAFKA-10606: If this request is to get metadata for all topics, auto topic creation should not be allowed
-          // The special handling is necessary on broker side because allowAutoTopicCreation is hard coded to true
-          // for backward compatibility on client side.
-          //
-          // However, in previous versions, UNKNOWN_TOPIC_OR_PARTITION won't happen on fetch all metadata,
-          // so, for backward-compatibility, we need to skip these not founds during fetch all metadata here.
+          // A metadata request for all topics should never result in topic auto creation, but a topic may be deleted
+          // in between the creation of the topics parameter and topicResponses, so make sure to return None for this case.
           None
         } else if (allowAutoTopicCreation && config.autoCreateTopicsEnable) {
           Some(createTopic(topic, config.numPartitions, config.defaultReplicationFactor))
--- a/core/src/main/scala/kafka/server/KafkaApis.scala
+++ b/core/src/main/scala/kafka/server/KafkaApis.scala
@@ -1236,7 +1236,8 @@ class KafkaApis(val requestChannel: RequestChannel,
     topicMetadata.headOption.getOrElse(createInternalTopic(topic))
   }
 
-  private def getTopicMetadata(allowAutoTopicCreation: Boolean, topics: Set[String], listenerName: ListenerName,
+  private def getTopicMetadata(allowAutoTopicCreation: Boolean, isFetchAllMetadata: Boolean,
+                               topics: Set[String], listenerName: ListenerName,
                                errorUnavailableEndpoints: Boolean,
                                errorUnavailableListeners: Boolean): Seq[MetadataResponseTopic] = {
     val topicResponses = metadataCache.getTopicMetadata(topics, listenerName,
@@ -1252,13 +1253,23 @@ class KafkaApis(val requestChannel: RequestChannel,
             metadataResponseTopic(Errors.INVALID_REPLICATION_FACTOR, topic, true, util.Collections.emptyList())
           else
             topicMetadata
-        } else if (allowAutoTopicCreation && config.autoCreateTopicsEnable) {
+        } else if (!isFetchAllMetadata && allowAutoTopicCreation && config.autoCreateTopicsEnable) {
+          // KAFKA-10606: If this request is to get metadata for all topics, auto topic creation should not be allowed
+          // The special handling is necessary on broker side because allowAutoTopicCreation is hard coded to true
+          // for backward compatibility on client side.
           createTopic(topic, config.numPartitions, config.defaultReplicationFactor)
         } else {
           metadataResponseTopic(Errors.UNKNOWN_TOPIC_OR_PARTITION, topic, false, util.Collections.emptyList())
         }
       }
-      topicResponses ++ responsesForNonExistentTopics
+      topicResponses ++ (
+        if (isFetchAllMetadata)
+          // KAFKA-10606: In previous versions, UNKNOWN_TOPIC_OR_PARTITION won't happen on fetch all metadata,
+          // so we filter out those UNKNOWN_TOPIC_OR_PARTITION during fetch all metadata for backward-compatibility.
+          responsesForNonExistentTopics.filter { _.errorCode() != Errors.UNKNOWN_TOPIC_OR_PARTITION.code() }
+        else
+          responsesForNonExistentTopics
+      )
     }
   }
 
@@ -1314,12 +1325,9 @@ class KafkaApis(val requestChannel: RequestChannel,
       if (authorizedTopics.isEmpty)
         Seq.empty[MetadataResponseTopic]
       else {
-        // KAFKA-10606: If this request is to get metadata for all topics, auto topic creation should not be allowed
-        // The special handling is necessary on broker side because allowAutoTopicCreation is hard coded to true
-        // for backward compatibility on client side.
-        val allowAutoTopicCreation = (!metadataRequest.isAllTopics) && metadataRequest.allowAutoTopicCreation
         getTopicMetadata(
-          allowAutoTopicCreation,
+          metadataRequest.allowAutoTopicCreation,
+          metadataRequest.isAllTopics,
           authorizedTopics,
           request.context.listenerName,
           errorUnavailableEndpoints,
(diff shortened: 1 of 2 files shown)


ijuma:
Thanks for the PR. Thoughts on how to test this?

Lincong:
@ijuma One way to test it could be to mock objects passed to the `KafkaApis` class. The `MetadataCache` object should be mocked in a way that upon the first invocation on the method `getAllTopics` to simulate the scenario in which the metadata cache got updated "concurrently" with `handleTopicMetadataRequest`.

However, IMO, this change is pretty minor. So, I am wondering how necessary it is to add tests for it?

ijuma:
`KafkaApisTest` has a few tests like that. Could we add one there? The main concern is that we will regress here if we don't have tests.

Lincong:
@ijuma  Will do! Thanks

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1236,9 +1236,19 @@ class KafkaApis(val requestChannel: RequestChannel,
...
       if (authorizedTopics.isEmpty)
         Seq.empty[MetadataResponseTopic]
-      else
-        getTopicMetadata(metadataRequest.allowAutoTopicCreation, authorizedTopics, request.context.listenerName,
-          errorUnavailableEndpoints, errorUnavailableListeners)
+      else {
+        // KAFKA-10606: If this request is to get metadata for all topics, auto topic creation should not be allowed
+        // The special handling is necessary on broker side because allowAutoTopicCreation is hard coded to true
+        // for backward compatibility on client side.
+        val allowAutoTopicCreation = (!metadataRequest.isAllTopics) && metadataRequest.allowAutoTopicCreation
```
chia7712: If we pass ```false``` to getTopicMetadata, it generates ```UNKNOWN_TOPIC_OR_PARTITION``` when the topic is removed, right? If so, does client-side need to handle such error? For example, KafkaAdminClient#listTopics should filter out those "nonexistent" topics (https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L1717). Otherwise, users want to get all "existent" topics but response say a_topic is "nonexistent".

chia7712 (commented):
@Lincong Nice finding and patch. I left a small question.

core/src/main/scala/kafka/server/KafkaApis.scala
Lincong: Hi @chia7712 , thank you for bringing up this issue and it's totally valid! 
In order to preserve its behavior (not including `UNKNOWN_TOPIC_OR_PARTITION ` in the response to fetch all topic metadata request), we implemented the below logic which filters out all entries in the response with `UNKNOWN_TOPIC_OR_PARTITION ` if the metadata request is to get metadata for all topisc.

```
val completeTopicMetadata = (if (metadataRequest.isAllTopics) {
    opicMetadata.filter(_.errorCode() != Errors.UNKNOWN_TOPIC_OR_PARTITION.code())
} else {
    topicMetadata
}) ++ unauthorizedForCreateTopicMetadata ++ unauthorizedForDescribeTopicMetadata
```
We have added a unit test as well. Hope to get some feedback from you soon!

chia7712:
@Lincong Thanks for updating code. Could you fix the conflicting files as well?

Lincong:
> @Lincong Thanks for updating code. Could you fix the conflicting files as well?

Conflicts resolved

chia7712:
@Lincong Could you take a look at those failed tests? I will give a review later :)

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1333,7 +1343,17 @@ class KafkaApis(val requestChannel: RequestChannel,
       }
     }
 
-    val completeTopicMetadata = topicMetadata ++ unauthorizedForCreateTopicMetadata ++ unauthorizedForDescribeTopicMetadata
+    val completeTopicMetadata = (if (metadataRequest.isAllTopics) {
```
chia7712: Could we do this filter early? For example, how about filtering ```getTopicMetadata``` directly? That makes all fixes be in one place and following authorized ops can skip "unknown partition/topic" as well.

chia7712 (approved):
@Lincong Thanks for your patch. overall LGTM. Just leave a small suggestion. Please take a look.

core/src/main/scala/kafka/server/KafkaApis.scala
ijuma: This is a good point. We should pass a boolean to `getTopicMetadata` indicating that it's an "allTopics" request and have that method handle everything.

lmr3796: Hi @chia7712 @ijuma ,

This is Joseph and I'm @Lincong 's colleague working on this patch with him.  I think this is a good point and just updated the PR accordingly

chia7712:
+1 to last commit. Will merge it to trunk tomorrow if no object.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -107,6 +108,11 @@ class KafkaApisTest {
   private val time = new MockTime
   private val clientId = ""
 
+  @Before
+  def setUp(): Unit = {
```
chia7712: Could we remove this ```@Before``` code? Changing modifier of ```metadataCache``` from "val" to "var" is enough.

lmr3796: hey @chia7712 ,

I add this to let each test case to have a clean, uncorrupted `MetadataCache`.  Is there any historical context that we should avoid that cleanup?

chia7712: It seems to me following code is able to generate a clean, uncorrupted ```MetadataCache``` for each test case.

```
var metadataCache = new MetadataCache(brokerId)
```

Also, it is simpler than ```@Before``` block.

lmr3796: Hey @chia7712 you're right.  I did some reading and found that I misunderstood JUnit's behavior.  Just updated that part and rebased latest trunk.

core/src/main/scala/kafka/server/KafkaApis.scala
chia7712: @ijuma I will merge this PR tomorrow if there is no objection.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1250,13 +1251,23 @@ class KafkaApis(val requestChannel: RequestChannel,
...
           else
             topicMetadata
-        } else if (allowAutoTopicCreation && config.autoCreateTopicsEnable) {
+        } else if (!isFetchAllMetadata && allowAutoTopicCreation && config.autoCreateTopicsEnable) {
+          // KAFKA-10606: If this request is to get metadata for all topics, auto topic creation should not be allowed
+          // The special handling is necessary on broker side because allowAutoTopicCreation is hard coded to true
+          // for backward compatibility on client side.
           createTopic(topic, config.numPartitions, config.defaultReplicationFactor)
         } else {
           metadataResponseTopic(Errors.UNKNOWN_TOPIC_OR_PARTITION, topic, false, util.Collections.emptyList())
```
ijuma: Could we not avoid this altogether instead of doing the `filter` later?

lmr3796: @ijuma I slightly changed it (and rebased latest) so the processing is a branch in the middle about `isFetchAllMetadata`.

It's still a `map` and `filter` structure because I don't want to sacrifice the readability, but it's handled in the middle instead of at the returning statement, which I think is more elegant and readable than the original version.

Let me know what you think!

lmr3796: Also, I added the assertion of no UNKNOWN_TOPIC_OR_PARTITION into the unit test

ijuma: I think if you use `flatMap`, you don't need the `filter` and it's still readable. Thoughts?

lmr3796: @ijuma That's a good point.  I've updated it.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1243,19 +1244,30 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+              topicMetadata
+          )
+        } else if (isFetchAllMetadata) {
+          // KAFKA-10606: If this request is to get metadata for all topics, auto topic creation should not be allowed
+          // The special handling is necessary on broker side because allowAutoTopicCreation is hard coded to true
+          // for backward compatibility on client side.
+          //
+          // However, in previous versions, UNKNOWN_TOPIC_OR_PARTITION won't happen on fetch all metadata,
+          // so, for backward-compatibility, we need to skip these not founds during fetch all metadata here.
+          Nil
```
chia7712: Is ```Some/None``` more readable than ```List/Nil```?

chia7712 (commented):
@Lincong @lmr3796 Thanks for the updating. one small question is raised. Please take a look.

core/src/main/scala/kafka/server/KafkaApis.scala
lmr3796: @chia7712 Thanks, you're right.  Just updated accordingly and rebased latest trunk

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1243,19 +1244,30 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+            else
+              topicMetadata
+          )
+        } else if (isFetchAllMetadata) {
+          // KAFKA-10606: If this request is to get metadata for all topics, auto topic creation should not be allowed
+          // The special handling is necessary on broker side because allowAutoTopicCreation is hard coded to true
+          // for backward compatibility on client side.
+          //
+          // However, in previous versions, UNKNOWN_TOPIC_OR_PARTITION won't happen on fetch all metadata,
+          // so, for backward-compatibility, we need to skip these not founds during fetch all metadata here.
```
ijuma: We don't usually include JIRA references in the code unless it's a very complex issue. I think the comment here could be something like:

"A metadata request for all topics should never result in topic auto creation. A topic may be deleted between the creation of `topics` and `topicResponses`, so we make sure to always return `None` for this case."

lmr3796: Thanks @ijuma , 

To follow the convention in Kafka, I just updated the comments.  Let me know if you have there are any other thoughts!

chia7712:
@lmr3796 Thanks for your updating. Could you rebase code to include recent big commits to trigger QA again?

lmr3796:
@chia7712 @ijuma I just rebased to latest trunk.  Please take a look

chia7712:
@ijuma I'd like to merge it tomorrow. would be helpful if you could give a final review :)

ijuma:
@chia7712 The code changes seen fine. The comment in KafkaApis still seems a bit convoluted. For example, not returning UnknownTopicOrPartition is not related to backwards compatibility, it doesn't make sense to return that if the given topic wasn't explicitly requested. Similarly, the fact that allowAutoTopicCreation is true for some clients isn't that relevant, the code needs to handle it irrespective of what some clients do or don't do.

I thought my suggested comment in the following was clear and concise, but I'm interested in other opinions:

https://github.com/apache/kafka/pull/9435#discussion_r535321971

chia7712:
> I thought my suggested comment in the following was clear and concise, but I'm interested in other opinions:
> #9435 (comment)

@ijuma Sorry that I neglected that conversation.

@lmr3796 FYI ~

lmr3796:
Hi thanks @ijuma @chia7712 ,

I updated the comments accordingly to make it more concise and rebased latest trunk.
Please let me know if there are any other issues.

chia7712:
> I updated the comments accordingly to make it more concise and rebased latest trunk.
Please let me know if there are any other issues.

@lmr3796 Thanks for updating. waiting for response from @ijuma :)

ijuma:
@chia7712 feel free to merge if it looks good to you. Thanks @lmr3796 for the patience.

chia7712:
@Lincong @lmr3796 Thanks for your efforts!

PR #9444, opened 2020-10-15
KAFKA-8305: Doc Fix default.replication.factor

garyrussell:
Doc fix for KAFKA-8305; mention its alternate use.

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaConfig.scala
+++ b/core/src/main/scala/kafka/server/KafkaConfig.scala
@@ -499,7 +499,7 @@ object KafkaConfig {
   /** ********* Replication configuration ***********/
   val ControllerSocketTimeoutMsDoc = "The socket timeout for controller-to-broker channels"
   val ControllerMessageQueueSizeDoc = "The buffer size for controller-to-broker-channels"
-  val DefaultReplicationFactorDoc = "default replication factors for automatically created topics"
+  val DefaultReplicationFactorDoc = "default replication factor for automatically created topics, and for AdminClient topic creation requests with -1 in the replication factor"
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -499,7 +499,7 @@ object KafkaConfig {
   /** ********* Replication configuration ***********/
   val ControllerSocketTimeoutMsDoc = "The socket timeout for controller-to-broker channels"
   val ControllerMessageQueueSizeDoc = "The buffer size for controller-to-broker-channels"
-  val DefaultReplicationFactorDoc = "default replication factors for automatically created topics"
+  val DefaultReplicationFactorDoc = "default replication factor for automatically created topics, and for AdminClient topic creation requests with -1 in the replication factor"
```
mimaison, line 502: Can we add `The` at the start?
I think we can remove AdminClient for the sentence too. Also would the sentence sound better with `as the replication factor` ?

mimaison (commented):
Sorry for the super late review, it seems like this fell through the cracks!
I think this would still be a nice improvement for this doc. Can you rebase and look at the suggestions I left? Thanks

core/src/main/scala/kafka/server/KafkaConfig.scala
garyrussell: @mimaison I have retired and currently don't have an environment set up on my home machines; feel free to take this over and make your suggested changes. Looks like it needs to be reworked anyway due to other changes in these property definitions.

mimaison: No worries, thanks!

PR #9447, opened 2020-10-15
MINOR: Fixed comment to refer to UpdateMetadataPartitionState 

jolshan:
This part of the code is a little confusing, so I thought I should fix this comment. 

When creating a builder for UpdateMetadataRequest, the topic name is set in UpdateMetadataPartitionState. For versions 5 and onward, the topic name is added to the UpdateMetadataTopicState, and this code does not null out the topic name in the  
UpdateMetadataPartitionState when it does this.


dajac (approved):
LGTM. Thanks for the fix.

PR #9454, opened 2020-10-19
KAFKA-10618: Add UUID class, use in protocols

jolshan:
In order to support topic IDs, we need to create a public UUID class. This class will be used in protocols. This PR creates the class, modifies code to use the class in the message protocol and changes the code surrounding the existing messages/json that used the old UUID class.

SimpleExampleMessage was used only for testing, so all usages of UUID have been switched to the new class.

SubscriptionInfoData uses UUID for processId extensively. It also utilizes java.util.UUID implementation of Comparable<UUID> so that UUIDs can be ordered. I felt that this functionality was not necessary for the UUIDs used for topic IDs so I decided to convert to java.util.UUID on the boundary of SubscriptionInfoData. I was only able to find the sorting necessary for testing, though, so this still may be changed.

I also added tests for the methods of the new UUID class. The existing SimpleExampleMessage tests should be sufficient for testing the new UUID class in message protocols.


clients/src/main/java/org/apache/kafka/common/protocol/types/Type.java
```diff
@@ -331,14 +331,14 @@ public String documentation() {
     public static final DocumentedType UUID = new DocumentedType() {
         @Override
         public void write(ByteBuffer buffer, Object o) {
-            final java.util.UUID uuid = (java.util.UUID) o;
+            final org.apache.kafka.common.UUID uuid = (org.apache.kafka.common.UUID) o;
```
rajinisivaram: We have already imported UUID, do we need to qualify here? I can see it was this way before, but not sure why.

clients/src/main/java/org/apache/kafka/common/protocol/types/Type.java
```diff
@@ -331,14 +331,14 @@ public String documentation() {
...
             buffer.putLong(uuid.getMostSignificantBits());
             buffer.putLong(uuid.getLeastSignificantBits());
         }
 
         @Override
         public Object read(ByteBuffer buffer) {
-            return new java.util.UUID(buffer.getLong(), buffer.getLong());
+            return new org.apache.kafka.common.UUID(buffer.getLong(), buffer.getLong());
```
rajinisivaram: Same as before, why don't we just use `UUID` since package was imported?

clients/src/test/java/org/apache/kafka/common/UUIDTest.java
```diff
@@ -0,0 +1,60 @@
...
+        assertNotEquals(id1, id3);
+
+        assertEquals(UUID.ZERO_UUID.hashCode(), UUID.ZERO_UUID.hashCode());
+        assertEquals(id1.hashCode(), id2.hashCode());
+        assertNotEquals(id1.hashCode(), id3.hashCode());
+    }
+
+    @Test
+    public void testStringConversion() {
+        UUID id = UUID.randomUUID();
```
rajinisivaram, line 51: Should we have a separate unit test for `UUID.randomUUID` to verify it is not zero or the sentinel?

rajinisivaram (commented):
@jolshan Thanks for the PR, looks good. Left just a few minor comments.

clients/src/test/java/org/apache/kafka/common/UUIDTest.java
jolshan: I thought about that, but since the likelihood we would get the zero/sentinel id is so low even if they were allowed, I didn't know if the test would be useful. The sentinel ID is also a private variable, but I suppose I could just construct a new one if we want to test.

clients/src/main/java/org/apache/kafka/common/protocol/types/Type.java
jolshan: My best guess is that since the type is also called UUID, originally the qualification was to differentiate. But it doesn't seem to be needed, so I'll update this to just say UUID.

rajinisivaram (approved):
@jolshan Thanks for the updates, LGTM

jolshan:
Two tests failed, but they seem to be flaky based on other open PRs.
StoreQueryIntegrationTest.shouldQueryOnlyActivePartitionStoresByDefault
EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]

rajinisivaram:
@jolshan Thanks for the PR, merging to trunk.

PR #9457, opened 2020-10-20
MINOR: Trivial Cleanups

dongjinleekr:
Remove unused variables, methods, parameters, unthrown exceptions, and fixing typos.

### DIFF
--- a/streams/test-utils/src/test/java/org/apache/kafka/streams/test/TestRecordTest.java
+++ b/streams/test-utils/src/test/java/org/apache/kafka/streams/test/TestRecordTest.java
@@ -42,7 +42,7 @@ public class TestRecordTest {
     private final Headers headers = new RecordHeaders(
             new Header[]{
                 new RecordHeader("foo", "value".getBytes()),
-                new RecordHeader("bar", (byte[]) null),
+                new RecordHeader("bar", null),
                 new RecordHeader("\"A\\u00ea\\u00f1\\u00fcC\"", "value".getBytes())
             });
     private final Instant recordTime = Instant.parse("2019-06-01T10:00:00Z");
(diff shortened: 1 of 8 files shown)


dongjinleekr:
Here is some explanation about the last commit.

`ProducerRunnable#shutdown` method is never called, and the variable with the same name also not changed from its default value, `false`.

Furthermore, this variable does not work as intended; From the non-main thread's context, the modification of this variable's state may not be visible. To make it work as intended, it should be a volatile variable.

Since this variable is not used anymore, I thought removing it would be better.

dongjinleekr:
@chia7712 @dengziming Could you have a look? This PR includes several trivial fixes I met reading the code, but `TestRecordTest#testProducerRecord`'s case is a little bit serious.

Since the types of `ProducerRecord` and `TestRecord` are different, calling `assertNotEquals` for these instances always succeeds.

streams/test-utils/src/test/java/org/apache/kafka/streams/test/TestRecordTest.java
```diff
@@ -166,6 +167,9 @@ public void testProducerRecord() {
         final TestRecord<String, Integer> testRecord = new TestRecord<>(producerRecord);
         final TestRecord<String, Integer> expectedRecord = new TestRecord<>(key, value, headers, recordTime);
         assertEquals(expectedRecord, testRecord);
-        assertNotEquals(expectedRecord, producerRecord);
+        assertFalse(expectedRecord.getHeaders() == producerRecord.headers() &&
```
chia7712: Could we address this issue in separate PR? It seem to me this issue does not belong to "trivial cleanups" :)

dongjinleekr: @chia7712 I see. Let me open a separate PR. :smile:

dongjinleekr:
@chia7712 Separated the testing issue and added some additional trivial cleanups.

@hachikuji the last commit on `NetworkChannel` is related to your commit on KAFKA-10842, which deleted `NetworkChannel#receive` but omitted to update the Javadoc accordingly. Is this what you meant?

dongjinleekr:
ping @chia7712

dongjinleekr:
Rebased onto the latest trunk. cc/ @chia7712

dongjinleekr:
@chia7712 Ready to merge!

chia7712 (approved):
@dongjinleekr thanks for this patch. LGTM :)

PR #9476, opened 2020-10-21
MINOR: Refactor RaftClientTest to be used by other tests

jsancio:
There is a lot of functionality in KafkaRaftClientTest that is useful for writing other tests. Refactor that functionality into another class that can be reused in other tests.

comment to ping reviewers.*


raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -1729,126 +1695,156 @@ public void testLeaderGracefulShutdownTimeout() throws Exception {
...
         int otherNodeId = 1;
         int epoch = 5;
+        Set<Integer> voters = Utils.mkSet(LOCAL_ID, otherNodeId);
 
-        Set<Integer> voters = Utils.mkSet(localId, otherNodeId);
-        quorumStateStore.writeElectionState(ElectionState.withElectedLeader(epoch, otherNodeId, voters));
-        KafkaRaftClient client = buildClient(voters);
-        assertEquals(ElectionState.withElectedLeader(epoch, otherNodeId, voters), quorumStateStore.readElectionState());
+        RaftClientTestContext context = new RaftClientTestContext.Builder()
+            .updateQuorumStateStore(quorumStateStore -> {
```
hachikuji, line 1701: I think just about every call to `updateQuorumStateStore` is just writing an initial state. Seems like we can introduce a more direct option to the builder.

By the way, one of the annoyances is needing to provide `voters` through the initial state and through `build` below. Since we always need `voters`, maybe we can provide it in the builder constructor. That would allow us to add helpers to construct the state. For example, we could turn this into:

```java
new RaftClientTestContext.Builder(voters)
  .initializeAsFollower(epoch, otherNodeId)
  .build()
```

hachikuji: I think nearly every call to `updateQuorumStateStore` is just writing an initial state. Seems like we can introduce a more direct option to the builder.

By the way, one of the annoyances is needing to provide `voters` through the initial state and through `build` below. Since we always need `voters`, maybe we can provide it in the builder constructor. That would allow us to add helpers to construct the state. For example, we could turn this into:

```java
new RaftClientTestContext.Builder(voters)
  .initializeAsFollower(epoch, otherNodeId)
  .build()
```

Similarly, we could probably do state assertions in the test context as well and save the need to always pass through `voters` (e.g. we could have `context.assertFollower(epoch, leaderId)` instead of the cumbersome `assertEquals(ElectionState.withElectedLeader(epoch, otherNodeId, voters), context.quorumStateStore.readElectionState())`).

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -1729,126 +1695,156 @@ public void testLeaderGracefulShutdownTimeout() throws Exception {
...
+                assertDoesNotThrow(() -> {
+                    quorumStateStore.writeElectionState(ElectionState.withElectedLeader(epoch, otherNodeId, voters));
+                });
+            })
+            .build(voters);
+
+        assertEquals(ElectionState.withElectedLeader(epoch, otherNodeId, voters), context.quorumStateStore.readElectionState());
```
hachikuji: nit: we have assertions like this in many test cases. With a more direct api to update quorum state, we can move these assertions into that api.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -1536,67 +1522,70 @@ public void testObserverLeaderRediscoveryAfterRequestTimeout() throws Exception
...
 
         // Now shutdown
         int shutdownTimeoutMs = 5000;
-        CompletableFuture<Void> shutdownFuture = client.shutdown(shutdownTimeoutMs);
+        CompletableFuture<Void> shutdownFuture = context.client.shutdown(shutdownTimeoutMs);
 
         // We should still be running until we have had a chance to send EndQuorumEpoch
-        assertTrue(client.isShuttingDown());
-        assertTrue(client.isRunning());
+        assertTrue(context.client.isShuttingDown());
```
hachikuji, line 1467: nit: it is a tad vexing to see all the `context` prefixes. I guess another option might be to define `RaftClientTestContext` as an abstract class so that the test method can define the test behavior within the scope of a subclass.

For example:
```java
new RaftClientTestContext(builder) {
  void run() {
    assertTrue(client.isShuttingDown());
    ...
  }
}
```
Not required, just an alternative to consider.

hachikuji (commented):
Looks pretty good. Left a few suggestions.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
jsancio: Yeah. Let me play around with this.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -90,470 +76,480 @@
...
-        Records records = log.read(0, Isolation.UNCOMMITTED).records;
+        Records records = context.log.read(0, Isolation.UNCOMMITTED).records;
         RecordBatch batch = records.batches().iterator().next();
         assertTrue(batch.isControlBatch());
 
         Record record = batch.iterator().next();
         assertEquals(electionTimestamp, record.timestamp());
-        verifyLeaderChangeMessage(localId, Collections.singletonList(otherNodeId),
-            record.key(), record.value());
+        RaftClientTestContext.verifyLeaderChangeMessage(LOCAL_ID, Collections.singletonList(otherNodeId), record.key(), record.value());
```
jsancio: @hachikuji is this a bug? Shouldn't the leader (LOCAL_ID) always be a voter (the second argument for this function)?

jsancio: This is what the LeaderChangeMessage says:
```
      {"name": "VotedIds", "type": "[]int32", "versions": "0+",
       "about": "The IDs of the voters who voted for the current leader"},
```

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
jsancio: Thanks for the suggestion. I implemented this.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
jsancio: Thanks for the suggestion. I implemented this.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
jsancio: I thought about this last night and hack some solutions. I wasn't very pleased with the result. Let's explore this improvement in a future PR.

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
```diff
@@ -0,0 +1,648 @@
...
+import org.mockito.Mockito;
+import static org.apache.kafka.raft.RaftUtil.hasValidTopicPartition;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+final class RaftClientTestContext {
+    private static final int FETCH_MAX_WAIT_MS = 0;
+
+    static final TopicPartition METADATA_PARTITION = new TopicPartition("metadata", 0);
+    static final int LOCAL_ID = 0;
```
hachikuji: I'm somewhat inclined to add the local id to the builder rather than making it constant. It makes the builder a bit more self-contained. 

On a similar note, it would be nice to push the other static config values into the builder as well.

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
```diff
@@ -0,0 +1,648 @@
...
+            Metrics metrics = new Metrics(time);
+            MockNetworkChannel channel = new MockNetworkChannel();
+            LogContext logContext = new LogContext();
+            QuorumState quorum = new QuorumState(LOCAL_ID, voters, ELECTION_TIMEOUT_MS, FETCH_TIMEOUT_MS,
+                    quorumStateStore, time, logContext, random);
+
+            Map<Integer, InetSocketAddress> voterAddresses = voters.stream().collect(Collectors.toMap(
+                        Function.identity(),
+                        RaftClientTestContext::mockAddress
+                        ));
```
hachikuji: nit: this indentation looks kind of funky

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
hachikuji: Yeah, that's fair. It looks like the code current just includes all followers. I guess we need to carry over the voters into the `LeaderState` if we want to implement the description above. Let's open a separate sub-task for https://issues.apache.org/jira/browse/KAFKA-9876 and decide what we want to do there.

jsancio: Done. https://issues.apache.org/jira/browse/KAFKA-10634

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
jsancio: Done.

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
jsancio: Okay. I made it possible to easily add support of this in the future without breaking the existing tests.

We can make this changeable in the `Builder` as we need it.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -278,13 +265,13 @@ public void testEndQuorumStartsNewElectionAfterBackoffIfReceivedFromVotedCandida
             .withVotedCandidate(epoch, otherNodeId)
```
hachikuji, line 274: The pattern I had in mind was a little different. I was thinking something like this:

```java
        int localId = 0;
        int otherNodeId = 1;
        int epoch = 2;
        Set<Integer> voters = Utils.mkSet(localId, otherNodeId);

        RaftClientTestContext context = new RaftClientTestContext.Builder(localId, voters)
          .withVotedCandidate(epoch, otherNodeId)
          .build()
```

Then we don't have the awkwardness of the partial reliance on the static `LOCAL_ID`. I like this better because the ids have to be explicitly declared in each test case, which makes it easier to follow.

jsancio: Makes sense. I implemented this suggestion.

hachikuji (approved):
Thanks for the patch. LGTM.

hachikuji:
The build failure appears to be unrelated. 
```

[2020-10-23T01:08:22.296Z] [ERROR] Failed to execute goal org.apache.maven.plugins:maven-archetype-plugin:3.2.0:generate (default-cli) on project standalone-pom: A Maven project already exists in the directory /home/jenkins/workspace/Kafka_kafka-pr_PR-9476/streams/quickstart/test-streams-archetype/streams.examples -> [Help 1]
[2020-10-23T01:08:22.296Z] [ERROR] 
[2020-10-23T01:08:22.296Z] [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[2020-10-23T01:08:22.297Z] [ERROR] Re-run Maven using the -X switch to enable full debug logging.
[2020-10-23T01:08:22.297Z] [ERROR] 
[2020-10-23T01:08:22.297Z] [ERROR] For more information about the errors and possible solutions, please read the following articles:
[2020-10-23T01:08:22.298Z] [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
```

I will merge to trunk.

jsancio:
Thanks @hachikuji for reviewing and merging the PR.

PR #9467, opened 2020-10-21
KAFKA-10515: Properly initialize nullable Serdes with default values

thake:
This is a cherry pick of PR #9338 on branch 2.6.

Also introduced the notion of WrappingNullableSerdes (aligned to the concept
of WrappingNullableSerializer and WrappingNullableDeserializer) and centralized
initialization in WrappingNullables.

The added integeration test KTableKTableForeignKeyJoinDistributedTest tests
whether all serdes are now correctly set on all stream clients.

Reviewers: John Roesler <vvcephei@apache.org>

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9478, opened 2020-10-22
Fixed unit test mocks and incorrect required context serdes.

thake:
Fixed an incorrect requireNonNull on the context Serdes introduced with #9338 

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/internals/WrappingNullableSerde.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/internals/WrappingNullableSerde.java
@@ -60,9 +60,7 @@ public abstract class WrappingNullableSerde<T, InnerK, InnerV> implements Serde<
     }
 
     public void setIfUnset(final Serde<InnerK> defaultKeySerde, final Serde<InnerV> defaultValueSerde) {
-        Objects.requireNonNull(defaultKeySerde);
-        Objects.requireNonNull(defaultValueSerde);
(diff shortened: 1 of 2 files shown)


vvcephei:
Thanks for this, @thake!

I'm wondering if this is the right direction to go here. In a real Streams application, as well as in MockProcessorContext and TopologyTestDriver, the context serdes can never be null, so it seems to be a problem only for EasyMock tests.

It's not that "easy" to begin with to set up an EasyMock of a context for testing Streams components because the set of required interactions is both non-trivial and an internal detail. That's why we offer the MockProcessorContext, which provides default behavior that should be suitable for most tests and still lets you capture and assert the things you'd need to assert.

Have I misunderstood the problem here?

Thanks,
-John

vvcephei:
Sorry, this is getting complicated. I've just realized that this PR is specifically addressing some internal test failures that came up in the 2.6 port: https://github.com/apache/kafka/pull/9467 .

If that's the sole motivation, then I think we can close this PR and instead just fix the tests in https://github.com/apache/kafka/pull/9467

What do you think, @thake ?

thake:
You were right. Somehow I thought that the tests were right and that keySerde and valueSerde are nullable. It would really help to have a null safe API description using annotations.

Closing this PR because it addresses a non issue :)

vvcephei:
Thanks @thake , I agree. I think someone proposed to add Nullable/NotNull annotations to the project; I'm not sure what happened to that proposal.

PR #9479, opened 2020-10-22
KAFKA-10631: Handle ProducerFencedException on offset commit

cadonna:
The transaction manager does currently not handle producer fenced errors returned from a offset commit request.

This PR adds the handling of the producer fenced errors.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
@@ -1742,6 +1742,7 @@ public class TransactionManager {
     private boolean isFatalException(Errors error) {
         return error == Errors.TRANSACTIONAL_ID_AUTHORIZATION_FAILED
                    || error == Errors.INVALID_PRODUCER_EPOCH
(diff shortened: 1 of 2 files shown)


abbccdda (approved):
LGTM!

apurvam:
I can't believe this bug has been lurking for so long.

vvcephei:
LGTM! Thanks, @cadonna .

vvcephei:

ableegoldman:
@cadonna did we double-check with the core/clients people to make sure this is right? It's not immediately obvious to me that this is a problem in the TransactionManager and not just in the way we handle it within Streams. But maybe it is obvious to someone who looked into the code more than I have

(And I agree with Apurva, how did this not come up sooner? 🤔 )

ableegoldman:
Ah nevermind, I see that this was introduced when we swapped INVALID_PRODUCER_EPOCH for PRODUCER_FENCED so it makes sense that we should handle PRODUCER_FENCED in the same way as the old error code

cadonna:
Yeah, that was also my conclusion!
Additionally, I trust @abbccdda's approval since -- according to the git history -- he is familiar with the code.

PR #9481, opened 2020-10-22
KAFKA-10284: Disable static membership test in 2.4

vvcephei:
This test was fixed in https://github.com/apache/kafka/pull/9270
for 2.5+, but the code in 2.4 is too different to have a clean backport.
Rather than risk introducing a worse bug in 2.4, and also because the
probability of a new bugfix release for 2.4 seems low, I'm proposing
just to disable this test in the 2.4 branch.

### DIFF
--- a/tests/kafkatest/tests/streams/streams_static_membership_test.py
+++ b/tests/kafkatest/tests/streams/streams_static_membership_test.py
@@ -49,6 +49,9 @@ class StreamsStaticMembershipTest(Test):
                                            throughput=1000,
                                            acks=1)
 
+    # This test fails due to a bug that is fixed in 2.5+ (KAFKA-10284). We opted not to backport
+    # the fix to 2.4 and instead marked this test as ignored. If desired, the fix can be backported,
+    # but it is non-trivial to do so.
     @ignore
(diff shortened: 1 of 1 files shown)


vvcephei:
Hey @abbccdda , @feyman2016 , what do you think about this, as opposed to spending more time doing a backport for 2.4?

abbccdda (approved):
LGTM, do you mind adding a comment on the test including the reasoning for ignoring this test due to 10284? Not super important, may just save some time for the next developer.

tests/kafkatest/tests/streams/streams_static_membership_test.py
```diff
@@ -48,6 +49,7 @@ def __init__(self, test_context):
                                            throughput=1000,
                                            acks=1)
 
+    @ignore
```
vvcephei, line 55:
```suggestion
    # This test fails due to a bug that is fixed in 2.5+ (KAFKA-10284). We opted not to backport
    # the fix to 2.4 and instead marked this test as ignored. If desired, the fix can be backported,
    # but it is non-trivial to do so.
    @ignore
```

vvcephei:
Thanks @abbccdda !

PR #9482, opened 2020-10-22
KAFKA-10632; Raft client should push all committed data to state machines

hachikuji:
In #9418, we add a listener to the `RaftClient` interface. In that patch, we used it only to send commit notifications for writes from the leader. In this PR, we extend the `handleCommit` API to accept all committed data and we remove the pull-based `read` API. Additionally, we add two new callbacks to the listener interface in order to notify the state machine when the raft client has claimed or resigned leadership.

Finally, this patch allows the `RaftClient` to support multiple listeners. This is necessary for KIP-500 because we will have one listener for the controller role and one for the broker role.

Note this patch builds on top of #9418. Once merged, I will rebase this patch and remove draft status.


raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -206,30 +234,77 @@ private void updateLeaderEndOffsetAndTimestamp(
...
 
-    @Override
-    public LeaderAndEpoch currentLeaderAndEpoch() {
-        return quorum.leaderAndEpoch();
+    private void maybeFireHandleCommit(long highWatermark) {
+        maybeFireHandleCommit(listenerContexts, highWatermark);
+    }
+
+    private void maybeFireHandleCommit(List<ListenerContext> listenerContexts, long highWatermark) {
+        // TODO: When there are multiple listeners, we can cache reads to save some work
```
jsancio: I would remove this comment. We ca file an issue and fix it if this becomes a performance issue.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -206,30 +234,77 @@ private void updateLeaderEndOffsetAndTimestamp(
...
+        for (ListenerContext listenerContext : listenerContexts) {
+            OptionalLong nextExpectedOffsetOpt = listenerContext.nextExpectedOffset();
+            if (!nextExpectedOffsetOpt.isPresent()) {
+                return;
```
jsancio: Why a `return`? Did you mean to use `continue`? If this is suppose to be a `continue` then maybe we can `Optional.ifPresent`.

Same comment for one of the other overloaded `maybeFireHandleCommit`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -206,30 +234,77 @@ private void updateLeaderEndOffsetAndTimestamp(
...
+    }
+
+    private void maybeFireHandleClaim(LeaderState state) {
+        for (ListenerContext listenerContext : listenerContexts) {
+            int leaderEpoch = state.epoch();
+
+            // We can fire `handleClaim` as soon as the listener has caught
+            // up to the start of the leader epoch. This guarantees that the
+            // state machine has seen the full committed state before it becomes
+            // leader and begins writing to the log.
```
jsancio: Interesting. It is good to hide this logic from the state machine. Looking at the epoch and not at the LEO is okay because at this point we guarantee that the only records with that epoch are control records (e.g. LeaderChangedMessage).

I am wondering if the state machine may want to know this before it can process state machine requests. Maybe this is okay because the brokers/replicas will learn about the new leader through the `Fetch` and `BeginQuorum` protocol and not from the state machine (Kafka Controller) itself.

It is possible that the leader will receive Kafka Controller message from replicas/broker before it knows that it is leader. Most likely the Kafka Controller will reject them but the replicas/brokers need to keep retrying. This is specially important for heartbeat messages.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -975,12 +1029,9 @@ private boolean handleFetchResponse(
...
                     logger.info("Truncated to offset {} from Fetch response from leader {}",
                         truncationOffset, quorum.leaderIdOrNil());
-
-                    // Since the end offset has been updated, we should complete any delayed
-                    // reads at the end offset.
-                    fetchPurgatory.maybeComplete(
-                        new LogOffset(Long.MAX_VALUE, Isolation.UNCOMMITTED),
-                        currentTimeMs);
+                    // After truncation, we complete all pending reads in order to
+                    // ensure that fetches account for the
```
jsancio: Incomplete sentence.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1757,35 +1809,86 @@ public void complete() {
...
+            return lastAckedOffset;
+        }
+
+        /**
+         * Get the next expected offset, which might be larger than the last acked
+         * offset if there are inflight batches which have not been acked yet.
+         * Note that when fetching from disk, we may not know the last offset of
+         * inflight data until it has been processed by the state machine. In this case,
+         * we delay sending additional data until the state machine has read to the
+         * end and the last offset is determined.
```
jsancio, line 1837: Does this mean that in practice, follower will have at most two batches in flight?
1. The one that they are currently processing
2. If they read the last message/record in the batch then the next batch in the log?

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
```diff
@@ -369,6 +379,17 @@ public void transitionToLeader(long epochStartOffset) throws IOException {
         if (!candidateState.isVoteGranted())
             throw new IllegalStateException("Cannot become leader without majority votes granted");
 
+        // Note that the leader does not retain the high watermark that was known
+        // in the previous state. The purpose of this is to protect the monotonicity
```
jsancio: What is "this" in this sentence? epochStartOffset?

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
```diff
@@ -16,57 +16,75 @@
...
+         * after consuming the reader.
+         *
+         * Note that there is not a one-to-one correspondence between writes through
+         * {@link #scheduleAppend(int, List)} and this callback. The Raft implementation
+         * is free to batch together the records from multiple append calls provided
+         * that batch boundaries are respected. This means that each batch specified
+         * through {@link #scheduleAppend(int, List)} is guaranteed to be a subset of
+         * a batch provided by the {@link BatchReader}.
+         *
+         * @param reader reader instance which must be iterated
```
jsancio: nit: "... must be iterated and closed".

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
```diff
@@ -16,57 +16,75 @@
...
+     * in the future. There is no guarantee that appended records will be written to
+     * the log and eventually committed. However, it is guaranteed that if any of the
+     * records become committed, then all of them will be.
      *
-     * @return Current leader and epoch information
+     * @param epoch the current leader epoch
+     * @param records the list of records to append
+     * @return the offset within the current epoch that the log entries will be appended,
+     *         or null if the leader was unable to accept the write (e.g. due to memory
+     *         being reached).
```
jsancio: We should also mention that `MAX_VALUE` is return if the RaftClient is not the leader.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
```diff
@@ -16,57 +16,75 @@
...
      */
-    LeaderAndEpoch currentLeaderAndEpoch();
+    Long scheduleAppend(int epoch, List<T> records);
```
jsancio, line 93: I think it is okay as the fix may be non-trivial but technically `scheduleAppend` will accept records even if no `Listener` has received a `handleClaim`.

raft/src/main/java/org/apache/kafka/raft/RecordSerde.java
```diff
@@ -0,0 +1,63 @@
...
+     * @param out the output stream to write the record to
+     */
+    void write(T data, Object context, DataOutputStreamWritable out);
+
+    /**
+     *
+     * @param input
+     * @param size
+     * @return
+     */
```
jsancio, line 59: TODO: missing comments.

jsancio (commented):
Some of the comments may not apply to this PR. Didn't realize that this PR also includes another PR against trunk.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
hachikuji: Agreed. I added this here: https://github.com/apache/kafka/pull/9418/files#diff-ac850e29114f9b5a03aaf3ccb8f07feda8b48e5de6912f4c527b8477aa3d6cbcR60.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
hachikuji: Yeah, I don't see a strong need to be too strict about this for now.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: When catching up from the log, yes. However, I have implemented an optimization for writes from the leader. We save the original batch in memory so that it can be sent back to the state machine after the write is committed. In this case, we know the last offset of the batch, so we can have multiple inflight batches sent to the controller. This is nice because it means the elected controller will not have to read from disk.

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
hachikuji: It was intended to refer to the behavior of not retaining the high watermark from the previous sentence. I will attempt to clarify.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Good catch. I think that I forgot to update this when moving from a single listener to multiple listeners.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: I thought a little about it. Right now the state machine has just two states: 1) i am not a leader, and 2) i am a leader and have caught up with all committed data from previous epochs.  An alternative design is to fire `handleClaim` immediately and provide the starting offset of the leader epoch. Then the controller can wait until its state machine has caught up to that offset before starting to write data. In the end, I decided not to do it because it adds a third state and I did not expect the controller would be able to do anything useful in the additional state. The point about heartbeats is interesting, but even that seems tricky since the controller would not know if a broker had been fenced until it has caught up. I think the only thing the controller could do is hold the requests in purgatory, which might be better than letting them retry, but not sure it's worth it.

jsancio: Yeah. I was thinking of the same thing, "hold the requests in purgatory". But like you said, maybe this optimization is not worth the added complexity.

hachikuji: I guess let's keep this option in our back pocket for now.

clients/src/main/java/org/apache/kafka/common/metrics/stats/Percentiles.java
```diff
@@ -113,11 +113,11 @@ protected HistogramSample newSample(long timeMs) {
     protected void update(Sample sample, MetricConfig config, double value, long timeMs) {
         final double boundedValue;
         if (value > max) {
-            log.warn("Received value {} which is greater than max recordable value {}, will be pinned to the max value",
+            log.debug("Received value {} which is greater than max recordable value {}, will be pinned to the max value",
```
abbccdda, line 116: Why do we change these values to debug?

clients/src/main/java/org/apache/kafka/common/protocol/DataInputStreamReadable.java
```diff
@@ -0,0 +1,130 @@
...
+        try {
+            return input.readByte();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public short readShort() {
+        try {
```
abbccdda, line 44: nit: we could refactor out the try-catch logic.

clients/src/main/java/org/apache/kafka/common/protocol/DataInputStreamReadable.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public DataInputStreamReadable(DataInputStream input) {
+        this.input = input;
+    }
+
+    @Override
+    public byte readByte() {
+        try {
+            return input.readByte();
+        } catch (IOException e) {
```
abbccdda, line 37: Why do we wrap the IO exception here?

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
```diff
@@ -0,0 +1,116 @@
...
+        public int epoch() {
+            return epoch;
+        }
+
+        @Override
+        public String toString() {
+            return "Batch(" +
+                "baseOffset=" + baseOffset +
+                ", epoch=" + epoch +
+                ", records=" + records +
```
abbccdda, line 95: We couldn't be sure whether the records could be read as string?

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            Batch<?> batch = (Batch<?>) o;
+            return baseOffset == batch.baseOffset &&
+                epoch == batch.epoch &&
+                Objects.equals(records, batch.records);
```
abbccdda, line 106: Do we compare the value or the address of records?

raft/src/main/java/org/apache/kafka/raft/ExpirationService.java
```diff
@@ -16,11 +16,17 @@
...
-public class LogTruncationException extends KafkaException {
-
-    public LogTruncationException(String message) {
-        super(message);
-    }
+public interface ExpirationService {
+    /**
+     * Get a new completable future which will automatically expire with a
+     * {@link org.apache.kafka.common.errors.TimeoutException} after the provided
+     * timeout passes if it is not completed before then through some other means.
```
abbccdda: `after the provided timeout passes if it is not completed`/ `if not completed within the provided time limit`

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
```diff
@@ -369,6 +379,17 @@ public void transitionToLeader(long epochStartOffset) throws IOException {
         if (!candidateState.isVoteGranted())
             throw new IllegalStateException("Cannot become leader without majority votes granted");
 
+        // Note that the leader does not retain the high watermark that was known
+        // in the previous state. The reason it does not is to protect the monotonicity
```
abbccdda: `it does not` could be removed.

raft/src/main/java/org/apache/kafka/raft/internals/MemoryBatchReader.java
```diff
@@ -0,0 +1,69 @@
...
+    private final Iterator<Batch<T>> iterator;
+    private final long baseOffset;
+    private final long lastOffset;
+
+    public MemoryBatchReader(
+        List<Batch<T>> batches,
+        CloseListener<BatchReader<T>> closeListener
+    ) {
+        if (batches.isEmpty()) {
+            throw new IllegalArgumentException();
```
abbccdda: Should we add description to the exception?

raft/src/main/java/org/apache/kafka/raft/internals/RecordsBatchReader.java
```diff
@@ -0,0 +1,210 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.OptionalLong;
+
+public class RecordsBatchReader<T> implements BatchReader<T> {
```
abbccdda, line 38: How do we decide to use either record batch reader or memory batch reader?

raft/src/main/java/org/apache/kafka/raft/internals/ThresholdPurgatory.java
```diff
@@ -0,0 +1,94 @@
...
+            completion.completeExceptionally(exception);
+        }
+    }
+
+    @Override
+    public int numWaiting() {
+        return thresholdMap.size();
+    }
+
+    private static class ThresholdKey<T extends Comparable<T>> implements Comparable<ThresholdKey<T>> {
```
abbccdda, line 74: Do we have other types of threshold than Long?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -228,35 +234,81 @@ private void updateLeaderEndOffsetAndTimestamp(
...
+            long nextExpectedOffset = nextExpectedOffsetOpt.getAsLong();
+            if (nextExpectedOffset == baseOffset) {
+                listenerContext.fireHandleCommit(baseOffset, epoch, records);
+            }
+        }
+    }
+
+    private void maybeFireHandleClaim(LeaderState state) {
+        for (ListenerContext listenerContext : listenerContexts) {
+            int leaderEpoch = state.epoch();
```
abbccdda: Why do we need to repeatedly get leader epoch?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -228,35 +234,81 @@ private void updateLeaderEndOffsetAndTimestamp(
...
+
+            long nextExpectedOffset = nextExpectedOffsetOpt.getAsLong();
+            if (nextExpectedOffset < highWatermark) {
+                LogFetchInfo readInfo = log.read(nextExpectedOffset, Isolation.COMMITTED);
+                listenerContext.fireHandleCommit(nextExpectedOffset, readInfo.records);
+            }
+        }
+    }
+
+    private void maybeFireHandleCommit(long baseOffset, int epoch, List<T> records) {
```
abbccdda, line 282: It looks a bit weird to have two versions of fire handle commit, could we name them differently or comment about their distinctive logics for determining when to fire commit callback?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -329,8 +387,9 @@ private void appendLeaderChangeMessage(LeaderState state, long currentTimeMs) {
     }
 
     private void flushLeaderLog(LeaderState state, long currentTimeMs) {
-        log.flush();
+        // We update the end offset before flushing so that parked fetches can return sooner
```
abbccdda, line 390: Why would this work? If the flush wasn't successful, could the fetched records be invalidated later?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1688,14 +1694,39 @@ private long pollCurrentState(long currentTimeMs) throws IOException {
         }
     }
 
+    private void pollListeners() {
+        // Register any listeners added since the last poll
+        while (!pendingListeners.isEmpty()) {
```
abbccdda, line 1696: Do we anticipate use cases to add listeners on the fly? Right now I could only see one case in static context from test raft server.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1688,14 +1694,39 @@ private long pollCurrentState(long currentTimeMs) throws IOException {
...
+            maybeFireHandleCommit(listenersToUpdate, highWatermarkMetadata.offset);
+        });
+    }
+
     public void poll() throws IOException {
         GracefulShutdown gracefulShutdown = shutdown.get();
         if (gracefulShutdown != null) {
             pollShutdown(gracefulShutdown);
         } else {
+            pollListeners();
```
abbccdda, line 1721: Should we call `pollListeners` after `pollCurrentState` to get more recent updates quicker?

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
```diff
@@ -26,30 +24,53 @@
...
+         * After becoming a leader, the client is eligible to write to the log
+         * using {@link #scheduleAppend(int, List)}.
          *
-         * @param epoch the epoch in which the write was accepted
-         * @param lastOffset the offset of the last record in the record list
-         * @param records the set of records that were committed
+         * @param epoch the claimed leader epoch
          */
-        void handleCommit(int epoch, long lastOffset, List<T> records);
+        default void handleClaim(int epoch) {}
```
abbccdda, line 52: What about just name as `handleBecomeLeader`?

abbccdda (commented):
Thanks for the PR, left some reviews on the non-testing part.

core/src/main/scala/kafka/raft/TimingWheelExpirationService.scala
```diff
@@ -0,0 +1,61 @@
...
+    }
+    timer.add(future)
+    future
+  }
+
+  private class ExpiredOperationReaper extends ShutdownableThread(
+    name = "raft-expiration-reaper", isInterruptible = false) {
+
+    override def doWork(): Unit = {
+      timer.advanceClock(200L)
```
abbccdda: Could we make 200L a constant?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: I doubt we would use it in practice, though I guess it would open the door to changing roles dynamically, which might be interesting in the future. That said, it was simple to add and useful in testing since it gave me an easy way to initialize a state where a listener had not caught up.

clients/src/main/java/org/apache/kafka/common/metrics/stats/Percentiles.java
hachikuji: The `warn` seemed excessive for a metric update, which could be done very frequently. I looked over the code and it looks like we don't really use this outside of tests (and now the code added in this patch). I think the user should just understand the contract, which is that anything outside of the specified range gets rounded down.

clients/src/main/java/org/apache/kafka/common/protocol/DataInputStreamReadable.java
hachikuji: Yeah, I could introduce a helper with a lambda, but that would add some unnecessary garbage to the deserialization path. Although it is ugly, I think the duplication is not a big deal. We probably won't touch this class after it is created.

clients/src/main/java/org/apache/kafka/common/protocol/DataInputStreamReadable.java
hachikuji: `IOException` is checked, so we cannot raise it from the current `Readable` interface, so the options are to either add the exception to the `Readable` interface or to rethrow it as an unchecked exception. I went with the latter to reduce the impact and because I think we tend to prefer unchecked exceptions in general since checked exceptions sort of end up leaking their way through a bunch of call stacks.

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
hachikuji: Yeah, I'm relying on the `toString`. I think this is only useful for debugging.

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
hachikuji: We are relying on the standard `equals` method. I think it's up to the user of the api to ensure a reasonable implementation if they expect to rely on batch equality. The raft implementation does not set any expectations on record equality, but it is useful in testing where we can control the record type.

raft/src/main/java/org/apache/kafka/raft/internals/RecordsBatchReader.java
hachikuji: `MemoryBatchReader` is only used for writes from the leader. We retain the original records from the call to `scheduleAppend` and send them to the listener in `handleCommit`. This is useful because it ensures the active controller will not need to read from disk.

raft/src/main/java/org/apache/kafka/raft/internals/ThresholdPurgatory.java
hachikuji: Not at the moment. I guess your point is that we might be able to drop the generics, which is fair. I think we can also drop the `FuturePurgatory` interface. Is it ok if we save this for a follow-up?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: I will move this outside the loop.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: The only difference is the input. I will add some comments to try and clarify the usage.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Yeah, it's ok for followers to see uncommitted or even unflushed data. The main thing is that we avoid advancing the high watermark until the fsync completes. Note that this is the main reason that we had to do KAFKA-10527. Without this fix, it was possible for the leader to continue in the same epoch after a start, which means that it could lose and overwrite unflushed data.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Hmm, that's a fair question. I think the listeners will tend to get new data in two cases: 1) high watermark advanced, or 2) a previous read completes. In the first case, the high watermark only advances in response to a request, so there should be no delay. In the second case, we call `wakeup()` to take us out of the network poll, so I think there also should be no delay. Can you think of a case where there would be a delay?

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
hachikuji: Yeah, I considered using `handleBecomeLeader` and `handleResignLeader`. In the end, I decided to use the more concise `handleClaim` and `handleResign` names which are used in the kip-500 branch.

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
```diff
@@ -0,0 +1,116 @@
...
+ * down to registered {@link RaftClient.Listener} instances.
+ *
+ * The advantage of hiding the consumption of committed batches behind an interface
+ * is that it allows us to push blocking operations such as reads from disk outside
+ * of the Raft IO thread. This helps to ensure that a slow state machine will not
+ * affect replication.
+ *
+ * @param <T> record type (see {@link org.apache.kafka.raft.RecordSerde})
+ */
+public interface BatchReader<T> extends Iterator<BatchReader.Batch<T>>, Closeable {
```
jsancio: How about extending `java.lang.Iterable<...>` instead of `Iterator`? This will allow the user to use `BatchReader::forEach`.

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
```diff
@@ -0,0 +1,116 @@
...
+     * @return optional last offset
+     */
+    OptionalLong lastOffset();
+
+    /**
+     * Close this reader. It is the responsibility of the {@link RaftClient.Listener}
+     * to close each reader passed to {@link RaftClient.Listener#handleCommit(BatchReader)}.
+     */
+    @Override
+    void close();
```
jsancio, line 61: If we are overriding `close` to not throw an `IOException`, then maybe extending `java.lang.AutoCloseable` instead of `java.io.Closeable` is better.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Yeah. You want to force an epoch change in the case that the old leader stays leader and partially replicated data was lost. This would force followers to truncate to the new leader's log state.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1017,12 +1028,9 @@ private boolean handleFetchResponse(
...
                         truncationOffset, quorum.leaderIdOrNil());
-
-                    // Since the end offset has been updated, we should complete any delayed
-                    // reads at the end offset.
-                    fetchPurgatory.maybeComplete(
-                        new LogOffset(Long.MAX_VALUE, Isolation.UNCOMMITTED),
-                        currentTimeMs);
+                    // After truncation, we complete all pending reads in order to
+                    // ensure that fetches account for the updated log end offset
+                    fetchPurgatory.completeAll(currentTimeMs);
```
jsancio: With the new `Listener` when is this not a noop? Looking at the code, we only add entries to `fetchPurgatory`when the replica is a leader and it receives a `Fetch` request.

In previous `read` based implementation we needed this because I think the `fetchPurgatory` contained both `reads` and `Fetch`.

I think the part that is missing is that the old leader should `fetchPurgatory.completeAll` when it loses leadership.

What do you think?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -228,35 +234,80 @@ private void updateLeaderEndOffsetAndTimestamp(
...
         state.highWatermark().ifPresent(highWatermark -> {
-            logger.debug("High watermark updated to {}", highWatermark);
+            logger.debug("Leader high watermark updated to {}", highWatermark);
             log.updateHighWatermark(highWatermark);
-
-            LogOffset offset = new LogOffset(highWatermark.offset, Isolation.COMMITTED);
-            appendPurgatory.maybeComplete(offset, currentTimeMs);
-            fetchPurgatory.maybeComplete(offset, currentTimeMs);
+            appendPurgatory.maybeComplete(highWatermark.offset, currentTimeMs);
+            maybeFireHandleCommit(highWatermark.offset);
```
jsancio, line 259: Is this needed because users of `KafkaRaftClient` can call `::register` before `::initizalize`? When else would this result on a call to `Listener::handleCommit`?

jsancio: Never mind. I think this can happens when the replica changes state from follower to leader.

I was having an issue if doing this would cause both `appendPurgatory.maybeComplete` and `maybeFireHandleCommit` to fire `Listener.handleCommit` for the same listener.

I don't think this can happened based on how `ListenerContext` is managing the `nextExpectedOffset`. If `appendPurgator.maybeComplete` fires then that means that `nextExpectedOffset` is greater that the high watermark. Since the `nextExpectedOffset` is greater than the high watermark then `maybeFireHandleCommit` will not fire.

I actually think that this order is important. Should we write a comment on the code explaining this if you agree with my analysis?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: That looks correct to me with the clarification that "in response to a request" has two cases:

1. The leader handles a fetch request. This implementation calls "update high watermark
2. The follower handle a fetch response. This implementation calls "update high watermark"

I think that `pollListeners` should only fire a `Listener::handleCommit` for new listeners in `pendingListeners`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1778,4 +1808,98 @@ public void complete() {
         }
     }
 
+    private final class ListenerContext implements CloseListener<BatchReader<T>> {
+        private final RaftClient.Listener<T> listener;
+        private BatchReader<T> lastSent = null;
+        private long lastAckedOffset = 0;
```
jsancio, line 1817: I would document that `synchronized` is protecting `lastSent` and `lastAckedOffset`.

`claimedEpoch` is okay because it is only used by the thread calling `RaftClient::poll`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1778,4 +1808,98 @@ public void complete() {
...
+        private int claimedEpoch = 0;
```
jsancio: Interesting that Java let's the outer class (`KafkaRaftClient`) access this private method. This makes reasoning about the concurrency non-trivial.

raft/src/main/java/org/apache/kafka/raft/ExpirationService.java
```diff
@@ -16,11 +16,17 @@
...
+    /**
+     * Get a new completable future which will automatically fail exceptionally with a
+     * {@link org.apache.kafka.common.errors.TimeoutException} if not completed before
+     * the provided time limit expires.
+     *
+     * @param timeoutMs the duration in milliseconds before the future is completed exceptionally
+     * @param <T> arbitrary future type (the service must set no expectation on the this type)
+     * @return the completable future
+     */
+    <T> CompletableFuture<T> await(long timeoutMs);
```
jsancio: How about `failAfter`?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1778,4 +1808,98 @@ public void complete() {
...
+            BatchReader.Batch<T> batch = new BatchReader.Batch<>(baseOffset, epoch, records);
+            MemoryBatchReader<T> reader = new MemoryBatchReader<>(Collections.singletonList(batch), this);
+            fireHandleCommit(reader);
+        }
+
+        private void fireHandleCommit(BatchReader<T> reader) {
+            synchronized (this) {
+                this.lastSent = reader;
+            }
+            listener.handleCommit(reader);
```
jsancio, line 1881: This applies to all of the `listener.handle...` on this file.

What are your thoughts on the `Listener` throwing an exception? I think with this implementation it will unwind all the way until `KafakRaftClient::poll`. Should we catch all non-fatal exceptions here and log an error instead?

jsancio (commented):
Thanks for the changes @hachikuji. Looking good.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
abbccdda: Sounds fair.

abbccdda (approved):
I think the testing part looks good, should be mergeable after addressing Jose's comments.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: > With the new Listener when is this not a noop? Looking at the code, we only add entries to fetchPurgatorywhen the replica is a leader and it receives a Fetch request.

Yeah, that's fair. I don't think we can truncate unless we are a follower and that implies we already cleared the purgatory in `onBecomeFollower`. So I think you are right that we are safe to remove this, though we'll probably need to add it back once we have follower fetching. 

> I think the part that is missing is that the old leader should fetchPurgatory.completeAll when it loses leadership.

I had considered this previously and decided to leave the fetches in purgatory while the election was in progress to prevent unnecessary retries since that is all the client can do while waiting for the outcome. On the other hand, some of the fetches in purgatory might be from other voters. It might be better to respond more quickly so that there are not any unnecessary election delays. I'd suggest we open a separate issue to consider this.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: I will add a comment. I agree it is a subtle point.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Let me add a helper to `ListenerContext` so that we can keep the field encapsulated.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Hmm... That's a good question. I guess the issue is that the listener will then be in an unknown state. Should the IO thread keep sending it updates or should it mark it as failed? This comes back to something I have been wondering in the KIP-500 world. Do we want the process to stay active if either the controller or broker listeners have failed or would it be better to shutdown? At the moment, I am leaning toward the latter. In any case, I suggest we let the errors propagate for now and file a jira to reconsider once we are closer to integration. Does that sound fair?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: > I had considered this previously and decided to leave the fetches in purgatory while the election was in progress to prevent unnecessary retries since that is all the client can do while waiting for the outcome. On the other hand, some of the fetches in purgatory might be from other voters. It might be better to respond more quickly so that there are not any unnecessary election delays. I'd suggest we open a separate issue to consider this.

Sounds good to create a Jira for this.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: >  At the moment, I am leaning toward the latter. In any case, I suggest we let the errors propagate for now and file a jira to reconsider once we are closer to integration. Does that sound fair?

Sounds fair to create a Jira and revisit this later.

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
hachikuji: I was thinking about this and having some trouble putting my hesitation into words. I guess there are two main reasons why I preferred the `Iterator`:

1. We need some way to communicate iteration progress back to the IO thread. It is probably still possible to do this with a layer of indirection through `Iterable`, but it seemed more natural if the IO thread had access to the `Iterator` object that was used by the listener.
2. Iteration is not necessarily cheap since it might involve reading from disk. I thought we may as well enforce a single-read pattern.

jsancio: > We need some way to communicate iteration progress back to the IO thread. It is probably still possible to do this with a layer of indirection through Iterable, but it seemed more natural if the IO thread had access to the Iterator object that was used by the listener.

Okay. Specifically, you are saying that this would be difficult to implement with an `Iterable`:
```java
lastReturnedOffset = res.lastOffset();
```
https://github.com/apache/kafka/pull/9482/files/867650fa2344497ac3f3505bd5058f2eae0cc0c4#diff-37728c07e52382d38a9db6f655c3921d274c4277a8909d7613fc433d6bf69636R140

hachikuji: Yeah. Not so much difficult, but awkward because of the one-to-many association between the `Iterable` and `Iterator`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Filed https://issues.apache.org/jira/browse/KAFKA-10677.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Filed https://issues.apache.org/jira/browse/KAFKA-10676.

PR #9486, opened 2020-10-23
KAFKA-9381: Fix publishing valid scaladoc for streams-scala

bbejeck:

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -919,7 +919,7 @@ project(':core') {
     from "$rootDir/LICENSE"
     from "$rootDir/NOTICE"
     from(configurations.runtime) { into("libs/") }
-    from(configurations.archives.artifacts.files) { into("libs/") }
+    from(configurations.archives.artifacts.files.filter({f -> f.name.matches(".*javadoc.*")})) { into("libs/") }
     from(project.siteDocsTar) { into("site-docs/") }
     from(project(':tools').jar) { into("libs/") }
     from(project(':tools').configurations.runtime) { into("libs/") }
@@ -942,7 +942,7 @@ project(':core') {
     from(project(':streams').jar) { into("libs/") }
     from(project(':streams').configurations.runtime) { into("libs/") }
     from(project(':streams:streams-scala').jar) { into("libs/") }
-    from(project(':streams:streams-scala').configurations.archives.artifacts.files.filter { file -> file.name.matches(".*scaladoc.*")}) { into("libs/") }
+    from(project(':streams:streams-scala').configurations.archives.artifacts.files.filter { f -> f.name.matches(".*scaladoc.*")}) { into("libs/") }
     from(project(':streams:streams-scala').configurations.runtime) { into("libs/") }
     from(project(':streams:test-utils').jar) { into("libs/") }
     from(project(':streams:test-utils').configurations.runtime) { into("libs/") }
--- a/build.gradle
+++ b/build.gradle
@@ -728,14 +728,13 @@ project(':core') {
   println "Building project 'core' with Scala version ${versions.scala}"
 
   apply plugin: 'scala'
-  /*
-    scaladoc generation is now specifically configured at
-    the sub-module level with an artifacts block (cf. see streams-scala).
-    This block is here to ensure that if scaladoc is enabled globally
-    we do not generate scaladoc for core as it's not public API
-  */
+  
+  // scaladoc generation is configured at the sub-module level with an artifacts
+  // block (cf. see streams-scala). If scaladoc generation is invoked explicitly
+  // for the `core` module, this ensures the generated jar doesn't include scaladoc
+  // files since the `core` module doesn't include public APIs.
   scaladoc {
-      enabled = false
+    enabled = false
   }
   if (userEnableTestCoverage)
     apply plugin: "org.scoverage"
--- a/build.gradle
+++ b/build.gradle
@@ -482,9 +482,6 @@ subprojects {
     //documentation task should also trigger building scala doc jar
     docsJar.dependsOn scaladocJar
 
-    artifacts {
-      archives scaladocJar
-    }
   }
 
   tasks.withType(ScalaCompile) {
@@ -1444,7 +1441,11 @@ project(':streams:streams-scala') {
     dependsOn 'copyDependantLibs'
   }
 
-  test.dependsOn(':spotlessScalaCheck')
+  artifacts {
+      archives scaladocJar
+  }
+
+    test.dependsOn(':spotlessScalaCheck')
 }
 
 project(':streams:test-utils') {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -1439,7 +1440,7 @@ project(':streams:streams-scala') {
   }
 
   jar {
-    dependsOn 'copyDependantLibs'
+    dependsOn 'copyDependantLibs' , 'scaladocJar'
```
ijuma: Hmm, seems weird to require scaladoc when building jars.

bbejeck: @ijuma good point, I've removed it and still get the required doc jars

bbejeck:
Results of rendering contents of scaladoc jar

Screen Shot 2020-10-23 at 2 59 32 PM

ijuma:
We probably only want to publish scaladoc for kafka streams since it's the only module that has a public Scala API.

mjsax:
@ijuma Correct! That is what KAFKA-9381 actually is all about.

LGTM.

bbejeck:
>We probably only want to publish scaladoc for kafka streams since it's the only module that has a public Scala API.

ack, working on that now

build.gradle
```diff
@@ -942,6 +942,7 @@ project(':core') {
     from(project(':streams').jar) { into("libs/") }
     from(project(':streams').configurations.runtime) { into("libs/") }
     from(project(':streams:streams-scala').jar) { into("libs/") }
+    from(project(':streams:streams-scala').configurations.archives.artifacts.files.filter { file -> file.name.matches(".*scaladoc.*")}) { into("libs/") }
```
mjsax: Do we actually need to do this? I don't see any other line here that would copy JavaDocs-jar into the `lib` directory?

\cc @ijuma

bbejeck:
Need to test locally that the `uploadArchives` task triggers `archive` blocks in the build.gradle file.  If correct then the `kafka-streams-scalaXXXX-scaladoc.jar` should get created properly, and with scaladoc disabled for core, we won't publish non-public scaladoc

bbejeck:
I was able to test this locally by setting the `mavenUrl` variable in `~/.gradle/gradle.properties` file to `/Users/bill/Downloads/testRepo/` on my laptop.  

Running the `uploadArchives` task locally produced these scaladoc jar files.

```
find . -name '*scaladoc.jar' | xargs ls -la
-rw-r--r--  1 bill  staff  1334323 Oct 26 10:48 ./org/apache/kafka/kafka-streams-scala_2.12/2.8.0-SNAPSHOT/kafka-streams-scala_2.12-2.8.0-20201026.144817-1-scaladoc.jar
-rw-r--r--  1 bill  staff  1326180 Oct 26 10:50 ./org/apache/kafka/kafka-streams-scala_2.13/2.8.0-SNAPSHOT/kafka-streams-scala_2.13-2.8.0-20201026.145048-1-scaladoc.jar
-rw-r--r--  1 bill  staff     9585 Oct 26 10:47 ./org/apache/kafka/kafka_2.12/2.8.0-SNAPSHOT/kafka_2.12-2.8.0-20201026.144723-1-scaladoc.jar
-rw-r--r--  1 bill  staff     9585 Oct 26 10:50 ./org/apache/kafka/kafka_2.13/2.8.0-SNAPSHOT/kafka_2.13-2.8.0-20201026.145017-1-scaladoc.jar
```
So the changes in this PR will produce a valid scaladoc jar file `streams-scala`, which has a public Scala API, but not for core, which is what we want.

So I think this PR is good now.

ping @ijuma, @mjsax  for another review

ijuma:
What's in the following two files?

```text
-rw-r--r--  1 bill  staff     9585 Oct 26 10:47 ./org/apache/kafka/kafka_2.12/2.8.0-SNAPSHOT/kafka_2.12-2.8.0-20201026.144723-1-scaladoc.jar
-rw-r--r--  1 bill  staff     9585 Oct 26 10:50 ./org/apache/kafka/kafka_2.13/2.8.0-SNAPSHOT/kafka_2.13-2.8.0-20201026.145017-1-scaladoc.jar
```

Can we disable them altogether?

bbejeck:
> What's in the following two files?

Just the manifest and the license 

```
jar tf ./org/apache/kafka/kafka_2.13/2.8.0-SNAPSHOT/kafka_2.13-2.8.0-20201026.145017-1-scaladoc.jar
META-INF/
META-INF/MANIFEST.MF
LICENSE
NOTICE
```
I'd have to look into what it would take to disable creating the scaladoc jar file, but since it's scala project I'm not sure.

bbejeck:
@ijuma 

This latest change gets us where I think we want to be.  

Using the same test procedure from before, we only have scaladoc files for streams-scala.

```
find . -name '*scaladoc.jar' | xargs ls -la
-rw-r--r--  1 bill  staff  1334323 Oct 26 11:53 ./org/apache/kafka/kafka-streams-scala_2.12/2.8.0-SNAPSHOT/kafka-streams-scala_2.12-2.8.0-20201026.155354-1-scaladoc.jar
-rw-r--r--  1 bill  staff  1326180 Oct 26 11:56 ./org/apache/kafka/kafka-streams-scala_2.13/2.8.0-SNAPSHOT/kafka-streams-scala_2.13-2.8.0-20201026.155648-1-scaladoc.jar
```

The `build.gradle` file now only generates scaladoc archives specifically for streams-scala vs. all sub-projects using the scala plugin.  
In the future, if we need to generate public scaladoc for another module, we'll have to make the same change to that project as we did for streams-scala.

build.gradle
```diff
@@ -731,6 +728,9 @@ project(':core') {
   println "Building project 'core' with Scala version ${versions.scala}"
 
   apply plugin: 'scala'
+  scaladoc {
+      enabled = false
+  }
```
ijuma, line 738: Do we still need this with your latest change? Also, can you please check the indenting generally? It should be two spaces.

bbejeck: >Do we still need this with your latest change?

I thought about taking it out, but I left it in as a fail-safe, thinking that if the build file changes somehow enabling scaladoc for all submodules, `core` will only provide the scaladoc jar file with the license and manifest.

I don't have a strong opinion on this, so I can remove if you want me to.

Ack on the indenting

bbejeck:
updated this \cc @ijuma

build.gradle
ijuma: ok, let's maybe add a comment explaining it then.

bbejeck: ack

bbejeck:
updated this PR

\cc @ijuma

build.gradle
ijuma: I tweaked it a bit

ijuma (approved):
LGTM

bbejeck:
test failure unrelated and is a known flaky test `org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]`

bbejeck:
Merged #9486 into trunk

bbejeck:
cherry-picked to 2.7

PR #9489, opened 2020-10-24
MINOR: demote "Committing task offsets" log to DEBUG

ableegoldman:
This message absolutely floods the logs, especially in an eos application where the commit interval is just 100ms. It's definitely a useful message but I don't think there's any justification for it being at the INFO level when it's logged 10 times a second.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -692,6 +692,7 @@ public class StreamThread extends Thread {
                     break;
                 } else if (Math.max(now - lastPollMs, 0) > maxPollTimeMs / 2) {
                     numIterations = numIterations > 1 ? numIterations / 2 : numIterations;
+                    log.info("Pause processing to call poll before we hit the max.poll.interval.ms");
                     break;
                 } else if (punctuated > 0 || committed > 0) {
                     numIterations = numIterations > 1 ? numIterations / 2 : numIterations;
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -692,7 +692,6 @@ public class StreamThread extends Thread {
                     break;
                 } else if (Math.max(now - lastPollMs, 0) > maxPollTimeMs / 2) {
                     numIterations = numIterations > 1 ? numIterations / 2 : numIterations;
-                    log.info("Pause processing to call poll before we hit the max.poll.interval.ms");
                     break;
                 } else if (punctuated > 0 || committed > 0) {
                     numIterations = numIterations > 1 ? numIterations / 2 : numIterations;
@@ -704,6 +703,14 @@ public class StreamThread extends Thread {
             // we record the ratio out of the while loop so that the accumulated latency spans over
             // multiple iterations with reasonably large max.num.records and hence is less vulnerable to outliers
             taskManager.recordTaskProcessRatio(totalProcessLatency, now);
+
+            // Only log this when we actually processed some records to avoid flooding INFO-level logs when there are
+            // no records on the input topics
+            if (totalProcessed > 0) {
+                log.info("Finished processing {} records, going to poll again", totalProcessed);
+            } else {
+                log.debug("Processed zero records, going to poll again");
+            }
         }
 
         now = time.milliseconds();
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -630,6 +630,8 @@ public class StreamThread extends Thread {
         advanceNowAndComputeLatency();
 
         int totalProcessed = 0;
+        int totalPunctuated = 0;
+        int totalCommitted = 0;
         long totalCommitLatency = 0L;
         long totalProcessLatency = 0L;
         long totalPunctuateLatency = 0L;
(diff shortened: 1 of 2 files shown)


ableegoldman:
@guozhangwang @vvcephei  WDYT? I'm generally all for more logs but this is pretty extreme 🙂

ghmulti:
hello @ableegoldman, I wonder why did you close the pr?

ableegoldman:
Eh, I'm going back and forth on this. It's definitely valuable information to have in the absence of any other logging within the main Streams loop at the INFO level

ghmulti:
There is a lot of useful information on an INFO level, but as you mentioned - having that line from several threads with 100ms frequency makes it way too noisy (especially when container is running on a cluster with log forwarder, which leads to waste of resources). Even though it is not complicated to change the log level configuration for that particular class, it is kind of inconvenient and prevents from using out-of-the-box default configs.

ableegoldman:
That's fair. Ok I'll go forward with demoting this and just add a new INFO-level log that's less frequent, maybe at the max poll interval

ableegoldman:
Thoughts? @wcarlson5 @vvcephei @abbccdda

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -703,6 +703,14 @@ void runOnce() {
             // we record the ratio out of the while loop so that the accumulated latency spans over
             // multiple iterations with reasonably large max.num.records and hence is less vulnerable to outliers
             taskManager.recordTaskProcessRatio(totalProcessLatency, now);
+
+            // Only log this when we actually processed some records to avoid flooding INFO-level logs when there are
+            // no records on the input topics
+            if (totalProcessed > 0) {
```
abbccdda: Could we log the owned task here as well?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -703,6 +703,14 @@ void runOnce() {
...
+                log.info("Finished processing {} records, going to poll again", totalProcessed);
+            } else {
+                log.debug("Processed zero records, going to poll again");
```
wcarlson5: Is it important to make the distinction between "Processed zero records, going to poll again" and "Finished processing 0 records, going to poll again"?

Also if we are not saying its processed anything maybe we should log what it is pulling from

ableegoldman: Reworked it a bit, lmk what you think

wcarlson5: Looks good

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -703,6 +703,16 @@ void runOnce() {
...
             // multiple iterations with reasonably large max.num.records and hence is less vulnerable to outliers
             taskManager.recordTaskProcessRatio(totalProcessLatency, now);
+
+            // Only log this at INFO when we actually processed some records to avoid flooding INFO-level logs when
+            // there are no records on the input topics
+            if (totalProcessed > 0) {
+                log.info("Processed {} records for tasks {}, polling for more records now",
+                         totalProcessed, taskManager.activeTaskIds());
+            } else {
+                log.debug("Processed {} records for tasks {}, polling for more records now",
```
vvcephei: This line looks about the same as L665. I'm wondering if we really need the "else" in this case.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -703,6 +703,16 @@ void runOnce() {
             // we record the ratio out of the while loop so that the accumulated latency spans over
             // multiple iterations with reasonably large max.num.records and hence is less vulnerable to outliers
             taskManager.recordTaskProcessRatio(totalProcessLatency, now);
+
+            // Only log this at INFO when we actually processed some records to avoid flooding INFO-level logs when
+            // there are no records on the input topics
+            if (totalProcessed > 0) {
+                log.info("Processed {} records for tasks {}, polling for more records now",
```
vvcephei: Since this is at the end of runOnce, I'm wondering if it also makes sense to log whether we committed or punctuated, and whether/how many records we polled at the beginning of the method.

Basically, it seems like, if it's a good idea to log some information once per cycle, then it's probably a good idea to summarize everything you'd want to know.

vvcephei (commented):
Thanks, @ableegoldman !

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
ableegoldman: Fair enough

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
ableegoldman: Yeah good call. I promoted one of the debug logs you added earlier to log the number of records polled at INFO. I think this should be reasonable but we can always demote it again if it's still showing up too frequently

vvcephei (approved):
Thanks, @ableegoldman !

ableegoldman:
Two flaky tests: 
`PurgeRepartitionTopicIntegrationTest.shouldRestoreState`
`SslSelectorTest.testCloseOldestConnection`

ableegoldman:
Merged to trunk. Waiting to cherrypick to 2.7 branch until 2.7.0 release has completed

ableegoldman:
Cherrypicked to 2.7

PR #9494, opened 2020-10-24
KAFKA-10641: ACL Command Exit properly always with error code

senthilm-ms:
When using ACL Command with SSL mode, the process is not terminating after successful ACL operation.


senthilm-ms:
@junrao @mjsax @guozhangwang

chia7712:
>  the process is not terminating after successful ACL operation.

@senthilm-ms Pardon me. Are there running thread which is not terminated?

PR #9495, opened 2020-10-24
KAFKA-10642: Expose the real stack trace if any exception occurred during SSL Client Trust Verification in extension

senthilm-ms:
If there is any exception occurred in the custom implementation of client trust verification (i.e. using security.provider), the inner exception is suppressed or hidden and not logged to the log file...

@junrao @mjsax @guozhangwang

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/security/ssl/SslFactory.java
+++ b/clients/src/main/java/org/apache/kafka/common/security/ssl/SslFactory.java
@@ -95,6 +95,7 @@ public class SslFactory implements Reconfigurable, Closeable {
             try {
                 SslEngineValidator.validate(builder, builder);
             } catch (Exception e) {
+                log.error("SSLFactory configure failed during validate", e);
                 throw new ConfigException("A client SSLEngine created with the provided settings " +
                         "can't connect to a server SSLEngine created with those settings.", e);
             }


junrao (commented):
@senthilm-ms : Thanks for the PR. When we log an exception, the cause of the exception is logged too, right?

senthilm-ms:
> @senthilm-ms : Thanks for the PR. When we log an exception, the cause of the exception is logged too, right?

Yes, however as it is rethrowing new exception with generic message that hides the real exception message. Also the stack trace in the log is against the new exception... with that said, the original exception is totally suppressed...

junrao:
@rajinisivaram : What do you think? Thanks.

senthilm-ms:
> @rajinisivaram : What do you think? Thanks.

@junrao @rajinisivaram - take a look at the stack trace mentioned in the JIRA bug where you don't see the extension/custom implementation stack trace

chia7712:
@senthilm-ms Could you share the screenshot to me? The error stack in the description is a bit hard to read. Also, why ```[2020-05-13 14:30:26,892] ERROR [KafkaServer id=423810470] Fatal error during ``` is printed two times?

clients/src/main/java/org/apache/kafka/common/security/ssl/SslFactory.java
```diff
@@ -95,6 +95,7 @@ public void configure(Map<String, ?> configs) throws KafkaException {
             try {
                 SslEngineValidator.validate(builder, builder);
             } catch (Exception e) {
+                log.error("SSLFactory configure failed during validate", e);
                 throw new ConfigException("A client SSLEngine created with the provided settings " +
```
rajinisivaram, line 99: We should probably change the exception class here since `ConfigException` takes a config name and value while this one is passing in the exception message and cause like other exception constructors. Then existing log entry would be sufficient as well.

PR #9498, opened 2020-10-25
MINOR: call super.close() when closing RocksDB options

rodesai:
Call super.close when closing rocksdb options. If we don't, then closing a state
store will leak the underlying rocksdb options object. This is actually somewhat
costly, since the default options object allocates it's own cache instance.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapter.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapter.java
@@ -1393,9 +1393,10 @@ public class RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapter extends 
 
     @Override
     public void close() {
-        // ColumnFamilyOptions should be closed last
+        // ColumnFamilyOptions should be closed after DBOptions
         dbOptions.close();
         columnFamilyOptions.close();
+        // close super last since we initialized it first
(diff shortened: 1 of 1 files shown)


streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapter.java
```diff
@@ -1396,5 +1396,6 @@ public void close() {
         // ColumnFamilyOptions should be closed last
         dbOptions.close();
         columnFamilyOptions.close();
+        super.close();
```
mjsax, line 1400: Are there any ordering constraints to call the three `close` methods?

mjsax (approved):

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapter.java
rodesai: I don't think it matters in this case as the parent's Options handle isn't actually used. Still, I've maintained the inverse of the initialization order here.

rodesai: updated the comment

guozhangwang (approved):
LGTM, Nice catch!

PR #9508, opened 2020-10-27
KAFKA-10648: Add Prefix Scan support to State Stores

vamossagar12:
comment to ping reviewers.*

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
@@ -16,6 +16,7 @@
  */
 package org.apache.kafka.streams.state;
 
+import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.streams.errors.InvalidStateStoreException;
 
 /**
@@ -94,6 +95,23 @@ public interface ReadOnlyKeyValueStore<K, V> {
      * @throws InvalidStateStoreException if the store is not initialized
      */
     default KeyValueIterator<K, V> reverseAll() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Return an iterator over all keys with the specified prefix.
+     * Since the type of the prefix can be different from that of the key, a serializer to convert the
+     * prefix into the format in which the keys are stored in the stores needs to be passed to this method.
+     * The returned iterator must be safe from {@link java.util.ConcurrentModificationException}s
+     * and must not return null values.
+     *
+     * @param prefix The prefix.
+     * @param prefixKeySerializer Serializer for the Prefix key type
+     * @param <PS> Prefix Serializer type
+     * @param <P> Prefix Type.
+     * @return The iterator for keys having the specified prefix.
+     */
+    default <PS extends Serializer<P>, P> KeyValueIterator<K, V> prefixScan(P prefix, PS prefixKeySerializer) {
         throw new UnsupportedOperationException();
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/CachingKeyValueStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/CachingKeyValueStore.java
@@ -16,6 +16,7 @@
  */
 package org.apache.kafka.streams.state.internals;
 
+import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.processor.ProcessorContext;
@@ -288,6 +289,16 @@ public class CachingKeyValueStore
         final KeyValueIterator<Bytes, byte[]> storeIterator =
             new DelegatingPeekingKeyValueIterator<>(this.name(), wrapped().all());
         final ThreadCache.MemoryLRUCacheBytesIterator cacheIterator = context.cache().all(cacheName);
+        return new MergedSortedCacheKeyValueBytesStoreIterator(cacheIterator, storeIterator, true);
+    }
+
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, final PS prefixKeySerializer) {
+        validateStoreOpen();
+        final KeyValueIterator<Bytes, byte[]> storeIterator = wrapped().prefixScan(prefix, prefixKeySerializer);
+        final Bytes from = Bytes.wrap(prefixKeySerializer.serialize(null, prefix));
+        final Bytes to = Bytes.increment(from);
+        final ThreadCache.MemoryLRUCacheBytesIterator cacheIterator = context.cache().range(cacheName, from, to, false);
         return new MergedSortedCacheKeyValueBytesStoreIterator(cacheIterator, storeIterator, true);
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStore.java
@@ -16,6 +16,7 @@
  */
 package org.apache.kafka.streams.state.internals;
 
+import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.processor.ProcessorContext;
@@ -95,6 +96,12 @@ public class ChangeLoggingKeyValueBytesStore
         for (final KeyValue<Bytes, byte[]> entry : entries) {
             log(entry.key, entry.value);
         }
+    }
+
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix,
+                                                                                    final PS prefixKeySerializer) {
+        return wrapped().prefixScan(prefix, prefixKeySerializer);
     }
 
     @Override
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
@@ -16,6 +16,7 @@
  */
 package org.apache.kafka.streams.state.internals;
 
+import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.processor.ProcessorContext;
@@ -31,6 +32,7 @@ import java.util.NavigableMap;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.Objects;
 
 public class InMemoryKeyValueStore implements KeyValueStore<Bytes, byte[]> {
 
@@ -101,6 +103,20 @@ public class InMemoryKeyValueStore implements KeyValueStore<Bytes, byte[]> {
         for (final KeyValue<Bytes, byte[]> entry : entries) {
             put(entry.key, entry.value);
         }
+    }
+
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, final PS prefixKeySerializer) {
+        Objects.requireNonNull(prefix, "prefix cannot be null");
+        Objects.requireNonNull(prefixKeySerializer, "prefixKeySerializer cannot be null");
+
+        final Bytes from = Bytes.wrap(prefixKeySerializer.serialize(null, prefix));
+        final Bytes to = Bytes.increment(from);
+
+        return new DelegatingPeekingKeyValueIterator<>(
+            name,
+            new InMemoryKeyValueIterator(map.subMap(from, true, to, false).keySet(), true)
+        );
     }
 
     @Override
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
@@ -18,6 +18,7 @@ package org.apache.kafka.streams.state.internals;
 
 import org.apache.kafka.common.metrics.Sensor;
 import org.apache.kafka.common.serialization.Serde;
+import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.streams.KeyValue;
@@ -67,6 +68,7 @@ public class MeteredKeyValueStore<K, V>
     private Sensor putAllSensor;
     private Sensor allSensor;
     private Sensor rangeSensor;
+    private Sensor prefixScanSensor;
     private Sensor flushSensor;
     private Sensor e2eLatencySensor;
     private InternalProcessorContext context;
@@ -127,6 +129,7 @@ public class MeteredKeyValueStore<K, V>
         getSensor = StateStoreMetrics.getSensor(threadId, taskId, metricsScope, name(), streamsMetrics);
         allSensor = StateStoreMetrics.allSensor(threadId, taskId, metricsScope, name(), streamsMetrics);
         rangeSensor = StateStoreMetrics.rangeSensor(threadId, taskId, metricsScope, name(), streamsMetrics);
+        prefixScanSensor = StateStoreMetrics.prefixScanSensor(taskId, metricsScope, name(), streamsMetrics);
         flushSensor = StateStoreMetrics.flushSensor(threadId, taskId, metricsScope, name(), streamsMetrics);
         deleteSensor = StateStoreMetrics.deleteSensor(threadId, taskId, metricsScope, name(), streamsMetrics);
         e2eLatencySensor = StateStoreMetrics.e2ELatencySensor(taskId, metricsScope, name(), streamsMetrics);
@@ -227,6 +230,12 @@ public class MeteredKeyValueStore<K, V>
             final String message = String.format(e.getMessage(), key);
             throw new ProcessorStateException(message, e);
         }
+    }
+
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<K, V> prefixScan(final P prefix, final PS prefixKeySerializer) {
+
+        return new MeteredKeyValueIterator(wrapped().prefixScan(prefix, prefixKeySerializer), prefixScanSensor);
     }
 
     @Override
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/NamedCache.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/NamedCache.java
@@ -280,8 +280,8 @@ class NamedCache {
         return cache.isEmpty();
     }
 
-    synchronized Iterator<Bytes> keyRange(final Bytes from, final Bytes to) {
-        return keySetIterator(cache.navigableKeySet().subSet(from, true, to, true), true);
+    synchronized Iterator<Bytes> keyRange(final Bytes from, final Bytes to, final boolean toInclusive) {
+        return keySetIterator(cache.navigableKeySet().subSet(from, true, to, toInclusive), true);
     }
 
     synchronized Iterator<Bytes> reverseKeyRange(final Bytes from, final Bytes to) {
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
@@ -17,6 +17,7 @@
 package org.apache.kafka.streams.state.internals;
 
 import org.apache.kafka.common.metrics.Sensor.RecordingLevel;
+import org.apache.kafka.common.serialization.Serializer;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.streams.KeyValue;
@@ -309,6 +310,21 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
     }
 
     @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix,
+                                                                                    final PS prefixKeySerializer) {
+        Objects.requireNonNull(prefix, "prefix cannot be null");
+        Objects.requireNonNull(prefixKeySerializer, "prefixKeySerializer cannot be null");
+
+        validateStoreOpen();
+        final Bytes prefixBytes = Bytes.wrap(prefixKeySerializer.serialize(null, prefix));
+
+        final KeyValueIterator<Bytes, byte[]> rocksDbPrefixSeekIterator = dbAccessor.prefixScan(prefixBytes);
+        openIterators.add(rocksDbPrefixSeekIterator);
+
+        return rocksDbPrefixSeekIterator;
+    }
+
+    @Override
     public synchronized byte[] get(final Bytes key) {
         validateStoreOpen();
         try {
@@ -510,6 +526,8 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
 
         KeyValueIterator<Bytes, byte[]> all(final boolean forward);
 
+        KeyValueIterator<Bytes, byte[]> prefixScan(final Bytes prefix);
+
         long approximateNumEntries() throws RocksDBException;
 
         void flush() throws RocksDBException;
@@ -580,7 +598,9 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
                 openIterators,
                 from,
                 to,
-                forward);
+                forward,
+                true
+            );
         }
 
         @Override
@@ -592,6 +612,20 @@ public class RocksDBStore implements KeyValueStore<Bytes, byte[]>, BatchWritingS
                 innerIterWithTimestamp.seekToLast();
             }
             return new RocksDbIterator(name, innerIterWithTimestamp, openIterators, forward);
+        }
+
+        @Override
+        public KeyValueIterator<Bytes, byte[]> prefixScan(final Bytes prefix) {
+            final Bytes to = Bytes.increment(prefix);
+            return new RocksDBRangeIterator(
+                name,
+                db.newIterator(columnFamily),
+                openIterators,
+                prefix,
+                to,
+                true,
+                false
+            );
         }
 
         @Override
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
@@ -201,7 +201,8 @@ public class RocksDBTimestampedStore extends RocksDBStore implements Timestamped
                 db.newIterator(oldColumnFamily),
                 from,
                 to,
-                forward);
+                forward,
+                true);
         }
 
         @Override
@@ -216,6 +217,20 @@ public class RocksDBTimestampedStore extends RocksDBStore implements Timestamped
                 innerIterNoTimestamp.seekToLast();
             }
             return new RocksDBDualCFIterator(name, innerIterWithTimestamp, innerIterNoTimestamp, forward);
+        }
+
+        @Override
+        public KeyValueIterator<Bytes, byte[]> prefixScan(final Bytes prefix) {
+            final Bytes to = Bytes.increment(prefix);
+            return new RocksDBDualCFRangeIterator(
+                name,
+                db.newIterator(newColumnFamily),
+                db.newIterator(oldColumnFamily),
+                prefix,
+                to,
+                true,
+                false
+            );
         }
 
         @Override
@@ -382,15 +397,18 @@ public class RocksDBTimestampedStore extends RocksDBStore implements Timestamped
         private final Comparator<byte[]> comparator = Bytes.BYTES_LEXICO_COMPARATOR;
         private final byte[] rawLastKey;
         private final boolean forward;
+        private final boolean toInclusive;
 
         RocksDBDualCFRangeIterator(final String storeName,
                                    final RocksIterator iterWithTimestamp,
                                    final RocksIterator iterNoTimestamp,
                                    final Bytes from,
                                    final Bytes to,
-                                   final boolean forward) {
+                                   final boolean forward,
+                                   final boolean toInclusive) {
             super(storeName, iterWithTimestamp, iterNoTimestamp, forward);
             this.forward = forward;
+            this.toInclusive = toInclusive;
             if (forward) {
                 iterWithTimestamp.seek(from.get());
                 iterNoTimestamp.seek(from.get());
@@ -416,8 +434,11 @@ public class RocksDBTimestampedStore extends RocksDBStore implements Timestamped
                 return allDone();
             } else {
                 if (forward) {
-                    if (comparator.compare(next.key.get(), rawLastKey) <= 0) {
+
+                    if (comparator.compare(next.key.get(), rawLastKey) < 0) {
                         return next;
+                    } else if (comparator.compare(next.key.get(), rawLastKey) == 0) {
+                        return toInclusive ? next : allDone();
                     } else {
                         return allDone();
                     }
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
@@ -187,11 +187,15 @@ public class ThreadCache {
     }
 
     public MemoryLRUCacheBytesIterator range(final String namespace, final Bytes from, final Bytes to) {
+        return range(namespace, from, to, true);
+    }
+
+    public MemoryLRUCacheBytesIterator range(final String namespace, final Bytes from, final Bytes to, final boolean toInclusive) {
         final NamedCache cache = getCache(namespace);
         if (cache == null) {
             return new MemoryLRUCacheBytesIterator(Collections.emptyIterator(), new NamedCache(namespace, this.metrics));
         }
-        return new MemoryLRUCacheBytesIterator(cache.keyRange(from, to), cache);
+        return new MemoryLRUCacheBytesIterator(cache.keyRange(from, to, toInclusive), cache);
     }
 
     public MemoryLRUCacheBytesIterator reverseRange(final String namespace, final Bytes from, final Bytes to) {
--- a/streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetrics.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetrics.java
@@ -110,6 +110,13 @@ public class StateStoreMetrics {
     private static final String RANGE_AVG_LATENCY_DESCRIPTION = AVG_LATENCY_DESCRIPTION_PREFIX + RANGE_DESCRIPTION;
     private static final String RANGE_MAX_LATENCY_DESCRIPTION = MAX_LATENCY_DESCRIPTION_PREFIX + RANGE_DESCRIPTION;
 
+    private static final String PREFIX_SCAN = "prefix-scan";
+    private static final String PREFIX_SCAN_DESCRIPTION = "calls to prefix-scan";
+    private static final String PREFIX_SCAN_RATE_DESCRIPTION =
+            RATE_DESCRIPTION_PREFIX + PREFIX_SCAN_DESCRIPTION + RATE_DESCRIPTION_SUFFIX;
+    private static final String PREFIX_SCAN_AVG_LATENCY_DESCRIPTION = AVG_LATENCY_DESCRIPTION_PREFIX + PREFIX_SCAN_DESCRIPTION;
+    private static final String PREFIX_SCAN_MAX_LATENCY_DESCRIPTION = MAX_LATENCY_DESCRIPTION_PREFIX + PREFIX_SCAN_DESCRIPTION;
+
     private static final String FLUSH = "flush";
     private static final String FLUSH_DESCRIPTION = "calls to flush";
     private static final String FLUSH_TOTAL_DESCRIPTION = TOTAL_DESCRIPTION + FLUSH_DESCRIPTION;
@@ -305,6 +312,33 @@ public class StateStoreMetrics {
             RecordingLevel.DEBUG,
             streamsMetrics
         );
+    }
+
+    public static Sensor prefixScanSensor(final String taskId,
+                                          final String storeType,
+                                          final String storeName,
+                                          final StreamsMetricsImpl streamsMetrics) {
+
+        final String latencyMetricName = PREFIX_SCAN + LATENCY_SUFFIX;
+        final Map<String, String> tagMap = streamsMetrics.storeLevelTagMap(taskId, storeType, storeName);
+
+        final Sensor sensor = streamsMetrics.storeLevelSensor(taskId, storeName, PREFIX_SCAN, RecordingLevel.DEBUG);
+        addInvocationRateToSensor(
+            sensor,
+            STATE_STORE_LEVEL_GROUP,
+            tagMap,
+            PREFIX_SCAN,
+            PREFIX_SCAN_RATE_DESCRIPTION
+        );
+        addAvgAndMaxToSensor(
+            sensor,
+            STATE_STORE_LEVEL_GROUP,
+            tagMap,
+            latencyMetricName,
+            PREFIX_SCAN_AVG_LATENCY_DESCRIPTION,
+            PREFIX_SCAN_MAX_LATENCY_DESCRIPTION
+        );
+        return sensor;
     }
 
     public static Sensor flushSensor(final String threadId,
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/CachingInMemoryKeyValueStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/CachingInMemoryKeyValueStoreTest.java
@@ -19,6 +19,7 @@ package org.apache.kafka.streams.state.internals;
 import org.apache.kafka.common.metrics.Metrics;
 import org.apache.kafka.common.serialization.Serde;
 import org.apache.kafka.common.serialization.StringDeserializer;
+import org.apache.kafka.common.serialization.StringSerializer;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.common.utils.LogContext;
 import org.apache.kafka.streams.KeyValue;
@@ -45,6 +46,7 @@ import java.util.List;
 
 import static org.apache.kafka.streams.state.internals.ThreadCacheTest.memoryCacheEntrySize;
 import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -357,6 +359,60 @@ public class CachingInMemoryKeyValueStoreTest extends AbstractKeyValueStoreTest 
             Bytes.wrap("1".getBytes()),
             Bytes.wrap("0".getBytes())
         ), results);
+    }
+
+    @Test
+    public void shouldGetRecordsWithPrefixKey() {
+        final List<KeyValue<Bytes, byte[]>> entries = new ArrayList<>();
+        entries.add(new KeyValue<>(bytesKey("p11"), bytesValue("2")));
+        entries.add(new KeyValue<>(bytesKey("k1"), bytesValue("1")));
+        entries.add(new KeyValue<>(bytesKey("k2"), bytesValue("2")));
+        entries.add(new KeyValue<>(bytesKey("p2"), bytesValue("2")));
+        entries.add(new KeyValue<>(bytesKey("p1"), bytesValue("2")));
+        entries.add(new KeyValue<>(bytesKey("p0"), bytesValue("2")));
+
+        store.putAll(entries);
+
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = store.prefixScan("p1", new StringSerializer());
+        final List<String> keys = new ArrayList<>();
+        final List<String> values = new ArrayList<>();
+        int numberOfKeysReturned = 0;
+
+        while (keysWithPrefix.hasNext()) {
+            final KeyValue<Bytes, byte[]> next = keysWithPrefix.next();
+            keys.add(next.key.toString());
+            values.add(new String(next.value));
+            numberOfKeysReturned++;
+        }
+        assertThat(numberOfKeysReturned, is(2));
+        assertThat(keys, is(Arrays.asList("p1", "p11")));
+        assertThat(values, is(Arrays.asList("2", "2")));
+    }
+
+    @Test
+    public void shouldGetRecordsWithPrefixKeyExcludingNextLargestKey() {
+        final List<KeyValue<Bytes, byte[]>> entries = new ArrayList<>();
+        entries.add(new KeyValue<>(bytesKey("abcd"), bytesValue("2")));
+        entries.add(new KeyValue<>(bytesKey("abcdd"), bytesValue("1")));
+        entries.add(new KeyValue<>(bytesKey("abce"), bytesValue("2")));
+        entries.add(new KeyValue<>(bytesKey("abc"), bytesValue("2")));
+
+        store.putAll(entries);
+
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = store.prefixScan("abcd", new StringSerializer());
+        final List<String> keys = new ArrayList<>();
+        final List<String> values = new ArrayList<>();
+        int numberOfKeysReturned = 0;
+
+        while (keysWithPrefix.hasNext()) {
+            final KeyValue<Bytes, byte[]> next = keysWithPrefix.next();
+            keys.add(next.key.toString());
+            values.add(new String(next.value));
+            numberOfKeysReturned++;
+        }
+        assertThat(numberOfKeysReturned, is(2));
+        assertThat(keys, is(Arrays.asList("abcd", "abcdd")));
+        assertThat(values, is(Arrays.asList("2", "1")));
     }
 
     @Test
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
@@ -206,8 +206,8 @@ public class ChangeLoggingKeyValueBytesStoreTest {
         store.put(hi, there);
         store.put(Bytes.increment(hi), world);
         final KeyValueIterator<Bytes, byte[]> keysWithPrefix = store.prefixScan(hi.toString(), new StringSerializer());
-        final List<String> keys = new ArrayList<>();
-        final List<String> values = new ArrayList<>();
+        final List<Bytes> keys = new ArrayList<>();
+        final List<Bytes> values = new ArrayList<>();
         int numberOfKeysReturned = 0;
 
         while (keysWithPrefix.hasNext()) {
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
@@ -212,8 +212,8 @@ public class ChangeLoggingKeyValueBytesStoreTest {
 
         while (keysWithPrefix.hasNext()) {
             final KeyValue<Bytes, byte[]> next = keysWithPrefix.next();
-            keys.add(next.key.toString());
-            values.add(new String(next.value));
+            keys.add(next.key);
+            values.add(Bytes.wrap(next.value));
             numberOfKeysReturned++;
         }
         assertThat(numberOfKeysReturned, is(1));
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
@@ -18,6 +18,7 @@ package org.apache.kafka.streams.state.internals;
 
 import org.apache.kafka.common.metrics.Metrics;
 import org.apache.kafka.common.serialization.Serdes;
+import org.apache.kafka.common.serialization.StringSerializer;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.common.utils.LogContext;
 import org.apache.kafka.streams.KeyValue;
@@ -25,6 +26,7 @@ import org.apache.kafka.streams.processor.ProcessorContext;
 import org.apache.kafka.streams.processor.StateStore;
 import org.apache.kafka.streams.processor.StateStoreContext;
 import org.apache.kafka.streams.processor.internals.MockStreamsMetrics;
+import org.apache.kafka.streams.state.KeyValueIterator;
 import org.apache.kafka.streams.state.KeyValueStore;
 import org.apache.kafka.test.InternalMockProcessorContext;
 import org.apache.kafka.test.MockRecordCollector;
@@ -34,7 +36,10 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
@@ -194,6 +199,26 @@ public class ChangeLoggingKeyValueBytesStoreTest {
     public void shouldReturnValueOnGetWhenExists() {
         store.put(hello, world);
         assertThat(store.get(hello), equalTo(world));
+    }
+
+    @Test
+    public void shouldGetRecordsWithPrefixKey() {
+        store.put(hi, there);
+        store.put(Bytes.increment(hi), world);
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = store.prefixScan("hi", new StringSerializer());
+        final List<String> keys = new ArrayList<>();
+        final List<String> values = new ArrayList<>();
+        int numberOfKeysReturned = 0;
+
+        while (keysWithPrefix.hasNext()) {
+            final KeyValue<Bytes, byte[]> next = keysWithPrefix.next();
+            keys.add(next.key.toString());
+            values.add(new String(next.value));
+            numberOfKeysReturned++;
+        }
+        assertThat(numberOfKeysReturned, is(1));
+        assertThat(keys, is(Collections.singletonList("hi")));
+        assertThat(values, is(Collections.singletonList("there")));
     }
 
     @Test
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
@@ -205,7 +205,7 @@ public class ChangeLoggingKeyValueBytesStoreTest {
     public void shouldGetRecordsWithPrefixKey() {
         store.put(hi, there);
         store.put(Bytes.increment(hi), world);
-        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = store.prefixScan("hi", new StringSerializer());
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = store.prefixScan(hi.toString(), new StringSerializer());
         final List<String> keys = new ArrayList<>();
         final List<String> values = new ArrayList<>();
         int numberOfKeysReturned = 0;
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
@@ -217,8 +217,8 @@ public class ChangeLoggingKeyValueBytesStoreTest {
             numberOfKeysReturned++;
         }
         assertThat(numberOfKeysReturned, is(1));
-        assertThat(keys, is(Collections.singletonList("hi")));
-        assertThat(values, is(Collections.singletonList("there")));
+        assertThat(keys, is(Collections.singletonList(hi)));
+        assertThat(values, is(Collections.singletonList(Bytes.wrap(there))));
     }
 
     @Test
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
@@ -27,6 +27,7 @@ import org.apache.kafka.common.serialization.Deserializer;
 import org.apache.kafka.common.serialization.Serde;
 import org.apache.kafka.common.serialization.Serdes;
 import org.apache.kafka.common.serialization.Serializer;
+import org.apache.kafka.common.serialization.StringSerializer;
 import org.apache.kafka.common.utils.Bytes;
 import org.apache.kafka.common.utils.MockTime;
 import org.apache.kafka.common.utils.Time;
@@ -431,6 +432,22 @@ public class MeteredKeyValueStoreTest {
         assertThat(storeMetrics(), not(empty()));
         assertThrows(RuntimeException.class, metered::close);
         assertThat(storeMetrics(), empty());
+        verify(inner);
+    }
+
+    @Test
+    public void shouldGetRecordsWithPrefixKey() {
+        final StringSerializer stringSerializer = new StringSerializer();
+        expect(inner.prefixScan(KEY, stringSerializer))
+            .andReturn(new KeyValueIteratorStub<>(Collections.singletonList(BYTE_KEY_VALUE_PAIR).iterator()));
+        init();
+
+        final KeyValueIterator<String, String> iterator = metered.prefixScan(KEY, stringSerializer);
+        assertThat(iterator.next().value, equalTo(VALUE));
+        iterator.close();
+
+        final KafkaMetric metric = metrics.metric(new MetricName("prefix-scan-rate", STORE_LEVEL_GROUP, "", tags));
+        assertTrue((Double) metric.metricValue() > 0);
         verify(inner);
     }
 
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
@@ -446,7 +446,7 @@ public class MeteredKeyValueStoreTest {
         assertThat(iterator.next().value, equalTo(VALUE));
         iterator.close();
 
-        final KafkaMetric metric = metric("prefix-scan-rate");
+        final KafkaMetric metric = metrics.metric(new MetricName("prefix-scan-rate", STORE_LEVEL_GROUP, "", tags));
         assertTrue((Double) metric.metricValue() > 0);
         verify(inner);
     }
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
@@ -446,7 +446,7 @@ public class MeteredKeyValueStoreTest {
         assertThat(iterator.next().value, equalTo(VALUE));
         iterator.close();
 
-        final KafkaMetric metric = metrics.metric(new MetricName("prefix-scan-rate", STORE_LEVEL_GROUP, "", tags));
+        final KafkaMetric metric = metric("prefix-scan-rate");
         assertTrue((Double) metric.metricValue() > 0);
         verify(inner);
     }
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
@@ -66,6 +66,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.UUID;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.easymock.EasyMock.eq;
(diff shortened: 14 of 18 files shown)


streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
```diff
@@ -107,4 +108,18 @@
      * @throws InvalidStateStoreException if the store is not initialized
      */
     long approximateNumEntries();
+
+    /**
+     * Get an iterator over keys which have the specified prefix. The type of the prefix can be different from that of
+     * the key. That's why, callers should also pass a serializer for the prefix to convert the prefix into the
+     * format in which the keys are stored underneath in the stores
+     * @param prefix The prefix.
```
cadonna, line 108: Please leave an empty line above to separate parameter description from the description of the method.

streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
```diff
@@ -107,4 +108,18 @@
      * @throws InvalidStateStoreException if the store is not initialized
      */
     long approximateNumEntries();
+
+    /**
+     * Get an iterator over keys which have the specified prefix. The type of the prefix can be different from that of
+     * the key. That's why, callers should also pass a serializer for the prefix to convert the prefix into the
+     * format in which the keys are stored underneath in the stores
```
cadonna: We usually put every sentence on its own line In javadocs. Unfortunately, we are not always consistent as you can see in the javadocs in this file.
I guess this iterator has similar requirements as the other iterators, i.e., it must be closed, it must be safe from `java.util.ConcurrentModificationException`, and it must not return `null` values. If so, please add this requirements to the javadocs.
```suggestion
     * Return an iterator over all keys with the specified prefix. 
     * Since the type of the prefix can be different from that of the key, a serializer to convert the
     * prefix into the format in which the keys are stored in the stores needs to be passed to this method.  
```

streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
```diff
@@ -107,4 +108,18 @@
...
+     * @param prefix The prefix.
+     * @param prefixKeySerializer Serializer for the Prefix key type
+     * @param <PS> Prefix Serializer type
+     * @param <P> Prefix Type.
+     * @return The iterator for keys having the specified prefix.
+     */
+    default <PS extends Serializer<P>, P> KeyValueIterator<K, V> prefixScan(P prefix, PS prefixKeySerializer) {
```
cadonna, line 114: I would move this method above `approximateNumEntries`, so that all methods that return iterators are in one block.

streams/src/main/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStore.java
```diff
@@ -97,6 +98,11 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
         }
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, final PS prefixKeySerializer) {
+        return wrapped().prefixScan(prefix, prefixKeySerializer);
+    }
```
cadonna:
```suggestion
    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, 
                                                                                    final PS prefixKeySerializer) {
        return wrapped().prefixScan(prefix, prefixKeySerializer);
    }
```

streams/src/main/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStore.java
```diff
@@ -97,6 +98,11 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
         }
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, final PS prefixKeySerializer) {
```
cadonna: Unit test for this method is missing.

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
```diff
@@ -103,6 +105,19 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
         }
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, final PS prefixKeySerializer) {
```
cadonna, line 109: This method needs unit testing.

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
```diff
@@ -103,6 +105,19 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
...
+        Objects.requireNonNull(prefix, "prefix cannot be null");
+        Objects.requireNonNull(prefixKeySerializer, "prefixKeySerializer cannot be null");
+
+        final Bytes from = Bytes.wrap(prefixKeySerializer.serialize(null, prefix));
+        final Bytes to = Bytes.increment(from);
+
+        return new DelegatingPeekingKeyValueIterator<>(
+                name,
+                new InMemoryKeyValueIterator(map.subMap(from, true, to, false).keySet(), true));
```
cadonna:
```suggestion
        return new DelegatingPeekingKeyValueIterator<>(
            name,
            new InMemoryKeyValueIterator(map.subMap(from, true, to, false).keySet(), true)
        );
```

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -229,6 +230,15 @@ public V delete(final K key) {
         }
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<K, V> prefixScan(final P prefix, final PS prefixKeySerializer) {
```
cadonna, line 236: This method needs unit testing.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
```diff
@@ -0,0 +1,55 @@
...
+    RocksDBPrefixIterator(final String name,
+                          final RocksIterator newIterator,
+                          final Set<KeyValueIterator<Bytes, byte[]>> openIterators,
+                          final Bytes prefix) {
+        super(name, newIterator, openIterators, true);
+        this.rawPrefix = prefix.get();
+        newIterator.seek(rawPrefix);
+    }
+
+    private boolean prefixEquals(final byte[] x, final byte[] y) {
```
cadonna: Please use meaningful names for the paramters instead of `x` and `y`.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.streams.KeyValue;
+import org.apache.kafka.streams.state.KeyValueIterator;
+import org.rocksdb.RocksIterator;
+
+import java.nio.ByteBuffer;
+import java.util.Set;
+
+class RocksDBPrefixIterator extends RocksDbIterator {
```
cadonna: This class needs unit testing.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -308,6 +309,20 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
         }
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, final PS prefixKeySerializer) {
```
cadonna: Please put the second parameter on a new line and align the two parameters like:
```suggestion
    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, 
                                                                                    final PS prefixKeySerializer) {
```
(On GitHub the suggestions seems a bit strange)

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
```diff
@@ -256,6 +256,11 @@ public void close() {
             oldColumnFamily.close();
             newColumnFamily.close();
         }
+
+        @Override
+        public KeyValueIterator<Bytes, byte[]> prefixScan(final Bytes prefix) {
+            throw new UnsupportedOperationException();
+        }
```
cadonna: Why is the prefix scan not supported here? The timstamped stores are the default stores in Kafka Streams.

streams/src/test/java/org/apache/kafka/streams/processor/internals/ProcessorTopologyTest.java
```diff
@@ -273,6 +275,43 @@ public void testDrivingStatefulTopology() {
         assertNull(store.get("key4"));
     }
 
+    @Test
+    public void testPrefixScanStatefulTopology() {
```
cadonna: You do not need to test the prefix scan by using the topology test driver because you actually do not need a topology. Your addition is limited to the state stores and those you should unit test. Furthermore, this test is misplaced because you did not change anything in `ProcessorTopology`. Please remove this test and add unit tests where I indicated the need for them.

Although, you will find a lot of test named `test*`, we name new tests with `should*` followed by the expected result, e.g., `shouldThrowIllegalStateException` or `shouldGetRecordsWithPrefixedKey`.

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
```diff
@@ -360,6 +361,118 @@ public void shouldPutAll() {
...
+        assertEquals(valuesWithPrefix[1], "d");
+        assertEquals(valuesWithPrefix[2], "b");
+
+        // Lastly, simple key value lookups should still work :)
+        assertEquals(
+                "c",
+                stringDeserializer.deserialize(
+                        null,
+                        rocksDBStore.get(new Bytes(stringSerializer.serialize(null, "k2")))));
+
```
cadonna: Please remove empty line.

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
```diff
@@ -360,6 +361,118 @@ public void shouldPutAll() {
...
+                stringSerializer.serialize(null, "c")));
+        entries.add(new KeyValue<>(
+                new Bytes(stringSerializer.serialize(null, "prefix_2")),
+                stringSerializer.serialize(null, "d")));
+        entries.add(new KeyValue<>(
+                new Bytes(stringSerializer.serialize(null, "k3")),
+                stringSerializer.serialize(null, "e")));
+        entries.add(new KeyValue<>(
+                new Bytes(stringSerializer.serialize(null, "prefix_1")),
+                stringSerializer.serialize(null, "f")));
```
cadonna: Please fix the indentation and the newlines also in the other tests as follows:
```suggestion
        entries.add(new KeyValue<>(
            new Bytes(stringSerializer.serialize(null, "k1")),
            stringSerializer.serialize(null, "a"))
        );
        entries.add(new KeyValue<>(
            new Bytes(stringSerializer.serialize(null, "prefix_3")),
            stringSerializer.serialize(null, "b"))
        );
        entries.add(new KeyValue<>(
            new Bytes(stringSerializer.serialize(null, "k2")),
            stringSerializer.serialize(null, "c"))
        );
        entries.add(new KeyValue<>(
            new Bytes(stringSerializer.serialize(null, "prefix_2")),
            stringSerializer.serialize(null, "d"))
        );
        entries.add(new KeyValue<>(
            new Bytes(stringSerializer.serialize(null, "k3")),
            stringSerializer.serialize(null, "e"))
        );
        entries.add(new KeyValue<>(
            new Bytes(stringSerializer.serialize(null, "prefix_1")),
            stringSerializer.serialize(null, "f"))
        );
```

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
```diff
@@ -360,6 +361,118 @@ public void shouldPutAll() {
...
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = rocksDBStore.prefixScan("prefix", stringSerializer);
+        final String[] valuesWithPrefix = new String[3];
+        int numberOfKeysReturned = 0;
+
+        while (keysWithPrefix.hasNext()) {
+            final KeyValue<Bytes, byte[]> next = keysWithPrefix.next();
+            valuesWithPrefix[numberOfKeysReturned++] = new String(next.value);
+        }
+        // Since there are 3 keys prefixed with prefix, the count should be 3
+        assertEquals(3, numberOfKeysReturned);
```
cadonna: We prefer to use `assertThat()` in new test, because it is better readable. This line would then become:
 ```suggestion
        assertThat(numberOfKeysReturned, is(3));
```
Please replace also the other verifications with `assertThat()` and fix the indentation (should be 4 spaces, not 8).

File: streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java [hunk 17]
```diff
@@ -360,6 +361,118 @@ public void shouldPutAll() {
...
+        final String prefix = uuid1.toString().substring(0, 4);
+        entries.add(new KeyValue<>(
+                new Bytes(uuidSerializer.serialize(null, uuid1)),
+                stringSerializer.serialize(null, "a")));
+
+        entries.add(new KeyValue<>(
+                new Bytes(uuidSerializer.serialize(null, uuid2)),
+                stringSerializer.serialize(null, "b")));
+
+
```
cadonna, line 448: Please remove empty line.

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
```diff
@@ -360,6 +361,118 @@ public void shouldPutAll() {
...
+        rocksDBStore.putAll(entries);
+        rocksDBStore.flush();
+
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = rocksDBStore.prefixScan("prefix", stringSerializer);
+        final String[] valuesWithPrefix = new String[3];
+        int numberOfKeysReturned = 0;
+
+        while (keysWithPrefix.hasNext()) {
+            final KeyValue<Bytes, byte[]> next = keysWithPrefix.next();
+            valuesWithPrefix[numberOfKeysReturned++] = new String(next.value);
```
cadonna: I would rather use a `List` instead of the array and then verify if the length of the list is three. That will give us a more meaningful assertion error than the `IndexOutOfBoundsException` that we would get with the array in case of failure.

cadonna (commented):
@vamossagar12, Thank you for the PR!

Here my feedback!

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
vamossagar12: This one, do I need explicit unit tests? I noticed that something like RocksDbRangeIterator also doesn't have unit tests.

vamossagar12:
> @vamossagar12, Thank you for the PR!
> 
> Here my feedback!

@cadonna thanks i made the requeiste changes barring 1.

vamossagar12:
@cadonna , The failing tests here https://github.com/apache/kafka/pull/9508/checks?check_run_id=1376108664 don't seem to be related to this PR. Would it be possible to retest?

vamossagar12:
hi @cadonna , wanted to know did you get a chance to look at my comment on the failing test case and the changes in general?

cadonna:
@vamossagar12 I am really sorry. I haven't found the time yet, but it is on my ToDo list. I hope I will manage to make a pass this week.

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
cadonna: This method still needs unit testing.

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
```diff
@@ -103,6 +105,19 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
...
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, final PS prefixKeySerializer) {
+        Objects.requireNonNull(prefix, "prefix cannot be null");
+        Objects.requireNonNull(prefixKeySerializer, "prefixKeySerializer cannot be null");
+
+        final Bytes from = Bytes.wrap(prefixKeySerializer.serialize(null, prefix));
+        final Bytes to = Bytes.increment(from);
+
+        return new DelegatingPeekingKeyValueIterator<>(
+            name,
+            new InMemoryKeyValueIterator(map.subMap(from, true, to, false).keySet(), true));
```
cadonna: nit: The last parenthesis should go to a new line. I know that we are not consistent throughout the code base (that is why this comment is prefixed with "nit") but that is actually the code style, we agreed upon. If you need to push another commit you can fix this, otherwise it's fine. 
```suggestion
        return new DelegatingPeekingKeyValueIterator<>(
            name,
            new InMemoryKeyValueIterator(map.subMap(from, true, to, false).keySet(), true)
        );
```

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -229,6 +230,15 @@ public V delete(final K key) {
...
+
+        return new MeteredKeyValueIterator(
+                wrapped().prefixScan(prefix, prefixKeySerializer),
+                rangeSensor
+        );
```
cadonna: nit: Sorry, I overlooked this last time. We actually use 4 space and not 8.
```suggestion
        return new MeteredKeyValueIterator(
            wrapped().prefixScan(prefix, prefixKeySerializer),
            rangeSensor
        );
```

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -229,6 +230,15 @@ public V delete(final K key) {
         }
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<K, V> prefixScan(final P prefix, final PS prefixKeySerializer) {
+
+        return new MeteredKeyValueIterator(
+                wrapped().prefixScan(prefix, prefixKeySerializer),
+                rangeSensor
```
cadonna: I am not sure whether we should use the `rangeSensor` here or introduce a new `prefixScanSensor`. I looked into the KIP discussion but could not find any reference to metrics. Either I missed it or we missed it in the KIP discussion. What do the reviewers of the KIP think @ableegoldman @vvcephei @guozhangwang?

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingInMemoryKeyValueStoreTest.java
```diff
@@ -359,6 +361,31 @@ public void shouldReverseIterateOverRange() {
...
+        while (keysWithPrefix.hasNext()) {
+            final KeyValue<Bytes, byte[]> next = keysWithPrefix.next();
+            keys.add(next.key.toString());
+            values.add(new String(next.value));
+            numberOfKeysReturned++;
+        }
+        assertThat(numberOfKeysReturned, is(3));
+        assertThat(keys, is(Arrays.asList("p0", "p1", "p2")));
+        assertThat(values, is(Arrays.asList("2", "2", "2")));
+    }
```
cadonna, line 390: The following request is not a requirement to get the PR approved, but rather optional extra work to improve the code base. Could you rewrite this test with a mock for the inner state store as in `MeteredKeyValueStoreTest`?

streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
```diff
@@ -196,6 +200,26 @@ public void shouldReturnValueOnGetWhenExists() {
...
+            final KeyValue<Bytes, byte[]> next = keysWithPrefix.next();
+            keys.add(next.key.toString());
+            values.add(new String(next.value));
+            numberOfKeysReturned++;
+        }
+        assertThat(numberOfKeysReturned, is(2));
+        assertThat(keys, is(Arrays.asList("hello", "hi")));
+        assertThat(values, is(Arrays.asList("world", "there")));
+    }
+
```
cadonna, line 223: See my comment in `CachingInMemoryKeyValueStoreTest`. Also this not required.

streams/src/main/java/org/apache/kafka/streams/state/internals/NamedCache.java
```diff
@@ -284,6 +284,12 @@ public boolean isEmpty() {
         return keySetIterator(cache.navigableKeySet().subSet(from, true, to, true), true);
     }
 
+    synchronized Iterator<Bytes> keyRange(final Bytes from, final Bytes to, final boolean toInclusive) {
+        if (toInclusive)
+            keyRange(from, to);
+        return keySetIterator(cache.navigableKeySet().subSet(from, true, to, false), true);
+    }
```
cadonna: Why not simply:
```suggestion
    synchronized Iterator<Bytes> keyRange(final Bytes from, final Bytes to, final boolean toInclusive) {
        return keySetIterator(cache.navigableKeySet().subSet(from, true, to, toInclusive), true);
    }
```

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public KeyValue<Bytes, byte[]> makeNext() {
+        final KeyValue<Bytes, byte[]> next = super.makeNext();
+        if (next == null) return allDone();
+        else {
+            if (prefixEquals(this.rawPrefix, next.key.get())) return next;
+            else return allDone();
+        }
```
cadonna:
```suggestion
        if (next == null || !prefixEquals(this.rawPrefix, next.key.get())) {
            return allDone();
        } else {
            return next;
        }
```

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
cadonna: Yes, you need explicit unit tests. If there are no unit tests for `RocksDbRangeIterator` then that is a hole in our test coverage. You do not need to add the tests for `RocksDbRangeIterator` but could you please create a ticket to document the missing unit tests? Of course, you are very welcome to assign the ticket to yourself and add the missing tests in a separate PR.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
```diff
@@ -0,0 +1,55 @@
...
+        super(name, newIterator, openIterators, true);
+        this.rawPrefix = prefix.get();
+        newIterator.seek(rawPrefix);
+    }
+
+    private boolean prefixEquals(final byte[] prefix1, final byte[] prefix2) {
+        final int min = Math.min(prefix1.length, prefix2.length);
+        final ByteBuffer prefix1Slice = ByteBuffer.wrap(prefix1, 0, min);
+        final ByteBuffer prefix2Slice = ByteBuffer.wrap(prefix2, 0, min);
+        return prefix1Slice.equals(prefix2Slice);
```
cadonna: With this code `abcd` would be a prefix of `abc`. Is this intended? Those cases should be tested in the missing unit tests for this class.

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
```diff
@@ -360,6 +361,115 @@ public void shouldPutAll() {
...
+        assertThat(numberOfKeysReturned, is(3));
+        // The order might seem inverted to the order in which keys were inserted, but since Rocksdb stores keys
+        // lexicographically, prefix_1 would still be the first key that is returned.
+        assertThat(valuesWithPrefix.get(0), is("f"));
+        assertThat(valuesWithPrefix.get(1), is("d"));
+        assertThat(valuesWithPrefix.get(2), is("b"));
+
+        // Lastly, simple key value lookups should still work :)
+        assertThat("c", is(stringDeserializer.deserialize(null,
+            rocksDBStore.get(new Bytes(stringSerializer.serialize(null, "k2"))))));
```
cadonna: Do we need this verification? This is covered by other unit test methods, isn't it? Could you please remove it from here?

cadonna: Could you please remove the inline comments? They do not add too much information.

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
```diff
@@ -360,6 +361,115 @@ public void shouldPutAll() {
...
+        rocksDBStore.flush();
+
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = rocksDBStore.prefixScan("d", stringSerializer);
+        int numberOfKeysReturned = 0;
+
+        while (keysWithPrefix.hasNext()) {
+            keysWithPrefix.next();
+            numberOfKeysReturned++;
+        }
+        // Since there are no keys prefixed with d, the count should be 0
```
cadonna: Please remove this comment.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
```diff
@@ -431,4 +441,36 @@ public Bytes peekNextKey() {
             }
         }
     }
+
+    private class RocksDBDualCFPrefixIterator extends RocksDBDualCFIterator {
+        private final byte[] rawPrefix;
+
+        // In Prefix scan mode, we always move in forward direction.
```
cadonna: Please remove comment.

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
```diff
@@ -201,6 +201,14 @@ public MemoryLRUCacheBytesIterator reverseAll(final String namespace) {
         return new MemoryLRUCacheBytesIterator(cache.reverseAllKeys(), cache);
     }
 
+    public MemoryLRUCacheBytesIterator prefixScan(final String namespace, final Bytes from, final Bytes to) {
```
cadonna: This method need unit testing. I realized that also here we missed to cover all methods with unit tests. For example, there are also no unit tests for `all()`. Could you please add the ones for `prefixScan()`? Also here you are really welcome to add unit tests for the other methods in a separate PR, but it is not a requirement for the approval of this PR.

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStoreTest.java
```diff
@@ -130,6 +134,8 @@ public void shouldMigrateDataFromDefaultToTimestampColumnFamily() throws Excepti
         // approx: 7 entries on old CF, 0 in new CF
         assertThat(rocksDBStore.approximateNumEntries(), is(7L));
 
+        // prefix scan should return 7 keys with prefix "key"
+        assertThat(runPrefixScan("key"), is(7));
```
cadonna: I guess, it would be more consistent to add your verifications to `iteratorsShouldNotMigrateData()` where all other iterators are tested.

cadonna (commented):
@vamossagar12 Thank you for the updates and sorry again for the long delay!

Here is my feedback!

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
vamossagar12: @cadonna , you are right, the KIP and the KIP discussion didn't involve a discussion on this. I would love to hear the thoughts of yours and the reviewers.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
vamossagar12: Ticket for unit test cases for RocksDBRangeIterator: https://issues.apache.org/jira/browse/KAFKA-10766.

Regarding `abcd` being returned as the prefix for `abc`, I added a local test where I added an entry with key `abc` 

`        entries.add(new KeyValue<>(new Bytes(stringSerializer.serialize(null, "abc")),stringSerializer.serialize(null, "f")));
`

and did a prefixScan for `abcd`

`
numberOfKeysReturned = 0;

        while (keysWithPrefix1.hasNext()) {
            System.out.println(Arrays.toString(keysWithPrefix1.next().key.get()));
            numberOfKeysReturned++;
        }

        assertThat(numberOfKeysReturned, is(0));
`
That is because the iterator first seeks to the rawPrefix passed (abcd) in this case and then finds keys which match the prefix and are lexicographically more than it.

But as you said, going by the method `prefixEquals`, yes there's a possibility there.

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
vamossagar12: I created a ticket for adding test cases for the missed methods in ThreadCache. https://issues.apache.org/jira/browse/KAFKA-10767

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingInMemoryKeyValueStoreTest.java
vamossagar12: @cadonna , I noticed that for this particular Test class, some test cases like `shouldDelegateDeprecatedInit` and `shouldDelegateInit`  use mocking the `inner` store and other ones use the explicitly created object. If you are ok, I can create a separate ticket and take care of making all test cases similar to `MeteredKeyValueStoreTest` where all of them use the same mocked inner store and let this one as is. Let me know if that's ok with you.

streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
vamossagar12: Same comment as in `CachingInMemoryKeyValueStoreTest`

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
cadonna: I think, to be consistent we should add a new sensor for the prefix scan. We have a sensor for each operation on a key-value store. The only exception is `reverseRange()`, which is a variant of range and does not necessarily qualify for its own sensor.

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingInMemoryKeyValueStoreTest.java
cadonna: Creating a ticket and take care that the ticket gets resolved, sound good to me. As I wote, my request is not a requirement for approval.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
cadonna: I see what you mean. But why do you bother then to find the minimum length of the two prefixes? You could just use the length of the first prefix.

cadonna:
@vamossagar12 I replied to your comments. Let me know when the PR is ready for review.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
vamossagar12: Created prefixScan sensor.

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingInMemoryKeyValueStoreTest.java
vamossagar12: Sure Thanks https://issues.apache.org/jira/browse/KAFKA-10788

streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
vamossagar12: Created ticket here: https://issues.apache.org/jira/browse/KAFKA-10789

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
vamossagar12: Yeah I have actually simplified the logic. Don't think I needed prefix1 and prefix2 as one of the prefixes is always fixed(i.e the prefix we are searching for). 
I have created a unit test class for RocksDBPrefixIterator.

vamossagar12:
Thanks @cadonna , I have resolved all comments. Plz review again if things look ok now..

vamossagar12:
hey @cadonna did you get a chance to look at the latest set of changes?

cadonna:
@vamossagar12 I am really sorry, but I haven't found the time yet to review the changes. I hope I will find the time next week.

streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
```diff
@@ -434,6 +435,23 @@ public void shouldRemoveMetricsEvenIfWrappedStoreThrowsOnClose() {
...
+    @Test
+    public void shouldGetRecordsWithPrefixKey() {
+        final StringSerializer stringSerializer = new StringSerializer();
+        expect(inner.prefixScan(KEY, stringSerializer))
+                .andReturn(new KeyValueIteratorStub<>(Collections.singletonList(BYTE_KEY_VALUE_PAIR).iterator()));
+        init();
+
+        final KeyValueIterator<String, String> iterator = metered.prefixScan(KEY, stringSerializer);
+        assertThat(iterator.next().value, equalTo(VALUE));
+        assertFalse(iterator.hasNext());
```
cadonna: This is not strictly necessary since you test the mock result you provide which has nothing to do with the code under test.

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetrics.java
```diff
@@ -110,6 +110,14 @@ private StateStoreMetrics() {}
     private static final String RANGE_AVG_LATENCY_DESCRIPTION = AVG_LATENCY_DESCRIPTION_PREFIX + RANGE_DESCRIPTION;
     private static final String RANGE_MAX_LATENCY_DESCRIPTION = MAX_LATENCY_DESCRIPTION_PREFIX + RANGE_DESCRIPTION;
 
+    private static final String PREFIX_SCAN = "prefixScan";
+    private static final String PREFIX_SCAN_DESCRIPTION = "calls to prefixScan";
```
cadonna: Please change `prefixScan` to `prefix-scan`. For metrics name we do not use camel case. See `put-all` or `put-if-absent` for examples.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -229,6 +232,15 @@ public V delete(final K key) {
         }
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<K, V> prefixScan(final P prefix, final PS prefixKeySerializer) {
+
+        return new MeteredKeyValueIterator(
+            wrapped().prefixScan(prefix, prefixKeySerializer),
+            prefixScanSensor
+        );
```
cadonna: nit: 
```suggestion
        return new MeteredKeyValueIterator(wrapped().prefixScan(prefix, prefixKeySerializer), prefixScanSensor);
```

streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
```diff
@@ -434,6 +435,23 @@ public void shouldRemoveMetricsEvenIfWrappedStoreThrowsOnClose() {
         verify(inner);
     }
 
+    @Test
+    public void shouldGetRecordsWithPrefixKey() {
+        final StringSerializer stringSerializer = new StringSerializer();
+        expect(inner.prefixScan(KEY, stringSerializer))
+                .andReturn(new KeyValueIteratorStub<>(Collections.singletonList(BYTE_KEY_VALUE_PAIR).iterator()));
```
cadonna: nit: Please use 4 instead of 8 spaces indentation.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -627,6 +645,11 @@ public void addToBatch(final byte[] key,
         public void close() {
             columnFamily.close();
         }
+
+        @Override
+        public KeyValueIterator<Bytes, byte[]> prefixScan(final Bytes prefix) {
+            return new RocksDBPrefixIterator(name, db.newIterator(columnFamily), openIterators, prefix);
+        }
```
cadonna: Sorry, if I haven't noticed it before. Could you move this method up before `approximateNumEntries()` to have all operations in one block?

streams/src/main/java/org/apache/kafka/streams/state/internals/NamedCache.java
```diff
@@ -284,6 +284,10 @@ public boolean isEmpty() {
         return keySetIterator(cache.navigableKeySet().subSet(from, true, to, true), true);
     }
 
+    synchronized Iterator<Bytes> keyRange(final Bytes from, final Bytes to, final boolean toInclusive) {
+        return keySetIterator(cache.navigableKeySet().subSet(from, true, to, toInclusive), true);
+    }
```
cadonna: nit: I would just add parameter `toInclusive` to the existing method instead of creating an overload.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.streams.KeyValue;
+import org.apache.kafka.streams.state.KeyValueIterator;
+import org.rocksdb.RocksIterator;
+
+import java.nio.ByteBuffer;
+import java.util.Set;
+
+class RocksDBPrefixIterator extends RocksDbIterator {
```
cadonna: I actually do not completely understand why we need a specific iterator for the prefix scan. We could just as good extend `RocksDBRangeIterator` to consider or not consider the end result of the range. We can do that because those iterator implementations are internal and the public API does not care which iterator is used as long as it implements interface `KeyValueIterator` and the behavior is correct. Extending and re-using `RocksDBRangeIterator` would lead to less code to maintain. Note, I agree that we need the public method `prefixScan()`, but what the implementation uses internally is not relevant for the KIP as long as it is correct. Did I miss something that imposes the implementation for a separate iterator?

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
```diff
@@ -201,6 +201,14 @@ public MemoryLRUCacheBytesIterator reverseAll(final String namespace) {
...
+        final NamedCache cache = getCache(namespace);
+        if (cache == null) {
+            return new MemoryLRUCacheBytesIterator(Collections.emptyIterator(), new NamedCache(namespace, this.metrics));
+        }
+        return new MemoryLRUCacheBytesIterator(cache.keyRange(from, to, false), cache);
```
cadonna: If you consider my comment in `NamedCache`, you could also just call `range()` here. Or -- if you want to be more descriptive -- having `prefixScan()` and `range()` in this class calling a private overload of `range()` with a flag that excludes or includes the end of the range. That would deduplicate code.

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetrics.java
```diff
@@ -307,6 +315,26 @@ public static Sensor rangeSensor(final String threadId,
         );
     }
 
+    public static Sensor prefixScanSensor(final String threadId,
+                                     final String taskId,
+                                     final String storeType,
+                                     final String storeName,
+                                     final StreamsMetricsImpl streamsMetrics) {
+        return throughputAndLatencySensor(
```
cadonna: For new sensors like this, we only need to consider built-in metrics version `LATEST`. Hence, you should not call `throughputAndLatencySensor()`, but only call the parts that are relevant for `LATEST` and not for `FROM_0100_TO_24`. You also need to adapt the corresponding unit test for that. See also KIP-444 for more details.

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingInMemoryKeyValueStoreTest.java
```diff
@@ -359,6 +361,34 @@ public void shouldReverseIterateOverRange() {
         ), results);
     }
 
+    @Test
+    public void shouldGetRecordsWithPrefixKey() {
```
cadonna, line 365: In all the tests for the prefix scan you should also verify boundary conditions, e.g., if you have a prefix `abcd`, you should verify that `abce` is not matched since this is the first key that should not be matched.

cadonna (commented):
Thanks @vamossagar12 for the updates!

I am wondering if we really need a separate prefix iterator or if we could just extend the range iterator and use that. See my comment in the respective iterator class.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBPrefixIterator.java
vamossagar12: Well the only reason i chose to add a separate iterator is that for prefix scan there is no end key which could be known upfront. We could pass a null end key but that's prohibited in RocksDBRangeIterator constructor and i don't think we should be changing that condition or we could pass in the same value or from and to and add a condition in makeNext() for prefix scan. i thought having a separate iterator might be cleaner. You have some other approach in mind?

cadonna: Could you not use the same technique as you use in the in-memory key-value state store and the cache?

```
        final Bytes from = Bytes.wrap(prefixKeySerializer.serialize(null, prefix));
        final Bytes to = Bytes.increment(from);
```

vamossagar12: I tested with that approach. It worked for all cases apart from the abcd/abce case that you mentioned below. So, we will need a way to signal to the Iterator that for the case of range iteration for prefixScan, it should ignore the last key. I suppose we will have to add another flag similar to the ones used in in memory key-value store.
BTW, the other reason why I created a separate prefix iterator was that there was an old unused prefix iterator, So, I thought of repurposing it :)

vamossagar12: hey @cadonna , did you get a chance to look at the comment?

cadonna: Sorry for the late reply but I was on holidays.

> I suppose we will have to add another flag similar to the ones used in in memory key-value store.

This is what I meant with "extend RocksDBRangeIterator". 

I think less code is easier maintainable. In addition, I think the code is easier to follow if cache, in-memory, and RocksDB use the same technique for the prefix scan unless we have a technique with better performance that is specific to one of those components. So I would be in favor of just using `RocksDBRangeIterator`.

vamossagar12: Thanks for the reply. Sure, I would make the relevant changes.

cadonna: Awesome!

vamossagar12: As discussed, I have extended RocksDBRangeIterator to include the toInclusive flag. 
1 point to note here is that with this change, RocksDBPrefixIterator is no longer needed so I have removed it and it's relevant test case. 
Also, the unit tests for prefix scan, i have a separate PR [9717](https://github.com/apache/kafka/pull/9717) where i am writing the unit tests for RangeIterator. I plan to add the prefix scan cases there. would that be ok?

streams/src/test/java/org/apache/kafka/streams/state/internals/CachingInMemoryKeyValueStoreTest.java
vamossagar12: i have added wherever the boundary conditions were not getting tested.

vamossagar12:
@cadonna , made further changes. Plz review whenever you get a chance to do so!

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStore.java
```diff
@@ -229,6 +232,13 @@ public V delete(final K key) {
         }
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<K, V> prefixScan(final P prefix, final PS prefixKeySerializer) {
+
+        return new MeteredKeyValueIterator(wrapped().prefixScan(prefix,
+            prefixKeySerializer), prefixScanSensor);
```
cadonna: nit:
```suggestion
        return new MeteredKeyValueIterator(wrapped().prefixScan(prefix, prefixKeySerializer), prefixScanSensor);
```

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -594,6 +616,20 @@ public void prepareBatch(final List<KeyValue<Bytes, byte[]>> entries,
...
+        public KeyValueIterator<Bytes, byte[]> prefixScan(final Bytes prefix) {
+            final Bytes to = Bytes.increment(prefix);
+            return new RocksDBRangeIterator(
+                name,
+                db.newIterator(columnFamily),
+                openIterators,
+                prefix,
+                to,
+                true,
+                false);
```
cadonna:
```suggestion
            return new RocksDBRangeIterator(
                name,
                db.newIterator(columnFamily),
                openIterators,
                prefix,
                to,
                true,
                false
            );
```

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -594,6 +616,20 @@ public void prepareBatch(final List<KeyValue<Bytes, byte[]>> entries,
...
+
```
cadonna: Please remove empty line.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -308,6 +309,24 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
...
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix,
+                                                                                    final PS prefixKeySerializer) {
+        Objects.requireNonNull(prefix, "prefix cannot be null");
+        Objects.requireNonNull(prefixKeySerializer, "prefixKeySerializer cannot be null");
+
+        validateStoreOpen();
+        final Bytes prefixBytes = Bytes.wrap(prefixKeySerializer.serialize(null, prefix));
+
+        /*final Bytes from = Bytes.wrap(prefixKeySerializer.serialize(null, prefix));
+        final Bytes to = Bytes.increment(from);*/
```
cadonna: I guess you forgot to remove this. :slightly_smiling_face:

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
```diff
@@ -360,6 +361,130 @@ public void shouldPutAll() {
...
+        entries.add(new KeyValue<>(
+            new Bytes(stringSerializer.serialize(null, "a")),
+            stringSerializer.serialize(null, "a")));
+        entries.add(new KeyValue<>(
+            new Bytes(stringSerializer.serialize(null, "b")),
+            stringSerializer.serialize(null, "c")));
+        entries.add(new KeyValue<>(
+            new Bytes(stringSerializer.serialize(null, "c")),
+            stringSerializer.serialize(null, "e")));
+
```
cadonna: Please remove empty line.

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
```diff
@@ -360,6 +361,130 @@ public void shouldPutAll() {
...
+    public void shouldReturnUUIDsWithStringPrefix() {
+        final List<KeyValue<Bytes, byte[]>> entries = new ArrayList<>();
+        final Serializer<UUID> uuidSerializer = Serdes.UUID().serializer();
+        final UUID uuid1 = UUID.randomUUID();
+        final UUID uuid2 = UUID.randomUUID();
+        final String prefix = uuid1.toString().substring(0, 4);
+        entries.add(new KeyValue<>(
+            new Bytes(uuidSerializer.serialize(null, uuid1)),
+            stringSerializer.serialize(null, "a")));
+
```
cadonna: Please remove empty line.

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBStoreTest.java
```diff
@@ -360,6 +361,130 @@ public void shouldPutAll() {
                 rocksDBStore.get(new Bytes(stringSerializer.serialize(null, "3")))));
     }
 
+    @Test
+    public void shouldReturnKeysWithGivenPrefix() {
```
cadonna, line 365: Could you split up this unit test into two, one for prefix `prefix` and one for prefix `abcd`? The former tests the general case whereas the latter tests the corner case where the the end of the range should not be returned.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -580,7 +601,8 @@ public void prepareBatch(final List<KeyValue<Bytes, byte[]>> entries,
                 openIterators,
                 from,
                 to,
-                forward);
+                forward,
+                true);
```
cadonna:
```suggestion
                true
            );
```

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStore.java
```diff
@@ -218,6 +219,19 @@ public void prepareBatch(final List<KeyValue<Bytes, byte[]>> entries,
...
+        public KeyValueIterator<Bytes, byte[]> prefixScan(final Bytes prefix) {
+            final Bytes to = Bytes.increment(prefix);
+            return new RocksDBDualCFRangeIterator(
+                name,
+                db.newIterator(newColumnFamily),
+                db.newIterator(oldColumnFamily),
+                prefix,
+                to,
+        true,
+      false);
```
cadonna: I think here IntelliJ's formatting confused you. 🙂 
```suggestion
                to,
                true,
                false
            );
```

streams/src/test/java/org/apache/kafka/streams/state/internals/RocksDBTimestampedStoreTest.java
```diff
@@ -337,6 +341,22 @@ private void iteratorsShouldNotMigrateData() {
             }
             assertFalse(it.hasNext());
         }
+
+        try (final KeyValueIterator<Bytes, byte[]> it =
+                     rocksDBStore.prefixScan("key1", stringSerializer)) {
```
cadonna:
```suggestion
        try (final KeyValueIterator<Bytes, byte[]> it = rocksDBStore.prefixScan("key1", stringSerializer)) {
```

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetrics.java
```diff
@@ -307,6 +315,34 @@ public static Sensor rangeSensor(final String threadId,
...
+
+        final Sensor sensor = streamsMetrics.storeLevelSensor(taskId, storeName, PREFIX_SCAN, RecordingLevel.DEBUG);
+        addInvocationRateToSensor(
+            sensor,
+            STATE_STORE_LEVEL_GROUP,
+            tagMap,
+            PREFIX_SCAN,
+            PREFIX_SCAN_RATE_DESCRIPTION
+        );
+
```
cadonna: Please remove empty line.

streams/src/main/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetrics.java
```diff
@@ -110,6 +110,14 @@ private StateStoreMetrics() {}
     private static final String RANGE_AVG_LATENCY_DESCRIPTION = AVG_LATENCY_DESCRIPTION_PREFIX + RANGE_DESCRIPTION;
     private static final String RANGE_MAX_LATENCY_DESCRIPTION = MAX_LATENCY_DESCRIPTION_PREFIX + RANGE_DESCRIPTION;
 
+    private static final String PREFIX_SCAN = "prefix-scan";
+    private static final String PREFIX_SCAN_DESCRIPTION = "calls to prefix-scan";
+    private static final String PREFIX_SCAN_TOTAL_DESCRIPTION = TOTAL_DESCRIPTION + PREFIX_SCAN_DESCRIPTION;
```
cadonna: This description is not needed since there is no total metric.

streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
```diff
@@ -434,6 +435,25 @@ public void shouldRemoveMetricsEvenIfWrappedStoreThrowsOnClose() {
...
+    @Test
+    public void shouldGetRecordsWithPrefixKey() {
+        final StringSerializer stringSerializer = new StringSerializer();
+        inner.put(eq(Bytes.increment(KEY_BYTES)), aryEq(VALUE_BYTES));
+        expectLastCall();
+        expect(inner.prefixScan(KEY, stringSerializer))
+            .andReturn(new KeyValueIteratorStub<>(Collections.singletonList(BYTE_KEY_VALUE_PAIR).iterator()));
+        init();
+
+        metered.put(Bytes.increment(KEY_BYTES).toString(), VALUE);
```
cadonna: I do not think you need to put an entry if you use mocks.

streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
```diff
@@ -434,6 +435,25 @@ public void shouldRemoveMetricsEvenIfWrappedStoreThrowsOnClose() {
         verify(inner);
     }
 
+    @Test
+    public void shouldGetRecordsWithPrefixKey() {
+        final StringSerializer stringSerializer = new StringSerializer();
+        inner.put(eq(Bytes.increment(KEY_BYTES)), aryEq(VALUE_BYTES));
+        expectLastCall();
```
cadonna: This line is not needed in this case. A method call without a return value is expected on the mock if you simply call the method on the mock in the replay phase.

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
```diff
@@ -218,6 +222,10 @@ public MemoryLRUCacheBytesIterator reverseAll(final String namespace) {
         return new MemoryLRUCacheBytesIterator(cache.reverseAllKeys(), cache);
     }
 
+    public MemoryLRUCacheBytesIterator prefixScan(final String namespace, final Bytes from, final Bytes to) {
```
cadonna: I think we do not need this method. We can just call `range()` in `CachingKeyValueStore`.

streams/src/test/java/org/apache/kafka/streams/state/internals/ThreadCacheTest.java
```diff
@@ -285,6 +285,25 @@ public void shouldPeekAndIterateOverRange() {
         assertEquals(5, bytesIndex);
     }
 
+    @Test
+    public void testPrefixScan() {
```
cadonna: If we remove `prefixScan()`, you can remove this test, but we need a test for `range()` that excludes the end limit.

streams/src/test/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetricsTest.java
```diff
@@ -204,6 +204,42 @@ public void shouldGetRangeSensor() {
...
+        expect(streamsMetrics.storeLevelTagMap(TASK_ID, STORE_TYPE, STORE_NAME)).andReturn(storeTagMap);
+
+        StreamsMetricsImpl.addInvocationRateToSensor(
+            expectedSensor,
+            STORE_LEVEL_GROUP,
+            storeTagMap,
+            metricName,
+            descriptionOfRate
+        );
+
```
cadonna: Please remove empty line.

streams/src/test/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetricsTest.java
```diff
@@ -204,6 +204,42 @@ public void shouldGetRangeSensor() {
...
+    public void shouldGetPrefixScanSensor() {
+        final String metricName = "prefix-scan";
+        final String descriptionOfRate = "The average number of calls to prefix-scan per second";
+        final String descriptionOfAvg = "The average latency of calls to prefix-scan";
+        final String descriptionOfMax = "The maximum latency of calls to prefix-scan";
+
+        expect(streamsMetrics.storeLevelSensor(TASK_ID, STORE_NAME, metricName, RecordingLevel.DEBUG))
+            .andReturn(expectedSensor);
+        expect(streamsMetrics.storeLevelTagMap(TASK_ID, STORE_TYPE, STORE_NAME)).andReturn(storeTagMap);
+
```
cadonna: Please remove empty line.

streams/src/test/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetricsTest.java
```diff
@@ -204,6 +204,42 @@ public void shouldGetRangeSensor() {
         );
     }
 
+    @Test
+    public void shouldGetPrefixScanSensor() {
+        final String metricName = "prefix-scan";
+        final String descriptionOfRate = "The average number of calls to prefix-scan per second";
+        final String descriptionOfAvg = "The average latency of calls to prefix-scan";
+        final String descriptionOfMax = "The maximum latency of calls to prefix-scan";
+
```
cadonna: Please remove empty line.

cadonna (commented):
@vamossagar12, Thanks  for the updates.

I have a couple of comments, but I also think that we are almost there.

vamossagar12:
@cadonna , done.

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingKeyValueStore.java
```diff
@@ -291,6 +292,16 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
         return new MergedSortedCacheKeyValueBytesStoreIterator(cacheIterator, storeIterator, true);
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, final PS prefixKeySerializer) {
```
cadonna, line 296: This method needs unit testing. Try to use a mock for the cache in the test.

streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
```diff
@@ -196,6 +201,26 @@ public void shouldReturnValueOnGetWhenExists() {
         assertThat(store.get(hello), equalTo(world));
     }
 
+    @Test
+    public void shouldGetRecordsWithPrefixKey() {
+        store.put(hi, there);
+        store.put(Bytes.increment(hi), world);
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = store.prefixScan("hi", new StringSerializer());
```
cadonna:
```suggestion
        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = store.prefixScan(hi.toString(), new StringSerializer());
```
In such a way, you can reuse variable `hi` and `there`. Similar is true for my suggestions below.

streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
```diff
@@ -196,6 +201,26 @@ public void shouldReturnValueOnGetWhenExists() {
...
+        final List<String> keys = new ArrayList<>();
+        final List<String> values = new ArrayList<>();
```
cadonna:
```suggestion
        final List<Bytes> keys = new ArrayList<>();
        final List<Bytes> values = new ArrayList<>();
```

streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
```diff
@@ -196,6 +201,26 @@ public void shouldReturnValueOnGetWhenExists() {
...
+        int numberOfKeysReturned = 0;
+
+        while (keysWithPrefix.hasNext()) {
+            final KeyValue<Bytes, byte[]> next = keysWithPrefix.next();
+            keys.add(next.key.toString());
+            values.add(new String(next.value));
```
cadonna:
```suggestion
            keys.add(next.key);
            values.add(Bytes.wrap(next.value));
```

streams/src/test/java/org/apache/kafka/streams/state/internals/ChangeLoggingKeyValueBytesStoreTest.java
```diff
@@ -196,6 +201,26 @@ public void shouldReturnValueOnGetWhenExists() {
...
+            numberOfKeysReturned++;
+        }
+        assertThat(numberOfKeysReturned, is(1));
+        assertThat(keys, is(Collections.singletonList("hi")));
+        assertThat(values, is(Collections.singletonList("there")));
```
cadonna:
```suggestion
        assertThat(keys, is(Collections.singletonList(hi)));
        assertThat(values, is(Collections.singletonList(Bytes.wrap(there))));
```

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
```diff
@@ -103,6 +105,20 @@ public void putAll(final List<KeyValue<Bytes, byte[]>> entries) {
         }
     }
 
+    @Override
+    public <PS extends Serializer<P>, P> KeyValueIterator<Bytes, byte[]> prefixScan(final P prefix, final PS prefixKeySerializer) {
```
cadonna, line 109: @vamossagar12 I can still not find the unit test for this method.

streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
```diff
@@ -434,6 +435,22 @@ public void shouldRemoveMetricsEvenIfWrappedStoreThrowsOnClose() {
...
+        final StringSerializer stringSerializer = new StringSerializer();
+        expect(inner.prefixScan(KEY, stringSerializer))
+            .andReturn(new KeyValueIteratorStub<>(Collections.singletonList(BYTE_KEY_VALUE_PAIR).iterator()));
+        init();
+
+        final KeyValueIterator<String, String> iterator = metered.prefixScan(KEY, stringSerializer);
+        assertThat(iterator.next().value, equalTo(VALUE));
+        iterator.close();
+
+        final KafkaMetric metric = metrics.metric(new MetricName("prefix-scan-rate", STORE_LEVEL_GROUP, "", tags));
```
cadonna, line 449: nit:
```suggestion
        final KafkaMetric metric = metric("prefix-scan-rate");
```

streams/src/test/java/org/apache/kafka/streams/state/internals/metrics/StateStoreMetricsTest.java
```diff
@@ -204,6 +204,39 @@ public void shouldGetRangeSensor() {
...
+            STORE_LEVEL_GROUP,
+            storeTagMap,
+            latencyMetricName(metricName),
+            descriptionOfAvg,
+            descriptionOfMax
+        );
+        replay(StreamsMetricsImpl.class, streamsMetrics);
+
+        final Sensor sensor =
+                StateStoreMetrics.prefixScanSensor(TASK_ID, STORE_TYPE, STORE_NAME, streamsMetrics);
```
cadonna: nit:
```suggestion
        final Sensor sensor = StateStoreMetrics.prefixScanSensor(TASK_ID, STORE_TYPE, STORE_NAME, streamsMetrics);
```

cadonna (commented):
Thanks for the updates, @vamossagar12 !

cadonna:
@vamossagar12 Could you please add a more descriptive commit message?

cadonna:
Call for committer review: @mjsax @guozhangwang @ableegoldman

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
vamossagar12: For this, do you want me to add the test cases here?https://github.com/apache/kafka/blob/17be91a37214bf77430c65d9300a5120e4348df9/streams/src/test/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStoreTest.java

There are tests in CachingInMemoryKeyValueStoreTest, which is where the tests for other methods like range etc have been added.

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingKeyValueStore.java
vamossagar12: Similar comment as below. Unit tests are in CachingInMemoryKeyValueStoreTest which already extends AbstractKeyValueStoreTest and creates an in memory cache store.

guozhangwang (approved):
I made a quick pass on the PR and it looks good to me. After test is green I can merge it.

@cadonna could you also help @vamossagar12 to come up with a PR on docs for this KIP?

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingKeyValueStore.java
cadonna: Oh, I see. I missed those. Sorry! That is fine then, although I think unit tests with mocks would be better.

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
cadonna: I think those tests never call the `prefixScan()` on the underlying in-memory state store because all entries fit into the cache. You would need to add another test that flushes the cache before you call `prefixScan()`. I would prefer a test that directly tests the in-memory store without any cache in between.

streams/src/test/java/org/apache/kafka/streams/state/internals/MeteredKeyValueStoreTest.java
```diff
@@ -434,6 +435,22 @@ public void shouldRemoveMetricsEvenIfWrappedStoreThrowsOnClose() {
...
+        final StringSerializer stringSerializer = new StringSerializer();
+        expect(inner.prefixScan(KEY, stringSerializer))
+            .andReturn(new KeyValueIteratorStub<>(Collections.singletonList(BYTE_KEY_VALUE_PAIR).iterator()));
+        init();
+
+        final KeyValueIterator<String, String> iterator = metered.prefixScan(KEY, stringSerializer);
+        assertThat(iterator.next().value, equalTo(VALUE));
+        iterator.close();
+
+        final KafkaMetric metric = metric("prefix-scan-rate");
```
cadonna: To get rid of the test failure, you need to change this:

```suggestion
        final KafkaMetric metric = metrics.metric(new MetricName("prefix-scan-rate", STORE_LEVEL_GROUP, "", tags));
```

Sorry, the failure of the test is my bad. I missed the issue with the different metrics versions when I requested to change this in a previous review.

cadonna:
@guozhangwang @vamossagar12 This failure is related to this PR:

```
org.apache.kafka.streams.state.internals.MeteredKeyValueStoreTest.shouldGetRecordsWithPrefixKey[0.10.0-2.4]
```

I provided the fix for this in the corresponding test.

cadonna:
@vamossagar12 Probably, it would also be good to rebase this PR to current trunk to get a green build since we merged a couple of fixes to flaky tests recently.

cadonna:
> @cadonna could you also help @vamossagar12 to come up with a PR on docs for this KIP?

Yes, @vamossagar12, you should open a PR and add a short paragraph about KIP-614 in the [upgrade guide](https://github.com/apache/kafka/blob/821d867c1bd01b88aa695827cb57c9aa764857a4/docs/streams/upgrade-guide.html#L124). Looking forward to reviewing it! 🙂

guozhangwang:
I've merged this PR to trunk. Thanks @vamossagar12 so much for your contribution! And thanks to @cadonna for the very thorough review!

vamossagar12:
Thanks @guozhangwang !
@cadonna I have added the doc here in this PR: https://github.com/apache/kafka/pull/10051.
I will create a separate. ticket for the missing test case for prefixScan. Thanks for your continued support @cadonna !

streams/src/main/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStore.java
vamossagar12: Here is the new ticket: https://issues.apache.org/jira/browse/KAFKA-12289 and the PR for the ticket:
https://github.com/apache/kafka/pull/10052

streams/src/main/java/org/apache/kafka/streams/state/internals/CachingKeyValueStore.java
vamossagar12: Actually, I had created another ticket to streamline tests for CachingKVStore: https://issues.apache.org/jira/browse/KAFKA-10788. @rohitrmd  had volunteered to take this up.

PR #9509, opened 2020-10-27
MINOR: rename metrics "iotime-total" to "io-time-total" and "io-waitt…

chia7712:
I believe that is a kind of typo since it is not consistent to other related metrics name.

1. io-wait-time-ns-avg
1. io-time-ns-avg
1. io-wait-ratio
1. io-ratio


dajac:
@chia7712 Thanks for the PR. Renaming it may break monitoring of existing users. As metrics are part of the public API, we may need a KIP to fix this...

chia7712:
@dajac thanks for reminder. I don't think this patch is worth doing that effort. close it.

PR #9512, opened 2020-10-27
KAFKA-10394: generate snapshot

jsancio:
This PR adds support for generating snapshot for KIP-630.

1. Adds the interfaces `RawSnapshotWriter` and `RawSnapshotReader` and the implementations `FileRawSnapshotWriter` and `FileRawSnapshotReader` respectively. These interfaces and implementations are low level API for writing and reading snapshots. They are internal to the Raft implementation and are not exposed to the users of `RaftClient`. They operation at the `Record` level. These types are exposed to the `RaftClient` through the `ReplicatedLog` interface.

2. Adds a buffered snapshot writer: `SnapshotWriter<T>`. This type is a higher-level type and it is exposed through the `RaftClient` interface. A future PR will add the related `SnapshotReader<T>`, which will be used by the state machine to load a snapshot.

comment to ping reviewers.*


raft/src/main/java/org/apache/kafka/snapshot/SnapshotReader.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Iterator;
+import org.apache.kafka.common.record.RecordBatch;
+import org.apache.kafka.raft.OffsetAndEpoch;
+
+// TODO: Write documentation for this type and all of the methods
+public interface SnapshotReader extends Closeable, Iterable<RecordBatch> {
```
hachikuji: It would be nice if we can figure out how to consolidate this and `BatchReader`. It seems like it should be doable.

raft/src/main/java/org/apache/kafka/snapshot/SnapshotReader.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public OffsetAndEpoch snapshotId();
+
+    public long sizeInBytes();
+
+    public Iterator<RecordBatch> iterator();
+
+    public int read(ByteBuffer buffer, long position) throws IOException;
```
hachikuji: So I guess we need this in order to be able to serve `FetchSnapshot` requests. I am wondering if it would be better to pull it into a separate object. Maybe we can have something like this:
```java
class SnapshotReader extends BatchReader;

class Snapshot extends Iterable<Batch> {
  public SnapshotReader iterator();
  public int read(ByteBuffer buffer, long position) throws IOException;
}
```

core/src/main/scala/kafka/snapshot/KafkaSnapshotReader.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import java.nio.ByteBuffer
+import java.nio.file.Path
+import java.util.{Iterator => JIterator}
+import org.apache.kafka.common.record.RecordBatch
+import org.apache.kafka.common.record.FileRecords
+import org.apache.kafka.raft.OffsetAndEpoch
+import org.apache.kafka.snapshot.SnapshotReader
+
+final class KafkaSnapshotReader private (fileRecords: FileRecords, snapshotId: OffsetAndEpoch) extends SnapshotReader {
```
hachikuji: So far we have preferred to keep implementations in the `raft` package except in cases where we depended on something that was in the server code (such as `Log`). Is there any reason to do snapshots differently? I think we would still be able to use the snapshot code from `raft` for other internal use cases (like __consumer_offsets).

raft/src/main/java/org/apache/kafka/snapshot/BatchedSnapshotWriter.java
```diff
@@ -0,0 +1,107 @@
...
+import org.apache.kafka.common.memory.MemoryPool;
+import org.apache.kafka.common.record.CompressionType;
+import org.apache.kafka.common.utils.Time;
+import org.apache.kafka.raft.OffsetAndEpoch;
+import org.apache.kafka.raft.RecordSerde;
+import org.apache.kafka.raft.internals.BatchAccumulator.CompletedBatch;
+import org.apache.kafka.raft.internals.BatchAccumulator;
+
+// TODO: Write documentation for this type and all of the methods
+final public class BatchedSnapshotWriter<T> implements Closeable {
```
hachikuji: The types surprised me a bit. I would have expected this to extend `SnapshotWriter`. I'm wondering why we don't push the type `<T>` up to `SnapshotWriter` and try to keep `Records` out of the interfaces?

hachikuji (commented):
Thanks for the PR. Left a few initial questions.

core/src/main/scala/kafka/snapshot/KafkaSnapshotReader.scala
jsancio: Sounds good. I'll move this code over and implement it in Java.

raft/src/main/java/org/apache/kafka/snapshot/SnapshotReader.java
jsancio: This comment applies to some of your other observations. At high-level there are 4 use cases that we need to design and implement. Two use cases are for the Raft implementation. Two use cases are for the state machine.

### Raft implementation
These types are internal to the raft implementation and don't have to be exposed to the state machine.
#### Leader Use Case
The leader needs to be able to send a part of the snapshot over the network. Something like this
```java
interface SnapshotReader extends Closeable {
    long transferTo(long position, long maxBytes, WritableChannel channel);
    int read(ByteBuffer buffer, long position);
}
```
#### Follower Use Case
The followers need to be able to copy bytes from the network and validate the snapshot on disk when fetching is done.
```java
interface SnapshotWriter extends Closeable {
    void append(ByteBuffer buffer);
    void validate();
    void freeze();
}
```
### State machine implementation
These types are exposed to the state machine.
#### Load Snapshot
The state machine needs to be able to load/scan the entire snapshot. The state machine can use `close` to tell the raft client that it finished loading the snapshot. This will be implemented in a future PR but it could look like this:
```java
interface BatchedSnapshotReader<T> extends Iterable<Iterable<T>>, Closeable {
}
```
#### Generate Snapshot
The state machine needs to be able to generate a snapshot by appending records/values and marking the snapshot as immutable (`freeze`) when it is done.
```java
interface BatchdSnapshotWriter<T> extends Closeable {
   void append(Iterable<T> records);
   void freeze();
}
```

### Notes

`SnapshotWriter` and `SnapshotReader` need to be interfaces because we will have a real implementation and a mocked implementation for testing. These two types are internal to raft and are not exposed to the state machine.

`BatchedSnapshotReader` and `BatchedSnapshotWriter` can depend on `SnapshotWriter` and `SnapshotReade` to reuse some code but this is not strictly required. These two type don't have to be `interfaces` if they delegate the IO to `SnapshotWriter` and `SnapshotReader`.

What do you think @hachikuji?

raft/src/main/java/org/apache/kafka/snapshot/SnapshotReader.java
jsancio: I think I address this in this comment https://github.com/apache/kafka/pull/9512#discussion_r517703078. If you agree let's move the conversation there.

raft/src/main/java/org/apache/kafka/snapshot/BatchedSnapshotWriter.java
jsancio: I cover some of the motivation for this here: https://github.com/apache/kafka/pull/9512#discussion_r517703078.

Let's move the conversation there if you agree.

core/src/main/scala/kafka/snapshot/KafkaSnapshotReader.scala
jsancio: Done.

raft/src/main/java/org/apache/kafka/snapshot/SnapshotReader.java
hachikuji: Ok, I think I see what you're saying. I guess I was not expecting that we would need a separate low-level interface when we already have `FileRecords` and `MemoryRecords`. I think `Records` already gives us a way to read from arbitrary positions:
```java
    long writeTo(GatheringByteChannel channel, long position, int length) throws IOException;
```
But there is no `Records` implementation currently that allows unaligned writes. We only have this:
```java
    public int append(MemoryRecords records) throws IOException;
```
Perhaps it would be possible to extend `Records` to provide what we need instead of creating a new interface?

As far as the naming, I wonder if we can reserve the nicer `SnapshotXXX` names for the state machine. Really what I would like is a common type that can be used by both `handleSnapshot` and `handleCommit` since both callbacks just need to provide a way to read through a set of records. I was thinking it could be `BatchReader`, but it doesn't have to be if there is something better.

(By the way, it's not too clear to me why we need `freeze` when we already have `close`.)

jsancio: > Perhaps it would be possible to extend Records to provide what we need instead of creating a new interface?

I think you are right that we don't need `SnapshotReader` as `Records` provides all of the functionality we need. `SnapshotReader` was added so that we didn't expose all of the mutating APIs in `FileRecords`. Snapshot are supposed to be immutable once frozen.

I think we still want `SnapshotWriter` or something similar as it provides 1. raw writes of bytes and 2. `freeze` which optionally marks the snapshot as immutable if the validation passes.

For 1., I don't think we should add raw writes to `Records` or `FileRecords` as in essence we are exposing an unsafe API and the user needs to make sure that they are writing the correct data.

For 2., I think we can get away from introducing a new type/interface and instead add that functionality as a static method in `Snapshots.java`. Unittest (mock tests) maybe difficult with this code organization.

> As far as the naming, I wonder if we can reserve the nicer SnapshotXXX names for the state machine.

Yes.I'll use the prettier name for the type exposed to the state machine.

> Really what I would like is a common type that can be used by both handleSnapshot and handleCommit since both callbacks just need to provide a way to read through a set of records. I was thinking it could be BatchReader, but it doesn't have to be if there is something better.

I think this should be possible. I haven't implemented this part so I don't have all of the details. I think the requirement for the type sent through `handleSnapshot` are a subset of the requirements for the type sent through `handleCommit`. In particular, the Raft Client (`ListenerContext`) only needs to know when the snapshot has been read fully (e.g. `close`). The Raft Client doesn't need so to know what is the "last seen offset + 1" as it already knows the `SnapshotId` sent through `handleSnapshot`.

> (By the way, it's not too clear to me why we need freeze when we already have close.)

I wanted to allow the user to abort a snapshot. If `close` is called without calling `freeze` then the partial snapshot is deleted and not made immutable.

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
```diff
@@ -32,25 +32,25 @@
  * This class is responsible for managing the current state of this node and ensuring only
  * valid state transitions.
  *
- * Unattached =>
+ * Unattached transitions to:
```
jsancio: Changes to this file are unrelated to this PR. Made them here so that `docsJar` would succeed.

raft/src/main/java/org/apache/kafka/snapshot/RawSnapshotWriter.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Fully appends the buffer to the snapshot.
+     *
+     * If the method returns without an exception the given buffer was fully writing the
+     * snapshot.
+     *
+     * @param buffer the buffer to append
+     * @throws IOException for any IO error during append
+     */
+    public void append(ByteBuffer buffer) throws IOException;
```
jsancio, line 50: As @hachikuji and I discussed offline, it is very like that this method will change to:

```java
public void append(MemoryRecords records) throws IOException;
```

After we implement https://issues.apache.org/jira/browse/KAFKA-10694

raft/src/main/java/org/apache/kafka/snapshot/RawSnapshotReader.java
```diff
@@ -0,0 +1,52 @@
...
+     * Reads bytes from position into the given buffer.
+     *
+     * It is not guarantee that the given buffer will be filled.
+     *
+     * @param buffer byte buffer to put the read files
+     * @param position the starting position in the snapshot to read
+     * @return the number of bytes read
+     * @throws IOException for any IO error while reading the snapshot
+     */
+    public int read(ByteBuffer buffer, long position) throws IOException;
```
jsancio, line 51: As @hachikuji and I discussed offline, it is very like that this method will change to:

```java
public BaseRecords slice(long position) throws IOException;
```

After we implement https://issues.apache.org/jira/browse/KAFKA-10694

raft/src/main/java/org/apache/kafka/snapshot/SnapshotReader.java
jsancio: @hachikuji and I discussed this offline. The latest PR should reflect our discussion.

The interfaces `RawSnapshotWriter` and `RawSnapshotReader` are expected to change after we address https://issues.apache.org/jira/browse/KAFKA-10694

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
```diff
@@ -100,4 +102,15 @@ default void handleResign() {}
...
     CompletableFuture<Void> shutdown(int timeoutMs);
 
+    /**
+     * Create a writable snapshot file for a given offset and epoch.
+     *
+     * The RaftClient assumes that the snapshot return will contain the records up to but
+     * not including the end offset in the snapshot id. See {@link SnapshotWriter} for
+     * details on how to use this object.
+     *
+     * @param snapshotId the end offset and epoch that identifies the snapshot
```
hachikuji, line 112: Do we assume the offset is below the high watermark?

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.text.NumberFormat;
+import org.apache.kafka.raft.OffsetAndEpoch;
+
+final class Snapshots {
+    private static final String SNAPSHOT_DIR = "snapshots";
+    private static final String SUFFIX =  ".snapshot";
```
hachikuji: This suffix is used for producer state snapshots already. Maybe we could use `.snap` or something like that.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
```diff
@@ -149,6 +152,29 @@ default OptionalLong truncateToEndOffset(OffsetAndEpoch endOffset) {
...
 
+    /**
+     * Create a writable snapshot for the given snapshot id.
+     *
+     * See {@link RawSnapshotWriter} for details on how to use this object.
+     *
+     * @param snapshotId the end offset and epoch that identifies the snapshot
+     * @return a writable snapshot
+     */
+    RawSnapshotWriter createSnapshot(OffsetAndEpoch snapshotId) throws IOException;
```
hachikuji, line 163: Do we also need an api to list snapshots?

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
```diff
@@ -0,0 +1,67 @@
...
+package org.apache.kafka.snapshot;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.text.NumberFormat;
+import org.apache.kafka.raft.OffsetAndEpoch;
+
+final class Snapshots {
+    private static final String SNAPSHOT_DIR = "snapshots";
```
hachikuji: Is there a tangible benefit to separating snapshots into a new directory? Currently the log directory is a flat structure. I'm wondering if we should stick with convention.

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        Path destination = Snapshots.moveRename(path, snapshotId);
+        Files.move(path, destination, StandardCopyOption.ATOMIC_MOVE);
+    }
+
+    @Override
+    public void close() throws IOException {
+        channel.close();
+        Files.deleteIfExists(path);
```
hachikuji: Maybe we can just a better name for `path` since it makes this code look suspicious.

raft/src/main/java/org/apache/kafka/snapshot/SnapshotWriter.java
```diff
@@ -0,0 +1,158 @@
...
+                "Append not supported. Snapshot is already frozen: id = {}.",
+                snapshot.snapshotId()
+            );
+
+            throw new IllegalStateException(message);
+        }
+
+        accumulator.append(snapshot.snapshotId().epoch, records);
+
+        if (!accumulator.needsDrain(time.milliseconds())) {
```
hachikuji: nit: I think this would be more natural:
```java
if (accumulator.needsDrain(time.milliseconds())) {
  appendBatches(accumulator.drain());
}
```

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotReader.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public long sizeInBytes() {
+        return fileRecords.sizeInBytes();
+    }
+
+    @Override
+    public Iterator<RecordBatch> iterator() {
+        return new Iterator<RecordBatch>() {
```
hachikuji: nit: kind of annoying to wrap the iterator just to avoid the generic warning. An alternative might be to use `<? extends RecordBatch>` in the interface.

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
```diff
@@ -0,0 +1,107 @@
...
+        channel.close();
+        frozen = true;
+
+        // Set readonly and ignore the result
+        if (!path.toFile().setReadOnly()) {
+            throw new IOException(String.format("Unable to set file %s as read-only", path));
+        }
+
+        Path destination = Snapshots.moveRename(path, snapshotId);
+        Files.move(path, destination, StandardCopyOption.ATOMIC_MOVE);
```
hachikuji: Wonder if we should consider using `Utils.atomicMoveWithFallback`?

raft/src/test/java/org/apache/kafka/raft/MockLog.java
```diff
@@ -472,4 +489,110 @@ private EpochStartOffset(int epoch, long startOffset) {
         }
     }
 
+    final class MockRawSnapshotWriter implements RawSnapshotWriter {
+        private final OffsetAndEpoch snapshotId;
+        private ByteBuffer data;
```
hachikuji: We could probably use `ByteBufferOutputStream` which already handles expansion.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
jsancio: Is there a specific reason why you are asking this? We don't currently check for this. I will add a check for this and we can relax this later if we need to.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
jsancio: Good catch. I mentioned using `.checkpoint` in KIP-630. I forgot to change it here. I'll change it to that but let me know if you have a preference.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
jsancio: I think this depends on if we need to scan the snapshot directory. Unfortunately, I don't have a concrete answer at the moment. When we implement the changes to the rest of the raft client. Log truncation, updating the start offset and LEO, we may need to scan the snapshot/checkpoint folder to determine the greatest log start offset and LEO. @lbradstreet suggested storing them in a different directory as part of the KIP-630 review process as Kafka already have a few files in the partition log directory.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
jsancio: Yeah. I think we will need that when we implement deleting snapshot. Do you mind if I add this later?

Also, I think we are going to need a `readLatestSnapshot()` when the state machine (controller or broker) needs to load the latest valid snapshot. I was planning to add this later when the case was clear to me.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
hachikuji: Just curious. Thought it might be worth mentioning the expectation in the javadoc.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
hachikuji: Sounds fine to me.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
hachikuji: Not sure if the snapshots will have much of an impact on this. It's not like we'll be storing more than a couple of them.

hachikuji: My preference is probably to keep things flat for now until we figure out what we want the long-term structure to look like. I guess it comes down to the implementation, but intuitively, the only time we'd need to scan would be on startup and we have to do that anyway.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
jsancio: Yeah. I'll create a Jira to validate the snapshot id. We should perform at least the following validations:
1. it is less than the high-watermark.
2. it is a valid epoch and end offset based on the log. I think we can use the leader epoch cache to check this.

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotReader.java
jsancio: I am not sure if we can do that. The interface `RawSnapshotReader` extends `Iterable<RecordBatch>`. `Iterable` defines that method as `Iterator<T> iterator()`. As you point out, they should have defined it as `Iterator<? extends T>  iterator()` since `Iterator` is covariant.

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
jsancio: I prefer to keep it this way. The atomic move (rename) should always succeed because we guarantee it is within the same directory. I think I prefer for Kafka to throw an exception than for the raft client or state machine to see a partial snapshot file because it performed a file copy instead of a file rename.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
jsancio: Cool. Yeah, let's use that. So many util goodies.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
jsancio: Done. Removed the `snapshots` directory.

lbradstreet: @jsancio @hachikuji we saw DeleteRecords calls became pretty expensive and tied up request handler threads when made on big partition directories and it scanned the directory to find deletable snapshot files. If we keep references to these snapshots in memory after log open this won't be a big deal to place them in the same directory. That's what we ended up doing for the producer state snapshots and it worked well.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
jsancio: Create https://issues.apache.org/jira/browse/KAFKA-10800

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
```diff
@@ -74,18 +74,19 @@ public void freeze() throws IOException {
...
+        Path destination = Snapshots.moveRename(tempSnapshotPath, snapshotId);
+        Files.move(tempSnapshotPath, destination, StandardCopyOption.ATOMIC_MOVE);
     }
 
     @Override
     public void close() throws IOException {
         channel.close();
-        Files.deleteIfExists(path);
+        // This is a noop if freeze was called before calling close
+        Files.deleteIfExists(tempSnapshotPath);
```
hachikuji: Would it be useful to call this in a `finally`?

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotReader.java
hachikuji: In that case, I would probably just do the cast to avoid the useless wrapper. Maybe we could create a helper like this in `Utils`:

```java
    private <S, T extends S> Iterator<S> covariantCast(Iterator<T> iterator) {
        return (Iterator<S>) iterator;
    }
```

raft/src/test/java/org/apache/kafka/snapshot/FileRawSnapshotTest.java
```diff
@@ -0,0 +1,289 @@
...
+        Path tempDir = TestUtils.tempDirectory().toPath();
+        OffsetAndEpoch offsetAndEpoch = new OffsetAndEpoch(10L, 3);
+        int bufferSize = 256;
+        int batches = 10;
+        int expectedSize = 0;
+
+        try (FileRawSnapshotWriter snapshot = FileRawSnapshotWriter.create(tempDir, offsetAndEpoch)) {
+            assertEquals(0, snapshot.sizeInBytes());
+
+            MemoryRecords records = buildRecords(new ByteBuffer[] {ByteBuffer.wrap(randomBytes(bufferSize))});
```
hachikuji: nit: can just do `buildRecords(ByteBuffer.wrap(randomBytes(bufferSize)))`

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
```diff
@@ -0,0 +1,108 @@
...
+        channel.close();
+        frozen = true;
+
+        // Set readonly and ignore the result
+        if (!tempSnapshotPath.toFile().setReadOnly()) {
+            throw new IOException(String.format("Unable to set file (%s) as read-only", tempSnapshotPath));
+        }
+
+        Path destination = Snapshots.moveRename(tempSnapshotPath, snapshotId);
+        Files.move(tempSnapshotPath, destination, StandardCopyOption.ATOMIC_MOVE);
```
hachikuji: It might be useful to review the history behind `Utils.atomicMoveWithFallback`. It's not clear to me why this case is different from some of the other situations that it is used.

jsancio: I investigated the issue a bit. It look like this utility was introduce to replace these lines in `OffsetCheckpoint.scala`:
```scala
// swap new offset checkpoint file with previous one
if(!temp.renameTo(file)) {
    // renameTo() fails on Windows if the destination file exists.
    file.delete()
    if(!temp.renameTo(file))
        throw new IOException(...)
}
```

Looking at the JDK implementation in Windows, `ATOMIC_MOVE` should work in Windows if the target file exists: https://github.com/openjdk/jdk/blob/master/src/java.base/windows/classes/sun/nio/fs/WindowsFileCopy.java#L298-L312

In other words, I think we can keep the code as is in this PR.

cc @ijuma

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
jsancio: Yes!

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
ijuma: It's not only Windows, NFS also has some restrictions. Why don't you want to use the utility method?

ijuma: Also, I would not rely on reading the code to assume one way or another. You'd want to test it too.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
```diff
@@ -472,4 +490,106 @@ private EpochStartOffset(int epoch, long startOffset) {
...
+        }
+
+        @Override
+        public long sizeInBytes() {
+            return data.sizeInBytes();
+        }
+
+        @Override
+        public Iterator<RecordBatch> iterator() {
+            return new Iterator<RecordBatch>() {
```
hachikuji: nit: we cab use `covariantCast`?

hachikuji (approved):
LGTM. I am going to push a trivial tweak to make use of `covariantCast` in `MockLog` and to use the atomic move utility. We can open a separate issue if using the utility is a mistake and discuss further, but it seems reasonable to have the initial patch follow the established pattern.

PR #9513, opened 2020-10-27
MINOR: improve `null` checks for headers

mjsax:
Headers are not allowed to be `null`, and thus the code does not check for `null` headers. However, we don't have proper guards in place and had NPE in the past if a header was `null`. This PR adds additional `null` checks to avoid that users create corrupted headers.

Cf https://issues.apache.org/jira/browse/KAFKA-8142 and https://issues.apache.org/jira/browse/KAFKA-10645

Call for review @mimaison

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/header/internals/RecordHeaders.java
+++ b/clients/src/main/java/org/apache/kafka/common/header/internals/RecordHeaders.java
@@ -23,7 +23,6 @@ import org.apache.kafka.common.utils.AbstractIterator;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


chia7712 (approved):
LGTM.

Does it need to add null checks to ```ConnectHeaders``` and ```ConsumerRecord``` ?

showuon:
LGTM. I'll close my PR then.

mjsax:
@chia7712 Good call about `ConnectHeaders`. I am not sure what we would need to change in `ConsumerRecord` though? Can you elaborate?

@showuon: what PR are you talking about?

chia7712:
> I am not sure what we would need to change in ConsumerRecord though? Can you elaborate?

this line (https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecord.java#L173). However, it may be overkill since users don't create ```ConsumerRecord```.

> what PR are you talking about?

see #9518

mjsax:
> However, it may be overkill since users don't create ConsumerRecord.

Yeah, I tend to agree. Would not hurt to add it, but should not be strictly necessary. Will merge this as-is.

PR #9515, opened 2020-10-28
KAFKA-10651: read  offsets directly from checkpoint for uninitialized tasks

ableegoldman:
Uninitialized tasks just return an empty collection in `changelogOffsets()` and are indistinguishable from genuinely stateless (or un-logged) tasks. We should just skip over these tasks and read directly from the checkpoint file when computing offset sums for a JoinGroup subscription

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
@@ -672,7 +672,8 @@ public class TaskManager {
         // just have an empty changelogOffsets map.
         for (final TaskId id : union(HashSet::new, lockedTaskDirectories, tasks.keySet())) {
             final Task task = tasks.get(id);
-            if (task != null && task.state() != State.CREATED) {
+            // Closed and uninitialized tasks don't have any offsets so we should read directly from the checkpoint
+            if (task != null && task.state() != State.CREATED && task.state() != State.CLOSED) {
                 final Map<TopicPartition, Long> changelogOffsets = task.changelogOffsets();
                 if (changelogOffsets.isEmpty()) {
                     log.debug("Skipping to encode apparently stateless (or non-logged) offset sum for task {}", id);
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
@@ -35,6 +35,7 @@ import org.apache.kafka.streams.errors.StreamsException;
 import org.apache.kafka.streams.errors.TaskIdFormatException;
 import org.apache.kafka.streams.errors.TaskMigratedException;
 import org.apache.kafka.streams.processor.TaskId;
+import org.apache.kafka.streams.processor.internals.Task.State;
 import org.apache.kafka.streams.state.internals.OffsetCheckpoint;
 import org.slf4j.Logger;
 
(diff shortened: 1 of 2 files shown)


streams/src/test/java/org/apache/kafka/streams/processor/internals/TaskManagerTest.java
```diff
@@ -2479,7 +2502,7 @@ public void shouldTransmitProducerMetrics() {
             allTasks.put(task.id(), (StateMachineTask) task);
         }
         for (final Task task : restoringTasks) {
-            assertThat(task.state(), not(Task.State.RUNNING));
```
ableegoldman, line 2482: Nothing really changed here, but we use to just leave tasks in CREATED and call them "restoring" so I had to fix this up so they really were RESTORING

vvcephei:
Thanks for this quick fix, @ableegoldman !

It looks like the bug before was that we would skip to encode these uninitialized tasks since they'd look "apparently stateless", and now we'll just try to read the checkpoint instead. If the checkpoint file is empty, though, it seems like the outcome is the same, though, right? We would not encode anything, just like a stateless task?

Although, as I'm typing this, I'm realizing the bug is probably that when the task _is_ initialized, we'd report some high offsetSum, which the assignor interprets as a low lag, and when it's _not_ initialized, then we report nothing, which the assignor interprets as a high lag (since the assignor will independently verify if it's a stateless task or not). In that case, when we legitimately have no checkpoint file, then it's ok to report nothing, because we legitimately have a high lag. And it won't flip-flop in any case, because once the task gets initialized, its lag will still be about the same.

Did I get that right?

ableegoldman:
> Although, as I'm typing this, I'm realizing the bug is probably that when the task is initialized, we'd report some high offsetSum, which the assignor interprets as a low lag, and when it's not initialized, then we report nothing, which the assignor interprets as a high lag (since the assignor will independently verify if it's a stateless task or not).

Exactly. It shouldn't report different offset sums in two adjacent rebalances if nothing changed except that it was assigned a task. One way to look at it is if we had just waited slightly longer to rejoin the group, then the task would have been initialized with the checkpoint offsets anyway, so the checkpoint file is the source of truth while the task is still in CREATED.

If the checkpoint file is empty (and the task uninitialized), then we _should_ report no offsets for that task because it doesn't have any actual state. Skipping the offset sum is technically how we handle stateless tasks, but it's also what happens for any stateful task we just don't happen to find on disk. Which is exactly what the task would be if the checkpoint is empty

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -671,7 +672,7 @@ void handleLostAll() {
         // just have an empty changelogOffsets map.
         for (final TaskId id : union(HashSet::new, lockedTaskDirectories, tasks.keySet())) {
             final Task task = tasks.get(id);
-            if (task != null) {
+            if (task != null && task.state() != State.CREATED) {
```
cadonna: Are there also other task states, we need to consider here except `CREATED`? What about `CLOSED` or `SUSPENDED`?

cadonna (commented):
@ableegoldman, Thank you for the fix!

I have one question!

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
ableegoldman: Ahh, good catch: the task returns changelog offsets ultimately from the ProcessorStateManager's `stores` map, which is cleared during close. So we should fall through to the checkpoint file for `CLOSED` tasks as well (technically there's currently no way for a CLOSED task to be in the TaskManager's `tasks` map when the member sends a JoinGroup, since we remove it from the `tasks` immediately upon `close`, but that may not always be the case so better safe than sorry.

For `SUSPENDED` we should still rely on the task's. changelog offsets since it doesn't do the checkpoint or clearing out of `stores` during suspend. (Actually very little happens in `suspend` at the moment)

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -671,7 +672,8 @@ void handleLostAll() {
         // just have an empty changelogOffsets map.
         for (final TaskId id : union(HashSet::new, lockedTaskDirectories, tasks.keySet())) {
             final Task task = tasks.get(id);
-            if (task != null) {
+            // Closed and uninitialized tasks don't have any offsets so we should read directly from the checkpoint
+            if (task != null && task.state() != State.CREATED && task.state() != State.CLOSED) {
```
cadonna, line 676: A unit test for the `CLOSED` case is missing.

cadonna (commented):
@ableegoldman Thank you for the update!

LGTM, but one unit test is missing.

Should we run the Streams system tests before we merge?

cadonna (approved):
LGTM

ableegoldman:
One test failure: `StreamTableJoinTopologyOptimizationIntegrationTest.shouldDoStreamTableJoinWithDifferentNumberOfPartitions[Optimization = all] failed`

This is pretty recently-added test  so I think it's just flaky. Created a ticket for it.

ableegoldman:
Merging to trunk

ableegoldman:
Cherry-picked to 2.7 (cc/ @bbejeck) and 2.6

PR #9516, opened 2020-10-28
MINOR: make Send and Receive work with TransferableChannel rather than Gat…

chia7712:
```BlockingChannel``` was removed by cc4dce94af8b19a796eeb7a9be78640739cb1a48 so it is time to refactor ```Send``` and ```NetworkReceive``` to remove ScatteringByteChannel and GatheringByteChannel.

### CODE REVIEW

**Comment by chia7712** [comment]
@ijuma Could you take a look? This PR is according to your comment in https://github.com/apache/kafka/blob/28c57b273a3620c18be0bba6d5f74f981d9ef47d/clients/src/main/java/org/apache/kafka/common/network/TransportLayers.java#L26

```
    // Once BlockingChannel is removed we can make Send and Receive work with TransportLayer rather than
    // GatheringByteChannel or ScatteringByteChannel.
```

File: clients/src/test/java/org/apache/kafka/common/network/TransportLayerUtils.java [hunk 1]
```diff
@@ -0,0 +1,156 @@
...
+    private TransportLayerUtils() {
+    }
+
+    /**
+     * Create a TransportLayer based on GatheringByteChannel. This helper method is used by testing only and noted that
+     * not all methods of TransportLayer are supported.
+     * @param channel channel
+     * @return a TransportLayer
+     */
+    public static TransportLayer toTransportLayer(GatheringByteChannel channel) {
```
ijuma: I had tried making this change, but it turns out we still have places where we use `GatheringByteChannel`, so not sure it's worth it.

chia7712: The benefit, personally, is that the ugly casting is removed from production code. However, you are right - ```GatheringByteChannel``` still makes us leave a wrapper in test scope...

ijuma: If we move the casting in `FileRecords` to `TransportLayers`, then it would be isolated in one class. Also, it's safe casting, not unsafe. That may be OK?

chia7712:
```java
        if (destChannel instanceof TransportLayer) {
            TransportLayer tl = (TransportLayer) destChannel;
            bytesTransferred = tl.transferFrom(channel, position, count);
        } else {
            bytesTransferred = channel.transferTo(position, count, destChannel);
        }
```
Did you mean that we could move above code to ```TransportLayers```. If so, it seems that we still don't replace ```GatheringByteChannel``` by ```TransportLayer``` in production code.

Most usage of ```GatheringByteChannel``` implementation (in test scope) is ```ByteBufferChannel```. We can rewrite ```ByteBufferChannel``` to ```ByteBufferTransportLayer``` so the ```toTransportLayer``` wrapper can be removed.
Also, the new ```ByteBufferTransportLayer ``` should be moved to test code as it is not used by production code.

chia7712: @ijuma please take a look at https://github.com/apache/kafka/pull/9516/commits/9369c7a67a66f55e1f0a93d8414d03a6e10ea410 to see what I described if you have free cycle :)

ijuma:
For some reason, I can't add a comment to the previous discussion. In my opinion TransportLayer is a fat interface and it seems like a step back to require it in so many places. Are `hasPendingWrites` and `transferFrom` the only methods we cast in order to access at the moment?

chia7712:
>  Are hasPendingWrites and transferFrom the only methods we cast in order to access at the moment?

yes

clients/src/main/java/org/apache/kafka/common/network/NetworkReceive.java
```diff
@@ -89,7 +88,8 @@ public boolean complete() {
         return !size.hasRemaining() && buffer != null && !buffer.hasRemaining();
     }
 
-    public long readFrom(ScatteringByteChannel channel) throws IOException {
+    @Override
+    public long readFrom(TransportLayer channel) throws IOException {
```
ijuma: Do we need to change this? It seems like what we had before was good enough.

ijuma:
Maybe we can introduce an interface that only adds `hasPendingWrites` and `transferFrom`. Then use that on the `Send` side. And revert the changes on the `Receive` side.

chia7712:
> Maybe we can introduce an interface that only adds hasPendingWrites and transferFrom. Then use that on the Send side. And revert the changes on the Receive side.

great idea. I have addressed it.

clients/src/main/java/org/apache/kafka/common/network/WritableChannel.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.common.network;
+
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+import java.nio.channels.GatheringByteChannel;
+
+public interface WritableChannel extends GatheringByteChannel {
```
ijuma: Maybe we should call it `TransferableChannel` since it adds a `transferFrom` method. And let's add some brief class javadoc explaining why we have this interface.

ijuma (commented):
I had a quick look and I think it works pretty well. One comment below.

clients/src/main/java/org/apache/kafka/common/network/WritableChannel.java
chia7712: will copy that

clients/src/main/java/org/apache/kafka/common/network/TransferableChannel.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package org.apache.kafka.common.network;
+
+import java.io.IOException;
+import java.nio.channels.FileChannel;
+import java.nio.channels.GatheringByteChannel;
+
+/**
+ * TransportLayer is a fat interface so it is a bit weird to use TransportLayer in write-only path.
+ * Hence, we extract all write-related methods from TransportLayer to be a new interface - TransferableChannel
```
ijuma: I would replace this comment with:

> Extends `GatheringByteChannel` with the minimal set of methods required by the `Send` interface. Supporting TLS and efficient `zero copy` transfers are the main reasons for the additional methods.

clients/src/test/java/org/apache/kafka/common/record/LazyDownConversionRecordsTest.java
```diff
@@ -160,20 +161,21 @@ private static MemoryRecords convertRecords(MemoryRecords recordsToConvert, byte
...
-
-            // cleanup
-            convertedRecords.close();
-            channel.close();
-
+            ByteBuffer convertedRecordsBuffer;
+            try (FileChannel fileChannel = FileChannel.open(outputFile.toPath(), StandardOpenOption.READ, StandardOpenOption.WRITE)) {
+                ByteBuffer buf;
+                try (org.apache.kafka.common.requests.ByteBufferChannel channel =
+                             new org.apache.kafka.common.requests.ByteBufferChannel(bytesToConvert)) {
```
ijuma: I don't think you can use this as a general class as it stands. A previous PR changed it to be non general, it seems:

```java
    @Override
    public long write(ByteBuffer[] srcs, int offset, int length) {
        int position = buf.position();
        for (int i = 0; i < length; i++) {
            ByteBuffer src = srcs[i].duplicate();
            if (i == 0) {
                src.position(src.position() + offset);
            }
            buf.put(src);
        }
        return buf.position() - position;
    }
```

ijuma: One more thing: why we are using a fully qualified name instead of relying on imports?

chia7712: IIRC, the import rule obstructs me from importing the class. Will check it later.

clients/src/main/java/org/apache/kafka/common/network/TransferableChannel.java
chia7712: will copy that.

clients/src/test/java/org/apache/kafka/common/record/LazyDownConversionRecordsTest.java
ijuma: We should fix the underlying case instead of working around it. :)

chia7712: @ijuma Thanks for your comments. I have fixed the method ```write``` and add more tests for it. Also, the disallowed package is removed.

chia7712:
> This seems less safe than what we had before. Same for the transferFrom implementation. I think we want implementations to decide the right behavior for themselves.

Fair enough. Will revert it.

ijuma (approved):
LGTM, thanks. I made a tiny adjustment to the javadoc of TransferableChannel.

chia7712:
> I made a tiny adjustment to the javadoc of TransferableChannel.

Thanks for kindly updating!

dengziming:
@chia7712 Hello, I have a question, why not also replace `ScatteringByteChannel` in `Receive` with a similar class?

chia7712:
hi @dengziming 

> I have a question, why not also replace ScatteringByteChannel in Receive with a similar class?

```TransportLayer``` is a heavy interface so it may causes trouble in testing if we replace ```ScatteringByteChannel ``` by ```TransportLayer``` for ```Receive```. By contrast, ```GatheringByteChannel``` lacks some methods from ```TransportLayer```  so we used to cast the ```GatheringByteChannel``` to ```TransportLayer ``` before. In this PR, we introduced a new interface ```TransferableChannel``` which is a subclass of ```GatheringByteChannel``` and have some methods from ```TransferableChannel```. Through ```TransferableChannel```, we can remove all casting code.

PR #9517, opened 2020-10-28
MINOR: add comments to explain why it needs to add synchronization on…

chia7712:
The reason of adding synchronization is obscure so it looks like a bug. It seems to me we should add comment to make it more readable.


clients/src/main/java/org/apache/kafka/common/security/authenticator/SaslClientAuthenticator.java
```diff
@@ -612,6 +612,9 @@ private void handleSaslHandshakeResponse(SaslHandshakeResponse response) {
      */
     public static String firstPrincipal(Subject subject) {
         Set<Principal> principals = subject.getPrincipals();
+        // getPrincipals() returns a SynchronizedSet and iteration over SynchronizedSet is not thread safe.
+        // Hence, we have to add synchronization on local variable "principals".
+        // see https://github.com/apache/kafka/pull/3208#discussion_r120130706 for discussion.
```
ijuma: Instead of referencing an older discussion, I would make sure the comment is self contained.

chia7712: fair enough. I will remove the link to older discussion. the remaining comment should be enough.

PR #9408, opened 2020-10-10
KAFKA-10598: Improve IQ name and type checks

vvcephei:
Previously, we would throw a confusing error, "the store has migrated,"
when users ask for a store that is not in the topology at all, or when the
type of the store doesn't match the QueryableStoreType parameter.

Adds an up-front check that the requested store is registered and also
a better error message when the QueryableStoreType parameter
doesn't match the store's type.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9521, opened 2020-10-28
KAFKA-10638: Fix QueryableStateIntegrationTest

vvcephei:
This test has been observed to have flaky failures.
Apparently, in the failed runs, Streams had entered a rebalance
before some of the assertions were made. We recently made
IQ a little stricter on whether it would return errors instead of
null responses in such cases:
KAFKA-10598: Improve IQ name and type checks (#9408)

As a result, we have started seeing failures now instead of
silently executing an invalid test (I.e., it was asserting the
return to be `null`, but the result was `null` for the wrong
reason).

Now, if the test discovers that Streams is no longer running,
it will repeat the verification until it actually gets a valid
positive or negative result.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/state/QueryableStoreType.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/QueryableStoreType.java
@@ -44,7 +44,7 @@ public interface QueryableStoreType<T> {
      * @param storeProvider     provides access to all the underlying StateStore instances
      * @param storeName         The name of the Store
      * @return a read-only interface over a {@code StateStore}
-     *        (cf. {@link org.apache.kafka.streams.state.QueryableStoreTypes.KeyValueStoreType})
+     *        (cf. {@link QueryableStoreTypes.KeyValueStoreType})
      */
     T create(final StateStoreProvider storeProvider,
              final String storeName);
--- a/streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
@@ -28,15 +28,16 @@ import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.StoreQueryParameters;
 import org.apache.kafka.streams.StreamsBuilder;
 import org.apache.kafka.streams.StreamsConfig;
+import org.apache.kafka.streams.errors.InvalidStateStoreException;
 import org.apache.kafka.streams.integration.utils.EmbeddedKafkaCluster;
 import org.apache.kafka.streams.integration.utils.IntegrationTestUtils;
 import org.apache.kafka.streams.kstream.Consumed;
 import org.apache.kafka.streams.kstream.Materialized;
 import org.apache.kafka.streams.state.KeyValueStore;
 import org.apache.kafka.streams.state.QueryableStoreType;
-import org.apache.kafka.streams.state.QueryableStoreTypes;
 import org.apache.kafka.streams.state.ReadOnlyKeyValueStore;
 import org.apache.kafka.test.IntegrationTest;
+import org.apache.kafka.test.TestCondition;
 import org.apache.kafka.test.TestUtils;
 import org.junit.After;
 import org.junit.Before;
@@ -44,6 +45,8 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.junit.rules.TestName;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.time.Duration;
 import java.util.ArrayList;
@@ -55,18 +58,24 @@ import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import static org.apache.kafka.streams.integration.utils.IntegrationTestUtils.getStore;
 import static org.apache.kafka.streams.integration.utils.IntegrationTestUtils.safeUniqueTestName;
 import static org.apache.kafka.streams.integration.utils.IntegrationTestUtils.startApplicationAndWaitUntilRunning;
+import static org.apache.kafka.streams.state.QueryableStoreTypes.keyValueStore;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
 
 @Category({IntegrationTest.class})
 public class StoreQueryIntegrationTest {
+
+    private static final Logger LOG = LoggerFactory.getLogger(StoreQueryIntegrationTest.class);
 
     private static final int NUM_BROKERS = 1;
     private static int port = 0;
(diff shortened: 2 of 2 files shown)


streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
```diff
@@ -102,10 +111,10 @@ public void shouldQueryOnlyActivePartitionStoresByDefault() throws Exception {
...
         final StreamsBuilder builder = new StreamsBuilder();
         builder.table(INPUT_TOPIC_NAME, Consumed.with(Serdes.Integer(), Serdes.Integer()),
-                        Materialized.<Integer, Integer, KeyValueStore<Bytes, byte[]>>as(TABLE_NAME)
-                                .withCachingDisabled())
-                .toStream()
-                .peek((k, v) -> semaphore.release());
+                      Materialized.<Integer, Integer, KeyValueStore<Bytes, byte[]>>as(TABLE_NAME)
+                          .withCachingDisabled())
+               .toStream()
+               .peek((k, v) -> semaphore.release());
```
vvcephei, line 117: I went ahead and fixed the whitespace also, since this PR is relatively small.

streams/src/main/java/org/apache/kafka/streams/state/QueryableStoreType.java
```diff
@@ -44,7 +44,7 @@
      * @param storeProvider     provides access to all the underlying StateStore instances
      * @param storeName         The name of the Store
      * @return a read-only interface over a {@code StateStore}
-     *        (cf. {@link org.apache.kafka.streams.state.QueryableStoreTypes.KeyValueStoreType})
+     *        (cf. {@link QueryableStoreTypes.KeyValueStoreType})
```
vvcephei, line 47: This class is in the same package, so the fully-qualified name is not necessary.

streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
```diff
@@ -117,17 +126,34 @@ public void shouldQueryOnlyActivePartitionStoresByDefault() throws Exception {
...
+                    assertThat(store1.get(key), is(nullValue()));
+                }
+                return true;
+            } catch (final InvalidStateStoreException exception) {
+                assertThat(
+                    exception.getMessage(),
+                    containsString("Cannot get state store source-table because the stream thread is PARTITIONS_ASSIGNED, not RUNNING")
+                );
+                LOG.info("Streams wasn't running. Will try again.");
+                return false;
```
vvcephei, line 154: This is the meat of this change. If we do get an exception, we can still verify the exception is the one we expected to get, and then we return `false` to indicate we should try again later to get a successful verification.

streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
```diff
@@ -153,51 +179,75 @@ public void shouldQuerySpecificActivePartitionStores() throws Exception {
...
+                    );
+                }
+                return true;
+            } catch (final InvalidStateStoreException exception) {
+                assertThat(
+                    exception.getMessage(),
+                    containsString("Cannot get state store source-table because the stream thread is PARTITIONS_ASSIGNED, not RUNNING")
+                );
+                LOG.info("Streams wasn't running. Will try again.");
+                return false;
```
vvcephei, line 248: Also, here, if we find that Streams is rebalancing, we'll try the whole verification again, including to re-discover the stores in case the stores have swapped ownership.

streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
```diff
@@ -337,34 +385,49 @@ public void shouldQuerySpecificStalePartitionStoresMultiStreamThreads() throws E
...
+            .withPartition(keyDontBelongPartition);
+        final ReadOnlyKeyValueStore<Integer, Integer> store3 = getStore(kafkaStreams1, otherParam);
+        final ReadOnlyKeyValueStore<Integer, Integer> store4 = getStore(kafkaStreams2, otherParam);
 
         // Assert that
         assertThat(store3.get(key), is(nullValue()));
         assertThat(store4.get(key), is(nullValue()));
     }
 
+    private static void until(final TestCondition condition) {
```
vvcephei, line 416: Note, this is different than `TestUtils.waitForCondition`, which does the inverse thing. That one will retry on exceptions and otherwise verify that the return is `true`. We need to fail on exceptions and retry as long as the return is `false`.

I opted to keep this method here, since it might be confusing next to the other util method.

vvcephei (commented):
@guozhangwang , @dongjinleekr , @chia7712 ,

Since you all were interested in https://github.com/apache/kafka/pull/9426, I'm wondering if you'd like to review this one as well?

Thanks,
-John

streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
```diff
@@ -117,17 +126,34 @@ public void shouldQueryOnlyActivePartitionStoresByDefault() throws Exception {
...
+            // Assert that only active is able to query for a key by default
+            assertThat(kafkaStreams1IsActive ? store1.get(key) : store2.get(key), is(notNullValue()));
+            try {
+                if (kafkaStreams1IsActive) {
+                    assertThat(store2.get(key), is(nullValue()));
+                } else {
+                    assertThat(store1.get(key), is(nullValue()));
+                }
+                return true;
+            } catch (final InvalidStateStoreException exception) {
```
chia7712, line 148: Why not handling the ```InvalidStateStoreException``` in the helper method ```until```

streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
```diff
@@ -153,51 +179,75 @@ public void shouldQuerySpecificActivePartitionStores() throws Exception {
...
+                    assertThat(getStore(kafkaStreams1, storeQueryParam2).get(key), is(nullValue()));
+                    final InvalidStateStoreException exception =
+                        assertThrows(InvalidStateStoreException.class, () -> getStore(kafkaStreams2, storeQueryParam2).get(key));
+                    assertThat(
+                        exception.getMessage(),
+                        containsString("The specified partition 1 for store source-table does not exist.")
+                    );
+                }
+                return true;
+            } catch (final InvalidStateStoreException exception) {
```
chia7712, line 242: which method can throw ```InvalidStateStoreException``` in this case? It seems to me the potential methods are caught by ```assertThrows```

chia7712 (commented):
@vvcephei Thanks for your patch. two little questions are left.

streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
vvcephei: I wanted to keep the concerns separate, so that unexpected exceptions would cause the test to fail fast. The idea is that `until` is the inverse of `while`, namely, it just loops as long as the condition evaluates to `false`. If the condition throws an exception, then the loop also throws, just like the real `while` loop.

streams/src/test/java/org/apache/kafka/streams/integration/StoreQueryIntegrationTest.java
vvcephei: It's a little subtle, but each block of the `if` has two verifications: that it can fetch data from one of the instances and that it gets an exception from the other instance. If something goes wrong the "can fetch data" verification could throw.

E.g., `assertThat(getStore(kafkaStreams1, storeQueryParam2).get(key), is(nullValue()));` could throw.

vvcephei:
The tests passed, and I answered the two review questions. I'll go ahead and merge.

PR #9526, opened 2020-10-29
KAFKA-10525: Emit JSONs with new auto-generated schema

anatasiavela:
Kafka’s request and response traces currently output in a format that is JSON-like and are not easily parsable. There is a new auto-generated schema for each request type that supports outputting JSON payloads for request and response payloads. These can be adapted to provide structured request tracing.

Includes tests that iterate through all the request types and ensure we handle all of them in RequestConvertToJson.

[KIP-673](https://cwiki.apache.org/confluence/display/KAFKA/KIP-673%3A+Emit+JSONs+with+new+auto-generated+schema)


clients/src/main/java/org/apache/kafka/common/requests/AddOffsetsToTxnRequest.java
```diff
@@ -57,6 +57,10 @@ public AddOffsetsToTxnRequest(Struct struct, short version) {
         this.data = new AddOffsetsToTxnRequestData(struct, version);
     }
 
+    public AddOffsetsToTxnRequestData data() {
+        return data;
+    }
```
dajac: This is not entirely related to your PR but it seems that we are little inconsistent wrt. how we expose/access `data` in the code base. As an example, here we add an accessor to access `data` but we keep `data` field public and keep accessing it directly from various places.

I think that we should do one the following:
1) Add an accessor and make `data` final private everywhere; or
2) Make `data` final public everywhere.

It seems that you suggest to go with 1) in the PR. Personally, I don't have a strong preference for any of them but I would like us to be consistent across the board.

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -162,7 +161,7 @@ object RequestChannel extends Logging {
       }
     }
 
-    trace(s"Processor $processor received request: ${requestDesc(true)}")
+    trace(s"Processor $processor received request: ${RequestConvertToJson.requestDesc(header, loggableRequest, true).toString}")
```
dajac: We may want to gate this by `isTraceEnabled` in order to not compute it necessarily.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+      case req: UpdateFeaturesRequest => UpdateFeaturesRequestDataJsonConverter.write(req.data(), request.version())
+      case req: UpdateMetadataRequest => UpdateMetadataRequestDataJsonConverter.write(req.data(), request.version())
+      case req: VoteRequest => VoteRequestDataJsonConverter.write(req.data, request.version())
+      case req: WriteTxnMarkersRequest => WriteTxnMarkersRequestDataJsonConverter.write(req.data(), request.version())
+    }
+  }
+
+  def response(response: AbstractResponse, version: Short): JsonNode = {
+    response match {
+      case res: AddOffsetsToTxnResponse => AddOffsetsToTxnResponseDataJsonConverter.write(res.data, version)
```
dajac, line 96: This is one case where we still access `data` without an accessor.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+import org.apache.kafka.common.record.RecordBatch
+import org.apache.kafka.common.requests._
+import org.apache.kafka.common.utils.CollectionUtils
+
+import scala.jdk.CollectionConverters._
+
+object RequestConvertToJson {
+  def request(request: AbstractRequest, verbose: Boolean): JsonNode = {
+    request match {
+      case req: AddOffsetsToTxnRequest => AddOffsetsToTxnRequestDataJsonConverter.write(req.data(), request.version())
```
dajac: It is a little frustrating to have to maintain this mapping for each request. I wonder if we could auto-generate such helper based on the protocol description. We already generate `DataJsonConverter` helpers so it should be feasible to also generate an overall helper.

Have you already considered this? I am fine with keep it as-is for now though and to investigate this in the future.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+      case req: SaslAuthenticateRequest => SaslAuthenticateRequestDataJsonConverter.write(req.data(), request.version())
+      case req: SaslHandshakeRequest => SaslHandshakeRequestDataJsonConverter.write(req.data(), request.version())
+      case req: StopReplicaRequest => StopReplicaRequestDataJsonConverter.write(req.data(), request.version())
+      case req: SyncGroupRequest => SyncGroupRequestDataJsonConverter.write(req.data(), request.version())
+      case req: TxnOffsetCommitRequest => TxnOffsetCommitRequestDataJsonConverter.write(req.data(), request.version())
+      case req: UpdateFeaturesRequest => UpdateFeaturesRequestDataJsonConverter.write(req.data(), request.version())
+      case req: UpdateMetadataRequest => UpdateMetadataRequestDataJsonConverter.write(req.data(), request.version())
+      case req: VoteRequest => VoteRequestDataJsonConverter.write(req.data, request.version())
+      case req: WriteTxnMarkersRequest => WriteTxnMarkersRequestDataJsonConverter.write(req.data(), request.version())
+    }
```
dajac, line 91: Should we handle the unknown cases explicitly and throw an exception with an meaningful message instead of relying on `MatchError`? What do you think?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+      case res: UpdateFeaturesResponse => UpdateFeaturesResponseDataJsonConverter.write(res.data(), version)
+      case res: UpdateMetadataResponse => UpdateMetadataResponseDataJsonConverter.write(res.data(), version)
+      case res: WriteTxnMarkersResponse => WriteTxnMarkersResponseDataJsonConverter.write(res.data, version)
+      case res: VoteResponse => VoteResponseDataJsonConverter.write(res.data, version)
+    }
+  }
+
+  def requestHeaderNode(header: RequestHeader): JsonNode = {
+    val node = RequestHeaderDataJsonConverter.write(header.data(), header.headerVersion()).asInstanceOf[ObjectNode]
+    node.set("requestApiKey", new TextNode(header.apiKey.toString))
```
dajac: nit: Could we add a comment here to explain why we overwrite the `requestApiKey`?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+    node
+  }
+
+  def requestDescMetrics(header: RequestHeader, res: Response, req: AbstractRequest,
+                  context: RequestContext, session: Session, verbose: Boolean,
+                  totalTimeMs: Double, requestQueueTimeMs: Double, apiLocalTimeMs: Double,
+                  apiRemoteTimeMs: Double, apiThrottleTimeMs: Long, responseQueueTimeMs: Double,
+                  responseSendTimeMs: Double, temporaryMemoryBytes: Long,
+                  messageConversionsTimeMs: Double): JsonNode = {
```
dajac: nit: Could we align them with `header`?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+    node.set("requestQueueTime", new DoubleNode(requestQueueTimeMs))
+    node.set("localTime", new DoubleNode(apiLocalTimeMs))
+    node.set("remoteTime", new DoubleNode(apiRemoteTimeMs))
+    node.set("throttleTime", new LongNode(apiThrottleTimeMs))
+    node.set("responseQueueTime", new DoubleNode(responseQueueTimeMs))
+    node.set("sendTime", new DoubleNode(responseSendTimeMs))
+    node.set("securityProtocol", new TextNode(context.securityProtocol.toString))
+    node.set("principal", new TextNode(session.principal.toString))
+    node.set("listener", new TextNode(context.listenerName.value))
+    node.set("clientInformation", new TextNode(context.clientInformation.toString))
```
dajac: I wonder if we should also structure the client information in JSON instead of relying on the string representation of the object. That would make the information more accessible.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+    val node = new ObjectNode(JsonNodeFactory.instance)
+    node.set("requestHeader", requestHeaderNode(header))
+    node.set("request", request(req, verbose))
+    node
+  }
+
+  /**
+   * Temporary until switch to use the generated schemas.
+   */
+  def offsetsForLeaderEpochRequestNode(request: OffsetsForLeaderEpochRequest, version: Short): JsonNode = {
```
dajac: Should we add unit tests for these helper methods? I mean that it would be great to ensure that the format of the JSON is equivalent to the JSON what will be generated by the auto-generated protocol.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+
+  def requestDesc(header: RequestHeader, req: AbstractRequest, verbose: Boolean): JsonNode = {
+    val node = new ObjectNode(JsonNodeFactory.instance)
+    node.set("requestHeader", requestHeaderNode(header))
+    node.set("request", request(req, verbose))
+    node
+  }
+
+  /**
+   * Temporary until switch to use the generated schemas.
```
dajac: I am working on this one :)

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+        node.set("transactionalId", NullNode.instance)
+      } else {
+        node.set("transactionalId", new TextNode(request.transactionalId))
+      }
+    }
+    node.set("acks", new ShortNode(request.acks))
+    node.set("timeoutMs", new IntNode(request.timeout))
+    if (verbose) {
+      val partSizes = new ArrayNode(JsonNodeFactory.instance)
+      for (partSize <- request.partitionSizes().entrySet.asScala) {
```
dajac: nit: We may be able to use `request.partitionSizes().forEach` directly.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+        val part = new ObjectNode(JsonNodeFactory.instance)
+        val topic = partSize.getKey
+        part.set("partition", new TextNode(topic.toString))
```
dajac: We are using the `topic` for the `partition` here. I suppose that we need both, isn't it?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+        part.set("size", new IntNode(partSize.getValue))
+        partSizes.add(part)
+      }
+      node.set("partitionSizes", partSizes)
+    } else {
+      node.set("partitionSizes", new IntNode(request.partitionSizes.size))
```
dajac: It is a bit weird to have `partitionSizes` used with two different type of value. It seems that we were previously using `numPartitions` for this case.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+      }
+      topicsData.set("partitions", partitionsArray)
+      topicsArray.add(topicsData)
+    }
+    node.set("topics", topicsArray)
+    node
+  }
+
+  /**
+   * Temporary until switch to use the generated schemas.
```
dajac: It is important to note that the format of the JSON will change when we move to the generated schema. `partitionSizes` does not exist in the schema. Thus, I wonder if we should better align directly; or keep this exception forever. Have we thought about this?

Similarly in the `FetchResponse`, we were outputting `recordsSizeInBytes` per partition and it seems that now we put an empty `recordSet` field. cc @lbradstreet

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,188 @@
...
+  @Test
+  def testAllRequestTypesHandled(): Unit = {
+    val unhandledKeys = ArrayBuffer[String]()
+    ApiKeys.values().foreach(key => {
+      val version: Short = 0
+      val req = createRequestsFromApiKey(key, version)
+      try {
+        RequestConvertToJson.request(req, false)
+      } catch {
+        case _ : Throwable => unhandledKeys += key.toString
```
dajac: Shouldn't we only catch `AssertionError` here and below?

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,188 @@
...
+    case ApiKeys.PRODUCE => ProduceRequest.Builder.forCurrentMagic(0.toShort, 10000, new HashMap[TopicPartition, MemoryRecords]()).build()
+    case ApiKeys.FETCH => new FetchRequest(new FetchRequestData(), version)
+    case ApiKeys.LIST_OFFSETS => new ListOffsetRequest(new ListOffsetRequestData().toStruct(version), version)
+    case ApiKeys.METADATA => new MetadataRequest(new MetadataRequestData(), version)
+    case ApiKeys.OFFSET_COMMIT => new OffsetCommitRequest(new OffsetCommitRequestData(), version)
+    case ApiKeys.OFFSET_FETCH => new OffsetFetchRequest(new OffsetFetchRequestData().toStruct(version), version)
+    case ApiKeys.FIND_COORDINATOR => new FindCoordinatorRequest(new FindCoordinatorRequestData().toStruct(version), version)
+    case ApiKeys.JOIN_GROUP => new JoinGroupRequest(new JoinGroupRequestData(), version)
+    case ApiKeys.HEARTBEAT => new HeartbeatRequest(new HeartbeatRequestData().toStruct(version), version)
+    case ApiKeys.LEAVE_GROUP => new LeaveGroupRequest(new LeaveGroupRequestData().toStruct(version), version)
```
dajac: Why are we using `toStruct` in some cases and passing the `*Data` object directly in others?

dajac (commented):
@anatasiavela Thanks for the PR. Overall, the PR looks good. I have left some comments/questions.

clients/src/main/java/org/apache/kafka/common/requests/AddOffsetsToTxnRequest.java
anatasiavela: Yea I definitely agree with wanting consistency. I'll stick to 1) in the PR.

core/src/main/scala/kafka/network/RequestChannel.scala
lbradstreet: Would we unnecessarily calculate it here? I think the macro will avoid it in this case, but I'm fine with gating it anyway.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
lbradstreet: Having thought about it a bit more, maybe we could include both the api key number and the api key name rather than overwriting it?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
lbradstreet: Yes, I agree, this would be helpful if it's not too difficult.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
lbradstreet: I think that is fair, it would be good to align now. With the FetchResponse I would like to circle back on it and provide more detail. If not the recordSizeInBytes in the FetchResponse I think we need to include a top level responseSizeBytes in the trace.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,343 @@
...
+                  totalTimeMs: Double, requestQueueTimeMs: Double, apiLocalTimeMs: Double,
+                  apiRemoteTimeMs: Double, apiThrottleTimeMs: Long, responseQueueTimeMs: Double,
+                  responseSendTimeMs: Double, temporaryMemoryBytes: Long,
+                  messageConversionsTimeMs: Double): JsonNode = {
+    val node = new ObjectNode(JsonNodeFactory.instance)
+    node.set("requestHeader", requestHeaderNode(header))
+    node.set("request", request(req, verbose))
+    node.set("response", res.responseLog.getOrElse(new TextNode("")))
+    node.set("connection", new TextNode(context.connectionId))
+    node.set("totalTime", new DoubleNode(totalTimeMs))
```
lbradstreet: I was thinking a bit more about the time fields, and I think we should add `Ms` to the end of all of them. What do you think?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: So I'll keep `requestApiKey` as it originally was and just add a new one `requestApiKeyName`.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: There was an issue before where a request type wasn't handled and it threw a `MatchError`. I added the system test to make sure all the request types were handled, but in case this happens again, I'll add the exception for unknown cases.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: Yay! Excited for this to make even more consistent with other request types.

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
anatasiavela: I guess this is related with the inconsistency issue, but I tried not to make too many changes. The constructors with the `data` parameters were private in some cases, so I used the `toStruct`. But for consistency, I'll change to use `toStruct` instead.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: I'm for it, that would be more descriptive.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
lbradstreet: How about requestApi or requestApiName?

core/src/main/scala/kafka/network/RequestChannel.scala
anatasiavela: The macro does check `isTraceEnabled`, and this check is done before the string is evaluated, so it wouldn't be unnecessarily computed. But I'm also fine with gating it.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: Good point, I'll add it.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: Changed to match the auto-generated schema.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: Changed to match the auto-generated schema.

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
anatasiavela: Actually, I found a workaround for this mapping and removed it all together, so there'd be one less apikey mapping to maintain.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: I've change it back to reflect what the auto-generated schema produces. The reason we didn't have it match the auto-generated schema was because it copies every byte of the record into the `records` field which could be expensive. If this isn't really a problem, I'll leave it as this.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -131,7 +131,7 @@ class KafkaApis(val requestChannel: RequestChannel,
    */
   override def handle(request: RequestChannel.Request): Unit = {
     try {
-      trace(s"Handling request:${request.requestDesc(true)} from connection ${request.context.connectionId};" +
+      trace(s"Handling request:${RequestConvertToJson.requestDesc(request.header, request.loggableRequest, true).toString} from connection ${request.context.connectionId};" +
```
splett2: nit: unnecessary call of `toString`.

core/src/main/scala/kafka/network/RequestChannel.scala
splett2: nit: unnecessary call of toString

`isTraceEnabled` shouldn't be necessary, if my understanding of the logging macro is correct.

https://github.com/lightbend/scala-logging#scala-logging-

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: I haven't thought about this, but it's a good thing to consider. It would definitely help with the hassle of having to maintain so many different request mappers. Although since it's working as-is, let's keep it and have a separate look into it later.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
dajac: Actually, it is probably better to not serialize the records in our case. Records could be large and bytes are not that useful. I suppose that having the number of bytes as a field like we had before is likely way more valuable. Should we bring back the amount of bytes per partition for both the Fetch response and the Produce request? @lbradstreet What do you think?

core/src/main/scala/kafka/network/RequestChannel.scala
dajac: Indeed, that is not necessary with the macro. My mistake.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,360 @@
...
+      case res: StopReplicaResponse => StopReplicaResponseDataJsonConverter.write(res.data, version)
+      case res: SyncGroupResponse => SyncGroupResponseDataJsonConverter.write(res.data, version)
+      case res: TxnOffsetCommitResponse => TxnOffsetCommitResponseDataJsonConverter.write(res.data, version)
+      case res: UpdateFeaturesResponse => UpdateFeaturesResponseDataJsonConverter.write(res.data, version)
+      case res: UpdateMetadataResponse => UpdateMetadataResponseDataJsonConverter.write(res.data, version)
+      case res: WriteTxnMarkersResponse => WriteTxnMarkersResponseDataJsonConverter.write(res.data, version)
+      case res: VoteResponse => VoteResponseDataJsonConverter.write(res.data, version)
+      case _ => throw new AssertionError(String.format("ApiKey %s is not currently handled in `response`, the " +
+        "code should be updated to do so.", response));
+    }
```
splett2: nit: we can use scala interpolated string instead of `String.format`, unnecessary semicolon.
also, our error message mentions `ApiKey`, but the templated argument is `response`.

Ditto for request.

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,140 @@
...
+        val struct = ApiMessageType.fromApiKey(key.id).newRequest().toStruct(version)
+        req = AbstractRequest.parseRequest(key, version, struct)
+      }
+      try {
+        RequestConvertToJson.request(req, false)
+      } catch {
+        case _ : AssertionError => unhandledKeys += key.toString
+      }
+    })
+    assert(unhandledKeys.isEmpty, String.format("%s request keys not handled in RequestConvertToJson", unhandledKeys))
```
splett2:
```suggestion
    assertEquals("Unhandled request keys", ArrayBuffer.empty, unhandledKeys))
```
nit: we should prefer to use junit assertions. 
ditto for `testAllResponseTypesHandled`.
These cases are not too bad because the assertion has a failure message.

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,140 @@
...
+    partitionDataMap.put(new TopicPartition("topic1", 0), new PartitionData(Optional.of(0),  1))
+
+    val version: Short = 3
+    val request = OffsetsForLeaderEpochRequest.Builder.forConsumer(partitionDataMap).build(version)
+    val manualGenNode = RequestConvertToJson.request(request, false)
+
+    val requestData = OffsetForLeaderEpochRequestDataJsonConverter.read(manualGenNode, version)
+    val autoGenNode = OffsetForLeaderEpochRequestDataJsonConverter.write(requestData, version)
+
+    assert(manualGenNode.equals(autoGenNode))
```
splett2:
```suggestion
    assertEquals(autoGenNode, manualGenNode)
```
using `assertEquals` instead of assert(`boolean`) makes this test easier to debug, since if it fails we get a helpful print out of the objects. ditto for other `testFormat` tests.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
lbradstreet: I agree that we shouldn't serialize the records. If it's easy to change the recordSet autogenerated JSON output I think it would be great to retain the sizes. This is something that I wanted to follow up with in a later PR but if we can do it now I think it would be very helpful.

anatasiavela: I changed the `recordSet` autogenerated JSON output to be the number of bytes, instead of an empty byte array.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,360 @@
...
+      case req: SaslAuthenticateRequest => SaslAuthenticateRequestDataJsonConverter.write(req.data, request.version)
+      case req: SaslHandshakeRequest => SaslHandshakeRequestDataJsonConverter.write(req.data, request.version)
+      case req: StopReplicaRequest => StopReplicaRequestDataJsonConverter.write(req.data, request.version)
+      case req: SyncGroupRequest => SyncGroupRequestDataJsonConverter.write(req.data, request.version)
+      case req: TxnOffsetCommitRequest => TxnOffsetCommitRequestDataJsonConverter.write(req.data, request.version)
+      case req: UpdateFeaturesRequest => UpdateFeaturesRequestDataJsonConverter.write(req.data, request.version)
+      case req: UpdateMetadataRequest => UpdateMetadataRequestDataJsonConverter.write(req.data, request.version)
+      case req: VoteRequest => VoteRequestDataJsonConverter.write(req.data, request.version)
+      case req: WriteTxnMarkersRequest => WriteTxnMarkersRequestDataJsonConverter.write(req.data, request.version)
+      case _ => throw new AssertionError(s"ApiKey ${request.api} is not currently handled in `request`, the " +
```
dajac: `IllegaleStateException` sounds more appropriate here and below.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,360 @@
...
+import org.apache.kafka.common.network.ClientInformation
+import org.apache.kafka.common.protocol.Errors
+import org.apache.kafka.common.record.RecordBatch
+import org.apache.kafka.common.requests._
+import org.apache.kafka.common.utils.CollectionUtils
+
+import scala.jdk.CollectionConverters._
+
+object RequestConvertToJson {
+  def request(request: AbstractRequest, verbose: Boolean): JsonNode = {
```
dajac: It seems that `verbose` is not used anymore. I think that we were dumping the full Produce request and Fetch response when `verbose` was `true`. Should we just remove it?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,360 @@
...
+  def requestHeaderNode(header: RequestHeader): JsonNode = {
+    val node = RequestHeaderDataJsonConverter.write(header.data(), header.headerVersion()).asInstanceOf[ObjectNode]
+    node.set("requestApiKeyName", new TextNode(header.apiKey.toString))
+    node
+  }
+
+  def clientInfoNode(clientInfo: ClientInformation): JsonNode = {
+    val node = new ObjectNode(JsonNodeFactory.instance)
+    node.set("softwareName", new TextNode(clientInfo.softwareName()))
+    node.set("softwareName", new TextNode(clientInfo.softwareVersion()))
```
dajac: `softwareName` -> `softwareVersion`?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,360 @@
...
+    }
+    node.set("acks", new ShortNode(request.acks))
+    node.set("timeoutMs", new IntNode(request.timeout))
+    val topics = CollectionUtils.groupPartitionDataByTopic(request.partitionRecordsOrFail())
+    val topicsArray = new ArrayNode(JsonNodeFactory.instance)
+    topics.forEach { (topicName, partitions) =>
+      val topicNode = new ObjectNode(JsonNodeFactory.instance)
+      topicNode.set("name", new TextNode(topicName))
+      val partitionsArray = new ArrayNode(JsonNodeFactory.instance)
+      partitions.forEach { (partitionIndex, partitionData)=>
```
dajac: nit: Missing space before `=>`.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,360 @@
...
+        val partitionNode = new ObjectNode(JsonNodeFactory.instance)
+        partitionNode.set("partitionIndex", new IntNode(partitionIndex))
+        if (partitionData == null)
+          partitionNode.set("records", NullNode.instance)
+        else
+          partitionNode.set("records", new BinaryNode(util.Arrays.copyOf(partitionData.buffer().array(), partitionData.validBytes())))
```
dajac: This is not consistent with the changes that have been made in the automated protocol.

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
```diff
@@ -380,8 +380,9 @@ private void generateVariableLengthTargetToJson(Target target, Versions versions
                         target.sourceVariable(), target.sourceVariable())));
             }
         } else if (target.field().type().isRecords()) {
-            headerGenerator.addImport(MessageGenerator.BINARY_NODE_CLASS);
-            buffer.printf("%s;%n", target.assignmentStatement("new BinaryNode(new byte[]{})"));
+            headerGenerator.addImport(MessageGenerator.INT_NODE_CLASS);
+            buffer.printf("%s;%n", target.assignmentStatement(
+                    String.format("new IntNode(%s.sizeInBytes())", target.sourceVariable())));
```
dajac: I am not entirely convinced by this change. The aim of the auto-generated protocol is to serialize/deserialise the requests/responses. Thus, it is a little weird that we only retain the size when serializing records to JSON and we actually expect bytes when deserializing from JSON. We are not consistent.

I can think of the following alternatives:
1) We keep exceptions for both the produce request and the fetch response in `RequestConvertToJson` where we could add a `recordsSizeInBytes` field instead of the `records` field.
2) We could post-process the JSON tree where we could discard `records` and add `recordsSizeInBytes`.
3) We could extend the autogenerated protocol to do only retain the size based on a flag. I am thinking about a `verbose` flag like we had before.

I lean towards 3) at the moment. By default, we should keep the current behavior and only retains the size when `verbose` is `false`. We may find a better name than `verbose` though.

@anatasiavela @lbradstreet Would this make sense?

dajac (commented):
@anatasiavela Thanks for the update. I have left few more comments.

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
lbradstreet: @dajac if I understand correctly, the current generated JSON does not print the recordSet either. Is it breaking anything to return the size rather than an empty array?
```
            }
            if (_object.recordSet == null) {
                _node.set("recordSet", NullNode.instance);
            } else {
                _node.set("recordSet", new BinaryNode(new byte[]{}));
            }
```

That said, implementing a verbose mode (3) where we print out a byte array seems reasonable if we don't use it in the trace logging path by default?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: Yes, I kept it because `requestDescMetrics` originally had the `detailsEnabled`, but I'll remove it since none of the JsonConverters differentiates between verbose and non-verbose.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: The `ProduceRequest` does not match the conditions for the change made in the automated protocol. The `records` field is of type `bytes` in `ProduceRequest`, whereas the change to the protocol only effects fields of type `records` like in `FetchResponse`.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: Just read through the comment below, so I'll keep the `verbose` flag for now, but I'll remove it depending on what we decide below.

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
anatasiavela: @lbradstreet The current generated JSON does not print the recordSet either. But when we are serializing a BinaryNode with an empty array, it is still being deserialized as a BinaryNode, so it doesn't break anything.
```
            buffer.printf("%s;%n", target.assignmentStatement(
                String.format("MemoryRecords.readableRecords(ByteBuffer.wrap(MessageUtil.jsonNodeToBinary(%s, \"%s\")))",
                    target.sourceVariable(), target.humanReadableName())));
```
So I believe the concern is that I had proposed to change the serialization to an IntNode, but the deserialization still expects a BinaryNode which I should have also changed.
Another alternative I could fix this by changing the deserialization to expect an IntNode, but I'm for adding a verbose flag

anatasiavela: Nvm, we can't deserialize `recordSet` to an IntNode because `FetchResponseData` expects `recordSet` to be of type `BaseRecords`, not `int`.
What I've done was left deserialization the same and added the verbose tag to serialization.

dajac: @lbradstreet No, it does not break anything but it is inconsistent. If we put our case aside, the purpose of the JSON converters is to convert to/from JSON so it is fair to expect that we have to maintain this contract by default. Changing it to return the size would break this contract.

Regarding `_node.set("recordSet", new BinaryNode(new byte[]{}));`, I do believe that this is a bug actually...

> That said, implementing a verbose mode (3) where we print out a byte array seems reasonable if we don't use it in the trace logging path by default?

Agreed. When logging requests/responses, we should stick to the non-verbose mode.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
```diff
@@ -346,6 +346,10 @@ public ProduceResponse getErrorResponse(int throttleTimeMs, Throwable e) {
         return partitionSizes.keySet();
     }
 
+    public Map<TopicPartition, Integer> partitionSizes() {
+        return partitionSizes;
+    }
```
dajac: Is this one still used? It seems that it is not the case.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  def clientInfoNode(clientInfo: ClientInformation): JsonNode = {
+    val node = new ObjectNode(JsonNodeFactory.instance)
+    node.set("softwareName", new TextNode(clientInfo.softwareName()))
+    node.set("softwareVersion", new TextNode(clientInfo.softwareVersion()))
+    node
+  }
+
+  def requestDescMetrics(header: RequestHeader, res: Response, req: AbstractRequest,
```
dajac: Could we add a unit test for this method?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
dajac: Right. I think that we will change it to `records` type when we will migrate the produce request. Anyway, we should not serialize the bytes to JSON here but rather put the size. We should retain the size of the records set for both the produce request and the fetch response.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
dajac: I think that we can remove this `verbose` flag here and only set it to false for both the produce request and the fetch response. I don't think that we will ever want to print out the bytes in the request log. Previously, we were printing out different things based on the flag but we never printed out the bytes: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java#L314.

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,141 @@
...
+        // a temporary workaround until ProduceRequest starts using ProduceRequestData
+        req = ProduceRequest.Builder.forCurrentMagic(0.toShort, 10000, new util.HashMap[TopicPartition, MemoryRecords]()).build()
+      } else {
+        val struct = ApiMessageType.fromApiKey(key.id).newRequest().toStruct(version)
+        req = AbstractRequest.parseRequest(key, version, struct)
+      }
+      try {
+        RequestConvertToJson.request(req, false)
+      } catch {
+        case _ : AssertionError => unhandledKeys += key.toString
```
dajac: `IllegalStateException`?

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,141 @@
...
+  def testAllResponseTypesHandled(): Unit = {
+    val unhandledKeys = ArrayBuffer[String]()
+    ApiKeys.values().foreach(key => {
+      val version: Short = 0
+      val struct = ApiMessageType.fromApiKey(key.id).newResponse().toStruct(version)
+      val res = AbstractResponse.parseResponse(key, struct, version)
+      try {
+        RequestConvertToJson.response(res, version, false)
+      } catch {
+        case _ : AssertionError => unhandledKeys += key.toString
```
dajac: `IllegalStateException`?

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
```diff
@@ -227,9 +227,13 @@ private void generateVariableLengthTargetFromJson(Target target, Versions curVer
             headerGenerator.addImport(MessageGenerator.MESSAGE_UTIL_CLASS);
             headerGenerator.addImport(MessageGenerator.BYTE_BUFFER_CLASS);
             headerGenerator.addImport(MessageGenerator.MEMORY_RECORDS_CLASS);
+            buffer.printf("if (_verbose) {%n");
+            buffer.incrementIndent();
```
dajac: I think that we can safely assume that when a request/response is serialized with `verbose` equals to `false`, we are not going to deserialize it. Therefore, I suggest to drop the handling of `verbose` on the read path.

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
```diff
@@ -380,8 +384,18 @@ private void generateVariableLengthTargetToJson(Target target, Versions versions
                         target.sourceVariable(), target.sourceVariable())));
             }
         } else if (target.field().type().isRecords()) {
+            headerGenerator.addImport(MessageGenerator.INT_NODE_CLASS);
             headerGenerator.addImport(MessageGenerator.BINARY_NODE_CLASS);
+            buffer.printf("if (_verbose) {%n");
```
dajac: I wonder if we could find a better name than `verbose`. Perhaps, we could be more explicit and use something like `serializeRecords` as we actually only use it for this at the moment. What do you think?

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
```diff
@@ -261,7 +265,7 @@ private void generateWrite(String className,
                                StructSpec struct,
                                Versions parentVersions) {
         headerGenerator.addImport(MessageGenerator.JSON_NODE_CLASS);
-        buffer.printf("public static JsonNode write(%s _object, short _version) {%n",
+        buffer.printf("public static JsonNode write(%s _object, short _version, boolean _verbose) {%n",
```
dajac: Instead of changing the usage of this method everywhere in the code base, how about generating an overloaded method which call this one with `verbose=true`? I only expect this one to be used by the request logger at the moment so it is also more convenient.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3354,7 +3354,7 @@ class KafkaApis(val requestChannel: RequestChannel,
       case Some(response) =>
         val responseSend = request.context.buildResponse(response)
         val responseString =
-          if (RequestChannel.isRequestLoggingEnabled) Some(response.toString(request.context.apiVersion))
+          if (RequestChannel.isRequestLoggingEnabled) Some(RequestConvertToJson.response(response, request.context.apiVersion, true))
```
dajac: I think that we should set this to `false`.

core/src/main/scala/kafka/server/KafkaApis.scala
dajac: `verbose` should be `false` here (if we keep it).

core/src/main/scala/kafka/tools/TestRaftRequestHandler.scala
```diff
@@ -39,7 +40,7 @@ class TestRaftRequestHandler(
 
   override def handle(request: RequestChannel.Request): Unit = {
     try {
-      trace(s"Handling request:${request.requestDesc(true)} from connection ${request.context.connectionId};" +
+      trace(s"Handling request:${RequestConvertToJson.requestDesc(request.header, request.loggableRequest, true)} from connection ${request.context.connectionId};" +
```
dajac: `verbose` should be `false` here (if we keep it).

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
dajac: Could we also add a comment here which explains why we are doing this? We can also add the KIP number.

dajac (commented):
@anatasiavela Thanks for the update! It looks pretty good now. There are few things to fix and to improve but I think that we are on the right track wrt. the overall approach. I have left some more comments.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: I'm not too sure what this should test, since we don't have an expected output for this besides all fields being filled. What do you think we should test?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
anatasiavela: This ends up breaking `testFormatOfProduceRequestNode()` because it can't deserialize and serialize the data because of the inconsistency problem discussed below.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
dajac: Yeah, I was thinking about a basic test like this which verifies the fields. We had few typos so it is good to test this.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
dajac: Right. We need to update that test for the time being until we migrate the produce request/response. We may be able to verify the output differently.

clients/src/test/java/org/apache/kafka/common/message/SimpleExampleMessageTest.java
```diff
@@ -394,7 +394,7 @@ private void testRoundTrip(SimpleExampleMessageData message,
         assertEquals(message.hashCode(), messageFromStruct.hashCode());
 
         // Check JSON serialization
-        JsonNode serializedJson = SimpleExampleMessageDataJsonConverter.write(message, version);
+        JsonNode serializedJson = SimpleExampleMessageDataJsonConverter.write(message, version, true);
```
dajac: It seems that we can revert this back as we have the overloaded method now.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,362 @@
...
+import org.apache.kafka.common.network.ClientInformation
+import org.apache.kafka.common.protocol.Errors
+import org.apache.kafka.common.record.RecordBatch
+import org.apache.kafka.common.requests._
+import org.apache.kafka.common.utils.CollectionUtils
+
+import scala.jdk.CollectionConverters._
+
+object RequestConvertToJson {
+  def request(request: AbstractRequest, verbose: Boolean): JsonNode = {
```
dajac: It seems that we can remove `verbose` as we don't really use it anymore. Instead, we could set it to false for the produce request below.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,362 @@
...
+      case req: UpdateFeaturesRequest => UpdateFeaturesRequestDataJsonConverter.write(req.data, request.version, verbose)
+      case req: UpdateMetadataRequest => UpdateMetadataRequestDataJsonConverter.write(req.data, request.version, verbose)
+      case req: VoteRequest => VoteRequestDataJsonConverter.write(req.data, request.version, verbose)
+      case req: WriteTxnMarkersRequest => WriteTxnMarkersRequestDataJsonConverter.write(req.data, request.version, verbose)
+      case _ => throw new IllegalStateException(s"ApiKey ${request.api} is not currently handled in `request`, the " +
+        "code should be updated to do so.");
+    }
+  }
+
+  def response(response: AbstractResponse, version: Short, verbose: Boolean): JsonNode = {
```
dajac: Same here. We can remove `verbose` and set it to false for the fetch response.

core/src/main/scala/kafka/tools/TestRaftRequestHandler.scala
```diff
@@ -93,10 +94,8 @@ class TestRaftRequestHandler(
     val response = responseOpt match {
       case Some(response) =>
         val responseSend = request.context.buildResponse(response)
-        val responseString =
-          if (RequestChannel.isRequestLoggingEnabled) Some(response.toString(request.context.apiVersion))
-          else None
-        new RequestChannel.SendResponse(request, responseSend, responseString, None)
+        val headerLog = RequestConvertToJson.requestHeaderNode(request.header)
+        new RequestChannel.SendResponse(request, responseSend, Some(headerLog), None)
```
dajac: I suppose that we need to keep checking `if (RequestChannel.isRequestLoggingEnabled)` here.

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,199 @@
...
+    val response = new ProduceResponse(responseData)
+    val actualNode = RequestConvertToJson.response(response, version, true)
+
+    val requestData = ProduceResponseDataJsonConverter.read(actualNode, version)
+    val expectedNode = ProduceResponseDataJsonConverter.write(requestData, version, true)
+
+    assertEquals(expectedNode, actualNode)
+  }
+
+  @Test def testFieldsRequestDescMetrics(): Unit = {
```
dajac: nit: Could we add a new line after `@Test`?

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,199 @@
...
+
+    val req = request(new AlterIsrRequest(new AlterIsrRequestData(), 0))
+    val byteBuffer = req.body[AbstractRequest].serialize(req.header)
+    val send = new NetworkSend(req.context.connectionId, byteBuffer)
+    val headerLog = RequestConvertToJson.requestHeaderNode(req.header)
+    val res = new RequestChannel.SendResponse(req, send, Some(headerLog), None)
+
+    val node = RequestConvertToJson.requestDescMetrics(req.header, res, req.loggableRequest, req.context, req.session,
+      1, 1, 1, 1, 1, 1, 1, 1, 1).asInstanceOf[ObjectNode]
+    val foundFields = getFieldNames(node)
```
dajac: Thinking a bit more about this. I think that it may be worth validating the values as well to ensure that the input maps to the output. If we use different values for each field, we could verify that the output JSON object is correct. `JSONNode` are comparable so we could construct the expected output and use it in `assertEquals`. What do you think?

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,199 @@
...
+
+    assertEquals(expectedFields, foundFields)
+  }
+
+  def request(req: AbstractRequest): RequestChannel.Request = {
```
dajac: nit: Could we make it private?

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
```diff
@@ -375,13 +385,36 @@ private void generateVariableLengthTargetToJson(Target target, Versions versions
                         target.sourceVariable())));
             } else {
                 headerGenerator.addImport(MessageGenerator.ARRAYS_CLASS);
+                headerGenerator.addImport(MessageGenerator.INT_NODE_CLASS);
+                buffer.printf("if (_serializeRecords) {%n");
```
dajac: I am not sure to understand why we are doing this change now. It was not there before. Could you elaborate?

Is it because the produce request does use `bytes` as type? It seems `bytes` was used there by mistake. FYI, the PR which will migrate the produce request to the automated protocol will change it to `records`: https://github.com/apache/kafka/pull/9401/files#diff-e6dde0832c4873b546db2ad0d37be899855a436972b0a2c5008b100a0ffff64dR50.

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,199 @@
...
+      if (key == ApiKeys.PRODUCE) {
+        // There's inconsistency with the toStruct schema in ProduceRequest
+        // and ProduceRequestDataJsonConverters where the field names don't
+        // match so the struct does not have the correct field names. This is
+        // a temporary workaround until ProduceRequest starts using ProduceRequestData
+        req = ProduceRequest.Builder.forCurrentMagic(0.toShort, 10000, new util.HashMap[TopicPartition, MemoryRecords]).build()
+      } else {
+        val struct = ApiMessageType.fromApiKey(key.id).newRequest().toStruct(version)
+        req = AbstractRequest.parseRequest(key, version, struct)
+      }
```
dajac: nit: We would usually use a `val` here and write the block as follow:
```
val req = if (key == ApiKeys.PRODUCE) {
  ...
} else {
  ...
}
```

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,199 @@
...
+import org.junit.Assert.assertEquals
+
+import scala.collection.mutable.ArrayBuffer
+
+class RequestConvertToJsonTest {
+
+  @Test
+  def testAllRequestTypesHandled(): Unit = {
+    val unhandledKeys = ArrayBuffer[String]()
+    ApiKeys.values().foreach(key => {
```
dajac: nit: We tend to write this as follow:
```
ApiKeys.values().foreach { key => 
  ...
}
```
This is another similar case in `testAllResponseTypesHandled`.

dajac (commented):
@anatasiavela Thanks for the update. The PR looks good overall. I have left minor suggestions and one question.

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
anatasiavela: Yea, this change is because ProduceRequest uses `bytes` type. It would be worthwhile if we waited until we migrated the produce request to the automated protocol so we wouldn't have to change this.

dajac: Understood. I think that we should revert this. I think that it makes sense to wait until we complete the migration of the remaining requests. We should have them pretty soon now.

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
anatasiavela: I've modified the test to validate the values, but I'm not sure the worth in doing this especially when constructing the expected output is just copying the code for the node construction into the test, unless you were thinking of another way of constructing the node. Plus it's kind of a hassle when we try to modify `requestDescMetric` because it would need to be changed in 2 places: the code and the test.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,362 @@
...
+      topicsArray.add(topicNode)
+    }
+    node.set("topics", topicsArray)
+    node
+  }
+
+  /**
+   * Temporary until switch to use the generated schemas.
+   */
+  def produceResponseNode(response: ProduceResponse, version: Short): JsonNode = {
```
lbradstreet: Seeing as https://github.com/apache/kafka/pull/9401 is about to merge, maybe we can wait for that before merging this and drop our manual JSON generation code for produce request/response?

dajac: We are actually waiting on that one and this one https://github.com/apache/kafka/pull/9547.

lbradstreet: Great

anatasiavela: Yes! we're waiting on #9401 and #9547 to merge. Once they're merged, I'll pick this PR back up and will remove all the temporary case handling done for the Produce and OffsetsForLeaderEpoch request/response.

dajac:
@anatasiavela Both PRs have been merged so we can proceed with this one.

There is something that we must consider that I was not aware of: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java#L235. When the ProduceRequest is processed in the KafkaApis layer, its internal data is set to null to free up the memory. That means that we won't have it to log the request. We need to take this into account.

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -121,9 +122,16 @@ object RequestChannel extends Logging {
       }
     }
 
-    def responseString(response: AbstractResponse): Option[String] = {
+    def requestLog: Option[JsonNode] = {
```
dajac: We must use a `val` to store the `requestLog` in order to guarantee that the JSON representation is computed before the request is processed by the api layer. Otherwise, we still have a ProduceRequest without the data. The `val` may be private as well and I would put it right below `bodyAndSize` with a comment which explains why we do this.

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -137,11 +145,8 @@ object RequestChannel extends Logging {
       }
     }
 
-    def requestDesc(details: Boolean): String = {
-      val forwardDescription = envelope.map { request =>
-        s"Forwarded request: ${request.context} "
-      }.getOrElse("")
-      s"$forwardDescription$header -- ${loggableRequest.toString(details)}"
+    def requestDesc: JsonNode = {
```
dajac: It seems that this one is used from `KafkaApis` and others. It may be better to keep the original implementation for these cases.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,225 @@
...
+      case req: LeaveGroupRequest => LeaveGroupRequestDataJsonConverter.write(req.data, request.version)
+      case req: ListGroupsRequest => ListGroupsRequestDataJsonConverter.write(req.data, request.version)
+      case req: ListOffsetRequest => ListOffsetRequestDataJsonConverter.write(req.data, request.version)
+      case req: ListPartitionReassignmentsRequest => ListPartitionReassignmentsRequestDataJsonConverter.write(req.data, request.version)
+      case req: MetadataRequest => MetadataRequestDataJsonConverter.write(req.data, request.version)
+      case req: OffsetCommitRequest => OffsetCommitRequestDataJsonConverter.write(req.data, request.version)
+      case req: OffsetDeleteRequest => OffsetDeleteRequestDataJsonConverter.write(req.data, request.version)
+      case req: OffsetFetchRequest => OffsetFetchRequestDataJsonConverter.write(req.data, request.version)
+      case req: OffsetsForLeaderEpochRequest => OffsetForLeaderEpochRequestDataJsonConverter.write(req.data, request.version)
+      case req: ProduceRequest => ProduceRequestDataJsonConverter.write(req.data, request.version)
```
dajac: Shouldn't we pass `false` here in order to not serialize the records?

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  def clientInfoNode(clientInfo: ClientInformation): JsonNode = {
+    val node = new ObjectNode(JsonNodeFactory.instance)
+    node.set("softwareName", new TextNode(clientInfo.softwareName))
+    node.set("softwareVersion", new TextNode(clientInfo.softwareVersion))
+    node
+  }
+
+  def topicPartitionNode(topicPartition: TopicPartition): JsonNode = {
```
dajac: It seems that we don't use this one any more.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,225 @@
...
+    node.set("securityProtocol", new TextNode(context.securityProtocol.toString))
+    node.set("listener", new TextNode(context.listenerName.value))
+    node.set("clientInformation", clientInfoNode(context.clientInformation))
+    node
+  }
+
+  def requestDesc(header: RequestHeader, requestNode: Option[JsonNode], envelope: Option[RequestChannel.Request] = None): JsonNode = {
+    val node = new ObjectNode(JsonNodeFactory.instance)
+    node.set("isForwarded", if (envelope.isDefined) BooleanNode.TRUE else BooleanNode.FALSE)
+    node.set("forwardedRequest", envelope.map(request => requestContextNode(request.context)).getOrElse(new TextNode("")))
```
dajac: I am not sure that this is necessary. Knowing that the request is forwarded is enough, I suppose.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  def topicPartitionNode(topicPartition: TopicPartition): JsonNode = {
+    val node = new ObjectNode(JsonNodeFactory.instance)
+    node.set("topic", new TextNode(topicPartition.topic))
+    node.set("partition", new IntNode(topicPartition.partition))
+    node
+  }
+
+  def requestDescMetrics(header: RequestHeader, requestNode: Option[JsonNode], responseNode: Option[JsonNode],
```
dajac, line 181: We must pass either the `envelope ` or `isForwarded` here as well. This is the main request log and, as it is, it does not tell if the request is forwarded or not.

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
```diff
@@ -377,11 +387,24 @@ private void generateVariableLengthTargetToJson(Target target, Versions versions
...
+            // KIP-673: When logging requests/responses, we do not serialize the record, instead we
+            // output its sizeInBytes, because outputting the bytes is not very useful and can be
+            // quite expensive. Otherwise, we will serialize the record.
+            buffer.printf("if (_serializeRecords) {%n");
+            buffer.incrementIndent();
             buffer.printf("%s;%n", target.assignmentStatement("new BinaryNode(new byte[]{})"));
+            buffer.decrementIndent();
+            buffer.printf("} else {%n");
+            buffer.incrementIndent();
+            buffer.printf("%s;%n", target.assignmentStatement(
```
dajac: I wonder if we should name the field `%sSizeInBytes`. I just looked at the result and having `"records":83` in the request log is not super clear to me.

dajac (commented):
@anatasiavela Thanks for the updates. I think that we are very close. I have left few small comments and suggestions. Could you also rebase the PR please?

generator/src/main/java/org/apache/kafka/message/JsonConverterGenerator.java
anatasiavela: I do see the issue of it not being super clear, but I don't think we can change the field name from here. Doing `%sSizeInBytes` would just add the name at the end of the line which would result in a compilation error. Unless you mean to change the field name in the `.json` file, but it would change the name for both the serialize and non-serialize case.

dajac: Would something like the following work?
```
buffer.printf("_node.set(\"%sSizeInBytes\", new IntNode(%s.sizeInBytes()));%n",
  target.field().camelCaseName(),
  target.sourceVariable());
```

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -97,7 +98,8 @@ import scala.annotation.nowarn
 /**
  * Logic to handle the various Kafka requests
  */
-class KafkaApis(val requestChannel: RequestChannel,
+class
+KafkaApis(val requestChannel: RequestChannel,
```
dajac: This is likely a mistake.

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -123,9 +131,9 @@ object RequestChannel extends Logging {
       }
     }
 
-    def responseString(response: AbstractResponse): Option[String] = {
+    def responseNode(response: AbstractResponse): Option[JsonNode] = {
       if (RequestChannel.isRequestLoggingEnabled)
-        Some(response.toString)
+        Some(RequestConvertToJson.response(response, context.apiVersion()))
```
dajac: nit: `context.apiVersion()` -> `context.apiVersion`

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -200,7 +208,7 @@ object RequestChannel extends Logging {
       }
     }
 
-    trace(s"Processor $processor received request: ${requestDesc(true)}")
+    trace(s"Processor $processor received request: ${RequestConvertToJson.requestDesc(header, requestLog, isForwarded)}")
```
dajac: As discussed offline, I think that we should revert this one. We want to use JSON in the request log (aka `requestLogger`), not as a replacement of all the `toString` usages.

core/src/main/scala/kafka/network/RequestConvertToJson.scala
```diff
@@ -0,0 +1,216 @@
...
+    }
+  }
+
+  def requestHeaderNode(header: RequestHeader): JsonNode = {
+    val node = RequestHeaderDataJsonConverter.write(header.data, header.headerVersion, false).asInstanceOf[ObjectNode]
+    node.set("requestApiKeyName", new TextNode(header.apiKey.toString))
+    node
+  }
+
+  def requestContextNode(context: RequestContext): JsonNode = {
```
dajac: It seems that this one is not used anymore.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -183,7 +185,8 @@ class KafkaApis(val requestChannel: RequestChannel,
    */
   override def handle(request: RequestChannel.Request): Unit = {
     try {
-      trace(s"Handling request:${request.requestDesc(true)} from connection ${request.context.connectionId};" +
+      trace(s"Handling request:${RequestConvertToJson.requestDesc(request.header, request.requestLog, request.isForwarded)} " +
```
dajac: We should also revert this one.

core/src/main/scala/kafka/tools/TestRaftRequestHandler.scala
```diff
@@ -39,7 +40,8 @@ class TestRaftRequestHandler(
 
   override def handle(request: RequestChannel.Request): Unit = {
     try {
-      trace(s"Handling request:${request.requestDesc(true)} from connection ${request.context.connectionId};" +
+      trace(s"Handling request:${RequestConvertToJson.requestDesc(request.header, request.requestLog, request.isForwarded)} " +
```
dajac: We should also revert this one.

core/src/test/scala/unit/kafka/network/RequestConvertToJsonTest.scala
```diff
@@ -0,0 +1,154 @@
...
+      val req = AbstractRequest.parseRequest(key, version, bytes.buffer).request
+      try {
+        RequestConvertToJson.request(req)
+      } catch {
+        case _ : IllegalStateException => unhandledKeys += key.toString
+      }
+    }}
+    assertEquals("Unhandled request keys", ArrayBuffer.empty, unhandledKeys)
+  }
+
```
dajac, line 67: Could we also add unit tests for `RequestConvertToJson.requestDesc`, `RequestConvertToJson.requestHeaderNode` and `RequestConvertToJson.clientInfoNode`?

dajac (commented):
@anatasiavela Thanks for the updates. The PR looks pretty good. I have left few more minor comments.

dajac (approved):
LGTM. Thanks for your contribution, @anatasiavela!

PR #9527, opened 2020-10-29
MINOR: avoid unnecessary conversion and tuple when updating error met…

chia7712:
The conversion (from java collection to scala collection) and tuple is unnecessary for hot method  ```updateErrorMetrics```

### DIFF
--- a/core/src/main/scala/kafka/network/RequestChannel.scala
+++ b/core/src/main/scala/kafka/network/RequestChannel.scala
@@ -27,7 +27,6 @@ import kafka.log.LogConfig
 import kafka.metrics.KafkaMetricsGroup
 import kafka.server.KafkaConfig
 import kafka.utils.{Logging, NotNothing, Pool}
-import kafka.utils.Implicits._
 import org.apache.kafka.common.config.types.Password
 import org.apache.kafka.common.config.ConfigResource
 import org.apache.kafka.common.memory.MemoryPool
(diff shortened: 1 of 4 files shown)


chia7712:
the failed tests are unrelated to this PR. rebase to trigger QA

ijuma:
Can you clarify what the issue is? The conversion from the Java to Scala collection is cheap. And `forKeyValue` doesn't necessarily generate a tuple. Is the issue that the Scala wrapper for the Java collection doesn't have an optimized implementation of `scala.Map.foreachEntry` and allocates a tuple even though it doesn't need to?

ijuma:
Something like https://github.com/scala/scala/pull/9423/files should fix this issue without making our code less readable.

chia7712:
> Something like https://github.com/scala/scala/pull/9423/files should fix this issue without making our code less readable.

It is so badass :)

close this PR.

ijuma:
On deeper analysis, the override is already there:

https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/convert/JavaCollectionWrappers.scala#L347

So, can you check where the tuple allocations were coming from?

chia7712:
> can you check where the tuple allocations were coming from?

It seems I made an incorrect check as the ```tuple``` was caused by https://github.com/chia7712/kafka/blob/750d24d999c3e642c61c04fe218656a16e4a039d/clients/src/main/java/org/apache/kafka/common/requests/ProduceResponse.java#L316 rather than ```forKeyValue```. The responses was a wrap of scala collection so it produced ```tuple``` for java foreach function.

ijuma:
Hmm, maybe it's a different override missing (since it's over the map values collection), I'll look again.

chia7712:
> Something like https://github.com/scala/scala/pull/9423/files should fix this issue without making our code less readable.

@ijuma Your comment encourage me to check the ```foreachEntry``` of all map collections used by ```KafkaApis```. I have filed a PR (https://github.com/scala/scala/pull/9425) to add custom ```foreachEntry``` to ```mutable.HashMap```.

ijuma:
Thanks @chia7712.

PR #9528, opened 2020-10-29
MINOR: Add releaseTarGz to args for building docs

bbejeck:
This PR adds the `releaseTarGz` command back into the `release.py::command_stage_docs` method when building docs for staging

### DIFF
--- a/release.py
+++ b/release.py
@@ -262,7 +262,7 @@ def command_stage_docs():
     # version due to already having bumped the bugfix version number.
     gradle_version_override = docs_release_version(version)
 
-    cmd("Building docs", "./gradlew -Pversion=%s clean aggregatedJavadoc" % gradle_version_override, cwd=REPO_HOME, env=jdk8_env)
+    cmd("Building docs", "./gradlew -Pversion=%s clean releaseTarGz aggregatedJavadoc" % gradle_version_override, cwd=REPO_HOME, env=jdk8_env)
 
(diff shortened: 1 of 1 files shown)


ijuma:
Thanks. What happens without this?

bbejeck:
>What happens without this?

The method fails on this line `cmd('Extracting site-docs', 'tar xf %s --strip-components 1' % docs_tar, cwd=versioned_docs_path)` 

Because even if you ran `releaseTarGz` separately ahead of time, the `clean` before calling `aggreagateJavadoc` wipes everything out, so the `tar` command can't find the `site-docs` file

ijuma:
@bbejeck do we need the clean?

bbejeck:
> do we need the clean?

No, but I think the `release.py::stage-docs` makes more sense to have the `releaseTarGz` target and run clean beforehand.  
If you prefer to leave it out, I can amend the PR to remove the `clean`

ijuma:
Would it be enough to call `siteDocsTar` instead? It's weird to generate the release tarball to just get the site docs.

bbejeck:
>Would it be enough to call siteDocsTar instead?

Yeah, my bad that makes more sense

bbejeck:
updated this

\cc @ijuma

ijuma (approved):
LGTM if we tested this and it works correctly.

bbejeck:
test locally and works as expected

bbejeck:
only a python file so merging now

PR #9531, opened 2020-10-29
KAFKA-10661; Add new resigned state for graceful shutdown/initialization

hachikuji:
When initializing the raft state machine after shutting down as a leader, we were previously entering the "unattached" state, which means we have no leader and no voted candidate. This was a bug because it allowed a reinitialized leader to cast a vote for a candidate in the same epoch that it was already the leader of. This patch fixes the problem by introducing a new "resigned" state which allows us to retain the leader state so that we cannot change our vote and we will not accept additional appends.

This patch also revamps the shutdown logic to make use of the new "resigned" state. Previously we had a separate path in `KafkaRaftClient.poll` for the shutdown logic which resulted in some duplication. Instead now we incorporate shutdown behavior into each state's respective logic.

### DIFF
--- a/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
+++ b/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
@@ -756,31 +756,26 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
             request.topics().get(0).partitions().get(0);
 
         int requestEpoch = partitionRequest.leaderEpoch();
-        int requestReplicaId = partitionRequest.replicaId();
+        int requestLeaderId = partitionRequest.leaderId();
 
-        Optional<Errors> errorOpt = validateVoterOnlyRequest(requestReplicaId, requestEpoch);
+        Optional<Errors> errorOpt = validateVoterOnlyRequest(requestLeaderId, requestEpoch);
         if (errorOpt.isPresent()) {
             return buildEndQuorumEpochResponse(errorOpt.get());
         }
-
-        OptionalInt requestLeaderId = optionalLeaderId(partitionRequest.leaderId());
-        maybeTransition(requestLeaderId, requestEpoch, currentTimeMs);
+        maybeTransition(OptionalInt.of(requestLeaderId), requestEpoch, currentTimeMs);
 
         if (quorum.isFollower()) {
             FollowerState state = quorum.followerStateOrThrow();
-            if (state.leaderId() == requestReplicaId) {
+            if (state.leaderId() == requestLeaderId) {
                 List<Integer> preferredSuccessors = partitionRequest.preferredSuccessors();
                 if (!preferredSuccessors.contains(quorum.localId)) {
                     return buildEndQuorumEpochResponse(Errors.INCONSISTENT_VOTER_SET);
                 }
                 long electionBackoffMs = endEpochElectionBackoff(preferredSuccessors);
+                logger.debug("Overriding follower fetch timeout to {} after receiving " +
+                    "EndQuorumEpoch request from leader {} in epoch {}", electionBackoffMs,
+                    requestLeaderId, requestEpoch);
                 state.overrideFetchTimeout(currentTimeMs, electionBackoffMs);
-            }
-        } else if (quorum.isVoted()) {
-            VotedState state = quorum.votedStateOrThrow();
-            if (state.votedId() == requestReplicaId) {
-                long electionBackoffMs = binaryExponentialElectionBackoffMs(1);
-                state.overrideElectionTimeout(currentTimeMs, electionBackoffMs);
             }
         }
         return buildEndQuorumEpochResponse(Errors.NONE);
@@ -1385,13 +1380,14 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
         return connection.remainingRequestTimeMs(currentTimeMs);
     }
 
-    private EndQuorumEpochRequestData buildEndQuorumEpochRequest(List<Integer> preferredSuccessors) {
+    private EndQuorumEpochRequestData buildEndQuorumEpochRequest(
+        ResignedState state
+    ) {
         return EndQuorumEpochRequest.singletonRequest(
             log.topicPartition(),
-            quorum.localId,
             quorum.epoch(),
-            quorum.leaderIdOrNil(),
-            preferredSuccessors
+            quorum.localId,
+            state.preferredSuccessors()
         );
     }
 
@@ -1519,15 +1515,14 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
     }
 
     private long pollResigned(long currentTimeMs) throws IOException {
-        GracefulShutdown shutdown = this.shutdown.get();
         ResignedState state = quorum.resignedStateOrThrow();
-
         long endQuorumBackoffMs = maybeSendRequests(
             currentTimeMs,
             state.unackedVoters(),
-            () -> buildEndQuorumEpochRequest(state.preferredSuccessors())
+            () -> buildEndQuorumEpochRequest(state)
         );
 
+        GracefulShutdown shutdown = this.shutdown.get();
         final long stateTimeoutMs;
         if (shutdown != null) {
             // If we are shutting down, then we will remain in the resigned state
@@ -1535,7 +1530,7 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
             stateTimeoutMs = shutdown.remainingTimeMs();
         } else if (state.hasElectionTimeoutExpired(currentTimeMs)) {
             transitionToCandidate(currentTimeMs);
-            stateTimeoutMs = 0;
+            stateTimeoutMs = 0L;
         } else {
             stateTimeoutMs = state.remainingElectionTimeMs(currentTimeMs);
         }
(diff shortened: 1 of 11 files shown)


raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1543,9 +1516,40 @@ private long maybeAppendBatches(
         return timeUnitFlush;
     }
 
+    private long pollResigned(long currentTimeMs) throws IOException {
+        GracefulShutdown shutdown = this.shutdown.get();
+        ResignedState state = quorum.resignedStateOrThrow();
+
+        long endQuorumBackoffMs = maybeSendRequests(
```
dengziming, line 1519: If the cluster resigns from candidateState, it will also send EndQuorumEpochRequest to all voters?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1543,9 +1516,40 @@ private long maybeAppendBatches(
...
+            currentTimeMs,
+            state.unackedVoters(),
+            () -> buildEndQuorumEpochRequest(state.preferredSuccessors())
+        );
+
+        final long stateTimeoutMs;
+        if (shutdown != null) {
+            // If we are shutting down, then we will remain in the resigned state
+            // until either the shutdown expires or an election bumps the epoch
```
guozhangwang, line 1529: If we are shutting down, why we still want to care about election at all?

guozhangwang (commented):
Had some high-level thoughts about how to fix the issue: I feel using the `resign` state for both cases 1) re-initializing from a previous leader/candidate state, 2) resigning due to shutdown is a bit confusing, since in the first case we may still want to transit to candidate or follower later upon election timeout / recipient of vote request, while in the second case we'd not want to do anything but just complete the shutdown smoothly.

On the other hand, upon becoming candidate node would bump up its epoch before sending out the vote request, and if the bumped epoch is indeed larger than the recipient's current epoch the recipient would transit to `unattached` first, and ONLY in that state we would possible grant the vote.

So what if we change the handle-vote-request logic, such that we do not rely on the `unattached` state to decide whether or not to grant vote, but to grant a vote only if its epoch is larger (and of course the end-offset epoch validation passes) first, and then consider transiting the state. In this case we can still re-initialize as `unattached` as long as our epoch is retained. And as for shutdown, personally I feel adding a ShuttingDown state may be more reasonable, in which case `pollShutdown` that basically ignores the election timer still make sense.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Yeah, that is why the `EndQuorumEpoch` specifies both a `leaderId` and a `replicaId`. The idea is that this allows a voter who had voted for the candidate to start a new election more quickly. Admittedly, I am not sure how valuable this optimization is in practice. We could consider simplifying the protocol so that it is only sent by leaders.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: This node can stick around long enough to help another leader get elected.

hachikuji:
@guozhangwang Thanks for the comment. It might be useful to consider this change with https://issues.apache.org/jira/browse/KAFKA-10655 in mind. The "resigned" state is not necessarily intended to be a final state which is only used on shutdown. We want to give the leader the ability to resign in order to trigger an election without necessarily shutting down. So it is not just for the case of starting up that we allow transitions from this state to other states.

Perhaps it is just me, but I actually liked the fact that the node would initialize in the same state that it was shutdown in. And it seemed to me that the clearest way to represent the situation where the node is still the current elected leader, but has basically self-fenced itself was through a separate state.

One thing I am strongly considering, however, is changing this state machine so that the resigned state is only for leaders. That would definitely simplify the logic. The optimization mentioned above in response to @dengziming 's question seems unlikely to have much benefit in practice. We could always reconsider it in the future of course.

hachikuji:
I decided to change the behavior so that the "resigned" state is reserved for leaders and `EndQuorumEpoch` is only sent by resigned leaders. It was always odd that a candidate could send `EndQuorumEpoch` request in an epoch that it had not yet sent `BeginQuorumEpoch` for. So this simplifies both the state machine and the protocol.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1560,9 +1558,34 @@ private long pollLeader(long currentTimeMs) {
...
+        GracefulShutdown shutdown = this.shutdown.get();
+
+        if (shutdown != null) {
+            // If we happen to shutdown while we are a candidate, we will continue
+            // with the current election until one of the following conditions is met:
+            //  1) we are elected as leader (which allows us to resign)
+            //  2) another leader is elected
+            //  3) the shutdown timer expires
+            long minRequestBackoffMs = maybeSendVoteRequests(state, currentTimeMs);
+            return Math.min(shutdown.remainingTimeMs(), minRequestBackoffMs);
```
dengziming, line 1591: should this be `Math.min(shutdown.remainingTimeMs(), minRequestBackoffMs, state.remainingElectionTimeMs(currentTimeMs))`

hachikuji: The intent is to ignore the election timeout in order to prevent a shutting down broker from becoming a candidate. Does that make sense?

dengziming: Understand, The candidate will try to complete only the current election when shutting down, so just ignore the election timeout.

guozhangwang:
> One thing I am strongly considering, however, is changing this state machine so that the resigned state is only for leaders. That would definitely simplify the logic. The optimization mentioned above in response to @dengziming 's question seems unlikely to have much benefit in practice. We could always reconsider it in the future of course.

Sounds good!

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1601,7 +1618,12 @@ private long pollFollower(long currentTimeMs) throws IOException {
     }
 
     private long pollFollowerAsVoter(FollowerState state, long currentTimeMs) throws IOException {
-        if (state.hasFetchTimeoutExpired(currentTimeMs)) {
+        GracefulShutdown shutdown = this.shutdown.get();
+        if (shutdown != null) {
+            // If we are a follower, then we can shutdown immediately. We want to
+            // skip the transition to candidate in any case.
+            return 0;
```
guozhangwang, line 1625: Why the behavior of `pollFollowerAsVoter` and `pollVoted` are different when shutting down? Could the former case still help in casting and completing a vote as well?

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
```diff
@@ -21,18 +21,21 @@
...
 
 /**
- * This class is responsible for managing the current state of this node and ensuring only
- * valid state transitions.
+ * This class is responsible for managing the current state of this node and ensuring
+ * only valid state transitions. Below we define the possible state transitions and
+ * how they are triggered:
  *
- * Unattached =>
+ * Unattached|Resigned =>
```
guozhangwang, line 38: Makes me thinking: if we have an even number sized quorum (say 2N), and the leader is resigning. Then before the leader shutdown we need N+1 votes, while after the leader shutdown, the quorum size shrink to 2N-1 and we would only need N votes. So if the resigning leader gives it a vote to a candidate and then shutdown, while the candidates thinks they only need N votes, would that potentially result in two candidates claiming victory --- somehow this sounds quite close to the real world :P --- each with N votes while one of them has the vote from the resigned leader?

guozhangwang (commented):
Made a pass over the code again. I liked the current FSM more after synching with @hachikuji.

Just one quick question about the resigning leader still being able vote: currently it may not be an issue since we do not support quorum size changes anyways, but wondering if in the future it may become a risk.

Otherwise, LGTM!

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
hachikuji: Hmm.. The quorum size does not change because a leader resigns. If there are 2N nodes in the cluster, then we always need N + 1 votes, so I don't think this case is possible.

hachikuji: In the future, when we have reassignment, we will still have to protect every quorum change with a majority of the current nodes. The proposal we had previously only allowed single-node changes, which meant that any majority was a majority before and after the state was applied.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: If we are a follower, then there is no election in progress to help with.

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
guozhangwang: Okay now I remembered what we discussed before.

What I was wondering is, say with quorum size 6, we would need 4 votes to elect leader; if the current leader shutdown and before it is restarted, the quorum size is 5 so logically we only need 3 votes --- but as long as we require that during this transition we still require 4 votes even with 5 alive quorum members we are fine.

guozhangwang:
LGTM.

PR #9535, opened 2020-10-30
MINOR: remove redundant return statement

dengziming:
the result of `GroupMetadataManager.storeOffsets` is Unit, so remove the redundant return statement


dengziming:
Hi, @hachikuji , please take a look.

chia7712 (approved):

PR #9539, opened 2020-10-31
KAFKA-10634: Adding LeaderId to Voters list in LeaderChangeMessage

vamossagar12:
comment to ping reviewers.*

### DIFF
--- a/clients/src/main/resources/common/message/LeaderChangeMessage.json
+++ b/clients/src/main/resources/common/message/LeaderChangeMessage.json
@@ -22,7 +22,9 @@
     {"name": "LeaderId", "type": "int32", "versions": "0+",
       "about": "The ID of the newly elected leader"},
     {"name": "Voters", "type": "[]Voter", "versions": "0+",
-      "about": "The voters who voted for the current leader"}
+      "about": "The set of voters in the quorum for this epoch"},
+    {"name": "GrantingVoters", "type": "[]Voter", "versions": "0+",
+      "about": "The voters who voted for the leader at the time of election"}
   ],
   "commonStructs": [
     { "name": "Voter", "versions": "0+", "fields": [
--- a/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
+++ b/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
@@ -314,7 +314,7 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
 
         long currentTimeMs = time.milliseconds();
         if (quorum.isLeader()) {
-            onBecomeLeader(currentTimeMs);
+            throw new IllegalStateException("Voter cannot initialize as a Leader");
         } else if (quorum.isCandidate()) {
             onBecomeCandidate(currentTimeMs);
         } else if (quorum.isFollower()) {
@@ -371,14 +371,23 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
         );
     }
 
-    private void appendLeaderChangeMessage(LeaderState state, long currentTimeMs) {
-        List<Voter> voters = state.followers().stream()
+    private static List<Voter> convertToVoters(Set<Integer> voterIds) {
+        return voterIds.stream()
             .map(follower -> new Voter().setVoterId(follower))
             .collect(Collectors.toList());
+    }
+
+    private void appendLeaderChangeMessage(LeaderState state, long currentTimeMs) {
+        List<Voter> voters = convertToVoters(state.followers());
+        List<Voter> grantingVoters = convertToVoters(state.grantingVoters());
+
+        // Adding the leader to the voters as any voter always votes for itself.
+        voters.add(new Voter().setVoterId(state.election().leaderId()));
 
         LeaderChangeMessage leaderChangeMessage = new LeaderChangeMessage()
             .setLeaderId(state.election().leaderId())
-            .setVoters(voters);
+            .setVoters(voters)
+            .setGrantingVoters(grantingVoters);
 
         MemoryRecords records = MemoryRecords.withLeaderChangeMessage(
             currentTimeMs, quorum.epoch(), leaderChangeMessage);
@@ -1556,7 +1565,7 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
 
         long timeUntilSend = maybeSendRequests(
             currentTimeMs,
-            state.nonEndorsingFollowers(),
+            state.nonEndorsingVoters(),
             this::buildBeginQuorumEpochRequest
         );
 
--- a/raft/src/main/java/org/apache/kafka/raft/LeaderState.java
+++ b/raft/src/main/java/org/apache/kafka/raft/LeaderState.java
@@ -34,13 +34,15 @@ public class LeaderState implements EpochState {
     private Optional<LogOffsetMetadata> highWatermark;
     private final Map<Integer, VoterState> voterReplicaStates = new HashMap<>();
     private final Map<Integer, ReplicaState> observerReplicaStates = new HashMap<>();
+    private final Set<Integer> grantingVoters = new HashSet<>();
     private static final long OBSERVER_SESSION_TIMEOUT_MS = 300_000L;
 
     protected LeaderState(
         int localId,
         int epoch,
         long epochStartOffset,
-        Set<Integer> voters
+        Set<Integer> voters,
+        Set<Integer> grantingVoters
     ) {
         this.localId = localId;
         this.epoch = epoch;
@@ -51,6 +53,7 @@ public class LeaderState implements EpochState {
             boolean hasEndorsedLeader = voterId == localId;
             this.voterReplicaStates.put(voterId, new VoterState(voterId, hasEndorsedLeader));
         }
+        this.grantingVoters.addAll(grantingVoters);
     }
 
     @Override
@@ -72,11 +75,15 @@ public class LeaderState implements EpochState {
         return voterReplicaStates.keySet().stream().filter(id -> id != localId).collect(Collectors.toSet());
     }
 
+    public Set<Integer> grantingVoters() {
+        return this.grantingVoters;
+    }
+
     public int localId() {
         return localId;
     }
 
-    public Set<Integer> nonEndorsingFollowers() {
+    public Set<Integer> nonEndorsingVoters() {
         Set<Integer> nonEndorsing = new HashSet<>();
         for (VoterState state : voterReplicaStates.values()) {
             if (!state.hasEndorsedLeader)
--- a/raft/src/main/java/org/apache/kafka/raft/QuorumState.java
+++ b/raft/src/main/java/org/apache/kafka/raft/QuorumState.java
@@ -420,7 +420,8 @@ public class QuorumState {
             localId,
             epoch(),
             epochStartOffset,
-            voters
+            voters,
+            candidateState.grantingVoters()
         ));
     }
 
--- a/raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
+++ b/raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
@@ -391,9 +391,9 @@ public final class RaftClientTestContext {
         return raftMessage.correlationId();
     }
 
-    int assertSentVoteRequest(int epoch, int lastEpoch, long lastEpochOffset) {
+    int assertSentVoteRequest(int epoch, int lastEpoch, long lastEpochOffset, int numVoteReceivers) {
         List<RaftRequest.Outbound> voteRequests = collectVoteRequests(epoch, lastEpoch, lastEpochOffset);
-        assertEquals(1, voteRequests.size());
+        assertEquals(numVoteReceivers, voteRequests.size());
         return voteRequests.iterator().next().correlationId();
     }
 
@@ -448,9 +448,9 @@ public final class RaftClientTestContext {
         channel.mockReceive(new RaftResponse.Inbound(correlationId, response, sourceId));
     }
 
-    int assertSentBeginQuorumEpochRequest(int epoch) {
+    int assertSentBeginQuorumEpochRequest(int epoch, int numBeginEpochRequests) {
         List<RaftRequest.Outbound> requests = collectBeginEpochRequests(epoch);
-        assertEquals(1, requests.size());
+        assertEquals(numBeginEpochRequests, requests.size());
         return requests.get(0).correlationId;
     }
 
@@ -723,6 +723,7 @@ public final class RaftClientTestContext {
     static void verifyLeaderChangeMessage(
         int leaderId,
         List<Integer> voters,
+        List<Integer> grantingVoters,
         ByteBuffer recordKey,
         ByteBuffer recordValue
     ) {
@@ -732,6 +733,8 @@ public final class RaftClientTestContext {
         assertEquals(leaderId, leaderChangeMessage.leaderId());
         assertEquals(voters.stream().map(voterId -> new Voter().setVoterId(voterId)).collect(Collectors.toList()),
             leaderChangeMessage.voters());
+        assertEquals(grantingVoters.stream().map(voterId -> new Voter().setVoterId(voterId)).collect(Collectors.toSet()),
+            new HashSet<>(leaderChangeMessage.grantingVoters()));
     }
 
     void assertFetchRequestData(
--- a/raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
+++ b/raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
@@ -165,7 +165,7 @@ public class KafkaRaftClientTest {
         context.time.sleep(context.electionTimeoutMs);
         context.pollUntilSend();
         context.assertVotedCandidate(epoch + 1, localId);
-        context.assertSentVoteRequest(epoch + 1, 0, 0L);
+        context.assertSentVoteRequest(epoch + 1, 0, 0L, 1);
     }
 
     @Test
@@ -238,7 +238,7 @@ public class KafkaRaftClientTest {
         context.pollUntilSend();
         context.assertVotedCandidate(1, context.localId);
 
-        int correlationId = context.assertSentVoteRequest(1, 0, 0L);
+        int correlationId = context.assertSentVoteRequest(1, 0, 0L, 1);
         context.deliverResponse(correlationId, otherNodeId, context.voteResponse(true, Optional.empty(), 1));
 
         // Become leader after receiving the vote
@@ -252,7 +252,7 @@ public class KafkaRaftClientTest {
 
         // Send BeginQuorumEpoch to voters
         context.client.poll();
-        context.assertSentBeginQuorumEpochRequest(1);
+        context.assertSentBeginQuorumEpochRequest(1, 1);
 
         Records records = context.log.read(0, Isolation.UNCOMMITTED).records;
         RecordBatch batch = records.batches().iterator().next();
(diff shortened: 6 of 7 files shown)


raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -317,6 +317,9 @@ private void appendLeaderChangeMessage(LeaderState state, long currentTimeMs) {
             .map(follower -> new Voter().setVoterId(follower))
             .collect(Collectors.toList());
 
+        // Adding the leader to the voters as the protocol ensures that leader always votes for itself.
+        voters.add(new Voter().setVoterId(state.election().leaderId()));
```
jsancio, line 385: Similar to `LeaderState::nonEndorsingFollower`, I think you want to add a method to `LeaderState` with the following signature `public Set<Integer> endorsingFollower()`.

jsancio (commented):
Thanks for the PR @vamossagar12

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
vamossagar12: Thanks @jsancio . I am slightly confused regarding your comment though(maybe because of my limited understanding of the protocol) so plz help me understand here:
The block of code where i made. the changes, picks all the followers and adds them as the voters:

`List<Voter> voters = state.followers().stream()
            .map(follower -> new Voter().setVoterId(follower))
            .collect(Collectors.toList());`

While the LeaderChange message schema, states that voterIds are all those voters who voted for the currently elected leader. 

Continuing on this, from what I understood in the handling of BeginQuorumEpoch response handling section in kIP-595, the moment a leader receives a response from a voter with no errors, it records an endorsement from the voter for itself. It also states that if a Voter accepts a BeginQuorumEpoch(and thereby sending the accepted response), it automatically transitions to a follower state for that leader. Lastly, the new leader will keep sending BeginQuorumEpoch requests to non endorsing voters until it doesn't receive them (or doesn't infer through the Fetch API in case of partitioning).

So, the questions I have are:

1. Since the leader waits for the endorsements from all voters, doesn't it mean that eventually the followers method in LeaderState would return all voters? While the LeaderState message schema mentions that VotedIDs are those voters are voted for this new leader. 
2. I had thought that the nonEndorsingFollower method was used to find all followers who haven't yet endorsed for the new leader and keep sending requests to them. In that assumption, I am not able to follow your statement of adding a new method to add endorsingFollower wrt this PR. 

Again, I apologise if I made some comments which don't make sense. As I said, I am not totally well verse with the internal implementations being done here.

hachikuji: I think the original KIP stated that the LeaderChange message would encode the set of voters that had voted for the leader. We thought this might be useful for debugging. Later on, we had a change of heart and decided it would just be the set of voters. Now I'm thinking it might be useful to have both. The log will always remember who the voters were at the time of the election and which voters had granted the leader's candidacy, which could be helpful in case of misconfigurations.

For the set of voters which voted for the current leader, I think what we want is `CandidateState.grantingVoters`. However, by the time `onBecomeLeader` is fired, we have already dropped the `CandidateState`. One option is to carry `grantingVoters` over to `LeaderState`. We might also be able to pass it through `onBecomeLeader`. This will be easier if we get rid of the call to `onBecomeLeader` in `initialize()`. Following KAFKA-10527, it is not possible to initialize as a leader, so we could raise an exception instead.

vamossagar12: Thanks @hachikuji . So, from the context of this PR, do you suggest to continue using the definition of Voters in the LeaderChange message as all the voters ? Maybe we can create separate issues for:
1) changing LeaderChange message to include both all voters and endorsing voters for that Leader.
2) Make relevant changes to be able to pass the grantingVoters information from CandidateState to LeaderState. These include the things you mentioned like removing `onBecomeLeader` from `initialize()` or throw exception.

Just curious on the last part. As per KAFKA-10527, a node can't be initialized as a leader. So, this block of code is effectively unused then:

`if (quorum.isLeader()) {
onBecomeLeader(currentTimeMs);
        } else if (quorum.isCandidate()) {
            onBecomeCandidate(currentTimeMs);
        } `


Or should we chase all the above as part of this PR itself? Plz let me know.

hachikuji: @vamossagar12 Up to you I guess. I'm ok doing it all here since the changes seem pretty small.

vamossagar12: Alright. Thanks @hachikuji . So, I need to add voters + voted for the current Leader in the LeaderChange message and see how to pass the latter to LeaderState.

vamossagar12: @hachikuji i have made the changes you had suggested. 
cc @jsancio

vamossagar12: hey @hachikuji, did you get a chance to review the changes?

clients/src/main/resources/common/message/LeaderChangeMessage.json
```diff
@@ -22,7 +22,9 @@
     {"name": "LeaderId", "type": "int32", "versions": "0+",
       "about": "The ID of the newly elected leader"},
     {"name": "Voters", "type": "[]Voter", "versions": "0+",
-      "about": "The voters who voted for the current leader"}
+      "about": "The voters who voted at the time of election"},
+    {"name": "EndorsingVoters", "type": "[]Voter", "versions": "0+",
```
hachikuji: Could we use "Granting" instead of "Endorsing"? I think this is a more consistent term considering `LeaderState`.

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
```diff
@@ -394,7 +394,7 @@ public void transitionToLeader(long epochStartOffset) throws IOException {
             localId,
             epoch(),
             epochStartOffset,
-            voters
+            voters, candidateState.grantingVoters()
```
hachikuji: nit: can we move to next line to follow the convention here?

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
```diff
@@ -709,6 +709,9 @@ static void verifyLeaderChangeMessage(
         assertEquals(leaderId, leaderChangeMessage.leaderId());
         assertEquals(voters.stream().map(voterId -> new Voter().setVoterId(voterId)).collect(Collectors.toList()),
             leaderChangeMessage.voters());
+        assertEquals(voters.stream().map(voterId -> new Voter().setVoterId(voterId)).collect(Collectors.toSet()),
```
hachikuji: Hmm.. It's curious that we can always rely on the full voter set. Do we not have any test cases where we have not received votes from all members?

hachikuji (commented):
Thanks for the patch. Left a few minor comments.

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
vamossagar12: Yeah at least I couldn't find one. This method `verifyLeaderChangeMessage` is invoked only from 2 methods, one being `testInitializeAsCandidateAndBecomeLeader` and the other one is `testLeaderAppendSingleMemberQuorum`. In both these, the full vote is granted. Plz let me know if you want another test which takes care of the case where not all votes are received.

vamossagar12:
> Thanks for the patch. Left a few minor comments.

Thanks @hachikuji , i have replied on the comments..

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
hachikuji: Yeah, if you don't mind, it seems like a gap. Thanks!

vamossagar12: @hachikuji , I have added a test case where i initialize a quorum of 3 and got votes from majority. Needed to tweak a couple of methods as some of them were having hardcoded values of 1 on the expected value. Plz review.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -370,14 +370,23 @@ private void onBecomeLeader(long currentTimeMs) {
         );
     }
 
+    private List<Voter> getVotersFromVoterId(Set<Integer> voterIds) {
```
jsancio: nit: we can make this a `static` function and rename it to something like `convertToVoters`.

clients/src/main/resources/common/message/LeaderChangeMessage.json
```diff
@@ -22,7 +22,9 @@
     {"name": "LeaderId", "type": "int32", "versions": "0+",
       "about": "The ID of the newly elected leader"},
     {"name": "Voters", "type": "[]Voter", "versions": "0+",
-      "about": "The voters who voted for the current leader"}
+      "about": "The voters who voted at the time of election"},
```
jsancio: Based on the implementation on this PR, this field contains all of the voters in the quorum irrespective of whether they voted or not. In other words I think we should change the description to something like "The set of voters in the quorum for this epoch".

clients/src/main/resources/common/message/LeaderChangeMessage.json
```diff
@@ -22,7 +22,9 @@
...
+    {"name": "GrantingVoters", "type": "[]Voter", "versions": "0+",
+      "about": "The voters who voted for the leader at the time of election"}
```
jsancio: I think we should add an `LeaderEpoch` field that documents the epoch associated with this `LeaderChangeMessage`. What do you think @vamossagar12 and @hachikuji ?

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
```diff
@@ -51,6 +53,7 @@ protected LeaderState(
             boolean hasEndorsedLeader = voterId == localId;
```
jsancio, line 53: I think we have two options here. This class can contain either:
```java
private final Map<Integer, VoterState> voterReplicaStates = new HashMap<>(); 
```
where we change this expression to `boolean hasEndorsedLeader = grantingVoters.contains(voterId);`
and remove the field `private final Set<Integer> grantingVoters = new HashSet<>();`

or:

```java
private final Set<Integer> voters = new HashSet<>(); 
private final Set<Integer> grantingVoters = new HashSet<>(); 
```
and change a few of the methods here use these two sets instead of `voterReplicaStates`.

I like the first option but it is up to you.

jsancio (commented):
Thanks for the additional changes @vamossagar12 . Left a few comments.

clients/src/main/resources/common/message/LeaderChangeMessage.json
vamossagar12: Yeah I think it makes sense to add the LeaderEpoch.

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
vamossagar12: @jsancio , even I would prefer option 1 but slightly confused on it. You suggest to check for endorsement using `boolean hasEndorsedLeader = grantingVoters.contains(voterId);` and then asking to remove the field `grantingVoters `. Am I missing something ?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
vamossagar12: Just wanted to know any particular reason you want this to be a static method?

clients/src/main/resources/common/message/LeaderChangeMessage.json
vamossagar12: Plz let me know if this needs to be added.

vamossagar12:
> Thanks for the additional changes @vamossagar12 . Left a few comments.

Thanks @jsancio , I had a couple of basic questions..

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Just a nitpick. It doesn't need to be a object method as it doesn't use any of the instance fields. When reading and maintaining a lot of code it is good to know that a method doesn't use instance fields.

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
jsancio: Yes, that's correct. You will have to change the implementation of `grantingVoters()` to something like:

```java
public Set<Integer> endorsingVoters() { 
    return voterReplicaStates
        .values()
        .stream()
        .filter(voter -> voter.hasEndorsedLeader)
        .map(voter -> voter.stateId)
        .collect(Collectors.toSet()); 
}
```

Notice the change in the method name. We can also change `nonEndorsingFollowers` to `nonEndorsingVoters` for consistency.

vamossagar12:
@jsancio , made the changes.. plz review when you get a chance.

vamossagar12:
@jsancio , I see some timeout related errors in the CI/CD checks. I am not sure if these are relevant to the changes I have made?

clients/src/main/resources/common/message/LeaderChangeMessage.json
```diff
@@ -22,7 +22,11 @@
     {"name": "LeaderId", "type": "int32", "versions": "0+",
       "about": "The ID of the newly elected leader"},
     {"name": "Voters", "type": "[]Voter", "versions": "0+",
-      "about": "The voters who voted for the current leader"}
+      "about": "The set of voters in the quorum for this epoch"},
+    {"name": "GrantingVoters", "type": "[]Voter", "versions": "0+",
+      "about": "The voters who voted for the leader at the time of election"},
+    {"name": "LeaderEpoch", "type": "int32", "versions": "0+",
```
hachikuji: The reason we didn't have this is because the batch header already has a field for the leader epoch.

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
```diff
@@ -72,11 +73,20 @@ public int epoch() {
         return voterReplicaStates.keySet().stream().filter(id -> id != localId).collect(Collectors.toSet());
     }
 
+    public Set<Integer> endorsingVoters() {
```
hachikuji: I think the terminology may be causing some confusion here. The "granting" voters are those who have voted for the leader and which the leader is using as the basis of its election. The "endorsing" voters are those who have acknowledged the leader. The set of endorsing voters changes over time as the voters convert to followers and begin fetching. We use `nonEndorsingVoters` to know which voters still need to be sent `BeginQuorumEpoch`. Note that it is possible for a voter to have granted a vote, but not yet endorsed the election (because it does not know the vote succeeded). 

We should not conflate these sets, but maybe we can come up with better terms to avoid the confusion. I would stick with "granting" voters as the set of voters who granted the leader's candidacy. However, maybe we can change "endorsing" to "acknowledged"? These are the voters who have acknowledged the election result. What do you think?

vamossagar12: Yeah that makes sense. As you explained, endorsing are those voters who have acknowledged the BeginQuorumEpoch request from the leader, so eventually if the leader is accepted, this set will be all the voters who voted. 
Regarding endorsing v/s acknowledged, can we also add what has been acknowledged? Like, `endoresedQuorumEpoch` or `endorsedLeaderCandidacy` or something like this. Just adding context, might clear the confusion. WDYT @hachikuji , @jsancio

clients/src/main/resources/common/message/LeaderChangeMessage.json
vamossagar12: hmm in that case, this might seem as redundant information then? I am sorry I have not seen the batch header message but is  that information would be handy when we view the LeaderChangeMessage? I felt adding it here might increase the debugging related information for Leadership changes.

jsancio: To me the question that we need to answer is if we want this messages to be self contained. Meaning do we only need to have this message to know the state of the election or leader change. Or do we also need to know where in the log this message was inserted.

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
jsancio: Good to know the subtle difference. "acknowledged" sounds good to me.

clients/src/main/resources/common/message/LeaderChangeMessage.json
vamossagar12: @jsancio , from what I understood, looking at just the LeaderChangeMessage , the epoch won't be available. i looked at the code in `appendControlRecord` which gets invoked for a LeaderChangeMessage but i couldn't find it. The record appends to the log on it's own would have it. That's my understanding. 
With this information that I have, I think it makes sense to still add it unless it isn't agreed upon?

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
vamossagar12: Ok. I will change it to acknowledged then..

vamossagar12: @jsancio , going by all the discussions here, I needed to revert some of the changes you had suggested in https://github.com/apache/kafka/pull/9539#discussion_r529592193 as a couple of test cases were failing. The issue is on this line:

https://github.com/apache/kafka/blob/ce275e29beb48611425708fbe67d28d17329845a/raft/src/main/java/org/apache/kafka/raft/LeaderState.java#L52

Note that, this is a deviation from the original code which was:

`boolean hasEndorsedLeader = voterId == localId;` 

and the test case which was failing because of this was this line in particular:

https://github.com/apache/kafka/pull/9539/files#diff-01c9950503611c357b85ef5b8850d2fbc6060075ce028bf4f86dfc6cbd63fd3cR255

Reason being, when `transitionToLeader` is invoked from `QuorumState` upon becoming the leader, it passes in the set of granting voters. Keeping in mind the explanation provided by @hachikuji above, these are all the voters who voted for the current leader.

If you check the above erroneous condition, we are checking if a voter has endorsed the voting by checking if it is present in the granting voters list:

`boolean hasEndorsedLeader = grantingVoters.contains(voterId);`

So, the problem is, this condition will force all voters to be considered as having endorsed the Leader even before the BeginQuorumEpoch request response dance has begun.

The test cases in question fails because the pollLeader() method in KafkaRaftClient, checks the nonEndorsingVoters and sends the BeginQuorumEpoch request to nonEndorsed ones here:

```
long timeUntilSend = maybeSendRequests(
            currentTimeMs,
            state.nonEndorsingVoters(),
            this::buildBeginQuorumEpochRequest
        );
```

but it never gets to do that because every granting voter has been marked as an endorsed voter.

Lastly, we don't really need the acknowledgedVoters here because all we need are the granting voters in the LeaderChangeMessage i.e the voters who voted for the Leader in the current  epoch. 

So, I am reverting the changes and now, the test cases have all passed as well. There were some timeout related errors as well which I believe which got resolved as well.

clients/src/main/resources/common/message/LeaderChangeMessage.json
vamossagar12: hi @hachikuji , @jsancio , wanted to know if we plan to have the LeaderEpoch field in the LeaderChangeMessage.

hachikuji: I think I'd rather leave it out. Conceptually it would be useful if it was included, but I am not sure what it actually buys us in practice and it introduces the possibility of inconsistency between the epoch in the message and that in the batch header.

vamossagar12: Thanks @hachikuji , I have removed it.

hachikuji (approved):
LGTM. Thanks for the patience on this PR. I think we should merge this and leave the discussed name change for a follow-up. I have filed https://issues.apache.org/jira/browse/KAFKA-10828.

vamossagar12:
Thanks @hachikuji . I have assigned the name change issue to myself. I see 1 test failures but those don't seem to be connected to this issue. 
1 of them is 
``` 
kafka.network.ConnectionQuotasTest > testListenerConnectionRateLimitWhenActualRateAboveLimit FAILED
08:54:04      java.util.concurrent.ExecutionException: java.lang.AssertionError: Expected rate (30 +- 7), but got 39.12107974180088 (600 connections / 15.337 sec) expected:<30.0> but was:<39.12107974180088>
08:54:04          at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
08:54:04          at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:205)
08:54:04          at kafka.network.ConnectionQuotasTest.$anonfun$testListenerConnectionRateLimitWhenActualRateAboveLimit$3(ConnectionQuotasTest.scala:414)
08:54:04          at scala.collection.immutable.List.foreach(List.scala:333)
08:54:04          at kafka.network.ConnectionQuotasTest.testListenerConnectionRateLimitWhenActualRateAboveLimit(ConnectionQuotasTest.scala:414)
08:54:04  
08:54:04          Caused by:
08:54:04          java.lang.AssertionError: Expected rate (30 +- 7), but got 39.12107974180088 (600 connections / 15.337 sec) expected:<30.0> but was:<39.12107974180088>
08:54:04              at org.junit.Assert.fail(Assert.java:89)
08:54:04              at org.junit.Assert.failNotEquals(Assert.java:835)
08:54:04              at org.junit.Assert.assertEquals(Assert.java:555)
08:54:04              at kafka.network.ConnectionQuotasTest.acceptConnectionsAndVerifyRate(ConnectionQuotasTest.scala:871)
08:54:04              at kafka.network.ConnectionQuotasTest.$anonfun$testListenerConnectionRateLimitWhenActualRateAboveLimit$2(ConnectionQuotasTest.scala:412)
08:54:04
``` 

and the other one being:
```
10:12:53  kafka.server.ScramServerStartupTest > testAuthentications FAILED
10:12:53      org.apache.zookeeper.KeeperException$NoAuthException: KeeperErrorCode = NoAuth for /config/changes
10:12:53          at org.apache.zookeeper.KeeperException.create(KeeperException.java:120)
10:12:53          at org.apache.zookeeper.KeeperException.create(KeeperException.java:54)
10:12:53          at kafka.zookeeper.AsyncResponse.maybeThrow(ZooKeeperClient.scala:583)
10:12:53          at kafka.zk.KafkaZkClient.createRecursive(KafkaZkClient.scala:1731)
```

PR #9541, opened 2020-11-02
KAFKA-10675: Add schema name to ConnectSchema.validateValue() error message

Iskuskov:
https://issues.apache.org/jira/browse/KAFKA-10675

The following error message
```java
org.apache.kafka.connect.errors.DataException: Invalid Java object for schema type INT64: class java.lang.Long for field: "moderate_time"
```
can be confusing because `java.lang.Long` is acceptable type for schema `INT64`. 

In fact, in this case `org.apache.kafka.connect.data.Timestamp` is used but this info is not logged.

### DIFF
--- a/connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java
+++ b/connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java
@@ -224,9 +224,7 @@ public class ConnectSchema implements Schema {
         List<Class<?>> expectedClasses = expectedClassesFor(schema);
 
         if (expectedClasses == null)
-            throw new DataException("Invalid Java object for schema type " + schema.type()
-                    + ": " + value.getClass()
-                    + " for field: \"" + name + "\"");
+            throw new DataException(buildNotFoundExpectedClassExceptionMessage(name, schema, value));
 
         boolean foundMatch = false;
         for (Class<?> expectedClass : expectedClasses) {
@@ -237,9 +235,7 @@ public class ConnectSchema implements Schema {
         }
 
         if (!foundMatch)
-            throw new DataException("Invalid Java object for schema type " + schema.type()
-                    + ": " + value.getClass()
-                    + " for field: \"" + name + "\"");
+            throw new DataException(buildNotFoundExpectedClassExceptionMessage(name, schema, value));
 
         switch (schema.type()) {
             case STRUCT:
@@ -268,6 +264,20 @@ public class ConnectSchema implements Schema {
         if (expectedClasses == null)
             expectedClasses = SCHEMA_TYPE_CLASSES.get(schema.type());
         return expectedClasses;
+    }
+
+    private static String buildNotFoundExpectedClassExceptionMessage(String name, Schema schema, Object value) {
+        return String.format("Invalid Java object for %s: %s for field: \"%s\"",
+            buildSchemaInfo(schema), value.getClass(), name);
+    }
+
+    private static String buildSchemaInfo(Schema schema) {
+        StringBuilder schemaInfo = new StringBuilder("schema");
+        if (schema.name() != null) {
+            schemaInfo.append(" \"").append(schema.name()).append("\"");
+        }
+        schemaInfo.append(" with type ").append(schema.type());
+        return schemaInfo.toString();
     }
 
     /**
--- a/connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java
+++ b/connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java
@@ -221,11 +221,7 @@ public class ConnectSchema implements Schema {
             return;
         }
 
-        List<Class<?>> expectedClasses = expectedClassesFor(schema);
-
-        if (expectedClasses == null)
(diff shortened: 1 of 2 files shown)


Iskuskov:
@C0urante, @gharris1727, could you take a look at this please?

connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java
```diff
@@ -266,10 +261,19 @@ public static void validateValue(String name, Schema schema, Object value) {
...
     }
 
+    private static String buildSchemaInfo(Schema schema) {
+        StringBuilder schemaInfo = new StringBuilder("schema");
+        if (schema.name() != null) {
+            schemaInfo.append(" \"").append(schema.name()).append("\"");
+        }
+        schemaInfo.append(" with type ").append(schema.type());
+        return schemaInfo.toString();
+    }
```
C0urante: Given that this is only used in one place, and the string that it creates is itself added to a formatted string, I think it'd be a bit more readable if we removed this method and inlined all the logic where it's being invoked right now.

C0urante (approved):
LGTM, one comment about readability but you can take it or leave it.

I'm also wondering if we might consider dropping the " for field ..." part from the exception message if `field` is null (which occurs with the two-argument variant of `ConnectSchema::validateValue`), but that can be left as follow-up work if you'd prefer.

connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java
Iskuskov: Thanks for reviewing! Fixes were pushed

Iskuskov:
@kkonstantine, could you take a look at this please?

Iskuskov:
@rhauch, could you take a look at this please?

Iskuskov:
@chia7712, @ijuma, could you take a look at this please?

connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java
```diff
@@ -266,7 +267,7 @@ public static void validateValue(String name, Schema schema, Object value) {
     private static List<Class> expectedClassesFor(Schema schema) {
         List<Class> expectedClasses = LOGICAL_TYPE_CLASSES.get(schema.name());
         if (expectedClasses == null)
-            expectedClasses = SCHEMA_TYPE_CLASSES.get(schema.type());
+            expectedClasses = SCHEMA_TYPE_CLASSES.getOrDefault(schema.type(), Collections.emptyList());
```
rhauch, line 270: Strictly speaking, this shouldn't be necessary as `SCHEMA_TYPE_CLASSES` should have a `Schema` instance for all `Schema.Type` literals. And with `SchemaBuilder` a connector or converter cannot create a schema instance with a null `Schema.Type`. 

However, it is possible to construct a `ConnectSchema` instance with a null `Type` reference (like what `FakeSchema` essentially does in the existing test), which of course without this change would result in this method returning a null list. 

So +1 for this line change since it simplifies the error handling in the calling code.

rhauch (approved):
Thanks, @Iskuskov. LGTM.

We're technically still frozen on the `2.6` and `2.8` branches, but both have RCs that passed their PMC votes and should be unblocked soon. I'll hopefully be able to merge and backport soon.

connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java
Iskuskov: Thank you! Highly appreciate your comment

kkonstantine:
@rhauch seems that this fix was approved but we were blocked on release candidates of previous releases. To include to 3.0 (as well as previous releases) and avoid entering a new code freeze phase we'd need to merge asap. Please take another look if possible. 

@Iskuskov in the meantime a merge conflict has been introduced. Would you mind rebasing on top of the latest changes on `trunk`?

Iskuskov:
> @Iskuskov in the meantime a merge conflict has been introduced. Would you mind rebasing on top of the latest changes on `trunk`?

@kkonstantine thank you! Rebased

connect/api/src/main/java/org/apache/kafka/connect/data/ConnectSchema.java
```diff
@@ -221,13 +221,7 @@ public static void validateValue(String name, Schema schema, Object value) {
...
         }
 
-        List<Class<?>> expectedClasses = expectedClassesFor(schema);
-
-        if (expectedClasses == null)
-            throw new DataException("Invalid Java object for schema type " + schema.type()
-                    + ": " + value.getClass()
-                    + " for field: \"" + name + "\"");
-
+        List<Class> expectedClasses = expectedClassesFor(schema);
```
kkonstantine:
```suggestion
        List<Class<?>> expectedClasses = expectedClassesFor(schema);
```

kkonstantine (commented):
We recently enabled warnings on the use of raw types. That's why the build failed after the rebase. 
Adding a suggestion and merging to the PR to allow the tests to run.

kkonstantine (approved):
Test failures are not relevant. 
LGTM

PR #9543, opened 2020-11-02
KAFKA-10500: Makes the Stream thread list resizable

wcarlson5:
Changes the StreamThreads to be held in a ArrayList so that we can dynamically change the number of threads easily.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -145,7 +145,7 @@ public class KafkaStreams implements AutoCloseable {
     private final String clientId;
     private final Metrics metrics;
     private final StreamsConfig config;
-    protected final ArrayList<StreamThread> threads;
+    protected final List<StreamThread> threads;
     private final StateDirectory stateDirectory;
     private final StreamsMetadataState streamsMetadataState;
     private final ScheduledExecutorService stateDirCleaner;
@@ -718,7 +718,7 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         // create the stream thread, global update thread, and cleanup thread
-        threads = new ArrayList<>(numStreamThreads);
+        threads = new LinkedList<>();
         globalTaskTopology = internalTopologyBuilder.buildGlobalStateTopology();
         final boolean hasGlobalTopology = globalTaskTopology != null;
 
@@ -782,12 +782,12 @@ public class KafkaStreams implements AutoCloseable {
                 delegatingStateRestoreListener,
                 i + 1);
             threads.add(i, streamThread);
-            threadState.put(threads.get(i).getId(), threads.get(i).state());
+            threadState.put(streamThread.getId(), streamThread.state());
             storeProviders.add(new StreamThreadStateStoreProvider(threads.get(i)));
         }
 
         ClientMetrics.addNumAliveStreamThreadMetric(streamsMetrics, (metricsConfig, now) ->
-            Math.toIntExact(Arrays.stream(threads.toArray(new StreamThread[numStreamThreads])).filter(thread -> thread.state().isAlive()).count()));
+            Math.toIntExact(threads.stream().filter(thread -> thread.state().isAlive()).count()));
 
         final StreamStateListener streamStateListener = new StreamStateListener(threadState, globalThreadState);
         if (hasGlobalTopology) {
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -718,7 +718,7 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         // create the stream thread, global update thread, and cleanup thread
-        threads = new LinkedList<>();
+        threads = Collections.synchronizedList(new LinkedList<>());
         globalTaskTopology = internalTopologyBuilder.buildGlobalStateTopology();
         final boolean hasGlobalTopology = globalTaskTopology != null;
 
@@ -781,9 +781,9 @@ public class KafkaStreams implements AutoCloseable {
                 stateDirectory,
                 delegatingStateRestoreListener,
                 i + 1);
-            threads.add(i, streamThread);
+            threads.add(streamThread);
             threadState.put(streamThread.getId(), streamThread.state());
-            storeProviders.add(new StreamThreadStateStoreProvider(threads.get(i)));
+            storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
         }
 
         ClientMetrics.addNumAliveStreamThreadMetric(streamsMetrics, (metricsConfig, now) ->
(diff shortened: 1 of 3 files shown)


wcarlson5:
@cadonna First part to make the streamThreads list resizable

wcarlson5:
retest this please

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -145,7 +145,7 @@
     private final String clientId;
     private final Metrics metrics;
     private final StreamsConfig config;
-    protected final StreamThread[] threads;
+    protected final ArrayList<StreamThread> threads;
```
cadonna: I would prefer to use `List` instead of `ArrayList` to be more generic.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -719,8 +718,7 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
         }
 
         // create the stream thread, global update thread, and cleanup thread
-        threads = new StreamThread[numStreamThreads];
-
+        threads = new ArrayList<>(numStreamThreads);
```
cadonna: I think, it is better to keep the default initial capacity of an `ArrayList`. Otherwise, the first time a stream thread is added, we immediately run into a memory allocation. Since we do not know how many stream thread we might expect, let's use the default.   
We could also consider using a `LinkedList` since we never access by index in production code.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -783,12 +781,13 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
                 stateDirectory,
                 delegatingStateRestoreListener,
                 i + 1);
-            threadState.put(threads[i].getId(), threads[i].state());
-            storeProviders.add(new StreamThreadStateStoreProvider(threads[i]));
+            threads.add(i, streamThread);
+            threadState.put(threads.get(i).getId(), threads.get(i).state());
+            storeProviders.add(new StreamThreadStateStoreProvider(threads.get(i)));
```
cadonna: You can simplify to
```suggestion
            threads.add(streamThread);
            threadState.put(streamThread.getId(), streamThread.state());
            storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -783,12 +781,13 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
         }
 
         ClientMetrics.addNumAliveStreamThreadMetric(streamsMetrics, (metricsConfig, now) ->
-            Math.toIntExact(Arrays.stream(threads).filter(thread -> thread.state().isAlive()).count()));
+            Math.toIntExact(Arrays.stream(threads.toArray(new StreamThread[numStreamThreads])).filter(thread -> thread.state().isAlive()).count()));
```
cadonna: Please simplify to
```suggestion
            Math.toIntExact(threads.stream().filter(thread -> thread.state().isAlive()).count()));
```

cadonna (commented):
@wcarlson5 Thank you for the PR!

Here my feedback!

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: That is fair

wcarlson5:
@cadonna Looks like all those changes made sense

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -719,8 +718,7 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
         }
 
         // create the stream thread, global update thread, and cleanup thread
-        threads = new StreamThread[numStreamThreads];
-
+        threads = new LinkedList<>();
```
vvcephei: Should this collection be threadsafe? (or are all accesses inside synchronized blocks anyway?)

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -783,12 +781,13 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
                 stateDirectory,
                 delegatingStateRestoreListener,
                 i + 1);
-            threadState.put(threads[i].getId(), threads[i].state());
-            storeProviders.add(new StreamThreadStateStoreProvider(threads[i]));
+            threads.add(i, streamThread);
```
vvcephei: A bit of a nitpick, but this operation is O(n) for LinkedList. Better to just `add(streamThread)` if you want to use LinkedList.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -783,12 +781,13 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+            threadState.put(streamThread.getId(), streamThread.state());
+            storeProviders.add(new StreamThreadStateStoreProvider(threads.get(i)));
```
vvcephei:
```suggestion
            storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
```

`get(i)` is also O(n) for a linked list. Again, this is admittedly a nit, since the list is small.

vvcephei (commented):
Thanks for the PR! Just a few remarks.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: They are not all in synchronized blocks so I think it probably should be thread safe

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: yep, should have changed that when I moved from ArrayList

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Good catch

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
cadonna: Currently only the constructor adds elements to the list. All other accesses afterwards are read-only. So, we would not need a synchronized list as far as I see. When we implement the add and remove streams thread APIs, we probably need synchronization. My proposal is to leave it a synchronized list now just in case we forget to think about it afterwards and then to reconsider how we synchronize the accesses.

cadonna (approved):
Thanks, @wcarlson5 !

LGTM!

vvcephei (approved):
Thanks for the update!

vvcephei:
The test failures look unrelated:
```

Build / JDK 8 / org.apache.kafka.connect.mirror.MirrorConnectorsIntegrationTest.testOneWayReplicationWithAutoOffsetSync | 34 sec | 1
-- | -- | --
Build / JDK 11 / kafka.api.PlaintextConsumerTest.testLowMaxFetchSizeForRequestAndPartition | 2 min 3 sec | 1
Build / JDK 11 / kafka.network.DynamicConnectionQuotaTest.testDynamicListenerConnectionCreationRateQuota | 59 sec | 1
```

PR #9545, opened 2020-11-03
[mm2] Allow Checkpoints for consumers using static partition assignments

scanterog:
Currently mm2 is not generating checkpoints for upstream consumers using static partition assignments. The reason is that these consumers are being explicitly filtered out. I couldn't find why this filter is being applied. This PR removes this limitation.

I've tested this change (but only in one direction: source->target) and it does not break any functionality. Even KAFKA-9076 continues working fine and allows consumer offset sync for consumers using static assignments.

### DIFF
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorCheckpointConnector.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorCheckpointConnector.java
@@ -145,7 +145,7 @@ public class MirrorCheckpointConnector extends SourceConnector {
         knownConsumerGroups = findConsumerGroups();
     }
 
-    private List<String> findConsumerGroups()
+    List<String> findConsumerGroups()
             throws InterruptedException, ExecutionException {
         return listConsumerGroups().stream()
                 .map(x -> x.groupId())
@@ -153,7 +153,7 @@ public class MirrorCheckpointConnector extends SourceConnector {
                 .collect(Collectors.toList());
     }
 
-    private Collection<ConsumerGroupListing> listConsumerGroups()
+    Collection<ConsumerGroupListing> listConsumerGroups()
             throws InterruptedException, ExecutionException {
         return sourceAdminClient.listConsumerGroups().valid().get();
     }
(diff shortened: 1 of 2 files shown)


scanterog:
cc: @ryannedolan who added this filter and might have more context on why this was needed.
cc: @mimaison for more eyes on this :)

Update: tests are working fine locally. Might want to rerun it here.

scanterog:
@ryannedolan friendly ping. Would love your feedback on this one.

mimaison:
Groups created on the target cluster by KAFKA-9076 are "simple groups" as there's no member information.
Not entirely sure why these were explicitly filtered. I can't immediately come up with a reason.

@scanterog Can you add a test as well?

scanterog:
> Groups created on the target cluster by KAFKA-9076 are "simple groups" as there's no member information.
> Not entirely sure why these were explicitly filtered. I can't immediately come up with a reason.
> 
> @scanterog Can you add a test as well?

Sure. Do we want a test for `findConsumerGroups` or what are you looking to test?

scanterog:
@mimaison Not sure what we can test here. It is basically just allowing consumers with no members information (generally the case of consumer with static partition assignments) when discovering the available consumers. Happy to test what needs to be tested but I personally don't think we need a unit test for this. Happy to read on @ryannedolan's feedback too.

mimaison:
@scanterog 
This PR changes the way the Checkpoint connector works. We need a test to exercise the new behaviour.
Something simple like this would do:
```java
@Test
public void testFindConsumerGroups() throws Exception {
    MirrorConnectorConfig config = new MirrorConnectorConfig(makeProps());
    MirrorCheckpointConnector connector = new MirrorCheckpointConnector(Collections.emptyList(), config);
    connector = spy(connector);

    Collection<ConsumerGroupListing> groups = Arrays.asList(
            new ConsumerGroupListing("g1", true),
            new ConsumerGroupListing("g2", false));
    doReturn(groups).when(connector).listConsumerGroups();
    doReturn(true).when(connector).shouldReplicate(anyString());
    List<String> groupFound = connector.findConsumerGroups();

    Set<String> expectedGroups = groups.stream().map(g -> g.groupId()).collect(Collectors.toSet());
    assertEquals(expectedGroups, new HashSet<>(groupFound));
}
```
On `trunk` this test fails. But with your PR it passes.

mimaison (approved):

ryannedolan:
Sorry, late to the convo, but lgtm, thanks!

PR #9546, opened 2020-11-03
MINOR: optimeze PartitionStateMachine handleStateChanges

dengziming:
The logic of state change to OfflinePartition or NonExistentPartition are identical, so just merge them.
If fact, the logic of the case NewPartition is also the same, but have different log format, so ignore.


dengziming:
@chia7712 , Hi, PTLA.

chia7712:
@dengziming Could you rebase PR to trigger QA again?

chia7712 (approved):
@dengziming LGTM. Thanks for your patch!

chia7712:
Merge to trunk

PR #9547, opened 2020-11-03
KAFKA-9630; Replace OffsetsForLeaderEpoch request/response with automated protocol

dajac:
This PR migrates the `OffsetsForLeaderEpoch` request/response to the automated protocol. It also refactors the `OffsetsForLeaderEpochClient` to use directly the internal structs generated by the automated protocol. It relies on the existing tests.

It seems that we could also refactor the broker (api layer, fetcher) to directly use the internal structs of the generated protocol but, as it is more involved, I propose to address it in a separate PR.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/OffsetsForLeaderEpochClient.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/OffsetsForLeaderEpochClient.java
@@ -19,7 +19,11 @@ package org.apache.kafka.clients.consumer.internals;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.TopicAuthorizationException;
+import org.apache.kafka.common.message.OffsetForLeaderEpochRequestData.OffsetForLeaderPartition;
+import org.apache.kafka.common.message.OffsetForLeaderEpochRequestData.OffsetForLeaderTopic;
+import org.apache.kafka.common.message.OffsetForLeaderEpochRequestData.OffsetForLeaderTopicCollection;
 import org.apache.kafka.common.protocol.Errors;
+import org.apache.kafka.common.record.RecordBatch;
 import org.apache.kafka.common.requests.AbstractRequest;
 import org.apache.kafka.common.requests.EpochEndOffset;
 import org.apache.kafka.common.requests.OffsetsForLeaderEpochRequest;
@@ -47,12 +51,23 @@ public class OffsetsForLeaderEpochClient extends AsyncClient<
     @Override
     protected AbstractRequest.Builder<OffsetsForLeaderEpochRequest> prepareRequest(
             Node node, Map<TopicPartition, SubscriptionState.FetchPosition> requestData) {
-        Map<TopicPartition, OffsetsForLeaderEpochRequest.PartitionData> partitionData = new HashMap<>(requestData.size());
-        requestData.forEach((topicPartition, fetchPosition) -> fetchPosition.offsetEpoch.ifPresent(
-            fetchEpoch -> partitionData.put(topicPartition,
-                new OffsetsForLeaderEpochRequest.PartitionData(fetchPosition.currentLeader.epoch, fetchEpoch))));
-
-        return OffsetsForLeaderEpochRequest.Builder.forConsumer(partitionData);
+        OffsetForLeaderTopicCollection topics = new OffsetForLeaderTopicCollection(requestData.size());
+        requestData.forEach((topicPartition, fetchPosition) ->
+            fetchPosition.offsetEpoch.ifPresent(fetchEpoch -> {
+                OffsetForLeaderTopic topic = topics.find(topicPartition.topic());
+                if (topic == null) {
+                    topic = new OffsetForLeaderTopic().setName(topicPartition.topic());
+                    topics.add(topic);
+                }
+                topic.partitions().add(new OffsetForLeaderPartition()
+                    .setPartitionIndex(topicPartition.partition())
+                    .setLeaderEpoch(fetchEpoch)
+                    .setCurrentLeaderEpoch(fetchPosition.currentLeader.epoch
+                        .orElse(RecordBatch.NO_PARTITION_LEADER_EPOCH))
+                );
+            })
+        );
+        return OffsetsForLeaderEpochRequest.Builder.forConsumer(topics);
     }
 
     @Override
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/OffsetsForLeaderEpochClient.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/OffsetsForLeaderEpochClient.java
@@ -27,7 +27,6 @@ import org.apache.kafka.common.message.OffsetForLeaderEpochResponseData.OffsetFo
 import org.apache.kafka.common.protocol.Errors;
 import org.apache.kafka.common.record.RecordBatch;
 import org.apache.kafka.common.requests.AbstractRequest;
-import org.apache.kafka.common.requests.EpochEndOffset;
 import org.apache.kafka.common.requests.OffsetsForLeaderEpochRequest;
 import org.apache.kafka.common.requests.OffsetsForLeaderEpochResponse;
 import org.apache.kafka.common.utils.LogContext;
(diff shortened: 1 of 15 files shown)


clients/src/main/resources/common/message/OffsetForLeaderEpochRequest.json
```diff
@@ -32,13 +32,13 @@
     { "name": "Topics", "type": "[]OffsetForLeaderTopic", "versions": "0+",
       "about": "Each topic to get offsets for.", "fields": [
       { "name": "Name", "type": "string", "versions": "0+", "entityType": "topicName",
```
chia7712: Previous "name" is ```topic``` (https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/protocol/CommonFields.java#L25) and the "name" in auto-generated protocol is ```name```. Does it break the compatibility? For example, a cluster mixed by auto-generated protocol and stale protocol.

chia7712: Please ignore above comment. I misunderstood the protocol :(

clients/src/main/java/org/apache/kafka/clients/consumer/internals/OffsetsForLeaderEpochClient.java
```diff
@@ -61,67 +77,76 @@ protected OffsetForEpochResult handleResponse(
             Map<TopicPartition, SubscriptionState.FetchPosition> requestData,
             OffsetsForLeaderEpochResponse response) {
 
+        Set<TopicPartition> missingPartitions = new HashSet<>(requestData.keySet());
```
hachikuji: Wonder if it might be simpler to initialize `partitionsToRetry` from the request key set.

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochRequest.java
```diff
@@ -51,169 +47,120 @@
...
         }
 
         public static Builder forFollower(short version, Map<TopicPartition, PartitionData> epochsByPartition, int replicaId) {
-            return new Builder(version, version, epochsByPartition, replicaId);
+            OffsetForLeaderEpochRequestData data = new OffsetForLeaderEpochRequestData();
+            data.setReplicaId(replicaId);
+
+            epochsByPartition.forEach((partitionKey, partitionValue) -> {
+                OffsetForLeaderTopic topic = data.topics().find(partitionKey.topic());
+                if (topic == null) {
```
hachikuji, line 77: We have similar logic in `OffsetsForLeaderEpochClient.prepareRequest`. Wonder if we should push it to the Builder?

hachikuji (commented):
Looks good, just a couple comments.

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochRequest.java
dajac: That's a good question. Actually, I was hoping to push this one out of the builder while removing `PartitionData` in profit of using the internal data structure of the auto-generated protocol. I'd like to do this in a follow-up PR.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/OffsetsForLeaderEpochClient.java
dajac: The downside of relying on `partitionsToRetry` is that we can't differentiate the partitions to be retried due to an error from the missing partitions in the response. It may not be that important but we can't preserve the above log otherwise.
```
logger().warn("Missing partition {} from response, retrying.", topicPartition);
```
I suppose that this log entry is there for a reason so I went with preserving it. What do you think?

hachikuji: No strong opinion. It's an unlikely case anyway, so I'm not sure it calls for special treatment.

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochRequest.java
hachikuji: Sounds fine. Do you have a jira?

hachikuji (approved):
LGTM.

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochRequest.java
```diff
@@ -51,169 +47,120 @@
...
+                        partition.leaderEpoch()))));
+
+        return epochsByTopicPartition;
+    }
+
+    public int replicaId() {
+        return data.replicaId();
+    }
+
+    public static OffsetsForLeaderEpochRequest parse(ByteBuffer buffer, short version) {
```
chia7712, line 137: not sure why this kind of method still exist in each request. It is not used anymore.

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochResponse.java
```diff
@@ -51,133 +41,82 @@
...
+                        partition.endOffset()))));
+
         return epochEndOffsetsByPartition;
     }
 
     @Override
     public Map<Errors, Integer> errorCounts() {
         Map<Errors, Integer> errorCounts = new HashMap<>();
-        epochEndOffsetsByPartition.values().forEach(response ->
+        responses().values().forEach(response ->
```
chia7712: Is it worth using ```data``` rather than ```responses()``` to avoid extra conversion?

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochResponse.java
```diff
@@ -51,133 +41,82 @@
...
     }
 
-    public OffsetsForLeaderEpochResponse(int throttleTimeMs, Map<TopicPartition, EpochEndOffset> epochsByTopic) {
-        this.throttleTimeMs = throttleTimeMs;
-        this.epochEndOffsetsByPartition = epochsByTopic;
+    public OffsetForLeaderEpochResponseData data() {
+        return data;
     }
 
     public Map<TopicPartition, EpochEndOffset> responses() {
```
chia7712, line 81: I hope we can get rid of those conversion in the future :)

chia7712 (approved):
I like this great patch. +1

some trivial comments are left. please take a look :)

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochRequest.java
dajac: Good question. As we have kept them in other requests/responses, I will keep it for now. We should consider removing all of them.

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochResponse.java
dajac: I will address this in a follow-up PR.

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochRequest.java
dajac: * https://issues.apache.org/jira/browse/KAFKA-10740
* https://issues.apache.org/jira/browse/KAFKA-10739

clients/src/main/java/org/apache/kafka/clients/consumer/internals/OffsetsForLeaderEpochClient.java
dajac: Yeah, you're right. Let me remove it.

dajac:
@chia7712 @hachikuji Thanks for your comments. I have addressed them. Could you have another quick look?

clients/src/main/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochRequest.java
hachikuji: I think these are still used in `RequestResponseTest`. There's probably a lot of cruft like this that we can start cleaning up though.

hachikuji:
@dajac The updates LGTM.

chia7712:
@dajac +1 to nice updating. sorry that my commit causes conflicting files to you.

dajac:
Failed test seems unrelated to this PR. Merging to trunk.

PR #9549, opened 2020-11-03
KIP-145: Add SMTs, HeaderFrom, DropHeaders and InsertHeader

tombentley:
These SMTs were originally specified in [KIP-145](https://cwiki.apache.org/confluence/display/KAFKA/KIP-145+-+Expose+Record+Headers+in+Kafka+Connect#KIP145ExposeRecordHeadersinKafkaConnect-Transformations) but never implemented at the time.

`HeaderTo` is not included since its original specification doesn't deal with the fact that there can be >1 header with the same name, but a field can only have a single value (while you could use an array, that doesn't work if the headers for the given name had different schemas).

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/tools/TransformationDoc.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/tools/TransformationDoc.java
@@ -18,11 +18,15 @@ package org.apache.kafka.connect.tools;
 
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.connect.transforms.Cast;
+import org.apache.kafka.connect.transforms.DropHeaders;
 import org.apache.kafka.connect.transforms.ExtractField;
 import org.apache.kafka.connect.transforms.Filter;
 import org.apache.kafka.connect.transforms.Flatten;
+import org.apache.kafka.connect.transforms.HeaderFrom;
+import org.apache.kafka.connect.transforms.HeaderTo;
 import org.apache.kafka.connect.transforms.HoistField;
 import org.apache.kafka.connect.transforms.InsertField;
+import org.apache.kafka.connect.transforms.InsertHeader;
 import org.apache.kafka.connect.transforms.MaskField;
 import org.apache.kafka.connect.transforms.RegexRouter;
 import org.apache.kafka.connect.transforms.ReplaceField;
@@ -62,7 +66,10 @@ public class TransformationDoc {
             new DocInfo(Flatten.class.getName(), Flatten.OVERVIEW_DOC, Flatten.CONFIG_DEF),
             new DocInfo(Cast.class.getName(), Cast.OVERVIEW_DOC, Cast.CONFIG_DEF),
             new DocInfo(TimestampConverter.class.getName(), TimestampConverter.OVERVIEW_DOC, TimestampConverter.CONFIG_DEF),
-            new DocInfo(Filter.class.getName(), Filter.OVERVIEW_DOC, Filter.CONFIG_DEF)
+            new DocInfo(Filter.class.getName(), Filter.OVERVIEW_DOC, Filter.CONFIG_DEF),
+            new DocInfo(InsertHeader.class.getName(), InsertHeader.OVERVIEW_DOC, InsertHeader.CONFIG_DEF),
+            new DocInfo(DropHeaders.class.getName(), DropHeaders.OVERVIEW_DOC, DropHeaders.CONFIG_DEF),
+            new DocInfo(HeaderFrom.class.getName(), HeaderFrom.OVERVIEW_DOC, HeaderFrom.CONFIG_DEF)
     );
 
     private static void printTransformationHtml(PrintStream out, DocInfo docInfo) {
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/tools/TransformationDoc.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/tools/TransformationDoc.java
@@ -23,7 +23,6 @@ import org.apache.kafka.connect.transforms.ExtractField;
 import org.apache.kafka.connect.transforms.Filter;
 import org.apache.kafka.connect.transforms.Flatten;
 import org.apache.kafka.connect.transforms.HeaderFrom;
-import org.apache.kafka.connect.transforms.HeaderTo;
 import org.apache.kafka.connect.transforms.HoistField;
 import org.apache.kafka.connect.transforms.InsertField;
 import org.apache.kafka.connect.transforms.InsertHeader;
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
@@ -29,7 +29,7 @@ public class DropHeaders<R extends ConnectRecord<R>> implements Transformation<R
     public static final String OVERVIEW_DOC =
             "Removes one or more headers from each record.";
 
-    public static final String HEADERS_FIELD = "header.names";
+    public static final String HEADERS_FIELD = "headers";
 
     public static final ConfigDef CONFIG_DEF = new ConfigDef()
             .define(HEADERS_FIELD, ConfigDef.Type.LIST, ConfigDef.Importance.HIGH,
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.connect.transforms;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.connector.ConnectRecord;
+import org.apache.kafka.connect.header.Headers;
+import org.apache.kafka.connect.transforms.util.SimpleConfig;
+
+import java.util.List;
+import java.util.Map;
+
+public class DropHeaders<R extends ConnectRecord<R>> implements Transformation<R> {
+
+    public static final String OVERVIEW_DOC =
+            "Removes one or more headers from each record.";
+
+    public static final String HEADERS_FIELD = "header.names";
+
+    public static final ConfigDef CONFIG_DEF = new ConfigDef()
+            .define(HEADERS_FIELD, ConfigDef.Type.LIST, ConfigDef.Importance.HIGH,
+                    "The name of the headers to be removed.");
+
+    private List<String> headers;
+
+    @Override
+    public R apply(R record) {
+        Headers updatedHeaders = record.headers().duplicate();
+        for (String name : headers) {
+            updatedHeaders.remove(name);
+        }
+        return record.newRecord(record.topic(), record.kafkaPartition(), record.keySchema(), record.key(),
+                record.valueSchema(), record.value(), record.timestamp(), updatedHeaders);
+    }
+
+
+    @Override
+    public ConfigDef config() {
+        return CONFIG_DEF;
+    }
+
+    @Override
+    public void close() {
+
+    }
+
+    @Override
+    public void configure(Map<String, ?> props) {
+        final SimpleConfig config = new SimpleConfig(CONFIG_DEF, props);
+        headers = config.getList(HEADERS_FIELD);
+    }
+}
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
@@ -18,11 +18,17 @@ package org.apache.kafka.connect.transforms;
 
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.connect.connector.ConnectRecord;
+import org.apache.kafka.connect.header.ConnectHeaders;
+import org.apache.kafka.connect.header.Header;
 import org.apache.kafka.connect.header.Headers;
+import org.apache.kafka.connect.transforms.util.NonEmptyListValidator;
 import org.apache.kafka.connect.transforms.util.SimpleConfig;
 
-import java.util.List;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
+
+import static org.apache.kafka.common.config.ConfigDef.NO_DEFAULT_VALUE;
 
 public class DropHeaders<R extends ConnectRecord<R>> implements Transformation<R> {
 
@@ -32,21 +38,24 @@ public class DropHeaders<R extends ConnectRecord<R>> implements Transformation<R
     public static final String HEADERS_FIELD = "headers";
 
     public static final ConfigDef CONFIG_DEF = new ConfigDef()
-            .define(HEADERS_FIELD, ConfigDef.Type.LIST, ConfigDef.Importance.HIGH,
+            .define(HEADERS_FIELD, ConfigDef.Type.LIST,
+                    NO_DEFAULT_VALUE, new NonEmptyListValidator(),
+                    ConfigDef.Importance.HIGH,
                     "The name of the headers to be removed.");
 
-    private List<String> headers;
+    private Set<String> headers;
 
     @Override
     public R apply(R record) {
-        Headers updatedHeaders = record.headers().duplicate();
-        for (String name : headers) {
-            updatedHeaders.remove(name);
+        Headers updatedHeaders = new ConnectHeaders();
+        for (Header header : record.headers()) {
+            if (!headers.contains(header.key())) {
+                updatedHeaders.add(header);
+            }
         }
         return record.newRecord(record.topic(), record.kafkaPartition(), record.keySchema(), record.key(),
                 record.valueSchema(), record.value(), record.timestamp(), updatedHeaders);
     }
-
 
     @Override
     public ConfigDef config() {
@@ -55,12 +64,11 @@ public class DropHeaders<R extends ConnectRecord<R>> implements Transformation<R
 
     @Override
     public void close() {
-
     }
 
     @Override
     public void configure(Map<String, ?> props) {
         final SimpleConfig config = new SimpleConfig(CONFIG_DEF, props);
-        headers = config.getList(HEADERS_FIELD);
+        headers = new HashSet<>(config.getList(HEADERS_FIELD));
     }
 }
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
@@ -28,6 +28,7 @@ import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.header.Header;
 import org.apache.kafka.connect.header.Headers;
+import org.apache.kafka.connect.transforms.util.NonEmptyListValidator;
 import org.apache.kafka.connect.transforms.util.Requirements;
 import org.apache.kafka.connect.transforms.util.SchemaUtil;
 import org.apache.kafka.connect.transforms.util.SimpleConfig;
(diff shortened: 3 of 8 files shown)


tombentley:
@kkonstantine would you be able to review this? These SMTs were originally specified in [KIP145](https://cwiki.apache.org/confluence/display/KAFKA/KIP-145+-+Expose+Record+Headers+in+Kafka+Connect#KIP145ExposeRecordHeadersinKafkaConnect-HeaderFrom) but never implemented. It seems they were forgotten about.

@C0urante maybe you'd also like to review?

C0urante:
Thanks for reaching out @tombentley! Happy to take a look.

For reference, maybe you could add a link to https://cwiki.apache.org/confluence/display/KAFKA/KIP-145+-+Expose+Record+Headers+in+Kafka+Connect#KIP145ExposeRecordHeadersinKafkaConnect-Transformations to the description?

Will begin reviewing shortly.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
```diff
@@ -0,0 +1,219 @@
...
+
+    public static final ConfigDef CONFIG_DEF = new ConfigDef()
+            .define(FIELDS_FIELD, ConfigDef.Type.LIST, ConfigDef.Importance.HIGH,
+                    "Field names in the record whose values are to be copied or moved to headers.")
+            .define(HEADERS_FIELD, ConfigDef.Type.LIST, ConfigDef.Importance.HIGH,
+                    "Header names, in the same order as the field names listed in the fields configuration property.")
+            .define(OPERATION_FIELD, ConfigDef.Type.STRING, NO_DEFAULT_VALUE,
+                    ConfigDef.ValidString.in("move", "copy"), ConfigDef.Importance.HIGH,
+                    "Either <code>move</code> if the fields are to be moved to the headers (removed from the key/value), " +
+                            "or <code>copy</code> if the fields are to be copied to the headers (retained in the key/value).");
```
C0urante: Might make sense to refer to the `MOVE.name` and `COPY.name` fields declared in the `Operation` enum below instead of using the literal `"move"` and `"copy"` strings in this section.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
```diff
@@ -0,0 +1,219 @@
...
+
+        if (operatingSchema == null) {
+            return applySchemaless(record, operatingValue);
+        } else {
+            return applyWithSchema(record, operatingValue, operatingSchema);
+        }
+    }
+
+    private R applyWithSchema(R record, Object operatingValue, Schema operatingSchema) {
+        Headers updatedHeaders = record.headers().duplicate();
```
C0urante, line 120: Why duplicate headers here? According to the [Header class's Javadocs](https://kafka.apache.org/26/javadoc/org/apache/kafka/connect/header/Headers.html), the collection should be mutable.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
```diff
@@ -0,0 +1,219 @@
...
+        final Schema updatedSchema;
+        final SchemaBuilder builder = SchemaUtil.copySchemaBasics(operatingSchema, SchemaBuilder.struct());
+        for (Field field : operatingSchema.fields()) {
+            if (!fields.contains(field.name())) {
+                builder.field(field.name(), field.schema());
+            }
+        }
+        updatedSchema = builder.build();
+        return updatedSchema;
+    }
```
C0urante: This looks fairly expensive to perform for every record. Do you think it might make sense to perform some caching, similarly to what's done in the [ReplaceField](https://github.com/apache/kafka/blob/d61dc0c1832935ae680388dcb8c12d1250dece33/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/ReplaceField.java#L169-L173) transform?

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
```diff
@@ -0,0 +1,219 @@
...
+
+    }
+
+    @Override
+    public void configure(Map<String, ?> props) {
+        final SimpleConfig config = new SimpleConfig(CONFIG_DEF, props);
+        fields = config.getList(FIELDS_FIELD);
+        headers = config.getList(HEADERS_FIELD);
+        if (headers.size() != fields.size()) {
+            throw new ConfigException("'fields' config must have the same number of elements as 'headers' config.");
```
C0urante: Probably want to refer to `FIELDS_FIELD` and `HEADERS_FIELD` here instead of the literal `"fields"` and `"headers"` strings.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
```diff
@@ -0,0 +1,219 @@
...
+            updatedSchema = operatingSchema;
+        }
+        final Struct updatedValue = new Struct(updatedSchema);
+        for (Field field : updatedSchema.fields()) {
+            updatedValue.put(field, value.get(field.name()));
+        }
+        for (int i = 0; i < fields.size(); i++) {
+            String fieldName = fields.get(i);
+            String headerName = headers.get(i);
+            Object fieldValue = value.get(fieldName);
```
C0urante: Worth noting that if there's no field with this name in the schema for the record, this will throw an exception and (unless it's configured with an error tolerance to ignore such issues), fail the task.

This isn't unreasonable behavior, especially given that that case isn't covered in the KIP, but it's different from what happens right now in the `applySchemaless` method, which is to silently add a null header instead.

It's probably best to ensure the same behavior happens in either case. I don't have a strong preference either way but I'm tentatively leaning towards silently adding `null` headers since it'll make working with heterogeneous topics easier.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
```diff
@@ -0,0 +1,219 @@
...
+        final Schema updatedSchema;
+        if (operation == Operation.MOVE) {
+            updatedSchema = moveSchema(operatingSchema);
+        } else {
+            updatedSchema = operatingSchema;
+        }
+        final Struct updatedValue = new Struct(updatedSchema);
+        for (Field field : updatedSchema.fields()) {
+            updatedValue.put(field, value.get(field.name()));
+        }
```
C0urante: Does this mean we'd be iterating over every field of every record? Not very performant 🙁 

I think we can improve the efficiency here for the `COPY` case, if not the `MOVE` case, by reusing the existing value and not creating a new one.

In the `MOVE` case though, it might be the best we have to settle for until/unless there's a richer API added for mutating Connect records and their keys, values, schemas, etc.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+public class DropHeaders<R extends ConnectRecord<R>> implements Transformation<R> {
+
+    public static final String OVERVIEW_DOC =
+            "Removes one or more headers from each record.";
+
+    public static final String HEADERS_FIELD = "header.names";
```
C0urante: Looks like this has two different names in the KIP; it's referenced as `headers` in the textual description but as `header.names` in the sample config.

I think `headers` makes more sense here given that it would align with the properties defined in the `InsertHeader` and `HeaderFrom` transformations.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public static final ConfigDef CONFIG_DEF = new ConfigDef()
+            .define(HEADERS_FIELD, ConfigDef.Type.LIST, ConfigDef.Importance.HIGH,
+                    "The name of the headers to be removed.");
+
+    private List<String> headers;
+
+    @Override
+    public R apply(R record) {
+        Headers updatedHeaders = record.headers().duplicate();
```
C0urante: Same question here RE: duplication

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/InsertHeader.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public static final ConfigDef CONFIG_DEF = new ConfigDef()
+            .define(HEADER_FIELD, ConfigDef.Type.STRING, ConfigDef.Importance.HIGH,
+                    "The name of the header.")
+            .define(VALUE_LITERAL_FIELD, ConfigDef.Type.STRING, ConfigDef.Importance.HIGH,
+                    "The literal value that is to be set as the header value on all records.");
+
+    private String header;
+
+    private String literalValue;
```
C0urante: I think this is supposed to be parsed using the `Values` class, which would allow users to specify integral, floating point, boolean, and even structured and nested types here that would then be correctly picked up by the framework and used in the resulting header.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/InsertHeader.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public R apply(R record) {
+        Headers updatedHeaders = record.headers().duplicate();
```
C0urante, line 54: Same question here RE: duplication

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/InsertHeader.java
```diff
@@ -0,0 +1,70 @@
...
+        updatedHeaders.add(header, literalValue, Schema.STRING_SCHEMA);
```
C0urante: If we use the `Values` class to parse the value literal, we can use the schema that it provides instead of hardcoding this to use a string schema.

C0urante (commented):
Looking pretty good! Will check out the tests after the core functional changes are settled.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
tombentley: `MOVE.name` is not a compile time constant (at least the compiler doesn't see it that way). I factored out constants to avoid the duplication of literals.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
tombentley: I wasn't completely sure whether a transformation was allowed to mutate headers, since it has to create a new `ConnectRecord`. Take this code from `WorkerSourceTask` as an example:

```java
            final SourceRecord record = transformationChain.apply(preTransformRecord);
            final ProducerRecord<byte[], byte[]> producerRecord = convertTransformedRecord(record);
            if (producerRecord == null || retryWithToleranceOperator.failed()) {
                counter.skipRecord();
                commitTaskRecord(preTransformRecord, null);
                continue;
            }
```

See how `preTransformRecord` can be used after the transformation chain has been applied? I think it would be incorrect for `commitTaskRecord` to commit the original record but with headers which had been mutated by transformations, right?

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
tombentley: Good idea.

tombentley:
@C0urante thanks for the review, some really helpful comments there.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
C0urante: Ah, that's fair. Does seem to be the pattern followed by the other out-of-the-box transformations as well; probably best to continue to follow that pattern.

I'm a little unnerved by this though, since as far as I can tell it's not publicly documented and so it's possible people writing their own transformations may be violating this implicit rule.

Out of scope, so I've filed [KAFKA-10720](https://issues.apache.org/jira/browse/KAFKA-10720) to track the need for possible documentation improvements.

connect/transforms/src/test/java/org/apache/kafka/connect/transforms/HeaderFromTest.java
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    @Test(expected = ConfigException.class)
+    public void invalidConfig() {
+        Map<String, Object> config = config();
+        List<String> headers = new ArrayList<>(this.headers);
+        headers.add("unexpected");
+        config.put(HeaderFrom.HEADERS_FIELD, headers);
+        xform.configure(config);
+    }
```
C0urante: Little silly to run this for every iteration of the parameterized test 😛 
I'm guessing there isn't an easy way to run this only once?

C0urante (approved):
LGTM, a few small comments but none of them are blockers. Thanks Tom!

connect/transforms/src/test/java/org/apache/kafka/connect/transforms/HeaderFromTest.java
tombentley: Yeah, parameterized tests do have their faults. AFAIK the only way to handle this using Junit 4 is to have two tests. You can arrange to have both tests in the same source file so both parameterised and non-parameterised tests are still easily discoverable (e.g. [using Enclosed](https://stackoverflow.com/questions/3361239/excluding-a-non-param-test-in-parameterized-test-class)), but I've never seen that done in practice and it looks rather fiddly. Since these tests execute pretty quickly, I'm inclined to think we should just live with it. The real solution might just be to adopt Junit 5.

tombentley:
@kkonstantine or perhaps @rhauch  please could one of you take a look?

tombentley:
@kkonstantine, @rhauch did you get a chance to look at this?

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/InsertHeader.java
```diff
@@ -0,0 +1,72 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.connect.transforms;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.connector.ConnectRecord;
+import org.apache.kafka.connect.data.Schema;
```
mimaison: Unused import

connect/runtime/src/main/java/org/apache/kafka/connect/tools/TransformationDoc.java
```diff
@@ -18,11 +18,15 @@
 
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.connect.transforms.Cast;
+import org.apache.kafka.connect.transforms.DropHeaders;
 import org.apache.kafka.connect.transforms.ExtractField;
 import org.apache.kafka.connect.transforms.Filter;
 import org.apache.kafka.connect.transforms.Flatten;
+import org.apache.kafka.connect.transforms.HeaderFrom;
+import org.apache.kafka.connect.transforms.HeaderTo;
```
mimaison: We don't have this transformation!

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
```diff
@@ -0,0 +1,66 @@
...
+        for (String name : headers) {
+            updatedHeaders.remove(name);
```
mimaison: Because `Headers` is a `LinkedList`, `remove()` has to iterate the whole list each time. I wonder if we could instead start from an empty headers list and add the headers not being removed?

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        return record.newRecord(record.topic(), record.kafkaPartition(), record.keySchema(), record.key(),
+                record.valueSchema(), record.value(), record.timestamp(), updatedHeaders);
+    }
+
```
mimaison, line 59: nit, extra line

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/DropHeaders.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public ConfigDef config() {
+        return CONFIG_DEF;
+    }
+
+    @Override
+    public void close() {
+
```
mimaison: nit, extra line

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/InsertHeader.java
```diff
@@ -0,0 +1,72 @@
...
+    public static final String OVERVIEW_DOC =
+            "Add a header to each record.";
+
+    public static final String HEADER_FIELD = "header";
+    public static final String VALUE_LITERAL_FIELD = "value.literal";
+
+    public static final ConfigDef CONFIG_DEF = new ConfigDef()
+            .define(HEADER_FIELD, ConfigDef.Type.STRING, ConfigDef.Importance.HIGH,
+                    "The name of the header.")
+            .define(VALUE_LITERAL_FIELD, ConfigDef.Type.STRING, ConfigDef.Importance.HIGH,
```
mimaison: Should we enforce these fields are not null?

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
```diff
@@ -0,0 +1,233 @@
...
+                    "Use the concrete transformation type designed for the record " +
+                    "key (<code>" + Key.class.getName() + "</code>) or value (<code>" + Value.class.getName() + "</code>).";
+
+    public static final ConfigDef CONFIG_DEF = new ConfigDef()
+            .define(FIELDS_FIELD, ConfigDef.Type.LIST, ConfigDef.Importance.HIGH,
+                    "Field names in the record whose values are to be copied or moved to headers.")
+            .define(HEADERS_FIELD, ConfigDef.Type.LIST, ConfigDef.Importance.HIGH,
+                    "Header names, in the same order as the field names listed in the fields configuration property.")
+            .define(OPERATION_FIELD, ConfigDef.Type.STRING, NO_DEFAULT_VALUE,
+                    ConfigDef.ValidString.in("move", "copy"), ConfigDef.Importance.HIGH,
```
mimaison: Should we reuse `MOVE_OPERATION` and `COPY_OPERATION` here?

mimaison (changes requested):
Thanks or the PR! Good to see these SMTs finally being added.
I've made a first quick pass (haven't looked at `HeaderFrom` yet) and left a few suggestions

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/InsertHeader.java
tombentley: I guess that makes sense @mimaison, but then for consistency shouldn't we make `DropHeaders.headers`, `HeaderFrom.headers` and `HeaderFrom.fields` reject empty lists?

tombentley:
@mimaison I went ahead and enforce non-nullness and non-emptiness, if you want to do another pass?

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/HeaderFrom.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        static Operation fromName(String name) {
+            switch (name) {
+                case MOVE_OPERATION:
+                    return MOVE;
+                case COPY_OPERATION:
+                    return COPY;
+                default:
+                    throw new IllegalArgumentException();
```
mimaison, line 90: Even though I don't think it's reachable by users, should we have a message here?

tombentley: It is impossible due to the `ConfigDef.ValidString.in(MOVE_OPERATION, COPY_OPERATION)`, so this is really an assertion failure. The line number in the stacktrace would be enough to track it down if it ever did happen due to a later refactoring, so imho an error message is of no value. But I'm happy to add one if you like.

mimaison: Ok, that's fair enough. Thanks

mimaison (approved):
Thanks Tom, it looks good!

mimaison:
Not sure why the build had failed, I've rekicked it.

tombentley:
Failing tests are those reported in [KAFKA-12629](https://issues.apache.org/jira/browse/KAFKA-12629), [KAFKA-12284](https://issues.apache.org/jira/browse/KAFKA-12284) and [KAFKA-9295](https://issues.apache.org/jira/browse/KAFKA-9295):

```
Build / JDK 11 and Scala 2.13 / testCreateClusterAndCreateAndManyTopicsWithManyPartitions() – kafka.server.RaftClusterTest32s
Build / JDK 8 and Scala 2.12 / testReplication() – org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationSSLTest1m 32s
Build / JDK 8 and Scala 2.12 / testReplicationWithEmptyPartition() – org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationSSLTest1m 26s
Build / JDK 8 and Scala 2.12 / testOneWayReplicationWithAutoOffsetSync() – org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationSSLTest1m 23s
Build / JDK 8 and Scala 2.12 / testReplication() – org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationTest1m 27s
Build / JDK 8 and Scala 2.12 / testReplicationWithEmptyPartition() – org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationTest1m 16s
Build / JDK 8 and Scala 2.12 / shouldInnerJoinMultiPartitionQueryable – org.apache.kafka.streams.integration.KTableKTableForeignKeyInnerJoinMultiIntegrationTest
```

All these pass using the relevant JDKs on my machine. Merging to trunk.

PR #9553, opened 2020-11-03
KAFKA-10427:  Fetch snapshot

jsancio:
Implements the code necessary for the leader to response to fetch snapshot requests and for the follower to fetch snapshots. It depends on PR https://github.com/apache/kafka/pull/9512 and that PR should be reviewed first. Included are the following changes.

Leader Changes:
1. Raft leader response to FetchSnapshot request by reading the local snapshot and sending the requested bytes in the response. This implementation currently copies the bytes to memory. This will be fixed in a future PR.

Follower Changes:
1. Raft followers will start fetching snapshot if the leader sends a Fetch response that includes a SnapshotId.

2. Raft followers send FetchSnapshot requests if there is a pending download. The same timer is used for both Fetch and FetchSnapshot requests.

3. Raft follower handle FetchSnapshot responses by comping the bytes to the pending SnapshotWriter. This implementation doesn't fix the replicated log after the snapshot has been downloaded. This will be implemented in a future PR.
 
comment to ping reviewers.*


raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1037,6 +1047,14 @@ private boolean handleFetchResponse(
                     logger.info("Truncated to offset {} from Fetch response from leader {}",
                         truncationOffset, quorum.leaderIdOrNil());
                 });
+            } else if (partitionResponse.snapshotId().epoch() >= 0 && partitionResponse.snapshotId().endOffset() >= 0) {
```
jsancio: Fix this logic to be more conservative. It should be a fatal error if one is negative and the other is positive.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1101,6 +1140,174 @@ private DescribeQuorumResponseData handleDescribeQuorumRequest(
...
+        if (snapshot.sizeInBytes() != partitionSnapshot.position()) {
+            throw new IllegalStateException(String.format("Received fetch snapshot response with an invalid position. Expected %s; Received %s", snapshot.sizeInBytes(), partitionSnapshot.position()));
+        }
+
+        snapshot.append(partitionSnapshot.bytes());
+
+        if (snapshot.sizeInBytes() == partitionSnapshot.size()) {
+            // Finished fetching the snapshot.
+            snapshot.freeze();
+            state.setFetchingSnapshot(Optional.empty());
```
jsancio, line 1328: The follower needs to update the log start offset and log end offset after it has successfully fetched a snapshot. I want to implement this part in this JIRA: https://issues.apache.org/jira/browse/KAFKA-10820

clients/src/main/java/org/apache/kafka/common/requests/FetchSnapshotResponse.java
```diff
@@ -0,0 +1,124 @@
...
+        return data;
+    }
+
+    /**
+     * Creates a FetchSnapshotResponseData with a top level error.
+     *
+     * @param error the top level error
+     * @return the created fetch snapshot response data
+     */
+    public static FetchSnapshotResponseData withTopError(Errors error) {
```
hachikuji: nit: maybe `withTopLevelError`?

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
```diff
@@ -287,4 +287,7 @@ public String name() {
         return "Leader";
     }
 
+    @Override
+    public void close() {}
```
hachikuji, line 305: Maybe we could add a default no-op implementation to EpochState?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -941,6 +949,8 @@ private FetchResponseData tryCompleteFetchRequest(
     ) {
         Optional<Errors> errorOpt = validateLeaderOnlyRequest(request.currentLeaderEpoch());
         if (errorOpt.isPresent()) {
+            // TODO: The replica should return what information it knows about the current epoch and
```
hachikuji: Isn't this already done by `buildFetchResponse`?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1037,6 +1047,35 @@ private boolean handleFetchResponse(
                     logger.info("Truncated to offset {} from Fetch response from leader {}",
                         truncationOffset, quorum.leaderIdOrNil());
                 });
+            } else if (partitionResponse.snapshotId().epoch() >= 0 ||
+                       partitionResponse.snapshotId().endOffset() >= 0) {
+                // The leader is asking us to fetch a snapshot
+
+                if (partitionResponse.snapshotId().epoch() < 0) {
+                    throw new KafkaException(
```
hachikuji: Hmm.. The leader has sent a bad response. I think logging an error and retrying might be better than crashing the followers.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1101,6 +1140,174 @@ private DescribeQuorumResponseData handleDescribeQuorumRequest(
...
+                        .setErrorCode(Errors.UNKNOWN_TOPIC_OR_PARTITION.code());
+                }
+            );
+        }
+
+        FetchSnapshotRequestData.PartitionSnapshot partitionSnapshot = partitionSnapshotOpt.get();
+        Optional<Errors> leaderValidation = validateLeaderOnlyRequest(
+                partitionSnapshot.currentLeaderEpoch()
+        );
+        if (leaderValidation.isPresent()) {
```
hachikuji, line 1185: nit: I guess you could use `leaderValidation.ifPresent`

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1101,6 +1140,174 @@ private DescribeQuorumResponseData handleDescribeQuorumRequest(
...
+            int maxSnapshotSize;
+            try {
+                maxSnapshotSize = Math.toIntExact(snapshot.sizeInBytes());
+            } catch (ArithmeticException e) {
+                maxSnapshotSize = Integer.MAX_VALUE;
+            }
+
+            // TODO: Make sure that we also limit based on the fetch max bytes configuration
+            ByteBuffer buffer = ByteBuffer.allocate(Math.min(data.maxBytes(), maxSnapshotSize));
+            snapshot.read(buffer, partitionSnapshot.position());
```
hachikuji, line 1229: Is it worth validating that `partitionSnapshot.position()` is non-negative?

clients/src/main/resources/common/message/FetchSnapshotResponse.json
```diff
@@ -0,0 +1,59 @@
...
+          { "name": "LeaderId", "type": "int32", "versions": "0+",
+            "about": "The ID of the current leader or -1 if the leader is unknown."},
+          { "name": "LeaderEpoch", "type": "int32", "versions": "0+",
+            "about": "The latest known leader epoch"}
+        ]},
+        { "name": "Size", "type": "int64", "versions": "0+",
+          "about": "The total size of the snapshot." },
+        { "name": "Position", "type": "int64", "versions": "0+",
+          "about": "The starting byte position within the snapshot included in the Bytes field." },
+        { "name": "Bytes", "type": "bytes", "versions": "0+", "zeroCopy": true,
```
hachikuji, line 54: Can you remind me if we are planning to change the type to "records"? I don't think we will get the benefit of `sendfile` unless we do so.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1101,6 +1140,174 @@ private DescribeQuorumResponseData handleDescribeQuorumRequest(
...
+    }
+
+    private boolean handleFetchSnapshotResponse(
+        RaftResponse.Inbound responseMetadata,
+        long currentTimeMs
+    ) throws IOException {
+        FetchSnapshotResponseData data = (FetchSnapshotResponseData) responseMetadata.data;
+        Errors topLevelError = Errors.forCode(data.errorCode());
+        if (topLevelError != Errors.NONE) {
+            // TODO: check what values this expression returns
```
hachikuji: Address TODO?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1101,6 +1140,174 @@ private DescribeQuorumResponseData handleDescribeQuorumRequest(
...
+            return handled.get();
+        }
+
+        FollowerState state = quorum.followerStateOrThrow();
+
+        if (Errors.forCode(partitionSnapshot.errorCode()) == Errors.SNAPSHOT_NOT_FOUND ||
+            partitionSnapshot.snapshotId().endOffset() < 0 ||
+            partitionSnapshot.snapshotId().epoch() < 0) {
+
+            /* The leader deleted the snapshot before the follower could download it. Start over by
```
hachikuji, line 1290: A log message would probably be helpful. It's probably worth doing one full pass over the logic here to see where we could add extra logging.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1629,16 +1872,29 @@ private long pollFollowerAsVoter(FollowerState state, long currentTimeMs) throws
             transitionToCandidate(currentTimeMs);
             return 0L;
         } else {
-            long backoffMs = maybeSendRequest(
-                currentTimeMs,
-                state.leaderId(),
-                this::buildFetchRequest
-            );
+            long backoffMs;
+            if (state.fetchingSnapshot().isPresent()) {
```
hachikuji: Can we move this logic to `pollFollower` somehow?

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientSnapshotTest.java
```diff
@@ -0,0 +1,765 @@
...
+        FetchSnapshotResponseData.PartitionSnapshot response =  context.assertSentFetchSnapshotResponse(context.metadataPartition).get();
+        assertEquals(Errors.NOT_LEADER_OR_FOLLOWER, Errors.forCode(response.errorCode()));
+        assertEquals(epoch, response.currentLeader().leaderEpoch());
+        assertEquals(leaderId, response.currentLeader().leaderId());
+    }
+
+    @Disabled
+    @Test
+    public void testFetchSnapshotRequestWithOlderEpoch() throws IOException {
+        assertTrue(false);
```
hachikuji: I guess you're still planning to implement these?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1101,6 +1140,174 @@ private DescribeQuorumResponseData handleDescribeQuorumRequest(
         );
     }
 
+    private FetchSnapshotResponseData handleFetchSnapshotRequest(
```
hachikuji, line 1154: Just checking my understanding. This patch adds the logic to respond to the snapshot id from a fetch response and to handle send/handle snapshots when needed. However, since it does not contain the logic to set the snapshot id in the fetch request handler, none of this logic will get exercised by the simulation test. Is that right?

jsancio: Correct. I am handling this case in https://issues.apache.org/jira/browse/KAFKA-10761. I think that after implementing that issue we should have an end-to-end working Raft Client with snapshot support that we test in the simulations.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Hmm. I don't think so. `ifPresent` returns `void` and this part of the code wants to return a `FetchSnapshotResponseData` if there was a validation error. We can do a `map` followed by a `isPresent` and `get` but I think this is more consistent with the rest of the code.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Good catch. In the KIP we added `POSITION_OUT_OF_RANGE` which I am not using any where in this PR.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1101,6 +1140,174 @@ private DescribeQuorumResponseData handleDescribeQuorumRequest(
...
+
+        try (RawSnapshotReader snapshot = snapshotOpt.get()) {
+            int maxSnapshotSize;
+            try {
+                maxSnapshotSize = Math.toIntExact(snapshot.sizeInBytes());
+            } catch (ArithmeticException e) {
+                maxSnapshotSize = Integer.MAX_VALUE;
+            }
+
+            // TODO: Make sure that we also limit based on the fetch max bytes configuration
```
jsancio: I created this issue: https://issues.apache.org/jira/browse/KAFKA-10884

We can fix this for Raft Client's implementation of both `Fetch` and `FetchSnapshot` at the same time.

clients/src/main/resources/common/message/FetchSnapshotResponse.json
jsancio: Yeah. I created this issue: https://issues.apache.org/jira/browse/KAFKA-10694

I think we have two options:
1. Use "records" as you have suggested.
2. Similar to "records" introduce a new type called for example `Memory` that has two implementations one that encapsulates `ByteBuffer` and another that encapsulates a file and a start position. This is similar to "records", `MemoryRecords` and `FileRecords` without the requirement that the start needs to be a valid `RecordBatch`.

I want to play around with this soon and restart the conversation on that issue and PR.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Yes. This was a note to me to confirm this as it wasn't clear to me when I first read the code.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Yeah. I am not sure. If these exception are thrown then it means that the epoch is "valid" but the end offset is not "valid" or vice versa. I think this could happen because of either a buggy remote replica or corrupted data in the remote replica.

I guess you can argue that it is safe to keep retrying.

In the case that the `Fetch` never succeed do we want to transition to the candidate the state? If so, let me add a test that confirms that.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Done. Excuse the noise.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: I cleaned this up by moving it to a method we can reuse. Let me know if that addresses your concern.

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
jsancio: I prefer to have each of the epoch states explicitly opt out of this `close` method. This makes it clear that this state doesn't have any resource that it wishes to clean/close during a transition. Instead of future code changes forgetting to override this method. What do you think?

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientSnapshotTest.java
jsancio: Yes. Doing that now...

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Hmm.. I would rather not handle this case specially by transitioning to a candidate. It's not a crazy idea, but I think we should attempt that in a separate PR and consider errors more holistically.

I would suggest that we log an error saying that the remote replica seemed to return an invalid response and just keep fetching. Then a user can see the log message and restart the remote replica.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Done.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: > I would suggest that we log an error saying that the remote replica seemed to return an invalid response and just keep fetching. Then a user can see the log message and restart the remote replica.

Yeah. This is what I implemented and added a test for it. In other words.
1. Log an error message
2. Tell the raft client that the response was handle successfully but the fetch timer was not reset

In practice this results in the follower continuing to send `Fetch` requests. After `fetchTimeoutMs` the follower will transition to candidate as the existing client code does. See https://github.com/apache/kafka/pull/9553/files#diff-86474ad1438150630c21b29a3da2f6dd79d1357e33ac034f00e5fcef0f2e889cR350

Let me know if this is what you were thinking.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1055,14 +1055,14 @@ private boolean handleFetchResponse(
                         partitionResponse.snapshotId().endOffset(),
                         partitionResponse.snapshotId().epoch()
                     );
-                    return false;
+                    return true;
```
hachikuji: Returning `false` seemed more appropriate. That ensures we will get a backoff before the next retry.

clients/src/main/resources/common/message/FetchSnapshotResponse.json
dengziming: @jsancio Can we add an index file to the snapshot file so that the `FetchSnaphotResponseData` can also satisfy the requirement that the start is a valid `RecordBatch`? I can help with some work of this issue: https://issues.apache.org/jira/browse/KAFKA-10694.

jsancio: @dengziming, Thanks for the comment. The `FetchSnapshotRequest` differs from `FetchRequest` in that `Position` is a byte offset. Because of this we don't need an `Offset` to `Byte` index.

If we implement option 1 in my comment above, we can still use `Records` by making sure that:
1. The `leaders` send at least one entire `RecordBatch` in the `FetchSnapshotResponse`.
2. The `followers` only write complete `RecordBatch`es by not writing bytes sent by the leader that are not complete `RecordBatch`.

This guarantees that the `Position` sent in `FetchSnapshotRequest` is `RecordBatch` aligned and it always makes progress because the leader sends at least one complete `RecordBatch`.

Having said that, I think we should try and implement option 2. Unfortunately, the Jira https://issues.apache.org/jira/browse/KAFKA-10694 doesn't document option 2.

hachikuji (approved):
LGTM

hachikuji:
I manually reran tests since it has been a week since the last update. Here is a link to the results since they have not been automatically linked: https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-9553/14/.

PR #9555, opened 2020-11-04
KAFKA-10673: Cache inter broker listener name used in connection quotas

splett2:
A profile from a moderately busy cluster shows that calls to `protectedListener` can make up more than 1% of the allocations on a broker.

`config.interBrokerListenerName` is a relatively expensive call that both makes a copy of the `KafkaConfig`'s backing map and performs string/regex parsing. Given that we call `protectedListener()` multiple times per call to `ConnectionQuotas.inc()`, we end up performing a lot of unnecessary allocations, particularly given that the inter broker listener is not reconfigurable. We can instead store the inter broker listener name and check against it rather than recompute from the config.

An additional, smaller optimization included is to check the size of listener counts instead of `config.listeners`, as `config.listeners()` does some additional string csv parsing to recompute the listeners each time.

### DIFF
--- a/core/src/main/scala/kafka/network/SocketServer.scala
+++ b/core/src/main/scala/kafka/network/SocketServer.scala
@@ -1278,6 +1278,7 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
   @volatile private var defaultMaxConnectionsPerIp: Int = config.maxConnectionsPerIp
   @volatile private var maxConnectionsPerIpOverrides = config.maxConnectionsPerIpOverrides.map { case (host, count) => (InetAddress.getByName(host), count) }
   @volatile private var brokerMaxConnections = config.maxConnections
+  private val interBrokerListenerName = config.interBrokerListenerName
   private val counts = mutable.Map[InetAddress, Int]()
 
   // Listener counts and configs are synchronized on `counts`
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1189,6 +1189,7 @@ class ConnectionQuotas(config: KafkaConfig, time: Time, metrics: Metrics) extend
   @volatile private var defaultMaxConnectionsPerIp: Int = config.maxConnectionsPerIp
   @volatile private var maxConnectionsPerIpOverrides = config.maxConnectionsPerIpOverrides.map { case (host, count) => (InetAddress.getByName(host), count) }
   @volatile private var brokerMaxConnections = config.maxConnections
+  @volatile private var interBrokerListenerName = config.interBrokerListenerName
```
apovzner: Good point and definitely agree that `config.interBrokerListenerName` could be expensive as we call it several times per accepting a new connection. 

The issue here is that interBrokerListenerName is a dynamic config. So, you will need to update the cached value on changes to that config; similar how we notify ConnectionQuotas about config changes from SocketServer.reconfigure().

splett2: `interBrokerListenerName` is apparently not a dynamic config.

see `DynamicBrokerReconfigurationTest.testAdvertisedListenerUpdate`
...
```
    // Verify updating inter-broker listener
    val props = new Properties
    props.put(KafkaConfig.InterBrokerListenerNameProp, SecureExternal)
    try {
      reconfigureServers(props, perBrokerConfig = true, (KafkaConfig.InterBrokerListenerNameProp, SecureExternal))
      fail("Inter-broker listener cannot be dynamically updated")
   }
```
I don't think we allow updating inter broker listener at all, so I think we can remove the test I added. I actually wasn't sure if we allowed it or not, but the code seems to suggest that it isn't.

apovzner: You are right -- I saw that test and thought it was a dynamic config, but the test was verifying that it cannot be updated. I see now that KIP-226 lists that as future work, cool.

apovzner (approved):
LGTM

dajac (approved):
LGTM, thanks for the PR!

PR #9560, opened 2020-11-05
KAFKA-10345: Add ZK-notification based update for trust/key store paths

abbccdda:
SSL trust store and key store paths update could no longer go through the direct per broker update due to forwarding. We need to add a mechanism to trigger the update through ZK notification.

### DIFF
--- a/core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
+++ b/core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
@@ -29,7 +29,7 @@ import kafka.utils.{CoreUtils, Logging, PasswordEncoder}
 import kafka.utils.Implicits._
 import kafka.zk.{AdminZkClient, KafkaZkClient}
 import org.apache.kafka.common.Reconfigurable
-import org.apache.kafka.common.config.{ConfigDef, ConfigException, SslConfigs, AbstractConfig}
+import org.apache.kafka.common.config.{AbstractConfig, ConfigDef, ConfigException, SslConfigs}
 import org.apache.kafka.common.metrics.MetricsReporter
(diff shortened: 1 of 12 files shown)


core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
```diff
@@ -331,6 +334,50 @@ class DynamicBrokerConfig(private val kafkaConfig: KafkaConfig) extends Logging
...
 
+  private[server] def maybeAugmentSslStorePaths(configProps: Properties, previousConfigProps: Map[String, String]): Unit ={
+    val processedFiles = new mutable.HashSet[String]
+    reconfigurables
+      .filter(reconfigurable => ReloadableFileConfigs.exists(reconfigurable.reconfigurableConfigs.contains))
+        .foreach({
+          case reconfigurable: ListenerReconfigurable =>
+            ReloadableFileConfigs.foreach(configName => {
+              val prefixedName = reconfigurable.listenerName.configPrefix + configName
+              if (!processedFiles.contains(prefixedName) && configProps.containsKey(prefixedName) &&
```
abbccdda, line 345: @rajinisivaram One case I'm not covering here is that whether we should compare the entire `configProps` with `previousConfigProps` first and skip the augment path when they are the same. Right now I'm only comparing the path change of key/trust store, do you think we should do the comparison on entire config change first?

rajinisivaram: I think the current logic seems reasonable since you can have an update with no changes where we just want to reload the key/trust stores. In that case `configProps` would match `previousConfigProps`.

PR #9561, opened 2020-11-05
KAFKA-10624: For FeatureZNodeStatus, use sealed trait instead of Enumeration

kowshik:
This is a follow-up to initial KIP-584 development. In this PR, I've switched the `FeatureZNodeStatus` enum to be a sealed trait. In Scala, we prefer sealed traits over Enumeration since the former gives you exhaustiveness checking. With Scala enumeration, you don't get a warning if you add a new value that is not handled in a given pattern match.

**Test plan:**
Rely on existing tests.

### DIFF
--- a/core/src/main/scala/kafka/zk/ZkData.scala
+++ b/core/src/main/scala/kafka/zk/ZkData.scala
@@ -798,12 +798,25 @@ object DelegationTokenInfoZNode {
  *             written by the controller to the FeatureZNode only when the broker IBP config
  *             is less than KAFKA_2_7_IV0.
  */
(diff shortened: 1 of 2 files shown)


kowshik:
@junrao @abbccdda: this PR is ready for review.

kowshik:
The build in CI is broken due to absence of https://github.com/apache/kafka/pull/9559. I'll rebase on top of latest AK trunk, after that PR is merged.

kowshik:
I've rebased on top of #9559 now. So the build failure(s) should be gone.

junrao (approved):

PR #7409, opened 2019-09-29
KAFKA-10818: Skip conversion to `Struct` when serializing generated requests/responses

ijuma:
Generated request/response classes have code to serialize/deserialize directly to
`ByteBuffer` so the intermediate conversion to `Struct` can be skipped for them.
We have recently completed the transition to generated request/response classes,
so we can also remove the `Struct` based fallbacks.

Additional noteworthy changes:
* `AbstractRequest.parseRequest` has a more efficient computation of request size that
relies on the received buffer instead of the parsed `Struct`.
* Use `SendBuilder` for `AbstractRequest/Response` `toSend`, made the superclass
implementation final and removed the overrides that are no longer necessary.
* Removed request/response constructors that assume latest version as they are unsafe
outside of tests.
* Removed redundant version fields in requests/responses.
* Removed unnecessary work in `OffsetFetchResponse`'s constructor when version >= 2.
* Made `AbstractResponse.throttleTimeMs()` abstract.
* Using `toSend` in `SaslClientAuthenticator` instead of `serialize`.
* Various changes in Request/Response classes to make them more consistent and to
rely on the Data classes as much as possible when it comes to their state.
* Remove the version argument from `AbstractResponse.toString`.
* Fix `getErrorResponse` for `ProduceRequest` and `DescribeClientQuotasRequest` to
use `ApiError` which processes the error message sent back to the clients. This was
uncovered by an accidental fix to a `RequestResponseTest` test (it was calling
`AbstractResponse.toString` instead of `AbstractResponse.toString(short)`).

Rely on existing protocol tests to ensure this refactoring does not change 
observed behavior (aside from improved performance).

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9563, opened 2020-11-05
KAFKA-10684; Avoid additional envelope copies during network transmission

hachikuji:
This patch creates a new `SendBuilder` class which allows us to avoid copying "bytes" types when transmitting an api message over the network. This is used in `EnvelopeRequest` and `EnvelopeResponse` to avoid copying the embedded data.

The patch also contains a few minor cleanups such as moving envelope parsing logic into `RequestContext`.


clients/src/main/resources/common/message/EnvelopeRequest.json
```diff
@@ -23,7 +23,7 @@
   "fields": [
     { "name": "RequestData", "type": "bytes", "versions": "0+", "zeroCopy": true,
       "about": "The embedded request header and data."},
-    { "name": "RequestPrincipal", "type": "bytes", "versions": "0+", "zeroCopy": true, "nullableVersions": "0+",
+    { "name": "RequestPrincipal", "type": "bytes", "versions": "0+", "nullableVersions": "0+",
```
chia7712, line 26: Is ```nullableVersions`` required? It seems there is no null handle/check in production for this field.

clients/src/main/java/org/apache/kafka/common/network/SendBuilder.java
```diff
@@ -0,0 +1,119 @@
...
+ * to allocate a buffer which accounts only for the additional request overhead.
+ *
+ * See {@link org.apache.kafka.common.requests.EnvelopeRequest#toSend(String, RequestHeader)}
+ * for example usage.
+ */
+public class SendBuilder implements Writable {
+    private final List<ByteBuffer> buffers = new ArrayList<>();
+    private final ByteBuffer buffer;
+
+    public SendBuilder(ByteBuffer buffer) {
```
chia7712: The ```buffer``` is used/owned by ```SendBuilder``` only. It seems to me this constructor should accept ```capacity(int type)``` rather than ```ByteBuffer```. (SendBuilder should create ```ByteBuffer``` in construction)

chia7712:
Does this improvement work for other requests?

hachikuji:
@chia7712 I think it could. I was actually thinking about #9401 when I was working on it. As far as I can tell, the client currently copies the whole produce buffer when the request reaches `NetworkClient`. One remaining thing to do here I think is converge `SendBuilder` and `RecordsWritable`, which are using similar tricks.

clients/src/main/java/org/apache/kafka/common/requests/RequestHeader.java
```diff
@@ -89,10 +89,11 @@ public ResponseHeader toResponseHeader() {
     public static RequestHeader parse(ByteBuffer buffer) {
         short apiKey = -1;
         try {
+            int position = buffer.position();
```
abbccdda, line 92: Is this specifically for Envelope case?

hachikuji: This was a bug. This logic assumes that the buffer is at position 0. Let me add a test case.

hachikuji:
I got inspired to try and extend this to apply to all message types. I've updated the patch to remove the custom response logic in `FetchResponse` in favor of a general pattern using `SendBuilder`. We had an existing benchmark `FetchResponseBenchmark.testSerializeFetchResponse`, so I tried it out.

Here is the results before the patch (10 topics and 500 topics):
```
Result "org.apache.kafka.jmh.common.FetchResponseBenchmark.testSerializeFetchResponse":
  6322.861 ±(99.9%) 310.785 ns/op [Average]
  (min, avg, max) = (6057.758, 6322.861, 7090.658), stdev = 290.708
  CI (99.9%): [6012.076, 6633.646] (assumes normal distribution)

Result "org.apache.kafka.jmh.common.FetchResponseBenchmark.testSerializeFetchResponse":
  323310.283 ±(99.9%) 25947.515 ns/op [Average]
  (min, avg, max) = (301370.273, 323310.283, 383716.556), stdev = 24271.322
  CI (99.9%): [297362.768, 349257.799] (assumes normal distribution)
```
Here is the new benchmark (10 topics and 500 topics):
```
Result "org.apache.kafka.jmh.common.FetchResponseBenchmark.testSerializeFetchResponse":
  5701.378 ±(99.9%) 100.848 ns/op [Average]
  (min, avg, max) = (5601.838, 5701.378, 5925.943), stdev = 94.333
  CI (99.9%): [5600.530, 5802.225] (assumes normal distribution)

Result "org.apache.kafka.jmh.common.FetchResponseBenchmark.testSerializeFetchResponse":
  298221.825 ±(99.9%) 8173.945 ns/op [Average]
  (min, avg, max) = (287615.891, 298221.825, 321499.618), stdev = 7645.913
  CI (99.9%): [290047.880, 306395.770] (assumes normal distribution)
```
So looks like a modest overall improvement.

Note I still need to polish up a few things in the PR.

clients/src/main/resources/common/message/EnvelopeRequest.json
hachikuji: I agree it should not be nullable. Will fix it.

clients/src/main/java/org/apache/kafka/common/protocol/Writable.java
```diff
@@ -33,6 +35,23 @@
...
 
+    default void writeApiMessage(
+        ApiMessage message,
+        ObjectSerializationCache serializationCache,
+        short version
+    ) {
+        message.write(this, serializationCache, version);
+    }
+
+    default void writeRecords(BaseRecords records) {
```
chia7712, line 38: Does it need null check (maybe no-op)?

clients/src/main/java/org/apache/kafka/common/protocol/Writable.java
```diff
@@ -33,6 +35,23 @@
     void writeVarint(int i);
     void writeVarlong(long i);
 
+    default void writeApiMessage(
```
chia7712: It is used by only ```SendBuilder```. How about moving it to ```SendBuilder```?

generator/src/main/java/org/apache/kafka/message/MessageDataGenerator.java
```diff
@@ -1581,56 +1570,56 @@ private void generateVariableLengthFieldSize(FieldSpec field,
...
                         headerGenerator.addImport(MessageGenerator.BYTE_UTILS_CLASS);
-                        buffer.printf("_size += _bytesSize + ByteUtils.sizeOfUnsignedVarint(_bytesSize);%n");
-                    } else {
-                        buffer.printf("_size += _bytesSize;%n");
+                        buffer.printf("_size.addBytes(ByteUtils.sizeOfUnsignedVarint(_bytesSize.totalSize()));%n");
                     }
+                    buffer.printf("_size.add(_bytesSize);%n");
                 } else if (field.type().isRecords()) {
-                    buffer.printf("int _recordsSize = %s.sizeInBytes();%n", field.camelCaseName());
+                    buffer.printf("_size.addBytes(%s.sizeInBytes());%n", field.camelCaseName());
```
chia7712: Why it is not ```addZeroCopyBytes```?

hachikuji: Yeah, my bad. Probably messed this up after renaming.

clients/src/main/java/org/apache/kafka/common/protocol/Writable.java
hachikuji: I was concerned about this also, but the generated code adds its own null check.

clients/src/main/java/org/apache/kafka/common/protocol/Writable.java
hachikuji: Fair enough.

clients/src/main/java/org/apache/kafka/common/requests/EnvelopeRequest.java
```diff
@@ -91,4 +91,14 @@ public AbstractResponse getErrorResponse(int throttleTimeMs, Throwable e) {
...
         return new EnvelopeRequest(ApiKeys.ENVELOPE.parseRequest(version, buffer), version);
     }
+
+    public EnvelopeRequestData data() {
+        return data;
+    }
+
+    @Override
+    public Send toSend(String destination, RequestHeader header) {
+        return SendBuilder.buildRequestSend(destination, header, this.data);
```
chia7712, line 101: not sure whether it is ok to ignore the version of EnvelopeRequest.

hachikuji: The version is contained in `RequestHeader`. This pattern was probably started before `AbstractRequest` included the version as well.

hachikuji:
@chia7712 Thanks for the reviews. I pushed an update to address your comments. I was a little concerned about the garbage created from the `MessageSize` objects, so I changed the generated code to use an accumulator pattern instead. Not sure it was really worth it, but let me know what you think.

hachikuji:
Here are updated results for `FetchResponseBenchmark`.

Trunk:
```
Benchmark                                          (partitionCount)  (topicCount)  Mode  Cnt        Score       Error  Units
FetchResponseBenchmark.testSerializeFetchResponse                 3            10  avgt   15     6037.058 ±    41.647  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                 3           500  avgt   15   296913.061 ±  2548.748  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                 3          1000  avgt   15   607466.969 ±  5149.176  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                10            10  avgt   15    17971.239 ±   141.934  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                10           500  avgt   15   888464.188 ±  6341.169  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                10          1000  avgt   15  1938108.325 ± 32056.839  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                20            10  avgt   15    34217.206 ±   234.166  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                20           500  avgt   15  1830593.478 ± 22601.172  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                20          1000  avgt   15  4983809.392 ± 38217.505  ns/op
```

This patch:
```
Benchmark                                          (partitionCount)  (topicCount)  Mode  Cnt        Score       Error  Units
FetchResponseBenchmark.testSerializeFetchResponse                 3            10  avgt   15     5686.644 ±    92.892  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                 3           500  avgt   15   294136.644 ±  6709.506  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                 3          1000  avgt   15   599678.709 ±  5837.803  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                10            10  avgt   15    17065.100 ±   241.715  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                10           500  avgt   15   907448.421 ± 14203.990  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                10          1000  avgt   15  1906411.458 ± 26450.041  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                20            10  avgt   15    33147.044 ±  1181.493  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                20           500  avgt   15  1861946.674 ± 25145.868  ns/op
FetchResponseBenchmark.testSerializeFetchResponse                20          1000  avgt   15  4339003.253 ± 69576.240  ns/op
```

So generally better, but not groundbreaking. That is fine since the main improvement is simplifying the generation of `Send` objects for all request types.

generator/src/main/java/org/apache/kafka/message/MessageDataGenerator.java
```diff
@@ -1579,58 +1566,58 @@ private void generateVariableLengthFieldSize(FieldSpec field,
...
+                        buffer.printf("int _arraySize = _size.totalSize() - _sizeBeforeArray;%n");
                         buffer.printf("_cache.setArraySizeInBytes(%s, _arraySize);%n",
                             field.camelCaseName());
-                        buffer.printf("_size += _arraySize + ByteUtils.sizeOfUnsignedVarint(_arraySize);%n");
-                    } else {
-                        buffer.printf("_size += _arraySize;%n");
+                        buffer.printf("_size.addBytes(ByteUtils.sizeOfUnsignedVarint(_arraySize));%n");
                     }
                 } else if (field.type().isBytes()) {
+                    buffer.printf("int _sizeBeforeBytes = _size.totalSize();%n");
```
chia7712:
```java
if (tagged) {
  buffer.printf("int _sizeBeforeBytes = _size.totalSize();%n");
}
```

clients/src/main/java/org/apache/kafka/common/requests/EnvelopeRequest.java
```diff
@@ -91,4 +91,14 @@ public AbstractResponse getErrorResponse(int throttleTimeMs, Throwable e) {
     public static EnvelopeRequest parse(ByteBuffer buffer, short version) {
         return new EnvelopeRequest(ApiKeys.ENVELOPE.parseRequest(version, buffer), version);
     }
+
+    public EnvelopeRequestData data() {
+        return data;
+    }
+
+    @Override
+    public Send toSend(String destination, RequestHeader header) {
```
chia7712, line 100: If all requests are using auto-generated data, should this be default implementation of ```AbstractRequest```?

clients/src/main/java/org/apache/kafka/common/protocol/Message.java
```diff
@@ -47,7 +47,20 @@
...
+    }
+
+    /**
+     * Add the size of this message to an accumulator.
+     *
+     * @param size          The size accumulator to add to
+     * @param cache         The serialization size cache to populate.
+     * @param version       The version to use.
+     */
+    void addSize(MessageSizeAccumulator size, ObjectSerializationCache cache, short version);
```
chia7712, line 63: I'm thinking about how to simplify this process.

Could we reuse the method ```void write(Writable writable, ObjectSerializationCache cache, short version)``` ? Maybe we can create a ```Writable``` instance but it does not write data to any output. Instead, it calculate the size of buffer according to input data.

chia7712 (commented):
@hachikuji This improvement LGTM overall.

It seems to me the serialization mechanism get more complicated since we are trying to add different implementation for IO-heavy requests get better memory usage. For more readable and consistent code, is there a follow-up to apply this improvement to all requests (if we complete all auto-generated protocol migration)?

clients/src/main/java/org/apache/kafka/common/requests/EnvelopeRequest.java
hachikuji: Yeah, I think so. And looks like we're almost there. After your patch for `Produce`, the only remaining unconverted API that I see is `OffsetsForLeaderEpoch`.

generator/src/main/java/org/apache/kafka/message/MessageDataGenerator.java
hachikuji: Good catch

clients/src/main/java/org/apache/kafka/common/requests/EnvelopeRequest.java
hachikuji: #7409 might be a good opportunity to complete this. What do you think @ijuma ?

generator/src/main/java/org/apache/kafka/message/MessageDataGenerator.java
```diff
@@ -1529,37 +1516,37 @@ private void generateVariableLengthFieldSize(FieldSpec field,
...
                                 throw new RuntimeException("Tagged field " + field.name() +
                                     " should not be present in non-flexible versions.");
                             }
-                            buffer.printf("_size += _stringBytes.length + 2;%n");
+                            buffer.printf("_size.addBytes(_stringBytes.length + 2);%n");
                         }).
                         generate(buffer);
                 } else if (field.type().isArray()) {
-                    buffer.printf("int _arraySize = 0;%n");
+                    buffer.printf("int _sizeBeforeArray = _size.totalSize();%n");
```
chia7712:
```java
if (tagged) {
  buffer.printf("int _sizeBeforeArray = _size.totalSize();%n");
}
```

clients/src/main/java/org/apache/kafka/common/requests/AbstractResponse.java
```diff
@@ -88,16 +80,11 @@ protected void updateErrorCounts(Map<Errors, Integer> errorCounts, Errors error)
...
-
-        return buffer;
-    }
-
-    public static AbstractResponse deserializeBody(ByteBuffer byteBuffer, RequestHeader header) {
+    /**
+     * Parse a response from the provided buffer. The buffer is expected to hold both
+     * the {@link ResponseHeader} as well as the response payload.
+     */
+    public static AbstractResponse parseResponse(ByteBuffer byteBuffer, RequestHeader header) {
```
chia7712, line 87: Is this method equal to https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java#L727 ?

clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
```diff
@@ -0,0 +1,194 @@
...
+        buffer.putLong(val);
+    }
+
+    @Override
+    public void writeDouble(double val) {
+        buffer.putDouble(val);
+    }
+
+    @Override
+    public void writeByteArray(byte[] arr) {
```
chia7712, line 94: Is it worth wrapping the byte array to a new ```ByteBufferSend``` to avoid array coping?

hachikuji: I did that in an earlier iteration, but I wasn't sure if the overhead of the `ByteBuffer` and `ByteBufferSend` made it a net win in the end. I also thought about adding a heuristic, such as looking for a minimum size. In the end, it seemed simpler to rely on "zeroCopy" to let us know when we are likely to get a benefit.

chia7712: Thanks for your explanation. make sense to me.

clients/src/main/java/org/apache/kafka/common/requests/AbstractResponse.java
hachikuji: If it's ok with you, I'd like to address this in a separate patch. The main difference is the presence of the correlation validation logic in `NetworkClient`, which has been tailored to a subtle case in `SaslClientAuthenticator`. I think the envelope parsing logic should also be checking the correlationId, but probably not with the same quirky behavior.

chia7712: sure. Open a jira as follow-up :)

chia7712 (approved):
@hachikuji this is a great improvement. +1

hachikuji:
@chia7712 Thanks for reviews, merging to trunk. I will follow up with the suggestion about `AbstractResponse.parseResponse` as well as a couple other improvements.

PR #9564, opened 2020-11-05
KAFKA-10667: add timeout for forwarding requests

abbccdda:
Right now the forwarding request will retry indefinitely, which is not the ideal behavior. We should timeout the enqueued request when it hits the request timeout.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/RequestCompletionHandler.java
+++ b/clients/src/main/java/org/apache/kafka/clients/RequestCompletionHandler.java
@@ -23,5 +23,4 @@ package org.apache.kafka.clients;
 public interface RequestCompletionHandler {
 
     void onComplete(ClientResponse response);
-
 }
--- a/core/src/main/scala/kafka/server/AlterIsrManager.scala
+++ b/core/src/main/scala/kafka/server/AlterIsrManager.scala
@@ -96,26 +96,25 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
       }
     }
 
-    class AlterIsrResponseHandler extends ControllerRequestCompletionHandler {
-      override def onComplete(response: ClientResponse): Unit = {
-        try {
-          val body = response.responseBody().asInstanceOf[AlterIsrResponse]
-          handleAlterIsrResponse(body, message.brokerEpoch, inflightAlterIsrItems)
-        } finally {
-          clearInflightRequests()
-        }
-      }
-
-      override def onTimeout(): Unit = {
-        throw new IllegalStateException("Encountered unexpected timeout when sending AlterIsr to the controller")
-      }
-    }
-
     debug(s"Sending AlterIsr to controller $message")
     // We will not timeout AlterISR request, instead letting it retry indefinitely
-    // until a response is received or the request is cancelled after receiving new LeaderAndIsr state.
+    // until a response is received, or a new LeaderAndIsr overwrites the existing isrState
+    // which causes the inflight requests to be ignored.
     controllerChannelManager.sendRequest(new AlterIsrRequest.Builder(message),
-      new AlterIsrResponseHandler, Long.MaxValue)
+      new ControllerRequestCompletionHandler {
+        override def onComplete(response: ClientResponse): Unit = {
+          try {
+            val body = response.responseBody().asInstanceOf[AlterIsrResponse]
+            handleAlterIsrResponse(body, message.brokerEpoch, inflightAlterIsrItems)
+          } finally {
+            clearInflightRequests()
+          }
+        }
+
+        override def onTimeout(): Unit = {
+          throw new IllegalStateException("Encountered unexpected timeout when sending AlterIsr to the controller")
+        }
+      }, Long.MaxValue)
   }
 
   private def buildRequest(inflightAlterIsrItems: Seq[AlterIsrItem]): AlterIsrRequestData = {
--- a/core/src/main/scala/kafka/server/AlterIsrManager.scala
+++ b/core/src/main/scala/kafka/server/AlterIsrManager.scala
@@ -96,11 +96,11 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
       }
     }
 
-    class AlterIsrResponseHandler extends BrokerToControllerRequestCompletionHandler {
+    class AlterIsrResponseHandler extends ControllerRequestCompletionHandler {
       override def onComplete(response: ClientResponse): Unit = {
         try {
           val body = response.responseBody().asInstanceOf[AlterIsrResponse]
-          handleAlterIsrResponse(body, message.brokerEpoch(), inflightAlterIsrItems)
+          handleAlterIsrResponse(body, message.brokerEpoch, inflightAlterIsrItems)
         } finally {
           clearInflightRequests()
         }
@@ -112,7 +112,8 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
     }
 
     debug(s"Sending AlterIsr to controller $message")
-    // We will not timeout AlterISR request, instead letting it retry indefinitely.
+    // We will not timeout AlterISR request, instead letting it retry indefinitely
+    // until a response is received or the request is cancelled after receiving new LeaderAndIsr state.
     controllerChannelManager.sendRequest(new AlterIsrRequest.Builder(message),
       new AlterIsrResponseHandler, Long.MaxValue)
   }
--- a/core/src/main/scala/kafka/server/AlterIsrManager.scala
+++ b/core/src/main/scala/kafka/server/AlterIsrManager.scala
@@ -88,20 +88,33 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
 
   private def sendRequest(inflightAlterIsrItems: Seq[AlterIsrItem]): Unit = {
     val message = buildRequest(inflightAlterIsrItems)
-    def responseHandler(response: ClientResponse): Unit = {
-      try {
-        val body = response.responseBody().asInstanceOf[AlterIsrResponse]
-        handleAlterIsrResponse(body, message.brokerEpoch(), inflightAlterIsrItems)
-      } finally {
-        // Be sure to clear the in-flight flag to allow future AlterIsr requests
-        if (!inflightRequest.compareAndSet(true, false)) {
-          throw new IllegalStateException("AlterIsr response callback called when no requests were in flight")
+
+    def clearInflightRequests(): Unit = {
+      // Be sure to clear the in-flight flag to allow future AlterIsr requests
+      if (!inflightRequest.compareAndSet(true, false)) {
+        throw new IllegalStateException("AlterIsr response callback called when no requests were in flight")
+      }
+    }
+
+    class AlterIsrResponseHandler extends BrokerToControllerRequestCompletionHandler {
+      override def onComplete(response: ClientResponse): Unit = {
+        try {
+          val body = response.responseBody().asInstanceOf[AlterIsrResponse]
+          handleAlterIsrResponse(body, message.brokerEpoch(), inflightAlterIsrItems)
+        } finally {
+          clearInflightRequests()
         }
+      }
+
+      override def onTimeout(): Unit = {
+        warn(s"Encountered request when sending AlterIsr to the controller")
       }
     }
 
     debug(s"Sending AlterIsr to controller $message")
-    controllerChannelManager.sendRequest(new AlterIsrRequest.Builder(message), responseHandler)
+    // We will not timeout AlterISR request, instead letting it retry indefinitely.
+    controllerChannelManager.sendRequest(new AlterIsrRequest.Builder(message),
+      new AlterIsrResponseHandler, Long.MaxValue)
   }
 
   private def buildRequest(inflightAlterIsrItems: Seq[AlterIsrItem]): AlterIsrRequestData = {
--- a/core/src/main/scala/kafka/server/AlterIsrManager.scala
+++ b/core/src/main/scala/kafka/server/AlterIsrManager.scala
@@ -107,7 +107,7 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
       }
 
       override def onTimeout(): Unit = {
-        warn(s"Encountered request when sending AlterIsr to the controller")
+        throw new IllegalStateException("Encountered unexpected timeout when sending AlterIsr to the controller")
       }
     }
 
--- a/core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
+++ b/core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
@@ -35,8 +35,15 @@ import scala.jdk.CollectionConverters._
 
 trait BrokerToControllerChannelManager {
 
-  // The retry deadline will only be checked after receiving a response. This means that in the worst case,
-  // the total timeout would be twice of the configured timeout.
+  /**
+   * Send request to the controller.
+   *
+   * @param request         The request to be sent.
+   * @param callback        Request completion callback.
+   * @param retryDeadlineMs The retry deadline which will only be checked after receiving a response.
+   *                        This means that in the worst case, the total timeout would be twice of
+   *                        the configured timeout.
+   */
   def sendRequest(request: AbstractRequest.Builder[_ <: AbstractRequest],
                   callback: ControllerRequestCompletionHandler,
                   retryDeadlineMs: Long): Unit
@@ -140,7 +147,7 @@ abstract class ControllerRequestCompletionHandler extends RequestCompletionHandl
 
   /**
    * Fire when the request transmission time passes the caller defined deadline on the channel queue.
-   * This is different from the original request's timeout.
+   * It covers the total waiting time including retries which might be the result of individual request timeout.
    */
   def onTimeout(): Unit
 }
--- a/core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
+++ b/core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
@@ -35,7 +35,8 @@ import scala.jdk.CollectionConverters._
 
 trait BrokerToControllerChannelManager {
   def sendRequest(request: AbstractRequest.Builder[_ <: AbstractRequest],
-                  callback: RequestCompletionHandler): Unit
+                  callback: BrokerToControllerRequestCompletionHandler,
+                  requestTimeout: Long): Unit
 
   def start(): Unit
 
@@ -125,15 +126,25 @@ class BrokerToControllerChannelManagerImpl(metadataCache: kafka.server.MetadataC
   }
 
   override def sendRequest(request: AbstractRequest.Builder[_ <: AbstractRequest],
-                           callback: RequestCompletionHandler): Unit = {
-    requestQueue.put(BrokerToControllerQueueItem(request, callback))
+                           callback: BrokerToControllerRequestCompletionHandler,
+                           requestTimeout: Long): Unit = {
+    requestQueue.put(BrokerToControllerQueueItem(request, callback, time.milliseconds() + requestTimeout))
     requestThread.wakeup()
   }
 
 }
 
+abstract class BrokerToControllerRequestCompletionHandler extends RequestCompletionHandler {
+
+  /**
+   * Fire when the request transmission hits a fatal exception.
+   */
+  def onTimeout(): Unit
+}
+
 case class BrokerToControllerQueueItem(request: AbstractRequest.Builder[_ <: AbstractRequest],
-                                       callback: RequestCompletionHandler)
+                                       callback: BrokerToControllerRequestCompletionHandler,
+                                       deadlineMs: Long)
 
 class BrokerToControllerRequestThread(networkClient: KafkaClient,
                                       metadataUpdater: ManualMetadataUpdater,
@@ -165,7 +176,9 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
   }
 
   private[server] def handleResponse(request: BrokerToControllerQueueItem)(response: ClientResponse): Unit = {
-    if (response.wasDisconnected()) {
+    if (isTimedOut(request)) {
+      request.callback.onTimeout()
+    } else if (response.wasDisconnected()) {
       activeController = None
       requestQueue.putFirst(request)
     } else if (response.responseBody().errorCounts().containsKey(Errors.NOT_CONTROLLER)) {
@@ -176,6 +189,10 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
     } else {
       request.callback.onComplete(response)
     }
+  }
+
+  private def isTimedOut(request: BrokerToControllerQueueItem): Boolean = {
+    time.milliseconds() > request.deadlineMs
   }
 
   private[server] def backoff(): Unit = pause(100, TimeUnit.MILLISECONDS)
--- a/core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
+++ b/core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
@@ -34,8 +34,11 @@ import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
 trait BrokerToControllerChannelManager {
+
+  // The retry deadline will only be checked after receiving a response. This means that in the worst case,
+  // the total timeout would be twice of the configured timeout.
   def sendRequest(request: AbstractRequest.Builder[_ <: AbstractRequest],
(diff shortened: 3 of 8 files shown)


core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -208,7 +209,9 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
   }
 
   private[server] def handleResponse(request: BrokerToControllerQueueItem)(response: ClientResponse): Unit = {
-    if (response.wasDisconnected()) {
+    if (isTimedOut(response)) {
+      debug(s"The request failed to send to the controller after timeout ${clientRequestTimeout} ms: $response")
+    } else if (response.wasDisconnected()) {
```
chia7712, line 191: (The following question is not related to this PR. just curious)

Is it possible to encounter authentication error when forwarding? If so, is it necessary to avoid retrying such (error) request?

abbccdda: Thanks, for authorization related exception, we could add the logic to stop retry as well.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -208,7 +209,9 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
   }
 
   private[server] def handleResponse(request: BrokerToControllerQueueItem)(response: ClientResponse): Unit = {
-    if (response.wasDisconnected()) {
```
mumrah, line 168: So before, if the request timed out, it would be marked as "disconnected"?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -208,7 +209,9 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
...
+    if (isTimedOut(response)) {
+      debug(s"The request failed to send to the controller after timeout ${clientRequestTimeout} ms: $response")
```
mumrah: Do we need to call the request callback here so we can propagate the error up to the client?

mumrah (commented):
I imagine timing out for forwarded requests is fine, but what about AlterIsr? 

Thinking out loud: If a leader gets partitioned off from the controller for too long, we'll eventually timeout and give up on an AlterIsr request. However, at this point the leader is already in a pending ISR state. If the controller never responds, we'll be stuck in that state until we get new metadata. Since this will only happen if the controller is down or we have a network partition, I think we can safely assume that once the controller is available again we should get new metadata. WDYT?

Also related to my comment on L213, AlterIsrManager is expecting the callback to always get called in all cases. If it doesn't, it won't clear the in-flight flag and will be stuck unable to make any new AlterIsr requests.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
abbccdda: I think so

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
abbccdda: Yea, I think we could actually fire the callback here.

clients/src/main/java/org/apache/kafka/clients/RequestCompletionHandler.java
```diff
@@ -24,4 +24,11 @@
 
     void onComplete(ClientResponse response);
 
+    /**
+     * Fire when the request transmission hits a fatal exception.
+     *
+     * @param exception the thrown exception
+     */
+    default void onFailure(RuntimeException exception) {
```
hachikuji: I am not so sure about adding this to `RequestCompletionHandler`. This interface is mainly serving `NetworkClient`, but there are no changes to `NetworkClient` here.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -178,6 +183,11 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
     }
   }
 
+  private def isTimedOut(response: ClientResponse): Boolean = {
+    val requestCreatedTime = response.receivedTimeMs() - response.requestLatencyMs()
```
hachikuji: The time we need to track should begin when the request is submitted to the manager. This looks like it is just checking how long an individual request is taking, which the request timeout in `NetworkClient` already takes care of.

hachikuji (commented):
Thanks, left a couple quick comments.

clients/src/main/java/org/apache/kafka/clients/RequestCompletionHandler.java
abbccdda: We do have a case in `ConsumerNetworkClient` which adds an `onFailure` callback. To me it makes sense to include it as part of the RequestCompletionHandler interface.

hachikuji: Here is the interface for `KafkaClient`:
```
 ClientRequest newClientRequest(String nodeId,
                                   AbstractRequest.Builder<?> requestBuilder,
                                   long createdTimeMs,
                                   boolean expectResponse,
                                   int requestTimeoutMs,
                                   RequestCompletionHandler callback);
```
It is misleading to add an `onFailure` callback to `RequestCompletionHandler` if it is not going to be used by `KafkaClient` implementations such as `NetworkClient`. The usage in `ConsumerNetworkClient` is different because it is internal. In general, we should avoid leaking implementation details up to the interfaces.

abbccdda: I see, if this is the case, we need a customized completion handler for both forwarding and AlterISR IMHO.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
abbccdda: Is it legitimate to compare with requestTimeout here since we actually measure the request buffered time on the broker-to-controller channel queue? Should we introduce a new timeout config here somehow?

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -88,20 +88,33 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
...
+          val body = response.responseBody().asInstanceOf[AlterIsrResponse]
+          handleAlterIsrResponse(body, message.brokerEpoch(), inflightAlterIsrItems)
+        } finally {
+          clearInflightRequests()
         }
       }
+
+      override def onFailure(exception: RuntimeException): Unit = {
+        error(s"Encountered $exception when sending AlterIsr to the controller, clearing all pending states")
+        unsentIsrUpdates.clear()
```
hachikuji: There may be requests that never had a chance to be sent?

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -88,20 +88,33 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
...
+      override def onComplete(response: ClientResponse): Unit = {
+        try {
+          val body = response.responseBody().asInstanceOf[AlterIsrResponse]
+          handleAlterIsrResponse(body, message.brokerEpoch(), inflightAlterIsrItems)
+        } finally {
+          clearInflightRequests()
         }
       }
+
+      override def onFailure(exception: RuntimeException): Unit = {
```
hachikuji: Don't we need to invoke the callbacks?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -178,6 +194,10 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
     }
   }
 
+  private def isTimedOut(request: BrokerToControllerQueueItem): Boolean = {
+    (time.milliseconds() - request.firstEnqueuedTimeMs) > clientRequestTimeout
```
hachikuji: We do not want to timeout requests from the broker such as `AlterIsr`. We only want this for client requests.

core/src/main/scala/kafka/server/AlterIsrManager.scala
abbccdda: Yea, I'm not sure whether we should clear all the pending updates here or not, just saw the comment as `Regardless of callback outcome, we need to clear from the unsent updates map to unblock further updates`, so thinking maybe we could just remove all pending updates here.

core/src/main/scala/kafka/server/AlterIsrManager.scala
abbccdda: You mean `handleAlterIsrResponse`?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -165,7 +176,9 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
   }
 
   private[server] def handleResponse(request: BrokerToControllerQueueItem)(response: ClientResponse): Unit = {
-    if (response.wasDisconnected()) {
+    if (isTimedOut(request)) {
```
hachikuji: We check for timeouts only after receiving a response. I guess this means that in the worst case, the total timeout would be request.timeout*2. This is probably not a big deal, but maybe worth documenting in a comment somewhere.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -35,7 +35,8 @@ import scala.jdk.CollectionConverters._
 
 trait BrokerToControllerChannelManager {
   def sendRequest(request: AbstractRequest.Builder[_ <: AbstractRequest],
-                  callback: RequestCompletionHandler): Unit
+                  callback: BrokerToControllerRequestCompletionHandler,
+                  requestTimeout: Long): Unit
```
hachikuji: Perhaps we could use a name like `retryTimeout` to distinguish this from the request timeout which only applies to individual requests. Alternatively we could let the caller provide the retry deadline explicitly. This would save the need for the extra `time.milliseconds` call.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -178,6 +191,10 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
     }
   }
 
+  private def isTimedOut(request: BrokerToControllerQueueItem): Boolean = {
```
hachikuji: nit: `hasTimedOut`?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -125,15 +126,25 @@ class BrokerToControllerChannelManagerImpl(metadataCache: kafka.server.MetadataC
...
+    requestQueue.put(BrokerToControllerQueueItem(request, callback, time.milliseconds() + requestTimeout))
     requestThread.wakeup()
   }
 
 }
 
+abstract class BrokerToControllerRequestCompletionHandler extends RequestCompletionHandler {
+
+  /**
+   * Fire when the request transmission hits a fatal exception.
```
hachikuji: This comment needs to be updated.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -125,15 +126,25 @@ class BrokerToControllerChannelManagerImpl(metadataCache: kafka.server.MetadataC
   }
 
   override def sendRequest(request: AbstractRequest.Builder[_ <: AbstractRequest],
-                           callback: RequestCompletionHandler): Unit = {
-    requestQueue.put(BrokerToControllerQueueItem(request, callback))
+                           callback: BrokerToControllerRequestCompletionHandler,
+                           requestTimeout: Long): Unit = {
+    requestQueue.put(BrokerToControllerQueueItem(request, callback, time.milliseconds() + requestTimeout))
```
hachikuji: Won't this overflow with `requestTimeout` set to `Long.MaxValue`. Do we have any test cases?

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -88,20 +88,33 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
...
+        try {
+          val body = response.responseBody().asInstanceOf[AlterIsrResponse]
+          handleAlterIsrResponse(body, message.brokerEpoch(), inflightAlterIsrItems)
+        } finally {
+          clearInflightRequests()
         }
       }
+
+      override def onTimeout(): Unit = {
+        warn(s"Encountered request when sending AlterIsr to the controller")
```
hachikuji: If we don't expect this, maybe we should just raise an exception. Alternatively if we are not going to fail, perhaps we should go ahead and invoke the callbacks.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -178,6 +191,10 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
     }
   }
 
+  private def isTimedOut(request: BrokerToControllerQueueItem): Boolean = {
+    time.milliseconds() > request.deadlineMs
```
hachikuji: Maybe we can avoid this call to `time.milliseconds` and use `ClientResponse.receivedTimeMs`?

core/src/main/scala/kafka/server/ForwardingManager.scala
```diff
@@ -44,26 +46,34 @@ class ForwardingManager(channelManager: BrokerToControllerChannelManager) extend
...
+          debug(s"Forwarded request $request failed with an error in the envelope response $envelopeError")
+          requestBody.getErrorResponse(Errors.UNKNOWN_SERVER_ERROR.exception)
+        } else {
+          parseResponse(envelopeResponse.responseData, requestBody, request.header)
+        }
+        responseCallback(response)
+      }
+
+      override def onTimeout(): Unit = {
+        error(s"Forwarding of the request $request failed due to timeout exception")
```
hachikuji: I think this should be debug. Users will already have visibility into the error through the request log and the error metrics. There's probably a stronger case to increase the level for the unknown error case in `onComplete` above, but I'm fine letting them both be debug.

core/src/main/scala/kafka/server/AlterIsrManager.scala
abbccdda: Not sure we could do the callback here, since the request failed already. Maybe just do nothing here?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
abbccdda: Good catch!

core/src/main/scala/kafka/server/AlterIsrManager.scala
hachikuji: Why not raise an exception?

abbccdda: That's what I decide to do eventually.

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -88,20 +88,33 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
...
+        try {
+          val body = response.responseBody().asInstanceOf[AlterIsrResponse]
+          handleAlterIsrResponse(body, message.brokerEpoch(), inflightAlterIsrItems)
+        } finally {
+          clearInflightRequests()
         }
       }
+
+      override def onTimeout(): Unit = {
+        throw new IllegalStateException("Encountered unexpected timeout when sending AlterIsr to the controller")
```
mumrah: Where will this exception get raised? I'm guessing the BrokerToControllerChannelManager thread?

If we timeout and give up on reading a response, I think we should still clear the `inflightRequest` flag. Otherwise AlterIsrManager will be stuck in-flight and be unable to send any more requests.

mumrah: We also might want to iterate through the in-flight ISR (`inflightAlterIsrItems`) and complete their callbacks with an error. Otherwise, these partitions will not know to retry their ISR changes. They will eventually get new metadata via LeaderAndIsr, but it would be good to prevent them from getting into a bad state.

Something like
```scala
inflightAlterIsrItems.foreach { item => item.callback.apply(Left(Errors.REQUEST_TIMED_OUT)) }
```

I don't _think_ we need this in the `finally` block at L104 since `handleAlterIsrResponse` should be pretty robust against raising an exception

abbccdda: We set the timeout of AlterIsrManager to infinity, so it should never happen. See comment from Jason: https://github.com/apache/kafka/pull/9564#discussion_r527326231

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -88,20 +88,33 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
...
+      if (!inflightRequest.compareAndSet(true, false)) {
+        throw new IllegalStateException("AlterIsr response callback called when no requests were in flight")
+      }
+    }
+
+    class AlterIsrResponseHandler extends BrokerToControllerRequestCompletionHandler {
+      override def onComplete(response: ClientResponse): Unit = {
+        try {
+          val body = response.responseBody().asInstanceOf[AlterIsrResponse]
+          handleAlterIsrResponse(body, message.brokerEpoch(), inflightAlterIsrItems)
```
hachikuji: nit: drop parenthesis after `brokerEpoch`

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -88,20 +88,33 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
...
       }
+
+      override def onTimeout(): Unit = {
+        throw new IllegalStateException("Encountered unexpected timeout when sending AlterIsr to the controller")
+      }
     }
 
     debug(s"Sending AlterIsr to controller $message")
-    controllerChannelManager.sendRequest(new AlterIsrRequest.Builder(message), responseHandler)
+    // We will not timeout AlterISR request, instead letting it retry indefinitely.
```
hachikuji: Perhaps add some more detail: "... letting it retry indefinitely until a response is received or the request is cancelled after receiving new `LeaderAndIsr` state".

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -125,15 +126,24 @@ class BrokerToControllerChannelManagerImpl(metadataCache: kafka.server.MetadataC
...
-                           callback: RequestCompletionHandler): Unit = {
-    requestQueue.put(BrokerToControllerQueueItem(request, callback))
+                           callback: BrokerToControllerRequestCompletionHandler,
+                           retryDeadlineMs: Long): Unit = {
+    requestQueue.put(BrokerToControllerQueueItem(request, callback, retryDeadlineMs))
     requestThread.wakeup()
   }
+}
+
+abstract class BrokerToControllerRequestCompletionHandler extends RequestCompletionHandler {
```
hachikuji: How about `ControllerRequestCompletionHandler`?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -125,15 +126,24 @@ class BrokerToControllerChannelManagerImpl(metadataCache: kafka.server.MetadataC
...
 
+  /**
+   * Fire when the request transmission hits timeout.
```
hachikuji: Can we document the difference between this and the request timeout?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -178,6 +190,12 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
     }
   }
 
+  // The timeout will only be checked after receiving a response. This means that in the worst case,
```
hachikuji: Can we move this comment to the doc for `sendRequest`?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
hachikuji: Note that we handle authentication exceptions in `InterBrokerSendThread`. We log a message and mark the response as "disconnected." I think it is reasonable to keep retrying in case the broker's configuration changes and so that the error messages will remain persistent in the logs.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -34,8 +34,12 @@ import scala.collection.mutable
 import scala.jdk.CollectionConverters._
 
 trait BrokerToControllerChannelManager {
+
+  // The retry deadline will only be checked after receiving a response. This means that in the worst case,
```
hachikuji: nit: turn into scaladoc

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -125,15 +129,25 @@ class BrokerToControllerChannelManagerImpl(metadataCache: kafka.server.MetadataC
...
+    requestQueue.put(BrokerToControllerQueueItem(request, callback, retryDeadlineMs))
     requestThread.wakeup()
   }
+}
+
+abstract class ControllerRequestCompletionHandler extends RequestCompletionHandler {
 
+  /**
+   * Fire when the request transmission time passes the caller defined deadline on the channel queue.
+   * This is different from the original request's timeout.
```
hachikuji: Can we explain how it is different? We want to emphasize that it covers the total time including retries which might be the result of request timeouts.

abbccdda:
Hit unrelated stream test failure, retest

core/src/main/scala/kafka/server/AlterIsrManager.scala
mumrah: Thanks, makes sense.

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -88,20 +88,34 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
...
-          throw new IllegalStateException("AlterIsr response callback called when no requests were in flight")
+
+    def clearInflightRequests(): Unit = {
+      // Be sure to clear the in-flight flag to allow future AlterIsr requests
+      if (!inflightRequest.compareAndSet(true, false)) {
+        throw new IllegalStateException("AlterIsr response callback called when no requests were in flight")
+      }
+    }
+
+    class AlterIsrResponseHandler extends ControllerRequestCompletionHandler {
```
mumrah: nit: could we inline this as an anonymous class down below rather than defining it here? Just seems a little odd to define it like this and then calling in to the containing class for `inflightRequest`.

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -88,20 +88,34 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
...
+
+      override def onTimeout(): Unit = {
+        throw new IllegalStateException("Encountered unexpected timeout when sending AlterIsr to the controller")
+      }
     }
 
     debug(s"Sending AlterIsr to controller $message")
-    controllerChannelManager.sendRequest(new AlterIsrRequest.Builder(message), responseHandler)
+    // We will not timeout AlterISR request, instead letting it retry indefinitely
+    // until a response is received or the request is cancelled after receiving new LeaderAndIsr state.
```
mumrah: I don't think we actually have the ability to cancel in-flight requests. The problem is that many partitions can have ISR changes batched together in a single request. So we couldn't cancel one without cancelling them all. 

A new LeaderAndIsr arriving will just overwrite the existing `isrState` in Partition which will effectively cause any in-flight AlterIsr response to be ignored (for that partition)

abbccdda: Updated the comment

hachikuji (approved):
LGTM

mumrah (approved):
Looks good, thanks @abbccdda!

PR #9566, opened 2020-11-05
KAFKA-10618: Update to Uuid class

jolshan:
As decided in KIP-516, the UUID class has been named Uuid. This PR changes all instances of org.apache.kafka.common.UUID to org.apache.kafka.common.Uuid.

It also modifies the Uuid class so that it no longer wraps a java.util.UUID object. Now it simply stores two longs. 


jolshan:
@cmccabe @ijuma 
Let me know if this looks ok.

clients/src/main/java/org/apache/kafka/common/protocol/types/Struct.java
```diff
@@ -266,12 +266,12 @@ public Long getLong(String name) {
...
 
-    public UUID getUUID(BoundField field) {
-        return (UUID) get(field);
+    public Uuid getUUID(BoundField field) {
+        return (Uuid) get(field);
     }
 
-    public UUID getUUID(String name) {
-        return (UUID) get(name);
+    public Uuid getUUID(String name) {
```
ijuma: Should we update the method name too?

clients/src/main/java/org/apache/kafka/common/Uuid.java
```diff
@@ -20,101 +20,100 @@
...
      */
     @Override
     public String toString() {
-        return Base64.getUrlEncoder().withoutPadding().encodeToString(getBytesFromUuid(uuid));
+        return Base64.getUrlEncoder().withoutPadding().encodeToString(getBytesFromUuid());
     }
 
     /**
-     * Creates a UUID based on a base64 string encoding used in the toString() method.
+     * Creates a Uuid based on a base64 string encoding used in the toString() method.
```
ijuma: Isn't it better to keep `UUID` in javadoc when we're referring to the concept?

ijuma (commented):
Thanks for the PR. A couple of minor comments below.

clients/src/test/java/org/apache/kafka/common/UuidTest.java
```diff
@@ -21,50 +21,50 @@
...
         assertEquals(id.getMostSignificantBits(), 34L);
         assertEquals(id.getLeastSignificantBits(), 98L);
     }
 
     @Test
-    public void testUUIDEquality() {
-        UUID id1 = new UUID(12L, 13L);
-        UUID id2 = new UUID(12L, 13L);
-        UUID id3 = new UUID(24L, 38L);
+    public void testUuidEquality() {
```
ijuma, line 35: Can we add a test that verifies that the `hashCode` is the same for our `Uuid` and Java's `UUID`? Or do we have that already?

jolshan: We don't have that yet. I also didn't specify this behavior. Should I?

ijuma: We don't have to specify it, but it would be good to ensure we have a test for the actual hashCode we're implementing. At the moment, we are only verifying that the hashCode is the same for two equal UUIDs and different for two unequal UUIDs. One option would be to have a few tests where we verify that the result is what we expect it to be.

ijuma (approved):
LGTM

PR #9568, opened 2020-11-06
KAFKA-10689: fix windowed FKJ topology and put checks in assignor to avoid infinite loops

ableegoldman:
Two pieces:
1) Fix bug in StreamSinkNode#writeToTopology to make sure it always calls the `addSink` overload with the specific topic name, when it exists, so that this topic gets tracked in the InternalTopologyBuilder's `nodeToSinkTopic` map. The sink topics are used by the StreamsPartitionAssignor to resolve the upstream subtopology of a repartition source topic, for whom that repartition topic will be a sink. Without this information the SPA gets stuck permanently during a rebalance
2) Improve the SPA's `setRepartitionTopicMetadataNumberOfPartitions()` method to break out of the loop if we aren't making any progress, to avoid infinitely looping if we ever have another bug like KAFKA-10689. If the SPA hasn't updated the known partition numbers for any repartition topic in the current outer loop, then we know that it's stuck and should throw a TaskAssignmentException to shut down the application

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
@@ -532,8 +532,8 @@ public class StreamsPartitionAssignor implements ConsumerPartitionAssignor, Conf
             numPartitionsNeeded = false;
 
             for (final TopicsInfo topicsInfo : topicGroups.values()) {
-                for (final String topicName : topicsInfo.repartitionSourceTopics.keySet()) {
-                    final Optional<Integer> maybeNumPartitions = repartitionTopicMetadata.get(topicName)
+                for (final String repartitionSourceTopic : topicsInfo.repartitionSourceTopics.keySet()) {
+                    final Optional<Integer> maybeNumPartitions = repartitionTopicMetadata.get(repartitionSourceTopic)
                                                                      .numberOfPartitions();
                     Integer numPartitions = null;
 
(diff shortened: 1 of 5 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -575,17 +577,22 @@ private void setRepartitionTopicMetadataNumberOfPartitions(final Map<String, Int
...
                         } else {
-                            repartitionTopicMetadata.get(topicName).setNumberOfPartitions(numPartitions);
+                            repartitionTopicMetadata.get(repartitionSourceTopic).setNumberOfPartitions(numPartitions);
+                            progressMadeThisIteration = true;
                         }
                     }
                 }
             }
-        } while (numPartitionsNeeded);
+        } while (numPartitionsNeeded && progressMadeThisIteration);
```
mjsax: Would it be easier to do
```
  if (!progressMadeThisIteration) {
    throw new TaskAssignmentException("Failed to compute number of partitions for all repartition topics");
  }
} while (numPartitionsNeeded);
```
(and don't have the `if (numPartitionsNeeded)` after the while loop)

I guess both is correct and it might be a matter of taste. Also don't have a strong opinion about it. It's just an idea.

streams/src/test/java/org/apache/kafka/streams/integration/InternalTopicIntegrationTest.java
```diff
@@ -135,6 +141,37 @@ private Admin createAdminClient() {
         return Admin.create(adminClientConfig);
     }
 
+    /*
+     * This test just ensures that that the assignor does not get stuck during partition number resolution
+     *  for internal repartition topics. See KAFKA-10689
+     */
```
mjsax, line 147: Do we need this commet?

mjsax (approved):
Just some nits/questions. Overall LGTM. Feel free to merge.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
ableegoldman: Mm yeah that does sound cleaner, will do

streams/src/test/java/org/apache/kafka/streams/integration/InternalTopicIntegrationTest.java
ableegoldman: I'd prefer to leave it in because otherwise the test looks totally out of place and doesn't seem to test anything. Historical context can be useful 🙂

ableegoldman: To quote a [very salty developer](https://pastebin.com/PTLeWhc2) in the comments of the leaked Windows Media Player source code:

```
/* The reason for requiring the following test is now lost
         * in the mists of time.  Now this app is 32-bit, these
         * bogus timer callbacks (if they really do still occur)
         * could be 16-bit, so we need to add yet more ugliness
         * in the form of assembler to an app which is already
         * hardly a paragon of pulchritude.
         *
         * A plea:
         *
         * If you add some obscure code such as below, to this or
         * any other app, even if it has only the teeniest chance
         * of being less blindingly obvious to someone else than
         * it is to you at the time of writing, please please please
         * add a f***ing comment.
         *
         * Respectfully,
         * A Developer
         */
```
😂

ableegoldman:
All tests pass, but the build overall is broken due to failure of the new `Travis CI - Pull Request` thing. I can't find anything in the results that indicate an actual problem, so I'm just going to merge this

ableegoldman:
Merged to trunk. Will cherrypick back to 2.7 once the ongoing release completes

ableegoldman:
Cherrypicked to 2.7

mjsax:
Back ported to `2.6` via #10020

PR #9570, opened 2020-11-06
KAFKA-9274: Handle TimeoutException on commit

mjsax:
- part of KIP-572
 - when KafkaStreams commit a task, a TimeoutException should not kill
   the thread but `task.timeout.ms` should be triggered and the commit
   should be retried in the next loop

Call for review @vvcephei

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/errors/TaskTimeoutExceptions.java
+++ b/streams/src/main/java/org/apache/kafka/streams/errors/TaskTimeoutExceptions.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.errors;
+
+import org.apache.kafka.common.errors.TimeoutException;
+import org.apache.kafka.streams.processor.internals.Task;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+public class TaskTimeoutExceptions extends StreamsException {
+    private static final long serialVersionUID = 1L;
+
+    private final TimeoutException timeoutException;
+    private final Map<Task, TimeoutException> exceptions;
+
+    public TaskTimeoutExceptions() {
+        super("");
+        timeoutException = null;
+        exceptions = new HashMap<>();
+    }
+
+    public TaskTimeoutExceptions(final TimeoutException timeoutException) {
+        super("");
+        this.timeoutException = timeoutException;
+        exceptions = null;
+    }
+
+    public void recordException(final Task task,
+                                final TimeoutException timeoutException) {
+        Objects.requireNonNull(exceptions)
+            .put(task, timeoutException);
+    }
+
+    public Map<Task, TimeoutException> exceptions() {
+        return exceptions;
+    }
+
+    public TimeoutException timeoutException() {
+        return timeoutException;
+    }
+
+}
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
@@ -164,9 +164,9 @@ public class InternalTopicManager {
                                 "Error message was: {}", topicName, cause.toString());
                             throw new StreamsException(String.format("Could not create topic %s.", topicName), cause);
                         }
-                    } catch (final TimeoutException retryableException) {
+                    } catch (final TimeoutException retriableException) {
                         log.error("Creating topic {} timed out.\n" +
-                            "Error message was: {}", topicName, retryableException.toString());
+                            "Error message was: {}", topicName, retriableException.toString());
                     }
                 }
             }
@@ -234,10 +234,10 @@ public class InternalTopicManager {
                         "Error message was: {}", topicName, cause.toString());
                     throw new StreamsException(String.format("Could not create topic %s.", topicName), cause);
                 }
-            } catch (final TimeoutException retryableException) {
+            } catch (final TimeoutException retriableException) {
                 tempUnknownTopics.add(topicName);
                 log.debug("Describing topic {} (to get number of partitions) timed out.\n" +
-                    "Error message was: {}", topicName, retryableException.toString());
+                    "Error message was: {}", topicName, retriableException.toString());
             }
         }
 
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
@@ -16,7 +16,6 @@
  */
 package org.apache.kafka.streams.processor.internals;
 
-import java.util.concurrent.atomic.AtomicReference;
 import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.KafkaException;
@@ -50,6 +49,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
 
 public class RecordCollectorImpl implements RecordCollector {
     private final static String SEND_EXCEPTION_MESSAGE = "Error encountered sending record to topic %s for task %s due to:%n%s";
@@ -111,6 +111,8 @@ public class RecordCollectorImpl implements RecordCollector {
             try {
                 partitions = streamsProducer.partitionsFor(topic);
             } catch (final KafkaException e) {
+                // TODO: KIP-572 need to handle `TimeoutException`
+                // -> should we throw a `TaskCorruptedException` for this case to reset the task and retry (including triggering `task.timeout.ms`) ?
                 // here we cannot drop the message on the floor even if it is a transient timeout exception,
                 // so we treat everything the same as a fatal exception
                 throw new StreamsException("Could not determine the number of partitions for topic '" + topic +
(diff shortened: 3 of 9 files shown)


streams/src/main/java/org/apache/kafka/streams/errors/TaskTimeoutExceptions.java
```diff
@@ -0,0 +1,58 @@
...
+package org.apache.kafka.streams.errors;
+
+import org.apache.kafka.common.errors.TimeoutException;
+import org.apache.kafka.streams.processor.internals.Task;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+public class TaskTimeoutExceptions extends StreamsException {
```
mjsax, line 26: If we commit multiple tasks individually (ie, eos-alpha), we use this class as an "exception container" to track the TimeoutException for each failed task individually.

To simplify the caller code, we also wrap a single `TimeoutException` if we commit all tasks at once (at-least-once, eos-beta)

streams/src/main/java/org/apache/kafka/streams/processor/internals/InternalTopicManager.java
```diff
@@ -164,9 +164,9 @@ public InternalTopicManager(final Time time,
                                 "Error message was: {}", topicName, cause.toString());
                             throw new StreamsException(String.format("Could not create topic %s.", topicName), cause);
                         }
-                    } catch (final TimeoutException retryableException) {
+                    } catch (final TimeoutException retriableException) {
```
mjsax, line 167: Just fixing some naming issue.

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
```diff
@@ -111,6 +111,8 @@ public void initialize() {
             try {
                 partitions = streamsProducer.partitionsFor(topic);
             } catch (final KafkaException e) {
+                // TODO: KIP-572 need to handle `TimeoutException`
+                // -> should we throw a `TaskCorruptedException` for this case to reset the task and retry (including triggering `task.timeout.ms`) ?
```
mjsax, line 115: This is an open question. Input is welcome. I would like to tackle it in a follow up PR.

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
```diff
@@ -202,6 +204,10 @@ private void recordSendError(final String topic, final Exception exception, fina
                 "indicating the task may be migrated out";
             sendException.set(new TaskMigratedException(errorMessage, exception));
         } else {
+            // TODO: KIP-572 handle `TimeoutException extends RetriableException`
```
mjsax, line 207: As above: This is an open question. Input is welcome. I would like to tackle it in a follow up PR.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -570,6 +575,11 @@ void handleRevocation(final Collection<TopicPartition> revokedPartitions) {
         // so we would capture any exception and throw
         try {
             commitOffsetsOrTransaction(consumedOffsetsPerTask);
+        } catch (final TaskTimeoutExceptions taskTimeoutExceptions) {
```
mjsax, line 578: If a task is revoked, we don't need to trigger `task.timeout.ms` but can re-throw the `TimeoutException` to trigger a "dirty close" instead.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -903,6 +913,15 @@ void shutdown(final boolean clean) {
                         tasksToCloseClean.remove(task);
                     }
                 }
+            } catch (final TaskTimeoutExceptions taskTimeoutExceptions) {
```
mjsax, line 916: During shutdown, we don't need to trigger `task.timeout.ms` but can re-throw the `TimeoutException` to trigger a "dirty close" instead.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -997,7 +1016,7 @@ void shutdown(final boolean clean) {
 
     // For testing only.
     int commitAll() {
-        return commit(tasks.values());
+        return commit(new HashSet<>(tasks.values()));
```
mjsax, line 1019: Need to make a copy (as all production calls do, too).

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -817,13 +817,15 @@ private void initializeMetadata() {
...
-                    e.toString(),
-                    ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG);
-
-            throw e;
+        } catch (final TimeoutException timeoutException) {
+            log.warn(
+                "Encountered {} while trying to fetch committed offsets, will retry initializing the metadata in the next loop." +
+                    "\nConsider overwriting consumer config {} to a larger value to avoid timeout errors",
+                time.toString(),
+                ConsumerConfig.DEFAULT_API_TIMEOUT_MS_CONFIG);
```
vvcephei, line 825: It might still be nice to see the stacktrace here (even if it also gets logged elsewhere). If you want to do it, don't forget you have to change to using `String.format` for the variable substitution.

I don't feel strongly in this case, so I'll defer to you whether you want to do this or not.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsProducer.java
```diff
@@ -262,13 +263,21 @@ void commitTransaction(final Map<TopicPartition, OffsetAndMetadata> offsets,
...
         if (!eosEnabled()) {
             throw new IllegalStateException(formatException("Exactly-once is not enabled"));
         }
         if (transactionInFlight) {
             try {
                 producer.abortTransaction();
+            } catch (final TimeoutException logAndSwallow) {
+                // no need to re-throw because we abort a TX only if we close a task dirty,
+                // and thus `task.timeout.ms` does not apply
+                log.debug(
```
vvcephei: I'm wondering if we should make this `info` or `warn` level. It doesn't seem like it would be very verbose, and it might be nice to see by default because it will have secondary effects later on when we try to start a new transaction, but get blocked.

But I also don't feel strongly about it, so I leave it to your discretion.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -1029,20 +1048,40 @@ int commit(final Collection<Task> tasksToCommit) {
...
+                        uncommittedTasks.add(task);
+                    }
+                }
+            }
 
             for (final Task task : tasksToCommit) {
-                if (task.commitNeeded()) {
+                if (!uncommittedTasks.contains(task)) {
                     ++committed;
                     task.postCommit(false);
```
vvcephei, line 1084: maybe we should move `clearTaskTimeout` here, in case some of the tasks timed out, but not all?

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -1029,20 +1048,40 @@ int commit(final Collection<Task> tasksToCommit) {
             return -1;
         } else {
             int committed = 0;
-            final Map<TaskId, Map<TopicPartition, OffsetAndMetadata>> consumedOffsetsAndMetadataPerTask = new HashMap<>();
-            for (final Task task : tasksToCommit) {
+            final Map<Task, Map<TopicPartition, OffsetAndMetadata>> consumedOffsetsAndMetadataPerTask = new HashMap<>();
+            final Iterator<Task> it = tasksToCommit.iterator();
```
vvcephei: I was initially worried about potential side-effects of removing from the input collection, but on second thought, maybe it's reasonable (considering the usage of this method) to assume that the input collection is always a copy already.

Still, it looks like we could actually simplify the implementation a little by _not_ reusing the input collection but instead only relying on the key set of `consumedOffsetsAndMetadataPerTask` as the list of "tasks to commit".

Here's what I'm thinking:
```java
int committed = 0;
            final Map<Task, Map<TopicPartition, OffsetAndMetadata>> consumedOffsetsAndMetadataPerTask = new HashMap<>();
            for (final Task task : tasksToCommit) {
                if (task.commitNeeded()) {
                    final Map<TopicPartition, OffsetAndMetadata> offsetAndMetadata = task.prepareCommit();
                    if (task.isActive()) {
                        consumedOffsetsAndMetadataPerTask.put(task, offsetAndMetadata);
                    }
                }
            }

            final Set<Task> uncommittedTasks = new HashSet<>();
            try {
                commitOffsetsOrTransaction(consumedOffsetsAndMetadataPerTask);
            } catch (final TaskTimeoutExceptions taskTimeoutExceptions) {
                final TimeoutException timeoutException = taskTimeoutExceptions.timeoutException();
                if (timeoutException != null) {
                    consumedOffsetsAndMetadataPerTask.keySet().forEach(t -> t.maybeInitTaskTimeoutOrThrow(time.milliseconds(), timeoutException));
                    uncommittedTasks.addAll(tasksToCommit);
                } else {
                    for (final Map.Entry<Task, TimeoutException> timeoutExceptions : taskTimeoutExceptions.exceptions().entrySet()) {
                        final Task task = timeoutExceptions.getKey();
                        task.maybeInitTaskTimeoutOrThrow(time.milliseconds(), timeoutExceptions.getValue());
                        uncommittedTasks.add(task);
                    }
                }
            }

            for (final Task task : tasksToCommit) {
                if (!uncommittedTasks.contains(task)) {
                    ++committed;
                    task.postCommit(false);
                    task.clearTaskTimeout();
                }
            }

            return committed;
```

It struck me as potentially incorrect to
* maybe init the timeout exception for standby tasks
* not clear the timeout for successful tasks when other tasks were timed out

But I'm also not sure whether these are big deals or not.

vvcephei (commented):
Thanks for this, @mjsax ! I just had a couple of comments, but it looks good to me in general!

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
mjsax: I guess I leave it as-is.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsProducer.java
mjsax: Ack. `warn` it is.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
mjsax: Good catch!

vvcephei (approved):
Thanks for the update, @mjsax !

PR #9571, opened 2020-11-06
KAFKA-10691: AlterIsr Respond with wrong Error Id

dengziming:
AlterIsr send by an unknown broker will respond with a STALE_BROKER_EPOCH, which should be UNKNOWN_MEMBER_ID.

### DIFF
--- a/core/src/main/scala/kafka/controller/KafkaController.scala
+++ b/core/src/main/scala/kafka/controller/KafkaController.scala
@@ -2261,7 +2261,7 @@ class KafkaController(val config: KafkaConfig,
     val brokerEpochOpt = controllerContext.liveBrokerIdAndEpochs.get(brokerId)
     if (brokerEpochOpt.isEmpty) {
       info(s"Ignoring AlterIsr due to unknown broker $brokerId")
-      callback.apply(Right(Errors.STALE_BROKER_EPOCH))
+      callback.apply(Right(Errors.UNKNOWN_MEMBER_ID))
       return
     }
 


dengziming:
@hachikuji  hi, PTAL.

core/src/main/scala/kafka/controller/KafkaController.scala
```diff
@@ -2261,7 +2261,7 @@ class KafkaController(val config: KafkaConfig,
     val brokerEpochOpt = controllerContext.liveBrokerIdAndEpochs.get(brokerId)
     if (brokerEpochOpt.isEmpty) {
       info(s"Ignoring AlterIsr due to unknown broker $brokerId")
-      callback.apply(Right(Errors.STALE_BROKER_EPOCH))
+      callback.apply(Right(Errors.UNKNOWN_MEMBER_ID))
```
chia7712, line 2264: not sure whether ```UNKNOWN_MEMBER_ID`` is more suitable than ```STALE_BROKER_EPOCH```. The concept of member id is not equal to the broker id here.

dengziming: Thank you for your reply,  here we have 2 judgement. the first is `if (brokerEpochOpt.isEmpty)` and second is `if(!brokerEpochOpt.contains(brokerEpoch))`, they both return `STALE_BROKER_EPOCH `. maybe the first is meant to return`UNKNOWN_MEMBER_ID `.

dengziming: You are right, I just inspect The KIP-497 and there isn't a `UNKNOWN_MEMBER_ID`.

PR #9572, opened 2020-11-06
KAFKA-10500: Thread Cache Resizes

wcarlson5:
The thread cache can now be resized. This will go towards being able to scale the number of threads

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -160,7 +160,7 @@ public class KafkaStreams implements AutoCloseable {
     private final StreamsMetricsImpl streamsMetrics;
     private final ProcessorTopology taskTopology;
     private final ProcessorTopology globalTaskTopology;
-    private Long totalCacheSize;
+    private final Long totalCacheSize;
 
     GlobalStreamThread globalStreamThread;
     private KafkaStreams.StateListener stateListener;
@@ -900,10 +900,6 @@ public class KafkaStreams implements AutoCloseable {
     }
 
     private long getCacheSizePerThread(final int numStreamThreads) {
-        if (totalCacheSize < 0) {
-            totalCacheSize = 0L;
-            log.warn("Negative cache size passed in. Reverting to cache size of 0 bytes.");
-        }
         return totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -160,7 +160,7 @@ public class KafkaStreams implements AutoCloseable {
     private final StreamsMetricsImpl streamsMetrics;
     private final ProcessorTopology taskTopology;
     private final ProcessorTopology globalTaskTopology;
-    private final Long totalCacheSize;
+    private final long totalCacheSize;
 
     GlobalStreamThread globalStreamThread;
     private KafkaStreams.StateListener stateListener;
@@ -823,7 +823,7 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         totalCacheSize = config.getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
-        final long cacheSizePerThread = getCacheSizePerThread(numStreamThreads);
+        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
         final boolean hasPersistentStores = taskTopology.hasPersistentLocalStore() ||
                 (hasGlobalTopology && globalTaskTopology.hasPersistentGlobalStore());
 
@@ -899,12 +899,8 @@ public class KafkaStreams implements AutoCloseable {
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
-    private long getCacheSizePerThread(final int numStreamThreads) {
-        return totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
-    }
-
     private void resizeThreadCache(final int numStreamThreads) {
-        final long cacheSizePreThread = getCacheSizePerThread(numStreamThreads);
+        final long cacheSizePreThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
         for (final StreamThread streamThread: threads) {
             streamThread.resizeCache(cacheSizePreThread);
         }
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -904,9 +904,6 @@ public class KafkaStreams implements AutoCloseable {
     }
 
     private void resizeThreadCache(final int numStreamThreads) {
-        if (numStreamThreads < 0) {
-            throw new IllegalArgumentException("There cannot be negative StreamThreads");
-        }
         final long cacheSizePreThread = getCacheSizePerThread(numStreamThreads);
         for (final StreamThread streamThread: threads) {
             streamThread.resizeCache(cacheSizePreThread);
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -823,7 +823,7 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         totalCacheSize = config.getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
-        final long cacheSizePerThread = getCacheSizePreThread(numStreamThreads);
+        final long cacheSizePerThread = getCacheSizePerThread(numStreamThreads);
         final boolean hasPersistentStores = taskTopology.hasPersistentLocalStore() ||
                 (hasGlobalTopology && globalTaskTopology.hasPersistentGlobalStore());
 
@@ -899,7 +899,7 @@ public class KafkaStreams implements AutoCloseable {
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
-    private long getCacheSizePreThread(final int numStreamThreads) {
+    private long getCacheSizePerThread(final int numStreamThreads) {
         if (totalCacheSize < 0) {
             totalCacheSize = 0L;
             log.warn("Negative cache size passed in. Reverting to cache size of 0 bytes.");
@@ -908,7 +908,7 @@ public class KafkaStreams implements AutoCloseable {
     }
 
     private void resizeThreadCache(final int numStreamThreads) {
-        final long cacheSizePreThread = getCacheSizePreThread(numStreamThreads);
+        final long cacheSizePreThread = getCacheSizePerThread(numStreamThreads);
         for (final StreamThread streamThread: threads) {
             streamThread.resizeCache(cacheSizePreThread);
         }
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -823,7 +823,7 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         totalCacheSize = config.getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
-        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + (hasGlobalTopology ? 1 : 0));
+        final long cacheSizePerThread = getCacheSizePerThread(numStreamThreads);
         final boolean hasPersistentStores = taskTopology.hasPersistentLocalStore() ||
                 (hasGlobalTopology && globalTaskTopology.hasPersistentGlobalStore());
 
@@ -899,11 +899,15 @@ public class KafkaStreams implements AutoCloseable {
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
+    private long getCacheSizePerThread(final int numStreamThreads) {
+        return totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
+    }
+
     private void resizeThreadCache(final int numStreamThreads) {
         if (numStreamThreads < 0) {
             throw new IllegalArgumentException("There cannot be negative StreamThreads");
         }
-        final long cacheSizePreThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
+        final long cacheSizePreThread = getCacheSizePerThread(numStreamThreads);
         for (final StreamThread streamThread: threads) {
             streamThread.resizeCache(cacheSizePreThread);
         }
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -823,7 +823,7 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         totalCacheSize = config.getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
-        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
+        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + (hasGlobalTopology ? 1 : 0));
         final boolean hasPersistentStores = taskTopology.hasPersistentLocalStore() ||
                 (hasGlobalTopology && globalTaskTopology.hasPersistentGlobalStore());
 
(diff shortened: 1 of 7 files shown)


wcarlson5:
@cadonna Part 2

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -589,6 +589,10 @@ private void subscribeConsumer() {
         }
     }
 
+    public void resizeCache(final long size) {
+        taskManager.resizeCache(size);
```
wcarlson5: I am not sure if I should just expose cache or pass it along. It is about 4 levels deep

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -806,6 +803,21 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
+    private long getCacheSizePerThread(final int numStreamThreads) {
+        if (totalCacheSize < 0) {
+            totalCacheSize = 0L;
+            log.warn("Negative cache size passed in. Reverting to cache size of 0 bytes.");
+        }
```
cadonna: I see that this check was there before, but I actually think it is not needed because the configs are validated and there `CACHE_MAX_BYTES_BUFFERING_CONFIG` is specified as at least 0.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -155,6 +155,7 @@
     private final StreamsMetricsImpl streamsMetrics;
     private final ProcessorTopology taskTopology;
     private final ProcessorTopology globalTaskTopology;
+    private Long totalCacheSize;
```
cadonna: I think this can be a `final long` if we remove the check as I proposed below.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
cadonna: I am in favour of keeping a reference to the thread cache in the `StreamThread` and do the re-sizing here. I think it makes the code a bit easier to follow. 
You will need synchronization, because the thread that will add the new stream thread will also resize the thread caches.

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
```diff
@@ -71,6 +71,16 @@ public long flushes() {
         return numFlushes;
     }
 
+    public void resize(final long maxCacheSizeBytes) {
+        final boolean shrink = maxCacheSizeBytes < this.maxCacheSizeBytes;
+        this.maxCacheSizeBytes = maxCacheSizeBytes;
+        if (shrink) {
+            for (final NamedCache cache : caches.values()) {
+                maybeEvict(cache.name());
+            }
```
cadonna: This loop has the disadvantage that it first evict entries of one named cache, if all entries are evicted and we still need to free space, it starts to evict entries of the next named cache etc. I guess it would be better to avoid such a skewed emission of records to downstream by continuously iterating over the named caches and evict one entry at a time from each named cache until enough space is freed.

cadonna (commented):
Thank you for the PR, @wcarlson5 !

Here my feedback.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: good to know

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: yes it can

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
wcarlson5: I would expect it to take a bit to repopulate the cache to be balanced but you are right it probably better to do so evenly

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: Are the thread caches not independent of each other? Also I was not planning on having the new thread resize the cache but the calling thread do so

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -155,6 +155,7 @@
     private final StreamsMetricsImpl streamsMetrics;
     private final ProcessorTopology taskTopology;
     private final ProcessorTopology globalTaskTopology;
+    private final Long totalCacheSize;
```
cadonna: Why does this need to be a `Long` instead of a `long`? The numerical value of the variable is only immutable if we use a `long` here.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -806,6 +803,17 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
+    private long getCacheSizePerThread(final int numStreamThreads) {
+        return totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
+    }
```
cadonna, line 904: IMO, the code would be easier navigable if you inline this method. Without the removed check, there is not really a reason to have a method here.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
cadonna: > I was not planning on having the new thread resize the cache but the calling thread do so

That is what I am saying "the thread that will add the new stream thread" is the calling thread. The new stream thread cannot resize the caches of the other stream threads because it is not aware of the other stream threads. Still we need synchronization because the calling thread will access and modify the thread caches of all stream threads and all stream threads will access and modify their own thread cache during normal processing.

wcarlson5: I'm not sure I follow. If we make it so that each thread is responsible for resizing their own independent cache then why would the method need to synchronized as each call should not affect the others. And if one thread does all the work there should only be one call. Unless you are thinking about multiple threads adding threads at the same time?

Maybe I don't understand how the cache is set up well enough?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I think that its just what the ide defaulted to. Ill change it.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I think I agree with you. fixed

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: Talked about off line. The named cache is already synchronized. Also made max cache size volatile

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -728,12 +729,8 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
                 "must subscribe to at least one source topic or global table.");
         }
 
-        long totalCacheSize = config.getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
-        if (totalCacheSize < 0) {
-            totalCacheSize = 0;
```
mjsax, line 826: Why do we remove this guard?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -728,12 +729,8 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
-            log.warn("Negative cache size passed in. Reverting to cache size of 0 bytes.");
-        }
-        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + (hasGlobalTopology ? 1 : 0));
+        totalCacheSize = config.getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
+        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
```
mjsax: Why move off using `hasGlobalTopology`?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -806,6 +803,13 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
+    private void resizeThreadCache(final int numStreamThreads) {
+        final long cacheSizePreThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
```
mjsax: Seems this duplicates `L733`. Might be good to extract into a small helper method.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -589,6 +593,10 @@ private void subscribeConsumer() {
         }
     }
 
+    public void resizeCache(final long size) {
+        cacheResizer.accept(size);
```
mjsax, line 624: Why do we need the `cacheResizer`? Can't we just call `cache.resize(size)` here?

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
```diff
@@ -71,6 +72,26 @@ public long flushes() {
         return numFlushes;
     }
 
+    public void resize(final long maxCacheSizeBytes) {
```
mjsax: nit: `newCachSizeBytes` ? (To avoid the "clash" with `this.maxCachSizeBytes`.)

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
```diff
@@ -71,6 +72,26 @@ public long flushes() {
...
+        final boolean shrink = maxCacheSizeBytes < this.maxCacheSizeBytes;
+        this.maxCacheSizeBytes = maxCacheSizeBytes;
+        if (shrink) {
+            final CircularIterator<NamedCache> circularIterator = new CircularIterator<>(caches.values());
+            while (sizeBytes() > maxCacheSizeBytes) {
+                if (!circularIterator.hasNext()) {
+                    log.error("Unable to remove any more entries as all caches are empty");
```
mjsax: Could this ever happen? If we the max cache size is smaller than a single entry, would we not evict the entry and the used cache size would always shrink to zero?

mjsax (commented):
A few nits. Overall LGTM.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: https://github.com/apache/kafka/pull/9572#discussion_r520444420

It seems it was not necessary

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I did have it in a separate method but helper but when removing the `totalCacheSize < 0 ` check @cadonna thought it would be more readable inline

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: It was in a separate method without access to `hasGlobalTopology`. I supposes if it stays we can move it back

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: The cache is not exposed in stream thread. Since I was only using one method I thought it best to only expose that.

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
wcarlson5: sure, that works

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
wcarlson5: If we add a check to make sure the number of threads is positive then probably not. Ill add that check then remove this one

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
mjsax: Not sure why the `totalCacheSize` check is relevant for avoiding code duplication?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
mjsax: Ah. I see. -- Should we pass `java.util.function.Consumer<Long> cacheResizer` into `StreamThread` constructor for this case instead?

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
mjsax: I see. -- I guess the miss-leading fact was, that this check was done inside the while-loop.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I think it was about readability. I might be misremembering though, as it was a conversation we had last week

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: I think we can, thats probably a good idea.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
cadonna: If this line is duplicated, it should go in a method. When I proposed to move it inline, I was apparently not aware that the same line was used somewhere else.

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
wcarlson5: Yeah, in retrospect it was not very clear. Hopefully its better this way now

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Moved to a new method. Glad we got that cleared up. LGTM?

mjsax: > LGTM?

If this is a question, should it be LGTY? 😂

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -806,6 +803,20 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
+    private long getCacheSizePerThread(final int numStreamThreads) {
+        return totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
+    }
+
+    private void resizeThreadCache(final int numStreamThreads) {
+        if (numStreamThreads < 0) {
```
mjsax: Can it be smaller than `0` ? Should the test be `<= 0` or `< 1` instead?

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
```diff
@@ -71,6 +72,22 @@ public long flushes() {
         return numFlushes;
     }
 
+    public void resize(final long newCacheSizeBytes) {
+        final boolean shrink = newCacheSizeBytes < this.maxCacheSizeBytes;
```
mjsax: nit: we can remove `this.` now (same next line)

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: It can be zero if you have a global thread, but since this is internal the check might not be entirely necessary

mjsax: Yes, it can be zero, but the check says `< 0`, so it would always evaluate to false?

And if we have zero threads, we should not resize the cache as we might end up in an infinite loop? But we would only call this method if we "shrink", ie, if the thread count grows, but it can never grow from negative to zero, right?

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
wcarlson5: yep

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: That is a good point. Maybe what we need to do it put a minimum size of cache to limit how many stream threads an instance can have?

mjsax: Well, `getCacheSizePerThread` would eventually return zero (with growing number of threads), what means that every put() into the cache would result in an immediate eviction. So I don't think we need to do anything for this corner case.

wcarlson5: that is a good point

mjsax:
Thanks for the PR @wcarlson5. Merged to `trunk`.

PR #9573, opened 2020-11-06
KAFKA-10693: Close quota managers created in tests

splett2:
We have various tests that call `QuotaFactory.instantiate()`, but don't close the returned `QuotaManagers`. This results in dangling threads from `ClientQuotaManager` and friends in test runs.

After these changes, I ran the unit tests locally and they passed.

### DIFF
--- a/core/src/test/scala/unit/kafka/server/IsrExpirationTest.scala
+++ b/core/src/test/scala/unit/kafka/server/IsrExpirationTest.scala
@@ -72,8 +72,8 @@ class IsrExpirationTest {
 
   @After
   def tearDown(): Unit = {
-    replicaManager.shutdown(false)
-    quotaManager.shutdown()
+    Option(replicaManager).foreach(_.shutdown(false))
+    Option(quotaManager).foreach(_.shutdown())
     metrics.close()
   }
 
--- a/core/src/test/scala/unit/kafka/server/IsrExpirationTest.scala
+++ b/core/src/test/scala/unit/kafka/server/IsrExpirationTest.scala
@@ -22,6 +22,7 @@ import java.util.concurrent.atomic.AtomicBoolean
 
 import kafka.cluster.Partition
 import kafka.log.{Log, LogManager}
+import kafka.server.QuotaFactory.QuotaManagers
 import kafka.utils.TestUtils.MockAlterIsrManager
 import kafka.utils._
 import org.apache.kafka.common.TopicPartition
@@ -51,6 +52,7 @@ class IsrExpirationTest {
   val time = new MockTime
   val metrics = new Metrics
 
+  var quotaManager: QuotaManagers = null
   var replicaManager: ReplicaManager = null
 
   var alterIsrManager: MockAlterIsrManager = _
@@ -62,14 +64,16 @@ class IsrExpirationTest {
     EasyMock.replay(logManager)
 
     alterIsrManager = TestUtils.createAlterIsrManager()
+    quotaManager = QuotaFactory.instantiate(configs.head, metrics, time, "")
     replicaManager = new ReplicaManager(configs.head, metrics, time, null, null, logManager, new AtomicBoolean(false),
-      QuotaFactory.instantiate(configs.head, metrics, time, ""), new BrokerTopicStats, new MetadataCache(configs.head.brokerId),
+      quotaManager, new BrokerTopicStats, new MetadataCache(configs.head.brokerId),
       new LogDirFailureChannel(configs.head.logDirs.size), alterIsrManager)
   }
 
   @After
   def tearDown(): Unit = {
     replicaManager.shutdown(false)
+    quotaManager.shutdown()
     metrics.close()
   }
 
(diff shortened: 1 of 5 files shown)


core/src/test/scala/unit/kafka/server/IsrExpirationTest.scala
```diff
@@ -62,14 +64,16 @@ class IsrExpirationTest {
...
     replicaManager = new ReplicaManager(configs.head, metrics, time, null, null, logManager, new AtomicBoolean(false),
-      QuotaFactory.instantiate(configs.head, metrics, time, ""), new BrokerTopicStats, new MetadataCache(configs.head.brokerId),
+      quotaManager, new BrokerTopicStats, new MetadataCache(configs.head.brokerId),
       new LogDirFailureChannel(configs.head.logDirs.size), alterIsrManager)
   }
 
   @After
   def tearDown(): Unit = {
     replicaManager.shutdown(false)
+    quotaManager.shutdown()
```
dajac: nit: We should check that `quotaManager` is not null. We can also do it for `replicaManager` above.

core/src/test/scala/unit/kafka/server/ReplicaManagerTest.scala
```diff
@@ -1506,7 +1508,7 @@ class ReplicaManagerTest {
       purgatoryName = "ElectLeader", timer, reaperEnabled = false)
 
     // Mock network client to show leader offset of 5
-    val quota = QuotaFactory.instantiate(config, metrics, time, "")
+    val quota = quotaManager
```
dajac: nit: Could we get rid of `quota`?

core/src/test/scala/unit/kafka/server/ReplicaManagerTest.scala
```diff
@@ -79,22 +81,24 @@ class ReplicaManagerTest {
...
+    val props = TestUtils.createBrokerConfig(1, TestUtils.MockZkConnect)
+    config = KafkaConfig.fromProps(props)
     alterIsrManager = EasyMock.createMock(classOf[AlterIsrManager])
+    quotaManager = QuotaFactory.instantiate(config, metrics, time, "")
   }
 
   @After
   def tearDown(): Unit = {
     TestUtils.clearYammerMetrics()
+    quotaManager.shutdown()
```
dajac: Should we check that `quotaManager` is not null?

core/src/test/scala/unit/kafka/server/epoch/OffsetsForLeaderEpochTest.scala
```diff
@@ -112,4 +120,11 @@ class OffsetsForLeaderEpochTest {
     //Then
     assertEquals(new EpochEndOffset(Errors.UNKNOWN_TOPIC_OR_PARTITION, UNDEFINED_EPOCH, UNDEFINED_EPOCH_OFFSET), response(tp))
   }
+
+  @After
+  def tearDown(): Unit = {
+    replicaManager.shutdown(checkpointHW = false)
+    quotaManager.shutdown()
```
dajac: We should check non null in both cases.

dajac (commented):
@splett2 Thanks for the PR. I have left few minor comments.

core/src/test/scala/unit/kafka/server/ReplicaManagerTest.scala
splett2: this is needed to avoid method variable name collision with the `ReplicationQuotaManager` in the overriden `createReplicaFetcherManager` a little bit further down in the test.

core/src/test/scala/unit/kafka/server/ReplicaManagerTest.scala
```diff
@@ -1894,10 +1894,10 @@ class ReplicaManagerTest {
 
     // each replica manager is for a broker
     val rm0 = new ReplicaManager(config0, metrics, time, kafkaZkClient, new MockScheduler(time), mockLogMgr0,
-      new AtomicBoolean(false), QuotaFactory.instantiate(config0, metrics, time, ""),
+      new AtomicBoolean(false), quotaManager,
       brokerTopicStats1, metadataCache0, new LogDirFailureChannel(config0.logDirs.size), alterIsrManager)
     val rm1 = new ReplicaManager(config1, metrics, time, kafkaZkClient, new MockScheduler(time), mockLogMgr1,
-      new AtomicBoolean(false), QuotaFactory.instantiate(config1, metrics, time, ""),
+      new AtomicBoolean(false), quotaManager,
```
splett2, line 1899: the two replica managers are intended to be separate. I used the same quota manager for both to avoid polluting the method signature/diff. The tests using this utility function don't rely on any quota manager behavior, so I felt this was okay.

core/src/test/scala/unit/kafka/server/ReplicaManagerTest.scala
dajac: Could we rename `quotaManager` used in `createReplicaFetcherManager` to `replicationQuotaManager` and use `quotaManager` as the global one everywhere?

dajac (approved):
LGTM, thanks.

PR #9574, opened 2020-11-07
MINOR: KIP-497 Not start Controller2ChannelManager when we are using zookeeper to notify isr change

dengziming:
KafkaServer Start alterIsrChannelManager even when we are using zookeeper to notify isr change.
 
BrokerToControllerChannelManager will consume some resources when started, so we should optimize them as possible.


dengziming:
@abbccdda @hachikuji ,Hi, PTAL.

dengziming:
The #9677 add an `ApiVersion.isAlterIsrSupported` method, so just use this method. 
@mumrah ,Hi, will this optimization take effect? PTAL.

mumrah (approved):
@dengziming, thanks for the patch! I think this change looks good. We don't need the AlterIsrManager propagation thread to run unless we are running IBP greater than or equal to 2.7-IV2. 

LGTM

dengziming:
@mumrah can you merge this pr? thank you.

dengziming:
Closed this since it has been resolved by #9732

PR #9575, opened 2020-11-07
MINOR: optimize lock operation

dengziming:
move lock backward.


clients/src/main/java/org/apache/kafka/clients/producer/internals/BufferPool.java
```diff
@@ -112,12 +112,11 @@ public ByteBuffer allocate(int size, long maxTimeToBlockMs) throws InterruptedEx
                                                + " on memory allocations.");
 
         ByteBuffer buffer = null;
-        this.lock.lock();
```
chia7712, line 115: The read/write of ```closed``` must be in lock so this improvement is not acceptable to ```BufferPool```.

dengziming: good catch!

PR #9578, opened 2020-11-09
MINOR: Log resource pattern of ACL updates at INFO level

rajinisivaram:
At the moment, we have one log entry for ACL updates that says:
```
Processing notification(s) to /kafka-acl-changes
```
For other updates like broker configuration updates, we have an additional entry at INFO level that shows what was updated when the change notification was processed. Since every resource pattern may have 100s or 1000s of access control entries associated with it, we don't want to log the entire contents on ACL update at INFO level. But it would be useful to log the resource pattern. This shows that we processed the notification in AclAuthorizer and gives the resource and version that was refreshed from ZK. This PR adds an additional INFO-level log entry for ACL updates in AclAuthorizer and retains the existing DEBUG level entry with full details.


omkreddy (approved):

rajinisivaram:
@omkreddy Thanks for the review, merging to trunk

PR #9579, opened 2020-11-09
KAFKA-9751: Forward CreateTopicsRequest for FindCoordinator/Metadata when topic creation is needed

abbccdda:
This PR forward the entire FindCoordinator request to the active controller when the internal topic being queried is not ready to be served yet.

### DIFF
--- a/core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
+++ b/core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
@@ -17,8 +17,9 @@
 
 package kafka.server
 
-import java.util.Properties
+import java.util.{Collections, Properties}
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicReference
 
 import kafka.cluster.Broker
 import kafka.controller.KafkaController
@@ -26,6 +27,7 @@ import kafka.coordinator.group.GroupCoordinator
 import kafka.coordinator.transaction.TransactionCoordinator
 import kafka.utils.Logging
 import org.apache.kafka.clients.ClientResponse
+import org.apache.kafka.common.errors.InvalidTopicException
 import org.apache.kafka.common.internals.Topic
 import org.apache.kafka.common.internals.Topic.{GROUP_METADATA_TOPIC_NAME, TRANSACTION_STATE_TOPIC_NAME}
 import org.apache.kafka.common.message.CreateTopicsRequestData
@@ -37,6 +39,7 @@ import org.apache.kafka.common.requests.{ApiError, CreateTopicsRequest}
 import org.apache.kafka.common.utils.Time
 
 import scala.collection.{Map, Seq, Set, mutable}
+import scala.jdk.CollectionConverters._
 
 trait AutoTopicCreationManager {
 
@@ -94,7 +97,7 @@ class DefaultAutoTopicCreationManager(
   txnCoordinator: TransactionCoordinator
 ) extends AutoTopicCreationManager with Logging {
 
-  private val inflightTopics = new ConcurrentHashMap[String, Boolean]()
+  private val inflightTopics = Collections.newSetFromMap(new ConcurrentHashMap[String, java.lang.Boolean]())
 
   override def start(): Unit = {
     channelManager.foreach(_.start())
@@ -109,51 +112,99 @@ class DefaultAutoTopicCreationManager(
     topics: Set[String],
     controllerMutationQuota: ControllerMutationQuota
   ): Seq[MetadataResponseTopic] = {
-    val (topicResponses, creatableTopics) = filterCreatableTopics(topics)
+    val (creatableTopics, uncreatableTopicResponses) = filterCreatableTopics(topics)
 
-    if (creatableTopics.nonEmpty) {
-      if (!controller.isActive && channelManager.isDefined) {
-        val topicsToCreate = new CreateTopicsRequestData.CreatableTopicCollection
-        creatableTopics.foreach(config => topicsToCreate.add(config._2))
-        val createTopicsRequest = new CreateTopicsRequest.Builder(
-          new CreateTopicsRequestData()
-            .setTimeoutMs(config.requestTimeoutMs)
-            .setTopics(topicsToCreate)
-        )
-
-        channelManager.get.sendRequest(createTopicsRequest, new ControllerRequestCompletionHandler {
-          override def onTimeout(): Unit = {
-            debug(s"Auto topic creation timed out for $topics.")
-            clearInflightRequests(creatableTopics)
-          }
-
-          override def onComplete(response: ClientResponse): Unit = {
-            debug(s"Auto topic creation completed for $topics.")
-            clearInflightRequests(creatableTopics)
-          }
-        })
-        info(s"Sent auto-creation request for $topics to the active controller.")
-      } else {
-        try {
-          def creationCallback(topicErrors: Map[String, ApiError]): Unit = {
-            info(s"Auto-creation of topics $topics returned with errors: $topicErrors")
-          }
-
-          // Note that we use timeout = 0 since we do not need to wait for metadata propagation
-          adminManager.createTopics(
-            timeout = 0,
-            validateOnly = false,
-            creatableTopics,
-            Map.empty,
-            controllerMutationQuota,
-            creationCallback
-          )
-        } finally {
-          clearInflightRequests(creatableTopics)
-        }
-      }
+    val creatableTopicResponses = if (creatableTopics.isEmpty) {
+      Seq.empty
+    } else if (!controller.isActive && channelManager.isDefined) {
+      sendCreateTopicRequest(creatableTopics)
+    } else {
+      createTopicsInZk(creatableTopics, controllerMutationQuota)
     }
-    topicResponses
+
+    uncreatableTopicResponses ++ creatableTopicResponses
+  }
+
+  private def createTopicsInZk(
+    creatableTopics: Map[String, CreatableTopic],
+    controllerMutationQuota: ControllerMutationQuota
+  ): Seq[MetadataResponseTopic] = {
+    val topicErrors = new AtomicReference[Map[String, ApiError]]()
+
+    // Note that we use timeout = 0 since we do not need to wait for metadata propagation
+    // and we want to get the response error immediately.
+    adminManager.createTopics(
+      timeout = 0,
+      validateOnly = false,
+      creatableTopics,
+      Map.empty,
+      controllerMutationQuota,
+      topicErrors.set
+    )
+
+    val creatableTopicResponses = Option(topicErrors.get) match {
+      case Some(errors) =>
+        errors.toSeq.map { case (topic, apiError) =>
+          val error = apiError.error match {
+            case Errors.TOPIC_ALREADY_EXISTS | Errors.REQUEST_TIMED_OUT =>
+              // The timeout error is expected because we set timeout=0. This
+              // nevertheless indicates that the topic metadata was created
+              // successfully, so we return LEADER_NOT_AVAILABLE.
+              Errors.LEADER_NOT_AVAILABLE
+            case error => error
+          }
+
+          new MetadataResponseTopic()
+            .setErrorCode(error.code)
+            .setName(topic)
+            .setIsInternal(Topic.isInternal(topic))
+        }
+
+      case None =>
+        creatableTopics.keySet.toSeq.map { topic =>
+          new MetadataResponseTopic()
+            .setErrorCode(Errors.UNKNOWN_TOPIC_OR_PARTITION.code)
+            .setName(topic)
+            .setIsInternal(Topic.isInternal(topic))
+        }
+    }
+
+    creatableTopicResponses
+  }
+
+  private def sendCreateTopicRequest(
+    creatableTopics: Map[String, CreatableTopic]
+  ): Seq[MetadataResponseTopic] = {
+    val topicsToCreate = new CreateTopicsRequestData.CreatableTopicCollection(creatableTopics.size)
+    topicsToCreate.addAll(creatableTopics.values.asJavaCollection)
+
+    val createTopicsRequest = new CreateTopicsRequest.Builder(
+      new CreateTopicsRequestData()
+        .setTimeoutMs(config.requestTimeoutMs)
+        .setTopics(topicsToCreate)
+    )
+
+    channelManager.get.sendRequest(createTopicsRequest, new ControllerRequestCompletionHandler {
+      override def onTimeout(): Unit = {
+        debug(s"Auto topic creation timed out for ${creatableTopics.keys}.")
+        clearInflightRequests(creatableTopics)
+      }
+
+      override def onComplete(response: ClientResponse): Unit = {
+        debug(s"Auto topic creation completed for ${creatableTopics.keys}.")
+        clearInflightRequests(creatableTopics)
+      }
+    })
+
+    val creatableTopicResponses = creatableTopics.keySet.toSeq.map { topic =>
+      new MetadataResponseTopic()
+        .setErrorCode(Errors.UNKNOWN_TOPIC_OR_PARTITION.code)
+        .setName(topic)
+        .setIsInternal(Topic.isInternal(topic))
+    }
+
+    info(s"Sent auto-creation request for ${creatableTopics.keys} to the active controller.")
+    creatableTopicResponses
   }
 
   private def clearInflightRequests(creatableTopics: Map[String, CreatableTopic]): Unit = {
@@ -195,36 +246,54 @@ class DefaultAutoTopicCreationManager(
     topicConfigs
   }
 
+  private def isValidTopicName(topic: String): Boolean = {
+    try {
+      Topic.validate(topic)
+      true
+    } catch {
+      case e: InvalidTopicException =>
+        false
+    }
+  }
+
   private def filterCreatableTopics(
     topics: Set[String]
-  ): (Seq[MetadataResponseTopic], Map[String, CreatableTopic]) = {
+  ): (Map[String, CreatableTopic], Seq[MetadataResponseTopic]) = {
 
     val aliveBrokers = metadataCache.getAliveBrokers
     val creatableTopics = mutable.Map.empty[String, CreatableTopic]
-    val topicResponses = topics.toSeq.map { topic =>
-      val error = if (hasEnoughLiveBrokers(topic, aliveBrokers)) {
-        val alreadyPresent = inflightTopics.putIfAbsent(topic, true)
-        if (!alreadyPresent) {
-          creatableTopics.put(topic, creatableTopic(topic))
-        }
+    val uncreatableTopics = mutable.Buffer.empty[MetadataResponseTopic]
 
-        Errors.UNKNOWN_TOPIC_OR_PARTITION
+    topics.foreach { topic =>
+      // Attempt basic topic validation before sending any requests to the controller.
+      val validationError: Option[Errors] = if (!isValidTopicName(topic)) {
+        Some(Errors.INVALID_TOPIC_EXCEPTION)
+      } else if (!hasEnoughLiveBrokers(topic, aliveBrokers)) {
+        Some(Errors.INVALID_REPLICATION_FACTOR)
+      } else if (!inflightTopics.add(topic)) {
+        Some(Errors.UNKNOWN_TOPIC_OR_PARTITION)
       } else {
-        Errors.INVALID_REPLICATION_FACTOR
+        None
       }
 
-      new MetadataResponseTopic()
-        .setErrorCode(error.code)
-        .setName(topic)
-        .setIsInternal(Topic.isInternal(topic))
+      validationError match {
+        case Some(error) =>
+          uncreatableTopics.addOne(new MetadataResponseTopic()
+            .setErrorCode(error.code)
+            .setName(topic)
+            .setIsInternal(Topic.isInternal(topic))
+          )
+        case None =>
+          creatableTopics.put(topic, creatableTopic(topic))
+      }
     }
 
-    (topicResponses, creatableTopics)
+    (creatableTopics, uncreatableTopics)
   }
 
   private def hasEnoughLiveBrokers(
     topicName: String,
-    aliveBrokers: Seq[Broker],
+    aliveBrokers: Seq[Broker]
   ): Boolean = {
     val (replicationFactor, replicationFactorConfig) = topicName match {
       case GROUP_METADATA_TOPIC_NAME =>
--- a/core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
+++ b/core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
@@ -17,7 +17,6 @@
 
 package kafka.server
 
-import java.util
 import java.util.Properties
 import java.util.concurrent.ConcurrentHashMap
 
@@ -31,13 +30,13 @@ import org.apache.kafka.common.internals.Topic
 import org.apache.kafka.common.internals.Topic.{GROUP_METADATA_TOPIC_NAME, TRANSACTION_STATE_TOPIC_NAME}
 import org.apache.kafka.common.message.CreateTopicsRequestData
 import org.apache.kafka.common.message.CreateTopicsRequestData.{CreatableTopic, CreateableTopicConfig, CreateableTopicConfigCollection}
-import org.apache.kafka.common.message.MetadataResponseData.{MetadataResponsePartition, MetadataResponseTopic}
+import org.apache.kafka.common.message.MetadataResponseData.MetadataResponseTopic
 import org.apache.kafka.common.metrics.Metrics
 import org.apache.kafka.common.protocol.Errors
-import org.apache.kafka.common.requests.CreateTopicsRequest
+import org.apache.kafka.common.requests.{ApiError, CreateTopicsRequest}
 import org.apache.kafka.common.utils.Time
 
-import scala.collection.{Map, Seq, Set}
+import scala.collection.{Map, Seq, Set, mutable}
 
 trait AutoTopicCreationManager {
 
@@ -80,12 +79,12 @@ object AutoTopicCreationManager {
         ))
       else
         None
-    new AutoTopicCreationManagerImpl(config, metadataCache, channelManager, adminManager,
+    new DefaultAutoTopicCreationManager(config, metadataCache, channelManager, adminManager,
       controller, groupCoordinator, txnCoordinator)
   }
 }
 
-class AutoTopicCreationManagerImpl(
+class DefaultAutoTopicCreationManager(
   config: KafkaConfig,
   metadataCache: MetadataCache,
   channelManager: Option[BrokerToControllerChannelManager],
@@ -95,7 +94,7 @@ class AutoTopicCreationManagerImpl(
   txnCoordinator: TransactionCoordinator
 ) extends AutoTopicCreationManager with Logging {
 
-  private val inflightTopics = new ConcurrentHashMap[String, CreatableTopic]()
+  private val inflightTopics = new ConcurrentHashMap[String, Boolean]()
 
   override def start(): Unit = {
     channelManager.foreach(_.start())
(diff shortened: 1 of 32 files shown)


core/src/test/scala/unit/kafka/server/MetadataRequestWithForwardingTest.scala
```diff
@@ -0,0 +1,92 @@
...
+  override def testAutoCreateOfCollidingTopics(): Unit = {
+    super.testAutoCreateOfCollidingTopics()
+  }
+
+  @Test
+  override def testAutoCreateTopicWithInvalidReplicationFactor(): Unit = {
+    super.testAutoCreateTopicWithInvalidReplicationFactor()
+  }
+
+  /* the rest of tests are not enabled */
```
hachikuji: An alternative that we have done elsewhere would be to introduce an `AbstractMetadataRequestTest` which we can pull the common cases up to. A more elegant option might be to figure out how to use `@ParameterizedTest` so that we can provide config overrides. This would be a little difficult at the moment because we initialize brokers in a `@Before` method. Probably means we need to move away from this approach long term. For now, the abstract class seems preferable. Similar for `CreateTopicsRequestWithForwardingTest`.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1370,55 +1345,164 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+        .setTopics(topicCollection)
+    )
+  }
+
+  private def getTopicConfigs(topic: String): CreatableTopic = {
+    topic match {
+      case GROUP_METADATA_TOPIC_NAME =>
+        new CreatableTopic()
+          .setName(topic)
+          .setNumPartitions(config.offsetsTopicPartitions)
```
hachikuji: The controller should have these configurations as well. Perhaps it is better to use -1 for this and replication factor and let the controller fill them in?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1370,55 +1345,164 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-              new FindCoordinatorResponseData()
-                .setErrorCode(error.code)
-                .setErrorMessage(error.message)
-                .setNodeId(node.id)
-                .setHost(node.host)
-                .setPort(node.port)
-                .setThrottleTimeMs(requestThrottleMs))
+      val topicCreationNeeded = topicMetadata.headOption.isEmpty
+      if (topicCreationNeeded) {
+        if (hasEnoughAliveBrokers(internalTopicName)) {
```
hachikuji: In the case of forwarding, maybe we can let the controller decide if there are enough alive brokers.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1110,82 +1098,39 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+       if (isFetchAllMetadata) {
           // A metadata request for all topics should never result in topic auto creation, but a topic may be deleted
           // in between the creation of the topics parameter and topicResponses, so make sure to return None for this case.
           None
-        } else if (allowAutoTopicCreation && config.autoCreateTopicsEnable) {
-          Some(createTopic(topic, config.numPartitions, config.defaultReplicationFactor))
-        } else {
-          Some(metadataResponseTopic(Errors.UNKNOWN_TOPIC_OR_PARTITION, topic, false, util.Collections.emptyList()))
+       } else {
+        Some(metadataResponseTopic(
```
hachikuji: nit: seems misaligned

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1084,24 +1087,9 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-                          numPartitions: Int,
-                          replicationFactor: Int,
-                          properties: util.Properties = new util.Properties()): MetadataResponseTopic = {
-    try {
-      adminZkClient.createTopic(topic, numPartitions, replicationFactor, properties, RackAwareMode.Safe)
-      info("Auto creation of topic %s with %d partitions and replication factor %d is successful"
-        .format(topic, numPartitions, replicationFactor))
-      metadataResponseTopic(Errors.LEADER_NOT_AVAILABLE, topic, isInternal(topic), util.Collections.emptyList())
-    } catch {
-      case _: TopicExistsException => // let it go, possibly another broker created this topic
```
hachikuji, line 1102: We seem to have lost this handling or am I missing something?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1110,82 +1098,39 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-        } else if (allowAutoTopicCreation && config.autoCreateTopicsEnable) {
-          Some(createTopic(topic, config.numPartitions, config.defaultReplicationFactor))
-        } else {
-          Some(metadataResponseTopic(Errors.UNKNOWN_TOPIC_OR_PARTITION, topic, false, util.Collections.emptyList()))
+       } else {
+        Some(metadataResponseTopic(
+          if (!hasEnoughAliveBrokers(topic))
+            Errors.INVALID_REPLICATION_FACTOR
+          else if (allowAutoTopicCreation && config.autoCreateTopicsEnable)
+            Errors.LEADER_NOT_AVAILABLE
```
hachikuji: Hmm.. In the old logic, we would attempt topic creation through zookeeper first. Then, if the topic was created successfully, we would return LEADER_NOT_AVAILABLE to give time for the controller to elect a leader. Now we return LEADER_NOT_AVAILABLE immediately and we send the CreateTopic request asynchronously. We don't know if the CreateTopic request will ultimately succeed or not. Perhaps it would be better to keep returning `UNKNOWN_TOPIC_OR_PARTITION` until we see that the topic exists.

hachikuji (commented):
Did a quick overview and just want to make sure I understand the high-level details. It looks like this patch is handling two cases:

1) automatic topic creation through `Metadata`
2) automatic internal topic creation through `FindCoordinator`

When the broker encounters one of these cases, it sends a `CreateTopic` request to the controller. However, the broker does not wait for the response before returning the client response. In the case of `FindCoordinator`, we return `COORDINATOR_NOT_AVAILABLE`; for `Metadata`, we return `LEADER_NOT_AVAILABLE`.

Do I have that right?

One question that occurred to me is whether we need logic to avoid spamming the controller with CreateTopic requests while we are waiting for the initial request to go through? For example, after seeing `COORDINATOR_NOT_AVAILABLE`, the consumer will backoff a few ms and then retry. Do we need somewhere to track the topics that are already awaiting creation?

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: I agree it's equivalent, but I think we could be conservative here to keep the logic on broker side for now, to reduce logical change in this PR.

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: I guess the purpose of doing live broker check here is to avoid sending excessive create topic requests.

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: We intentionally avoid using adminZkClient so that we could go through topic creation rules through `zkAdminManager`. TopicExistsException is handled there.

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: That makes sense

core/src/main/scala/kafka/server/BrokerToControllerChannelManager.scala
```diff
@@ -61,7 +61,6 @@ class BrokerToControllerChannelManager(
 
   def shutdown(): Unit = {
     requestThread.shutdown()
-    requestThread.awaitShutdown()
```
abbccdda, line 136: Side cleanup which is called within `requestThread.shutdown()`

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -0,0 +1,139 @@
...
+import scala.collection.Map
+
+trait AutoTopicCreationManager {
+
+  def createTopics(
+    topicNames: Set[CreatableTopic],
+    controllerMutationQuota: ControllerMutationQuota
+  ): Unit
+
+  def start(): Unit = {}
```
hachikuji: nit: do we need default implementations?

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -0,0 +1,139 @@
...
+    metrics: Metrics,
+    threadNamePrefix: Option[String],
+    adminManager: ZkAdminManager,
+    controller: KafkaController,
+    enableForwarding: Boolean
+  ): AutoTopicCreationManager = {
+
+    val channelManager =
+      if (enableForwarding)
+        Some(new BrokerToControllerChannelManager(
```
hachikuji: Let's leave this for a follow-up, but just want to mention that it is probably better if we can reuse the same `BrokerToControllerChannelManager` as `ForwardingManager`. Can you file a JIRA for a follow-up?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1113,82 +1097,36 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-        if (isInternal(topic)) {
-          val topicMetadata = createInternalTopic(topic)
-          Some(
-            if (topicMetadata.errorCode == Errors.COORDINATOR_NOT_AVAILABLE.code)
-              metadataResponseTopic(Errors.INVALID_REPLICATION_FACTOR, topic, true, util.Collections.emptyList())
-            else
-              topicMetadata
-          )
-        } else if (isFetchAllMetadata) {
+       if (isFetchAllMetadata) {
```
hachikuji: Hmm, I agree the ordering of these checks was weird. So if `isFetchAllMetadata` is set, then `responsesForNonExistentTopics` will be empty and we will return `(topicResponses, Seq.empty[MetadataResponseTopic])`. Does that mean we can add this check to the first clause?
```scala 
  if (isFetchAllMetadata || topics.isEmpty || topicResponses.size == topics.size) {
       (topicResponses, Seq.empty[MetadataResponseTopic])
```

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1234,19 +1171,28 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-          request.context.listenerName,
-          errorUnavailableEndpoints,
-          errorUnavailableListeners
-        )
+        (Seq.empty[MetadataResponseTopic], Seq.empty[MetadataResponseTopic])
+      else
+        getTopicMetadata(metadataRequest.isAllTopics, authorizedTopics,
+          request.context.listenerName, errorUnavailableEndpoints, errorUnavailableListeners)
+
+    nonExistTopicMetadata.foreach(metadata =>
```
hachikuji: nit: usually we write as
```scala
nonExistTopicMetadata.foreach { metadata =>
```

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: I don't think it is equivalent, at least not completely. My thought was to reduce the reliance on the broker's configuration, which is more likely to be stale than the controller. This actually raises an interesting question about the `CreateTopic` API which I had not thought of before. If we receive a `CreateTopic` request for an internal topic, which configuration should we use? Currently it looks like we will apply the standard topic defaults, but that does not seem right. I filed https://issues.apache.org/jira/browse/KAFKA-12280, so we can consider this later.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1375,55 +1325,137 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+  private def getTopicConfigs(topic: String): CreatableTopic = {
+    topic match {
+      case GROUP_METADATA_TOPIC_NAME =>
+        new CreatableTopic()
+          .setName(topic)
+          .setNumPartitions(config.offsetsTopicPartitions)
+          .setReplicationFactor(config.offsetsTopicReplicationFactor)
+          .setConfigs(convertToTopicConfigCollections(groupCoordinator.offsetsTopicConfigs))
+      case TRANSACTION_STATE_TOPIC_NAME =>
+        new CreatableTopic()
```
hachikuji: Feels to me like we would benefit by moving this logic into `AutoTopicCreationManager`. The configuration for auto-created topics can always be derived from the broker configuration. Hence we could simplify the interface by letting it take only the topic names. The advantage is that we can move all of this logic out of `KafkaApis` (which is now up to 3500 LOC).

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1375,55 +1325,137 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+  private def hasEnoughAliveBrokers(topic: String): Boolean = {
+    if (topic == null)
+      throw new IllegalArgumentException("topic must not be null")
+
+    val aliveBrokers = metadataCache.getAliveBrokers
+
+    topic match {
+      case GROUP_METADATA_TOPIC_NAME =>
+        if (aliveBrokers.size < config.offsetsTopicReplicationFactor) {
+          error(s"Number of alive brokers '${aliveBrokers.size}' does not meet the required replication factor " +
```
hachikuji: The differences between these log lines are minor. Can we factor out a helper?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1375,55 +1325,137 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-          throw new InvalidRequestException("Unknown coordinator type in FindCoordinator request")
+      val topicMetadata = metadataCache.getTopicMetadata(Set(internalTopicName), request.context.listenerName)
+      def createFindCoordinatorResponse(error: Errors,
+                                        node: Node,
+                                        requestThrottleMs: Int,
+                                        errorMessage: Option[String] = None): FindCoordinatorResponse = {
+        new FindCoordinatorResponse(
+          new FindCoordinatorResponseData()
+            .setErrorCode(error.code)
+            .setErrorMessage(errorMessage.getOrElse(error.message))
```
hachikuji: nit: is `error.message` actually useful to send back? It doesn't provide any information beyond the error code. Could we just use `errorMessage.orNull`?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1375,55 +1325,137 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+          (txnCoordinator.partitionFor(findCoordinatorRequest.data.key), TRANSACTION_STATE_TOPIC_NAME)
+      }
 
-        case _ =>
-          throw new InvalidRequestException("Unknown coordinator type in FindCoordinator request")
+      val topicMetadata = metadataCache.getTopicMetadata(Set(internalTopicName), request.context.listenerName)
+      def createFindCoordinatorResponse(error: Errors,
+                                        node: Node,
+                                        requestThrottleMs: Int,
+                                        errorMessage: Option[String] = None): FindCoordinatorResponse = {
```
hachikuji: Maybe I'm missing it, but where is this argument used?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1375,55 +1325,137 @@ class KafkaApis(val requestChannel: RequestChannel,
         !authHelper.authorize(request.context, DESCRIBE, TRANSACTIONAL_ID, findCoordinatorRequest.data.key))
       requestHelper.sendErrorResponseMaybeThrottle(request, Errors.TRANSACTIONAL_ID_AUTHORIZATION_FAILED.exception)
     else {
-      // get metadata (and create the topic if necessary)
-      val (partition, topicMetadata) = CoordinatorType.forId(findCoordinatorRequest.data.keyType) match {
+      val (partition, internalTopicName) = CoordinatorType.forId(findCoordinatorRequest.data.keyType) match {
```
hachikuji, line 1322: Pre-existing issue. `CoordinatorType.forId` returns `IllegalArgumentException` if the key type is unknown. That will get translated to UNKNOWN_SERVER_ERROR. It would be better to return INVALID_REQUEST.

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: Inside `ZkAdminManager.createTopics`, I see that we catch `TopicExistsException`. However, I do not see any logic to translate it to `LEADER_NOT_AVAILABLE`. Can you show me where this happens?

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: Ok. My thought is to reduce the reliance on the broker metadata. We will need the replication factor check on the controller anyway, so I'm not sure it is worth optimizing for the case when the cluster is initializing.

Perhaps at least we can move this validation into `AutoTopicCreationManager` to better encapsulate the logic.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1234,19 +1171,28 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-        )
+        (Seq.empty[MetadataResponseTopic], Seq.empty[MetadataResponseTopic])
+      else
+        getTopicMetadata(metadataRequest.isAllTopics, authorizedTopics,
+          request.context.listenerName, errorUnavailableEndpoints, errorUnavailableListeners)
+
+    nonExistTopicMetadata.foreach(metadata =>
+      try {
+        // Validate topic name and propagate error if failed
+        Topic.validate(metadata.name())
```
hachikuji: Is the idea to validate before sending the CreateTopic? Could we move this to `AutoTopicManager`?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1234,19 +1171,28 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+      } catch {
+        case e: Exception =>
+          metadata.setErrorCode(Errors.forException(e).code)
       }
+    )
+
+    if (nonExistTopicMetadata.nonEmpty && metadataRequest.allowAutoTopicCreation && config.autoCreateTopicsEnable) {
+      val controllerMutationQuota = quotas.controllerMutation.newQuotaFor(request, strictSinceVersion = 6)
+      autoTopicCreationManager.createTopics(
+        nonExistTopicMetadata.map(metadata => getTopicConfigs(metadata.name())).toSet, controllerMutationQuota)
```
hachikuji: Not sure if I am missing something, but doesn't `nonExistTopicMetadata` include topics which failed the `Topic.validate` check above?

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  override def shutdown(): Unit = {
+    channelManager.foreach(_.shutdown())
+  }
+
+  override def createTopics(topics: Set[CreatableTopic],
+                            controllerMutationQuota: ControllerMutationQuota): Unit = {
+    val topicConfigs = topics
+      .filter(topic => !inflightTopics.contains(topic.name()))
+      .map(topic => {(topic.name(), topic)}).toMap
```
hachikuji: nit: unnecessary braces

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  override def shutdown(): Unit = {
+    channelManager.foreach(_.shutdown())
+  }
+
+  override def createTopics(topics: Set[CreatableTopic],
+                            controllerMutationQuota: ControllerMutationQuota): Unit = {
+    val topicConfigs = topics
+      .filter(topic => !inflightTopics.contains(topic.name()))
```
hachikuji: nit: there are a few of these throughout, but the parenthesis are unnecessary

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -0,0 +1,139 @@
...
+        adminManager.createTopics(
+          requestTimeout,
+          validateOnly = false,
+          topicConfigs,
+          Map.empty,
+          controllerMutationQuota,
+          _ => ())
+      }
+    } else {
+      debug(s"Topics $topics are under creation already, skip sending additional " +
```
hachikuji: We only get this message if _all_ of the topics are already inflight. Perhaps it is still useful if only some of them are inflight?

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -0,0 +1,139 @@
...
+  override def start(): Unit = {
+    channelManager.foreach(_.start())
+  }
+
+  override def shutdown(): Unit = {
+    channelManager.foreach(_.shutdown())
+  }
+
+  override def createTopics(topics: Set[CreatableTopic],
+                            controllerMutationQuota: ControllerMutationQuota): Unit = {
```
hachikuji: Eventually we need to figure out quota behavior for forwarded requests. I am wondering if it makes sense to apply the quota on each broker separately before sending the `CreateTopic` to the controller or if we rely on the controller exclusively.

cc @dajac

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -0,0 +1,139 @@
...
+}
+
+class AutoTopicCreationManagerImpl(
+  channelManager: Option[BrokerToControllerChannelManager],
+  adminManager: ZkAdminManager,
+  controller: KafkaController,
+  requestTimeout: Int
+) extends AutoTopicCreationManager with Logging {
+
+  private val inflightTopics = new ConcurrentHashMap[String, CreatableTopic]
```
hachikuji: Could this be a set? As far as I can tell, we do not rely on the value.

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -0,0 +1,139 @@
...
+        None
+    new AutoTopicCreationManagerImpl(channelManager, adminManager, controller, config.requestTimeoutMs)
+  }
+}
+
+class AutoTopicCreationManagerImpl(
+  channelManager: Option[BrokerToControllerChannelManager],
+  adminManager: ZkAdminManager,
+  controller: KafkaController,
+  requestTimeout: Int
```
hachikuji: nit: requestTimeoutMs?

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -0,0 +1,139 @@
...
+
+        val topicsToCreate = new CreateTopicsRequestData.CreatableTopicCollection
+        topicConfigs.foreach(config => topicsToCreate.add(config._2))
+        val createTopicsRequest = new CreateTopicsRequest.Builder(
+          new CreateTopicsRequestData()
+            .setTimeoutMs(requestTimeout)
+            .setTopics(topicsToCreate)
+        )
+
+        channelManager.get.sendRequest(createTopicsRequest, new ControllerRequestCompletionHandler {
```
hachikuji: Probably useful to have some logging when we send the CreateTopic request and in the callbacks.

hachikuji:
@abbccdda I had one additional thought here. When we receive a normal `CreateTopic` request, we forward it to the controller through the `Envelope` request. This allows us to preserve the original client principal, which is useful for auditing. We are losing that here for auto-created topics, which is unfortunate. One idea I was thinking about is whether we can wrap the `CreateTopics` request for the auto-created topic in an `Envelope` in order to preserve the client principal. So basically we take the implicit topic creation request from the client and turn it into an explicit forwarded request on behalf of the client.

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
abbccdda: Seems I could omit the value here: https://stackoverflow.com/questions/40993683/scala-thread-safe-hashset

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: Makes sense

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: Seems no longer in use, will remove and revert back to using error code message.

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
abbccdda: Sure, we do have https://issues.apache.org/jira/browse/KAFKA-10348 to track.

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
abbccdda: Actually I don't think this logging is very useful, will replace it with something more explicit about state change.

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: Yes, I think so.

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: Actually after looking into the zk admin manager logic, I don't think it's necessary to do the topic validation here.

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
abbccdda: You mean omit () for `topic.name()`?

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: I guess we could rely on admin manager to do the validation for us.

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: The problem we have is that `ZkAdminManager.createTopics` only takes a callback instead of responding to you in realtime whether we hit TopicExists. Right now we are doing the topic creation async, so unless this is necessary to be fixed (which today we would just return UNKNOWN_PARTITION which seems to be semantically similar to LEADER_NOT_AVAILABLE), I think we could just returning unknown partition immediately without waiting for the async creation?

core/src/main/scala/kafka/server/KafkaApis.scala
abbccdda: Sounds good

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
dajac: @hachikuji Sorry for my late reply. I've missed the notification. We have to enforce the quota on the controller exclusively. It is a global quota and we can't really distribute it fairly in the cluster. In this case, it would be great if we could propagate the principal and clientId to the controller to enforce the quota. However, I wonder how we could propagate the error and the delay to the client if the topic creation is throttled. Perhaps, we could reply with `UNKNOW_TOPIC_OR_PARTITION` until the topic can be created.

hachikuji:
@abbccdda Thanks for the updates. I opened a PR with a few fixes to speed this along since we're trying to get it checked in today: https://github.com/abbccdda/kafka/pull/6. The tests that were previously failing now seem to be passing (at least when testing locally).

hachikuji (approved):
LGTM

core/src/main/scala/kafka/server/AutoTopicCreationManager.scala
```diff
@@ -169,6 +169,8 @@ class DefaultAutoTopicCreationManager(
         }
     }
 
+    clearInflightRequests(creatableTopics)
```
hachikuji: Can you use a `try/finally` here?

ijuma:
I triggered a new build. Is there a reason why you aborted the previous one @abbccdda?

abbccdda:
only one no-related test failure in connect. Verified on local, will merge the PR.

PR #9580, opened 2020-11-09
KAFKA-10350: add forwarding manager implementation with metrics

abbccdda:
Add metric for forwarding request tracking. Note that the implementation is slightly diverged from the KIP, where we decide to get rid of the client.id tag since most admin clients would only have one inflight forwarding request.


core/src/main/scala/kafka/server/ForwardingManager.scala
```diff
@@ -0,0 +1,85 @@
...
+
+class ForwardingManager(metadataCache: kafka.server.MetadataCache,
+                        time: Time,
+                        metrics: Metrics,
+                        config: KafkaConfig,
+                        threadNamePrefix: Option[String] = None) extends
+  BrokerToControllerChannelManagerImpl(metadataCache, time, metrics,
+    config, "forwardingChannel", threadNamePrefix) with KafkaMetricsGroup {
+
+  private val forwardingMetricName = "NumRequestsForwardingToControllerPerSec"
```
mumrah, line 38: Wdyt about also adding a "forwarded" boolean to the request log? (in `RequestChannel#updateRequestMetrics`)

Would this be useful? cc @cmccabe

core/src/main/scala/kafka/server/ForwardingManager.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  def forwardRequest(request: RequestChannel.Request,
+                     responseCallback: AbstractResponse => Unit): Unit = {
+    val principalSerde = request.context.principalSerde.asScala.getOrElse(
```
mumrah, line 42: nit: should we do this get-or-else-throw thing, or do a match/case?

core/src/main/scala/kafka/server/ForwardingManager.scala
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.kafka.common.utils.Time
+
+import scala.compat.java8.OptionConverters._
+
+class ForwardingManager(metadataCache: kafka.server.MetadataCache,
+                        time: Time,
+                        metrics: Metrics,
+                        config: KafkaConfig,
+                        threadNamePrefix: Option[String] = None) extends
+  BrokerToControllerChannelManagerImpl(metadataCache, time, metrics,
```
mumrah, line 35: Should we extend the BrokerToControllerChannel here? It looks like we only ever call `forwardRequest`, so maybe it's better for ForwardingManager _not_ to extend BrokerToControllerChannel and instead take it as a dependency. Otherwise we are exposing the `sendRequest` method which we don't really want anyone to use on the `forwardingManager` instance.

core/src/main/scala/kafka/server/ForwardingManager.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  override def start(): Unit = {
+    super.start()
+    newGauge(forwardingMetricName, () => requestQueue.size())
+  }
+
+  override def shutdown(): Unit = {
+    super.shutdown()
+    removeMetric(forwardingMetricName)
```
mumrah, line 82: nit: I normally reverse the order of stuff in shutdown methods so when the super is called anything the child class has done is undone. Probably doesn't matter in this case

mumrah (commented):
Thanks for the patch, @abbccdda. Few questions inline

core/src/main/scala/kafka/server/ForwardingManager.scala
abbccdda: We should have the principal serde emptiness check in the caller, so here would be an illegal state.

core/src/main/scala/kafka/server/ForwardingManager.scala
abbccdda: Actually we do need `sendRequest` on the forwarding manager interface in the topic creation routing (WIP [here.](https://github.com/apache/kafka/pull/9579/). The `forwardRequest` is serving the purpose of embedding client request inside an Envelope, while for certain cases broker would like to send direct CreateTopicsRequest on behalf of its own principal instead of the client principal. I think we could debate whether to rename `forwardingManager` to a different name to also include the possibility of sending direct request, just not have yet come up a better name here.

core/src/main/scala/kafka/server/ForwardingManager.scala
abbccdda: Sg.

core/src/main/scala/kafka/server/ForwardingManager.scala
abbccdda: Added the logging in RequestChannel.

core/src/main/scala/kafka/server/ForwardingManager.scala
mumrah: I think logic is fine, I'm just wondering if it's more idiomatic to do:

```scala
request.context.principalSerde.asScala match {
      case Some(principalSerde) => ...
      case None => throw new IllegalArgumentException
```

core/src/main/scala/kafka/server/ForwardingManager.scala
mumrah: Ah, ok 👍 Thanks for the explanation. Another option would be to let KafkaApis have BrokerToControllerChannelManager and ForwardingManager as separate dependencies. However, we already have lots of dependencies there, so I could see us wanting to use one instance for forwarding and regular requests. I don't really have a strong opinion either way, I'm ok leaving it as-is

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -279,6 +279,7 @@ object RequestChannel extends Logging {
           .append(",principal:").append(session.principal)
           .append(",listener:").append(context.listenerName.value)
           .append(",clientInformation:").append(context.clientInformation)
+          .append(",isForwarded:").append(envelope.isDefined)
```
mumrah: Not 100% sure about our metrics conventions, but I think this should be `forwarded`

core/src/main/scala/kafka/server/ForwardingManager.scala
mumrah: Looks like we actually use this getOrElse + throw pattern in a few places already.

mumrah (approved):
LGTM!

PR #9583, opened 2020-11-10
[KAFKA-10705]: Make state stores not readable by others

lct45:
Change permissions on the folders for the state store so they're no readable or writable by "others", but still accessible by owner and group members.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
@@ -31,10 +31,15 @@ import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
 import java.nio.channels.OverlappingFileLockException;
 import java.nio.file.NoSuchFileException;
+import java.nio.file.Paths;
 import java.nio.file.Path;
+import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
+import java.nio.file.attribute.PosixFilePermission;
+import java.nio.file.attribute.PosixFilePermissions;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import static org.apache.kafka.streams.processor.internals.StateManagerUtil.CHECKPOINT_FILE_NAME;
@@ -101,6 +106,15 @@ public class StateDirectory {
         if (hasPersistentStores && stateDirName.startsWith("/tmp")) {
             log.warn("Using /tmp directory in the state.dir property can cause failures with writing the checkpoint file" +
                 " due to the fact that this directory can be cleared by the OS");
+        }
+        final Path basePath = Paths.get(baseDir.getPath());
+        final Path statePath = Paths.get(stateDir.getPath());
+        final Set<PosixFilePermission> perms = PosixFilePermissions.fromString("rwxr-x---");
+        try {
+            Files.setPosixFilePermissions(basePath, perms);
+            Files.setPosixFilePermissions(statePath, perms);
+        } catch (final IOException e) {
+            log.error("Error changing permissions for the state or base directory {} ", stateDir.getPath(), e);
         }
     }
 
--- a/streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
@@ -34,10 +34,14 @@ import java.io.IOException;
 import java.nio.channels.FileChannel;
 import java.nio.channels.OverlappingFileLockException;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
+import java.nio.file.attribute.PosixFilePermission;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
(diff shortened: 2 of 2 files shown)


lct45:
@wcarlson5 @cadonna @ableegoldman for review

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -30,11 +30,16 @@
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
 import java.nio.channels.OverlappingFileLockException;
-import java.nio.file.NoSuchFileException;
```
wcarlson5, line 33: Did we need to move this?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -102,6 +107,13 @@ public StateDirectory(final StreamsConfig config, final Time time, final boolean
             log.warn("Using /tmp directory in the state.dir property can cause failures with writing the checkpoint file" +
                 " due to the fact that this directory can be cleared by the OS");
         }
+        final Path path = Paths.get(baseDir.getPath());
```
wcarlson5: Is changing the `baseDir` sufficient?

wcarlson5 (commented):
Just a couple small questions

Thanks for the PR @lct45 !

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
lct45: We did not

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
lct45: It should be, the `stateDir` is inside of the `baseDir` so the permissions should apply there as well. I guess if we wanted to be _more_ selective we could apply permissions to `stateDir` and not `baseDir` but from what I see it doesn't make a big difference. In the test though, there is an `appDir` that had broader permissions even with the changes to `baseDir` so maybe we _do_ need to change permissions to both `baseDir` and `stateDir`

wcarlson5: I'm not sure off the top of my head I just wanted to verify, but its probably best to be overly specific. LGTM

streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
```diff
@@ -106,6 +110,27 @@ public void shouldCreateBaseDirectory() {
...
+            PosixFilePermission.GROUP_EXECUTE,
+            PosixFilePermission.OWNER_READ);
+
+        final Path statePath = Paths.get(stateDir.getPath());
+        final Path basePath = Paths.get(appDir.getPath());
+        try {
+            final Set<PosixFilePermission> baseFilePermissions = Files.getPosixFilePermissions(statePath);
+            final Set<PosixFilePermission> appFilePermissions = Files.getPosixFilePermissions(basePath);
+            assertEquals(expectedPermissions, baseFilePermissions);
+            assertEquals(expectedPermissions, appFilePermissions);
```
cadonna: Recently, we prefer to use `assertThat()` instead of `assertEquals()`.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
```diff
@@ -106,6 +110,27 @@ public void shouldCreateBaseDirectory() {
...
+        } catch (final IOException e) {
+            // okay
```
cadonna: If we swallow the exception here, and the test always throws an IO exception, we will never notice. I guess it would be better to use `fail()` with a message.

cadonna (commented):
Thank you for PR, @lct45 !

I think, you should create a AK ticket and change the PR title accordingly. It would be good if people are made aware of this change. 

Here my feedback:

wcarlson5:
retest this please

cadonna (approved):
LGTM!

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -102,6 +107,15 @@ public StateDirectory(final StreamsConfig config, final Time time, final boolean
             log.warn("Using /tmp directory in the state.dir property can cause failures with writing the checkpoint file" +
                 " due to the fact that this directory can be cleared by the OS");
         }
+        final Path basePath = Paths.get(baseDir.getPath());
+        final Path statePath = Paths.get(stateDir.getPath());
+        final Set<PosixFilePermission> perms = PosixFilePermissions.fromString("rwxr-x---");
```
ableegoldman, line 112: Just wondering, why "read" and "execute" permissions for the group?

cadonna: I thought it would be handy to be able to log into a node to debug state store issues, e.g., with some RocksDB command line tools, with a user that is able to read not able to write, i.e., is in the group but it is not the owner.

lct45: I agree with what Bruno said for the reading. I think allowing execute for group members isn't a security risk, since it's still somewhat contained. I ran the streams test without having execute for group and I got an exception in `StreamTableJoinTopologyOptimizationIntegrationTest` test `shouldDoStreamTableJoinWithDifferentNumberOfPartitions`. It looks like the integration test utils weren't able to clear the directory because groups didn't have execute authorization. Since it doesn't seem like allowing groups to execute would constitute a security risk, it seems like this is a good thing to keep for testing capabilities

ableegoldman: >  I ran the streams test without having execute for group and I got an exception in StreamTableJoinTopologyOptimizationIntegrationTest test

I've noticed that test is actually pretty flaky at the moment, IIRC it's been due to a `Directory not empty` exception). So the failure you saw might not be related to this PR. I'm pretty sure it's write permissions, not execute, that let you delete files within a directory. Execute just lets you `cd` in and traverse the directory. So I guess read+execute does make sense here, and probably we wouldn't want anyone other than the owner to be able to clear the directory anyway

lct45: Hmm yeah that was the exception so it probably wasn't related to this PR. I would err on the side of leaving those permissions but can take them out if you think that's better

ableegoldman: No I think this looks ok, just wanted to ask

ableegoldman (approved):
LGTM, merging to trunk

ableegoldman:
Cherrypicked to 2.6, will wait for 2.7.0 release to finish up before merging to 2.7

mimaison:
@ableegoldman I see you merged this in 2.6. How critical is this fix? Do I need to spin a new RC?

ableegoldman:
I don't think so. It would be nice to have if you happen to end up cutting a new RC, but I wouldn't delay the ongoing release over this

ableegoldman:
Cherrypicked to 2.7

PR #9584, opened 2020-11-10
[KAFKA-10708]: Add "group-id" Tag to Kafka Consumer Metrics

brianwyka:
Add the "group-id" to the metrics created during instantiation of the `KafkaConsumer`

https://issues.apache.org/jira/browse/KAFKA-10708

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
@@ -872,8 +872,13 @@ public class KafkaConsumer<K, V> implements Consumer<K, V> {
                 .timeWindow(config.getLong(ConsumerConfig.METRICS_SAMPLE_WINDOW_MS_CONFIG), TimeUnit.MILLISECONDS)
                 .recordLevel(Sensor.RecordingLevel.forName(config.getString(ConsumerConfig.METRICS_RECORDING_LEVEL_CONFIG)))
                 .tags(metricsTags);
+        final Map<String, Object> reporterConfigOverrides = new HashMap<>();
+        reporterConfigOverrides.put(ConsumerConfig.CLIENT_ID_CONFIG, clientId);
+        if (groupId != null && !groupId.isEmpty()) {
+            reporterConfigOverrides.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
+        }
         List<MetricsReporter> reporters = config.getConfiguredInstances(ConsumerConfig.METRIC_REPORTER_CLASSES_CONFIG,
-                MetricsReporter.class, Collections.singletonMap(ConsumerConfig.CLIENT_ID_CONFIG, clientId));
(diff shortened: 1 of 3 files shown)


brianwyka:
I couldn't find any tests that are currently testing out the `client-id` tag for reference.  Would appreciate a point in the right direction.

clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
```diff
@@ -858,14 +862,23 @@ public KafkaConsumer(Map<String, Object> configs,
         this.kafkaConsumerMetrics = new KafkaConsumerMetrics(metrics, "consumer");
     }
 
-    private static Metrics buildMetrics(ConsumerConfig config, Time time, String clientId) {
-        Map<String, String> metricsTags = Collections.singletonMap(CLIENT_ID_METRIC_TAG, clientId);
+    private static Metrics buildMetrics(ConsumerConfig config, Time time, String clientId, String groupId) {
+        Map<String, String> metricsTags = new HashMap<>();
+        metricsTags.put(CLIENT_ID_METRIC_TAG, clientId);
+        if (groupId != null && !groupId.trim().isEmpty()) {
```
kirktrue: The secondary `!groupId.trim().isEmpty()` clause of the `if` statement shouldn't ever catch given that that case is  checked for in the constructor before calling `buildMetrics`, right?

clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
```diff
@@ -858,14 +862,23 @@ public KafkaConsumer(Map<String, Object> configs,
...
+            metricsTags.put(GROUP_ID_METRIC_TAG, groupId);
+        }
         MetricConfig metricConfig = new MetricConfig().samples(config.getInt(ConsumerConfig.METRICS_NUM_SAMPLES_CONFIG))
                 .timeWindow(config.getLong(ConsumerConfig.METRICS_SAMPLE_WINDOW_MS_CONFIG), TimeUnit.MILLISECONDS)
                 .recordLevel(Sensor.RecordingLevel.forName(config.getString(ConsumerConfig.METRICS_RECORDING_LEVEL_CONFIG)))
                 .tags(metricsTags);
+        final Map<String, Object> reporterConfigOverrides = new HashMap<>();
+        reporterConfigOverrides.put(ConsumerConfig.CLIENT_ID_CONFIG, clientId);
+        if (groupId != null && !groupId.trim().isEmpty()) {
```
kirktrue: Same minor nitpick about whether or not we need to check for an empty group ID.

kirktrue (approved):
This looks pretty clear-cut to me.

Is there anywhere in the existing unit tests that verify that the client ID or group ID tags are actually coming through correctly?

brianwyka:
I'm not sure @kirktrue.

PR #9589, opened 2020-11-12
KAFKA-10710 - Mirror Maker 2 - Create herders only if source->target.enabled=true

jchanaud:
By default Mirror Maker 2 creates herders for all the possible combinations even if the "links" are not enabled.

This is because the beats are emitted from the "opposite" herder.
If there is a replication flow from A to B and heartbeats are required, 2 herders are needed :

- A->B for the MirrorSourceConnector
- B->A for the MirrorHeartbeatConnector

The MirrorHeartbeatConnector on B->A emits beats into topic heartbeats on cluster A.
The MirrorSourceConnector on A->B then replicates whichever topic is configured as well as heartbeats.

In cases with multiple clusters (10 and more), this leads to an incredible amount of connections, file descriptors and configuration topics created in every target clusters that are not necessary.

With this code change, we add a new top level property "emit.heartbeats.enabled" which defaults to "true".
We skip creating the A->B herder whenever A->B.emit.heartbeats.enabled=false (defaults to true) and A->B.enabled=false (defaults to false). 

Existing users will not see any change and if they depend on these "opposites" herders for their monitoring, it will still work.
New users with more complex use case can change this property and fine tune their heartbeat generation.

### DIFF
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
@@ -89,12 +89,12 @@ public class MirrorMakerConfig extends AbstractConfig {
     public List<SourceAndTarget> clusterPairs() {
         List<SourceAndTarget> pairs = new ArrayList<>();
         Set<String> clusters = clusters();
+        Map<String, String> originalStrings = originalsStrings();
         for (String source : clusters) {
             for (String target : clusters) {
-                SourceAndTarget sourceAndTarget = new SourceAndTarget(source, target);
                 if (!source.equals(target)) {
-                    if (this.stringsWithPrefixStripped(source + "->" + target + ".").getOrDefault("enabled","false").equals("true")){
-                        pairs.add(sourceAndTarget);
+                    if (originalStrings.getOrDefault(source + "->" + target + ".enabled", "false").equals("true")) {
+                        pairs.add(new SourceAndTarget(source, target));
                     }
                 }
             }
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
@@ -90,10 +90,22 @@ public class MirrorMakerConfig extends AbstractConfig {
         List<SourceAndTarget> pairs = new ArrayList<>();
         Set<String> clusters = clusters();
         Map<String, String> originalStrings = originalsStrings();
+        boolean globalHeartbeatsEnabled = MirrorConnectorConfig.EMIT_HEARTBEATS_ENABLED_DEFAULT;
+        if (originalStrings.containsKey(MirrorConnectorConfig.EMIT_HEARTBEATS_ENABLED)) {
+            globalHeartbeatsEnabled = Boolean.valueOf(originalStrings.get(MirrorConnectorConfig.EMIT_HEARTBEATS_ENABLED));
+        }
+
         for (String source : clusters) {
             for (String target : clusters) {
                 if (!source.equals(target)) {
-                    if (originalStrings.getOrDefault(source + "->" + target + ".enabled", "false").equals("true")) {
+                    String clusterPairConfigPrefix = source + "->" + target + ".";
+                    boolean clusterPairEnabled = Boolean.valueOf(originalStrings.getOrDefault(clusterPairConfigPrefix + "enabled", "false"));
+                    boolean clusterPairHeartbeatsEnabled = globalHeartbeatsEnabled;
+                    if (originalStrings.containsKey(clusterPairConfigPrefix + MirrorConnectorConfig.EMIT_HEARTBEATS_ENABLED)) {
+                        clusterPairHeartbeatsEnabled = Boolean.valueOf(originalStrings.get(clusterPairConfigPrefix + MirrorConnectorConfig.EMIT_HEARTBEATS_ENABLED));
+                    }
+
+                    if (clusterPairEnabled || clusterPairHeartbeatsEnabled) {
                         pairs.add(new SourceAndTarget(source, target));
                     }
                 }
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
@@ -93,7 +93,9 @@ public class MirrorMakerConfig extends AbstractConfig {
             for (String target : clusters) {
                 SourceAndTarget sourceAndTarget = new SourceAndTarget(source, target);
                 if (!source.equals(target)) {
-                    pairs.add(sourceAndTarget);
+                    if (this.stringsWithPrefixStripped(source + "->" + target + ".").getOrDefault("enabled","false").equals("true")){
+                        pairs.add(sourceAndTarget);
+                    }
                 }
             }
         }
--- a/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
+++ b/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java
@@ -105,6 +105,11 @@ public class MirrorMakerConfig extends AbstractConfig {
                         clusterPairHeartbeatsEnabled = Boolean.valueOf(originalStrings.get(clusterPairConfigPrefix + MirrorConnectorConfig.EMIT_HEARTBEATS_ENABLED));
                     }
 
+                    // By default, all source->target Herder combinations are created even if `x->y.enabled=false`
+                    // Unless `emit.heartbeats.enabled=false` or `x->y.emit.heartbeats.enabled=false`
+                    // Reason for this behavior: for a given replication flow A->B with heartbeats, 2 herders are required :
+                    // B->A for the MirrorHeartbeatConnector (emits heartbeats into A for monitoring replication health)
+                    // A->B for the MirrorSourceConnector (actual replication flow)
                     if (clusterPairEnabled || clusterPairHeartbeatsEnabled) {
                         pairs.add(new SourceAndTarget(source, target));
                     }
--- a/connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorMakerConfigTest.java
+++ b/connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorMakerConfigTest.java
@@ -28,6 +28,7 @@ import java.util.Set;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Arrays;
+import java.util.List;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -254,6 +255,33 @@ public class MirrorMakerConfigTest {
             "secret2", bProps.get("ssl.key.password"));
         assertEquals("security properties should be transformed in worker producer config",
             "secret2", bProps.get("producer.ssl.key.password"));
+    }
+
+    @Test
+    public void testClusterPairs() {
+        MirrorMakerConfig mirrorConfig = new MirrorMakerConfig(makeProps(
+                "clusters", "a, b, c, d, e, f",
+                "a->b.enabled", "true",
+                "a->c.enabled", "true",
+                "a->d.enabled", "true",
+                "a->e.enabled", "false",
+                "a->f.enabled", "false"));
+        List<SourceAndTarget> clusterPairs = mirrorConfig.clusterPairs();
+        assertEquals("clusterPairs count should match x->y.enabled=true count",
+                3, clusterPairs.size());
+
+        // Link b->a.enabled doesn't exist therefore it must not be in clusterPairs
+        SourceAndTarget sourceAndTarget = new SourceAndTarget("b", "a");
+        assertFalse("disabled/unset link x->y should not be in clusterPairs", clusterPairs.contains(sourceAndTarget));
+    }
+
+    @Test
+    public void testEmptyClusterPairs() {
+        MirrorMakerConfig mirrorConfig = new MirrorMakerConfig(makeProps(
+                "clusters", "a, b, c, d, e, f"));
+        // expect 0 since no link x->y is enabled
+        assertEquals("clusterPairs count should match x->y.enabled=true count",
+                0, mirrorConfig.clusterPairs().size());
     }
 
     public static class FakeConfigProvider implements ConfigProvider {
--- a/connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorMakerConfigTest.java
+++ b/connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorMakerConfigTest.java
@@ -258,16 +258,59 @@ public class MirrorMakerConfigTest {
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


jchanaud:
After fixing the issue : 
````
[2020-11-13 15:53:53,935] INFO creating herder for replica_CENTRAL->replica_ZAR (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:53:58,597] INFO creating herder for replica_CENTRAL->replica_UGO (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:01,180] INFO creating herder for replica_CENTRAL->replica_HBG (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:03,656] INFO creating herder for replica_CENTRAL->replica_OLS (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:06,097] INFO creating herder for replica_CENTRAL->replica_UMO (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:08,186] INFO creating herder for replica_OLS->replica_CENTRAL (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:09,774] INFO creating herder for replica_CENTRAL->replica_CNO (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:12,253] INFO creating herder for replica_CENTRAL->replica_TST (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:14,407] INFO creating herder for replica_CENTRAL->replica_VLD (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:16,599] INFO creating herder for replica_CENTRAL->replica_ARA (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:19,005] INFO creating herder for replica_CENTRAL->replica_VIT (org.apache.kafka.connect.mirror.MirrorMaker)
[2020-11-13 15:54:21,789] INFO Kafka MirrorMaker starting with 11 herders. (org.apache.kafka.connect.mirror.MirrorMaker)
````

**Review by hachikuji — COMMENTED** [review]
Thanks for the patch. It would be helpful to have a test case.

**Comment by hachikuji** [comment]
Thanks for the patch. The proposed fix makes sense. I am only trying to confirm that this is a defect and not intended. The code seems to explicitly allow the connector to be created even when the link is disabled. Perhaps the cost of a disabled herder was not understood.

@ryannedolan maybe you could clarify?

**Comment by jchanaud** [comment]
Thank you for your help @hachikuji. I agree with your analysis and it kind of makes sense for most use-cases with 2-3 clusters as described in the original KIP.
I also understand that the original intent was that the ``clusterPairs`` variable would not grow exponential thanks to the command line parameter ``--clusters`` which is target-based (not source-based).
 
But this parameter doesn't help the business case we are implementing which can be better viewed as one "central cluster" and multiple "local clusters":
- Some topics (_schema is a perfect example) must be replicated down to every local cluster, 
- and some "local" topics will be replicated up to the central cluster. 

As stated in the KAFKA-10710, I cherry picked this commit in 2.5.2 and we are now running this build of MirrorMaker in production, because we can't ramp up deployment with current code as it stands.

**Comment by ryannedolan** [comment]
> The code seems to explicitly allow the connector to be created even when the link is disabled.
> @ryannedolan maybe you could clarify?

My intention was to ensure that MirrorHeartbeatConnector always runs, even when a link/flow is not otherwise needed. This is because MirrorHeartbeatConnector is most useful when it emits to all clusters, not just those clusters targeted by some flow.

For example, in a two-cluster environment with only A->B replicated, it is nice to have heartbeats emitted to A s.t. they get replicated to B. Without a Herder targeting A, there can be no MirrorHeartbeatConnector emitting heartbeats there, and B will see no heartbeats from A.

I know that some vendors/providers use heartbeats in this way, e.g. for discovering which flows are active and healthy. And I know that some vendors/providers don't use heartbeats at all, or use something else to send them (instead of MirrorHeartbeatConnector). Hard to say whether anything would break if we nixed these extra herders without addressing the heartbeats that would go missing.

IMO, we'd ideally skip creating the A->B herder whenever A->B.emit.heartbeats.enabled=false (defaults to true) and A->B.enabled=false (defaults to false). A top-level emit.heartbeats.enabled=false would then disable heartbeats altogether, which would trivially eliminate the extra herders. N.B. this would just be an optimization and wouldn't required a KIP, IMO.

Ryanne

**Comment by jchanaud** [comment]
@ryannedolan @hachikuji 
If I understand correctly, when setting up a link A->B.enabled=true (with defaults settings regarding heartbeat), it creates a topic heartbeat and produce beats on cluster B

I figured it was the other way around, so that the topic could be picked up by replication process (A->B), leading to a replicated topic A.heartbeat on cluster B (which we monitor actually). 

Looking into my configuration, I now understand that I was lucky because I have one link going up back up ``replica_OLS->replica_CENTRAL`` which is now the single emitter of beats (which are then replicated in every other cluster)

Reading the KIP led me to interpret that the production of heartbeat would be done within the same herder (the source one)  : 
> Internal Topics
> MM2 emits a heartbeat topic in each source cluster, which is replicated to demonstrate connectivity through the connectors.

**Comment by jchanaud** [comment]
@hachikuji ready for review, I believe failed CI is unrelated

**Comment by ryannedolan** [comment]
> Looking into my configuration, I now understand that I was lucky because I have one link going up back up replica_OLS->replica_CENTRAL which is now the single emitter of beats (which are then replicated in every other cluster)

Ah, that's lucky indeed. If you sever that link, I suppose your heartbeats will stop flowing across the entire topology. Also, if you look at the heartbeat message content, you'll probably find that all heartbeats are originating from that same cluster.

Sounds like disabling heartbeats altogether isn't what you want either. So we may need a way to emit heartbeats to source clusters without spinning up N*(N-1) herders just to do so.

What if we created N additional herders just for heartbeats to each cluster? Say you had 20 clusters and only one flow enabled (A->B). In that case you could have 21 herders total: one for the A->B flow and 20 for per-cluster heartbeats. That's at least better than 20*19=380 herders. This raises the question of what "source" to associated with heartbeats, but seems like a solvable problem.

Another idea is to have herders emit heartbeats _upstream_ within MirrorSourceConnector. In that case, the A->B herder would target B but, paradoxically, emit heartbeats to A. Hard to say if that would break anything, but as @twobeeb  points out, this may be a reasonable interpretation of the documentation.

**Comment by jchanaud** [comment]
@ryannedolan I'm fine with either solution you suggest. What's the next step here ? I'm happy to contribute but I need your help on which direction we should take.

**Comment by ryannedolan** [comment]
What if we just run herders in both directions (A->B and B->A) whenever replication is enabled in one or both directions? That would require very minimal code changes, and only in the driver, while yielding the results you are expecting. Namely, we'd still see heartbeats on every _source_ cluster, and we'd still see heartbeats get replicated along each replication flow, but we would eliminate a ton of extraneous herders in the scenario you've illustrated.

Specifically, if we see A->B.enabled, we run MirrorSourceConnector and MirrorCeckpointConnector in herder A->B, and we run MirrorHeartbeatConnector in herder B->A. But if neither A->B nor B->A is enabled, then we don't need to create either herder at all.

This would change the behavior of the MM2 driver very slightly, since we'd no longer emit heartbeats to clusters that aren't part of any flow. But that doesn't really seem like a breaking change to me. "If a tree falls in the woods" and all that.

What do you think @twobeeb ?

**Comment by jchanaud** [comment]
@ryannedolan 
I have a preference for your first suggestion (the one for which there is currently a code proposition) because it doesn't alter the behavior of MM2 in any way for existing users.  

> IMO, we'd ideally skip creating the A->B herder whenever A->B.emit.heartbeats.enabled=false (defaults to true) and A->B.enabled=false (defaults to false). A top-level emit.heartbeats.enabled=false would then disable heartbeats altogether, which would trivially eliminate the extra herders. N.B. this would just be an optimization and wouldn't required a KIP, IMO.

That being said, I'm also fine with your latest proposition. It fits perfectly to my personnal use case because topics will be replicated local to central as well as central to local in the near future ; all instanciated herders will then transmit data (and not just beats).

Could you confirm and/or elaborate on the behavior you want to see when playing with ``emit.heartbeats.enabled``  ?

I guess that'd be good test cases : 

**Simple A to B**
````
A->B.enabled=true
````
Expected : 2 herders
  
**A to B and A to C**
```` 
clusters=A, B, C
A->B.enabled=true
A->C.enabled=true
````
 Expected : 4 herders (AB BA AC CA)

**Disabled heartbeats**
```` 
clusters=A, B, C
A->B.enabled=true
A->B.emit.heartbeats.enabled=false
B->A.emit.heartbeats.enabled=false
A->C.enabled=true
A->C.emit.heartbeats.enabled=false
C->A.emit.heartbeats.enabled=false
````
Expected : 4 herders (2 of which are unused, but still much better than N*N-1)

**Comment by jchanaud** [comment]
Hi @ryannedolan @hachikuji 

I'd like a definitive direction from you on this subject. 

Once again, I reiterate on my personal preference as expressed previously by @ryannedolan :
> IMO, we'd ideally skip creating the A->B herder whenever A->B.emit.heartbeats.enabled=false (defaults to true) and A->B.enabled=false (defaults to false). A top-level emit.heartbeats.enabled=false would then disable heartbeats altogether, which would trivially eliminate the extra herders. N.B. this would just be an optimization and wouldn't required a KIP, IMO.

Please advise.

Julien

**Comment by ryannedolan** [comment]
> I have a preference for your first suggestion ... because it doesn't alter the behavior of MM2 in any way for existing users.

Good point, and also it doesn't conflict with the other ideas if we want to try something later. Works for me.

File: connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorMakerConfig.java [hunk 1]
```diff
@@ -89,11 +89,25 @@ public MirrorMakerConfig(Map<?, ?> props) {
...
                 if (!source.equals(target)) {
-                    pairs.add(sourceAndTarget);
+                    String clusterPairConfigPrefix = source + "->" + target + ".";
+                    boolean clusterPairEnabled = Boolean.valueOf(originalStrings.getOrDefault(clusterPairConfigPrefix + "enabled", "false"));
+                    boolean clusterPairHeartbeatsEnabled = globalHeartbeatsEnabled;
+                    if (originalStrings.containsKey(clusterPairConfigPrefix + MirrorConnectorConfig.EMIT_HEARTBEATS_ENABLED)) {
+                        clusterPairHeartbeatsEnabled = Boolean.valueOf(originalStrings.get(clusterPairConfigPrefix + MirrorConnectorConfig.EMIT_HEARTBEATS_ENABLED));
+                    }
+
+                    if (clusterPairEnabled || clusterPairHeartbeatsEnabled) {
```
skaundinya15, line 113: Question @twobeeb and @ryannedolan about this: is there a use case for which a cluster pair would be disabled yet its heartbeats enabled? I'm wondering because in the case someone disabled the cluster pair but didn't think to disable the heartbeats (since it defaults to true), these herders would still be created, but I'm not sure if it would be useful to have them created since there is no replication going on. Would it make more sense to just not create a herder if the cluster pair was disabled (instead of checking the heartbeats as well)? If not can we make sure to add documentation about this behavior?

jchanaud: That's actually the current behavior, all herders are created and the beats are emitted from the "opposite" herder  

If you have a replication flow from A to B **and you want heartbeats**, you need 2 herders :  
- A->B for the MirrorSourceConnector 
- B->A for the MirrorHeartbeatConnector 

The MirrorHeartbeatConnector on B->A emits beats into topic heartbeats on cluster A.
The MirrorSourceConnector on A->B then replicates whichever topic is configured as well as heartbeats.

All of this is fine with 2-3 clusters. It starts to make less sense when using 10+ clusters.

Because this is the current implementation, we decided to keep it as-is so as to not cause change in behavior, and add a new top-level property "emit.heartbeats.enabled" which defaults to "true".

Existing users will not see any change and if they depend on these "opposites" herders for their monitoring, it will still work.  
New users with more complex use case can change this property and fine tune their heartbeat generation.

skaundinya15: Thanks for the explanation @twobeeb, this makes sense. It would be good to add some comments explaining this in the code as this isn't immediately obvious. Other than that it looks good to me overall.

jchanaud: Thanks for your review @skaundinya15.
I'm having a hard time phrasing this properly, suggestions would be welcome.
Is this comment proposition aligned with what you had in mind ?
```suggestion
                    // By default, all source->target Herder combinations are created even if `x->y.enabled=false`
                    // Unless `emit.heartbeats.enabled=false` or `x->y.emit.heartbeats.enabled=false`
                    // Reason for this behavior: for a given replication flow A->B with heartbeats, 2 herders are required :
                    // B->A for the MirrorHeartbeatConnector (emits heartbeats into A)
                    // A->B for the MirrorSourceConnector (actual replication flow)
                    if (clusterPairEnabled || clusterPairHeartbeatsEnabled) {
```

skaundinya15:
```suggestion
                    // By default, all source->target Herder combinations are created even if `x->y.enabled=false`
                    // Unless `emit.heartbeats.enabled=false` or `x->y.emit.heartbeats.enabled=false`
                    // Reason for this behavior: for a given replication flow A->B with heartbeats, 2 herders are required :
                    // B->A for the MirrorHeartbeatConnector (emits heartbeats into A for monitoring replication health)
                    // A->B for the MirrorSourceConnector (actual replication flow)
                    if (clusterPairEnabled || clusterPairHeartbeatsEnabled) {
```
Looks good to me, just had a small tweak for the `B->A` comment. Thanks!

hachikuji (approved):
LGTM. Thanks for the patch!

hachikuji:
@twobeeb Before I merge, would you mind updating the PR description? Also, I will leave it to you to add the doc suggestion from @skaundinya15.

jchanaud:
@hachikuji
I included the comment from @skaundinya15 and changed the PR description to reflect the actual design choice.

Let me know if that's ok for you.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorMakerConfigTest.java
```diff
@@ -256,6 +257,80 @@ public void testWorkerConfigs() {
...
+        // b->a.enabled=false
+        // b->a.emit.heartbeat.enabled=true
+        // b->c.enabled=false
+        // b->c.emit.heartbeat.enabled=true
+        // c->a.enabled=false
+        // c->a.emit.heartbeat.enabled=true
+        // c->b.enabled=false
+        // c->b.emit.heartbeat.enabled=true
+        List<SourceAndTarget> clusterPairs = mirrorConfig.clusterPairs();
+        assertEquals("clusterPairs count should match all combinations count",
```
aloknnikhil, line 278: @twobeeb Could you migrate these to use the new JUnit 5 `assertEquals` API? Looks like the build actually failed here - https://github.com/apache/kafka/runs/1782294502

ijuma: I fixed the build via https://github.com/apache/kafka/commit/bf4afae8f53471ab6403cbbfcd2c4e427bdd4568

PR #9590, opened 2020-11-12
KAFKA-7556: KafkaConsumer.beginningOffsets does not return actual first offsets

jolshan:
The current behavior for KafkaConsumer.beginningOffsets on compacted topics is to always return offset 0. However, when looking at the record with the earliest timestamp in the log with KafkaConsumer.offsetsForTimes, the offset is nonzero. 

This PR seeks to correct this issue by updating the segment baseOffsets and the logStartOffset during compaction. 
The segment offset will be the offset of the first non-compacted batch, or the original baseOffset if the segment is empty. 
Generally, all updates to logStartOffset must maintain the invariant logStartOffset <= end offset of first batch that can be returned by Fetch.

There are a few key changes/decisions to note that may not be immediately apparent:

1. **Producer Snapshot Behavior** Currently the behavior of producer snapshots is that during compaction, if the segment offset of the producer snapshot file is not in the new segment offsets, the snapshot is deleted. This will lead to behavior like the following:

- So for example we have segments with offsets 0 and 1, we want to delete first segment
New segment is base offset 0 so snapshot for segment 1 deleted. One segment, no snapshot files

Under this PR, the new behavior is this

- We have segments with offsets 0 and 1, we want to delete first segment
New segment is base offset 1 so we don't delete snapshot. One segment, one snapshot file

I was not sure if this is behavior we wanted, so I'm pointing it out.

2.  **High Watermark Compaction** I wasn't sure if we allowed compaction past the high watermark. It seems like there is not a check to prevent it. I've added such a check for this PR.


core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1306,7 +1309,7 @@ class Log(@volatile private var _dir: File,
     // in an unclean manner within log.flush.start.offset.checkpoint.interval.ms. The chance of this happening is low.
     maybeHandleIOException(s"Exception while increasing log start offset for $topicPartition to $newLogStartOffset in dir ${dir.getParent}") {
       lock synchronized {
-        if (newLogStartOffset > highWatermark)
+        if (reason != SegmentCompaction && newLogStartOffset > highWatermark)
```
jsancio: I don't think it is safe for replicas to compact past the `highWatermark`. Should we remove this change and make sure that log compaction doesn't compact past the `highWatermark`.

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -538,6 +539,14 @@ private[log] class Cleaner(val id: Int,
     for (group <- groupedSegments)
       cleanSegments(log, group, offsetMap, deleteHorizonMs, stats, transactionMetadata)
 
+    if (baseOffset == log.logStartOffset) {
```
jsancio: Why the check? Shouldn't the log cleaner always perform the algorithm in the `if` block?

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -711,6 +723,9 @@ private[log] class Cleaner(val id: Int,
           shallowOffsetOfMaxTimestamp = result.shallowOffsetOfMaxTimestamp,
           records = retained)
         throttler.maybeThrottle(outputBuffer.limit())
+        if (newCanUpdateBaseOffset)
+          dest.updateBaseOffset(result.minOffset())
+        newCanUpdateBaseOffset = false
```
jsancio: Can we move this code to `LogSegment::append`? Meaning shouldn't the `baseOffset` of a segment always be the `baseOffset` of the first record batch appended?

Is the issue that the new clean segment is created before the log cleaner knows the first clean record (record batch). If so, why don't we delay creating of the new segment until that point?

jsancio (commented):
Thanks for the PR @jolshan. Left some high-level comments and questions.

core/src/main/scala/kafka/log/LogCleaner.scala
jolshan: Good catch. The if statement is left from an earlier implementation. I believe cleanable.log will always be the entire log and always have baseOffset be log.logStartOffset.

core/src/main/scala/kafka/log/LogCleaner.scala
jolshan: Yeah. I'm not totally happy with how this is handled, but I'm not sure if there is another better way. So the new cleaned segment is created in cleanSegments which calls this method. The issue is that multiple segments can be compacted into a single segment and each time a segment is compacted this method is called. Thus, this method does not always know if there is a record in the new segment yet.  This is why we have this extra boolean variable. 

This is also why delaying might not work without a variable like this. The method will not know if the new segment has been created already or not.

core/src/main/scala/kafka/log/Log.scala
jolshan: I talked to Jason about this a bit and since we don't clean the active segment, cleaning past the high watermark is pretty rare. But I think it makes sense to explicitly enforce this in code.

I think one other change we will need to make is to update a lot of our log cleaner tests since a large portion of them clean past the high watermark. (They don't update the high watermark at all)

core/src/main/scala/kafka/log/LogCleaner.scala
jsancio: Okay. We can do this in a future PR if you like. Do you want to file an issue for this?

core/src/main/scala/kafka/log/Log.scala
jsancio: Okay. We can do this in a future PR if you like. Do you want to file an issue for this?

core/src/main/scala/kafka/log/LogCleaner.scala
jsancio: I think one solution is to change the signature to:

```scala
private[log] def cleanInto(
    log: Log
    sourceRecords: FileRecords,
    dest: Option[LogSegment],
    map: OffsetMap,
    retainDeletesAndTxnMarkers: Boolean,
    maxLogMessageSize: Int,
    transactionMetadata: CleanedTransactionMetadata,
    lastRecordsOfActiveProducers: Map[Long, LastRecord],
    stats: CleanerStats): Option[LogSegment]
```

This should allow you to call `LogCleaner.createNewCleanedSegment(log, result.minOffset)` when a cleaned batch has been identified. What do you think?

jolshan: This is much more elegant! A question I would have is what we do if we return from trying all the segments and we end up never making a new segment. Currently the behavior is to return an empty segment with the original baseOffset. I'd lean towards keeping the existing logic, but I suppose we could also simply use deleteSegments instead of replaceSegments in this case too.

jolshan: I've updated to delete segments that end up being empty.

jolshan:
I've updated the code to delay creating a new segment until there is a non-compacted record. If a segment is never created in cleanSegments, the old segments are simply deleted rather than replaced. I had to change some code surrounding the transactionMetadata that allows a delay before updating the transactionIndex of a segment until the segment is actually created. Any aborted transactions will be added once the segment is created.

jolshan:
Also looks like the test I added may be flaky, so I'll take a look at that.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1749,6 +1752,9 @@ class Log(@volatile private var _dir: File,
           checkIfMemoryMappedBufferClosed()
           // remove the segments for lookups
           removeAndDeleteSegments(deletable, asyncDelete = true, reason)
+          if (reason == LogCompaction) {
+            maybeIncrementLogStartOffset(segments.firstEntry.getValue.baseOffset, SegmentCompaction)
+          }
           maybeIncrementLogStartOffset(segments.firstEntry.getValue.baseOffset, SegmentDeletion)
```
ijuma: Shouldn't we be passing `reason` instead of what we're doing here?

ijuma: Looks like the types are not the same, but this approach seems error prone.

chia7712:
(I'm still reading this story so pardon me for asking stupid question.)

If using timestamp=0 can bring correct offset, why not following that way to handle ```ListOffsetRequest.EARLIEST_TIMESTAMP```? For example, (https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/log/Log.scala#L1643) if timestamp is equal to ```ListOffsetRequest.EARLIEST_TIMESTAMP```, we pass 0 to find offset. Does it work?

core/src/main/scala/kafka/log/Log.scala
jolshan: Sorry if this is a little confusing. `reason` is for deletion and the other parameter `SegmentCompaction` is the reason for changing the offset. Currently that is used to handle the issue for updating past the high watermark. I agree that it is not the cleanest way and we should figure out what we want to do with compacting past the high watermark.

jolshan:
@chia7712 The code path you linked is the code path beginningOffsets uses. Are you suggesting removing `ListOffsetRequest.EARLIEST_TIMESTAMP` and replacing it with 0?

chia7712:
> Are you suggesting removing ListOffsetRequest.EARLIEST_TIMESTAMP and replacing it with 0? 

yep, it seems like a simple solution without much changes. However, I have not understood this issue totally. It is just my imagination.

jolshan:
@chia7712 I did think about this solution initially. I'm wondering if we do want to update the segment offsets and logStartOffsets correctly though. If that isn't as important then maybe we can go with the simpler solution. This solution removes empty segments and keeps the baseOffsets updated. If these baseOffsets are used in other places, then maybe this solution is better.

jolshan:
@chia7712 Were you able to take a closer look at this?

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
```diff
@@ -386,6 +390,10 @@ public long maxOffset() {
             return maxOffset;
         }
 
+        public long minOffset() {
```
hachikuji: Can we modify some of the test cases in `MemoryRecordsTest` to ensure this is getting set correctly?

It might also be useful to document what exactly the min offset is with respect to batches and records.

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
```diff
@@ -198,7 +198,7 @@ private static FilterResult filterTo(TopicPartition partition, Iterable<MutableR
             if (!retainedRecords.isEmpty()) {
                 if (writeOriginalBatch) {
                     batch.writeTo(bufferOutputStream);
-                    filterResult.updateRetainedBatchMetadata(batch, retainedRecords.size(), false);
+                    filterResult.updateRetainedBatchMetadata(batch, retainedRecords.get(0).offset(), retainedRecords.size(), false);
```
hachikuji: Hmm... It seems a little inconsistent to use the offset of the first record. When the batch is empty, we use the base offset of the batch. Shouldn't we do the same here? Otherwise we will end up with some batches which have base offset smaller than the segment base offset. Note that the base offset is always preserved by compaction.

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
jolshan: Sounds good

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
jolshan: I see. I guess that the goal here was to give the actual offset of the first record. If we choose to give the batch offset instead, we may return an offset that is still not the actual first offset (but will be much closer than 0). If this behavior is acceptable we can go with this option, but it will not match the behavior of consumer.offsetsForTimes() with timestamp 0L, as mentioned in KAFKA-7556.

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
jolshan: I've added some unresolved comments about min offsets below.

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
hachikuji: I don't think we have much choice given retention semantics. I think we have to keep the invariant that the segment base offset is less than or equal to the base offset of any batch contained in it. Note that these batches are still returned in Fetch requests, so it seems reasonable to let the log start offset reflect them.

jolshan: Just to clarify, I understand that there are issues with the segment base offset mismatching with the base offset. 
But would there also be issues if the batch offset was lower than the log start offset as well?

hachikuji: I believe that a call to `DeleteRecords` can advance the start offset to an arbitrary offset which could be in the middle of a batch. I do not think today that we have any hard guarantees that we won't return batches or even records that are lower than the current log start offset. It might be worth experimenting with this to clarify what the behavior is today. With that said, given the fact that we always return the whole batch anyway, my preference would probably be to try and keep the log start offset aligned on a batch boundary.

jolshan: I see. I'll look into that behavior.

jolshan: Another option I considered that could give the exact offset without changing any offsets in the log would be to handle the listOffset request for the earliest timestamp on compacted topics differently. Instead of grabbing the log start offset, it could behave as though the request was a call for timestamp 0 and search for the first offset that is >=0 (as it does for any other timestamp search) Would this be a preferable solution if we wanted to ensure the exact offset?

hachikuji: I think we're on the right track really. We don't have to implement exactly what the jira describes. One thing to keep in mind is that the first record in the log might be a control record which a consumer won't return to applications anyway. Also note that the log start offset has implications for replication. Only the records above the log start offset will be replicated to a new replica, but all of the retained records after cleaning should be replicated. Hence we already have some constraints which make keeping the log start offset aligned with the first returnable record difficult.

I would just say that we define and document a reasonable invariant. I think it should be clear that we need to ensure the segment base offset is less than or equal to the base offset of the first batch contained in it. The invariant I would suggest is that the log start offset should be less than or equal to the end offset of the first batch that can be returned from `Fetch`. For the case of log cleaning, the simplest thing to do is keep the log start offset aligned with the base offset of the first batch. In the case of `DeleteRecords`, the log start offset will point to some offset within the first batch.

jolshan:
Updated to match invariants discussed above, still need to look at a few other areas like producer snapshots and high watermark

jolshan:
Updated to not clean past high watermark. If this should be its own ticket/PR, let me know.

core/src/main/scala/kafka/log/Log.scala
jolshan: I've removed this now that I've changed to not clean past the high watermark.

core/src/main/scala/kafka/log/Log.scala
jolshan: Added to this PR, may still want to do separately if this has gotten too big.

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
```diff
@@ -198,7 +198,7 @@ private static FilterResult filterTo(TopicPartition partition, Iterable<MutableR
             if (!retainedRecords.isEmpty()) {
                 if (writeOriginalBatch) {
                     batch.writeTo(bufferOutputStream);
-                    filterResult.updateRetainedBatchMetadata(batch, retainedRecords.size(), false);
+                    filterResult.updateRetainedBatchMetadata(batch, batch.baseOffset(), retainedRecords.size(), false);
```
hachikuji: nit: why do we need to pass the base offset as a separate parameter? We are already passing `batch`.

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
```diff
@@ -330,27 +332,31 @@ public int hashCode() {
         private int messagesRetained = 0;
         private int bytesRetained = 0;
         private long maxOffset = -1L;
+        // Retains invariant `minOffset` = the baseOffset of the first batch of retained records
+        // or -1 if no batches are retained
+        private long minOffset = -1L;
```
hachikuji: It's more verbose, but perhaps we should choose a more explicit name like `baseOffsetOfFirstBatch`. Then we could either get rid of the comment or move to the getter for this field (as a javadoc).

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -582,13 +586,16 @@ private[log] class Cleaner(val id: Int,
         transactionMetadata.addAbortedTransactions(abortedTransactions)
 
         val retainDeletesAndTxnMarkers = currentSegment.lastModified > deleteHorizonMs
-        info(s"Cleaning $currentSegment in log ${log.name} into ${cleaned.baseOffset} " +
+        info(s"Cleaning $currentSegment in log ${log.name} into ${currentSegment.baseOffset} " +
```
hachikuji: This log message no longer makes sense. Is the intent to use `cleanedSegment` instead of `currentSegment`? Perhaps we should consider moving this log line to after the cleaning when we know what the new segment is.

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -582,13 +586,16 @@ private[log] class Cleaner(val id: Int,
...
           s"with deletion horizon $deleteHorizonMs, " +
           s"${if(retainDeletesAndTxnMarkers) "retaining" else "discarding"} deletes.")
 
         try {
-          cleanInto(log.topicPartition, currentSegment.log, cleaned, map, retainDeletesAndTxnMarkers, log.config.maxMessageSize,
+           cleanedSegment = cleanInto(log, log.topicPartition, currentSegment.log, cleanedSegment, map, retainDeletesAndTxnMarkers, log.config.maxMessageSize,
```
hachikuji: nit: it shouldn't be necessary to pass both `log` and `log.topicPartition`.

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -599,21 +606,30 @@ private[log] class Cleaner(val id: Int,
...
-      // update the modification date to retain the last modified date of the original files
-      val modified = segments.last.lastModified
-      cleaned.lastModified = modified
-
-      // swap in new segment
-      info(s"Swapping in cleaned segment $cleaned for segment(s) $segments in log $log")
-      log.replaceSegments(List(cleaned), segments)
+      
+      // Result of cleaning included at least one record.
+      if (cleanedSegment.isDefined) {
```
hachikuji: nit: use `match`

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -599,21 +606,30 @@ private[log] class Cleaner(val id: Int,
...
+        // flush new segment to disk before swap
+        cleaned.flush()
+
+        // update the modification date to retain the last modified date of the original files
+        val modified = segments.last.lastModified
+        cleaned.lastModified = modified
+
+        // swap in new segment
+        info(s"Swapping in cleaned segment $cleaned for segment(s) $segments in log $log")
+        log.replaceSegments(List(cleaned), segments)
```
hachikuji: Would it make more sense to update log start offset in `replaceSegments` when we are protected by the lock? Otherwise, it seems like the log start offset will be out of sync with the segments until all of them have been cleaned and `maybeIncrementLogStartOffset` is called. 

Note that `deleteSegments` does update the log start offset. Interestingly, `SegmentDeletion` is used for the reason in this case, and not `SegmentCompaction`.

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -1094,6 +1119,8 @@ private[log] class CleanedTransactionMetadata {
 
   // Output cleaned index to write retained aborted transactions
   var cleanedIndex: Option[TransactionIndex] = None
+  
+  var toAppend: List[AbortedTxn] = List.empty
```
hachikuji: Can you help me understand why we needed to change the logic for appending to the transaction index?

jolshan: I believe the issue is that this code expects the log to be created already but it is not. I think my intention was to delay this operation until the log is created. I'll double check this is actually the case.

core/src/main/scala/kafka/log/LogCleaner.scala
jolshan: Hmmm I think I didn't really understand this log at first. It does make sense to move it to when we know what the new segment is. I think the issue I had was what to do with including `retainDeletesAndTxnMarkers`. Maybe it makes sense to have a log saying the current segment and the deletion info and a second log to say what the new segment's baseOffset is later.

core/src/main/scala/kafka/log/LogCleaner.scala
jolshan: Right. I wasn't sure if it made sense to change the deleteSegments code to support the SegmentCompaction reason. The entire segment is being deleted here, but I see how this may be confusing when reading in the log.

jolshan: As for the first part about `replaceSegments`. It seems that there are a few other times we call this method. I'm wondering if we would want to update the logStartOffset in these cases (probably) and what the reason should be. I'm also wondering if we should include the reason as a parameter to `replaceSegments`.

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -701,11 +719,17 @@ private[log] class Cleaner(val id: Int,
       // if any messages are to be retained, write them out
       val outputBuffer = result.outputBuffer
       if (outputBuffer.position() > 0) {
+        if (destSegment.isEmpty) {
+          // create a new segment with a suffix appended to the name of the log and indexes
+          destSegment = Some(LogCleaner.createNewCleanedSegment(log, result.minOffset()))
+          transactionMetadata.cleanedIndex = Some(destSegment.get.txnIndex)
```
jolshan: Re: changing transaction code
Since cleanedIndex is not created until here, we need to delay appends. Originally it was created at the start of `cleanSegments`. In that method, we now can not guarantee the log has been created until `cleanInto` returns with a defined `cleanedSegment`. In `cleanInto` there is a chance to call `onControlBatchRead` which previously just appeneded transactions, but can no longer do so if `cleanedIndex` is not defined.

jolshan: It may be possible to call `transactionMetadata.appendTransactionIndex()` here though to append a little bit sooner.

jolshan: Another option is to check if the `cleanedIndex` exists and append immediately if it does, otherwise put it in `toAppend`. Then we apply the changes only when the `cleanedIndex` is first defined.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -209,6 +209,11 @@ case object LeaderOffsetIncremented extends LogStartOffsetIncrementReason {
 case object SegmentDeletion extends LogStartOffsetIncrementReason {
   override def toString: String = "segment deletion"
 }
+
+case object SegmentCompaction extends LogStartOffsetIncrementReason {
```
hachikuji: It seems a little more intuitive if we swap the two new names:

```scala
case object LogCompaction extends LogStartOffsetIncrementReason
case object SegmentCompaction extends SegmentDeletionReason
```

core/src/main/scala/kafka/log/LogCleaner.scala
hachikuji: There are really only two cases for `replaceSegments`: log compaction, and segment splitting. We also call it in `completeSwapOperations`, but that is for recovery if the broker crashes in the middle of `replaceSegments`. It should be safe in either case to adjust the log start offset. Probably something similar to the deletion case which just ensures that the log start offset did not fall off the log:
```scala
maybeIncrementLogStartOffset(segments.firstEntry.getValue.baseOffset, SegmentDeletion)
```
Ideally `LogCompaction` would be the reason, at least in the log compaction case. It's becoming cumbersome to pass through the reason in all cases it is used, but I'm not sure I see an alternative. Perhaps we can consider how to refactor this in the future. Maybe there should be some kind of context object which encapsulates the higher level operation that is being done (such as log compaction or retention enforcement).

core/src/main/scala/kafka/log/LogCleaner.scala
hachikuji: Not sure I follow what the complication is. We can still mention whether deletes/markers have been retained after the fact. It would be useful to clearly indicate when no records were retained from a segment though.

jolshan: That makes sense.

core/src/main/scala/kafka/log/LogCleaner.scala
hachikuji: Ok, I think I understand. One thing that makes this a little confusing is the need to reverse the collection. Why don't just build the index in order?

I am debating whether this is good enough. A potential problem is that we don't have a guarantee on the size of the index. In the common case, it should be small, but there is nothing to prevent an entire segment from being full of aborted transaction markers. Currently we wait until `filterInto` returns before creating the new segment, but maybe instead we could do it in `checkBatchRetention` after the first retained batch is observed?

core/src/main/scala/kafka/log/LogCleaner.scala
jolshan: I think I was concerned about putting this information in the try block and it not logging due to LogSegmentOffsetOverflowException. But I looking at it again, it's not a huge deal not to log in that case.

core/src/main/scala/kafka/log/Log.scala
jolshan: I took a quick look at the others that extend SegementDeletionReason and LogStartOffsetIncrementReason. I think this is why I chose Log and Segment respectively.
```
RetentionMsBreach extends SegmentDeletionReason 
RetentionSizeBreach extends SegmentDeletionReason 
StartOffsetBreach extends SegmentDeletionReason
LogRecovery extends SegmentDeletionReason
LogTruncation extends SegmentDeletionReason 
LogRoll extends SegmentDeletionReason 
LogDeletion extends SegmentDeletionReason 
```

```
ClientRecordDeletion extends LogStartOffsetIncrementReason
LeaderOffsetIncremented extends LogStartOffsetIncrementReason 
SegmentDeletion extends LogStartOffsetIncrementReason 
SnapshotGenerated extends LogStartOffsetIncrementReason 
```

jolshan: But I think it also makes sense to switch them.

core/src/main/scala/kafka/log/LogCleaner.scala
jolshan: Oh I think the reverse is just because of how I append the transactions to the front of the list instead of the end. There's a better way to do that, so I'll fix it.

jolshan: Locally I have it filterTo returns. I'll look for a better way to do it.

jolshan: I can do it in `onControlBatchRead` directly I believe. I'll add to toAppend when there is no `cleanedIndex` and as soon as there is, I'll add those transactions in `filterInto` as soon as the `cleanedIndex` is created. From then on, I can just add the transactions in `onControlBatchRead` as before.

hachikuji:
@jolshan In `Log.completeSwapOperations`, we have the following comment:
```
      // We create swap files for two cases:
      // (1) Log cleaning where multiple segments are merged into one, and
      // (2) Log splitting where one segment is split into multiple.
      //
      // Both of these mean that the resultant swap segments be composed of the original set, i.e. the swap segment
      // must fall within the range of existing segment(s). If we cannot find such a segment, it means the deletion
      // of that segment was successful. In such an event, we should simply rename the .swap to .log without having to
      // do a replace with an existing segment.
```
This is no longer accurate following the change here. As I understand it, the swap recovery logic looks for each file with the `.swap` extension. It then tries to find all log segments which have been deleted by the compaction process by searching for any segments which have an end offset larger than the swap segment's base offset, but less than the offset of the first record in the swapped segment. This is handling the case when the broker crashes before it could delete the old segments.

The recovery algorithm is described in the comment above `replaceSegments`. I believe the logic still works correctly if we change it so that segments that are smaller than the first swap segment get removed. This should work because the swap suffix is removed in descending order, which means that the smallest swap file is the last one to be updated. I think we will need some test cases to ensure that the various cases work correctly.

jolshan:
@hachikuji thanks for the review. Just to clarify, you are suggesting updating/changing
```
val oldSegments = logSegments(swapSegment.baseOffset, swapSegment.readNextOffset).filter { segment =>
        segment.readNextOffset > swapSegment.baseOffset
      }
``` 
in `Log.completeSwapOperations`?
When you say "removed" do you mean the segments should be deleted or not included in `oldSegments`? 
When you say "smaller" do you mean a lower offset? 
My understanding is that the swap segment's baseOffset can now have a higher offset than the baseOffsets of the oldSegments it is replacing. So I'm thinking we are trying to delete all segments before the swap segment's baseOffset as well? Is there a case where we can't assume all segments before the baseOffset can be deleted?

And yes, will definitely add tests

hachikuji:
@jolshan The idea is to mimic during recovery the same operation that the log cleaner was trying to do. So if we were trying to replace segments [0, 10, 20] with [15] when the operation failed, then we'd want to do the same thing during recovery. The recovery logic currently assumes that a replacement like this is not possible, so we need to adjust it. For example, we might be left with the follower files on disk when the broker starts up:
```
0.log
10.log
20.log
30.log
40.log
15.swap
30.swap
```
We would want the recovery to generate two replacements: [0, 10, 20] -> 15 and [30, 40] -> 30. The tricky thing is that the recovery operation itself might fail, so we have to basically pick up the operation from the same point and continue with the same algorithm (preserving order and all)

jolshan:
@hachikuji In the example above, we could also have something like `35.swap` too, right, since we are updating the baseOffset of each segment. (Which means we will need to look at the next segment's offsets as well) Can we also assume that all the log will be replaced? That is, `15.swap` is the lowest swap file, so it must cover `0.log`

hachikuji:
@jolshan The active segment would not be replaced during cleaning at a minimum. I think the recovery logic we want is to scan backwards through the `.swap` files and collect the segments to replace. Maybe something like this?

For each swap file in reverse order:
- If the next earliest swap file exists, then collect any segments which have a has a base offset larger than the end offset in the next earliest swap and an end offset less than that of the swap segment itself.
- If no next earliest swap file exists, then collect all segments which have an end offset less than that of the swap file itself.
- Call `replaceSegments` with the swap file and the collected log segments

jolshan:
@hachikuji updated based on the conversations here + those we had offline. Quick summary is that we will only have one swap file upon crash in the log cleaning case, so we will just accept that there may be segment(s) before or after the cleaned segment that won't get replaced upon crash recovery, but will upon subsequent cleanings.

Two things about the latest commit:
1) I decided that in the failure case the logStartOffset will be updated outside of replace segments. In the recovery path there is already code to update the logStartOffset. It must follow changes to update the log end offset so it made sense to handle that all there. Added a test for this scenario. 

2) **EDIT: fixed** 
~Let me know if any of the tests seem unnecessary. I think we should definitely keep `testRecoveryAfterCrashBaseOffsetUpdatedFirstSegmentChunk` and `testLogStartOffsetUpdatedUponFailureRecovery`. Let me know if any others seem superfluous.~

jolshan:
Realized I left out one case. Going to add that and clean up tests.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -760,6 +768,12 @@ class Log(@volatile private var _dir: File,
       // must fall within the range of existing segment(s). If we cannot find such a segment, it means the deletion
       // of that segment was successful. In such an event, we should simply rename the .swap to .log without having to
       // do a replace with an existing segment.
+      //
+      // For case 1 (log cleaning), we may have old segments before or after the swap segment that were cleaned.
```
hachikuji: Can we clarify the comment above that "the swap segment must fall within the range of existing segment(s)"? I find it a bit confusing in the context of the new behavior.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -760,6 +768,12 @@ class Log(@volatile private var _dir: File,
...
+      // Unfortunately, since the baseOffset and the readNextOffset were changed, these segments will not be removed on
+      // recovery if they were not yet given a DeletedFileSuffix. A subsequent cleaning that succeeds will correctly remove these segments.
+      // ie. segments [0, 1000), [1000, 2000), [2000, 3000) cleaned into [1500, 1750).swap without marking old segments with DeletedFileSuffix
+      // -> [0. 1000), [1500, 1750), [2000, 3000)
```
hachikuji: nit: `[0, 1000)`

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -760,6 +768,12 @@ class Log(@volatile private var _dir: File,
       // must fall within the range of existing segment(s). If we cannot find such a segment, it means the deletion
       // of that segment was successful. In such an event, we should simply rename the .swap to .log without having to
       // do a replace with an existing segment.
+      //
+      // For case 1 (log cleaning), we may have old segments before or after the swap segment that were cleaned.
+      // Unfortunately, since the baseOffset and the readNextOffset were changed, these segments will not be removed on
+      // recovery if they were not yet given a DeletedFileSuffix. A subsequent cleaning that succeeds will correctly remove these segments.
+      // ie. segments [0, 1000), [1000, 2000), [2000, 3000) cleaned into [1500, 1750).swap without marking old segments with DeletedFileSuffix
```
hachikuji: Maybe this would be a more interesting example:
```
// ie. segments [0, 1000), [1000, 2000), [2000, 3000), [3000, 4000) cleaned into [1500, 2500).swap without marking old segments with DeletedFileSuffix
```
The filter logic below would capture both [1000, 2000) and [2000, 3000) as the old segments which will be replaced with [1500, 2500). Do I have that right? So in other words, the recovery logic replaces any overlapping segments.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2393,6 +2407,10 @@ class Log(@volatile private var _dir: File,
       }
       // okay we are safe now, remove the swap suffix
       sortedNewSegments.foreach(_.changeFileSuffixes(Log.SwapFileSuffix, ""))
+
+      // If not recovered swap file we need to increment logStartOffset here. Otherwise, we do this when loading the log.
+      if (!isRecoveredSwapFile)
```
hachikuji, line 1916: Can you explain why this is necessary? I understand that there is logic to initialize the log start offset after loading segments, but why do we need a special check to prevent updating the log start offset here?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2393,6 +2407,10 @@ class Log(@volatile private var _dir: File,
...
+         maybeIncrementLogStartOffset(segments.firstEntry.getValue.baseOffset, LogCompaction)
```
hachikuji: Maybe we need a separate reason for the segment splitting case?

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -600,20 +608,31 @@ private[log] class Cleaner(val id: Int,
         currentSegmentOpt = nextSegmentOpt
       }
 
-      cleaned.onBecomeInactiveSegment()
-      // flush new segment to disk before swap
-      cleaned.flush()
+      cleanedSegment match {
+        case Some(cleaned) => {
```
hachikuji: nit: braces are unnecessary. Similar for `None` case

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -600,20 +608,31 @@ private[log] class Cleaner(val id: Int,
...
+        }
+        case None => {
+          info(s"Deleting segment(s) $segments in log $log")
+          log.deleteSegments(segments, SegmentCompaction)
+        }
+      }
     } catch {
       case e: LogCleaningAbortedException =>
-        try cleaned.deleteIfExists()
+          try if (cleanedSegment.isDefined) {
```
hachikuji: nit: this alignment looks off. Also, can we do a `foreach`?

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -683,6 +702,8 @@ private[log] class Cleaner(val id: Int,
     }
 
     var position = 0
+    var destSegment = dest
+    val topicPartition = log.topicPartition
```
hachikuji: nit: does not feel like a very useful redefinition.

core/src/main/scala/kafka/log/LogCleaner.scala
hachikuji: Sorry to keep pressing this, but do you have an update here? It would be nice to keep the behavior here on par with the existing implementation. Really if you think about the way this is working, we could write to the cleaned segment directly from `filterTo` instead of building a buffer in memory, but we can leave that as a follow-up improvement.

core/src/main/scala/kafka/log/LogCleanerManager.scala
```diff
@@ -601,6 +601,9 @@ private[log] object LogCleanerManager extends Logging {
       // the active segment is always uncleanable
       Option(log.activeSegment.baseOffset),
 
+      // we do not want to clean past the high watermark
+      Option(log.highWatermark),
```
hachikuji: I agree this probably makes sense. It is surprising that it was not there already. Was there a specific reason related to this patch that you decided to do it here? An alternative by the way is to replace the high watermark and first unstable offset here with `lastStableOffset`, which would represent the lower bound of the two.

core/src/test/scala/integration/kafka/api/AbstractConsumerTest.scala
```diff
@@ -116,6 +116,18 @@ abstract class AbstractConsumerTest extends BaseRequestTest {
     records
   }
 
+  protected def sendDuplicateRecords(producer: KafkaProducer[Array[Byte], Array[Byte]], numRecords: Int,
```
hachikuji: How about renaming this to `sendRecordsWithKey` and include the key as a parameter? Otherwise it is a little strange as a general utility since it is tailored so narrowly for compaction. Also, perhaps we could consider moving this to `TestUtils`?

core/src/main/scala/kafka/log/Log.scala
jolshan: Yes. That is right. `logSegments` does the following:
```
 Get all segments beginning with the segment that includes "from" and ending with the segment
 that includes up to "to-1" or the end of the log (if to > logEndOffset).
 ```
 So that grabs [1000,2000) and [2000,3000) as you said. Both of the read next offsets of these segments are larger than 1500. So these two segments are replaced with [1500,2500)
 
 I can include this example and the simple summary.

File: core/src/main/scala/kafka/log/Log.scala (hunk 19 above)
Comment (jolshan) [reply core/src/main/scala/kafka/log/Log.scala:1916]: I suppose we could remove the check and update the offset twice in the case of loading with swap files.

File: core/src/main/scala/kafka/log/Log.scala (hunk 20 above)
Comment (jolshan) [reply core/src/main/scala/kafka/log/Log.scala]: Is this referring to the comment about excluding recovered swap files, or is there an issue that will occur with segment splitting?

File: core/src/main/scala/kafka/log/LogCleanerManager.scala (hunk 24 above)
Comment (jolshan) [reply core/src/main/scala/kafka/log/LogCleanerManager.scala]: This would cause issues during tests where if we didn't update the highWatermark and tried to compact (and update the logStartOffset) we would throw an exception since we can't update the logStartOffset higher than the HW. This made me think we also shouldn't compact beyond the HW. I'm good with using `lastStableOffset`

File: core/src/test/scala/integration/kafka/api/AbstractConsumerTest.scala (hunk 25 above)
Comment (jolshan) [reply core/src/test/scala/integration/kafka/api/AbstractConsumerTest.scala]: I also thought it was a little out of place. I'll change it.

File: core/src/main/scala/kafka/log/Log.scala (hunk 16 above)
Comment (jolshan) [reply core/src/main/scala/kafka/log/Log.scala]: I think I'm a little confused too. My understanding of this line is that we can't have offsets in the swap segment that are outside the range of the segments in log we started from. Is there a case with this new behavior that I'm missing where this isn't the case? Or is it the "existing segments" part that is confusing.

Comment (hachikuji) [reply core/src/main/scala/kafka/log/Log.scala]: Initially I interpreted the comment as suggesting a stronger statement about how each swapped segment related to the set of segments that it would replace. Perhaps it was really the comment that the "swap segments be composed of the original set" which confused me. 

I would suggest that we use more precise language. For example, can we say that all offsets in the swapped segment must be present in the segments that are to be replaced? It would also be helpful to clarify the following sentence:

> If we cannot find such a segment, it means the deletion of that segment was successful.

Really we are referring to a group of segments. How does the conclusion that deletion was successful follow from the fact that the swapped segment is "composed" of the original segments?

File: core/src/main/scala/kafka/log/Log.scala (hunk 20 above)
Comment (hachikuji) [reply core/src/main/scala/kafka/log/Log.scala]: What I am referring to is the use of the `LogCompaction` reason passed to `maybeIncrementLogStartOffset`. I believe this logic also applies when splitting a segment and that does not involve compaction.

File: core/src/main/scala/kafka/log/Log.scala (hunk 16 above)
Comment (jolshan) [reply core/src/main/scala/kafka/log/Log.scala]: Right. Ok, I see what you are saying now. I can clarify these points. 
As for the deletion. I think the idea is that the segments that will be replaced (that "compose" the swap segment) are marked for deletion after the swap segment is created. I think the idea is that if a segment containing an offset in the range of the swap segment's offsets is missing, that means that the deletion completed. Again, I can clarify this.

File: core/src/main/scala/kafka/log/Log.scala (hunk 20 above)
Comment (jolshan) [reply core/src/main/scala/kafka/log/Log.scala]: Ah. I see what you mean. I can add such a reason as well as a parameter for this method. In the case of `completeSwapOperations` I'm a little unsure how we can tell. We can check the number of swap files since cleaning will always make just 1. But I think in the case of splitting, we may also have just 1 if we fail right before handling the last swap file. This seems like a pretty rare case though.

Comment (jolshan) [reply core/src/main/scala/kafka/log/Log.scala]: Decided to rename the reason to generalize.

File: core/src/main/scala/kafka/log/Log.scala (hunk 19 above)
Comment (jolshan) [reply core/src/main/scala/kafka/log/Log.scala:1916]: Ah I remember now. The highwatermark is not updated until later so this throws an error.

Comment (jolshan) [reply core/src/main/scala/kafka/log/Log.scala:1916]: Or rather, my test remembers.

File: core/src/main/scala/kafka/log/LogCleanerManager.scala (hunk 24 above)
Comment (hachikuji) [reply core/src/main/scala/kafka/log/LogCleanerManager.scala]: In practice, we do not allow the log start offset to get ahead of the high watermark, so I am not sure that is a real issue. Still, I cannot think of a strong reason to allow cleaning above the high watermark, so maybe it is simpler to prevent it.

Comment (jolshan) [reply core/src/main/scala/kafka/log/LogCleanerManager.scala]: I switched this to `lastStableOffset` replacing unstable offset above.

File: clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java [hunk 26]
```diff
@@ -340,17 +343,18 @@ private FilterResult(ByteBuffer outputBuffer) {
         private void updateRetainedBatchMetadata(MutableRecordBatch retainedBatch, int numMessagesInBatch, boolean headerOnly) {
             int bytesRetained = headerOnly ? DefaultRecordBatch.RECORD_BATCH_OVERHEAD : retainedBatch.sizeInBytes();
             updateRetainedBatchMetadata(retainedBatch.maxTimestamp(), retainedBatch.lastOffset(),
-                    retainedBatch.lastOffset(), numMessagesInBatch, bytesRetained);
+                    retainedBatch.lastOffset(), retainedBatch.baseOffset(), numMessagesInBatch, bytesRetained);
         }
 
         private void updateRetainedBatchMetadata(long maxTimestamp, long shallowOffsetOfMaxTimestamp, long maxOffset,
-                                                int messagesRetained, int bytesRetained) {
+                                                long baseOffset, int messagesRetained, int bytesRetained) {
```
hachikuji: nit: this looks misaligned

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
```diff
@@ -340,17 +343,18 @@ private FilterResult(ByteBuffer outputBuffer) {
...
             validateBatchMetadata(maxTimestamp, shallowOffsetOfMaxTimestamp, maxOffset);
             if (maxTimestamp > this.maxTimestamp) {
                 this.maxTimestamp = maxTimestamp;
                 this.shallowOffsetOfMaxTimestamp = shallowOffsetOfMaxTimestamp;
             }
             this.maxOffset = Math.max(maxOffset, this.maxOffset);
+            if (this.bytesRetained == 0) this.baseOffsetOfFirstBatch = baseOffset;
```
hachikuji: It seems like it would be more direct to check `baseOffsetOfFirstBatch < 0`.

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
```diff
@@ -386,6 +390,13 @@ public long maxOffset() {
             return maxOffset;
         }
 
+        /**
+         * @return  the baseOffset of the first batch of retained records or -1 if no batches are retained
+         */
+        public long baseOffsetOfFirstBatch() {
```
hachikuji: Would it be worth using `OptionalLong`?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2659,3 +2672,9 @@ case object LogDeletion extends SegmentDeletionReason {
     log.info(s"Deleting segments as the log has been deleted: ${toDelete.mkString(",")}")
   }
 }
+
+case object SegmentCompaction extends SegmentDeletionReason {
+  override def logReason(log: Log, toDelete: List[LogSegment]): Unit = {
+    log.info(s"Deleting segments as all records were deleted in log compaction: ${toDelete.mkString(",")}")
```
hachikuji: nit: "deleted during log compaction"?

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -599,20 +607,27 @@ private[log] class Cleaner(val id: Int,
...
+          cleaned.lastModified = modified
 
-      // swap in new segment
-      info(s"Swapping in cleaned segment $cleaned for segment(s) $segments in log $log")
-      log.replaceSegments(List(cleaned), segments)
+          // swap in new segment
+          info(s"Swapping in cleaned segment $cleaned for segment(s) $segments in log $log")
+          log.replaceSegments(List(cleaned), segments)
+        case None =>
+          info(s"Deleting segment(s) $segments in log $log")
```
hachikuji: nit: maybe add some explanation? Maybe `Deleting segments $segments in log $log since no records were retained after cleaning`?

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -700,11 +716,18 @@ private[log] class Cleaner(val id: Int,
       // if any messages are to be retained, write them out
       val outputBuffer = result.outputBuffer
       if (outputBuffer.position() > 0) {
+        if (destSegment.isEmpty) {
+          // create a new segment with a suffix appended to the name of the log and indexes
+          destSegment = Some(LogCleaner.createNewCleanedSegment(log.dir, log.config, result.baseOffsetOfFirstBatch()))
```
hachikuji: nit: unnecessary parenthesis `baseOffsetOfFirstBatch()`

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -700,11 +716,18 @@ private[log] class Cleaner(val id: Int,
...
+          transactionMetadata.cleanedIndex = Some(destSegment.get.txnIndex)
+          transactionMetadata.appendTransactionIndex()
+        }
+        
         outputBuffer.flip()
         val retained = MemoryRecords.readableRecords(outputBuffer)
         // it's OK not to hold the Log's lock in this case, because this segment is only accessed by other threads
         // after `Log.replaceSegments` (which acquires the lock) is called
-        dest.append(largestOffset = result.maxOffset,
+        destSegment.get.append(largestOffset = result.maxOffset,
```
hachikuji: nit: we could probably do something like this to avoid the nasty `get` calls in here
```scala
val segment = destSegment.getOrElse {
  val newSegment = LogCleaner.createNewCleanedSegment(log.dir, log.config, result.baseOffsetOfFirstBatch())
  transactionMetadata.cleanedIndex = Some(newSegment.txnIndex)
  transactionMetadata.appendTransactionIndex()
  destSegment = Some(newSegment)
  newSegment
}
```

core/src/main/scala/kafka/log/LogCleaner.scala
```diff
@@ -1161,6 +1191,14 @@ private[log] class CleanedTransactionMetadata {
     }
   }
 
+  /**
+   * Apply transactions that accumulated before cleanedIndex was applied
+   */
+  def appendTransactionIndex(): Unit = {
+    toAppend.foreach(transaction => cleanedIndex.foreach(_.append(transaction)))
+    toAppend = ListBuffer.empty
```
hachikuji: Maybe we can just `clear()`?

core/src/main/scala/kafka/log/LogCleanerManager.scala
```diff
@@ -595,8 +595,8 @@ private[log] object LogCleanerManager extends Logging {
     // may be cleaned
     val firstUncleanableDirtyOffset: Long = Seq(
 
-      // we do not clean beyond the first unstable offset
-      log.firstUnstableOffset,
+      // we do not clean beyond the lastStableOffset
```
hachikuji: Maybe worth emphasizing that this also implies that we do not clean beyond the high watermark?
```scala
// we do not clean beyond the last stable offset (and therefore the high watermark)
```

core/src/test/scala/integration/kafka/api/PlaintextConsumerTest.scala
```diff
@@ -1267,6 +1266,64 @@ class PlaintextConsumerTest extends BaseConsumerTest {
...
+    val t1p1 = new TopicPartition(topic1, 1)
+    val partitions = Set(t0p0, t1p0, t1p1).asJava
+
+    val producerProps = new Properties()
+    // Each batch will hold about 10 records
+    producerProps.setProperty(ProducerConfig.BATCH_SIZE_CONFIG, "256")
+    val producer = createProducer(configOverrides = producerProps)
+    // First topic will not have compaction. Simply a sanity test.
+    createTopicAndSendRecords(producer, topicName = topic0, numPartitions = 1, recordsPerPartition = 100)
+
```
hachikuji: nit: unneeded newline

core/src/test/scala/integration/kafka/api/PlaintextConsumerTest.scala
```diff
@@ -1267,6 +1266,64 @@ class PlaintextConsumerTest extends BaseConsumerTest {
...
+    // Send records to first partition -- all duplicates.
+    createTopic(topic1, numPartitions = 2, replicationFactor = 1, props)
+    TestUtils.sendRecordsWithKey(producer, 100, 0L, new TopicPartition(topic1, 0), "key")
+
+    // Send records fo second partition -- only last 50 records are duplicates.
+    sendRecords(producer, 50, t1p1)
+    TestUtils.sendRecordsWithKey(producer, 50, 50L, new TopicPartition(topic1, 1), "key")
+
+    // Sleep to allow compaction to take place.
+    Thread.sleep(25000)
```
hachikuji: Ouch. 25s of sleep is significant. I wonder if this test is overkill given testing we have in `LogCleanerTest`?

core/src/test/scala/unit/kafka/log/LogCleanerManagerTest.scala
```diff
@@ -540,13 +540,34 @@ class LogCleanerManagerTest extends Logging {
...
+    log.updateHighWatermark(log.activeSegment.baseOffset)
 
     val lastCleanOffset = Some(0L)
     val cleanableOffsets = LogCleanerManager.cleanableOffsets(log, lastCleanOffset, time.milliseconds)
     assertEquals(0L, cleanableOffsets.firstDirtyOffset, "The first cleanable offset starts at the beginning of the log.")
     assertEquals(log.activeSegment.baseOffset, cleanableOffsets.firstUncleanableDirtyOffset, "The first uncleanable offset begins with the active segment.")
   }
 
+  @Test
+  def testCleanableOffsetsForNoneWithLowerHighWatermark(): Unit = {
```
hachikuji: nit: the name is not very clear. What does `None` refer to?

core/src/test/scala/unit/kafka/log/LogCleanerTest.scala
```diff
@@ -984,19 +1003,26 @@ class LogCleanerTest {
...
 
     val distinctValuesBySegmentAfterClean = distinctValuesBySegment
 
-    assertTrue(disctinctValuesBySegmentBeforeClean.zip(distinctValuesBySegmentAfterClean)
-      .take(numCleanableSegments).forall { case (before, after) => after < before },
+    // One segment should have been completely deleted, so there will be fewer segments.
+    assertTrue(distinctValuesBySegmentAfterClean.size < distinctValuesBySegmentBeforeClean.size)
+
+    // Drop the first segment from before cleaning since it was removed. Also subtract 1 from numCleanableSegments
+    val normalizedDistinctValuesBySegmentBeforeClean = distinctValuesBySegmentBeforeClean.drop(1)
```
hachikuji: The logic in this test case has become rather obscure after the change. Maybe we could do something simpler than comparing segment by segment. As far as I can tell, all the test is doing is ensuring that the first uncleanable offset is respected. Maybe a simpler test would just write the same key over and over and then assert that all records below the uncleanable offset are removed and all records above that offset are retained?

core/src/test/scala/integration/kafka/api/PlaintextConsumerTest.scala
jolshan: Yeah. I was wondering about this too. I originally wrote this test to describe the exact case KAFKA-7556 is describing. I'll take a look at this again and see if we can bring it down.

core/src/test/scala/unit/kafka/log/LogCleanerTest.scala
jolshan: I'll take another look at this one

clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
jolshan: In the non-test code, this is only used when we have bytes in the segment (so it will not be -1). In tests, this is sometimes called when the value is -1. I switched to OptionalLong, but we can also switch back.

It is a little awkward, since we will need to call get() and there isn't really an alternative if the optional is empty. (I guess we could throw an error)

core/src/test/scala/unit/kafka/log/LogCleanerManagerTest.scala
jolshan: I think the None may have been taken from the test above. In that test, it was referring to "no minimum compaction lag settings active". Changed name and added a comment to describe what the test is doing.

core/src/test/scala/integration/kafka/api/PlaintextConsumerTest.scala
jolshan: I was able to get it to 15 seconds consistently (50 passed tests) but once we get to 13, it starts flaking. Not sure if this is good enough to keep the test

core/src/test/scala/unit/kafka/log/LogCleanerTest.scala
jolshan: This test is a little tricky, but I've updated it. Now it only uses duplicate keys. It's a little confusing because the first uncleanable offset is not actually the point at which records below are cleaned. The segments cleaned are the full segments below the uncleanable offset (so if the segments before the uncleanable offset in this case). And even then, one record (the last record in the last segment) will be retained due to how cleaning works.

core/src/test/scala/integration/kafka/api/PlaintextConsumerTest.scala
hachikuji: I'd rather just remove it if we don't have a good way to keep it from flaking.

jolshan: Sounds good. On further inspection looks like `log.cleaner.backoff.ms` is 15 seconds, which is why this is happening. I could try to figure out how to change this property for this test, but it may be easier to remove it.

jolshan:
all 3 jdk versions are failing with something like this:
```
 Execution failed for task ':core:integrationTest'.
> Process 'Gradle Test Executor 127' finished with non-zero exit value 1
This problem might be caused by incorrect test process configuration.
```
They build locally, so I'm not sure what's up with Jenkins. Been seeing this with at least one jdk build failing in many PRs in this repo.

ijuma:
What's remaining to get this merged?

jolshan:
I think just the merge conflicts, a final lookover and an lgtm

PR #9594, opened 2020-11-13
Improve JavaDoc in KeyValueBytesStoreSupplier

fml2:
In the JavaDoc, the implemented interface was described inaccurately.
Also, the ordered list was formatted as plain text, not as html "ol".

No testing necessary since only a comment was changed.


abbccdda (approved):
LGTM, thanks for the patch!

PR #9596, opened 2020-11-15
KAFKA-10723: Fix LogManager shutdown error handling

kowshik:
The asynchronous shutdown in `LogManager` has the shortcoming that if during shutdown any of the internal futures fail, then we do not always ensure that all futures are completed before `LogManager.shutdown` returns. This is because, [this line](https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/log/LogManager.scala#L502) in the `finally` clause shuts down the thread pools asynchronously. As a result, despite the `shut down completed` message from KafkaServer is seen in the error logs, some futures continue to run from inside `LogManager` attempting to close some logs. This is misleading during debugging. Also sometimes it introduces an avoidable post-shutdown activity where resources (such as file handles) are released or persistent state is checkpointed in the Broker.

In this PR, we fix the above behavior such that we prevent leakage of threads. If any of the futures throw an error, we  skip creating of checkpoint and clean shutdown file only for the affected log directory. We continue to wait for all futures to complete for all the directories.

**Test plan:**

Added a new unit test: `LogManager.testHandlingExceptionsDuringShutdown`.

### DIFF
--- a/core/src/main/scala/kafka/log/LogManager.scala
+++ b/core/src/main/scala/kafka/log/LogManager.scala
@@ -477,27 +477,41 @@ class LogManager(logDirs: Seq[File],
       jobs(dir) = jobsForDir.map(pool.submit).toSeq
     }
 
+    var firstExceptionOpt: Option[Throwable] = Option.empty
     try {
       for ((dir, dirJobs) <- jobs) {
-        dirJobs.foreach(_.get)
+        val errorsForDirJobs = dirJobs.map {
+          future =>
+            try {
+              future.get
+              Option.empty
+            } catch {
+              case e: ExecutionException =>
+                error(s"There was an error in one of the threads during LogManager shutdown: ${e.getCause}")
+                Some(e.getCause)
+            }
+        }.filter{ e => e.isDefined }.map{ e => e.get }
 
-        val logs = logsInDir(localLogsByDir, dir)
+        if (firstExceptionOpt.isEmpty) {
+          firstExceptionOpt = errorsForDirJobs.headOption
+        }
 
-        // update the last flush point
-        debug(s"Updating recovery points at $dir")
-        checkpointRecoveryOffsetsInDir(dir, logs)
+        if (errorsForDirJobs.isEmpty) {
+          val logs = logsInDir(localLogsByDir, dir)
 
-        debug(s"Updating log start offsets at $dir")
-        checkpointLogStartOffsetsInDir(dir, logs)
+          // update the last flush point
+          debug(s"Updating recovery points at $dir")
+          checkpointRecoveryOffsetsInDir(dir, logs)
 
-        // mark that the shutdown was clean by creating marker file
-        debug(s"Writing clean shutdown marker at $dir")
-        CoreUtils.swallow(Files.createFile(new File(dir, Log.CleanShutdownFile).toPath), this)
+          debug(s"Updating log start offsets at $dir")
+          checkpointLogStartOffsetsInDir(dir, logs)
+
+          // mark that the shutdown was clean by creating marker file
+          debug(s"Writing clean shutdown marker at $dir")
+          CoreUtils.swallow(Files.createFile(new File(dir, Log.CleanShutdownFile).toPath), this)
+        }
       }
-    } catch {
-      case e: ExecutionException =>
-        error(s"There was an error in one of the threads during LogManager shutdown: ${e.getCause}")
-        throw e.getCause
+      firstExceptionOpt.foreach{ e => throw e}
     } finally {
       threadPools.foreach(_.shutdown())
       // regardless of whether the close succeeded, we need to unlock the data directories
--- a/core/src/main/scala/kafka/log/LogManager.scala
+++ b/core/src/main/scala/kafka/log/LogManager.scala
@@ -477,26 +477,21 @@ class LogManager(logDirs: Seq[File],
       jobs(dir) = jobsForDir.map(pool.submit).toSeq
     }
 
-    var firstExceptionOpt: Option[Throwable] = Option.empty
     try {
       for ((dir, dirJobs) <- jobs) {
-        val errorsForDirJobs = dirJobs.map {
+        val hasErrors = dirJobs.exists {
           future =>
             try {
               future.get
-              Option.empty
+              false
             } catch {
               case e: ExecutionException =>
-                error(s"There was an error in one of the threads during LogManager shutdown: ${e.getCause}")
-                Some(e.getCause)
+                warn(s"There was an error in one of the threads during LogManager shutdown: ${e.getCause}")
+                true
             }
-        }.filter{ e => e.isDefined }.map{ e => e.get }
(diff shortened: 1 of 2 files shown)


kowshik:
cc @dhruvilshah3 @junrao @hachikuji for review

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
```diff
@@ -83,6 +87,51 @@ class LogManagerTest {
...
+    log1.appendAsLeader(TestUtils.singletonRecords("test1".getBytes()), leaderEpoch = 0)
+    log1.takeProducerSnapshot()
+    log1.appendAsLeader(TestUtils.singletonRecords("test1".getBytes()), leaderEpoch = 0)
+
+    log2.appendAsLeader(TestUtils.singletonRecords("test2".getBytes()), leaderEpoch = 0)
+    log2.takeProducerSnapshot()
+    log2.appendAsLeader(TestUtils.singletonRecords("test2".getBytes()), leaderEpoch = 0)
+
+    // This should cause log1.close() to fail during LogManger shutdown sequence.
+    FileUtils.deleteDirectory(logFile1)
```
lqjack: If the end user delete the log files Manually , the server cannot be stopped. and The cannot startup it again? so in this case ,how do they resolve it ?

kowshik: Sorry I do not understand the question.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -477,27 +477,41 @@ class LogManager(logDirs: Seq[File],
...
+          // mark that the shutdown was clean by creating marker file
+          debug(s"Writing clean shutdown marker at $dir")
+          CoreUtils.swallow(Files.createFile(new File(dir, Log.CleanShutdownFile).toPath), this)
+        }
       }
-    } catch {
-      case e: ExecutionException =>
-        error(s"There was an error in one of the threads during LogManager shutdown: ${e.getCause}")
-        throw e.getCause
+      firstExceptionOpt.foreach{ e => throw e}
```
junrao: Hmm, since we are about to shut down the JVM, should we just log a WARN here instead of throwing the exception?

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
```diff
@@ -83,6 +87,51 @@ class LogManagerTest {
     log.appendAsLeader(TestUtils.singletonRecords("test".getBytes()), leaderEpoch = 0)
   }
 
+  /**
+   * Tests that all internal futures are completed before LogManager.shutdown() returns to the
+   * caller during error situations.
+   */
+  @Test
+  def testHandlingExceptionsDuringShutdown(): Unit = {
+    logManager.shutdown()
```
junrao: Hmm, do we need this given that we do this in tearDown() already?

junrao (commented):
@kowshik : Thanks for the PR. A couple of comments below.

core/src/main/scala/kafka/log/LogManager.scala
kowshik: Great point. I've changed the code to do the same.
My understanding is that the exception swallow safety net exists inside `KafkaServer.shutdown()` today, but it makes sense to also just log a warning here instead instead of relying on the safety net: https://github.com/apache/kafka/blob/bb34c5c8cc32d1b769a34329e34b83cda040aafc/core/src/main/scala/kafka/server/KafkaServer.scala#L732.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
kowshik: Yeah this explicit shutdown is needed to:
1) Re-create a new `LogManager` instance with multiple `logDirs` for this test. This is different from the default one provided in `setUp()`.
2) Help do some additional checks post shutdown (towards the end of this test).

kowshik: Thinking about it again, you are right. I have eliminated the need for the `shutdown()` now by using a `LogManager` instance specific to the test.

kowshik:
Thanks for the review @junrao! I have addressed the comments in f917f0c24cebbb0fb5eb7029ccb6676734b60b3e.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
lqjack: What if error occur during the shutdown of the broker ?  should we log the error info to the log or just throw the exception ?

kowshik: It depends on the kind of error, but we do log the error information to the log today from within `KafkaServer.shutdown()`.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -479,25 +479,33 @@ class LogManager(logDirs: Seq[File],
 
     try {
       for ((dir, dirJobs) <- jobs) {
-        dirJobs.foreach(_.get)
+        val hasErrors = dirJobs.exists {
+          future =>
```
ijuma: Nit: this should be in the previous line.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -479,25 +479,33 @@ class LogManager(logDirs: Seq[File],
 
     try {
       for ((dir, dirJobs) <- jobs) {
-        dirJobs.foreach(_.get)
+        val hasErrors = dirJobs.exists {
```
ijuma: This looks wrong. `exists` short-circuits. I think you want `map` followed by `exists`.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -479,25 +479,33 @@ class LogManager(logDirs: Seq[File],
...
+          future =>
+            try {
```
ijuma: You can use `scala.util.Try` to wrap the call and get a `Success` or `Failure`.

core/src/main/scala/kafka/log/LogManager.scala
kowshik: Thats a really good point. Done.

core/src/main/scala/kafka/log/LogManager.scala
kowshik: Done.

core/src/main/scala/kafka/log/LogManager.scala
kowshik: Good idea, done.

kowshik:
Thanks for the review @ijuma ! I have addressed the comments in 8716429b48cad8af6ad73109c1d9f7442823c02f .

junrao (approved):
@kowshik : Thanks for the latest PR. LGTM

PR #9598, opened 2020-11-16
KAFKA-10701 : First line of detailed stats from consumer-perf-test.sh incorrect

lijubjohn:
Corrected the initialization of joinStart variable to fix the first line of the consumer performance stats. Since the joinStart was initialized with 0 , so when the first time onPartitionAssigned method is called it makes the joinTime as System.currentTimeMillis which inturn made the fetchTimeMs incorrect (-ve)

### DIFF
--- a/core/src/main/scala/kafka/tools/ConsumerPerformance.scala
+++ b/core/src/main/scala/kafka/tools/ConsumerPerformance.scala
@@ -106,7 +106,7 @@ object ConsumerPerformance extends LazyLogging {
     var messagesRead = 0L
     var lastBytesRead = 0L
     var lastMessagesRead = 0L
-    var joinStart = 0L
+    var joinStart = System.currentTimeMillis
     var joinTimeMsInSingleRound = 0L
 
     consumer.subscribe(topics.asJava, new ConsumerRebalanceListener {


mumrah (commented):
Thanks for the patch, @lijubjohn! Can you see how difficult it would be to add or modify a test to capture this fix?

lijubjohn:
@mumrah Thanks for reviewing the pr. 
The return type of the method which is modified is void and the purpose of the method is to print the performance stats to console , so there isn't any direct hook to test this method

quanuw:
Hi @lijubjohn, can you explain how having joinStart initialized to 0L led to a negative fetchTimeMs? I'm new to the project and not sure how having joinStart initialized as 0L could have been the problem.

mumrah (approved):
LGTM

mumrah:
@quanuw there are some statements that expect `joinStart` to be a timestamp in order to calculate durations. When it's zero, those durations become huge which leads to subsequent calculations being incorrect. 

You should be able to run the consumer perf test with `KAFKA_DEBUG=1` flag and attach a debugger from IntelliJ or Eclipse. This will allow more insight as to what's going on.

PR #9599, opened 2020-11-16
MINOR: Include connector name in error message

C0urante:
These log messages aren't triggered very frequently, but when they are it can indicate a serious problem with the connector, and it'd be nice to know exactly which connector is having that problem without having to dig through other log messages and try to correlate the stack trace here with existing connector configs.


rhauch (approved):
Looks useful. Thanks!

chia7712 (approved):
make sense to me. +1

chia7712:
@C0urante Thanks for your patch. Merge to trunk!

C0urante:
Thanks @chia7712!

PR #9601, opened 2020-11-16
 KAFKA-10729: Bump remaining RPC's to use tagged fields.

gardnervickers:
As a follow-up from [KIP-482](https://cwiki.apache.org/confluence/display/KAFKA/KIP-482%3A+The+Kafka+Protocol+should+Support+Optional+Tagged+Fields), this PR bumps the version for several
RPC's to enable tagged fields via the flexible versioning mechanism.

Additionally, a new IBP version `KAFKA_2_8_IV0` is introduced to
allow replication to take advantage of these new RPC versions for
OffsetForLeaderEpoch and ListOffset.


core/src/main/scala/kafka/api/ApiVersion.scala
```diff
@@ -424,6 +426,13 @@ case object KAFKA_2_7_IV2 extends DefaultApiVersion {
   val id: Int = 30
 }
 
+case object KAFKA_2_7_IV3 extends DefaultApiVersion {
```
ijuma: 2.7 has been branched. It should be 2.8, right?

gardnervickers: Thanks, that makes more sense.

gardnervickers:
I realize now that we have several requests which have not been converted to use the generated RPC's. It appears we need to do that to support tagged fields.

gardnervickers:
OffsetForLeaderEpoch and Produce are not yet generated RPCs, but will be once #9401 and #9547 are merged. I've removed the taggedFields bump for these RPC's. We can bump them once their respective PR's are merged.

dajac:
@gardnervickers https://github.com/apache/kafka/pull/9401 and https://github.com/apache/kafka/pull/9547 have been merged. You can bring them back in this PR if you like.

core/src/main/scala/kafka/api/ApiVersion.scala
```diff
@@ -108,7 +108,9 @@ object ApiVersion {
     // Bup Fetch protocol for Raft protocol (KIP-595)
     KAFKA_2_7_IV1,
     // Introduced AlterIsr (KIP-497)
-    KAFKA_2_7_IV2
+    KAFKA_2_7_IV2,
+    // Flexible versioning on ListOffsets
```
hachikuji: For what it's worth, `WriteTxnMarkers` and `OffsetsForLeaderEpoch` are also inter-broker APIs.

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
```diff
@@ -613,12 +613,12 @@ private void sendResponse(ResponseHeader respHeader, Struct response) {
         selector.completeReceive(new NetworkReceive(node.idString(), buffer));
     }
 
-    private void sendThrottledProduceResponse(int correlationId, int throttleMs) {
+    private void sendThrottledProduceResponse(int correlationId, int throttleMs, short version) {
```
gardnervickers, line 616: This is to fix `testThrottlingNotEnabledForConnectionToOlderBroker`. The test forces ApiVersionsResponse to version 5, but relied on the fact that nothing really changed between version 5 and version 8 for PRODUCE responses. With flex versions we need to ensure the response matches the ApiVersions response

clients/src/main/java/org/apache/kafka/common/requests/AlterReplicaLogDirsResponse.java
```diff
@@ -38,10 +38,6 @@
 
     private final AlterReplicaLogDirsResponseData data;
 
-    public AlterReplicaLogDirsResponse(Struct struct) {
```
gardnervickers, line 41: Having this override seemed a bit error prone, and was causing failures for the `NetworkClientTest`. I opted to remove it entirely in favor of forcing the caller to specify the response version.

core/src/test/scala/unit/kafka/server/ReplicaFetcherThreadTest.scala
```diff
@@ -437,8 +437,8 @@ class ReplicaFetcherThreadTest {
     thread.doWork()
     assertEquals(2, mockNetwork.epochFetchCount)
     assertEquals(1, mockNetwork.fetchCount)
-    assertEquals("OffsetsForLeaderEpochRequest version.",
-      3, mockNetwork.lastUsedOffsetForLeaderEpochVersion)
+    assertTrue("OffsetsForLeaderEpochRequest version.",
+      mockNetwork.lastUsedOffsetForLeaderEpochVersion >= 3)
```
gardnervickers, line 441: Admittedly I'm not sure why we check the version of the last `OffsetForLeaderEpoch` response is 3 here. I decided to widen the check a bit so this won't break for future versions.

clients/src/main/java/org/apache/kafka/common/requests/ProduceRequest.java
```diff
@@ -286,6 +286,7 @@ public static byte requiredMagicForVersion(short produceRequestVersion) {
             case 6:
             case 7:
             case 8:
+            case 9:
```
hachikuji: I wonder if we may as well make this the default case. Not sure we're getting much by forcing ourselves to update this logic after each bump. Maybe the range validation is still useful, but that could be done by using `oldestVersion` and `latestVersion`.

clients/src/main/java/org/apache/kafka/common/requests/WriteTxnMarkersRequest.java
```diff
@@ -105,8 +105,8 @@ public int hashCode() {
 
         public final WriteTxnMarkersRequestData data;
 
-        public Builder(final List<TxnMarkerEntry> markers) {
-            super(ApiKeys.WRITE_TXN_MARKERS);
+        public Builder(final List<TxnMarkerEntry> markers, short latestAllowedVersion) {
+            super(ApiKeys.WRITE_TXN_MARKERS, ApiKeys.WRITE_TXN_MARKERS.oldestVersion(), latestAllowedVersion);
```
hachikuji: I think this is probably ok, but it is a little inconsistent with how we handle the versions for other inter-broker RPCs. Since we rely on the IBP, we always set the version explicitly in the caller, which means there is exactly one allowable version for the builder to use. See for example `LeaderAndIsrRequest.Builder`.

core/src/test/scala/unit/kafka/server/EdgeCaseRequestTest.scala
```diff
@@ -84,19 +85,24 @@ class EdgeCaseRequestTest extends KafkaServerTestHarness {
...
+    // Check for flex versions, some tests here verify that an invalid apiKey is detected properly, so if -1 is used,
+    // assume the request is not using flex versions.
+    val flexVersion = if (apiKey >= 0) ApiKeys.forId(apiKey).requestHeaderVersion(apiVersion) >= 2 else false
     val size = {
       2 /* apiKey */ +
         2 /* version id */ +
         4 /* correlation id */ +
-        Type.NULLABLE_STRING.sizeOf(clientId) /* client id */
+        Type.NULLABLE_STRING.sizeOf(clientId)  /* client id */ +
+        (if (flexVersion) ByteUtils.sizeOfUnsignedVarint(0) else 0)
```
hachikuji: nit: maybe add a comment that this field is for the number of tagged fields?

hachikuji (approved):
LGTM. Thanks for the patch!

core/src/test/scala/unit/kafka/server/ReplicaFetcherThreadTest.scala
hachikuji: I think this was originally using `1` in order to ensure that we were using a version which included the epoch in the response. Since then it looks like it has been updated blindly every time we've bumped the protocol. I'm ok leaving this as is, but we could probably also get rid of it.

PR #9602, opened 2020-11-17
MINOR: Use string interpolation in FinalizedFeatureCache

kowshik:
This is a small change. In this PR, I'm using string interpolation in `FinalizedFeatureCache` at places where string format was otherwise used. This just ensures uniformity, with this change we ensure that throughout the file we just use string interpolation.

**Test plan:**

Rely on existing tests since this PR is not changing behavior.


dajac (approved):
LGTM, thanks for the patch.

dajac:
Failed test is unrelated. Merging to trunk.

kowshik:
@dajac thanks for the review :)

PR #9603, opened 2020-11-17
MINOR: Initialize ConnectorConfig constructor with emptyMap and avoid instantiating a new Map

UnityLung:
It's a better way to return an empty Map.
Use Collections.emptyMap() instead HashMap


UnityLung:
@chia7712  Could you review it please

kkonstantine (approved):
Passed map is not mutated, indeed. 

LGTM

UnityLung:
@kkonstantine  
Thanks a lot.

PR #9604, opened 2020-11-17
[MINOR] adjust the log level to error

lisa2lisa:
we recently encounter this log which should be more of a warn, or error message. Even though it checks acks=0, it still should bring people's attention on
```
[2020-11-17 01:05:17,005] INFO [KafkaApi-2] Closing connection due to error during produce request with correlation id 1159730542 from client id -0 with ack=0
Topic and partition to exceptions:some-topic-name-30 -> org.apache.kafka.common.errors.RecordTooLargeException (kafka.server.KafkaApis)
```

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaApis.scala
+++ b/core/src/main/scala/kafka/server/KafkaApis.scala
@@ -477,7 +477,7 @@ class KafkaApis(val requestChannel: RequestChannel,
           val exceptionsSummary = mergedResponseStatus.map { case (topicPartition, status) =>
             topicPartition -> status.error.exceptionName
           }.mkString(", ")
-          info(
+          error(
             s"Closing connection due to error during produce request with correlation id ${request.header.correlationId} " +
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -477,7 +477,7 @@ class KafkaApis(val requestChannel: RequestChannel,
           val exceptionsSummary = mergedResponseStatus.map { case (topicPartition, status) =>
             topicPartition -> status.error.exceptionName
           }.mkString(", ")
-          info(
+          error(
```
chia7712, line 480: Is this message duplicate to https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/KafkaApis.scala#L613 ? If so, is it better to remove one of them?

chia7712 (commented):
@lisa2lisa Thanks for your patch. one minor question. please take a look :)

PR #9605, opened 2020-11-17
KAFKA-10727; Handle Kerberos error during re-login as transient failure in clients

rajinisivaram:
We use a background thread for Kerberos to perform re-login before tickets expire. The thread performs logout() followed by login(), relying on the Java library to clear and then populate credentials in Subject. This leaves a timing window where clients fail to authenticate because credentials are not available. We cannot introduce any form of locking since authentication is performed on the network thread. So this PR treats NO_CRED as a transient failure rather than a fatal authentication exception in clients.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/network/SaslChannelBuilder.java
+++ b/clients/src/main/java/org/apache/kafka/common/network/SaslChannelBuilder.java
@@ -327,7 +327,7 @@ public class SaslChannelBuilder implements ChannelBuilder, ListenerReconfigurabl
         }
     }
 
-    private Class<? extends Login> defaultLoginClass() {
+    protected Class<? extends Login> defaultLoginClass() {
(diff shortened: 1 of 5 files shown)


rondagostino (commented):
Hi Rajini.  LGTM.  I was wondering if we had a similar possibility of currently-non-retriable failure with OAuth Bearer tokens, but it appears that we support multiple simultaneous tokens (see ExpiringCredentialRefreshingLogin; and OAuthBearerLoginModule sets loginRefreshReloginAllowedBeforeLogout to true) -- so what happens there is a new token is retrieved/logged-in before the first one is logged-out, and there is never a moment without valid credentials.

omkreddy (approved):

rajinisivaram:
@omkreddy  @rondagostino Thanks for the reviews. 
@rondagostino Yes, I had checked the OAuth code path to see if it had the same issue and found that you had already taken care of that :-)

Quota test failure not related, merging to trunk.

PR #9606, opened 2020-11-17
[KAFKA-10722] doc: Improve JavaDoc for KGroupedStream.aggregate and other similar methods

fml2:
Tell that the store used internally is always a timestamped one.

This is related to KAFKA-10722.

No tests are necessary because only JavaDoc was changed.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/KGroupedStream.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/KGroupedStream.java
@@ -55,7 +55,8 @@ public interface KGroupedStream<K, V> {
      * {@link StreamsConfig#CACHE_MAX_BYTES_BUFFERING_CONFIG cache size}, and
      * {@link StreamsConfig#COMMIT_INTERVAL_MS_CONFIG commit intervall}.
      * <p>
-     * For failure and recovery the store will be backed by an internal changelog topic that will be created in Kafka.
+     * For failure and recovery the store (which always will be of type {@link TimestampedKeyValueStore}) will be backed by
+     * an internal changelog topic that will be created in Kafka.
      * The changelog topic will be named "${applicationId}-${internalStoreName}-changelog", where "applicationId" is
      * user-specified in {@link StreamsConfig} via parameter
      * {@link StreamsConfig#APPLICATION_ID_CONFIG APPLICATION_ID_CONFIG}, "internalStoreName" is an internal name
@@ -82,7 +83,8 @@ public interface KGroupedStream<K, V> {
      * {@link StreamsConfig#CACHE_MAX_BYTES_BUFFERING_CONFIG cache size}, and
      * {@link StreamsConfig#COMMIT_INTERVAL_MS_CONFIG commit intervall}.
      * <p>
-     * For failure and recovery the store will be backed by an internal changelog topic that will be created in Kafka.
+     * For failure and recovery the store (which always will be of type {@link TimestampedKeyValueStore}) will be backed by
+     * an internal changelog topic that will be created in Kafka.
      * The changelog topic will be named "${applicationId}-${internalStoreName}-changelog", where "applicationId" is
      * user-specified in {@link StreamsConfig} via parameter
      * {@link StreamsConfig#APPLICATION_ID_CONFIG APPLICATION_ID_CONFIG}, "internalStoreName" is an internal name
@@ -125,7 +127,8 @@ public interface KGroupedStream<K, V> {
      * query the value of the key on a parallel running instance of your Kafka Streams application.
      *
      * <p>
-     * For failure and recovery the store will be backed by an internal changelog topic that will be created in Kafka.
+     * For failure and recovery the store (which always will be of type {@link TimestampedKeyValueStore} -- regardless of what
+     * is specified in the parameter {@code materialized}) will be backed by an internal changelog topic that will be created in Kafka.
      * Therefore, the store name defined by the Materialized instance must be a valid Kafka topic name and cannot contain characters other than ASCII
      * alphanumerics, '.', '_' and '-'.
      * The changelog topic will be named "${applicationId}-${storeName}-changelog", where "applicationId" is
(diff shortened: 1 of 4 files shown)


streams/src/main/java/org/apache/kafka/streams/kstream/KGroupedStream.java
```diff
@@ -381,7 +381,8 @@
      * {@link StreamsConfig#COMMIT_INTERVAL_MS_CONFIG commit intervall}.
      *
      * <p>
-     * For failure and recovery the store will be backed by an internal changelog topic that will be created in Kafka.
+     * For failure and recovery the store (which always will be of type {@link TimestampedKeyValueStore}) will be backed by
+     * an internal changelog topic that will be created in Kafka.
```
mjsax, line 392: Should apply the same improvement to `reduce()` and `count()` overloads? Also for `CogroupedKStream#aggregate()`?

What about `TimeWindowedKStream` and `TimeWindowedCogroupedKStream` ?

Also `StreamsBuilder#table()` (and `#globalTable()`) might need an update?

streams/src/main/java/org/apache/kafka/streams/kstream/KGroupedStream.java
```diff
@@ -438,7 +439,8 @@
      * query the value of the key on a parallel running instance of your Kafka Streams application.
      *
      * <p>
-     * For failure and recovery the store will be backed by an internal changelog topic that will be created in Kafka.
+     * For failure and recovery the store (which always will be of type {@link TimestampedKeyValueStore} -- regardless of what
+     * is specified in the parameter {@materialized}) will be backed by an internal changelog topic that will be created in Kafka.
```
mjsax: `{@materialized}` is not valid markup as far as I know. Should we `{@code materialized}`? (same below)

mjsax (commented):
Thanks for the PR. A few comments.

streams/src/main/java/org/apache/kafka/streams/kstream/KGroupedStream.java
fml2: Yes, this should be probably done. But since I don't know how they work and what to write there, I'd prefer to do it in another PR (or someone else should do it).

streams/src/main/java/org/apache/kafka/streams/kstream/KGroupedStream.java
fml2: Yes, actually, I wanted to write this but somehow... I will correct it. Shoudl I then squash the commits?

fml2: Done

fml2:
???

streams/src/main/java/org/apache/kafka/streams/kstream/KGroupedStream.java
mjsax: `reduce()` and `count()` are similar to `aggregate()`: reduce only does not allow you to change the value type (ie, output type == input type) and count, well implements an aggregate() that counts :)

And they all work the same for `KGroupedStream`, `CogroupedKStream`, `TimeWindowedKStream` and `TimeWindowedCoGroupedKStream`.

`table()` and `globalTable()` just read a topic and upsert the data into a state store / table.

So it would be great to do them all in one PR?

streams/src/main/java/org/apache/kafka/streams/kstream/KGroupedStream.java
mjsax: No need to squash commits. We squash them automatically when merging.

streams/src/main/java/org/apache/kafka/streams/kstream/KGroupedStream.java
fml2: I'd prefer to commit this chages and create an issue referring to this PR. What do you think?

mjsax: We can still do multiple PRs -- feel free to open them in parallel reusing the existing JIRA ticket.

Will hold off to merge this because if we can detect other things we need to change in the other PRs, we can keep the applied changes in-sync more easily.

fml2:
I've made the changes (rather mechanically) as requested.

streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedCogroupedKStream.java
```diff
@@ -77,7 +77,8 @@
      * {@link StreamsConfig#CACHE_MAX_BYTES_BUFFERING_CONFIG cache size}, and
      * {@link StreamsConfig#COMMIT_INTERVAL_MS_CONFIG commit intervall}.
      * <p>
-     * For failure and recovery the store will be backed by an internal changelog topic that will be created in Kafka.
+     * For failure and recovery the store (which always will be of type {@link TimestampedKeyValueStore}) will be backed by
```
mjsax: For time-windows, it would be a `TimestampedWindowStore` (not tkv-store). (same below for the other method of this class)

Note the signature of `Materialized<K, V, WindowStore<Bytes, byte[]>>` that uses `WindowStore`, not `KeyValueStore`.

streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedKStream.java
```diff
@@ -67,7 +67,8 @@
      * {@link StreamsConfig#CACHE_MAX_BYTES_BUFFERING_CONFIG cache size}, and
      * {@link StreamsConfig#COMMIT_INTERVAL_MS_CONFIG commit intervall}.
      * <p>
-     * For failure and recovery the store will be backed by an internal changelog topic that will be created in Kafka.
+     * For failure and recovery the store (which always will be of type {@link TimestampedKeyValueStore}) will be backed by
```
mjsax: As above. Should we `TimestampedWindowStore` for this class.

mjsax (commented):
Thanks a lot for updating the PR! I understand that it's rather mechanical, but that is (unfortunately) how JavaDocs work and it is quite some effort to maintain them correctly!

One more comment that needs to be addressed.

streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedCogroupedKStream.java
fml2: Updated

streams/src/main/java/org/apache/kafka/streams/kstream/TimeWindowedKStream.java
fml2: Updated

mjsax:
Thanks for updating the PR. Merged to `trunk`.

PR #9608, opened 2020-11-18
MINOR: Enable testLogCleanerStats

mattwong949:
The `testLogCleanerStats` test in `LogCleanerTest.scala` was not enabled but it was implemented. This PR adds the `@Test` annotation, and also gives it a larger map to allow the test to pass as intended.

### DIFF
--- a/core/src/test/scala/unit/kafka/log/LogCleanerTest.scala
+++ b/core/src/test/scala/unit/kafka/log/LogCleanerTest.scala
@@ -1450,9 +1450,9 @@ class LogCleanerTest {
   @Test
   def testBuildPartialOffsetMap(): Unit = {
     // because loadFactor is 0.75, this means we can fit 2 messages in the map
-    val map = new FakeOffsetMap(3)
     val log = makeLog()
-    val cleaner = makeCleaner(2)
+    val cleaner = makeCleaner(3)
+    val map = cleaner.offsetMap
 
     log.appendAsLeader(record(0,0), leaderEpoch = 0)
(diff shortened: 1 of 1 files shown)


hachikuji (approved):
Nice find. LGTM

core/src/test/scala/unit/kafka/log/LogCleanerTest.scala
```diff
@@ -815,9 +815,10 @@ class LogCleanerTest {
                (0 until leo.toInt by 2).forall(!keys.contains(_)))
   }
 
+  @Test
   def testLogCleanerStats(): Unit = {
-    // because loadFactor is 0.75, this means we can fit 2 messages in the map
-    val cleaner = makeCleaner(2)
+    // because loadFactor is 0.75, this means we can fit 3 messages in the map
```
hachikuji, line 820: Sorry, but we have several identical comments in other test cases. Are those comments also wrong?

mattwong949: I took a look at the `testPartialSegmentClean` which also uses `makeCleaner(2)`, and noticed it does multiple clean attempts since it keeps filling its map, so the logic of the tests are ok. I think the comments are misleading, I traced the code to `LogCleaner.buildOffsetMapForSegment` , and there's this line:
```
val maxDesiredMapSize = (map.slots * this.dupBufferLoadFactor).toInt
```
So we will only be able to put one offset in the map, and won't attempt to put anything else after that one even if it's the same key. I am going to change the comments

hachikuji (approved):

PR #9609, opened 2020-11-18
KAFKA-6687: restrict DSL to allow only Streams from the same source topics

ableegoldman:
Followup to https://github.com/apache/kafka/pull/9582

Will leave the ability to create multiple KTables from the same source topic as followup work. Similarly, creating a KStream and a KTable from the same topic can be tackled later if need be

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedKStreamImpl.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedKStreamImpl.java
@@ -129,7 +129,7 @@ public class CogroupedKStreamImpl<K, VOut> extends AbstractStream<K, VOut> imple
             subTopologySourceNodes,
             name,
             aggregateBuilder,
-                                                         graphNode,
+            graphNode,
             groupPatterns);
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedKStreamImpl.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedKStreamImpl.java
@@ -30,7 +30,7 @@ import org.apache.kafka.streams.kstream.SlidingWindows;
 import org.apache.kafka.streams.kstream.TimeWindowedCogroupedKStream;
 import org.apache.kafka.streams.kstream.Window;
 import org.apache.kafka.streams.kstream.Windows;
-import org.apache.kafka.streams.kstream.internals.graph.StreamsGraphNode;
+import org.apache.kafka.streams.kstream.internals.graph.GraphNode;
 import org.apache.kafka.streams.state.KeyValueStore;
 
 import java.util.LinkedHashMap;
@@ -48,9 +48,9 @@ public class CogroupedKStreamImpl<K, VOut> extends AbstractStream<K, VOut> imple
 
     CogroupedKStreamImpl(final String name,
                          final Set<String> subTopologySourceNodes,
-                         final StreamsGraphNode streamsGraphNode,
+                         final GraphNode graphNode,
                          final InternalStreamsBuilder builder) {
-        super(name, null, null, subTopologySourceNodes, streamsGraphNode, builder);
+        super(name, null, null, subTopologySourceNodes, graphNode, builder);
         groupPatterns = new LinkedHashMap<>();
         aggregateBuilder = new CogroupedStreamAggregateBuilder<>(builder);
     }
@@ -104,7 +104,7 @@ public class CogroupedKStreamImpl<K, VOut> extends AbstractStream<K, VOut> imple
             subTopologySourceNodes,
             name,
             aggregateBuilder,
-            streamsGraphNode,
+            graphNode,
             groupPatterns);
     }
 
@@ -117,7 +117,7 @@ public class CogroupedKStreamImpl<K, VOut> extends AbstractStream<K, VOut> imple
             subTopologySourceNodes,
             name,
             aggregateBuilder,
-            streamsGraphNode,
+            graphNode,
             groupPatterns);
     }
 
@@ -129,7 +129,7 @@ public class CogroupedKStreamImpl<K, VOut> extends AbstractStream<K, VOut> imple
             subTopologySourceNodes,
             name,
             aggregateBuilder,
-            streamsGraphNode,
+                                                         graphNode,
             groupPatterns);
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.streams.kstream.internals.graph;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Pattern;
+import org.apache.kafka.common.serialization.Serde;
+import org.apache.kafka.streams.kstream.internals.ConsumedInternal;
+
+abstract public class SourceGraphNode<K, V> extends StreamsGraphNode {
+
+    private Collection<String> topicNames;
+    private Pattern topicPattern;
+    private final ConsumedInternal<K, V> consumedInternal;
+
+    public SourceGraphNode(final String nodeName,
+                            final Collection<String> topicNames,
+                            final ConsumedInternal<K, V> consumedInternal) {
+        super(nodeName);
+
+        this.topicNames = topicNames;
+        this.consumedInternal = consumedInternal;
+    }
+
+    public SourceGraphNode(final String nodeName,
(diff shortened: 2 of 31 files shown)


streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.kafka.streams.kstream.internals.ConsumedInternal;
+
+abstract public class SourceGraphNode<K, V> extends StreamsGraphNode {
+
+    private Collection<String> topicNames;
+    private Pattern topicPattern;
+    private final ConsumedInternal<K, V> consumedInternal;
+
+    public SourceGraphNode(final String nodeName,
+                            final Collection<String> topicNames,
```
mjsax: nit: fix indention (same below in other constructor)

streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Pattern;
+import org.apache.kafka.common.serialization.Serde;
+import org.apache.kafka.streams.kstream.internals.ConsumedInternal;
+
+abstract public class SourceGraphNode<K, V> extends StreamsGraphNode {
+
+    private Collection<String> topicNames;
```
mjsax: Can me make those field `final`?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
```diff
@@ -0,0 +1,71 @@
...
+    private Pattern topicPattern;
```
mjsax: `final`?

mjsax (commented):
Just some nits. As I did not review the original PR, might be good to get a review from @cadonna and/or @lct45 too.

I think I understand the PR: changing the class hierarchy ensures that only `KStream` source nodes are merged?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
```diff
@@ -0,0 +1,71 @@
...
+                            final ConsumedInternal<K, V> consumedInternal) {
+
+        super(nodeName);
+
+        this.topicPattern = topicPattern;
+        this.consumedInternal = consumedInternal;
+    }
+
+    public Set<String> topicNames() {
+        return new HashSet<>(topicNames);
```
cadonna: Should we already demand a set of topics in the constructors of  `SourceGraphNode()` and its children?

streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
```diff
@@ -0,0 +1,71 @@
...
+package org.apache.kafka.streams.kstream.internals.graph;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Pattern;
+import org.apache.kafka.common.serialization.Serde;
+import org.apache.kafka.streams.kstream.internals.ConsumedInternal;
+
+abstract public class SourceGraphNode<K, V> extends StreamsGraphNode {
```
cadonna: Can we also rename `StreamsGraphNode` to `GraphNode`? The `Streams` prefix is a bit confusing, IMO, because `StreamSourceNode` and `StreamsGraphNode` seem really similar although they are quite different.

cadonna (commented):
@ableegoldman Thanks for the PR!

I have some minor comments, in addition to @mjsax comments.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
ableegoldman: I'm not sure I understand exactly what you're asking, but I made a few changes to this topic collection/method. Please lmk if it hasn't addressed your question

streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
ableegoldman: Ok, but don't come crying when the PR blows up in length 😉 (but yeah that makes sense to me)

ableegoldman:
@mjsax @cadonna I addressed your feedback, lmk if there's anything else (to answer your question, Matthias, this fixes the problem because we use `instanceof StreamsSourceNode` to determine which nodes to merge. Reclassifying things makes this pass over the table nodes, and imo makes the class hierarchy easier to understand anyway

ableegoldman:
Also just FYI since the suggested renaming touched on a lot of files, the only logical changes are to the SourceGraphNode,  StreamSourceNode, and StreamsBuilderTest classes

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedKStreamImpl.java
```diff
@@ -129,7 +129,7 @@
             subTopologySourceNodes,
             name,
             aggregateBuilder,
-            streamsGraphNode,
+                                                         graphNode,
```
mjsax: nit: indention

mjsax:
I don't want to spoil the party, but as this is a bug-fix PR, that we want to back-port, we should not to massive renaming... Happy, do rename in follow up for `trunk` (and 2.7). -- Or do a new PR without the renaming for 2.6.

mjsax (approved):
Overall LGTM. I think it's ok to have the renaming in `trunk` (and maybe `2.7`), but I have doubts to cherry-pick ii to `2.6`...

From a bug-fix/cherry-picking perspective, it might be best to undo the renaming, and cherry-pick the fix to 2.7 and 2.6 such that we can track the fix via git easily. And do a follow up PR for `trunk` only.

Feel free to merge.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedKStreamImpl.java
ableegoldman: wth happened here 🤔

ableegoldman:
@mjsax it's not a bugfix PR, this is only going to trunk. Well technically it is fixing a bug, but that bug was only merged to trunk a few weeks ago. Otherwise I would agree, that would be merge hell 🔥

streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
cadonna: I have never said you need to do it in this PR 😉 . Jokes apart, I think in general it would be better to do such things in a separate PR, but when I wrote my comment, I completely forgot about it. Sorry about that!

streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
cadonna: I like your changes. What I meant is that we could change the constructor of `SourceGraphNode` to:

```
public SourceGraphNode(final String nodeName,
                       final Set<String> topicNames,
                       final ConsumedInternal<K, V> consumedInternal)
``` 

and the one of `StreamSourceNode` to:

```
public StreamSourceNode(final String nodeName,
                        final Set<String> topicNames,
                        final ConsumedInternal<K, V> consumedInternal)
```

In this way, we have a set of topics as soon as possible in the code path from the public API. I think this makes it clearer that it is not possible to have duplicates of topics internally.

To keep this PR small, I would propose to just do the changes for `SourceGraphNode`, and do the other changes in a separate PR.

cadonna (approved):
LGTM!

I just had one minor comment about the set of topics.

ableegoldman:
Merged to trunk

streams/src/main/java/org/apache/kafka/streams/kstream/internals/graph/SourceGraphNode.java
ableegoldman: Ah ok you meant making it a Set vs a Collection -- I do agree with the principle, and I did push the Set-ification of the topics up one level so that the actual class field is a Set. But I don't think it's really worth it to push it up another layer and Set-ify the constructor argument. For one thing we would just have to do the same conversion to a Set but in more places, and more importantly, the actual callers of the constructor don't care at all whether it's a Set or any other Collection. So I think it actually does make sense to convert to a Set inside the constructor body

cadonna: Fair enough and I think that this is nothing urgent or absolute necessary. However, I would like to explain my line of thoughts. I think an interface of a class should also describe the constraints on the the object and as far as I see it does not make any sense to pass the same topic name multiple times to a source node. 
I do not see an issue with making the same conversion in more places and actually this is even not true because the only place we would do a conversion is in `StreamsBuilder#stream()`. All other dependent calls create a singleton collection which can be easily replaced with a singleton set. Actually, I do not understand why `StreamsBuilder#stream()` takes a collection instead of a set.

I am not sure I can follow your other argument

>  the actual callers of the constructor don't care at all whether it's a Set or any other Collection
 
Do you refer to the creation of the singleton collection in the callers?

As I said, I do not say we need to follow my proposal. I just wanted to argue in favor of a cleaner and more descriptive interface.

ableegoldman: > All other dependent calls create a singleton collection which can be easily replaced with a singleton set

Ah ok, I didn't notice that. I guess I only looked at `StreamsBuilder#stream`

> Actually, I do not understand why StreamsBuilder#stream() takes a collection instead of a set.

This I totally agree with. I suspect the intention was just for convenience, so users don't have to do a list->set conversion themselves, but I personally don't find that to be a very strong argument. It doesn't seem worth doing a KIP over, but maybe if we rewrite some large parts of the DSL in the future, we can fix this as well

By "callers" I meant the method body of `StreamsBuilder#stream`, which doesn't really care whether there are duplicates in the collection because its only job is to pass the topics straight from the user to this source node.

But I see your point. If I touch on some related code in a future PR I can fix this on the side, or I'd be happy to review a PR if you want to submit one. Thanks for the discussion

PR #9615, opened 2020-11-18
KAFKA-10500: Add thread option

wcarlson5:
Can add stream threads now

replace https://github.com/apache/kafka/pull/9581 after rebase

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -175,6 +175,7 @@ public class KafkaStreams implements AutoCloseable {
     private StateRestoreListener globalStateRestoreListener;
     private boolean oldHandler;
     private java.util.function.Consumer<Throwable> streamsUncaughtExceptionHandler;
+    private final Object newThread = new Object();
 
     // container states
     /**
@@ -923,24 +924,26 @@ public class KafkaStreams implements AutoCloseable {
      * @return name of the added stream thread or empty if a new stream thread could not be added
      */
     public Optional<String> addStreamThread() {
-        if (isRunningOrRebalancing()) {
-            final int threadIdx = getNextThreadIndex();
-            final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
-            resizeThreadCache(cacheSizePerThread);
-            final StreamThread streamThread = createStreamThread(cacheSizePerThread, threadIdx);
-            synchronized (stateLock) {
-                if (isRunningOrRebalancing()) {
-                    streamThread.start();
-                    return Optional.of(streamThread.getName());
-                } else {
-                    threads.remove(streamThread);
-                    resizeThreadCache(getCacheSizePerThread(threads.size()));
-                    return Optional.empty();
+        synchronized (newThread) {
+            if (isRunningOrRebalancing()) {
+                final int threadIdx = getNextThreadIndex();
+                final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
+                resizeThreadCache(cacheSizePerThread);
+                final StreamThread streamThread = createStreamThread(cacheSizePerThread, threadIdx);
+                synchronized (stateLock) {
+                    if (isRunningOrRebalancing()) {
+                        streamThread.start();
+                        return Optional.of(streamThread.getName());
+                    } else {
+                        streamThread.shutdown();
+                        threads.remove(streamThread);
+                        resizeThreadCache(getCacheSizePerThread(threads.size()));
+                        return Optional.empty();
+                    }
                 }
             }
-        } else {
-            return Optional.empty();
         }
+        return Optional.empty();
     }
 
     private int getNextThreadIndex() {
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -822,7 +822,7 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         totalCacheSize = config.getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
-        final long cacheSizePerThread = getCacheSizePerThread(numStreamThreads);
+        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
         final boolean hasPersistentStores = taskTopology.hasPersistentLocalStore() ||
                 (hasGlobalTopology && globalTaskTopology.hasPersistentGlobalStore());
 
@@ -898,12 +898,8 @@ public class KafkaStreams implements AutoCloseable {
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
-    private long getCacheSizePerThread(final int numStreamThreads) {
-        return totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
-    }
-
     private void resizeThreadCache(final int numStreamThreads) {
-        final long cacheSizePreThread = getCacheSizePerThread(numStreamThreads);
+        final long cacheSizePreThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
         for (final StreamThread streamThread: threads) {
             streamThread.resizeCache(cacheSizePreThread);
         }
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -923,9 +923,9 @@ public class KafkaStreams implements AutoCloseable {
     public Optional<String> addStreamThread() {
         synchronized (stateLock) {
             if (state == State.RUNNING || state == State.REBALANCING) {
-                final int threadIdx = threads.size() + 1;
-                final long cacheSizePerThread = getCacheSizePerThread(threadIdx);
-                resizeThreadCache(threadIdx);
+                final int threadIdx = getNextThreadIndex();
+                final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
+                resizeThreadCache(threads.size() + 1);
                 final StreamThread streamThread = StreamThread.create(
                         internalTopologyBuilder,
                         config,
@@ -947,11 +947,26 @@ public class KafkaStreams implements AutoCloseable {
                 threadState.put(streamThread.getId(), streamThread.state());
                 storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
                 streamThread.setStateListener(streamStateListener);
+                streamThread.start();
                 return Optional.of(streamThread.getName());
             } else {
                 return Optional.empty();
             }
         }
+    }
+
+    private int getNextThreadIndex() {
+        final HashSet<String> names = new HashSet<>();
+        for (final StreamThread streamThread: threads) {
+            names.add(streamThread.getName());
+        }
+        for (int i = 0; i < threads.size(); i++) {
+            final String name = clientId + "-StreamThread-" + i;
+            if (!names.contains(name)) {
+                return i;
+            }
+        }
+        return threads.size() + 1;
     }
 
     private long getCacheSizePerThread(final int numStreamThreads) {
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -822,7 +822,7 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         totalCacheSize = config.getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
-        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + (hasGlobalTopology ? 1 : 0));
+        final long cacheSizePerThread = getCacheSizePerThread(numStreamThreads);
         final boolean hasPersistentStores = taskTopology.hasPersistentLocalStore() ||
                 (hasGlobalTopology && globalTaskTopology.hasPersistentGlobalStore());
 
@@ -898,11 +898,15 @@ public class KafkaStreams implements AutoCloseable {
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
+    private long getCacheSizePerThread(final int numStreamThreads) {
+        return totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
+    }
+
     private void resizeThreadCache(final int numStreamThreads) {
         if (numStreamThreads < 0) {
             throw new IllegalArgumentException("There cannot be negative StreamThreads");
         }
-        final long cacheSizePreThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
+        final long cacheSizePreThread = getCacheSizePerThread(numStreamThreads);
         for (final StreamThread streamThread: threads) {
             streamThread.resizeCache(cacheSizePreThread);
         }
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -866,7 +866,7 @@ public class KafkaStreams implements AutoCloseable {
         threadState = new HashMap<>(numStreamThreads);
         storeProviders = new ArrayList<>();
         for (int i = 0; i < numStreamThreads; i++) {
-            makeThread(cacheSizePerThread, i + 1);
+            createStreamThread(cacheSizePerThread, i + 1);
         }
 
         ClientMetrics.addNumAliveStreamThreadMetric(streamsMetrics, (metricsConfig, now) ->
@@ -888,7 +888,7 @@ public class KafkaStreams implements AutoCloseable {
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
-    private StreamThread makeThread(final long cacheSizePerThread, final int threadIdx) {
+    private StreamThread createStreamThread(final long cacheSizePerThread, final int threadIdx) {
         final StreamThread streamThread = StreamThread.create(
                 internalTopologyBuilder,
                 config,
@@ -926,18 +926,22 @@ public class KafkaStreams implements AutoCloseable {
      * @return name of the added stream thread or empty if a new stream thread could not be added
      */
     public Optional<String> addStreamThread() {
-        synchronized (stateLock) {
-            if (isRunningOrRebalancing()) {
-                final int threadIdx = getNextThreadIndex();
-                final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
-                resizeThreadCache(threads.size() + 1);
-                final StreamThread streamThread = makeThread(cacheSizePerThread, threadIdx);
-                streamThread.setStateListener(streamStateListener);
-                streamThread.start();
-                return Optional.of(streamThread.getName());
-            } else {
-                return Optional.empty();
+        if (isRunningOrRebalancing()) {
+            final int threadIdx = getNextThreadIndex();
+            final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
+            resizeThreadCache(cacheSizePerThread);
+            final StreamThread streamThread = createStreamThread(cacheSizePerThread, threadIdx);
+            streamThread.setStateListener(streamStateListener);
+            synchronized (stateLock) {
+                if (isRunningOrRebalancing()) {
+                    streamThread.start();
+                } else {
+                    return Optional.empty();
+                }
             }
+            return Optional.of(streamThread.getName());
+        } else {
+            return Optional.empty();
         }
     }
 
@@ -959,10 +963,9 @@ public class KafkaStreams implements AutoCloseable {
         return totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
     }
 
-    private void resizeThreadCache(final int numStreamThreads) {
-        final long cacheSizePreThread = getCacheSizePerThread(numStreamThreads);
+    private void resizeThreadCache(final long cacheSizePerThread) {
         for (final StreamThread streamThread: threads) {
-            streamThread.resizeCache(cacheSizePreThread);
+            streamThread.resizeCache(cacheSizePerThread);
         }
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -866,26 +866,7 @@ public class KafkaStreams implements AutoCloseable {
         threadState = new HashMap<>(numStreamThreads);
         storeProviders = new ArrayList<>();
         for (int i = 0; i < numStreamThreads; i++) {
-            final StreamThread streamThread = StreamThread.create(
-                internalTopologyBuilder,
-                config,
-                clientSupplier,
-                adminClient,
-                processId,
-                clientId,
-                streamsMetrics,
-                time,
-                streamsMetadataState,
-                cacheSizePerThread,
-                stateDirectory,
-                delegatingStateRestoreListener,
-                i + 1,
-                KafkaStreams.this::closeToError,
-                streamsUncaughtExceptionHandler
-            );
-            threads.add(streamThread);
-            threadState.put(streamThread.getId(), streamThread.state());
-            storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
+            makeThread(cacheSizePerThread, i + 1);
         }
 
         ClientMetrics.addNumAliveStreamThreadMetric(streamsMetrics, (metricsConfig, now) ->
@@ -907,6 +888,30 @@ public class KafkaStreams implements AutoCloseable {
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
+    private StreamThread makeThread(final long cacheSizePerThread, final int threadIdx) {
+        final StreamThread streamThread = StreamThread.create(
+                internalTopologyBuilder,
+                config,
+                clientSupplier,
+                adminClient,
+                processId,
+                clientId,
+                streamsMetrics,
+                time,
+                streamsMetadataState,
+                cacheSizePerThread,
+                stateDirectory,
+                delegatingStateRestoreListener,
+                threadIdx,
+                KafkaStreams.this::closeToError,
+                streamsUncaughtExceptionHandler
+        );
+        threads.add(streamThread);
+        threadState.put(streamThread.getId(), streamThread.state());
+        storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
+        return streamThread;
+    }
+
     /**
      * Adds and starts a stream thread in addition to the stream threads that are already running in this
      * Kafka Streams client.
@@ -922,30 +927,11 @@ public class KafkaStreams implements AutoCloseable {
      */
     public Optional<String> addStreamThread() {
         synchronized (stateLock) {
-            if (state == State.RUNNING || state == State.REBALANCING) {
+            if (isRunningOrRebalancing()) {
                 final int threadIdx = getNextThreadIndex();
                 final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
                 resizeThreadCache(threads.size() + 1);
-                final StreamThread streamThread = StreamThread.create(
-                        internalTopologyBuilder,
-                        config,
-                        clientSupplier,
-                        adminClient,
-                        processId,
-                        clientId,
-                        streamsMetrics,
-                        time,
-                        streamsMetadataState,
-                        cacheSizePerThread,
-                        stateDirectory,
-                        delegatingStateRestoreListener,
-                        threadIdx,
-                        KafkaStreams.this::closeToError,
-                        streamsUncaughtExceptionHandler
-                );
-                threads.add(streamThread);
-                threadState.put(streamThread.getId(), streamThread.state());
-                storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
+                final StreamThread streamThread = makeThread(cacheSizePerThread, threadIdx);
                 streamThread.setStateListener(streamStateListener);
                 streamThread.start();
                 return Optional.of(streamThread.getName());
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -822,7 +822,7 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         totalCacheSize = config.getLong(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG);
-        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
+        final long cacheSizePerThread = totalCacheSize / (numStreamThreads + (hasGlobalTopology ? 1 : 0));
         final boolean hasPersistentStores = taskTopology.hasPersistentLocalStore() ||
                 (hasGlobalTopology && globalTaskTopology.hasPersistentGlobalStore());
 
@@ -899,6 +899,9 @@ public class KafkaStreams implements AutoCloseable {
     }
 
     private void resizeThreadCache(final int numStreamThreads) {
+        if (numStreamThreads < 0) {
+            throw new IllegalArgumentException("There cannot be negative StreamThreads");
+        }
         final long cacheSizePreThread = totalCacheSize / (numStreamThreads + ((globalTaskTopology != null) ? 1 : 0));
         for (final StreamThread streamThread: threads) {
             streamThread.resizeCache(cacheSizePreThread);
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -175,7 +175,7 @@ public class KafkaStreams implements AutoCloseable {
     private StateRestoreListener globalStateRestoreListener;
     private boolean oldHandler;
     private java.util.function.Consumer<Throwable> streamsUncaughtExceptionHandler;
-    private final Object newThread = new Object();
+    private final Object changeThreadCount = new Object();
 
     // container states
     /**
@@ -887,21 +887,21 @@ public class KafkaStreams implements AutoCloseable {
 
     private StreamThread createStreamThread(final long cacheSizePerThread, final int threadIdx) {
         final StreamThread streamThread = StreamThread.create(
-                internalTopologyBuilder,
-                config,
-                clientSupplier,
-                adminClient,
-                processId,
-                clientId,
-                streamsMetrics,
-                time,
-                streamsMetadataState,
-                cacheSizePerThread,
-                stateDirectory,
-                delegatingStateRestoreListener,
-                threadIdx,
-                KafkaStreams.this::closeToError,
-                streamsUncaughtExceptionHandler
+            internalTopologyBuilder,
+            config,
+            clientSupplier,
+            adminClient,
+            processId,
+            clientId,
+            streamsMetrics,
+            time,
+            streamsMetadataState,
+            cacheSizePerThread,
+            stateDirectory,
+            delegatingStateRestoreListener,
+            threadIdx,
+            KafkaStreams.this::closeToError,
+            streamsUncaughtExceptionHandler
         );
         streamThread.setStateListener(streamStateListener);
         threads.add(streamThread);
@@ -924,7 +924,7 @@ public class KafkaStreams implements AutoCloseable {
      * @return name of the added stream thread or empty if a new stream thread could not be added
      */
     public Optional<String> addStreamThread() {
-        synchronized (newThread) {
+        synchronized (changeThreadCount) {
             if (isRunningOrRebalancing()) {
                 final int threadIdx = getNextThreadIndex();
                 final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
@@ -952,7 +952,7 @@ public class KafkaStreams implements AutoCloseable {
             names.add(streamThread.getName());
         }
         final String baseName = clientId + "-StreamThread-";
-        for (int i = 0; i < threads.size(); i++) {
+        for (int i = 1; i <= threads.size(); i++) {
             final String name = baseName + i;
             if (!names.contains(name)) {
                 return i;
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -903,9 +903,6 @@ public class KafkaStreams implements AutoCloseable {
     }
 
     private void resizeThreadCache(final int numStreamThreads) {
-        if (numStreamThreads < 0) {
-            throw new IllegalArgumentException("There cannot be negative StreamThreads");
-        }
         final long cacheSizePreThread = getCacheSizePerThread(numStreamThreads);
         for (final StreamThread streamThread: threads) {
             streamThread.resizeCache(cacheSizePreThread);
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -865,20 +865,16 @@ public class KafkaStreams implements AutoCloseable {
 
         threadState = new HashMap<>(numStreamThreads);
         storeProviders = new ArrayList<>();
-        for (int i = 0; i < numStreamThreads; i++) {
+        streamStateListener = new StreamStateListener(threadState, globalThreadState);
+        if (hasGlobalTopology) {
+            globalStreamThread.setStateListener(streamStateListener);
+        }
+        for (int i = 1; i <= numStreamThreads; i++) {
             createStreamThread(cacheSizePerThread, i + 1);
         }
 
         ClientMetrics.addNumAliveStreamThreadMetric(streamsMetrics, (metricsConfig, now) ->
             Math.toIntExact(threads.stream().filter(thread -> thread.state().isAlive()).count()));
-
-        streamStateListener = new StreamStateListener(threadState, globalThreadState);
-        if (hasGlobalTopology) {
-            globalStreamThread.setStateListener(streamStateListener);
-        }
-        for (final StreamThread thread : threads) {
-            thread.setStateListener(streamStateListener);
-        }
 
         final GlobalStateStoreProvider globalStateStoreProvider = new GlobalStateStoreProvider(internalTopologyBuilder.globalStateStores());
         queryableStoreProvider = new QueryableStoreProvider(storeProviders, globalStateStoreProvider);
@@ -906,6 +902,7 @@ public class KafkaStreams implements AutoCloseable {
                 KafkaStreams.this::closeToError,
                 streamsUncaughtExceptionHandler
         );
+        streamThread.setStateListener(streamStateListener);
         threads.add(streamThread);
         threadState.put(streamThread.getId(), streamThread.state());
         storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
@@ -915,12 +912,12 @@ public class KafkaStreams implements AutoCloseable {
     /**
      * Adds and starts a stream thread in addition to the stream threads that are already running in this
      * Kafka Streams client.
-     *
+     * <p>
      * Since the number of stream threads increases, the sizes of the caches in the new stream thread
      * and the existing stream threads are adapted so that the sum of the cache sizes over all stream
      * threads does not exceed the total cache size specified in configuration
-     * {@code cache.max.bytes.buffering}.
-     *
+     * {@link StreamsConfig#CACHE_MAX_BYTES_BUFFERING_CONFIG}.
+     * <p>
      * Stream threads can only be added if this Kafka Streams client is in state RUNNING or REBALANCING.
      *
      * @return name of the added stream thread or empty if a new stream thread could not be added
@@ -931,12 +928,12 @@ public class KafkaStreams implements AutoCloseable {
             final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
             resizeThreadCache(cacheSizePerThread);
             final StreamThread streamThread = createStreamThread(cacheSizePerThread, threadIdx);
-            streamThread.setStateListener(streamStateListener);
             synchronized (stateLock) {
                 if (isRunningOrRebalancing()) {
                     streamThread.start();
                     return Optional.of(streamThread.getName());
                 } else {
+                    threads.remove(streamThread);
                     return Optional.empty();
                 }
             }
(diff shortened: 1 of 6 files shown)


wcarlson5:
@cadonna @mjsax @vvcephei Part 3 of KIP-663 add threads is ready for review

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +904,57 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+     * {@code cache.max.bytes.buffering}.
+     *
+     * Stream threads can only be added if this Kafka Streams client is in state RUNNING or REBALANCING.
+     *
+     * @return name of the added stream thread or empty if a new stream thread could not be added
+     */
+    public Optional<String> addStreamThread() {
+        synchronized (stateLock) {
+            if (state == State.RUNNING || state == State.REBALANCING) {
+                final int threadIdx = threads.size() + 1;
```
cadonna: Assume the following thread list [t2, t3, t4], `threadIdx` would be 4, which is already there. You should keep the currently used `threadIdx`s and check those to decide on the next `threadIdx`.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +904,57 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+                final long cacheSizePerThread = getCacheSizePerThread(threadIdx);
```
cadonna: This should be:
```suggestion
                final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +904,57 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+                resizeThreadCache(threadIdx);
```
cadonna:
```suggestion
                resizeThreadCache(threads.size() + 1);
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +904,57 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+                threads.add(streamThread);
+                threadState.put(streamThread.getId(), streamThread.state());
+                storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
+                streamThread.setStateListener(streamStateListener);
+                return Optional.of(streamThread.getName());
+            } else {
+                return Optional.empty();
+            }
+        }
+    }
```
cadonna, line 962: Where is the stream thread started?

cadonna (commented):
Thank you for the PR, @wcarlson5 !

Here is my feedback

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Looks like I didn't understand threadIdx. that makes sense now

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: right before the positive return :)

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -588,6 +592,29 @@ public void testCloseIsIdempotent() {
...
+    public void testAddThread() {
+        props.put(StreamsConfig.NUM_STREAM_THREADS_CONFIG, 1);
+        final KafkaStreams streams = new KafkaStreams(getBuilderWithSource().build(), props, supplier, time);
+        streams.start();
+        final int oldSize = streams.threads.size();
+        try {
+            TestUtils.waitForCondition(() -> streams.state() == KafkaStreams.State.RUNNING, 15L, "wait until running");
+        } catch (final InterruptedException e) {
+            e.printStackTrace();
+        }
```
cadonna: You should not use `try-catch` here but just add `throws InterruptedException` to the method signature.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +904,72 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+     * and the existing stream threads are adapted so that the sum of the cache sizes over all stream
+     * threads does not exceed the total cache size specified in configuration
+     * {@code cache.max.bytes.buffering}.
+     *
+     * Stream threads can only be added if this Kafka Streams client is in state RUNNING or REBALANCING.
+     *
+     * @return name of the added stream thread or empty if a new stream thread could not be added
+     */
+    public Optional<String> addStreamThread() {
+        synchronized (stateLock) {
```
cadonna: Why do we need to synchronize the whole method on `stateLock`?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +904,72 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+            if (state == State.RUNNING || state == State.REBALANCING) {
```
cadonna: Could we also use `isRunningOrRebalancing()` here?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +904,72 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+                        cacheSizePerThread,
+                        stateDirectory,
+                        delegatingStateRestoreListener,
+                        threadIdx,
+                        KafkaStreams.this::closeToError,
+                        streamsUncaughtExceptionHandler
+                );
+                threads.add(streamThread);
+                threadState.put(streamThread.getId(), streamThread.state());
+                storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
```
cadonna: We do something really similar when we start the stream threads at startup. Could you try to extract this part to a method?

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -588,6 +592,29 @@ public void testCloseIsIdempotent() {
             closeCount, MockMetricsReporter.CLOSE_COUNT.get());
     }
 
+    @Test
+    public void testAddThread() {
```
cadonna: I would prefer to use `shouldAddThread()` as name although the pattern is different for the other test methods.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -588,6 +592,29 @@ public void testCloseIsIdempotent() {
...
+            TestUtils.waitForCondition(() -> streams.state() == KafkaStreams.State.RUNNING, 15L, "wait until running");
+        } catch (final InterruptedException e) {
+            e.printStackTrace();
+        }
+        assertThat(streams.addStreamThread(), equalTo(Optional.of("newThread")));
+        assertThat(streams.threads.size(), equalTo(oldSize + 1));
+    }
+
+    @Test
+    public void testAddThreadNotDuringStart() {
```
cadonna: I would prefer to use `shouldNotAddThread()` as name although the pattern is different for the other test methods.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +904,72 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+     * Since the number of stream threads increases, the sizes of the caches in the new stream thread
+     * and the existing stream threads are adapted so that the sum of the cache sizes over all stream
+     * threads does not exceed the total cache size specified in configuration
+     * {@code cache.max.bytes.buffering}.
+     *
+     * Stream threads can only be added if this Kafka Streams client is in state RUNNING or REBALANCING.
+     *
+     * @return name of the added stream thread or empty if a new stream thread could not be added
+     */
+    public Optional<String> addStreamThread() {
```
cadonna, line 926: Additionally to the unit tests that you wrote, I think we also need integration tests.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Well we don't want it changing state while adding a thread

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: you are right, Ill add one

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +885,77 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+        synchronized (stateLock) {
+            if (isRunningOrRebalancing()) {
+                final int threadIdx = getNextThreadIndex();
+                final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
+                resizeThreadCache(threads.size() + 1);
```
cadonna: I think it would be cleaner to pass `cacheSizePerThread` to `resizeThreadCache()` instead of the number of stream threads. We would then just call `getCacheSizePerThread()` once instead of once in `addStreamThread()` and once in `resizeThreadCache()`. We would also just need to compute `threads.size() + 1` once.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,11 +885,77 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
         queryableStoreProvider = new QueryableStoreProvider(storeProviders, globalStateStoreProvider);
 
         stateDirCleaner = setupStateDirCleaner();
-        oldHandler = false;
         maybeWarnAboutCodeInRocksDBConfigSetter(log, config);
         rocksDBMetricsRecordingService = maybeCreateRocksDBMetricsRecordingService(clientId, config);
     }
 
+    private StreamThread makeThread(final long cacheSizePerThread, final int threadIdx) {
```
cadonna: IMO, `createStreamThread()` would describe the behavior better.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
cadonna: Wouldn't it be sufficent to check in the beginning if the Streams client is in `RUNNING` or `REBALANCING`, then optimistically create the stream thread, and before we start the stream thread check if the client is not in `PENDING_SHUTDOWN` and not in `ERROR` and not in `NOT_RUNNING`? State changes between `RUNNING` and `REBALANCING` should not affect adding a stream thread, right?

wcarlson5: Okay we can check then only synchronize around the start of the thread to make sure it doesn't shutdown between the check and the starting

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: @cadonna added

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+        properties  = mkObjectProperties(
+                mkMap(
+                        mkEntry(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, CLUSTER.bootstrapServers()),
+                        mkEntry(StreamsConfig.APPLICATION_ID_CONFIG, appId),
+                        mkEntry(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getPath()),
+                        mkEntry(StreamsConfig.NUM_STREAM_THREADS_CONFIG, 2),
+                        mkEntry(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.StringSerde.class),
+                        mkEntry(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.StringSerde.class)
+                )
```
cadonna: Please adjust indentation:

```suggestion
            mkMap(
                mkEntry(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, CLUSTER.bootstrapServers()),
                mkEntry(StreamsConfig.APPLICATION_ID_CONFIG, appId),
                mkEntry(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getPath()),
                mkEntry(StreamsConfig.NUM_STREAM_THREADS_CONFIG, 2),
                mkEntry(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.StringSerde.class),
                mkEntry(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.StringSerde.class)
            )
```

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -0,0 +1,115 @@
...
+    private void produceMessages(final long timestamp, final String streamOneInput, final String msg) {
+        IntegrationTestUtils.produceKeyValuesSynchronouslyWithTimestamp(
+                streamOneInput,
+                Collections.singletonList(new KeyValue<>("1", msg)),
+                TestUtils.producerConfig(
+                        CLUSTER.bootstrapServers(),
+                        StringSerializer.class,
+                        StringSerializer.class,
+                        new Properties()),
+                timestamp);
```
cadonna: wrong indentation

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -0,0 +1,115 @@
...
+    @Test
+    public void shouldAddStreamThread() throws Exception {
+        try (final KafkaStreams kafkaStreams = new KafkaStreams(builder.build(), properties)) {
+            StreamsTestUtils.startKafkaStreamsAndWaitForRunningState(kafkaStreams);
+            final int oldThreadCount = kafkaStreams.localThreadsMetadata().size();
+
+            kafkaStreams.addStreamThread();
+            produceMessages(0L, inputTopic, "A");
+
+            assertThat(kafkaStreams.localThreadsMetadata().size(), equalTo(oldThreadCount + 1));
```
cadonna, line 106: We should wait for the new stream thread with a timeout, otherwise we risk that this test may become flaky.

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Test
+    public void shouldAddStreamThread() throws Exception {
+        try (final KafkaStreams kafkaStreams = new KafkaStreams(builder.build(), properties)) {
+            StreamsTestUtils.startKafkaStreamsAndWaitForRunningState(kafkaStreams);
+            final int oldThreadCount = kafkaStreams.localThreadsMetadata().size();
+
+            kafkaStreams.addStreamThread();
+            produceMessages(0L, inputTopic, "A");
```
cadonna: Do we need this line?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,19 +885,88 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+            final StreamThread streamThread = createStreamThread(cacheSizePerThread, threadIdx);
+            streamThread.setStateListener(streamStateListener);
+            synchronized (stateLock) {
+                if (isRunningOrRebalancing()) {
+                    streamThread.start();
+                } else {
+                    return Optional.empty();
+                }
+            }
+            return Optional.of(streamThread.getName());
```
cadonna:
```suggestion
            synchronized (stateLock) {
                if (isRunningOrRebalancing()) {
                    streamThread.start();
                    return Optional.of(streamThread.getName());
                } else {
                    return Optional.empty();
                }
            }
```

cadonna (commented):
@wcarlson5 Thank you for the updates!

I have just some minor comments.

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
wcarlson5: no, we don't

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: sure

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
wcarlson5: we can wait for it to be added to the thread meta data. I assume that is what you mean

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -0,0 +1,104 @@
...
+    public void shouldAddStreamThread() throws Exception {
+        try (final KafkaStreams kafkaStreams = new KafkaStreams(builder.build(), properties)) {
+            StreamsTestUtils.startKafkaStreamsAndWaitForRunningState(kafkaStreams);
+            final int oldThreadCount = kafkaStreams.localThreadsMetadata().size();
+
+            final Optional<String> name = kafkaStreams.addStreamThread();
+            assertThat(name, CoreMatchers.not(Optional.empty()));
+            TestUtils.waitForCondition(() -> kafkaStreams.localThreadsMetadata().stream().sequential().map(ThreadMetadata::threadName).anyMatch(t -> t.equals(name.orElse(""))),
+                "Wait for the thread to be added");
+            assertThat(kafkaStreams.localThreadsMetadata().size(), equalTo(oldThreadCount + 1));
```
cadonna: nit: If not all parameters fit on one line, we put each of them on a new line. Additionally we put also the closing parenthesis on a new line.

nit: Lines should -- if possible -- not exceed 120 characters.

nit: I like when tests are visually structured with one block for setup, the call under test, and one block for verifications. 
 
```suggestion
            StreamsTestUtils.startKafkaStreamsAndWaitForRunningState(kafkaStreams);
            final int oldThreadCount = kafkaStreams.localThreadsMetadata().size();

            final Optional<String> name = kafkaStreams.addStreamThread();
            
            assertThat(name, CoreMatchers.not(Optional.empty()));
            TestUtils.waitForCondition(
                () -> kafkaStreams.localThreadsMetadata().stream().sequential()
                    .map(ThreadMetadata::threadName).anyMatch(t -> t.equals(name.orElse(""))), 
                "Wait for the thread to be added"
            );
            assertThat(kafkaStreams.localThreadsMetadata().size(), equalTo(oldThreadCount + 1));
```

cadonna (approved):
LGTM!

I have just a nit-comment.

Call for committer review and merge: @mjsax @vvcephei @ableegoldman @guozhangwang @abbccdda

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -398,6 +407,7 @@ public void setUncaughtExceptionHandler(final StreamsUncaughtExceptionHandler st
         final Consumer<Throwable> handler = exception -> handleStreamsUncaughtException(exception, streamsUncaughtExceptionHandler);
         synchronized (stateLock) {
             if (state == State.CREATED) {
+                this.streamsUncaughtExceptionHandler = handler;
```
mjsax, line 410: nit. remove unnecessary `this.`

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -846,43 +856,24 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
-                stateDirectory,
-                delegatingStateRestoreListener,
-                i + 1,
-                KafkaStreams.this::closeToError,
-                this::defaultStreamsUncaughtExceptionHandler
-            );
-            threads.add(streamThread);
-            threadState.put(streamThread.getId(), streamThread.state());
-            storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
+            createStreamThread(cacheSizePerThread, i + 1);
```
mjsax: Nit: can we change the loop to `int = 1; i <= numStreamThreads` and just pass in `i` here?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,19 +885,88 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+        threads.add(streamThread);
+        threadState.put(streamThread.getId(), streamThread.state());
+        storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
+        return streamThread;
+    }
+
+    /**
+     * Adds and starts a stream thread in addition to the stream threads that are already running in this
+     * Kafka Streams client.
+     *
```
mjsax: Add missing `<p>` tag

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,19 +885,88 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+     * Since the number of stream threads increases, the sizes of the caches in the new stream thread
+     * and the existing stream threads are adapted so that the sum of the cache sizes over all stream
+     * threads does not exceed the total cache size specified in configuration
+     * {@code cache.max.bytes.buffering}.
+     *
```
mjsax: As above.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,19 +885,88 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+    }
+
+    /**
+     * Adds and starts a stream thread in addition to the stream threads that are already running in this
+     * Kafka Streams client.
+     *
+     * Since the number of stream threads increases, the sizes of the caches in the new stream thread
+     * and the existing stream threads are adapted so that the sum of the cache sizes over all stream
+     * threads does not exceed the total cache size specified in configuration
+     * {@code cache.max.bytes.buffering}.
```
mjsax: Should we link to `StreamConfig` instead?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -846,43 +856,24 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
-            threadState.put(streamThread.getId(), streamThread.state());
-            storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
+            createStreamThread(cacheSizePerThread, i + 1);
         }
 
         ClientMetrics.addNumAliveStreamThreadMetric(streamsMetrics, (metricsConfig, now) ->
             Math.toIntExact(threads.stream().filter(thread -> thread.state().isAlive()).count()));
 
-        final StreamStateListener streamStateListener = new StreamStateListener(threadState, globalThreadState);
+        streamStateListener = new StreamStateListener(threadState, globalThreadState);
```
mjsax, line 869: Can we create the `StreamStateListener` before we call `createStreamThread` and do `setStateListener` within `createStreamThread` ? If yes, we also don't need to call `setStateListener` within `addStreamThread`

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,19 +885,88 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+    public Optional<String> addStreamThread() {
+        if (isRunningOrRebalancing()) {
+            final int threadIdx = getNextThreadIndex();
+            final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
+            resizeThreadCache(cacheSizePerThread);
+            final StreamThread streamThread = createStreamThread(cacheSizePerThread, threadIdx);
+            streamThread.setStateListener(streamStateListener);
+            synchronized (stateLock) {
+                if (isRunningOrRebalancing()) {
+                    streamThread.start();
```
mjsax: Just to clarify for myself: if we don't `start()` the thread, no harm is done creating it? Or would we need to do some cleanup even if we don't start the thread?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,19 +885,88 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+     * @return name of the added stream thread or empty if a new stream thread could not be added
+     */
+    public Optional<String> addStreamThread() {
+        if (isRunningOrRebalancing()) {
+            final int threadIdx = getNextThreadIndex();
+            final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
+            resizeThreadCache(cacheSizePerThread);
+            final StreamThread streamThread = createStreamThread(cacheSizePerThread, threadIdx);
+            streamThread.setStateListener(streamStateListener);
+            synchronized (stateLock) {
```
mjsax: Don't we need to get the `stateLock` as an outer most guard (and not check `isRunningOrRebalancing()` twice)? It seems weird to create a thread but later not start it and throw it away -- especially because we resize the caches (but also don't undo the resizing)?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,19 +885,88 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+                    return Optional.empty();
+                }
+            }
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    private int getNextThreadIndex() {
+        final HashSet<String> names = new HashSet<>();
```
mjsax, line 950: Why do we compute the names from scratch, but not incrementally maintain them as member variable?

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
```diff
@@ -76,6 +76,9 @@ public void resize(final long newCacheSizeBytes) {
         final boolean shrink = newCacheSizeBytes < maxCacheSizeBytes;
         maxCacheSizeBytes = newCacheSizeBytes;
         if (shrink) {
+            if (caches.values().isEmpty()) {
```
mjsax, line 79: Why this check?

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -588,6 +592,25 @@ public void testCloseIsIdempotent() {
             closeCount, MockMetricsReporter.CLOSE_COUNT.get());
     }
 
+    @Test
+    public void shouldAddThread() throws InterruptedException {
```
mjsax: `shouldAddThreadWhenRunning`

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -588,6 +592,25 @@ public void testCloseIsIdempotent() {
...
+        final KafkaStreams streams = new KafkaStreams(getBuilderWithSource().build(), props, supplier, time);
+        streams.start();
+        final int oldSize = streams.threads.size();
+        TestUtils.waitForCondition(() -> streams.state() == KafkaStreams.State.RUNNING, 15L, "wait until running");
+        assertThat(streams.addStreamThread(), equalTo(Optional.of("newThread")));
+        assertThat(streams.threads.size(), equalTo(oldSize + 1));
+    }
+
+    @Test
+    public void shouldNotAddThread() {
```
mjsax: `shouldNotAddThreadWhenCreated`

Should we also `close()` `KafkaStreams` and check that adding a thread is not possible and/or even put the client into ERROR state and test?

mjsax (commented):
Couple of comments/questions.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: the `this.` is necessary. the parameter is the same name

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I think so

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: good idea. I don't know why the `SteamStateListener` is created after the stream threads are made but it seems to work.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I think that makes more sense.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I think I explain this above. But we can remove from the thread list.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: If it is not running or rebalancing we after it was already running or rebalancing on line 930 we know the client has stopped. The number of threads will be reset to the config and everything will be rebuilt anyways, so changing the cache size should not matter.

The state lock is so that in between the second check and starting the thread the state does not change to pending shutdown or something else. I don't think its necessary to guard the whole method as the cache should be thrown out if it's not being started.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: As threads are removed we want to reuse those names, so incrementing would not work for us. Maybe there is away to store a next name, but then the logic would have to be spread out in a few places and I prefer to just compute a few names.

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
wcarlson5: Apparently `CircularIterators` throw an error if they are made on empty lists. And if there are no caches to resize we don't need to bother with the rest of this.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
wcarlson5: Good idea I'll add some tests.

wcarlson5 (commented):
@mjsax I think I got all your comments

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
mjsax: Ah. Was missing that as you assign `handler` that is not a `StreamsUncaughtExceptionHandler` but a `Consumer<Throwable>`...

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
mjsax: Maybe we had some cyclic dependency at some point in the past? Not sure.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -846,18 +856,37 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
-        for (int i = 0; i < numStreamThreads; i++) {
-            final StreamThread streamThread = StreamThread.create(
+        threadState = new HashMap<>(numStreamThreads);
+        storeProviders = new ArrayList<>();
+        streamStateListener = new StreamStateListener(threadState, globalThreadState);
+        if (hasGlobalTopology) {
+            globalStreamThread.setStateListener(streamStateListener);
+        }
+        for (int i = 1; i <= numStreamThreads; i++) {
+            createStreamThread(cacheSizePerThread, i + 1);
```
mjsax: `i + 1` -> `i` 

Wondering why this does not result in a test failure? (Or does it; Jenkins is still running.) -- Maybe we want to add a small test that verifies that we name threads correctly.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
mjsax: That is not what I meant. But it might not matter much anyway.

While we need to loop over all used names in L951 below to reuse, we don't need to compute `names` from scratch but would just modify `names` each time we add/remove a thread. But it's not perf-critical so re-doing the computation is fine, too.

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
mjsax: I see -- so it's a bug fix on the side (not directly related to this RP) -- the original PR that added this method should have added this check.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
mjsax: I see. So we exploit that possible state transitions are limited. Thanks for explaining. Makes sense.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
mjsax: Maybe. We should ensure that we do proper cleanup for all cases.

What make we wonder: I don't see any code (except the `remove` you added below) that would remove a `StreamThread` from the list? Will this be done in a different PR?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: good catch. I don't think we make sure the thread index starts at 1. But let me fix that

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: There will be two more cases of remove. In the replace thread option and in the remove thread option.

I'm not really convinced it is necessary but I don't see a problem with re-resizing the cache if we do not start the thread. I don't think there will be any side affects as the client should be shutting down, but if we resize there should be a little extra info in the state and store providers but it would not get used

streams/src/main/java/org/apache/kafka/streams/state/internals/ThreadCache.java
wcarlson5: yeah, I didn't realize this was a problem, but when I added more test coverage it showed up

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I'll remove a few of the unnecessary `+` operations then

wcarlson5 (commented):
@mjsax god catch on the +1. Hopefully I cleared up the remaining concerns, I think I took care of what you meant by "We should ensure that we do proper cleanup for all cases." for threads

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
cadonna: What about checking for the state and do the clean-up only if the state is not `PENDING_SHUTDOWN` and not `ERROR` and not `NOT_RUNNING`? In this way we are safe for future changes that break our assumption on state transitions and we make sure not to do unnecessary stuff when we are shutting down.

wcarlson5: From running or rebalancing aren't those the only states we can get to?

cadonna: Yes, currently this assumption is correct, but if the state transitions change in future, we would be safe if we do the cleanup.
On a second thought, we are probably not 100% safe because if a transition from `NOT_RUNNING` to `RUNNING` is added (or any other transition that goes from the above mentioned states to `RUNNING` or `REBALANCING`), we would still not do the clean up.

wcarlson5: I think when a state transition is changed or add is when these changes should be made. Removing from the thread list is low cost as is increasing the size of the cache, so it won't be expensive to make these changes for all cases.

I think the two good options we have is that we can move the cache resize and create thread into the stateLock or we can undo the changes we made if we have to abort starting the new thread.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -870,43 +899,73 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+            final int threadIdx = getNextThreadIndex();
+            final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
+            resizeThreadCache(cacheSizePerThread);
+            final StreamThread streamThread = createStreamThread(cacheSizePerThread, threadIdx);
+            synchronized (stateLock) {
+                if (isRunningOrRebalancing()) {
+                    streamThread.start();
+                    return Optional.of(streamThread.getName());
+                } else {
+                    threads.remove(streamThread);
```
ableegoldman: We should also shutdown the thread if it doesn't get started, otherwise me may leak (consumer or producer) clients. But I'm actually not sure why we don't just do everything (resize cache, create thread) inside the synchronized block? I'm guessing it would deadlock due to locking on the `statelock` but can't we just synchronize on something else that wouldn't interfere with the StreamThread creation?

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -0,0 +1,108 @@
...
+            final Optional<String> name = kafkaStreams.addStreamThread();
+
+            assertThat(name, CoreMatchers.not(Optional.empty()));
+            TestUtils.waitForCondition(
+                () -> kafkaStreams.localThreadsMetadata().stream().sequential()
+                        .map(ThreadMetadata::threadName).anyMatch(t -> t.equals(name.orElse(""))),
+                "Wait for the thread to be added"
+            );
+            assertThat(kafkaStreams.localThreadsMetadata().size(), equalTo(oldThreadCount + 1));
+        }
```
ableegoldman, line 109: Can we also assert that the state gets to `RUNNING` after the new thread has joined

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
cadonna: Good point about the shutdown of the stream thread!

Actually, I did not want to have everything in the synchronized block because I thought blocking the client state more than needed was not a good idea. I thought decreasing the size of the cache might be costly if the evicted records are forwarded downstream. 
Now that you mention to synchronize on a separate lock, I noticed that we probably need to put resize, start, and cleanup in the same synchronized block. The reason is that if two threads call `addStreamThread()` one after the other and the later thread passes

```
final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
```

before the earlier thread adds the new stream thread to `threads` in `createStreamThread()`, the later thread would compute the wrong cache size.

So, I am in favor of having a separate lock that just synchronizes the threads calling `addStreamThread()`. Maybe we can simply synchronize the whole method (which means to synchronize with `start()` and `close()`).

Still a minor issue seems to be the synchronization between`isRunningOrRebalancing()` and `streamThread.start()`. If between these two calls the Streams client transits to `ERROR` (the global stream thread died) an `IllegalStateException` would be thrown from the `StreamStateListener` because the Streams client would try to transit from `ERROR` to `REBALANCING`. But I guess that would also happen if the Streams client transits to `ERROR` before the new stream thread transits to `PARTITION_ASSIGNED` and calls the `StreamStateListener` that would transit the Streams client to `REBALANCING`. So it needs to be fixed somewhere else.

Did I miss something?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
cadonna: See my other comment https://github.com/apache/kafka/pull/9615#discussion_r534261843

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Unfortunately I don't think we can shutdown a thread until we have started it.

 I don't think there should be a dead lock by just using the state lock around most of the method. It as mostly about cost. However I think that its is probably the safest way as it solves our problem about needing to remove a thread we have just created and we won't potentially waste time resizing the cache as such.

If we synchronize on a new lock we end up with all the problems we were trying to solve earlier anyways with the possible state changes and having to clean up un-started threads

cadonna: > Unfortunately I don't think we can shutdown a thread until we have started it.

Have a look at https://github.com/apache/kafka/blob/aeeb7b2f9a9abe8f49543a2278757722e5974cb3/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java#L976-L983

wcarlson5: Oh okay, when I shutdown unstarted threads in a test I got a `java.lang.IllegalStateException Unexpected state transition`. But it looks like that is the client.

I added a new lock for the add threads, and shutdown the thread. I think this address the problem you found with concurrent resizes. As well as @ableegoldman 's concerns

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -870,43 +900,75 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
-                i + 1,
+                threadIdx,
                 KafkaStreams.this::closeToError,
-                this::defaultStreamsUncaughtExceptionHandler
-            );
-            threads.add(streamThread);
-            threadState.put(streamThread.getId(), streamThread.state());
-            storeProviders.add(new StreamThreadStateStoreProvider(streamThread));
-        }
+                streamsUncaughtExceptionHandler
```
cadonna: nit: If it happens that you need to push another commit, could you fix the indentation here? Sorry that I haven't noticed this before.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -870,43 +900,75 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+     * @return name of the added stream thread or empty if a new stream thread could not be added
+     */
+    public Optional<String> addStreamThread() {
+        synchronized (newThread) {
+            if (isRunningOrRebalancing()) {
+                final int threadIdx = getNextThreadIndex();
+                final long cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
+                resizeThreadCache(cacheSizePerThread);
+                final StreamThread streamThread = createStreamThread(cacheSizePerThread, threadIdx);
+                synchronized (stateLock) {
```
cadonna, line 933: Sorry to bother you again with the synchronization on the `stateLock`, but could you explain why we still need it after we synchronize on `newThread`?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -870,43 +900,75 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
-            globalStreamThread.setStateListener(streamStateListener);
+    private int getNextThreadIndex() {
+        final HashSet<String> names = new HashSet<>();
+        for (final StreamThread streamThread: threads) {
+            names.add(streamThread.getName());
         }
-        for (final StreamThread thread : threads) {
-            thread.setStateListener(streamStateListener);
+        final String baseName = clientId + "-StreamThread-";
+        for (int i = 0; i < threads.size(); i++) {
```
cadonna: Shouldn't that be `int i = 1; i <= threads.size(); i++`? Otherwise, we would look up `*-StreamThread-0"` and we would not look up `"*-StreamThread-" + threads.size()`.

Could you add some tests that check the correct naming as @mjsax suggested?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Well newThread only syncs the addThread method. There is still the race condition between the second check of is running and starting the thread. It seems like a bad idea to leave that open as it could cause thread state changes when there shouldn't be. Starting the thread is relatively low cost so this shouldn't have much impact perf wise.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: ah good catch. the diff makes that hard to see as it was actually moved to a new method.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Sure, I missed that suggestion

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Expanding on this, the problem in the shutdown thread. When the join only waits for alive threads, and to be alive the thread needs to be started. 

So if in between the check and the start thread another thread transitions the state to NOT_RUNNING the thread will not join in the shutdown thread. Then when it continues it will start as it passed the check and we will have a thread running after the client is shutdown.

This would be extremely though race condition to find or reproduce so best to just avoid it.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -845,67 +856,118 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+     * and the existing stream threads are adapted so that the sum of the cache sizes over all stream
+     * threads does not exceed the total cache size specified in configuration
+     * {@link StreamsConfig#CACHE_MAX_BYTES_BUFFERING_CONFIG}.
+     * <p>
+     * Stream threads can only be added if this Kafka Streams client is in state RUNNING or REBALANCING.
+     *
+     * @return name of the added stream thread or empty if a new stream thread could not be added
+     */
+    public Optional<String> addStreamThread() {
+        synchronized (changeThreadCount) {
```
mjsax, line 927: Why do we need this new lock-object? Would it not be simpler to just reuse `stateLock` ?

mjsax (commented):
Just one final question...

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: We thought it would be better use a separate lock because it is serving a different purpose. It will also use used in remove thread. It might be simpler to reuse the statelock but I don’t think that would be cleaner. We are really locking on the thread cache access and the thread indexes

PR #9616, opened 2020-11-18
KAFKA-10091: KIP-695: Deterministic semantics for task idling

vvcephei:
Expose fetched metadata per partition so that Streams
can enqueue records for each input and select among
them for processing in timestamp order.

Implements KIP-695

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/PartitionGroup.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/PartitionGroup.java
@@ -17,16 +17,22 @@
 package org.apache.kafka.streams.processor.internals;
 
 import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.metrics.Sensor;
+import org.apache.kafka.streams.StreamsConfig;
+import org.apache.kafka.streams.processor.TaskId;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.PriorityQueue;
 import java.util.Set;
-import java.util.Iterator;
-import java.util.HashSet;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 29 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/PartitionGroup.java
```diff
@@ -78,15 +89,147 @@ RecordQueue queue() {
...
+
+        final Set<TopicPartition> queued = new HashSet<>();
+        Map<TopicPartition, Long> enforced = null;
+
+        for (final Map.Entry<TopicPartition, RecordQueue> entry : partitionQueues.entrySet()) {
+            final TopicPartition partition = entry.getKey();
+            final RecordQueue queue = entry.getValue();
+
+            // TODO: hardcoded stale metadata at 100s for POC, need to compute it more intelligently
+            fetchedMetadata.computeIfPresent(partition, (tp, m) -> wallClockTime - m.receivedTimestamp > 100_000L ? null : m);
```
vvcephei: @guozhangwang , do you have any better ideas of how we can heuristically determine whether the lag is stale or not?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -331,9 +331,29 @@ public static StreamThread create(final InternalTopologyBuilder builder,
...
+            consumerConfigs.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "none");
+        }
+
+        final int maxPollTimeMs =
+            new InternalConsumerConfig(config.getMainConsumerConfigs("", "", 0))
+                .getInt(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG);
+
+        // Heuristically bound the topic-partition metadata staleness at twice the max poll interval.
+        // This should typically mean that we get at least four chances to refresh the metadata before it expires.
+        final long metadataStalenessBoundMs = 2L * maxPollTimeMs;
```
guozhangwang, line 351: Why it's four changes?

BTW I think we should test his heuristics a bit in soak: normally for partitions that have empty data, the metadata would still be refreshed for each poll; and if we observe the same maybe we can set this a bit more strict.

vvcephei: Good question. It's four because we "learn" the numIterations in StreamThread so that processing takes no more than 1/2 the max poll interval. If we wait for 2 max poll intervals, then we should have at least 4 calls to poll. If Streams runs out of records to process, we'll call poll sooner, so we're probably calling poll quite a bit more frequently.

Note that if we do get fresher metadata, we overwrite the older copy and use the new information instead. I'm not sure if there's any real benefit in being even stricter. What I was aiming for was rather a bound that we'd probably only hit if something is really wrong, eg talking to the brokers.

guozhangwang:
Hey @vvcephei , as the KIP is closing to be accepted, do you plan to break down this PR to smaller, multi-staged PRs? :)

vvcephei:
Implemented in:
a92b986c855592d630fbabf49d1e9a160ad5b230
4d28391480fd8c547a63af119bba67fceb5d2ede
fdcf8fbf72bee9e672d0790cdbe5539846f7dc8e
2bfe7a9d1ec7a39bb6cb290ed91644f6ad14cb3f

vvcephei:
Implemented in:
a92b986c855592d630fbabf49d1e9a160ad5b230
4d28391480fd8c547a63af119bba67fceb5d2ede
fdcf8fbf72bee9e672d0790cdbe5539846f7dc8e
2bfe7a9d1ec7a39bb6cb290ed91644f6ad14cb3f

PR #9617, opened 2020-11-18
MINOR: Factor out common response parsing logic

hachikuji:
This patch factors out some common parsing logic from `NetworkClient.parseResponse` and `AbstractResponse.parseResponse`. As a result of this refactor, we are now verifying the correlationId in forwarded requests. This patch also adds a test case to verify handling in this case.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java
+++ b/clients/src/main/java/org/apache/kafka/clients/NetworkClient.java
@@ -30,14 +30,13 @@ import org.apache.kafka.common.network.NetworkReceive;
 import org.apache.kafka.common.network.Selectable;
 import org.apache.kafka.common.network.Send;
 import org.apache.kafka.common.protocol.ApiKeys;
-import org.apache.kafka.common.protocol.CommonFields;
 import org.apache.kafka.common.protocol.Errors;
 import org.apache.kafka.common.protocol.types.SchemaException;
-import org.apache.kafka.common.protocol.types.Struct;
 import org.apache.kafka.common.requests.AbstractRequest;
 import org.apache.kafka.common.requests.AbstractResponse;
 import org.apache.kafka.common.requests.ApiVersionsRequest;
 import org.apache.kafka.common.requests.ApiVersionsResponse;
+import org.apache.kafka.common.requests.CorrelationIdMismatchException;
 import org.apache.kafka.common.requests.MetadataRequest;
 import org.apache.kafka.common.requests.MetadataResponse;
 import org.apache.kafka.common.requests.RequestHeader;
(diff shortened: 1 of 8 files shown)


hachikuji:
This is a follow-up from #9563. cc @chia7712 @abbccdda 

While I was working on this patch, I noticed this: https://issues.apache.org/jira/browse/KAFKA-10743. In this patch, I have just removed the metric.

clients/src/main/java/org/apache/kafka/common/requests/CorrelationIdMismatchException.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.common.requests;
+
+/**
+ * Raised if the correlationId in a response header does not match
+ * the expected value from the request header.
+ */
+public class CorrelationIdMismatchException extends IllegalStateException {
```
abbccdda, line 23: nit: we could put this exception into `common/internals` to indicate that we don't plan to let external user catch it.

clients/src/main/java/org/apache/kafka/common/requests/CorrelationIdMismatchException.java
```diff
@@ -0,0 +1,45 @@
...
+        int requestCorrelationId,
+        int responseCorrelationId
+    ) {
+        super(message);
+        this.requestCorrelationId = requestCorrelationId;
+        this.responseCorrelationId = responseCorrelationId;
+    }
+
+    public int requestCorrelationId() {
+        return requestCorrelationId;
```
abbccdda, line 38: This field is not used.

clients/src/main/java/org/apache/kafka/clients/NetworkClient.java
```diff
@@ -974,21 +969,6 @@ private void handleInitiateApiVersionRequests(long now) {
         }
     }
 
-    /**
-     * Validate that the response corresponds to the request we expect or else explode
-     */
-    private static void correlate(RequestHeader requestHeader, ResponseHeader responseHeader) {
```
abbccdda, line 980: nit: I don't think we need to remove this helper

abbccdda (commented):
Overall LGTM, just some minor comments

core/src/main/scala/kafka/server/ForwardingManager.scala
```diff
@@ -17,25 +17,17 @@
...
-class ForwardingManager(metadataCache: kafka.server.MetadataCache,
-                        time: Time,
-                        metrics: Metrics,
-                        config: KafkaConfig,
-                        threadNamePrefix: Option[String] = None) extends
-  BrokerToControllerChannelManagerImpl(metadataCache, time, metrics,
-    config, "forwardingChannel", threadNamePrefix) with KafkaMetricsGroup {
-
-  private val forwardingMetricName = "NumRequestsForwardingToControllerPerSec"
+class ForwardingManager(channelManager: BrokerToControllerChannelManager) extends Logging {
```
chia7712, line 30: As ```ForwardingManager``` does not extend ```BrokerToControllerChannelManagerImpl```, the ```requestQueue``` can be declared as private now.

https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala#L58

clients/src/main/java/org/apache/kafka/clients/NetworkClient.java
hachikuji: Why do you want to keep it if it is not used?

clients/src/main/java/org/apache/kafka/common/requests/CorrelationIdMismatchException.java
hachikuji: Yeah, I added it for completeness. It is a little odd if the mismatch exception only tells you what the response correlationId was.

clients/src/main/java/org/apache/kafka/common/requests/CorrelationIdMismatchException.java
hachikuji: Yeah, I was debating the location. I decided to put it in `common/requests` since it is not a public package and the error is specific to the request protocol.

PR #9619, opened 2020-11-18
MINOR: Reduce sends created by `SendBuilder`

hachikuji:
In #9401, I observed some strange latency behavior when testing locally (on macos). The behavior seemed to be caused by slightly different write behavior as a result of the new `SendBuilder` class. After the change, we were making several calls down to both `GatheringByteChannel.write(ByteBuffer[])` as well as `WritableByteChannel.write(ByteBuffer)` whereas previously there would be just one path through the former. The difference is clearly seen in the flame graphs posted here: https://github.com/apache/kafka/pull/9401#issuecomment-729258346. 

This patch changes the grouping of `Send` objects created by `SendBuilder` to try and restore the previous write behavior. Rather than creating a single `ByteBufferSend` for each `ByteBuffer`, we attempt to group consecutive buffers into a single `ByteBufferSend`. I confirmed that did indeed address the latency issue I observed using the same test setup with a single broker that was discussed in #9401.

```
bin/kafka-topics.sh --create --topic foo --replication-factor 1 --partitions 10 --bootstrap-server $BROKER
bin/kafka-producer-perf-test.sh --topic foo --num-records 250000000 --throughput -1  --record-size 256 --producer-props bootstrap.servers=$BROKER
```

Here are the results:

```
Before KAFKA-9628:
250000000 records sent, 1398546.630342 records/sec (341.44 MB/sec), 39.71 ms avg latency, 424.00 ms max latency, 16 ms 50th, 137 ms 95th, 194 ms 99th, 270 ms 99.9th.

Trunk:
250000000 records sent, 1413619.374502 records/sec (345.12 MB/sec), 87.72 ms avg latency, 2058.00 ms max latency, 64 ms 50th, 247 ms 95th, 429 ms 99th, 790 ms 99.9th.                                             

This patch:
250000000 records sent, 1469723.691946 records/sec (358.82 MB/sec), 12.72 ms avg latency, 320.00 ms max latency, 2 ms 50th, 83 ms 95th, 162 ms 99th, 256 ms 99.9th.                                                
```

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
+++ b/clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
@@ -41,7 +41,7 @@ import java.util.Queue;
  * for example usage.
  */
 public class SendBuilder implements Writable {
-    private final Queue<Send> sends = new ArrayDeque<>();
+    private final Queue<Send> sends = new ArrayDeque<>(1);
     private final List<ByteBuffer> buffers = new ArrayList<>();
     private final ByteBuffer buffer;
     private final String destinationId;
--- a/clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
+++ b/clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
@@ -19,6 +19,7 @@ package org.apache.kafka.common.protocol;
 import org.apache.kafka.common.network.ByteBufferSend;
 import org.apache.kafka.common.network.Send;
 import org.apache.kafka.common.record.BaseRecords;
+import org.apache.kafka.common.record.MemoryRecords;
 import org.apache.kafka.common.record.MultiRecordsSend;
 import org.apache.kafka.common.requests.RequestHeader;
 import org.apache.kafka.common.requests.ResponseHeader;
@@ -26,6 +27,8 @@ import org.apache.kafka.common.utils.ByteUtils;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Queue;
 
 /**
@@ -39,6 +42,7 @@ import java.util.Queue;
  */
 public class SendBuilder implements Writable {
     private final Queue<Send> sends = new ArrayDeque<>();
+    private final List<ByteBuffer> buffers = new ArrayList<>();
     private final ByteBuffer buffer;
     private final String destinationId;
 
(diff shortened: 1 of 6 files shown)


hachikuji:
cc @chia7712 @dajac

clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
```diff
@@ -39,6 +42,7 @@
  */
 public class SendBuilder implements Writable {
     private final Queue<Send> sends = new ArrayDeque<>();
+    private final List<ByteBuffer> buffers = new ArrayList<>();
```
chia7712: why ```sends``` is ```ArrayDeque``` and ```buffers``` is ```ArrayList```?

clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
```diff
@@ -130,13 +139,25 @@ public void writeVarlong(long i) {
...
     public void writeRecords(BaseRecords records) {
-        flushCurrentBuffer();
-        sends.add(records.toSend(destinationId));
+        flushPendingBuffer();
+
+        if (records instanceof MemoryRecords) {
+            buffers.add(((MemoryRecords) records).buffer());
+        } else {
+            flushPendingSend();
+            sends.add(records.toSend(destinationId));
```
chia7712: How about setting initial size of ```sends``` to 1 as auto-generated code use only ```MemoryRecords```

chia7712 (approved):
@hachikuji +1 to this improvement.

just curious. Why the "big" latency can be made without this improvement. I run the perf on my machine and can't get such big different latency. Is ```MultiRecordsSend``` a bad pattern to our code?

hachikuji:
@chia7712 To be honest, I'm not sure. I think it makes sense in general to try and batch the writes together to reduce number of calls to the kernel in any case, but I am not sure why it has such a big impact locally. I also didn't see anything like this when I was testing in on linux, but I didn't feel great about the difference anyway.

clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
hachikuji: Oh, just because `MultiRecordsSend` needs a Queue.

clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
hachikuji: That's fair. `FetchResponse` is the only case where we would use a different `Records` type, so may as well optimize for the common case.

chia7712:
+1 to last commit

clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
```diff
@@ -122,6 +123,14 @@ public void writeVarlong(long i) {
         ByteUtils.writeVarlong(i, buffer);
     }
 
+    private void flushPendingSend() {
+        if (!buffers.isEmpty()) {
+            ByteBuffer[] byteBufferArray = buffers.toArray(new ByteBuffer[0]);
+            sends.add(new ByteBufferSend(destinationId, byteBufferArray));
```
dajac: I just noticed that `ByteBufferSend` re-iterates over the byte buffers to compute the total size. We could compute the size while we accumulate them. I suppose that the number of buffers is usually small so it should not make a big difference.

dajac (approved):
LGTM, nice optimization!

clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
hachikuji: I was somewhat inclined to leave this as is, but finally I decided to do it since it might help (even if only a little) with fetch overhead.

dajac (approved):
LGTM, thanks for the update.

PR #9620, opened 2020-11-19
MINOR: Get rid of generic from FetchResponse

chia7712:
Except for LazyDownConversionRecords, we don't use the generic. We should handle the specify case directly instead of making generic in whole project.

In short, this patch is a kind of cleanup.


ijuma:
It's a bit hard to know when the Lazy version can be used though, right?

chia7712:
> It's a bit hard to know when the Lazy version can be used though, right?

You are right :(

Will go back when I have a better idea.

chia7712:
superseded by #9758

PR #9621, opened 2020-11-19
KAFKA-9892: Producer state snapshot needs to be forced to disk

bristy:
FileChannel.close() does not guarantee modified buffer would be written on the file system. We are changing  it with force() semantics to enforce file buffer and metadata written to filesystem ( FileChannel.force(true) updates buffer and metadata).

I have run unittests after making the changes.

### DIFF
--- a/core/src/main/scala/kafka/log/ProducerStateManager.scala
+++ b/core/src/main/scala/kafka/log/ProducerStateManager.scala
@@ -436,8 +436,12 @@ object ProducerStateManager {
     ByteUtils.writeUnsignedInt(buffer, CrcOffset, crc)
 
     val fileChannel = FileChannel.open(file.toPath, StandardOpenOption.CREATE, StandardOpenOption.WRITE)
-    try fileChannel.write(buffer)
-    finally fileChannel.force(true)
+    try {
+      fileChannel.write(buffer)
+      fileChannel.force(true)
+    } finally {
+      fileChannel.close()
+    }
   }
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/log/ProducerStateManager.scala
```diff
@@ -437,7 +437,7 @@ object ProducerStateManager {
 
     val fileChannel = FileChannel.open(file.toPath, StandardOpenOption.CREATE, StandardOpenOption.WRITE)
     try fileChannel.write(buffer)
-    finally fileChannel.close()
+    finally fileChannel.force(true)
```
junrao: I think we need to do force and then close.

junrao (commented):
@bristy : Thanks for the PR. Left a comment below.

hachikuji:
@bristy Thanks for the patch. I'd suggest changing to this:
```java
    try {
      fileChannel.write(buffer)
      fileChannel.force(true)
    } finally {
      fileChannel.close()
    }
```
If you don't have time to get back to this, I can push a commit.

core/src/main/scala/kafka/log/ProducerStateManager.scala
bristy: Fixed. Please check

hachikuji (approved):
LGTM. Thanks for the patch!

junrao (approved):
@bristy : Thanks for the updated PR. LGTM

PR #9622, opened 2020-11-19
KAFKA-10547; add topicId in MetadataResp

dengziming:

1. Bump the version of MetadataReq and MetadataResp, add topicId in MetadataResp
2. Alter describeTopic in AdminClientTopicService and ZookeeperTopicService
3. TopicMetadata is cached in MetadataCache, so we need to add topicId to MetadataCache
4. MetadataCache is updated by UpdateMetadataRequest, bump the version of UpdateMetadataReq and UpdateMetadataResp, add topicId in UpdateMetadataReq.

Tested locally, here is some result:

New server + new Client : 

kafka-topics.sh --describe --zookeeper localhost:2181 --topic old-version-topic 
Topic: old-version-topic	TopicId: wRPl6VAlQeyE77bDxEESzg	PartitionCount: 2	ReplicationFactor: 1	Configs: 
	Topic: old-version-topic	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
	Topic: old-version-topic	Partition: 1	Leader: 0	Replicas: 0	Isr: 0

kafka-topics.sh --describe --bootstrap-server localhost:9092 --topic old-version-topic 
Topic: old-version-topic	TopicId: wRPl6VAlQeyE77bDxEESzg	PartitionCount: 2	ReplicationFactor: 1	Configs: segment.bytes=1073741824
	Topic: old-version-topic	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
	Topic: old-version-topic	Partition: 1	Leader: 0	Replicas: 0	Isr: 0

Old Server + new Client

kafka-topics.sh --describe --bootstrap-server localhost:9092 --topic old-version-topic 
Topic: old-version-topic	PartitionCount: 2	ReplicationFactor: 1	Configs: segment.bytes=1073741824
	Topic: old-version-topic	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
	Topic: old-version-topic	Partition: 1	Leader: 0	Replicas: 0	Isr: 0

New server + old client

kafka-topics.sh --describe --bootstrap-server localhost:9092 --topic old-version-topic 
Topic: old-version-topic	PartitionCount: 2	ReplicationFactor: 1	Configs: segment.bytes=1073741824
	Topic: old-version-topic	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
	Topic: old-version-topic	Partition: 1	Leader: 0	Replicas: 0	Isr: 0

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/Cluster.java
+++ b/clients/src/main/java/org/apache/kafka/common/Cluster.java
@@ -46,6 +46,7 @@ public final class Cluster {
     private final Map<Integer, List<PartitionInfo>> partitionsByNode;
     private final Map<Integer, Node> nodesById;
     private final ClusterResource clusterResource;
+    private final Map<String, Uuid> topicIds;
 
     /**
      * Create a new cluster with the given id, nodes and partitions
@@ -57,7 +58,7 @@ public final class Cluster {
                    Collection<PartitionInfo> partitions,
                    Set<String> unauthorizedTopics,
                    Set<String> internalTopics) {
-        this(clusterId, false, nodes, partitions, unauthorizedTopics, Collections.emptySet(), internalTopics, null);
+        this(clusterId, false, nodes, partitions, unauthorizedTopics, Collections.emptySet(), internalTopics, null, Collections.emptyMap());
     }
 
     /**
@@ -71,7 +72,7 @@ public final class Cluster {
                    Set<String> unauthorizedTopics,
                    Set<String> internalTopics,
                    Node controller) {
-        this(clusterId, false, nodes, partitions, unauthorizedTopics, Collections.emptySet(), internalTopics, controller);
+        this(clusterId, false, nodes, partitions, unauthorizedTopics, Collections.emptySet(), internalTopics, controller, Collections.emptyMap());
     }
 
     /**
@@ -86,7 +87,23 @@ public final class Cluster {
                    Set<String> invalidTopics,
                    Set<String> internalTopics,
                    Node controller) {
-        this(clusterId, false, nodes, partitions, unauthorizedTopics, invalidTopics, internalTopics, controller);
+        this(clusterId, false, nodes, partitions, unauthorizedTopics, invalidTopics, internalTopics, controller, Collections.emptyMap());
+    }
+
+    /**
+     * Create a new cluster with the given id, nodes, partitions and topicIds
+     * @param nodes The nodes in the cluster
+     * @param partitions Information about a subset of the topic-partitions this cluster hosts
+     */
+    public Cluster(String clusterId,
+                   Collection<Node> nodes,
+                   Collection<PartitionInfo> partitions,
+                   Set<String> unauthorizedTopics,
+                   Set<String> invalidTopics,
+                   Set<String> internalTopics,
+                   Node controller,
+                   Map<String, Uuid> topicIds) {
+        this(clusterId, false, nodes, partitions, unauthorizedTopics, invalidTopics, internalTopics, controller, topicIds);
     }
 
     private Cluster(String clusterId,
@@ -96,7 +113,8 @@ public final class Cluster {
                     Set<String> unauthorizedTopics,
                     Set<String> invalidTopics,
                     Set<String> internalTopics,
-                    Node controller) {
+                    Node controller,
+                    Map<String, Uuid> topicIds) {
         this.isBootstrapConfigured = isBootstrapConfigured;
         this.clusterResource = new ClusterResource(clusterId);
         // make a randomized, unmodifiable copy of the nodes
@@ -165,6 +183,7 @@ public final class Cluster {
         this.partitionsByTopic = Collections.unmodifiableMap(tmpPartitionsByTopic);
         this.availablePartitionsByTopic = Collections.unmodifiableMap(tmpAvailablePartitionsByTopic);
         this.partitionsByNode = Collections.unmodifiableMap(tmpPartitionsByNode);
+        this.topicIds = Collections.unmodifiableMap(topicIds);
 
         this.unauthorizedTopics = Collections.unmodifiableSet(unauthorizedTopics);
         this.invalidTopics = Collections.unmodifiableSet(invalidTopics);
@@ -191,7 +210,7 @@ public final class Cluster {
         for (InetSocketAddress address : addresses)
             nodes.add(new Node(nodeId--, address.getHostString(), address.getPort()));
         return new Cluster(null, true, nodes, new ArrayList<>(0),
-            Collections.emptySet(), Collections.emptySet(), Collections.emptySet(), null);
+            Collections.emptySet(), Collections.emptySet(), Collections.emptySet(), null, Collections.emptyMap());
     }
 
     /**
@@ -325,6 +344,14 @@ public final class Cluster {
 
     public Node controller() {
         return controller;
+    }
+
+    public Collection<Uuid> topicIds() {
+        return topicIds.values();
+    }
+
+    public Uuid topicId(String topic) {
+        return topicIds.getOrDefault(topic, Uuid.ZERO_UUID);
     }
 
     @Override
--- a/clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
+++ b/clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
@@ -20,6 +20,7 @@ import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.Node;
 import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.Uuid;
 import org.apache.kafka.common.message.MetadataResponseData;
 import org.apache.kafka.common.message.MetadataResponseData.MetadataResponseBroker;
 import org.apache.kafka.common.message.MetadataResponseData.MetadataResponsePartition;
@@ -125,18 +126,22 @@ public class MetadataResponse extends AbstractResponse {
     public Cluster cluster() {
         Set<String> internalTopics = new HashSet<>();
         List<PartitionInfo> partitions = new ArrayList<>();
+        Map<String, Uuid> topicIds = new HashMap<>();
 
         for (TopicMetadata metadata : topicMetadata()) {
             if (metadata.error == Errors.NONE) {
                 if (metadata.isInternal)
                     internalTopics.add(metadata.topic);
+                if (metadata.topicId() != null && metadata.topicId() != Uuid.ZERO_UUID) {
+                    topicIds.put(metadata.topic, metadata.topicId());
+                }
                 for (PartitionMetadata partitionMetadata : metadata.partitionMetadata) {
                     partitions.add(toPartitionInfo(partitionMetadata, holder().brokers));
                 }
             }
         }
         return new Cluster(data.clusterId(), brokers(), partitions, topicsByError(Errors.TOPIC_AUTHORIZATION_FAILED),
-                topicsByError(Errors.INVALID_TOPIC_EXCEPTION), internalTopics, controller());
+                topicsByError(Errors.INVALID_TOPIC_EXCEPTION), internalTopics, controller(), topicIds);
     }
 
     public static PartitionInfo toPartitionInfo(PartitionMetadata metadata, Map<Integer, Node> nodesById) {
@@ -251,17 +256,20 @@ public class MetadataResponse extends AbstractResponse {
     public static class TopicMetadata {
         private final Errors error;
         private final String topic;
+        private final Uuid topicId;
         private final boolean isInternal;
         private final List<PartitionMetadata> partitionMetadata;
         private int authorizedOperations;
 
         public TopicMetadata(Errors error,
                              String topic,
+                             Uuid topicId,
                              boolean isInternal,
                              List<PartitionMetadata> partitionMetadata,
                              int authorizedOperations) {
             this.error = error;
             this.topic = topic;
+            this.topicId = topicId;
             this.isInternal = isInternal;
             this.partitionMetadata = partitionMetadata;
             this.authorizedOperations = authorizedOperations;
@@ -271,7 +279,7 @@ public class MetadataResponse extends AbstractResponse {
                              String topic,
                              boolean isInternal,
                              List<PartitionMetadata> partitionMetadata) {
-            this(error, topic, isInternal, partitionMetadata, AUTHORIZED_OPERATIONS_OMITTED);
+            this(error, topic, Uuid.ZERO_UUID, isInternal, partitionMetadata, AUTHORIZED_OPERATIONS_OMITTED);
         }
 
         public Errors error() {
@@ -280,6 +288,10 @@ public class MetadataResponse extends AbstractResponse {
 
         public String topic() {
             return topic;
+        }
+
+        public Uuid topicId() {
+            return topicId;
         }
 
         public boolean isInternal() {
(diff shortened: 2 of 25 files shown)


dengziming:
@rajinisivaram @jolshan Hi, PTAL.

jolshan:
Hi @dengziming! Thanks for the PR! I was hoping to add LeaderAndIsrRequests before UpdateMetadata/Metadata, following the ordering of the  [JIRA tickets.](https://issues.apache.org/jira/browse/KAFKA-8872) There are just a few features for persisting the topic IDs I wanted to include. I'm thinking we could review this PR and my PR: https://github.com/apache/kafka/pull/9626 at the same time and merge mine first and yours immediately after.

clients/src/main/java/org/apache/kafka/common/Cluster.java
```diff
@@ -327,6 +350,18 @@ public Node controller() {
...
 
+    public Collection<Uuid> topicIds() {
+        return topicIds.values();
+    }
+
+    public Uuid getTopicId(String topic) {
+        return topicIds.getOrDefault(topic, Uuid.ZERO_UUID);
+    }
+
+    public String getTopicName(Uuid topiId) {
```
jolshan: nit: `Uuid topicId`

jolshan: Also, for getTopicId, we do getOrDefault. Is there a reason we wouldn't do that here?

core/src/main/scala/kafka/admin/TopicCommand.scala
```diff
@@ -113,6 +114,7 @@ object TopicCommand extends Logging {
     def printDescription(): Unit = {
       val configsAsString = config.entries.asScala.filter(!_.isDefault).map { ce => s"${ce.name}=${ce.value}" }.mkString(",")
       print(s"Topic: $topic")
+      print(s"\tTopicId: ${if(topicId == Uuid.ZERO_UUID) "" else topicId}")
```
jolshan: Do we want to print an empty string for TopicId? Would it make sense to include the Zero Uuid or not print a topic ID label at all?

core/src/main/scala/kafka/api/ApiVersion.scala
```diff
@@ -424,6 +426,13 @@ case object KAFKA_2_7_IV2 extends DefaultApiVersion {
   val id: Int = 30
 }
 
+case object KAFKA_2_7_IV3 extends DefaultApiVersion {
```
jolshan: nit: this should be KAFKA_2_8_IV0, or KAFKA_2_8_IV1 if https://github.com/apache/kafka/pull/9601 merges

jolshan:
@dengziming Looks pretty good so far to me! I think it would be useful to write a few unit/integration tests to ensure the metadata snapshot behavior and describe topics work as expected.

clients/src/main/resources/common/message/MetadataRequest.json
```diff
@@ -31,9 +31,11 @@
     // Starting in version 8, authorized operations can be requested for cluster and topic resource.
     //
     // Version 9 is the first flexible version.
+    // Version 10 add topicId
     { "name": "Topics", "type": "[]MetadataRequestTopic", "versions": "0+", "nullableVersions": "1+",
       "about": "The topics to fetch metadata for.", "fields": [
-      { "name": "Name", "type": "string", "versions": "0+", "entityType": "topicName",
+      { "name": "TopicId", "type": "uuid", "versions": "10+", "ignorable": true, "about": "The topic id." },
+      { "name": "Name", "type": "string", "versions": "0+", "entityType": "topicName", "ignorable": true,
```
jolshan: As discussed in the mailing thread and the KIP writeup, this should be "nullable" rather than ignorable

clients/src/main/java/org/apache/kafka/common/Cluster.java
dengziming: Here, I try to use getOrDefault but don't know what the default value is(it seems that "" is not a good default topicName). so just return null and let the caller decide, suggestions are welcomed if you have good ideas.

clients/src/main/resources/common/message/MetadataRequest.json
dengziming: I tried to use nullable but got an Exceptioin with `Unrecognized field "nullable"` message.

clients/src/main/java/org/apache/kafka/common/Cluster.java
jolshan: I see...maybe I missed it, but I did not see the method used. What use case would you expect for topic names in Cluster?

dengziming: This is not used so far but will be used when we get TopicMetadata by topicId, so just add it in advance.

dengziming:
> @dengziming Looks pretty good so far to me! I think it would be useful to write a few unit/integration tests to ensure the metadata snapshot behavior and describe topics work as expected.

I added/altered some unit/integration tests in `MetadataRequestTest` and `ControllerChannelManagerTest`, take a look.

core/src/main/scala/kafka/log4j.properties
```diff
@@ -0,0 +1,91 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Unspecified loggers and loggers with additivity=true output to server.log and stdout
+# Note that INFO only applies to unspecified loggers, the log level of the child logger is used otherwise
+log4j.rootLogger=INFO, stdout, kafkaAppender
```
rajinisivaram: Was this file checked in by mistake? Can we remove?

core/src/main/scala/kafka/server/MetadataCache.scala
```diff
@@ -372,6 +380,8 @@ class MetadataCache(brokerId: Int) extends Logging {
   }
 
   case class MetadataSnapshot(partitionStates: mutable.AnyRefMap[String, mutable.LongMap[UpdateMetadataPartitionState]],
+                              topicIds: Map[String, Uuid],
+                              topicNames: Map[Uuid, String],
```
rajinisivaram: It seems unnecessary to include both topicIds and topicNames in the constructor of this case class. We can add `val topicNames` inside this class that creates the second map from the first.

core/src/test/scala/unit/kafka/controller/ControllerChannelManagerTest.scala
```diff
@@ -873,6 +883,11 @@ class ControllerChannelManagerTest {
       context.updatePartitionFullReplicaAssignment(partition, ReplicaAssignment(replicas))
       leaderIndex += 1
     }
+
+    context.allTopics ++= topics
+    for (topic <- topics if topicIds.contains(topic)) {
```
rajinisivaram: Couldn't we just do this instead of the for loop:
```
 topicIds.foreach { case (name, id) => context.addTopicId(name, id) }
```

clients/src/main/java/org/apache/kafka/common/Cluster.java
```diff
@@ -46,6 +46,8 @@
     private final Map<Integer, List<PartitionInfo>> partitionsByNode;
     private final Map<Integer, Node> nodesById;
     private final ClusterResource clusterResource;
+    private final Map<String, Uuid> topicIds;
+    private final Map<Uuid, String> topicNames;
```
rajinisivaram: Do we have a use case that requires both maps to be stored in Cluster for fast lookups? Or would it be sufficient to store it one-way and look up the single map for the other way?

core/src/main/scala/kafka/server/MetadataCache.scala
rajinisivaram: Also same question as for Cluster: Do we have a use case that requires both maps to be stored for fast lookups? Or would it be sufficient to store it one-way and look up the single map for the other way?

core/src/main/scala/kafka/server/MetadataCache.scala
```diff
@@ -314,9 +315,16 @@ class MetadataCache(brokerId: Int) extends Logging {
           error(s"Listeners are not identical across brokers: $aliveNodes")
       }
 
+      val newTopicIds = updateMetadataRequest.topicStates().asScala
+        .map(topicState => (topicState.topicName(), topicState.topicId()))
+        .filter(_._2 != Uuid.ZERO_UUID).toMap
+      val topicIds = mutable.Map.empty[String, Uuid]
+      topicIds.addAll(metadataSnapshot.topicIds)
+      topicIds.addAll(newTopicIds)
```
rajinisivaram: We seem to be adding to the topicIds/topicNames map, but never removing anything.

rajinisivaram (commented):
@dengziming Thanks for the PR, looks good. Left some comments.

core/src/test/scala/unit/kafka/controller/ControllerChannelManagerTest.scala
dengziming: done!

clients/src/main/java/org/apache/kafka/common/Cluster.java
dengziming: Thank you for your suggestions, the `topicNames` is not used so I just removed it. Maybe we will add them if it's necessary in a future use case.

core/src/main/scala/kafka/server/MetadataCache.scala
dengziming: Hi, @rajinisivaram ,Thank you for your reply. I find that when KafkaController deletes a topic it will send `updateMetadataRequest` with all the partitions in it, and set `state.leader=-2` indicating the partition is deleted. And the `MetadataCache` will not remove the topic from cache but just keep the `UpdateMetadataPartitionState` of the deleted topics, so I also keep the topicId when deleted.

here I tried to delete a topic and then describe it.

➜  ~ kafka-topics.sh --delete --zookeeper localhost:2181 --topic old-version-topic
Topic old-version-topic is marked for deletion.
Note: This will have no impact if delete.topic.enable is not set to true.

➜  ~ kafka-topics.sh --describe --zookeeper localhost:2181 --topic old-version-topic
Topic: old-version-topic TopicId: wRPl6VAlQeyE77bDxEESzg  PartitionCount: 1	ReplicationFactor: 1 Configs: 	MarkedForDeletion: true
	Topic: old-version-topic	Partition: 0	Leader: 0	Replicas: 0	Isr: 0	MarkedForDeletion: true

So the topicId of the deleted topic should also remain when deleted, when do you think we should delete the topicId from MetadataCache? suggestions are welcomed.

clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
```diff
@@ -320,6 +333,7 @@ public String toString() {
             return "TopicMetadata{" +
                 "error=" + error +
                 ", topic='" + topic + '\'' +
+                ", topicId='" + topicId.toString() + '\'' +
```
rajinisivaram: nit: toString() unnecessary

core/src/test/scala/unit/kafka/server/MetadataRequestTest.scala
```diff
@@ -223,6 +224,31 @@ class MetadataRequestTest extends BaseRequestTest {
...
+      assertEquals(Errors.NONE, topicMetadata.error)
+      assertEquals(Uuid.ZERO_UUID, topicMetadata.topicId())
+    }
+
+    // from version 10, UUID will be included in MetadataResponse
+    val resp2 = sendMetadataRequest(new MetadataRequest.Builder(Seq(topic1, topic2).asJava, true, 10, 10).build(), Some(notControllerSocketServer))
+    assertEquals(2, resp2.topicMetadata.size)
+    resp2.topicMetadata.forEach { topicMetadata =>
+      assertEquals(Errors.NONE, topicMetadata.error)
+      assertNotEquals(Uuid.ZERO_UUID, topicMetadata.topicId())
```
rajinisivaram, line 248: we probably also want to assert that the topic id is not null here (even though we currently never return null).

core/src/main/scala/kafka/server/MetadataCache.scala
rajinisivaram: When a topic is deleted, brokers process UpdateMetadataRequest and remove deleted topics from their cache. We track deletion state in ZooKeeper and as you mentioned, you can get this information by directly going to ZK in kafka-topics.sh. But we don't retain that information in every broker. I would remove topic id in the code segment just below this when the partition state is removed from the MetadataCache since we cannot clearly have a map that keeps growing in brokers. Is there a reason why we would want to retain topic id in every broker even after the topic has been deleted? We can't get this information through existing metadata request from brokers anyway. I guess in future, we can add additional metadata to track deleted topic ids if we wanted to, but for now it seems better to delete topic ids from MetadataCache when we delete the partition from the cache.  What do you think?

rajinisivaram (commented):
@dengziming Thanks for the updates, looks good, just a couple more comments left.

core/src/main/scala/kafka/server/MetadataCache.scala
jolshan: So it seems like there is some logic to remove the partition states of deleted topics from the MetadataSnapshot. Would we want to do something similar there but with topic Ids? Apologies if I'm missing something

rajinisivaram: Yes, that was my suggestion too, but my wording wasn't right. I will reword that to avoid confusion.

dengziming: Thank you for your suggestions, I now understand the solution. I remove the topicId in `removePartitionInfo`.

clients/src/main/java/org/apache/kafka/common/Cluster.java
```diff
@@ -327,6 +346,14 @@ public Node controller() {
         return controller;
     }
 
+    public Collection<Uuid> topicIds() {
+        return topicIds.values();
+    }
+
+    public Uuid getTopicId(String topic) {
```
rajinisivaram: nit: We don't use `get` prefix in other methods. Since this is part of the public API, can we change this to `topicId(String topic)` before merging?

rajinisivaram (commented):
@dengziming Thanks for the update, just one minor comment left, apart from that LGTM.

clients/src/main/java/org/apache/kafka/common/Cluster.java
dengziming: Thank you for your reminder, it's helpful to me.

rajinisivaram (approved):
@dengziming Thanks for the update, LGTM. Will review PR #9626 as well and merge these two together, hopefully later today.

rajinisivaram:
There is a compilation failure in the Java 8 build because the PR uses an API not available in Scala 2.12. Will push a fix.

rajinisivaram:
@dengziming Thanks for the PR, merging to trunk.

clients/src/main/java/org/apache/kafka/common/Cluster.java
jolshan: Can we also add this and any other new public APIs to the KIP page?

dengziming: Thank you for your reminder, I have added these 2 methods to the KIP page, take a look if you have time.

jolshan: Thanks. looks good!

PR #9473, opened 2020-10-21
KAFKA-10545: Create topic IDs in ZooKeeper and Controller

jolshan:
Topic IDs must be created for all new topics and all existing topics that do not yet have a topic ID. In ZooKeeper, the ID is written to the TopicZNode, and in the controller, it is stored in a map. 

This is a preliminary change before the second part of KAFKA-10545, which will propagate these IDs to brokers.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9626, opened 2020-11-19
KAFKA-10545: Create topic IDs and propagate to brokers

jolshan:
This change takes the topic IDs created in https://github.com/apache/kafka/pull/9473 and propagates them to brokers using LeaderAndIsr Request. It also removes the topic name from the LeaderAndIsr Response, reorganizes the response to be sorted by topic, and includes the topic ID.

In addition, the topic ID is persisted to each replica in Log as well as in a file on disk. This file is read on startup and if the topic ID exists, it will be reloaded. 

This PR bumps the IBP and is expected to be merged at the same time as https://github.com/apache/kafka/pull/9622 as to not bump the protocol twice


core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1445,15 +1483,38 @@ class ReplicaManager(val config: KafkaConfig,
...
+                .setPartitionIndex(tp.partition)
+                .setErrorCode(error.code)
+            }.toBuffer
+            new LeaderAndIsrResponse(new LeaderAndIsrResponseData()
+              .setErrorCode(Errors.NONE.code)
+              .setPartitionErrors(responsePartitions.asJava))
+          } else {
+            val topics = new mutable.HashMap[String, List[LeaderAndIsrPartitionError]]
+            responseMap.asJava.forEach { case (tp, error) =>
+              if (topics.get(tp.topic) == None) {
```
dengziming: can be changed into topics.contains()

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1445,15 +1483,38 @@ class ReplicaManager(val config: KafkaConfig,
...
-          val responsePartitions = responseMap.iterator.map { case (tp, error) =>
-            new LeaderAndIsrPartitionError()
-              .setTopicName(tp.topic)
-              .setPartitionIndex(tp.partition)
-              .setErrorCode(error.code)
-          }.toBuffer
-          new LeaderAndIsrResponse(new LeaderAndIsrResponseData()
-            .setErrorCode(Errors.NONE.code)
-            .setPartitionErrors(responsePartitions.asJava))
+          if (leaderAndIsrRequest.version() < 4) {
```
dengziming: Should here be version() < 5 ?

dengziming (commented):
ApiVersionTest need to be changed because you changed ApiVersion, and a few suggesions.

core/src/main/scala/kafka/server/ReplicaManager.scala
jolshan: Good catch. Thanks!

dengziming:
KAFKA-10729 add `KAFKA_2_8_IV0`, so you should bump the version again.

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrResponse.java
```diff
@@ -58,8 +68,17 @@ public Errors error() {
         Errors error = error();
         if (error != Errors.NONE)
             // Minor optimization since the top-level error applies to all partitions
-            return Collections.singletonMap(error, data.partitionErrors().size());
-        return errorCounts(data.partitionErrors().stream().map(l -> Errors.forCode(l.errorCode())));
+            if (data.topics().isEmpty()) {
+                return Collections.singletonMap(error, data.partitionErrors().size());
+            } else {
```
rite2nikhil: nit: is the else statement needed all it has is return?

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrRequest.java
```diff
@@ -138,14 +145,32 @@ public LeaderAndIsrResponse getErrorResponse(int throttleTimeMs, Throwable e) {
         Errors error = Errors.forException(e);
         responseData.setErrorCode(error.code());
 
-        List<LeaderAndIsrPartitionError> partitions = new ArrayList<>();
-        for (LeaderAndIsrPartitionState partition : partitionStates()) {
-            partitions.add(new LeaderAndIsrPartitionError()
-                .setTopicName(partition.topicName())
-                .setPartitionIndex(partition.partitionIndex())
-                .setErrorCode(error.code()));
+        if (version() < 5) {
```
rite2nikhil, line 139: may be i missed is there a test checking correctness of versioning ?

rite2nikhil (commented):
thanks for the pr, left a few commrnts

clients/src/main/resources/common/message/LeaderAndIsrRequest.json
```diff
@@ -21,8 +21,12 @@
   //
   // Version 2 adds broker epoch and reorganizes the partitions by topic.
   //
-  // Version 3 adds AddingReplicas and RemovingReplicas
-  "validVersions": "0-4",
+  // Version 3 adds AddingReplicas and RemovingReplicas.
+  //
+  // Version 4 is the first flexible version.
+  //
+  // Version 5 adds Topic ID to the TopicStates.
```
dajac: `Type` field is also added. I would also mention the KIP for reference.

clients/src/main/resources/common/message/LeaderAndIsrRequest.json
```diff
@@ -31,6 +35,8 @@
       "about": "The current controller epoch." },
     { "name": "BrokerEpoch", "type": "int64", "versions": "2+", "ignorable": true, "default": "-1",
       "about": "The current broker epoch." },
+    { "name":  "Type", "type":  "int8", "versions": "5+",
```
dajac: nit: There are two spaces before `Type` and `int8`.

clients/src/main/resources/common/message/LeaderAndIsrResponse.json
```diff
@@ -22,15 +22,28 @@
   // Version 2 is the same as version 1.
   //
   // Version 3 is the same as version 2.
-  "validVersions": "0-4",
+  //
+  // Version 4 is the first flexible version.
+  //
+  // Version 5 removes TopicName and replaces it with TopicId and and reorganizes the partitions by topic.
```
dajac: nit: There are two consecutive `and`.

clients/src/main/resources/common/message/LeaderAndIsrResponse.json
```diff
@@ -22,15 +22,28 @@
...
   "fields": [
     { "name": "ErrorCode", "type": "int16", "versions": "0+",
       "about": "The error code, or 0 if there was no error." },
-    { "name": "PartitionErrors", "type": "[]LeaderAndIsrPartitionError", "versions": "0+",
-      "about": "Each partition.", "fields": [
-      { "name": "TopicName", "type": "string", "versions": "0+", "entityType": "topicName",
-        "about": "The topic name." },
+    { "name": "PartitionErrors", "type": "[]LeaderAndIsrPartitionError", "versions": "0-4",
+      "about": "Each partition in v0 to v4 message."},
+    {"name":  "Topics", "type":  "[]LeaderAndIsrTopicError", "versions": "5+",
```
dajac: nit: There are two spaces before `Topics` and `[]LeaderAndIsrTopicError`. We could also add a space before `name` to remain consistent with the other fields.

clients/src/main/resources/common/message/LeaderAndIsrResponse.json
```diff
@@ -22,15 +22,28 @@
...
+      "about": "Each topic", "fields": [
+      { "name": "TopicId", "type": "uuid", "versions": "5+", "about": "The unique topic ID" },
+      { "name": "PartitionErrors", "type": "[]LeaderAndIsrPartitionError", "versions": "0+",
```
dajac: Should the version of `PartitionErrors` be `5+`?

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrRequest.java
```diff
@@ -138,14 +145,32 @@ public LeaderAndIsrResponse getErrorResponse(int throttleTimeMs, Throwable e) {
...
+                        .setErrorCode(error.code()));
+            }
+            responseData.setPartitionErrors(partitions);
+            return new LeaderAndIsrResponse(responseData);
+        }
+
+        List<LeaderAndIsrTopicError> topics = new ArrayList<>();
+        for (LeaderAndIsrTopicState topicState : data.topicStates()) {
+            LeaderAndIsrTopicError topicError = new LeaderAndIsrTopicError();
+            topicError.setTopicId(topicIds().get(topicState.topicName()));
```
dajac: `topicIds()` recomputes the `Map` so it would be better to keep a local reference to it.

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrRequest.java
```diff
@@ -138,14 +145,32 @@ public LeaderAndIsrResponse getErrorResponse(int throttleTimeMs, Throwable e) {
...
+                partitions.add(new LeaderAndIsrPartitionError()
+                        .setTopicName(partition.topicName())
+                        .setPartitionIndex(partition.partitionIndex())
+                        .setErrorCode(error.code()));
+            }
+            responseData.setPartitionErrors(partitions);
+            return new LeaderAndIsrResponse(responseData);
+        }
+
+        List<LeaderAndIsrTopicError> topics = new ArrayList<>();
```
dajac: nit: Could we directly allocate the `ArrayList` with the correct capacity? The same for `partitions` above and below.

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrResponse.java
```diff
@@ -45,8 +47,16 @@ public LeaderAndIsrResponse(Struct struct, short version) {
...
+    public List<LeaderAndIsrTopicError> topics() {
+        return this.data.topics();
+    }
+
+    public Iterable<LeaderAndIsrPartitionError> partitions() {
+        if (data.topics().isEmpty()) {
+            return data.partitionErrors();
+        }
+        return () -> new FlattenedIterator<>(data.topics().iterator(),
+            topic -> topic.partitionErrors().iterator());
```
dajac: It would be better to explicitly handle the version here instead of relying on `topics()` to be empty or not. It is easier to reason about for the reader and it also makes the handling very explicit instead of being implicit.

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrResponse.java
```diff
@@ -58,8 +68,17 @@ public Errors error() {
...
+                return Collections.singletonMap(error, data.partitionErrors().size());
+            } else {
+                return Collections.singletonMap(error,
+                        data.topics().stream().mapToInt(t -> t.partitionErrors().size()).sum());
+            }
+        if (data.topics().isEmpty()) {
+            return errorCounts(data.partitionErrors().stream().map(l -> Errors.forCode(l.errorCode())));
+        }
+        return errorCounts(data.topics().stream().flatMap(t -> t.partitionErrors().stream()).map(l ->
+                Errors.forCode(l.errorCode())));
```
dajac: ditto here. It would be better to be explicit wrt. the handling of the version.

clients/src/test/java/org/apache/kafka/common/requests/LeaderAndIsrRequestTest.java
```diff
@@ -51,15 +53,15 @@
...
+                Collections.emptyList(), Collections.emptyMap(), Collections.emptySet());
         assertThrows(UnsupportedVersionException.class, builder::build);
     }
 
     @Test
     public void testGetErrorResponse() {
         for (short version = LEADER_AND_ISR.oldestVersion(); version < LEADER_AND_ISR.latestVersion(); version++) {
             LeaderAndIsrRequest.Builder builder = new LeaderAndIsrRequest.Builder(version, 0, 0, 0,
-                    Collections.emptyList(), Collections.emptySet());
+                    Collections.emptyList(), Collections.emptyMap(), Collections.emptySet());
```
dajac: Shouldn't we verify that topic ids are correctly set in the generated response as well?

clients/src/test/java/org/apache/kafka/common/requests/LeaderAndIsrRequestTest.java
```diff
@@ -116,8 +118,13 @@ public void testVersionLogic() {
                 new Node(0, "host0", 9090),
                 new Node(1, "host1", 9091)
             );
+
+            HashMap<String, Uuid> topicIds = new HashMap<>();
```
dajac: nit: `HashMap<String, Uuid>` to `Map<String, Uuid>`. I have seen this in a couple of places.

clients/src/test/java/org/apache/kafka/common/requests/LeaderAndIsrResponseTest.java
```diff
@@ -57,29 +60,32 @@ public void testErrorCountsFromGetErrorResponse() {
             .setZkVersion(20)
             .setReplicas(Collections.singletonList(10))
             .setIsNew(false));
+        HashMap<String, Uuid> topicIds = new HashMap<>();
+        topicIds.put("foo", Uuid.randomUuid());
```
dajac: nit: You could use `Collections.singletonMap` here.

clients/src/test/java/org/apache/kafka/common/requests/LeaderAndIsrResponseTest.java
```diff
@@ -57,29 +60,32 @@ public void testErrorCountsFromGetErrorResponse() {
...
     public void testErrorCountsWithTopLevelError() {
-        List<LeaderAndIsrPartitionError> partitions = createPartitions("foo",
-            asList(Errors.NONE, Errors.NOT_LEADER_OR_FOLLOWER));
+        Uuid id = Uuid.randomUuid();
+        List<LeaderAndIsrTopicError> topics = createTopic(id, asList(Errors.NONE, Errors.NOT_LEADER_OR_FOLLOWER));
         LeaderAndIsrResponse response = new LeaderAndIsrResponse(new LeaderAndIsrResponseData()
             .setErrorCode(Errors.UNKNOWN_SERVER_ERROR.code())
-            .setPartitionErrors(partitions));
+            .setTopics(topics));
         assertEquals(Collections.singletonMap(Errors.UNKNOWN_SERVER_ERROR, 2), response.errorCounts());
```
dajac: Should we keep testing the older version as well? Tests assume the newest version only now.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -21,6 +21,7 @@
 import org.apache.kafka.common.IsolationLevel;
```
dajac, line 21: It would be good to verify that all versions are tested in `testSerialization`.

core/src/main/scala/kafka/controller/ControllerChannelManager.scala
```diff
@@ -482,8 +483,13 @@ abstract class AbstractControllerBrokerRequestBatch(config: KafkaConfig,
           _.node(config.interBrokerListenerName)
         }
         val brokerEpoch = controllerContext.liveBrokerIdAndEpochs(broker)
+        val topicIds = leaderAndIsrPartitionStates.keys
+          .map(_.topic)
+          .toSet
+          .map((topic: String) => (topic, controllerContext.topicIds(topic)))
```
dajac: nit: `(topic: String)` -> `topic`. We rarely specify the type in lambdas.

core/src/test/scala/unit/kafka/controller/ControllerChannelManagerTest.scala
```diff
@@ -72,6 +73,8 @@ class ControllerChannelManagerTest {
     assertEquals(1, updateMetadataRequests.size)
 
     val leaderAndIsrRequest = leaderAndIsrRequests.head
+    val topicIds = leaderAndIsrRequest.topicIds();
+    val topicNames = topicIds.asScala.map{ case (k,v) => (v, k)}
```
dajac: nit: We usually put a space before `{` and `}` when we use curly braces inline. We could also add a space after the coma.

core/src/test/scala/unit/kafka/controller/ControllerChannelManagerTest.scala
```diff
@@ -87,7 +90,10 @@ class ControllerChannelManagerTest {
     val LeaderAndIsrResponseReceived(leaderAndIsrResponse, brokerId) = batch.sentEvents.head
     assertEquals(2, brokerId)
     assertEquals(partitions.keySet,
-      leaderAndIsrResponse.partitions.asScala.map(p => new TopicPartition(p.topicName, p.partitionIndex)).toSet)
+      leaderAndIsrResponse.topics.asScala.map(t => t.partitionErrors.asScala.map(p =>
+        new TopicPartition(topicNames.get(t.topicId).get, p.partitionIndex))).flatMap(f => f).toSet)
```
dajac: * `topicNames.get(t.topicId).get` -> `topicNames(t.topicId)`. It is a bit more concise when you know that the Map contains what you are looking up.
* `flatMap(f => f)` looks weird. I suppose that we could use `flatMap` instead of the first `map`.

core/src/test/scala/unit/kafka/controller/ControllerChannelManagerTest.scala
```diff
@@ -818,15 +825,18 @@ class ControllerChannelManagerTest {
   private def applyLeaderAndIsrResponseCallbacks(error: Errors, sentRequests: List[SentRequest]): Unit = {
     sentRequests.filter(_.request.apiKey == ApiKeys.LEADER_AND_ISR).filter(_.responseCallback != null).foreach { sentRequest =>
       val leaderAndIsrRequest = sentRequest.request.build().asInstanceOf[LeaderAndIsrRequest]
-      val partitionErrors = leaderAndIsrRequest.partitionStates.asScala.map(p =>
-        new LeaderAndIsrPartitionError()
-          .setTopicName(p.topicName)
-          .setPartitionIndex(p.partitionIndex)
-          .setErrorCode(error.code))
+      val topicIds = leaderAndIsrRequest.topicIds()
+      val topicErrors = leaderAndIsrRequest.data.topicStates().asScala.map(t =>
```
dajac: nit: The parenthesis after `topicStates`, `partitionStates`, and `partitionIndex` are not mandatory. We tend to not put it when they are not.

core/src/test/scala/unit/kafka/controller/ControllerChannelManagerTest.scala
```diff
@@ -157,7 +163,8 @@ class ControllerChannelManagerTest {
 
     for (apiVersion <- ApiVersion.allVersions) {
       val leaderAndIsrRequestVersion: Short =
-        if (apiVersion >= KAFKA_2_4_IV1) 4
+        if (apiVersion >= KAFKA_2_8_IV0) 5
```
dajac: I wonder if we should extend `testLeaderAndIsrRequestFollowsInterBrokerProtocolVersion` to verifies the topic ids based on the different supported versions. What do you think?

core/src/main/scala/kafka/controller/KafkaController.scala
```diff
@@ -1378,12 +1378,26 @@ class KafkaController(val config: KafkaConfig,
     val offlineReplicas = new ArrayBuffer[TopicPartition]()
     val onlineReplicas = new ArrayBuffer[TopicPartition]()
 
-    leaderAndIsrResponse.partitions.forEach { partition =>
-      val tp = new TopicPartition(partition.topicName, partition.partitionIndex)
-      if (partition.errorCode == Errors.KAFKA_STORAGE_ERROR.code)
-        offlineReplicas += tp
-      else if (partition.errorCode == Errors.NONE.code)
-        onlineReplicas += tp
+    if (leaderAndIsrResponse.topics().isEmpty) {
```
dajac: It may be better to also have an explicit handling of the version here. Alternatively, we could push this into the `LeaderAndIsrResponse` and provides a method `Map<TopicPartition, ...> partitions()` which handles the version.

core/src/main/scala/kafka/controller/KafkaController.scala
```diff
@@ -1378,12 +1378,26 @@ class KafkaController(val config: KafkaConfig,
...
+        val tp = new TopicPartition(topicName, partition.partitionIndex)
+        if (partition.errorCode == Errors.KAFKA_STORAGE_ERROR.code)
+          offlineReplicas += tp
+        else if (partition.errorCode == Errors.NONE.code)
+          onlineReplicas += tp
+      }
+    }
+
+    leaderAndIsrResponse.topics.forEach { topic =>
+      val topicName = controllerContext.topicNames.get(topic.topicId).get
```
dajac: Do we need to handle the case when the topic may not be there anymore when the response is received? If not, we could use `controllerContext.topicNames(topic.topicId)`.

dajac (commented):
Thanks for the PR. I have left some comments. I have only looked at the request/response and the controller so far. I will look at the remaining parts tomorrow.

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrResponse.java
jolshan: I agree. I think I had some problems with defining the version when the constructor only provides the data. Is there a way to get the version with just the data?

clients/src/main/resources/common/message/LeaderAndIsrRequest.json
jolshan: Thanks for catching that. Will add the KIP too.

core/src/main/scala/kafka/controller/KafkaController.scala
dajac: Actually, this may not work as we don't have the topic name in the latest version...

core/src/test/scala/unit/kafka/controller/ControllerChannelManagerTest.scala
jolshan: I thought I removed most of the get(topic).gets, but I guess I missed a few. Thanks for catching. And I agree about flatMap as well.

core/src/test/scala/unit/kafka/controller/ControllerChannelManagerTest.scala
jolshan: That's a good idea

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrResponse.java
jolshan: I'm thinking that it most cases where I create a LeaderAndIsrResponse with the data object, I have access to the request and can grab the version there. Then I can add it to the constructor.

core/src/main/scala/kafka/controller/KafkaController.scala
jolshan: Yeah. One option I thought of would be to do something like partitions(Map<id, string> topicNames) and handle it inside the response with the version. That might be a little cleaner but I'm not sure.

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrRequest.java
jolshan: There is LeaderAndIsrResponseTest.java (which doesn't try all versions) and testGetErrorResponse() in LeaderAndIsrRequestTest.java that does test all versions. I will add version tests to the former.

core/src/main/scala/kafka/controller/KafkaController.scala
jolshan: Oops. Looks like I missed another one.

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrRequest.java
jolshan: So the partition above uses an iterable, and I'm not sure if there is a way to grab the size without iterating through. Let me know if I'm forgetting something.

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrRequest.java
```diff
@@ -171,6 +196,14 @@ public long brokerEpoch() {
         return data.ungroupedPartitionStates();
     }
 
+    public Map<String, Uuid> topicIds() {
+        Map<String, Uuid> topicIds = new HashMap<>();
+        for (LeaderAndIsrTopicState ts : data.topicStates()) {
+            topicIds.put(ts.topicName(), ts.topicId());
+        }
+        return topicIds;
```
rajinisivaram: Should be able to replace this with something like:
```
return data.topicStates().stream()
  .collect(Collectors.toMap(LeaderAndIsrTopicState::topicName, LeaderAndIsrTopicState::topicId));
```

clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrResponse.java
```diff
@@ -58,8 +68,17 @@ public Errors error() {
         Errors error = error();
         if (error != Errors.NONE)
```
rajinisivaram: nit: We should add braces here since there are multiple lines inside the `if` statement

rajinisivaram: nit: We should add braces here since there are multiple lines inside the `if` statement

core/src/main/scala/kafka/controller/KafkaController.scala
rajinisivaram: Should we do `controllerContext.topicNames.get(topic.topicId).foreach` instead of get to avoid throwing exception if topic is not in the context?

core/src/main/scala/kafka/controller/KafkaController.scala
```diff
@@ -1378,12 +1378,26 @@ class KafkaController(val config: KafkaConfig,
...
+      topic.partitionErrors().forEach { partition =>
+        val tp = new TopicPartition(topicName, partition.partitionIndex)
+        if (partition.errorCode == Errors.KAFKA_STORAGE_ERROR.code)
+          offlineReplicas += tp
+        else if (partition.errorCode == Errors.NONE.code)
+          onlineReplicas += tp
```
rajinisivaram: We can make an inner method within this method with the common logic.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -322,6 +327,11 @@ class Log(@volatile private var _dir: File,
     // deletion.
     producerStateManager.removeStraySnapshots(segments.values().asScala.map(_.baseOffset).toSeq)
     loadProducerState(logEndOffset, reloadFromCleanShutdown = hadCleanShutdown)
+
+    // Recover topic ID if present
+    if (!partitionMetadataFile.get.isEmpty()) {
+      topicId = partitionMetadataFile.get.read().topicId
+    }
```
rajinisivaram, line 337: Seems neater to use `partitionMetadataFile.foreach` instead of using `.get` twice.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1021,6 +1036,13 @@ class Log(@volatile private var _dir: File,
           // re-initialize leader epoch cache so that LeaderEpochCheckpointFile.checkpoint can correctly reference
           // the checkpoint file in renamed log directory
           initializeLeaderEpochCache()
+          if (!partitionMetadataFile.isEmpty && !partitionMetadataFile.get.isEmpty()) {
+            val partitionMetadata = partitionMetadataFile.get.read()
+            initializePartitionMetadata()
+            partitionMetadataFile.get.write(partitionMetadata.topicId)
```
rajinisivaram: Why are we reading and writing again?

core/src/main/scala/kafka/server/PartitionMetadataFile.scala
```diff
@@ -0,0 +1,141 @@
...
+
+import kafka.utils.Logging
+import org.apache.kafka.common.Uuid
+import org.apache.kafka.common.errors.KafkaStorageException
+import org.apache.kafka.common.utils.Utils
+
+
+
+object PartitionMetadataFile {
+  private val LeaderEpochCheckpointFilename = "partition.metadata"
```
rajinisivaram: `LeaderEpochCheckpointFilename` => `PartitionMetadataFilename`?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1021,6 +1036,13 @@ class Log(@volatile private var _dir: File,
           // re-initialize leader epoch cache so that LeaderEpochCheckpointFile.checkpoint can correctly reference
           // the checkpoint file in renamed log directory
           initializeLeaderEpochCache()
+          if (!partitionMetadataFile.isEmpty && !partitionMetadataFile.get.isEmpty()) {
```
rajinisivaram: `.foreach` may be better than lots of .`get`s

core/src/main/scala/kafka/server/PartitionMetadataFile.scala
```diff
@@ -0,0 +1,141 @@
...
+          case Array(_, version) =>
+            if (version.toInt == CurrentVersion) {
+              line = reader.readLine()
+              WhiteSpacesPattern.split(line) match {
+                case Array(_, topicId) => metadataTopicId = Uuid.fromString(topicId)
+                case _ => throw malformedLineException(line)
+              }
+              new PartitionMetadata(CurrentVersion, metadataTopicId)
+            } else {
+              throw new IOException(s"Unrecognized version of the checkpoint file ($location): " + version)
```
rajinisivaram: update error message?

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1434,6 +1447,31 @@ class ReplicaManager(val config: KafkaConfig,
...
+                    s" match the topic Id provided in the request: " +
+                    s"${topicIds.get(topicPartition.topic).toString}.")
+                } else {
+                  // There is not yet a topic ID stored in the log.
+                  // Write the partition metadata file if it is empty.
+                  if (log.partitionMetadataFile.get.isEmpty()) {
+                    log.partitionMetadataFile.get.write(topicIds.get(topicPartition.topic))
+                    log.topicId = topicIds.get(topicPartition.topic)
+                  } else {
+                    stateChangeLogger.warn("Partition metadata file already contains content.")
```
rajinisivaram: why is this a warning?

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1434,6 +1447,31 @@ class ReplicaManager(val config: KafkaConfig,
...
+                // If the topic ID in the log does not match the one in the request, the broker's topic must be stale.
+                if (!log.topicId.equals(Uuid.ZERO_UUID) && !log.topicId.equals(topicIds.get(topicPartition.topic))) {
+                  stateChangeLogger.warn(s"Topic Id in memory: ${log.topicId.toString} does not" +
+                    s" match the topic Id provided in the request: " +
+                    s"${topicIds.get(topicPartition.topic).toString}.")
+                } else {
+                  // There is not yet a topic ID stored in the log.
+                  // Write the partition metadata file if it is empty.
+                  if (log.partitionMetadataFile.get.isEmpty()) {
+                    log.partitionMetadataFile.get.write(topicIds.get(topicPartition.topic))
```
rajinisivaram: we already have the topic id in `id`.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1434,6 +1447,31 @@ class ReplicaManager(val config: KafkaConfig,
...
+                    log.topicId = topicIds.get(topicPartition.topic)
```
rajinisivaram: As above, we can use `id`

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1445,15 +1483,38 @@ class ReplicaManager(val config: KafkaConfig,
...
+            val topics = new mutable.HashMap[String, List[LeaderAndIsrPartitionError]]
+            responseMap.asJava.forEach { case (tp, error) =>
+              if (!topics.contains(tp.topic)) {
+                topics.put(tp.topic, List(new LeaderAndIsrPartitionError()
+                                                                .setPartitionIndex(tp.partition)
+                                                                .setErrorCode(error.code)))
+              } else {
+                topics.put(tp.topic, new LeaderAndIsrPartitionError()
+                  .setPartitionIndex(tp.partition)
+                  .setErrorCode(error.code)::topics.get(tp.topic).get)
```
rajinisivaram: Replace `topics.get(tp.topic).get` with `topics(tp.topic)`

rajinisivaram (commented):
@jolshan Thanks for the PR, looks good. Left some minor comments.

core/src/main/scala/kafka/controller/KafkaController.scala
jolshan: I ended up moving some of this to the response class so the duplicate logic is removed. I just haven't pushed yet

core/src/main/scala/kafka/server/PartitionMetadataFile.scala
jolshan: Good catch.

core/src/main/scala/kafka/log/Log.scala
jolshan: Perhaps my understading is wrong, but if we rename the directory, do we need to create the file in the new directory? With the same topic ID and such as before.

core/src/main/scala/kafka/server/ReplicaManager.scala
jolshan: I think that if we reach here, we are in an unexpected state. The partitionMetadata file should have loaded in the topic ID if the file already contained content. I left as a warn rather than an exception.

core/src/main/scala/kafka/log/Log.scala
jolshan: Oh I did understand wrong. The file is already in place, so initializePartitionMetadata is all that is needed to get the correct file path.

core/src/main/scala/kafka/controller/KafkaController.scala
jolshan: I moved this logic to LeaderAndIsrResponse which is java code, so it behaves a bit differently. (We may want to move back though, and if so, I'll do this.) For now I do a null check in the java code. 

One question I have though is what does it mean to not have a topic in the context? Is it possible to send a leaderAndIsrRequest with an unknown partition/one not in the context? I can only think maybe the topic gets deleted while handling the request. What would the normal error handling case be in that scenario?

core/src/main/scala/kafka/server/ReplicaManager.scala
rajinisivaram: Hmm, looking at the conditional statements here, it looks like we would write the file the first time we get here because `log.partitionMetadataFile.get.isEmpty()` and the second time we would print a warning even if the id in the file matches the expected id. Unless I missed something.

rajinisivaram (commented):
@jolshan Thanks for the updates, looks good, just one comment/question left. The changes to process errors in LeaderAndIsrResponse.java seem fine. As you said, the only case where topic is not in the controller context would be where a topic was deleted. Don't think we need any other special handling for that. The PR needs rebasing to resolve conflicts btw.

jolshan:
@rajinisivaram Yup, will look at the rebase next.

core/src/main/scala/kafka/server/ReplicaManager.scala
jolshan: Oops. I think I cleaned up this block and deleted something. There should be a check if log.topicId.equals(id). If so, then the file exists and we shouldn't go in to the block that says "// There is not yet a topic ID stored in the log."

I should also fix the topicIds.get(topicPartition.topic) above and replace with id.

jolshan: I might think about making this code cleaner in general to avoid so many nested if statements

jolshan: So I thought through these cases some more and realized that the metadata file will fail to open if formatted incorrectly. So the only case where there could be data written to the file is if the ID is the zero UUID. So I decided to just fail on reading the file if the zero ID is provided. (We will never write zero ID to file.) The rest of this cleaned up pretty nicely.

core/src/test/scala/unit/kafka/server/ReplicaManagerTest.scala
```diff
@@ -2176,4 +2212,84 @@ class ReplicaManagerTest {
...
+          .setIsr(brokerList)
+          .setZkVersion(0)
+          .setReplicas(brokerList)
+          .setIsNew(true)).asJava,
+        topicIds,
+        Set(new Node(0, "host1", 0), new Node(1, "host2", 1)).asJava).build()
+
+      // The file has no contents if the topic does not have an associated topic ID.
+      replicaManager.becomeLeaderOrFollower(0, leaderAndIsrRequest(0, "fakeTopic"), (_, _) => ())
+      assertTrue(!replicaManager.localLog(topicPartition).isEmpty)
```
rajinisivaram: nit: `assertFalse` instead of `assertTrue(!...)`. There are a few of these below as well.

rajinisivaram (commented):
@jolshan Thanks for the updates. Left a minor comment, apart from that LGTM. PR #9622 is also ready to merge, so we can merge these two together later today.

rajinisivaram:
@jolshan Thanks for the PR, the last build was good, so merging to trunk

PR #9628, opened 2020-11-20
KAFKA-10747: Implement APIs for altering and describing IP connection rate quotas

splett2:
This PR adds support for IP entities to the `DescribeClientQuotas` and `AlterClientQuotas` APIs. This change does not require any protocol version bumps. 

This PR also adds support for describing/altering IP quotas via `kafka-configs` tooling.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/quota/ClientQuotaEntity.java
+++ b/clients/src/main/java/org/apache/kafka/common/quota/ClientQuotaEntity.java
@@ -32,7 +32,7 @@ public class ClientQuotaEntity {
      */
     public static final String USER = "user";
     public static final String CLIENT_ID = "client-id";
-    public static final String IP = "IP";
+    public static final String IP = "ip";
 
     /**
      * Constructs a quota entity for the given types and names. If a name is null,
--- a/clients/src/main/java/org/apache/kafka/common/quota/ClientQuotaEntity.java
+++ b/clients/src/main/java/org/apache/kafka/common/quota/ClientQuotaEntity.java
@@ -32,6 +32,7 @@ public class ClientQuotaEntity {
      */
     public static final String USER = "user";
     public static final String CLIENT_ID = "client-id";
+    public static final String IP = "IP";
 
     /**
      * Constructs a quota entity for the given types and names. If a name is null,
--- a/core/src/main/scala/kafka/admin/ConfigCommand.scala
+++ b/core/src/main/scala/kafka/admin/ConfigCommand.scala
@@ -755,7 +755,7 @@ object ConfigCommand extends Config {
     val entityName = parser.accepts("entity-name", "Name of entity (topic name/client id/user principal name/broker id/ip)")
             .withRequiredArg
             .ofType(classOf[String])
-    val entityDefault = parser.accepts("entity-default", "Default entity name for clients/users/brokers (applies to corresponding entity type in command line)")
+    val entityDefault = parser.accepts("entity-default", "Default entity name for clients/users/brokers/ips (applies to corresponding entity type in command line)")
 
     val nl = System.getProperty("line.separator")
     val addConfig = parser.accepts("add-config", "Key Value pairs of configs to add. Square brackets can be used to group values which contain commas: 'k1=v1,k2=[v1,v2,v2],k3=v3'. The following is a list of valid configurations: " +
@@ -901,7 +901,7 @@ object ConfigCommand extends Config {
             entityTypeVals.contains(ConfigType.Broker) ||
             entityTypeVals.contains(ConfigType.Ip)) {
           if (!hasEntityName && !hasEntityDefault)
-            throw new IllegalArgumentException("an entity-name or default entity must be specified with --alter of users, clients or brokers")
+            throw new IllegalArgumentException("an entity-name or default entity must be specified with --alter of users, clients, brokers or ips")
         } else if (!hasEntityName)
           throw new IllegalArgumentException(s"an entity name must be specified with --alter of ${entityTypeVals.mkString(",")}")
 
--- a/core/src/main/scala/kafka/admin/ConfigCommand.scala
+++ b/core/src/main/scala/kafka/admin/ConfigCommand.scala
@@ -886,7 +886,10 @@ object ConfigCommand extends Config {
       }
 
       if (hasEntityName && entityTypeVals.contains(ConfigType.Ip)) {
-        Seq(entityName, ip).filter(options.has(_)).map(options.valueOf(_)).foreach(DynamicConfig.Ip.validateIpOrHost)
+        Seq(entityName, ip).filter(options.has(_)).map(options.valueOf(_)).foreach { ipEntity =>
+          if (!DynamicConfig.Ip.isValidIpEntity(ipEntity))
+            throw new IllegalArgumentException(s"The entity name for ${entityTypeVals.head} must be a valid IP or resolvable host, but it is: $ipEntity")
+        }
       }
 
       if (options.has(describeOpt) && entityTypeVals.contains(BrokerLoggerConfigType) && !hasEntityName)
--- a/core/src/main/scala/kafka/admin/ConfigCommand.scala
+++ b/core/src/main/scala/kafka/admin/ConfigCommand.scala
@@ -886,10 +886,7 @@ object ConfigCommand extends Config {
       }
 
       if (hasEntityName && entityTypeVals.contains(ConfigType.Ip)) {
-        Seq(entityName, ip).filter(options.has(_)).map(options.valueOf(_)).foreach { ipAddress =>
-          if (!Utils.validHostPattern(ipAddress))
-            throw new IllegalArgumentException(s"The entity name for ${entityTypeVals.head} must be a valid IP, but it is: $ipAddress")
-        }
+        Seq(entityName, ip).filter(options.has(_)).map(options.valueOf(_)).foreach(DynamicConfig.Ip.validateIpOrHost)
       }
 
       if (options.has(describeOpt) && entityTypeVals.contains(BrokerLoggerConfigType) && !hasEntityName)
--- a/core/src/main/scala/kafka/admin/ConfigCommand.scala
+++ b/core/src/main/scala/kafka/admin/ConfigCommand.scala
@@ -46,7 +46,7 @@ import scala.jdk.CollectionConverters._
 import scala.collection._
 
 /**
- * This script can be used to change configs for topics/clients/users/brokers/IPs dynamically
+ * This script can be used to change configs for topics/clients/users/brokers/ips dynamically
  * An entity described or altered by the command may be one of:
  * <ul>
  *     <li> topic: --topic <topic> OR --entity-type topics --entity-name <topic>
@@ -56,12 +56,12 @@ import scala.collection._
  *                          --entity-type users --entity-name <user-principal> --entity-type clients --entity-name <client-id>
  *     <li> broker: --broker <broker-id> OR --entity-type brokers --entity-name <broker-id>
  *     <li> broker-logger: --broker-logger <broker-id> OR --entity-type broker-loggers --entity-name <broker-id>
- *     <li> ip: --IP <ip> OR --entity-type IPs --entity-name <ip>
+ *     <li> ip: --ip <ip> OR --entity-type ips --entity-name <ip>
  * </ul>
- * --entity-type <users|clients|brokers|IPs> --entity-default may be specified in place of --entity-type <users|clients|brokers|IPs> --entity-name <entityName>
- * when describing or altering default configuration for users, clients, brokers, or IPs, respectively.
- * Alternatively, --user-defaults, --client-defaults, --broker-defaults, or --IP-defaults may be specified in place of
- * --entity-type <users|clients|brokers|IPs> --entity-default, respectively.
+ * --entity-type <users|clients|brokers|ips> --entity-default may be specified in place of --entity-type <users|clients|brokers|ips> --entity-name <entityName>
+ * when describing or altering default configuration for users, clients, brokers, or ips, respectively.
+ * Alternatively, --user-defaults, --client-defaults, --broker-defaults, or --ip-defaults may be specified in place of
+ * --entity-type <users|clients|brokers|ips> --entity-default, respectively.
  */
 object ConfigCommand extends Config {
 
@@ -85,7 +85,7 @@ object ConfigCommand extends Config {
     try {
       val opts = new ConfigCommandOptions(args)
 
-      CommandLineUtils.printHelpAndExitIfNeeded(opts, "This tool helps to manipulate and describe entity config for a topic, client, user, broker or IP")
+      CommandLineUtils.printHelpAndExitIfNeeded(opts, "This tool helps to manipulate and describe entity config for a topic, client, user, broker or ip")
 
       opts.checkArgs()
 
@@ -567,7 +567,7 @@ object ConfigCommand extends Config {
           val typeStr = entityType match {
             case ClientQuotaEntity.USER => "user-principal"
             case ClientQuotaEntity.CLIENT_ID => "client-id"
-            case ClientQuotaEntity.IP => "IP"
+            case ClientQuotaEntity.IP => "ip"
           }
           if (name != null) s"$typeStr '$name'"
           else s"the default $typeStr"
@@ -749,10 +749,10 @@ object ConfigCommand extends Config {
     val describeOpt = parser.accepts("describe", "List configs for the given entity.")
     val allOpt = parser.accepts("all", "List all configs for the given topic, broker, or broker-logger entity (includes static configuration when the entity type is brokers)")
 
-    val entityType = parser.accepts("entity-type", "Type of entity (topics/clients/users/brokers/broker-loggers/IPs)")
+    val entityType = parser.accepts("entity-type", "Type of entity (topics/clients/users/brokers/broker-loggers/ips)")
             .withRequiredArg
             .ofType(classOf[String])
-    val entityName = parser.accepts("entity-name", "Name of entity (topic name/client id/user principal name/broker id/IP)")
+    val entityName = parser.accepts("entity-name", "Name of entity (topic name/client id/user principal name/broker id/ip)")
             .withRequiredArg
             .ofType(classOf[String])
     val entityDefault = parser.accepts("entity-default", "Default entity name for clients/users/brokers (applies to corresponding entity type in command line)")
@@ -793,8 +793,8 @@ object ConfigCommand extends Config {
     val brokerLogger = parser.accepts("broker-logger", "The broker's ID for its logger config.")
       .withRequiredArg
       .ofType(classOf[String])
-    val ipDefaults = parser.accepts("IP-defaults", "The config defaults for all IPs.")
-    val ip = parser.accepts("IP", "The IP address.")
+    val ipDefaults = parser.accepts("ip-defaults", "The config defaults for all IPs.")
+    val ip = parser.accepts("ip", "The IP address.")
       .withRequiredArg
       .ofType(classOf[String])
     val zkTlsConfigFile = parser.accepts("zk-tls-config-file",
--- a/core/src/main/scala/kafka/admin/ConfigCommand.scala
+++ b/core/src/main/scala/kafka/admin/ConfigCommand.scala
@@ -46,7 +46,7 @@ import scala.jdk.CollectionConverters._
 import scala.collection._
 
 /**
- * This script can be used to change configs for topics/clients/users/brokers dynamically
+ * This script can be used to change configs for topics/clients/users/brokers/IPs dynamically
  * An entity described or altered by the command may be one of:
  * <ul>
  *     <li> topic: --topic <topic> OR --entity-type topics --entity-name <topic>
@@ -56,11 +56,12 @@ import scala.collection._
  *                          --entity-type users --entity-name <user-principal> --entity-type clients --entity-name <client-id>
  *     <li> broker: --broker <broker-id> OR --entity-type brokers --entity-name <broker-id>
  *     <li> broker-logger: --broker-logger <broker-id> OR --entity-type broker-loggers --entity-name <broker-id>
+ *     <li> ip: --IP <ip> OR --entity-type IPs --entity-name <ip>
  * </ul>
- * --entity-type <users|clients|brokers> --entity-default may be specified in place of --entity-type <users|clients|brokers> --entity-name <entityName>
- * when describing or altering default configuration for users, clients, or brokers, respectively.
- * Alternatively, --user-defaults, --client-defaults, or --broker-defaults may be specified in place of
- * --entity-type <users|clients|brokers> --entity-default, respectively.
+ * --entity-type <users|clients|brokers|IPs> --entity-default may be specified in place of --entity-type <users|clients|brokers|IPs> --entity-name <entityName>
+ * when describing or altering default configuration for users, clients, brokers, or IPs, respectively.
+ * Alternatively, --user-defaults, --client-defaults, --broker-defaults, or --IP-defaults may be specified in place of
+ * --entity-type <users|clients|brokers|IPs> --entity-default, respectively.
  */
 object ConfigCommand extends Config {
 
@@ -84,7 +85,7 @@ object ConfigCommand extends Config {
     try {
       val opts = new ConfigCommandOptions(args)
 
-      CommandLineUtils.printHelpAndExitIfNeeded(opts, "This tool helps to manipulate and describe entity config for a topic, client, user or broker")
+      CommandLineUtils.printHelpAndExitIfNeeded(opts, "This tool helps to manipulate and describe entity config for a topic, client, user, broker or IP")
 
       opts.checkArgs()
 
@@ -401,7 +402,11 @@ object ConfigCommand extends Config {
             throw new IllegalStateException(s"Altering user SCRAM credentials should never occur for more zero or multiple users: $entityNames")
           alterUserScramCredentialConfigs(adminClient, entityNames.head, scramConfigsToAddMap, scramConfigsToDelete)
         }
-
+      case ConfigType.Ip =>
+        val unknownConfigs = (configsToBeAdded.keys ++ configsToBeDeleted).filterNot(key => DynamicConfig.Ip.names.contains(key))
+        if (unknownConfigs.nonEmpty)
+          throw new IllegalArgumentException(s"Only connection quota configs can be added for '${ConfigType.Ip}' using --bootstrap-server. Unexpected config names: ${unknownConfigs.mkString(",")}")
+        alterQuotaConfigs(adminClient, entityTypes, entityNames, configsToBeAddedMap, configsToBeDeleted)
       case _ => throw new IllegalArgumentException(s"Unsupported entity type: $entityTypeHead")
     }
 
@@ -443,12 +448,11 @@ object ConfigCommand extends Config {
     if (invalidConfigs.nonEmpty)
       throw new InvalidConfigurationException(s"Invalid config(s): ${invalidConfigs.mkString(",")}")
 
-    val alterEntityTypes = entityTypes.map { entType =>
-      entType match {
-        case ConfigType.User => ClientQuotaEntity.USER
-        case ConfigType.Client => ClientQuotaEntity.CLIENT_ID
-        case _ => throw new IllegalArgumentException(s"Unexpected entity type: ${entType}")
-      }
+    val alterEntityTypes = entityTypes.map {
+      case ConfigType.User => ClientQuotaEntity.USER
+      case ConfigType.Client => ClientQuotaEntity.CLIENT_ID
+      case ConfigType.Ip => ClientQuotaEntity.IP
+      case entType => throw new IllegalArgumentException(s"Unexpected entity type: $entType")
     }
     val alterEntityNames = entityNames.map(en => if (en.nonEmpty) en else null)
 
@@ -461,7 +465,7 @@ object ConfigCommand extends Config {
     val alterOps = (configsToBeAddedMap.map { case (key, value) =>
       val doubleValue = try value.toDouble catch {
         case _: NumberFormatException =>
-          throw new IllegalArgumentException(s"Cannot parse quota configuration value for ${key}: ${value}")
+          throw new IllegalArgumentException(s"Cannot parse quota configuration value for $key: $value")
       }
       new ClientQuotaAlteration.Op(key, doubleValue)
     } ++ configsToBeDeleted.map(key => new ClientQuotaAlteration.Op(key, null))).asJavaCollection
@@ -480,6 +484,8 @@ object ConfigCommand extends Config {
         describeResourceConfig(adminClient, entityTypes.head, entityNames.headOption, describeAll)
       case ConfigType.User | ConfigType.Client =>
         describeClientQuotaAndUserScramCredentialConfigs(adminClient, entityTypes, entityNames)
+      case ConfigType.Ip =>
+        describeQuotaConfigs(adminClient, entityTypes, entityNames)
       case entityType => throw new IllegalArgumentException(s"Invalid entity type: $entityType")
     }
   }
@@ -551,7 +557,7 @@ object ConfigCommand extends Config {
       }).toSeq
   }
 
-  private def describeClientQuotaAndUserScramCredentialConfigs(adminClient: Admin, entityTypes: List[String], entityNames: List[String]) = {
+  private def describeQuotaConfigs(adminClient: Admin, entityTypes: List[String], entityNames: List[String]) = {
     val quotaConfigs = getAllClientQuotasConfigs(adminClient, entityTypes, entityNames)
     quotaConfigs.forKeyValue { (entity, entries) =>
       val entityEntries = entity.entries.asScala
@@ -561,15 +567,22 @@ object ConfigCommand extends Config {
           val typeStr = entityType match {
             case ClientQuotaEntity.USER => "user-principal"
             case ClientQuotaEntity.CLIENT_ID => "client-id"
+            case ClientQuotaEntity.IP => "IP"
           }
-          if (name != null) s"${typeStr} '${name}'"
-          else s"the default ${typeStr}"
+          if (name != null) s"$typeStr '$name'"
+          else s"the default $typeStr"
         }
 
-      val entityStr = (entitySubstr(ClientQuotaEntity.USER) ++ entitySubstr(ClientQuotaEntity.CLIENT_ID)).mkString(", ")
+      val entityStr = (entitySubstr(ClientQuotaEntity.USER) ++
+                       entitySubstr(ClientQuotaEntity.CLIENT_ID) ++
+                       entitySubstr(ClientQuotaEntity.IP)).mkString(", ")
       val entriesStr = entries.asScala.map(e => s"${e._1}=${e._2}").mkString(", ")
-      println(s"Quota configs for ${entityStr} are ${entriesStr}")
+      println(s"Quota configs for $entityStr are $entriesStr")
     }
+  }
+
+  private def describeClientQuotaAndUserScramCredentialConfigs(adminClient: Admin, entityTypes: List[String], entityNames: List[String]) = {
+    describeQuotaConfigs(adminClient, entityTypes, entityNames)
     // we describe user SCRAM credentials only when we are not describing client information
     // and we are not given either --entity-default or --user-defaults
     if (!entityTypes.contains(ConfigType.Client) && !entityNames.contains("")) {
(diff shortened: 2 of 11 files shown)


clients/src/main/java/org/apache/kafka/common/quota/ClientQuotaEntity.java
```diff
@@ -32,6 +32,7 @@
      */
     public static final String USER = "user";
     public static final String CLIENT_ID = "client-id";
+    public static final String IP = "IP";
```
bdbyrne: Consider making lower-case for consistency, but otherwise not a big issue, as I see that's how it was proposed in the KIP.

core/src/main/scala/kafka/server/AdminManager.scala
```diff
@@ -833,45 +835,76 @@ class AdminManager(val config: KafkaConfig,
...
         case "" =>
           throw new InvalidRequestException(s"Unexpected empty filter component entity type")
         case et =>
           // Supplying other entity types is not yet supported.
           throw new UnsupportedVersionException(s"Custom entity type '${et}' not supported")
       }
     }
-    handleDescribeClientQuotas(userComponent, clientIdComponent, filter.strict)
+    if ((userComponent.isDefined || clientIdComponent.isDefined) && ipComponent.isDefined)
+      throw new InvalidRequestException(s"Invalid entity filter component combination")
```
bdbyrne: May want to provide a little more detail in the response

core/src/test/scala/unit/kafka/utils/TestUtils.scala
```diff
@@ -1529,6 +1530,16 @@ object TestUtils extends Logging {
     adminClient.incrementalAlterConfigs(configs)
   }
 
+  def alterClientQuotas(adminClient: Admin, request: Map[ClientQuotaEntity, Map[String, Option[Double]]]): AlterClientQuotasResult = {
+    val entries = request.map { case (entity, alter) =>
+      val ops = alter.map { case (key, value) =>
+        new ClientQuotaAlteration.Op(key, value.map(Double.box).getOrElse(null))
+      }.asJavaCollection
+        new ClientQuotaAlteration(entity, ops)
```
bdbyrne: nit: indenting

core/src/main/scala/kafka/server/AdminManager.scala
```diff
@@ -807,23 +807,25 @@ class AdminManager(val config: KafkaConfig,
       case name => Sanitizer.desanitize(name)
     }
 
-  private def entityToSanitizedUserClientId(entity: ClientQuotaEntity): (Option[String], Option[String]) = {
+  private def entityToSanitizedUserClientId(entity: ClientQuotaEntity): (Option[String], Option[String], Option[String]) = {
```
bdbyrne: May want to rename (generalize) since this now includes IP

bdbyrne (approved):
LG - only minor comments, thanks for the change!

clients/src/main/java/org/apache/kafka/common/quota/ClientQuotaEntity.java
splett2: I have no objection to editing the KIP to changing this (and other uses of IP) to lower case, I just made it upper case because the original iteration of IP quotas in KIP-612 had IP in upper-case.

@apovzner @dajac for any thoughts on upper vs lower.
FWIW, the `kafka-configs` help output is better ordered when IP is in lower-case due to how it sorts flags.

dajac: I would also make it lower-case for consistency.

core/src/main/scala/kafka/server/AdminManager.scala
```diff
@@ -920,32 +954,49 @@ class AdminManager(val config: KafkaConfig,
...
-        case (Some(user), Some(clientId)) => (user + "/clients/" + clientId, ConfigType.User, DynamicConfig.User.configKeys)
-        case (Some(user), None) => (user, ConfigType.User, DynamicConfig.User.configKeys)
-        case (None, Some(clientId)) => (clientId, ConfigType.Client, DynamicConfig.Client.configKeys)
-        case _ => throw new InvalidRequestException("Invalid empty client quota entity")
+      val (path, configType, configKeys) = parseAndSanitizeQuotaEntity(entity) match {
+        case (Some(user), Some(clientId), None) => (user + "/clients/" + clientId, ConfigType.User, DynamicConfig.User.configKeys)
+        case (Some(user), None, None) => (user, ConfigType.User, DynamicConfig.User.configKeys)
+        case (None, Some(clientId), None) => (clientId, ConfigType.Client, DynamicConfig.Client.configKeys)
+        case (None, None, Some(ip)) => (ip, ConfigType.Ip, DynamicConfig.Ip.configKeys)
+        case _ => throw new InvalidRequestException("Invalid client quota entity")
```
dajac, line 1004: Should we also provide a better error message if the user mixes up IP and user/clientId?

core/src/main/scala/kafka/server/AdminManager.scala
```diff
@@ -920,32 +954,49 @@ class AdminManager(val config: KafkaConfig,
...
-
-    (userEntries ++ clientIdEntries ++ bothEntries).map { case ((u, c), p) =>
+    (userEntries ++ clientIdEntries ++ bothEntries).flatMap { case ((u, c), p) =>
       val quotaProps = p.asScala.filter { case (key, _) => QuotaConfigs.isQuotaConfig(key) }
       if (quotaProps.nonEmpty && matches(userComponent, u) && matches(clientIdComponent, c))
         Some(userClientIdToEntity(u, c) -> fromProps(quotaProps))
       else
         None
-    }.flatten.toMap
+    }.toMap
```
dajac, line 963: I am not sure to understand why you have removed `flatten` here. Could you elaborate a bit on this please?

core/src/main/scala/kafka/admin/ConfigCommand.scala
```diff
@@ -56,11 +56,12 @@ import scala.collection._
...
  *     <li> broker-logger: --broker-logger <broker-id> OR --entity-type broker-loggers --entity-name <broker-id>
+ *     <li> ip: --IP <ip> OR --entity-type IPs --entity-name <ip>
  * </ul>
- * --entity-type <users|clients|brokers> --entity-default may be specified in place of --entity-type <users|clients|brokers> --entity-name <entityName>
- * when describing or altering default configuration for users, clients, or brokers, respectively.
- * Alternatively, --user-defaults, --client-defaults, or --broker-defaults may be specified in place of
- * --entity-type <users|clients|brokers> --entity-default, respectively.
+ * --entity-type <users|clients|brokers|IPs> --entity-default may be specified in place of --entity-type <users|clients|brokers|IPs> --entity-name <entityName>
+ * when describing or altering default configuration for users, clients, brokers, or IPs, respectively.
+ * Alternatively, --user-defaults, --client-defaults, --broker-defaults, or --IP-defaults may be specified in place of
```
dajac: `--IP-defaults` looks pretty bad. I think that we should definitely use lower-case.

dajac (commented):
@splett2 Thanks for the PR. It looks pretty good. I have left some comments. I will need to do another pass over it, especially over the tests.

core/src/main/scala/kafka/server/AdminManager.scala
splett2: this was just a coding style change, from calling `entries.map(...).flatten()` => `entries.flatMap()`

core/src/main/scala/kafka/admin/ConfigCommand.scala
splett2: sounds good. In this case, do we want to change all mentions of IP to lower-case? including entity type/ZK path.

apovzner: yes, agreed about `--IP-defaults` looking bad. 

The main reason I originally proposed upper case "IP" in the KIP is because `ips` as in `--entity-type ips` does not seem very recognizable as plural for IP (but that could be just me), i.e., not very user friendly. But if we do `--entity-type IPs` and then `--ip-defaults`, I think that would cause more issues later because people would assume it's consistent and use wrong case.

Ok, I am convinced that we lower-case is better, since there are more advantages to this vs. upper-case.

clients/src/main/java/org/apache/kafka/common/quota/ClientQuotaEntity.java
apovzner: as I mentioned in my other comment, agreed with using lower-case.

core/src/main/scala/kafka/admin/ConfigCommand.scala
dajac: @splett2 Isn't the ZK path already lower-case?

core/src/main/scala/kafka/server/AdminManager.scala
dajac: Ah. I missed the `flatMap`. That makes sense, thanks.

core/src/main/scala/kafka/admin/ConfigCommand.scala
splett2: @dajac 
it was originally lower-case in #9386
after reading the KIP, i realized it was described as being upper case, e.g.
```
Default quota for <IPs> will be stored in Zookeeper at /config/IPs/<default>
Quota for a specific IP address for which quota override is defined will be stored in Zookeeper at /config/IPs/<ip-address>
```
and changed it to upper in this PR with the change in
```
core/src/main/scala/kafka/server/DynamicConfigManager.scala
```

core/src/test/scala/integration/kafka/network/DynamicConnectionQuotaTest.scala
```diff
@@ -259,13 +264,19 @@ class DynamicConnectionQuotaTest extends BaseRequestTest {
   }
 
   private def updateIpConnectionRate(ip: Option[String], updatedRate: Int): Unit = {
-    adminZkClient.changeIpConfig(ip.getOrElse(ConfigEntityName.Default),
-      CoreUtils.propsWith(DynamicConfig.Ip.IpConnectionRateOverrideProp, updatedRate.toString))
+    val initialConnectionCount = connectionCount
+    val adminClient = createAdminClient()
```
apovzner, line 268: `TestUtils.alterClientQuotas` called before `adminClient.close` can throw an exception. Admin extends AutoCloseable, so you can use try-with-resources.

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
```diff
@@ -472,6 +512,134 @@ class ConfigCommandTest extends ZooKeeperTestHarness with Logging {
...
+      val opts = new ConfigCommandOptions(Array("--bootstrap-server", "localhost:9092",
+        "--entity-type", entityType, "--entity-name", "admin",
+        "--alter") ++ alterOpts)
+      val e = intercept[IllegalArgumentException] {
+        ConfigCommand.alterConfig(mockAdminClient, opts)
+      }
+      assertTrue(s"Unexpected exception: $e", e.getMessage.contains("some_config"))
+    }
+
+    verifyCommand("ips", "--add-config", "connection_creation_rate=10000,some_config=10")
```
apovzner: super tiny nit: `verifyCommand` name makes the code a bit confusing since the command is expected to fail. What about `verifyCommandFails` or `verifyFails`?

apovzner (approved):
LGTM! Left couple of minor comments.

core/src/main/scala/kafka/admin/ConfigCommand.scala
```diff
@@ -864,11 +885,21 @@ object ConfigCommand extends Config {
...
         throw new IllegalArgumentException(s"an entity name must be specified with --describe of ${entityTypeVals.mkString(",")}")
 
       if (options.has(alterOpt)) {
-        if (entityTypeVals.contains(ConfigType.User) || entityTypeVals.contains(ConfigType.Client) || entityTypeVals.contains(ConfigType.Broker)) {
+        if (entityTypeVals.contains(ConfigType.User) ||
+            entityTypeVals.contains(ConfigType.Client) ||
+            entityTypeVals.contains(ConfigType.Broker) ||
+            entityTypeVals.contains(ConfigType.Ip)) {
           if (!hasEntityName && !hasEntityDefault)
             throw new IllegalArgumentException("an entity-name or default entity must be specified with --alter of users, clients or brokers")
```
dajac: nit: Could we add ip in the list as well?

core/src/main/scala/kafka/server/AdminManager.scala
```diff
@@ -920,32 +954,51 @@ class AdminManager(val config: KafkaConfig,
...
   }
 
   def alterClientQuotas(entries: Seq[ClientQuotaAlteration], validateOnly: Boolean): Map[ClientQuotaEntity, ApiError] = {
     def alterEntityQuotas(entity: ClientQuotaEntity, ops: Iterable[ClientQuotaAlteration.Op]): Unit = {
-      val (path, configType, configKeys) = entityToSanitizedUserClientId(entity) match {
-        case (Some(user), Some(clientId)) => (user + "/clients/" + clientId, ConfigType.User, DynamicConfig.User.configKeys)
-        case (Some(user), None) => (user, ConfigType.User, DynamicConfig.User.configKeys)
-        case (None, Some(clientId)) => (clientId, ConfigType.Client, DynamicConfig.Client.configKeys)
-        case _ => throw new InvalidRequestException("Invalid empty client quota entity")
+      val (path, configType, configKeys) = parseAndSanitizeQuotaEntity(entity) match {
```
dajac, line 994: It seems that we should validate the provided `ip` somewhere here. I just tried with the following command:
```
./bin/kafka-configs.sh --bootstrap-server localhost:9092 --alter --entity-type ips --entity-name ip --add-config  connection_creation_rate=10
```
The command completes but then the broker fails while applying the configuration:
```
[2020-12-04 15:31:01,951] ERROR error processing change notification {"version":2,"entity_path":"ips/ip"} from /config/changes/config_change_0000000001 (kafka.common.ZkNodeChangeNotificationListener)
java.lang.IllegalArgumentException: Unable to resolve address ip
	at kafka.server.IpConfigHandler.processConfigChanges(ConfigHandler.scala:199)
	at kafka.server.DynamicConfigManager$ConfigChangedNotificationHandler$.processEntityConfigChangeVersion2(DynamicConfigManager.scala:151)
	at kafka.server.DynamicConfigManager$ConfigChangedNotificationHandler$.$anonfun$processNotification$1(DynamicConfigManager.scala:105)
	at kafka.server.DynamicConfigManager$ConfigChangedNotificationHandler$.processNotification(DynamicConfigManager.scala:96)
	at kafka.common.ZkNodeChangeNotificationListener.$anonfun$processNotification$1(ZkNodeChangeNotificationListener.scala:106)
	at kafka.common.ZkNodeChangeNotificationListener.processNotification(ZkNodeChangeNotificationListener.scala:106)
	at kafka.common.ZkNodeChangeNotificationListener.$anonfun$processNotifications$2(ZkNodeChangeNotificationListener.scala:90)
	at kafka.common.ZkNodeChangeNotificationListener.$anonfun$processNotifications$2$adapted(ZkNodeChangeNotificationListener.scala:87)
	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:563)
	at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:561)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:919)
	at kafka.common.ZkNodeChangeNotificationListener.kafka$common$ZkNodeChangeNotificationListener$$processNotifications(ZkNodeChangeNotificationListener.scala:87)
	at kafka.common.ZkNodeChangeNotificationListener$ChangeNotification.process(ZkNodeChangeNotificationListener.scala:120)
	at kafka.common.ZkNodeChangeNotificationListener$ChangeEventProcessThread.doWork(ZkNodeChangeNotificationListener.scala:146)
	at kafka.utils.ShutdownableThread.run(ShutdownableThread.scala:96)
```

It would be good to cover this in the tests, both for the command and the api.

This is also true with:
```
./bin/kafka-configs.sh --zookeeper localhost:2181 --alter --ip ip --add-config connection_creation_rate=2
```

core/src/main/scala/kafka/admin/ConfigCommand.scala
```diff
@@ -778,6 +793,10 @@ object ConfigCommand extends Config {
     val brokerLogger = parser.accepts("broker-logger", "The broker's ID for its logger config.")
       .withRequiredArg
       .ofType(classOf[String])
+    val ipDefaults = parser.accepts("ip-defaults", "The config defaults for all IPs.")
+    val ip = parser.accepts("ip", "The IP address.")
```
dajac, line 797: It would be good to add these two flags in the KIP as well for completeness.

core/src/test/scala/integration/kafka/network/DynamicConnectionQuotaTest.scala
dajac: I second this.

core/src/main/scala/kafka/admin/ConfigCommand.scala
```diff
@@ -864,11 +885,21 @@ object ConfigCommand extends Config {
         }
       }
 
+      if (hasEntityName && entityTypeVals.contains(ConfigType.Ip)) {
+        Seq(entityName, ip).filter(options.has(_)).map(options.valueOf(_)).foreach { ipAddress =>
+          if (!Utils.validHostPattern(ipAddress))
```
dajac: I think that we should validate that `ipAddress` is actually an IP. `validHostPattern` only ensure that it is a valid host name.

core/src/main/scala/kafka/admin/ConfigCommand.scala
```diff
@@ -864,11 +885,21 @@ object ConfigCommand extends Config {
         }
       }
 
+      if (hasEntityName && entityTypeVals.contains(ConfigType.Ip)) {
+        Seq(entityName, ip).filter(options.has(_)).map(options.valueOf(_)).foreach { ipAddress =>
```
dajac: nit: We can probably remove `(_)` in `filter` and `map`.

core/src/main/scala/kafka/server/AdminManager.scala
```diff
@@ -920,32 +954,51 @@ class AdminManager(val config: KafkaConfig,
...
+        case (Some(user), Some(clientId), None) => (user + "/clients/" + clientId, ConfigType.User, DynamicConfig.User.configKeys)
+        case (Some(user), None, None) => (user, ConfigType.User, DynamicConfig.User.configKeys)
+        case (None, Some(clientId), None) => (clientId, ConfigType.Client, DynamicConfig.Client.configKeys)
+        case (None, None, Some(ip)) => (ip, ConfigType.Ip, DynamicConfig.Ip.configKeys)
+        case (_, _, Some(_)) => throw new InvalidRequestException(s"Invalid quota entity combination, " +
```
dajac, line 1002: nit: Shouldn't we use `(Some(_), Some(_), Some(_))` here in order to be really accurate? As it is, the first two could be anything. That works because it is after all the others but I would prefer to be explicit if possible.

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
```diff
@@ -398,19 +418,39 @@ class ConfigCommandTest extends ZooKeeperTestHarness with Logging {
...
+        assertEquals("b", configChange.get("a"))
+        assertEquals("d", configChange.get("c"))
       }
     }
+
+    ConfigCommand.alterConfigWithZk(null, createOpts, new TestAdminZkClient(zkClient))
+  }
+
+  private def toValues(entityName: Option[String], entityType: String, command: String):
+  (Array[String], Option[String], Option[ClientQuotaFilterComponent]) = {
```
dajac: nit: Indentation seems wrong here.

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
```diff
@@ -472,6 +512,134 @@ class ConfigCommandTest extends ZooKeeperTestHarness with Logging {
...
+  @Test
+  def shouldNotAlterNonQuotaIpConfigsUsingBootstrapServer(): Unit = {
+    // when using --bootstrap-server, it should be illegal to alter anything that is not a connection quota
+    // for ip entities
+    val node = new Node(1, "localhost", 9092)
+    val mockAdminClient = new MockAdminClient(util.Collections.singletonList(node), node)
+
+    def verifyCommand(entityType: String, alterOpts: String*): Unit = {
+      val opts = new ConfigCommandOptions(Array("--bootstrap-server", "localhost:9092",
+        "--entity-type", entityType, "--entity-name", "admin",
```
dajac: Actually, I believe that the command should fail due to `admin` which is not valid IP. It may be better to use a valid IP address here as you want to tests the configs.

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
dajac: I do agree. It took me a while to grasp it. Moreover, the very same `verifyCommand` is defined in other places. I wonder if we could consolidate them into one. What do you think?

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
```diff
@@ -472,6 +512,134 @@ class ConfigCommandTest extends ZooKeeperTestHarness with Logging {
...
+
+      val node = new Node(1, "localhost", 9092)
+      val mockAdminClient = new MockAdminClient(util.Collections.singletonList(node), node) {
+        override def describeClientQuotas(filter: ClientQuotaFilter, options: DescribeClientQuotasOptions): DescribeClientQuotasResult = {
+          assertTrue(filter.strict)
+          assertEquals(expectedFilter, filter)
+          describedConfigs = true
+          describeResult
+        }
+      }
```
dajac: Have you considered using a real mock instead of doing this? Is there a reason not to? It seems to be that it would be more appropriate.

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
```diff
@@ -472,6 +512,134 @@ class ConfigCommandTest extends ZooKeeperTestHarness with Logging {
...
+      EasyMock.expect(describeResult.entities()).andReturn(describeFuture)
+
+      var alteredConfigs = false
+      val alterFuture = new KafkaFutureImpl[Void]
+      alterFuture.complete(null)
+      val alterResult: AlterClientQuotasResult = EasyMock.createNiceMock(classOf[AlterClientQuotasResult])
+      EasyMock.expect(alterResult.all()).andReturn(alterFuture)
+
+      val node = new Node(1, "localhost", 9092)
+      val mockAdminClient = new MockAdminClient(util.Collections.singletonList(node), node) {
```
dajac: ditto

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
```diff
@@ -187,6 +190,59 @@ class ClientQuotasRequestTest extends BaseRequestTest {
     ))
   }
 
+  @Test
+  def testAlterIpQuotasRequest(): Unit = {
```
dajac, line 194: We must add tests with invalid IPs.

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
```diff
@@ -472,6 +512,134 @@ class ConfigCommandTest extends ZooKeeperTestHarness with Logging {
...
+
+    verifyCommand("ips", "--add-config", "connection_creation_rate=10000,some_config=10")
+    verifyCommand("ips", "--add-config", "some_config=10")
+    verifyCommand("ips", "--delete-config", "connection_creation_rate=10000,some_config=10")
+    verifyCommand("ips", "--delete-config", "some_config=10")
+  }
+
+  @Test
+  def testDescribeIpConfigs(): Unit = {
+    def testShouldDescribeIpConfig(ip: Option[String]): Unit = {
```
dajac: nit: I feel like this inner method is quite large and may merit to be extracted. It feels a bit weird when the inner method is way larger than the actual body of the method. I guess that it is a matter of taste but I believe that the code would be more readable. I leave this up to you.

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
```diff
@@ -236,13 +310,22 @@ class ClientQuotasRequestTest extends BaseRequestTest {
...
+      e -> Map((RequestPercentageProp, Some(v)))
+    }.toMap
+    val ipQuotas = matchIpEntities.map { case (e, v) =>
+      e -> Map((IpConnectionRateProp, Some(v)))
+    }.toMap
+    val result = alterClientQuotas(userClientQuotas ++ ipQuotas, validateOnly = false)
+    (matchUserClientEntities ++ matchIpEntities).foreach(e => result(e._1).get(10, TimeUnit.SECONDS))
 
     // Allow time for watch callbacks to be triggered.
     Thread.sleep(500)
```
dajac: This is not related to your PR but I wonder if it would be possible to replace this sleep by `retry` or `waitUntilTrue`. Thoughts?

dajac (commented):
@splett2 Thanks for the updates. I just made another pass and left few more comments.

core/src/main/scala/kafka/admin/ConfigCommand.scala
splett2: will do, along with the change to lower

core/src/main/scala/kafka/server/AdminManager.scala
splett2: i used `(_, _, Some(_))` for this case, because we want to throw the more verbose exception in all of these cases:

`(Some(_), None, Some(_))` => specifying a user and IP, but no client
`(None, Some, Some)` => specifying client and IP, but no user
`(Some, Some, Some)` => specifying client, user and IP

basically any time there is a client OR user specified along with IP. I felt like adding all three cases would be pretty verbose, so I relied on the ordering of the cases instead.

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
splett2: makes sense, will do.
Actually, taking a closer look, there is no need for the `Thread.sleep()` call. None of the tests that use the match setup rely on ZK watches to trigger. I'll just remove the call to `Thread.sleep()` instead.

core/src/main/scala/kafka/admin/ConfigCommand.scala
splett2: unexpectedly, I get a compile error if I remove `(_)` in `map`.

```
[Error]~/kafka/core/src/main/scala/kafka/admin/ConfigCommand.scala:889: type mismatch;
 found   : [V](x$1: joptsimple.OptionSpec[V]): V <and> (x$1: String): Object
 required: joptsimple.ArgumentAcceptingOptionSpec[String] => ?
```
I will keep it as is for both cases to avoid confusion, e.g. why do we use a function reference in `filter`, but not `map`

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
splett2: hmm, good point. I did this because it's what a similar quotas config command test did, but you're right that a real mock would be more appropriate here.

core/src/main/scala/kafka/admin/ConfigCommand.scala
splett2: it looks like `max.connections.per.ip` permits hosts in addition to IPs.

```
    val invalidAddresses = maxConnectionsPerIpOverrides.keys.filterNot(address => Utils.validHostPattern(address))
    if (!invalidAddresses.isEmpty)
      throw new IllegalArgumentException(s"${KafkaConfig.MaxConnectionsPerIpOverridesProp} contains invalid addresses : ${invalidAddresses.mkString(",")}")
```
Do you think we should prefer consistency with the existing IP quota prop (e.g., permit any resolvable host address) or go for more strict enforcement (only IPs permitted)?

dajac: I think that we also permit hosts in our case as `InetAddress.getByName` resolves names that are not IP addresses. The issue is that `Utils.validHostPattern` is just a regex so it does not validate much.

`max.connections.per.ip` is validated in two phases. The code that you outlines is from the `KafkaConfig`. I suppose that we do a lightweight verification there in order to avoid doing any DNS resolution while parsing the config. This is later complemented in the socket server.

```
  private[network] def updateMaxConnectionsPerIpOverride(overrideQuotas: Map[String, Int]): Unit = {
    maxConnectionsPerIpOverrides = overrideQuotas.map { case (host, count) => (InetAddress.getByName(host), count) }
  }
```

In our case, it is important to remain consistent everywhere.

core/src/main/scala/kafka/server/AdminManager.scala
dajac: Thanks for the explanation. That makes sense.

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
splett2: I took an attempt at this, and realized that the reason we use `MockAdminClient` is because `ClientQuotaFilter` and `ClientQuotaAlteration` take `Collection` in their constructors.

If we used a real mock, it would make the tests unnecessarily brittle, since we'd have to match the collection type used in the `ConfigCommand` implementation.

core/src/test/scala/unit/kafka/admin/ConfigCommandTest.scala
splett2: done. refactored the method to be more generic as well, to reduce some code duplication with an existing test.

splett2:
Thanks for the feedback @dajac 
I went ahead and also refactored some tests in `ConfigCommandTest` that were a bit repetitive.
Let me know what you think

core/src/main/scala/kafka/admin/ConfigCommand.scala
splett2: done. while I was updating the KIP, i noticed that the previous config documentation was inconsistent with existing configs.

```
Default connection rate quotas for an IP address can be configured by omitting entity name.
bin/kafka-configs  --bootstrap-server localhost:9091 --alter --add-config 'connection_creation_rate=100' --entity-type IPs
```
This isn't the case for any of the existing configs, you get an error:
`an entity-name or default entity must be specified with --alter of users, clients, brokers`
so I also updated the KIP for default IP to specify --entity-default.

core/src/main/scala/kafka/zk/AdminZkClient.scala
```diff
@@ -394,8 +393,7 @@ class AdminZkClient(zkClient: KafkaZkClient) extends Logging {
    * @param configs properties to validate for the IP
    */
   def validateIpConfig(ip: String, configs: Properties): Unit = {
-    if (ip != ConfigEntityName.Default && !Utils.validHostPattern(ip))
-      throw new AdminOperationException(s"IP $ip is not a valid address.")
+    DynamicConfig.Ip.validateIpOrHost(ip)
```
dajac: It is a little weird to throw `InvalidRequestException` here because there is not request involved. Users of `AdminZkClient` may not expect this. I think that it would be better to throw a `IllegalArgumentException`. Then in the `AdminManager#alterClientQuotas`, we can catch it and transform it into `InvalidRequestException` by copying the internal message.

core/src/main/scala/kafka/admin/ConfigCommand.scala
```diff
@@ -864,13 +885,20 @@ object ConfigCommand extends Config {
         }
       }
 
+      if (hasEntityName && entityTypeVals.contains(ConfigType.Ip)) {
+        Seq(entityName, ip).filter(options.has(_)).map(options.valueOf(_)).foreach(DynamicConfig.Ip.validateIpOrHost)
```
dajac: ditto here. It is a bit weird to have an `InvalidRequestException` thrown here.

core/src/main/scala/kafka/server/DynamicConfig.scala
```diff
@@ -141,6 +144,18 @@ object DynamicConfig {
     def names = ipConfigs.names
 
     def validate(props: Properties) = DynamicConfig.validate(ipConfigs, props, customPropsAllowed = false)
+
+    def validateIpOrHost(ip: String): Unit = {
+      if (ip != ConfigEntityName.Default) {
+        if (!Utils.validHostPattern(ip))
+          throw new InvalidRequestException(s"$ip is not a valid hostname")
```
dajac: I wonder if this step is really necessary as `InetAddress.getByName` also ensures that the hostname is valid. What do you think?

core/src/main/scala/kafka/server/DynamicConfig.scala
```diff
@@ -141,6 +144,18 @@ object DynamicConfig {
...
+        try {
+          InetAddress.getByName(ip)
+        } catch {
+          case _ :UnknownHostException => throw new InvalidRequestException(s"$ip is not a valid IP or resolvable hostname")
```
dajac: nit: `_ :UnknownHostException` => `_: UnknownHostException`

dajac (commented):
@splett2 Thanks for the updates. I have left few minor comments.

core/src/main/scala/kafka/server/AdminManager.scala
```diff
@@ -920,32 +954,57 @@ class AdminManager(val config: KafkaConfig,
...
+      val (path, configType, configKeys) = parseAndSanitizeQuotaEntity(entity) match {
+        case (Some(user), Some(clientId), None) => (user + "/clients/" + clientId, ConfigType.User, DynamicConfig.User.configKeys)
+        case (Some(user), None, None) => (user, ConfigType.User, DynamicConfig.User.configKeys)
+        case (None, Some(clientId), None) => (clientId, ConfigType.Client, DynamicConfig.Client.configKeys)
+        case (None, None, Some(ip)) =>
+          try {
+            DynamicConfig.Ip.validateIpOrHost(ip)
+          } catch {
+            case e: IllegalArgumentException => throw new InvalidRequestException(e.getMessage)
+          }
```
dajac: * For my understanding, have you chosen to do the validation here to ensure that requests with `validateOnly = true` are fully validated?
* Initially, I thought that we could just catch the `IllegalArgumentException` here https://github.com/apache/kafka/blob/1b0ec2a3ca79dc8da5908c41afc4ef87d2288d84/core/src/main/scala/kafka/server/AdminManager.scala#L1043. Doing the conversion here seems reasonable as well. Alternatively, we could also change `validateIpOrHost` to return a boolean and let the caller decides which exception to throw.

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
```diff
@@ -223,8 +279,34 @@ class ClientQuotasRequestTest extends BaseRequestTest {
     alterEntityQuotas(entity, Map((ProducerByteRateProp -> Some(10000.5))), validateOnly = true)
   }
 
+  @Test
+  def testAlterClientQuotasInvalidEntityCombination(): Unit = {
```
dajac, line 288: For these three tests, would it make sense to also verify that the error message? `InvalidRequestException` are thrown all over the place so we may get one but not for the expected cause. Your last commit suggests that you got caught by this already, is it? https://github.com/apache/kafka/pull/9628/commits/1b0ec2a3ca79dc8da5908c41afc4ef87d2288d84#diff-f02b619b5cd14e83f7e21dbe211b3f88336f825e1ee5c630fee32b8a0fbe3d20R294

dajac (commented):
@splett2 Thanks for the updates. I have left two small suggestions. Besides this, the PR looks good. One last iteration and we can merge it.

core/src/main/scala/kafka/server/AdminManager.scala
splett2: @dajac 
yes, I thought that catching in that block would work, but after taking a closer look, it seems like entity validation is only done when altering, so we wouldn't hit validation in the `validateOnly = true` case.

I think changing `validateIpOrHost` to return a boolean is reasonable. I originally wanted to have it throw exception instead of boolean since we had two validation steps (`validHostPattern` and then resolvable IP), but since we simplified that code block to only do host resolution, it makes more sense to have it just return a boolean.

dajac (approved):
LGTM.

dajac:
Failed test is not related:
* JDK 11 - kafka.api.PlaintextAdminIntegrationTest.testAlterReplicaLogDirs

dajac:
@splett2 Thanks for your contribution!

PR #9630, opened 2020-11-20
KAFKA-10739; Replace EpochEndOffset with automated protocol

dajac:
This patch follows up https://github.com/apache/kafka/pull/9547. It refactors KafkaApis, ReplicaManager and Partition to use `OffsetForLeaderPartitionResult` instead of `EpochEndOffset`. In the mean time, it removes `OffsetsForLeaderEpochRequest#epochsByTopicPartition` and `OffsetsForLeaderEpochResponse#responses` and replaces their usages to use the automated protocol directly. Finally, it removes old constructors in `OffsetsForLeaderEpochResponse`. The patch relies on existing tests.


dajac:
cc @hachikuji @chia7712

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1862,23 +1864,51 @@ class ReplicaManager(val config: KafkaConfig,
...
-        case HostedPartition.None if metadataCache.contains(tp) =>
-          new EpochEndOffset(Errors.NOT_LEADER_OR_FOLLOWER, UNDEFINED_EPOCH, UNDEFINED_EPOCH_OFFSET)
-
-        case HostedPartition.None =>
-          new EpochEndOffset(Errors.UNKNOWN_TOPIC_OR_PARTITION, UNDEFINED_EPOCH, UNDEFINED_EPOCH_OFFSET)
+  def lastOffsetForLeaderEpoch(
+    requestedEpochInfo: Seq[OffsetForLeaderTopic]
+  ): Seq[OffsetForLeaderTopicResult] = {
+    requestedEpochInfo.map { offsetForLeaderTopic =>
+      val partitions = offsetForLeaderTopic.partitions.asScala.map { offsetForLeaderPartition =>
```
chia7712, line 1883: Could we avoid duplicate conversion between scala and java? It can be rewrite by java stream APIs so the  ```asScala```  can be avoid.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -2592,25 +2596,39 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+    val (authorizedTopics, unauthorizedTopics) =
       if (authorize(request.context, CLUSTER_ACTION, CLUSTER, CLUSTER_NAME, logIfDenied = false))
-        (requestInfo, Map.empty[TopicPartition, OffsetsForLeaderEpochRequest.PartitionData])
-      else partitionMapByAuthorized(request.context, DESCRIBE, TOPIC, requestInfo)(_.topic)
+        (topics, Seq.empty[OffsetForLeaderTopic])
+      else partitionSeqByAuthorized(request.context, DESCRIBE, TOPIC, topics)(_.topic)
+
+    val endOffsetsForAuthorizedPartitions = replicaManager.lastOffsetForLeaderEpoch(authorizedTopics)
+    val endOffsetsForUnauthorizedPartitions = unauthorizedTopics.map { offsetForLeaderTopic =>
+      val partitions = offsetForLeaderTopic.partitions.asScala.map { offsetForLeaderPartition =>
```
chia7712, line 2610: ditto

core/src/test/scala/unit/kafka/server/epoch/util/ReplicaFetcherMockBlockingSend.scala
```diff
@@ -78,7 +80,19 @@ class ReplicaFetcherMockBlockingSend(offsets: java.util.Map[TopicPartition, Epoc
...
+
+        val data = new OffsetForLeaderEpochResponseData()
+        currentOffsets.forEach((tp, offsetForLeaderPartition) => {
+          var topic = data.topics.find(tp.topic)
+          if (topic == null) {
+            topic = new OffsetForLeaderTopicResult()
+              .setTopic(tp.topic)
+            data.topics.add(topic)
+          }
+          topic.partitions.add(offsetForLeaderPartition.setPartition(tp.partition))
```
chia7712: Should it be ```topic.partitions.add(offsetForLeaderPartition.duplicate())```?

chia7712 (approved):
overall +1. Three minor questions are left.

Also, there are a lot of data structure like ```Map[TopicPartition, OffsetForLeaderPartitionResult]```. It contains duplicate (int) partition info. Not sure whether
it is worth replacing it by ```Map[String,List[OffsetForLeaderPartitionResult]]``` to eliminate duplicate partition info. It will bring a bunch of changes :(

core/src/main/scala/kafka/server/ReplicaManager.scala
dajac: `asScala` and `asJava` just wrap the collections to make it accessible from respectively Scala and Java. In that regards, I am not sure that we would gain much by using the stream API here as it also creates a Stream. It seems more natural to keep using Scala here. If you look in the API layer, we do this everywhere.

core/src/test/scala/unit/kafka/server/epoch/util/ReplicaFetcherMockBlockingSend.scala
dajac: Hum.. I am not sure to understand the `duplicate` suggestion here. Could you elaborate?

dajac:
> Also, there are a lot of data structure like `Map[TopicPartition, OffsetForLeaderPartitionResult]`. It contains duplicate (int) partition info. Not sure whether
> it is worth replacing it by `Map[String,List[OffsetForLeaderPartitionResult]]` to eliminate duplicate partition info. It will bring a bunch of changes :(

Yeah, I considered this actually. Finally, I went with keeping TopicPartition as a key mainly because TopicPartition is required by most of the callers.

core/src/main/scala/kafka/server/ReplicaManager.scala
chia7712: Make sense to me :)

core/src/test/scala/unit/kafka/server/epoch/util/ReplicaFetcherMockBlockingSend.scala
chia7712: oh, it is just personal taste. the code ```topic.partitions.add(offsetForLeaderPartition.setPartition(tp.partition))``` adds inner mutable object (```offsetForLeaderPartition```) to returned response. I prefer to add a copy of ```offsetForLeaderPartition```.

dajac: Good point. Actually, `setPartition(tp.partition)` is not needed. Let me remove it.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1862,23 +1864,51 @@ class ReplicaManager(val config: KafkaConfig,
...
+            partition.lastOffsetForLeaderEpoch(
+              currentLeaderEpochOpt,
+              offsetForLeaderPartition.leaderEpoch,
+              fetchOnlyFromLeader = true)
+
+          case HostedPartition.Offline =>
+            new OffsetForLeaderPartitionResult()
+              .setPartition(offsetForLeaderPartition.partition)
+              .setErrorCode(Errors.KAFKA_STORAGE_ERROR.code)
+              .setLeaderEpoch(UNDEFINED_EPOCH)
```
hachikuji: I'm wondering if it is reasonable to rely on defaults for some of these. I guess there's value in being explicit, but it is a tad vexing to see the same code repeated for a few of these cases.

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
```diff
@@ -3727,15 +3733,24 @@ public void testOffsetValidationRequestGrouping() {
...
-                tp -> new EpochEndOffset(Errors.NONE, 4, 0)
-            ));
+            OffsetForLeaderEpochResponseData data = new OffsetForLeaderEpochResponseData();
+            expectedPartitions.forEach(tp -> {
+                OffsetForLeaderTopicResult topic = data.topics().find(tp.topic());
+                if (topic == null) {
+                    topic = new OffsetForLeaderTopicResult().setTopic(tp.topic());
+                    data.topics().add(topic);
+                }
+                topic.partitions().add(new OffsetForLeaderPartitionResult()
```
hachikuji: Maybe just me, but `OffsetForLeaderPartitionResult` seems more cumbersome and less descriptive than `EpochEndOffset`. Would it be worth changing the name in the generated schema?

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
```diff
@@ -3727,15 +3733,24 @@ public void testOffsetValidationRequestGrouping() {
...
             allRequestedPartitions.addAll(expectedPartitions);
 
-            Map<TopicPartition, EpochEndOffset> endOffsets = expectedPartitions.stream().collect(Collectors.toMap(
-                Function.identity(),
-                tp -> new EpochEndOffset(Errors.NONE, 4, 0)
-            ));
+            OffsetForLeaderEpochResponseData data = new OffsetForLeaderEpochResponseData();
+            expectedPartitions.forEach(tp -> {
+                OffsetForLeaderTopicResult topic = data.topics().find(tp.topic());
+                if (topic == null) {
```
hachikuji, line 3739: Not for this patch, but all of this boilerplate we need to build the topic groupings gets annoying. It is such a common case that it might be worth writing a special type that lets the parser construct `Map<TopicPartition, Data>` directly since that is really what the code wants. Alternatively, maybe we could flatten the schemas and introduce compression.

hachikuji (commented):
Nice improvement. Left a couple small comments.

core/src/main/scala/kafka/server/ReplicaManager.scala
dajac: That makes sense. Let me use them where ever possible. I will also add a default value (-1) for `EndOffset` in the schema.

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
dajac: I do agree with you. Using `EpochEndOffset` sounds much better.

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
dajac: That would be nice, indeed! I have been thinking about this as well but did not have the time to tackle this yet. I have opened a JIRA to track this: https://issues.apache.org/jira/browse/KAFKA-10795.

dajac:
@hachikuji Thanks for your feedback. I have addressed your points.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -1839,20 +1839,17 @@ private OffsetsForLeaderEpochResponse createLeaderEpochResponse() {
             .setPartitions(Arrays.asList(
                 new OffsetForLeaderPartitionResult()
                     .setPartition(0)
-                    .setErrorCode(Errors.NONE.code())
                     .setLeaderEpoch(1)
                     .setEndOffset(0),
                 new OffsetForLeaderPartitionResult()
                     .setPartition(1)
-                    .setErrorCode(Errors.NONE.code())
```
hachikuji: In the case of the error code, I think it might be better to be explicit.

hachikuji (approved):
Thanks for the updates. LGTM. Just a minor comment you can consider (no need for re-review).

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
dajac: I don't feel strong either ways so I will follow your suggestion.

dajac:
I just rebased the PR cause it had conflicts with 7ecc3a579a4b13e0cef4bd3129982ea3bc1a9341.

dajac:
Test failure is unrelated:
* Build / JDK 8 / kafka.api.TransactionsTest.testBumpTransactionalEpoch

PR #9631, opened 2020-11-20
KAFKA-9672: Leader with ISR as a superset of replicas

jsancio:
It is possible for the the controller to send LeaderAndIsr requests with
an ISR that contains ids not in the replica set. This is used during
reassignment so that the partition leader doesn't add replicas back to
the ISR. This is needed because the controller updates ZK and the
replicas through two rounds:

1. The first round of ZK updates and LeaderAndIsr requests shrinks the ISR.

2. The second round of ZK updates and LeaderAndIsr requests shrinks the replica
set.

This could be avoided by doing 1. and 2. in one round. Unfortunately the
current controller implementation makes that non-trivial.

This commit changes the leader to allow the state where the ISR contains
ids that are not in the replica set and to remove such ids from the ISR
if required.

comment to ping reviewers.*


core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -947,9 +947,10 @@ class Partition(val topicPartition: TopicPartition,
                                   leaderEndOffset: Long,
                                   currentTimeMs: Long,
                                   maxLagMs: Long): Boolean = {
-    val followerReplica = getReplicaOrException(replicaId)
-    followerReplica.logEndOffset != leaderEndOffset &&
-      (currentTimeMs - followerReplica.lastCaughtUpTimeMs) > maxLagMs
+    getReplica(replicaId).fold(true) { followerReplica =>
```
junrao, line 943: For the reassignment case, once the controller shrinks the assigned replica set, the next step is for the controller to remove the remove replica from ISR and bump up the leader epoch. The shrunk isr will then be propagated to the leader. With `fold(true)`, we allow to leader to shrink ISR immediately. This might be ok, but is unnecessary work since the controller will be doing that soon.

Another option is to use `fold(false)`. This way, the leader won't shrink the removed replica from ISR. Only the controller will do.

junrao (commented):
@jsancio : Thanks for the PR. Just one comment below.

core/src/main/scala/kafka/cluster/Partition.scala
jsancio: Thanks for the review!

> This might be ok, but is unnecessary work since the controller will be doing that soon.

According to some users and the report from KAFKA-9672, it looks like under some conditions the controller is writing to ZK that it removed the replica from the assignment but not from the ISR. I am unable to reproduce this or convince myself from the code on how this can happen.

I was thinking of defensively letting the leader also remove the replica from the ISR so that Kafka can recover from this case. If the leader is not allowed to do this then `ack=all` produce messages will continue to fail.

What do you think @junrao?

junrao: @jsancio : Yes, we can keep the logic in the PR. On the leader, the logic for shrinking ISR is checked every 10 secs by default. So, in the common case when completing a reassignment, the reduced ISR by the controller will be propagated to the leader before the leader's ISR shrinking logic kicks in.

junrao (approved):
@jsancio : Thanks for the explanation. LGTM.

Do you know why the java 8 tests timed out?

jsancio:
@junrao I merged the latest trunk to this PR. Jenkins ran the tests and all the failures look unrelated.

PR #9635, opened 2020-11-21
KAFKA-10756; Add missing unit test for `UnattachedState`

dengziming:
Accidentally find that there is no unit test for UnattachedState, We should add missing unit tests, code is similar to ResignedStateTest and VotedStateTest.


dengziming:
@hachikuji @guozhangwang ,Hi, PTAL.

hachikuji (approved):
LGTM. Thanks for the patch!

PR #9637, opened 2020-11-21
MONOR; Wrong command line suggestion

dengziming:
When I launch Kafka in IDE or command line, it will print `USAGE: java [options] KafkaServer server.properties [--override property=value]*`, but the real command is `java [options] kafka.Kafka server.properties`.


chia7712 (approved):
@dengziming Thanks for your patch! LGTM. Merge it to trunk

PR #9638, opened 2020-11-22
MINOR; Remove duplicate updateLeaderEndOffsetAndTimestamp

dengziming:
The `KafkaRaftClient.onBecomeLeader` will invoke `appendLeaderChangeMessage`, the call stack are:
```
1. appendLeaderChangeMessage
    1.1 flushLeaderLog
        1.1.1 updateLeaderEndOffsetAndTimestamp
        1.1.2 log.flush()
```
so the `updateLeaderEndOffsetAndTimestamp ` is already invoked, and `updateLeaderEndOffsetAndTimestamp` should only be invoked after leo change(or time change), since `log.flush()` will not change leo(and time),  it's unnessary to invoke twice.


dengziming:
ping @hachikuji to have a look.

PR #9639, opened 2020-11-22
KAFKA-10677; Complete fetches in purgatory immediately after resigning

dengziming:

If the condition of fetch is satisfied, `BROKER_NOT_AVAILABLE` or `NOT_LEADER_OR_FOLLOWER` is returned when the leader is shutting down. so we just return `BROKER_NOT_AVAILABLE` with a message.

*Summary of testing strategy*
A simple unit test to verify fetches in purgatory is completed after resigning.

### DIFF
--- a/raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
+++ b/raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
@@ -31,6 +31,7 @@ import org.apache.kafka.common.record.MutableRecordBatch;
 import org.apache.kafka.common.record.Record;
 import org.apache.kafka.common.record.RecordBatch;
 import org.apache.kafka.common.record.Records;
+import org.apache.kafka.common.record.SimpleRecord;
 import org.apache.kafka.common.requests.DescribeQuorumRequest;
 import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.test.TestUtils;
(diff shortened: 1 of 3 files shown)


dengziming:
ping @hachikuji to have a look.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -207,6 +208,41 @@ public void testEndQuorumEpochRetriesWhileResigned() throws Exception {
...
+        assertEquals(context.client.numWaitingFetch(), 1);
+
+        // append some record, but the fetch in purgatory will still fail
+        context.log.appendAsLeader(Collections.singleton(new SimpleRecord("raft".getBytes())), epoch);
+
+        // when transition to resign, all request in fetchPurgatory will fail
+        context.client.shutdown(1000);
+        context.client.poll();
+        assertTrue(context.client.quorumState().isResigned());
+        assertEquals(context.client.numWaitingFetch(), 0);
```
hachikuji: I wonder if we need this. Is it not good enough to verify the response was received?

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -207,6 +208,41 @@ public void testEndQuorumEpochRetriesWhileResigned() throws Exception {
...
+        context.client.poll();
+        assertEquals(context.client.numWaitingFetch(), 1);
+
+        // append some record, but the fetch in purgatory will still fail
+        context.log.appendAsLeader(Collections.singleton(new SimpleRecord("raft".getBytes())), epoch);
+
+        // when transition to resign, all request in fetchPurgatory will fail
+        context.client.shutdown(1000);
+        context.client.poll();
+        assertTrue(context.client.quorumState().isResigned());
```
hachikuji: Maybe we can add a helper to `RaftClientTestContext` similar to `assertElectedLeader` and get rid of `quorumState()`?

hachikuji (commented):
@dengziming Thanks for the patch. Just had a couple minor comments on the test case.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
dengziming: This is just a double-check, it's better to be removed.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
dengziming: Thank you for your suggestion, I added a method to `RaftClientTestContext`.

hachikuji (approved):
LGTM. Note I pushed a trivial alignment fix in `RaftClientTestContext`.

PR #9640, opened 2020-11-22
KAFKA-10283; Consolidate client-level and consumer-level assignment within ClientState

highluck:
KAFKA-10283 Consolidate client-level and consumer-level assignment within ClientState
https://issues.apache.org/jira/browse/KAFKA-10283


highluck:
@guozhangwang 
Please review
thanks!

guozhangwang:
@ableegoldman could you also take a look at this PR? Thanks!

highluck:
@ableegoldman 
please review!
thanks!

highluck:
**AS-IS log**
Client {clientId} per-consumer assignment:
prev owned active {taskIds...}
prev owned standby  {taskIds...}
assigned active  {taskIds...}
trevoking active  {taskIds...}
tassigned standby  {taskIds...}

**TO-BE log**
Client {clientId} per-consumer assignment:
taskId : {taskId} state : {state}
taskId : {taskId} state : PREV OWNED ACTIVE
taskId : {taskId} state : PREV OWNED STANDBY
taskId : {taskId} state : PREV OWNED ACTIVE
....

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/ClientState.java
```diff
@@ -54,15 +58,17 @@
     private final Map<String, Set<TaskId>> consumerToPreviousStatefulTaskIds = new TreeMap<>();
 
     // the following four maps are used only for logging purposes;
```
ableegoldman: no need for this comment anymore

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/ClientState.java
```diff
@@ -54,15 +58,17 @@
...
-    // TODO KAFKA-10283: we could consider merging them with other book-keeping maps at client-levels
-    //                   so that they would not be inconsistent
-    private final Map<String, Set<TaskId>> consumerToPreviousActiveTaskIds = new TreeMap<>();
-    private final Map<String, Set<TaskId>> consumerToAssignedActiveTaskIds = new TreeMap<>();
-    private final Map<String, Set<TaskId>> consumerToAssignedStandbyTaskIds = new TreeMap<>();
-    private final Map<String, Set<TaskId>> consumerToRevokingActiveTaskIds = new TreeMap<>();
-
+    private final Map<String, Map<TaskId, Set<ConsumerState>>> consumerToTaskIdsConsumerStates = new TreeMap<>();
```
ableegoldman: This map is great as a concise & efficient way to keep track of the tasks, but I'm worried that we're sacrificing some readability. I'm not sure we necessarily want to stuff everything into a single data structure, just reduce the duplication of effort in tracking tasks at both the client and consumer level. If you take a look at the `activeTasks` family of sets above, that's basically just a superset of all tasks in the `consumerToAssignedActive` map.

But `consumerToAssignedActive` and `consumerToAssignedStandby`, for example, actually track two completely orthogonal things, so I think it seems reasonable for them to be in two different data structures. WDYT?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignor.java
```diff
@@ -996,18 +996,19 @@ private void populatePartitionsByHostMaps(final Map<HostInfo, Set<TopicPartition
...
+                                         .entrySet()
+                                         .stream()
+                                         .map(e -> "\tconsumer : " + e.getKey() + '\n' +
+                                                   e.getValue().entrySet()
+                                                    .stream()
+                                                    .map(taskIdEntry -> "\ttaskId : " + taskIdEntry.getKey()
+                                                                        + " states : " + taskIdEntry.getValue())
+                                                    .collect(Collectors.joining("\n")))
+                                         .collect(Collectors.joining("\n"))
+            );
```
ableegoldman: This will print everything out with one task per line, right? That might be a problem (or at least, an annoyance) in apps with hundreds or more tasks in total. Also, just speaking from experience in debugging an application, it's more useful to present the information in terms of "here is all the assigned active tasks, here are the revoking active, etc..." just so you can see all the tasks of each type together. So we may want to keep the original format, or something similar at least

highluck:
@ableegoldman
thanks for review 
i updated code! 
please review thanks!

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/ClientState.java
```diff
@@ -53,16 +56,17 @@
     private final Map<TopicPartition, String> ownedPartitions = new TreeMap<>(TOPIC_PARTITION_COMPARATOR);
     private final Map<String, Set<TaskId>> consumerToPreviousStatefulTaskIds = new TreeMap<>();
 
-    // the following four maps are used only for logging purposes;
-    // TODO KAFKA-10283: we could consider merging them with other book-keeping maps at client-levels
-    //                   so that they would not be inconsistent
```
ableegoldman, line 58: I think maybe @guozhangwang and I haven't really explained the purpose of this cleanup ticket very well...it's Guozhang's ticket so I'll let him correct me if I've misunderstood the ticket as well, but my impression was that we wanted to consolidate the client-level sets -- eg `activeTasks` or `prevActiveTasks`, on lines 46-49 in the current code -- with the consumer-level maps, eg `consumerToPrevActiveTaskIds`. So the point is not so much to combine any of the maps on lines 59 - 62, but instead to combine each map with its corresponding set in the ones on lines 46 - 49. Does that make sense?

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/ClientState.java
```diff
@@ -16,28 +16,31 @@
...
+import java.util.Map.Entry;
 import java.util.Set;
-import java.util.stream.Collectors;
 import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
-import static java.util.Collections.emptyMap;
```
ableegoldman: Looks like you're IDE might have a conflicting style for import order. There's no "official" import order in the project, but in general we put static imports last rather than first. 

There's actually an ongoing KIP/PR for cleaning this up but obviously it's a pretty big undertaking, so it's been in limbo a while. For now it may be best to just avoid reordering any of the imports, and just follow the existing order in the file

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/ClientState.java
highluck: Oh, I was misunderstanding
@ableegoldman 

Integrating two structures, for example

`class ClientStateTask {
   private final Set<TaskId> clientTaskIds
  private filnal Map<String, Set<TaskId>> consumerTaskIds
}`

Is this form correct?
thanks!

highluck: @ableegoldman 
thanks for review
i updated code
add `class ClientStateTask`
Thanks!

guozhangwang: @ableegoldman yes you're right. On the high-level I want to consolidate sets that *always* get updated at the same time, since it is vulnerable to bugs when we forget to update some of them while modifying the others. But for maps that are logically orthogonal (i.e. they are not always updated at the same time) then we could still keep them separated.

guozhangwang:
Thanks @ableegoldman for your review and @highluck for your patience! I left a comment confirming my original intention. Would leave to Sophie to decide when it is mergable.

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/ClientState.java
```diff
@@ -100,40 +94,41 @@ boolean reachedCapacity() {
...
     }
 
     public void assignActiveTasks(final Collection<TaskId> tasks) {
-        activeTasks.addAll(tasks);
+        assignedActiveTasks.taskIds().addAll(tasks);
     }
 
     public void assignActiveToConsumer(final TaskId task, final String consumer) {
-        consumerToAssignedActiveTaskIds.computeIfAbsent(consumer, k -> new HashSet<>()).add(task);
+        assignedActiveTasks.consumerToTaskIds().computeIfAbsent(consumer, k -> new HashSet<>()).add(task);
```
ableegoldman: Thanks for the update, I think we're closer but still not quite there yet since we've kind of just wrapped the existing data structures up into the new `ClientStateTask` class but still access those maps/sets like we did before. So there's not much added safety/consistency, which was the original intention of this ticket IIUC -- ie to make sure that `activeTasks` and `consumerToAssignedActiveTaskIds` hold the same set of tasks overall.

Granted, this might turn out to be pretty difficult to do without a large refactoring of the assignment code. Part of the problem is just that we first assigned tasks at the client-level, at which point we don't know anything about the consumers for each task. Then later we assign tasks to consumers from inside the `StreamsPartitionAssignor` class, at which point we don't really care so much about the client-level tasks.

But I think we can still make some solid improvements here. For example what if we forget about refactoring any of the data structures for now, and just focus on adding some sanity checks -- eg inside `assignActiveToConsumer` we make sure that the `task` actually is assigned to this client to begin with, ie that  it exists in the `activeTasks` set. WDYT, does that make sense?

highluck: @ableegoldman thanks for review!
I have a question.
1. Do you mean that the same structure is assigned to the client level when assigned to ActiveToConsumer?
2. Do you mean that the set of TaskId and Client TaskId should be the same as the Consumer's set?

highluck: ping @ableegoldman 
thanks :)

guozhangwang:
@ableegoldman are you still on top of reviewing this PR?

ableegoldman:
@guozhangwang @highluck Sorry, I've been swamped with other things lately. If you have a minute to review it yourself Guozhang I would appreciate it! Also ping @cadonna or @vvcephei for reviews since they're familiar with this part of the code

guozhangwang:
> @guozhangwang @highluck Sorry, I've been swamped with other things lately. If you have a minute to review it yourself Guozhang I would appreciate it! Also ping @cadonna or @vvcephei for reviews since they're familiar with this part of the code

Thanks @ableegoldman , I will review it :)

guozhangwang:
@highluck I read through the code, and it looks good to me overall. However the unit test failure seems relevant: 

```
org.apache.kafka.streams.processor.internals.assignment.ClientStateTest.shouldReturnAssignedTasksForConsumer
```

Could you check and resolve it before we can merge?

highluck:
@guozhangwang 
There was an error in the code. Now I have fixed it and the tests are running fine.
Thanks for review!

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/ClientState.java
guozhangwang: @highluck I think the main point here is that, inside the `assignActiveToConsumer` we first make sure that the `task` is already in the `ClientStateTask.taskIds` set, i.e. that the task is already assigned to the client before we assign it to the client's consumer in case any bugs caused inconsistent assignment: for example, we assigned the task to instance `A` but also at the same time assign it to consumer `2` of instance `B`: currently the data structure still does not guarantee that would never happen, but at least we can add such sanity check for now.

guozhangwang (commented):
@highluck Just one more comment reflecting @ableegoldman 's comment, otherwise LGTM!

highluck:
@guozhangwang
The code has been modified.
Thank you for the review

streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/ClientState.java
```diff
@@ -110,7 +110,10 @@ public void assignActiveTasks(final Collection<TaskId> tasks) {
     }
 
     public void assignActiveToConsumer(final TaskId task, final String consumer) {
-        assignedActiveTasks.consumerToTaskIds().computeIfAbsent(consumer, k -> new HashSet<>()).add(task);
+        if (assignedActiveTasks.taskIds().contains(task)) {
+            assignedActiveTasks.consumerToTaskIds()
+                               .computeIfAbsent(consumer, k -> new HashSet<>()).add(task);
```
guozhangwang: If the condition does not satisfy, then we should throw an illegal state exception indicating a bug.

highluck:
@guozhangwang
The code has been modified.
If the phrase is strange, please tell me Thanks!

guozhangwang:
@highluck I'm re-triggering the unit test again, at the mean time, could you also run them locally and see if the failed tests are consistent?

highluck:
@guozhangwang 
Yes, I checked, but the logic was not changed except for assignActiveToConsumer, so the test is consistent.
There seems to be no problem locally!

guozhangwang:
LGTM! And merged to trunk. Thanks @highluck

PR #9641, opened 2020-11-23
MINOR: Convert connect assignment schemas to use generated protocol

chia7712:
as title.


connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/ConnectProtocol.java
```diff
@@ -43,96 +27,9 @@
...
-    public static final String CONFIG_OFFSET_KEY_NAME = "config-offset";
-    public static final String CONNECTOR_KEY_NAME = "connector";
-    public static final String LEADER_KEY_NAME = "leader";
-    public static final String LEADER_URL_KEY_NAME = "leader-url";
-    public static final String ERROR_KEY_NAME = "error";
-    public static final String TASKS_KEY_NAME = "tasks";
-    public static final String ASSIGNMENT_KEY_NAME = "assignment";
     public static final int CONNECTOR_TASK = -1;
-
     public static final short CONNECT_PROTOCOL_V0 = 0;
```
dengziming: The `CONNECT_PROTOCOL_V0` is only used in `ConnectProtocolCompatibility.fromProtocolVersion()`, I think we shouldn't depend on the version to compute `ConnectProtocolCompatibility` because every time we bump the version we should change the method, maybe we can add a field in the ExtendedWorkerMetadata.json to represent the corresponding `ConnectProtocolCompatibility`.

connect/runtime/src/main/resources/common/message/ConnectAssignment.json
```diff
@@ -0,0 +1,35 @@
...
+{
+  // Assignments for each worker are a set of connectors and tasks. These are categorized by
+  // connector ID. A sentinel task ID (CONNECTOR_TASK) is used to indicate the connector itself
+  // (i.e. that the assignment includes responsibility for running the Connector instance in
+  // addition to any tasks it generates).
+  "type": "data",
+  "name": "ConnectAssignment",
+  "validVersions": "0",
+  "fields": [
+    { "name": "error", "type": "int16", "versions": "0" },
```
dengziming: We can add some annotations here.

ijuma:
cc @kkonstantine

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/ConnectProtocol.java
chia7712: I feel the version is bump only if we add new protocol (see https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/WorkerCoordinator.java#L190). 

For another, your comment inspires me that the constant field can be moved to ```ConnectProtocolCompatibility``` as it is used by ```ConnectProtocolCompatibility``` only

connect/runtime/src/main/resources/common/message/ConnectAssignment.json
```diff
@@ -0,0 +1,36 @@
...
+{
+  // Assignments for each worker are a set of connectors and tasks. These are categorized by
+  // connector ID. A sentinel task ID (CONNECTOR_TASK) is used to indicate the connector itself
+  // (i.e. that the assignment includes responsibility for running the Connector instance in
+  // addition to any tasks it generates).
+  "type": "data",
+  "name": "ConnectAssignment",
+  "validVersions": "0",
+  "fields": [
+    // the error code of this assignment; 0 signals successful assignment
```
dengziming: Why not add the note to the `about` field of the json?

chia7712: oh, I misunderstood your comment. Will copy that.

chia7712:
convert this patch to draft as some  system tests can't pass on my local

PR #9643, opened 2020-11-23
MINOR: Upgrade to Scala 2.13.4

ijuma:
Scala 2.13.4 restores default global `ExecutionContext` to 2.12 behavior
(to fix a perf regression in some use cases) and improves pattern matching
(especially exhaustiveness checking). Most of the changes are related
to the latter as I have enabled the newly introduced `-Xlint:strict-unsealed-patmat`.

More details on the code changes:
* Don't swallow exception in `ReassignPartitionsCommand.topicDescriptionFutureToState`.
* `RequestChannel.Response` should be `sealed`.
* Introduce sealed ClientQuotaManager.BaseUserEntity to avoid false positive
exhaustiveness warning.
* Handle a number of cases where pattern matches were not exhaustive:
either by marking them with @unchecked or by adding a catch-all clause.
* Workaround scalac bug related to exhaustiveness warnings in ZooKeeperClient
* Remove warning suppression annotations related to the optimizer that are no
longer needed in ConsumerGroupCommand and AclAuthorizer.
* Use `forKeyValue` in `AclAuthorizer.acls` as the scala bug preventing us from
using it seems to be fixed.
* Also update scalaCollectionCompat to 2.3.0, which includes minor improvements.

Full release notes:
* https://github.com/scala/scala/releases/tag/v2.13.4
* https://github.com/scala/scala-collection-compat/releases/tag/v2.3.0

### DIFF
--- a/core/src/main/scala/kafka/admin/ConsumerGroupCommand.scala
+++ b/core/src/main/scala/kafka/admin/ConsumerGroupCommand.scala
@@ -44,8 +44,6 @@ import org.apache.kafka.common.requests.ListOffsetResponse
 import org.apache.kafka.common.ConsumerGroupState
 import joptsimple.OptionException
 
-import scala.annotation.nowarn
-
 object ConsumerGroupCommand extends Logging {
 
   def main(args: Array[String]): Unit = {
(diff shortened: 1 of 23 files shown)


core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2418,13 +2418,11 @@ class Log(@volatile private var _dir: File,
...
-        newSegments.foreach { splitSegment =>
-          splitSegment.close()
-          splitSegment.deleteIfExists()
-        }
-        throw e
+    } finally {
+      newSegments.foreach { splitSegment =>
+        splitSegment.close()
+        splitSegment.deleteIfExists()
+      }
```
ijuma, line 2425: @dhruvilshah3 Is there a reason why we had a try/catch with a rethrow instead of using `finally`? spotBugs complained about it and hence the question.

omkreddy (approved):
@ijuma Thanks for the PR. LGTM of tests pass,

core/src/main/scala/kafka/admin/ConsumerGroupCommand.scala
```diff
@@ -151,21 +149,22 @@ object ConsumerGroupCommand extends Logging {
...
       val clazz = implicitly[ClassTag[T]].runtimeClass
       mapper.writerFor(clazz).`with`(schema)
     }
-    private def getSchema[T <: CsvRecord: ClassTag] = {
+    private def getSchema[T <: CsvRecord : ClassTag] = {
       val clazz = implicitly[ClassTag[T]].runtimeClass
-      val fields = clazz match {
-        case _ if classOf[CsvRecordWithGroup] == clazz => CsvRecordWithGroup.fields
-        case _ if classOf[CsvRecordNoGroup]   == clazz => CsvRecordNoGroup.fields
+      val fields = {
```
chia7712: useless brackets

chia7712 (approved):
+1 and a trivial comment.

core/src/main/scala/kafka/admin/ConsumerGroupCommand.scala
ijuma: Fixed.

ijuma:
I reverted the Scala 2.12.13 change, looks like this hasn't been published to Maven yet.

ijuma:
Tests passed, merging.

PR #9644, opened 2020-11-23
KAFKA-10565: Only print console producer prompt with a tty

tombentley:
This fixes https://issues.apache.org/jira/browse/KAFKA-10565 by using `System.console` to determine whether we're connected to a TTY before printing the `kafka-console-producer.sh` prompt. That's not a perfect solution (since stdin might be connected to the tty but not stdout, or vice versa), but it's still better that printing many prompts when reading from a pipe.

### DIFF
--- a/core/src/main/scala/kafka/tools/ConsoleProducer.scala
+++ b/core/src/main/scala/kafka/tools/ConsoleProducer.scala
@@ -266,6 +266,7 @@ object ConsoleProducer {
     var keySeparator = "\t"
     var ignoreError = false
     var lineNumber = 0
+    var printPrompt = System.console != null
 
     override def init(inputStream: InputStream, props: Properties): Unit = {
       topic = props.getProperty("topic")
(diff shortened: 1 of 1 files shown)


tombentley:
@chia7712 might you be able to review this? It's trivial, but I couldn't find a good way to test it except manually.

chia7712:
> I couldn't find a good way to test it except manually.

@tombentley Could you attach the console screenshot ? I will test this PR manually later.

tombentley:
@chia7712 current behaviour:

Screenshot from 2020-11-24 10-40-59


With this patch:
Screenshot from 2020-11-24 10-45-06

chia7712:
@tombentley Thanks for your testing. Is it worth having a configuration to disable/enable to print prompt? The check of ```System.console``` is useful and the configuration make it more flexible. WDYT?

tombentley:
@chia7712 what's the use case for when you'd need to override the effects of using `System.console`? It would require a KIP, so adding a config is quite a lot more effort and I can't immediately see the benefit, but maybe you have something in mind.

chia7712 (approved):
> what's the use case for when you'd need to override the effects of using System.console? It would require a KIP, so adding a config is quite a lot more effort and I can't immediately see the benefit, but maybe you have something in mind.

oh, I have no use case for that. just a question :)

patch LGTM. Will commit it tomorrow if no objection.

PR #9647, opened 2020-11-24
KAFKA-10500: Remove thread

wcarlson5:
Add the ability to remove running threads

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -946,6 +946,41 @@ public class KafkaStreams implements AutoCloseable {
         }
     }
 
+    /**
+     * Removes one stream thread out of the running stream threads from this Kafka Streams client.
+     *
+     * The removed stream thread is gracefully shut down. This method does not specify which stream
+     * thread is shut down.
+     *
+     * Since the number of stream threads decreases, the sizes of the caches in the remaining stream
+     * threads are adapted so that the sum of the cache sizes over all stream threads equals the total
+     * cache size specified in configuration {@code cache.max.bytes.buffering}.
+     *
+     * @return name of the removed stream thread or empty if a stream thread could not be removed because
+     *         no stream threads are alive
+     */
+    public Optional<String> removeStreamThread() {
+        if (isRunningOrRebalancing()) {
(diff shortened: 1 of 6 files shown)


streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,19 +885,126 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+     *
+     * @return name of the removed stream thread or empty if a stream thread could not be removed because
+     *         no stream threads are alive
+     */
+    public Optional<String> removeStreamThread() {
+        if (isRunningOrRebalancing()) {
+            for (final StreamThread streamThread : threads) {
+                if (streamThread.isAlive()) {
+                    streamThread.shutdown();
+                    while (streamThread.isAlive() && streamThread.getName().equals(Thread.currentThread().getName())) {
```
wcarlson5, line 967: In the Kip we promise to block until the thread is finished being shutdown, I don't think this is best way to do this. However I am not sure we will really want to at all. There is also a problem if the calling thread is the thread to be shutdown.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -894,19 +885,126 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+     * threads are adapted so that the sum of the cache sizes over all stream threads equals the total
+     * cache size specified in configuration {@code cache.max.bytes.buffering}.
+     *
+     * @return name of the removed stream thread or empty if a stream thread could not be removed because
+     *         no stream threads are alive
+     */
+    public Optional<String> removeStreamThread() {
+        if (isRunningOrRebalancing()) {
+            for (final StreamThread streamThread : threads) {
+                if (streamThread.isAlive()) {
```
wcarlson5, line 965: In the kip we say remove a running thread, but that does not really make a lot of sense as we should be able to remove an alive thread. Just because all the threads are in partition assigned does not mean we should fail this call.

wcarlson5 (commented):
@cadonna Here is the PR for remove thread, It will look a bit wonky until the add thread option is merged

wcarlson5:
closed in favor of https://github.com/apache/kafka/pull/9695

PR #9648, opened 2020-11-24
KAFKA-10758: ProcessorTopology should only consider its own nodes when updating regex source topics

ableegoldman:
The problem is basically just that we compare two incompatible sets in ProcessorTopology#updateSourceTopics: the local `sourceNodesByName` map only contains nodes that correspond to a particular subtopology whereas the passed-in `nodeToSourceTopics` ultimately comes from the InternalTopologyBuilder's map, which contains nodes for the entire topology. So we would end up hitting the IllegalStateException thrown in #updateSourceTopics` any time we tried to update an application with more than one subtopology.

The fix is simple, we just need to ignore any source nodes that aren't part of the ProcessorTopology's subtopology.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
@@ -154,9 +154,9 @@ public abstract class AbstractTask implements Task {
     }
 
     @Override
-    public void update(final Set<TopicPartition> topicPartitions, final Map<String, List<String>> nodeToSourceTopics) {
+    public void update(final Set<TopicPartition> topicPartitions, final Map<String, List<String>> allTopologyNodesToSourceTopics) {
         this.inputPartitions = topicPartitions;
-        topology.updateSourceTopics(nodeToSourceTopics);
+        topology.updateSourceTopics(allTopologyNodesToSourceTopics);
     }
 
     @Override
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorTopology.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorTopology.java
@@ -149,24 +149,30 @@ public class ProcessorTopology {
         return false;
     }
 
-    public void updateSourceTopics(final Map<String, List<String>> sourceTopicsByName) {
-        if (!sourceTopicsByName.keySet().equals(sourceNodesByName.keySet())) {
-            log.error("Set of source nodes do not match: \n" +
-                "sourceNodesByName = {}\n" +
-                "sourceTopicsByName = {}",
-                sourceNodesByName.keySet(), sourceTopicsByName.keySet());
-            throw new IllegalStateException("Tried to update source topics but source nodes did not match");
-        }
+    public void updateSourceTopics(final Map<String, List<String>> allSourceTopicsByNodeName) {
         sourceNodesByTopic.clear();
-        for (final Map.Entry<String, List<String>> sourceEntry : sourceTopicsByName.entrySet()) {
-            final String nodeName = sourceEntry.getKey();
-            for (final String topic : sourceEntry.getValue()) {
-                if (sourceNodesByTopic.containsKey(topic)) {
-                    throw new IllegalStateException("Topic " + topic + " was already registered to source node "
-                        + sourceNodesByTopic.get(topic).name());
-                }
-                sourceNodesByTopic.put(topic, sourceNodesByName.get(nodeName));
+        for (final Map.Entry<String, SourceNode<?, ?, ?, ?>> sourceNodeEntry : sourceNodesByName.entrySet()) {
+            final String sourceNodeName = sourceNodeEntry.getKey();
+            final SourceNode<?, ?, ?, ?> sourceNode = sourceNodeEntry.getValue();
+
+            final List<String> updatedSourceTopics = allSourceTopicsByNodeName.get(sourceNodeName);
+            if (updatedSourceTopics == null) {
+                log.error("Unable to find source node {} in updated topics map {}",
+                          sourceNodeName, allSourceTopicsByNodeName);
+                throw new IllegalStateException("Node " + sourceNodeName + " not found in full topology");
             }
+
+            log.trace("Updating source node {} with new topics {}", sourceNodeName, updatedSourceTopics);
+            for (final String topic : updatedSourceTopics) {
+                if (sourceNodesByTopic.containsKey(topic)) {
+                    log.error("Tried to subscribe topic {} to two nodes when updating topics from {}",
+                              topic, allSourceTopicsByNodeName);
+                    throw new IllegalStateException("Topic " + topic + " was already registered to source node "
+                                                        + sourceNodesByTopic.get(topic).name());
+                }
+                sourceNodesByTopic.put(topic, sourceNode);
+            }
+
         }
     }
 
--- a/streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
@@ -31,6 +31,7 @@ import org.apache.kafka.streams.KafkaStreams;
 import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.StreamsBuilder;
 import org.apache.kafka.streams.StreamsConfig;
+import org.apache.kafka.streams.Topology;
 import org.apache.kafka.streams.TopologyWrapper;
 import org.apache.kafka.streams.errors.StreamsUncaughtExceptionHandler;
 import org.apache.kafka.streams.integration.utils.EmbeddedKafkaCluster;
@@ -72,6 +73,7 @@ import static org.apache.kafka.streams.integration.utils.IntegrationTestUtils.st
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.greaterThan;
 
 /**
  * End-to-end integration test based on using regex and named topics for creating sources, using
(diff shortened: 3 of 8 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/AbstractTask.java
```diff
@@ -154,9 +154,9 @@ final void transitionTo(final Task.State newState) {
     }
 
     @Override
-    public void update(final Set<TopicPartition> topicPartitions, final Map<String, List<String>> nodeToSourceTopics) {
```
ableegoldman, line 157: Since the root cause of this bug was basically just confusion over what exactly this set contains, a renaming feels in order

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorTopology.java
```diff
@@ -149,24 +149,30 @@ public boolean hasPersistentGlobalStore() {
...
-                "sourceTopicsByName = {}",
-                sourceNodesByName.keySet(), sourceTopicsByName.keySet());
-            throw new IllegalStateException("Tried to update source topics but source nodes did not match");
-        }
+    public void updateSourceTopics(final Map<String, List<String>> allSourceTopicsByNodeName) {
         sourceNodesByTopic.clear();
-        for (final Map.Entry<String, List<String>> sourceEntry : sourceTopicsByName.entrySet()) {
-            final String nodeName = sourceEntry.getKey();
-            for (final String topic : sourceEntry.getValue()) {
+        for (final Map.Entry<String, SourceNode<?, ?, ?, ?>> sourceNodeEntry : sourceNodesByName.entrySet()) {
```
ableegoldman, line 154: In addition to removing the fault check, I slightly refactored this loop so that we only loop over the source nodes in this particular subtopology. Previously we would have added entries for all source nodes across the entire topology to our `sourceNodesByTopic` map

streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
```diff
@@ -198,9 +200,16 @@ public void testRegexRecordsAreProcessedAfterReassignment() throws Exception {
 
             final StreamsBuilder builder = new StreamsBuilder();
             final KStream<String, String> pattern1Stream = builder.stream(Pattern.compile("TEST-TOPIC-\\d"));
-            pattern1Stream.to(outputTopic, Produced.with(Serdes.String(), Serdes.String()));
+            final KStream<String, String> otherStream = builder.stream(Pattern.compile("not-a-match"));
```
ableegoldman, line 205: We didn't catch the bug in this test for two reasons: it has only one subtopology, and it didn't wait for Streams to get to RUNNING before it created the new topic. So we weren't even covering the "update source topics" code path since all topics existed by the first assignment

streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
```diff
@@ -198,9 +200,16 @@ public void testRegexRecordsAreProcessedAfterReassignment() throws Exception {
...
+
+            pattern1Stream
```
ableegoldman, line 207: Technically it's sufficient to just add the second KStream above for a multi-subtopology application, but I felt the test coverage could only stand to benefit with (slightly) more complicated examples

mjsax (approved):
Thanks for the swift fix @ableegoldman

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorTopology.java
```diff
@@ -149,24 +149,30 @@ public boolean hasPersistentGlobalStore() {
...
 
-    public void updateSourceTopics(final Map<String, List<String>> sourceTopicsByName) {
-        if (!sourceTopicsByName.keySet().equals(sourceNodesByName.keySet())) {
-            log.error("Set of source nodes do not match: \n" +
-                "sourceNodesByName = {}\n" +
-                "sourceTopicsByName = {}",
-                sourceNodesByName.keySet(), sourceTopicsByName.keySet());
-            throw new IllegalStateException("Tried to update source topics but source nodes did not match");
-        }
+    public void updateSourceTopics(final Map<String, List<String>> allSourceTopicsByNodeName) {
```
cadonna, line 152: Could you also add a unit test that verifies that the issue is gone?

streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
```diff
@@ -198,9 +200,16 @@ public void testRegexRecordsAreProcessedAfterReassignment() throws Exception {
...
+                .selectKey((k, v) -> k)
+                .groupByKey()
+                .aggregate(() -> "", (k, v, a) -> v)
+                .toStream().to(outputTopic, Produced.with(Serdes.String(), Serdes.String()));
+
             streams = new KafkaStreams(builder.build(), streamsConfiguration);
```
cadonna: I would add the following assert here to ensure we have more than one sub-topology:

```
            final Topology topology = builder.build();
            assertThat(topology.describe().subtopologies().size(), greaterThan(1));
            streams = new KafkaStreams(topology, streamsConfiguration);
```

Just to make it clear that we want to test with multiple sub-topologies.

streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
```diff
@@ -198,9 +200,16 @@ public void testRegexRecordsAreProcessedAfterReassignment() throws Exception {
...
             final StreamsBuilder builder = new StreamsBuilder();
             final KStream<String, String> pattern1Stream = builder.stream(Pattern.compile("TEST-TOPIC-\\d"));
-            pattern1Stream.to(outputTopic, Produced.with(Serdes.String(), Serdes.String()));
+            final KStream<String, String> otherStream = builder.stream(Pattern.compile("not-a-match"));
+
+            pattern1Stream
+                .selectKey((k, v) -> k)
+                .groupByKey()
+                .aggregate(() -> "", (k, v, a) -> v)
+                .toStream().to(outputTopic, Produced.with(Serdes.String(), Serdes.String()));
```
cadonna, line 211: I would even add `otherStream` like:
```suggestion
                .toStream()
                .merge(otherStream)
                .to(outputTopic, Produced.with(Serdes.String(), Serdes.String()));
```

cadonna (commented):
@ableegoldman Thank you for fixing this so quickly!

I just have some comments about testing.

streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
ableegoldman: But if we merge then won't that merge the subtopologies as well? (We would still have two subtopologies due to the upstream key-changing operation/repartition, but I wanted the test to cover different "kinds" of subtopologies like this)

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorTopology.java
ableegoldman: I modified the existing test rather than add a new one, and verified that the modified test does fail without the changes. Since the original test wasn't actually testing anything prior to the modifications I thought it made sense to fix the test rather than add a new one and retain a useless test in addition

streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
ableegoldman: Ack good call

streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
cadonna: You would have a source node in one sub-topology an a source node in the other sub-topology. I thought that was the pattern in the bug report, but I now realized that the bug report uses a pattern similar to the one you specified. Wouldn't it make sense to test both?

streams/src/main/java/org/apache/kafka/streams/processor/internals/ProcessorTopology.java
cadonna: The scope of an integration test and of a unit test are quite different. The integration test specifically tests the pattern subscription whereas the unit test just tests the update of the source topics. Probably at the moment the update of the source topics is only called when pattern subscription is used, but that assumption seems rather brittle to me because   it may change in future. I think that neither of the tests is useless. If one were useless, I would keep the unit test for test execution performance reasons.

ableegoldman: Added a unit test

streams/src/test/java/org/apache/kafka/streams/integration/RegexSourceIntegrationTest.java
ableegoldman: There are two ways to have more than one source node: either reading from a different input topic/pattern, or via a repartition. I agree that we should test both of these cases, but I'd prefer to do so in a single integration test rather than in two separate integration tests, to avoid making the test suite even longer

ableegoldman:
The Java 15 tests passed, Java 8 build failed with flaky `TransactionsTest.testBumpTransactionalEpoch` and the Java 11 build failed with `StreamTableJoinTopologyOptimizationIntegrationTest.shouldDoStreamTableJoinWithDifferentNumberOfPartitions` (which is known to be flaky, confirmed that the failure was unrelated -- due to `DirectoryNotEmptyException`)

Will merge & cherrypick to older branches to unblock the 2.6.1 and 2.7.0 releases

ableegoldman:
Cherrypicked to 2.7 cc @bbejeck

ableegoldman:
Cherrypicked to 2.6 cc/ @mimaison

PR #9649, opened 2020-11-24
MINOR: Combine repeated top-level error test in AlterIsrManagerTest

dengziming:
The content of the 3 top-level errors are identical, just merge them.

### DIFF
--- a/core/src/test/scala/unit/kafka/server/AlterIsrManagerTest.scala
+++ b/core/src/test/scala/unit/kafka/server/AlterIsrManagerTest.scala
@@ -126,21 +126,30 @@ class AlterIsrManagerTest {
   }
 
   @Test
-  def testTopLevelErrors(): Unit = {
-    val errors = Seq(Errors.CLUSTER_AUTHORIZATION_FAILED, Errors.STALE_BROKER_EPOCH, Errors.UNKNOWN_SERVER_ERROR)
-    errors.foreach(error => {
-      val isrs = Seq(AlterIsrItem(tp0, new LeaderAndIsr(1, 1, List(1,2,3), 10), _ => { }, 0))
-      val manager = testTopLevelError(isrs, error)
-      // Any top-level error, we want to retry, so we don't clear items from the pending map
-      assertFalse(manager.submit(AlterIsrItem(tp0, null, _ => { }, 0)))
-    })
-
+  def testAuthorizationFailed(): Unit = {
+    checkTopLevelError(Errors.CLUSTER_AUTHORIZATION_FAILED)
   }
 
-  def testTopLevelError(isrs: Seq[AlterIsrItem], error: Errors): AlterIsrManager = {
+  @Test
+  def testStaleBrokerEpoch(): Unit = {
+    checkTopLevelError(Errors.STALE_BROKER_EPOCH)
+  }
+
+  @Test
+  def testUnknownServer(): Unit = {
+    checkTopLevelError(Errors.UNKNOWN_SERVER_ERROR)
+  }
+
+  private def checkTopLevelError(error: Errors): Unit = {
+    val isrs = Seq(AlterIsrItem(tp0, new LeaderAndIsr(1, 1, List(1,2,3), 10), _ => { }, 0))
+    val alterIsrManager = testTopLevelError(isrs, error)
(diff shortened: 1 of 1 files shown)


dengziming:
@cmccabe  hi, PTAL.

dengziming:
@mumrah  Hi, this class is created by you so I think you can have a look, thank you.

dengziming:
@mumrah Hi, David, PTAL.

chia7712 (commented):
@dengziming Thanks for your patch. It seems to me the name of test case is good to clarify the purpose. Another way to remove duplicate code is that we can add a private helper method to handle duplicate parts. For example:
```scala
def testTopLevelError(error: Error): Unit = {
  val sirs = ...
  val manager = ...
  assertFalse(...
}

@Test
def testOtherErrors(): Unit = {
  testTopLevelError(Errors.UNKNOWN_SERVER_ERROR)
}
``` 

Each test case can call the helper method to complete testing. WDYT?

dengziming:
@chia7712 ,Thank you for your suggestion. In fact, I get the idea from `AlterIsrManagerTest.testPartitionErrors` which test 3 errors in a method, so I also merge the 3 top errors in a method. I think you offer a better idea and I also changed the 3 partition level error.

core/src/test/scala/unit/kafka/server/AlterIsrManagerTest.scala
```diff
@@ -127,26 +127,24 @@ class AlterIsrManagerTest {
...
+    testTopLevelError(Errors.STALE_BROKER_EPOCH)
   }
 
   @Test
-  def testOtherErrors(): Unit = {
+  def testUnknownServer(): Unit = {
+    testTopLevelError(Errors.UNKNOWN_SERVER_ERROR)
+  }
+
+  def testTopLevelError(error: Errors): Unit = {
```
mumrah: Maybe rename this and "testPartitionError" to something slightly different so they don't look like actual test cases? Maybe something like "assertTopLevelError" or "checkTopLevelError".

mumrah (commented):
@dengziming thanks for the patch. This cleanup looks fine, just one comment inline.

core/src/test/scala/unit/kafka/server/AlterIsrManagerTest.scala
```diff
@@ -127,26 +127,24 @@ class AlterIsrManagerTest {
...
+    checkTopLevelError(Errors.STALE_BROKER_EPOCH)
   }
 
   @Test
-  def testOtherErrors(): Unit = {
+  def testUnknownServer(): Unit = {
+    checkTopLevelError(Errors.UNKNOWN_SERVER_ERROR)
+  }
+
+  def checkTopLevelError(error: Errors): Unit = {
```
chia7712: Could you change it to ```private``` method?

core/src/test/scala/unit/kafka/server/AlterIsrManagerTest.scala
```diff
@@ -174,13 +172,24 @@ class AlterIsrManagerTest {
...
+  def testUnknownTopicPartition(): Unit = {
+    checkPartitionError(Errors.UNKNOWN_TOPIC_OR_PARTITION)
+  }
+
+  @Test
+  def testNotLeaderOrFollower(): Unit = {
+    checkPartitionError(Errors.NOT_LEADER_OR_FOLLOWER)
+  }
+
+  def checkPartitionError(error: Errors): Unit = {
```
chia7712: ditto

dengziming: Thanks, done!

dengziming: Thanks, done.

chia7712:
@dengziming Could you trigger QA again?

dengziming:
@chia7712 I tried to rebase my pr and push it again, but still failed. this is weird, maybe I triggered a bug of jenkins and should tried again tomorrow.

chia7712 (approved):

chia7712:
I will merge this PR tomorrow if there is no objection. @mumrah FYI

mumrah (approved):
LGTM, thanks!

PR #9653, opened 2020-11-25
MINOR: Remove unnecessary statement from WorkerConnector#doRun

UnityLung:
Continue is unnecessary as the last statement in a loop.
So I remove it from WorkerConnector.


UnityLung:
@chia7712  Could you review it please

chia7712 (approved):
@abc863377 Thanks for your patch. LGTM

PR #9655, opened 2020-11-25
MINOR: fix listeners doc to close <code> properly

tombentley:
Fixes a problem with an incorrectly closed `<code>` in the doc for `listeners`.


tombentley:
This is how it renders without this fix

image

@bbejeck this also affects the 2.7RC3 docs, but you could always just fix the generated HTML there.

bbejeck (approved):
Good catch, thanks for the fix @tombentley.

bbejeck:
Merged 9655 into trunk

bbejeck:
cherry-picked to 2.7

PR #9656, opened 2020-11-25
repro demonstrating non-owner commit

vvcephei:


vvcephei:
Not sure if this is expected or not, but it _seems_ wrong.

When this test fails, it prints:

```
cons1 owns it
Committed 11 with assignment [inputTopicLeft-0]
Committed 8 with assignment []
{inputTopicLeft-0=OffsetAndMetadata{offset=8, leaderEpoch=null, metadata='[]'}}
```
which indicates that the consumer that does _not_ own the partition is allowed to overwrite the owner's prior commit.

PR #9658, opened 2020-11-27
Linking to config parameters

tombentley:
This demos an idea I had for improving the website docs. 

The docs often mention config parameters. With recent changes it's now possible to deep link directly to the the doc for a given config parameter. It would be useful for readers if we made use of this in our own docs, so when talking about `listeners` in prose we link to the doc for it. 

This PR achieves that in a relatively easy-to-author way using a CSS class on the `<code>` used to wrap some config parameters and some JS which takes the class and the included text and turns it into one or more links. For example `<code class="brokerconfig">listeners</code>` gets turned into a link to the doc for [`listeners`](https://kafka.apache.org/27/documentation.html#brokerconfigs_listeners), with similar classes for topic, producer consumer etc. configs.

Here a're some screenshots of what the markup changes in the patch result in, first from the upgrade section:

Screenshot from 2020-11-27 12-05-57

The script knows to ignore whatever is after the `=` sign. Here's another example from the security section (please excuse the magenta find-in-page highlighting):

Screenshot from 2020-11-27 12-06-58

Because the script checks for the existence of the link before generating the link it doesn't linkify the `username` or `password` (the regex isn't powerful enough to properly understand that lines ending with `\` are continuation lines). This is a bit of a hack, but it works well enough.

There's two potential issues I can see with this:

1. Lots of client configs are shared between both producer and consumer (and admin client...), so which should we link to? Probably a link is better than no link, but equally we could chose not to link such configs.
2. This is not really compatible with https://issues.apache.org/jira/browse/KAFKA-2967 though it might be possible to do something similar in ReStructuredText with custom roles.

### DIFF
--- a/docs/documentation.html
+++ b/docs/documentation.html
@@ -86,3 +86,4 @@
 
 <!--#include virtual="../includes/_footer.htm" -->
 <!--#include virtual="../includes/_docs_footer.htm" -->
+<script><!--#include virtual="js/kafkaConfigs.js" --></script>
--- a/docs/js/kafkaConfigs.js
+++ b/docs/js/kafkaConfigs.js
@@ -1,0 +1,28 @@
+$(document).ready(function() {
+    $.each([ "brokerconfig", "topicconfig", "producerconfig", "consumerconfig", 
+    "connectconfig", "streamsconfig", "adminclientconfig" ], function( index, value ) {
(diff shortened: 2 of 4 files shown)


tombentley:
Thoughts?

@mimaison, @chia7712, @omkreddy

PR #9660, opened 2020-11-28
Kafka 10629 - TopologyTestDriver should not require a Properties argument

rohitrmd:
- KIP-680: TopologyTestDriver should not require a Properties argument.
https://cwiki.apache.org/confluence/display/KAFKA/KIP-680%3A+TopologyTestDriver+should+not+require+a+Properties+argument
- Jira for the KIP: 
https://issues.apache.org/jira/browse/KAFKA-10629


streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
```diff
@@ -471,7 +471,7 @@ public void shouldCloseProcessor() {
 
     @Test
     public void shouldThrowForUnknownTopic() {
-        testDriver = new TopologyTestDriver(new Topology(), config);
+        testDriver = new TopologyTestDriver(new Topology());
```
chia7712, line 474: the ```config``` is not from empty properties so it may be incorrect to remove ```config```.

streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
```diff
@@ -298,8 +322,14 @@ private TopologyTestDriver(final InternalTopologyBuilder builder,
                                final long initialWallClockTimeMs) {
         final Properties configCopy = new Properties();
         configCopy.putAll(config);
-        configCopy.putIfAbsent(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy-bootstrap-host:0");
-        configCopy.putIfAbsent(StreamsConfig.APPLICATION_ID_CONFIG, "dummy-topology-test-driver-app-id");
+        if (!configCopy.containsKey(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG)) {
```
chia7712: It seems to me ```putIfAbsent``` is more readable than ```containsKey + setProperty```.

streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
chia7712: there are a lot of similar code change.

streams/src/test/java/org/apache/kafka/streams/TopologyTest.java
```diff
@@ -353,8 +349,7 @@ public void shouldThrowOnUnassignedStateStoreAccess() {
             .addProcessor(badNodeName, new LocalMockProcessorSupplier(), sourceNodeName);
 
         try {
-            new TopologyTestDriver(topology, mkProperties(
-                mkMap(mkEntry(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getAbsolutePath()))));
+            new TopologyTestDriver(topology);
```
chia7712, line 352: Is it ok to remove ```StreamsConfig.STATE_DIR_CONFIG``` here?

chia7712 (commented):
@rohitrmd Thanks for your patch. Some suggestions are left. please take a look :)

streams/src/test/java/org/apache/kafka/streams/TopologyTest.java
rohitrmd: That's a good point. This test is validating if exception is thrown when topology is created with processor node which doesn't have state store specified. Providing STATE_DIR_CONFIG config parameter in properties won't alter desired outcome of the test. Hence I removed it.

streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
rohitrmd: @chia7712 fixed it.

streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
rohitrmd: Default config in the test provides [application_id and state_dir_config](https://github.com/apache/kafka/blob/324c0ea0ae2c4ade0f5320ac2e794b020f57ab52/streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java#L145-L146) parameters. Majority of the tests like [shouldThrowForUnknownTopic](https://github.com/apache/kafka/blob/324c0ea0ae2c4ade0f5320ac2e794b020f57ab52/streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java#L473),  [shouldThrowForMissingTime](https://github.com/apache/kafka/pull/9660/files#diff-a10cd21c3ae928cc682ee13fa43fa02168fdebb5a319ccf8ecf5d69daac1de19L487),  [shouldThrowNoSuchElementExceptionForUnusedOutputTopicWithDynamicRoutingdon't](https://github.com/apache/kafka/pull/9660/files#diff-a10cd21c3ae928cc682ee13fa43fa02168fdebb5a319ccf8ecf5d69daac1de19R519),  [shouldCaptureSinkTopicNamesIfWrittenInto](https://github.com/apache/kafka/pull/9660/files#diff-a10cd21c3ae928cc682ee13fa43fa02168fdebb5a319ccf8ecf5d69daac1de19R533),  [shouldProcessRecordForTopic](https://github.com/apache/kafka/pull/9660/files#diff-a10cd21c3ae928cc682ee13fa43fa02168fdebb5a319ccf8ecf5d69daac1de19R592), [shouldSetRecordMetadata](https://github.com/apache/kafka/pull/9660/files#diff-a10cd21c3ae928cc682ee13fa43fa02168fdebb5a319ccf8ecf5d69daac1de19R604) don't require these config parameters.                                                                                                                                                                    But tests using state stores such as [shouldPopulateGlobalStore](https://github.com/apache/kafka/pull/9660/files#diff-a10cd21c3ae928cc682ee13fa43fa02168fdebb5a319ccf8ecf5d69daac1de19R911), [shouldReturnAllStores](https://github.com/apache/kafka/pull/9660/files#diff-a10cd21c3ae928cc682ee13fa43fa02168fdebb5a319ccf8ecf5d69daac1de19R1032) and [shouldThrowIfBuiltInStoreIsAccessedWithUntypedMethod](https://github.com/apache/kafka/pull/9660/files#diff-a10cd21c3ae928cc682ee13fa43fa02168fdebb5a319ccf8ecf5d69daac1de19R1156) may need specific dir config to clean up after the test. Right now I am relying on default state-dir-config value for these few tests which use state stores. I am open to change those specific tests which use state stores.

streams/test-utils/src/main/java/org/apache/kafka/streams/TopologyTestDriver.java
```diff
@@ -254,6 +265,19 @@ public TopologyTestDriver(final Topology topology,
         this(topology, config, null);
     }
 
+    /**
+     * Create a new test diver instance.
+     * Initialized the internally mocked wall-clock time with {@link System#currentTimeMillis() current system time}.
```
mjsax: This overload takes `initialWallClockTimeMs`, thus this sentence seems to be incorrect? Should go to `TopologyTestDriver(final Topology topology)` instead?

rohitrmd: @mjsax fixed the comment.

mjsax:
Thanks for the KIP and PR @rohitrmd!

mjsax:
@rohitrmd Seems we forgot to update the docs with this PR, in particular `streams/upgarde_guide.html` and maybe also the section about testing? Would you like to do a follow up PR to close this gap?

mjsax:
Follow up PR to update the docs: https://github.com/apache/kafka/pull/10048

PR #9662, opened 2020-11-30
KAFKA-10130; Rewrite FeatureZNode struct with auto-generated protocol

dengziming:
*More detailed description of your change*

1. remove FeatureZNode and replace it with FeatureZNode.json
2. Change code where FeatureZNode is used
3. copy some code of `org.apache.kafka.raft.FileBasedStateStore` to generate json data from and to FeatureZNodeData
4. I will also replace the rest ZNode with auto-generated protocol if this was approved.

*Summary of testing strategy (including rationale)*
Replace FeatureZNode  in unit test and integration test

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9664, opened 2020-12-01
KAFKA-10780; Rewrite ControllerZNode struct with auto-generated protocol

dengziming:
1. The #9662 rewrite FeatureZNode struct with auto-generated protocol, but it's a non-trivial change, so we can just review this simple pr first.
2. Copy some code of `org.apache.kafka.raft.FileBasedStateStore` to generate json data from and to FeatureZNodeData

unit test

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -830,7 +830,7 @@ project(':core') {
     args = [ "-p", "kafka.internals.generated",
              "-o", "src/generated/java/kafka/internals/generated",
              "-i", "src/main/resources/common/message",
-             "-m", "MessageDataGenerator"
+             "-m", "MessageDataGenerator", "JsonConverterGenerator"
     ]
     inputs.dir("src/main/resources/common/message")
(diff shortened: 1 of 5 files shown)


dengziming:
Hi, @abbccdda PTAL.

ijuma:
Thanks for the PR. I don't think we should do this since KIP-500 will replace all this code.

dengziming:
@ijuma Thank you, I ignored this fact. 
Some of the ZkNode will be removed because we no longer need them after KIP-500, for example `IsrChangeNotificationZNode` and `ControllerZNode`, however Some ZkNode will just be moved from ZkNode to RaftKafkaMetadataLog , for example `TopicZNode` and `FeatureZNode`.
So can we still rewrite their struct use auto-generated protocol which can also be used after KIP-500? also ping @abbccdda as this is a major direction discussion.
I will close this pr and create another for `TopicZNode` if you think it's reasonable, Thank you.

PR #9652, opened 2020-11-25
MINOR: Align the UID inside/outside container

chia7712:
The root cause of causing permission issue is that the uid outside container is not 1000 so container (has uid 1000) can't modify the file of ```/opt/kafka-dev```

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9666, opened 2020-12-01
MINOR: Remove broken `.travis.yml` with system test hooks

hachikuji:
This patch removes `.travis.yml` which has never worked as far as I know, and is lately cluttering up PR requests.


hachikuji:
cc @chia7712 I guess you were trying to get this working instead? I can close this if you prefer.

dajac:
I am +1 on removing it. We can always bring it back to trunk if we can make it work later.

chia7712:
> guess you were trying to get this working instead? 

#9652

Removing travis test is ok to me.

hachikuji:
@chia7712 Nice that you got it working! I think there is value running a subset of system tests automatically. It might even encourage us to write more system tests if we can get the feedback more easily. Not having access to logs is annoying, but might not be a dealbreaker. It looks like we need a bit of initial work to tune the build though. I'm happy to close this and merge #9652. Then perhaps we can file a few jiras for any follow-up actions. What do you think?

chia7712:
> Nice that you got it working! I think there is value running a subset of system tests automatically. It might even encourage us to write more system tests if we can get the feedback more easily. Not having access to logs is annoying, but might not be a dealbreaker. It looks like we need a bit of initial work to tune the build though. I'm happy to close this and merge #9652. Then perhaps we can file a few jiras for any follow-up actions. What do you think?

+1

hachikuji:
Closing in favor of #9652.

PR #9667, opened 2020-12-01
MINOR: Do not print log4j for memberId required

guozhangwang:
For MemberIdRequiredException, we would not print the exception at INFO with a full exception message since it may introduce more confusion that clearance.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
@@ -465,7 +465,11 @@ public abstract class AbstractCoordinator implements Closeable {
                 }
             } else {
                 final RuntimeException exception = future.exception();
-                log.info("Rebalance failed.", exception);
+
+                if (!(exception instanceof MemberIdRequiredException)) {
+                    log.info("Rebalance failed.", exception);
+                }
+
                 resetJoinGroupFuture();
                 if (exception instanceof UnknownMemberIdException ||
                     exception instanceof RebalanceInProgressException ||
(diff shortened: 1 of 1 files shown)


guozhangwang:
@hachikuji @abbccdda

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -465,7 +465,11 @@ boolean joinGroupIfNeeded(final Timer timer) {
                 }
             } else {
                 final RuntimeException exception = future.exception();
-                log.info("Rebalance failed.", exception);
+
+                if (!(exception instanceof MemberIdRequiredException)) {
```
chia7712, line 471: Is ```JoinGroupResponseHandler``` a better place to log error? For example, the error ```UNKNOWN_MEMBER_ID``` is log twice. 
(https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java#L605)

chia7712: For another, does it need some comment to explain why ```MemberIdRequiredException``` is excluded.

chia7712 (commented):
@guozhangwang Nice improvement. It makes sense to me to eliminate this "error message" as it is NOT fatal.

abbccdda (commented):
Maybe add an explanation here for why to skip the member.id required exception? (as it's transient)

guozhangwang:
> Maybe add an explanation here for why to skip the member.id required exception? (as it's transient)

Ack, will do.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
guozhangwang: I think I agree with you: it is a bit redundant. I think originally it is because this future may contain the error that is either from join-response or from sync-response, but at the moment both errors are logged inside the Join/SyncGroupHandler anyways. Just to maintain compatibility in case some trouble shooting scenarios are dependent on this summary entry for now.

abbccdda (approved):

bbejeck:
cherry-picked to 2.7

PR #9668, opened 2020-12-01
MINOR: add test for repartition/source-topic/changelog optimization

mjsax:
If topology optimization is enabled, KafkaStreams does not create store
changelog topics but re-uses source input topics if possible. However,
this optimization should not be applied to internal repartition topics,
because those are actively purged.

Call for review @ableegoldman

### DIFF
--- a/streams/src/test/java/org/apache/kafka/streams/StreamsBuilderTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/StreamsBuilderTest.java
@@ -16,7 +16,6 @@
  */
 package org.apache.kafka.streams;
 
-import java.util.regex.Pattern;
 import org.apache.kafka.common.serialization.LongSerializer;
 import org.apache.kafka.common.serialization.Serdes;
 import org.apache.kafka.common.serialization.StringSerializer;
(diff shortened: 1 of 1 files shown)


streams/src/test/java/org/apache/kafka/streams/StreamsBuilderTest.java
```diff
@@ -463,7 +463,36 @@ public void shouldReuseSourceTopicAsChangelogsWithOptimization20() {
...
+            equalTo(true)
+        );
+        assertThat(
+            internalTopologyBuilder.topicGroups().get(1).stateChangelogTopics.keySet(),
+            equalTo(Collections.singleton("appId-store-changelog"))
+        );
+    }
+
+    @Test
+    public void shouldNotReuseRepartitionTopicAsChangelogsByDefault() {
```
ableegoldman: This should still be called `shouldNotReuseSourceTopicAsChangelogsByDefault`,  right?

ableegoldman (approved):
LGTM, modulo one question about the existing test name. Thanks for the test coverage to verify this!

streams/src/test/java/org/apache/kafka/streams/StreamsBuilderTest.java
mjsax: Ah. Yes, the test name should not change.

PR #5020, opened 2018-05-15
KAFKA-6566: Improve Connect Resource Cleanup

rayokota:
This is a change to improve resource cleanup for sink tasks and source tasks.  Now `Task.stop()` is called from both `WorkerSinkTask.close()` and `WorkerSourceTask.close()`.

It is called from `WorkerXXXTask.close()` since this method is called in the `finally` block of `WorkerTask.run()`, and Connect developers use `stop()` to clean up resources.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9669, opened 2020-12-01
KAFKA-10792: Prevent source task shutdown from blocking herder thread

C0urante:
[Jira](https://issues.apache.org/jira/browse/KAFKA-10792)

The functional changes are simple: change the `WorkerSourceTask` class to only call `SourceTask::stop` from one location, during task shutdown, and only if an attempt has been made to start the task (which will not be the case if it was created in the paused state and then shut down before being started). This is important in order to prevent `SourceTask::stop` from being indirectly invoked on the herder's thread, which can have adverse effects if the task is unable to shut down promptly.

Unit tests are tweaked where necessary to account for this new logic, which covers some edge cases mentioned in https://github.com/apache/kafka/pull/5020 that were unaddressed up until now.

The existing integration tests for blocking connectors are expanded to also include cases for blocking source and sink tasks. Full coverage of every source/sink task method is intentionally omitted from these expanded tests in order to avoid inflating test runtime (each one adds an extra 5 seconds at minimum) and because the tests that are added here were sufficient to reproduce the bug with source task shutdown.

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
@@ -103,7 +103,6 @@ class WorkerSourceTask extends WorkerTask {
 
     private Map<String, String> taskConfig;
     private boolean started = false;
-    private boolean stopped = false;
 
     public WorkerSourceTask(ConnectorTaskId id,
                             SourceTask task,
@@ -165,7 +164,13 @@ class WorkerSourceTask extends WorkerTask {
 
     @Override
     protected void close() {
-        tryStop();
+        if (started) {
+            try {
+                task.stop();
+            } catch (Throwable t) {
+                log.warn("Could not stop task", t);
+            }
+        }
         if (producer != null) {
             try {
                 producer.close(Duration.ofSeconds(30));
@@ -203,20 +208,6 @@ class WorkerSourceTask extends WorkerTask {
     public void stop() {
         super.stop();
         stopRequestedLatch.countDown();
-    }
-
-    // Note: This method is not thread-safe
-    private void tryStop() {
-        // If the task is scheduled for shutdown before we invoke initialize or start on it (which
-        // can happen reliably if it's started in the PAUSED state), we don't have to invoke stop on it
-        if (!stopped && started) {
-            try {
-                task.stop();
-                stopped = true;
-            } catch (Throwable t) {
-                log.warn("Could not stop task", t);
-            }
-        }
     }
 
     @Override
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
@@ -102,8 +102,7 @@ class WorkerSourceTask extends WorkerTask {
     private CountDownLatch stopRequestedLatch;
 
     private Map<String, String> taskConfig;
-    private boolean finishedStart = false;
-    private boolean startedShutdownBeforeStartCompleted = false;
+    private boolean started = false;
     private boolean stopped = false;
 
     public WorkerSourceTask(ConnectorTaskId id,
@@ -166,9 +165,7 @@ class WorkerSourceTask extends WorkerTask {
 
     @Override
     protected void close() {
-        if (!shouldPause()) {
-            tryStop();
-        }
+        tryStop();
         if (producer != null) {
             try {
                 producer.close(Duration.ofSeconds(30));
@@ -206,16 +203,13 @@ class WorkerSourceTask extends WorkerTask {
     public void stop() {
         super.stop();
         stopRequestedLatch.countDown();
-        synchronized (this) {
-            if (finishedStart)
-                tryStop();
-            else
-                startedShutdownBeforeStartCompleted = true;
-        }
     }
 
-    private synchronized void tryStop() {
-        if (!stopped) {
+    // Note: This method is not thread-safe
+    private void tryStop() {
+        // If the task is scheduled for shutdown before we invoke initialize or start on it (which
+        // can happy reliably if it's started in the PAUSED state), we don't have to invoke stop on it
+        if (!stopped && started) {
             try {
                 task.stop();
                 stopped = true;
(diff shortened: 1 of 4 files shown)


C0urante:
@tombentley @gharris1727 @ncliang care to take a look?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
```diff
@@ -206,16 +203,13 @@ public void cancel() {
...
-                startedShutdownBeforeStartCompleted = true;
-        }
     }
 
-    private synchronized void tryStop() {
-        if (!stopped) {
+    // Note: This method is not thread-safe
+    private void tryStop() {
+        // If the task is scheduled for shutdown before we invoke initialize or start on it (which
+        // can happy reliably if it's started in the PAUSED state), we don't have to invoke stop on it
```
ncliang:
```suggestion
        // can happen reliably if it's started in the PAUSED state), we don't have to invoke stop on it
```

ncliang (approved):
LGTM. Thanks @C0urante .

C0urante:
Thanks, Nigel.

@chia7712 @rhauch @kkonstantine would you like to take a look?

tombentley (approved):
LGTM.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
```diff
@@ -206,16 +203,13 @@ public void cancel() {
...
-            if (finishedStart)
-                tryStop();
-            else
-                startedShutdownBeforeStartCompleted = true;
-        }
     }
 
-    private synchronized void tryStop() {
-        if (!stopped) {
+    // Note: This method is not thread-safe
```
rhauch: Maybe a nit, but this comment seems a bit misleading. It sounds as if the method need be called from a thread safe point, but otherwise there are no restrictions on how it's used.

Strictly speaking, this method need not be thread safe because it should *only* be called from within `close()`, which is called from the task's own thread and therefore is thread safe. Perhaps this design assumption should be mentioned here.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
```diff
@@ -228,17 +222,16 @@ private synchronized void tryStop() {
     @Override
     public void execute() {
         try {
+            // If we try to start the task at all (by invoking initialize and possibly start), we count this as
+            // "started" in order to properly clean up any resources allocated by those invocations when the task is
+            // shut down by calling stop. If the task throws an exception during startup, it should still be able to
+            // clean up any allocated resources when stop is called, and if it isn't able to or even throws another
+            // exception during stop, the worst thing that happens is another exception gets logged for an already-
+            // failed task
```
rhauch: Minor suggestions to improve wording and to be a bit more explicit about expectations for Task implementations:
```suggestion
            // If we try to start the task at all by invoking initialize, then count this as
            // "started" and expect a subsequent call to the task's stop() method
            // to properly clean up any resources allocated by its initialize() or 
            // start() methods. If the task throws an exception during stop(),
            // the worst thing that happens is another exception gets logged for an already-
            // failed task
```

connect/runtime/src/test/java/org/apache/kafka/connect/integration/BlockingConnectorTest.java
```diff
@@ -263,145 +332,153 @@ private void verifyNormalConnector() throws InterruptedException {
...
+            this(new AbstractConfig(config(), props).getString(BLOCK_CONFIG));
+        }
+
+        public Block(String block) {
+            this.block = block;
+            synchronized (Block.class) {
+                if (blockLatch != null) {
+                    blockLatch.countDown();
+                }
+                blockLatch = new CountDownLatch(1);
```
rhauch, line 389: How does this static field initialization work within a blocking connector that also has a blocking task instance? Aren't both constructors called, resulting in the task constructor re-initializing the static `blockLatch` instance? Does it work because the tests don't block both a connector method and a task method? If so, doesn't that make it somewhat brittle if people want to add more tests but don't infer that limitation?

Would it be better to store a static list of all Block latches, have the constructor simply create a block latch and add it to that static list, and then have the `resetBlockLatch()` method remove each of the Block latches and call `countDown()` on them? (This still doesn't bode well for running the tests in parallel, but ATM that's not an issue.)

rhauch (commented):
Thanks, @C0urante, for the nice improvement and cleanup of the stop behavior. I do have a few minor suggestions to make two comments a bit more clear, and one question regarding the `BlockingConnectorTest` class and whether the cleanup of the static methods is actually complete.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
C0urante: This is a fair point. TBH I don't think this method deserves to exist anymore at all and can just be folded into `close` where a bunch of similar blocks already exist. There's also no need to track `stopped` since `close` is only ever invoked once. I'll refactor and remove the comment entirely since everything should be more readable after that change even without commenting.

connect/runtime/src/test/java/org/apache/kafka/connect/integration/BlockingConnectorTest.java
C0urante: > Does it work because the tests don't block both a connector method and a task method?

Yep, that's exactly it. Actually, it's even broader--it works because no test involves blocking in more than one method.

> If so, doesn't that make it somewhat brittle if people want to add more tests but don't infer that limitation?

It does make it brittle for that case, although since connectors and tasks are completely decoupled w/r/t allocation in a Connect cluster (they frequently exist on separate workers, for example), I'm a little skeptical of the need for a test that touches on a connector that blocks in its `Connector` and `Task` classes, either now or in the future.

> Would it be better to store a static list of all Block latches, have the constructor simply create a block latch and add it to that static list, and then have the resetBlockLatch() method remove each of the Block latches and call countDown() on them?

I'm happy to leave a comment on the limitations of the current test setup and how it could be expanded to cover more ground in the future, but it seems a bit premature to put in the necessary infrastructure for that when there aren't any identified cases yet that would require it, especially since the PR already greatly expands on the capabilities for testing blocks in connectors and tasks but only uses a portion of it. Is that fair?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
rhauch: Yeah, it's unclear whether to keep the `tryStop()` method to minimize deltas, or remove it to simplify things. +1 for simplifying things by moving the logic into `close()`.

connect/runtime/src/test/java/org/apache/kafka/connect/integration/BlockingConnectorTest.java
rhauch: Yeah, maybe the simplest fix is to add a comment on the `resetBlockLatch()` method mentioning the limitation.

C0urante: Ack, added a comment. LMKWYT

rhauch (approved):
LGTM. Thanks, @C0urante.

The build failures all appear to be flaky tests outside of Connect.

PR #7312, opened 2019-09-07
KAFKA-8421: Still return data during rebalance

guozhangwang:
1. Not wait until `updateAssignmentMetadataIfNeeded` returns true, but only call it once with 0 timeout. Also do not return empty if in rebalance.

2. Trim the pre-fetched records after long polling since assignment may have been changed.

3. Also need to update SubscriptionState to retain the state in `assignFromSubscribed` if it already exists (similar to `assignFromUser`), so that we do not need the transition of INITIALIZING to FETCHING.

4. Unit test: this actually took me the most time :)

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9671, opened 2020-12-02
KAFKA-10793: move handling of FindCoordinatorFuture to fix race condition

ableegoldman:
A race condition between the consumer and hb thread can lead to a failed but non-null `findCoordinatorFuture`, causing the AbstractCoordinator to wait endlessly on the request which it thinks is still in flight. We should move the handling of this future out of the listener callbacks and into the `ensureCoordinatorReady()` method where we can check the exception and clear the future all in one place.  

See ticket for full analysis.

Also starts logging a warning if the consumer is unable to connect to the coordinator for longer than the max poll interval.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
@@ -860,7 +860,7 @@ public abstract class AbstractCoordinator implements Closeable {
 
         @Override
         public void onFailure(RuntimeException e, RequestFuture<Void> future) {
-            log.debug("FindCoordinator request failed due to {}", e.getMessage());
+            log.debug("FindCoordinator request failed due to {}", e);
 
             if (!(e instanceof RetriableException)) {
                 // Remember the exception if fatal so we can ensure it gets thrown by the main thread
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
@@ -1369,10 +1369,8 @@ public abstract class AbstractCoordinator implements Closeable {
 
                         if (coordinatorUnknown()) {
                             if (findCoordinatorFuture != null) {
-                                // if it has failed, clear it so that hb thread can try discover again in the next loop in case main thread is busy
-                                if (findCoordinatorFuture.failed()) {
-                                    clearFindCoordinatorFuture();
-                                }
+                                // clear it so that hb thread can try discover again in the next loop in case main thread cannot
+                                clearFindCoordinatorFuture();
 
                                 // backoff properly
                                 AbstractCoordinator.this.wait(rebalanceConfig.retryBackoffMs);
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
@@ -1368,16 +1368,16 @@ public abstract class AbstractCoordinator implements Closeable {
                         long now = time.milliseconds();
 
                         if (coordinatorUnknown()) {
-                            if (findCoordinatorFuture != null || lookupCoordinator().failed()) {
-                                if (findCoordinatorFuture != null && findCoordinatorFuture.failed()) {
-                                    // if the future did complete and has failed, clear it so that the hb thread
-                                    // can send a new FindCoordinator request in case the main thread is busy
+                            if (findCoordinatorFuture != null) {
+                                // if it has failed, clear it so that hb thread can try discover again in the next loop in case main thread is busy
+                                if (findCoordinatorFuture.failed()) {
                                     clearFindCoordinatorFuture();
                                 }
 
-                                // the immediate future check ensures that we backoff properly in the case that no
-                                // brokers are available to connect to.
+                                // backoff properly
                                 AbstractCoordinator.this.wait(rebalanceConfig.retryBackoffMs);
+                            } else {
+                                lookupCoordinator();
                             }
                         } else if (heartbeat.sessionTimeoutExpired(now)) {
                             // the session timeout has expired without seeing a successful heartbeat, so we should
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
@@ -1368,13 +1368,17 @@ public abstract class AbstractCoordinator implements Closeable {
                         long now = time.milliseconds();
 
                         if (coordinatorUnknown()) {
-                            if (findCoordinatorFuture != null || lookupCoordinator().failed())
+                            if (findCoordinatorFuture != null || lookupCoordinator().failed()) {
                                 if (findCoordinatorFuture != null && findCoordinatorFuture.failed()) {
+                                    // if the future did complete and has failed, clear it so that the hb thread
+                                    // can send a new FindCoordinator request in case the main thread is busy
                                     clearFindCoordinatorFuture();
                                 }
+
                                 // the immediate future check ensures that we backoff properly in the case that no
                                 // brokers are available to connect to.
                                 AbstractCoordinator.this.wait(rebalanceConfig.retryBackoffMs);
+                            }
                         } else if (heartbeat.sessionTimeoutExpired(now)) {
                             // the session timeout has expired without seeing a successful heartbeat, so we should
                             // probably make sure the coordinator is still healthy.
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
@@ -137,10 +137,10 @@ public abstract class AbstractCoordinator implements Closeable {
     private HeartbeatThread heartbeatThread = null;
     private RequestFuture<ByteBuffer> joinFuture = null;
     private RequestFuture<Void> findCoordinatorFuture = null;
-    volatile private RuntimeException findCoordinatorException = null;
     private Generation generation = Generation.NO_GENERATION;
     private long lastRebalanceStartMs = -1L;
     private long lastRebalanceEndMs = -1L;
+    private long lastTimeOfConnectionMs = -1L; // starting logging a warning only after unable to connect for a while
 
     protected MemberState state = MemberState.UNJOINED;
 
@@ -235,11 +235,6 @@ public abstract class AbstractCoordinator implements Closeable {
             return true;
 
         do {
-            if (findCoordinatorException != null && !(findCoordinatorException instanceof RetriableException)) {
-                final RuntimeException fatalException = findCoordinatorException;
-                findCoordinatorException = null;
-                throw fatalException;
-            }
             final RequestFuture<Void> future = lookupCoordinator();
             client.poll(future, timer);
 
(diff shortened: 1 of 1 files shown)


ableegoldman:
Waiting to add tests until I get some sanity checks on this proposal

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -853,7 +844,7 @@ public void onSuccess(ClientResponse resp, RequestFuture<Void> future) {
 
         @Override
         public void onFailure(RuntimeException e, RequestFuture<Void> future) {
-            clearFindCoordinatorFuture();
+            log.debug("FindCoordinator request failed", e);
```
guozhangwang: nit: I think it's better to just print the e.message in a single line.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -235,11 +235,6 @@ protected synchronized boolean ensureCoordinatorReady(final Timer timer) {
             return true;
 
         do {
-            if (findCoordinatorException != null && !(findCoordinatorException instanceof RetriableException)) {
```
guozhangwang, line 238: The main reason for https://github.com/apache/kafka/pull/7312/files#diff-15efe9b844f78b686393b6c2e2ad61306c3473225742caed05c7edab9a138832R230-R234 is the following:

* inside ensureCoordinatorReady called by the main thread, we may break out of the loop at line 248 below, without knowing what's the final state of the future.
* and that future could be completed by the other thread (hb) later, and replaced by a new `future` object.

In that case, when the main thread calls ensureCoordinatorReady again, it will "miss" the previous future's contained fatal error.

So thinking about it again, I think we would still want to maintain the exception but only if it is a fatal one inside the handler (i.e. we do not probably need to register another listener just to bookkeep that exception, but just piggy-back this logic inside the handler listener directly), and then inside the while loop, we check if a previous future already gets a fatal exception and if yes, throw it to fail the whole client.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -248,18 +243,26 @@ protected synchronized boolean ensureCoordinatorReady(final Timer timer) {
...
+            RuntimeException fatalException = null;
+
             if (future.failed()) {
                 if (future.isRetriable()) {
                     log.debug("Coordinator discovery failed, refreshing metadata", future.exception());
                     client.awaitMetadataUpdate(timer);
-                } else
-                    throw future.exception();
+                } else {
+                    log.info("FindCoordinator request hit fatal  exception", fatalException);
```
guozhangwang: nit: extra space.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -248,18 +243,26 @@ protected synchronized boolean ensureCoordinatorReady(final Timer timer) {
...
+                    fatalException = future.exception();
+                }
             } else if (coordinator != null && client.isUnavailable(coordinator)) {
                 // we found the coordinator, but the connection has failed, so mark
                 // it dead and backoff before retrying discovery
                 markCoordinatorUnknown();
                 timer.sleep(rebalanceConfig.retryBackoffMs);
             }
+
+            clearFindCoordinatorFuture();
```
guozhangwang, line 269: I think this is a better approach, but we need to be careful about the callee inside hb thread:

```
if (findCoordinatorFuture != null || lookupCoordinator().failed())
```

i.e. a hb thread sending a discover-coordinator request would also cause a future to be assigned, but that future would only be cleared by the main thread caller. Thinking about that for a sec I think this is okay, but maybe worth having a second pair of eyes over it.

guozhangwang (commented):
Thanks for the PR. As we discussed so far this seems to be the most promising bug that caused the consumer never re-join issue. Let's see if it actually resolves it in the testbed.

One minor comment about bookkeeping the last fatal exception, lmk what do you think.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
ableegoldman: You mean like 
```
log.debug("FindCoordinator request failed due to {}", e.getMessage());
```
?

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
ableegoldman: Ah, good point...actually I think that's probably not ok for it to only ever be cleared in the main thread, since eg the main thread might be stuck in long processing while the hb threads should not be blocked from looking up the coordinator.
So, maybe we should also call `clearFindCoordinatorFuture` inside the hb thread in the `if (findCoordinatorFuture != null || lookupCoordinator().failed())` block (if it did indeed finish and has failed) -- WDYT?

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
ableegoldman: Ok yeah that makes sense. Thanks

ableegoldman:
Kicked off 30 versions of the system test which has seemed to be flaky due to this bug:
https://jenkins.confluent.io/job/system-test-kafka-branch-builder/4302/

ijuma:
Is this a 2.7.0 blocker? cc @bbejeck

ableegoldman:
@ijuma no, I don't think it should be a 2.7 blocker. It's definitely not a regression, AFAICT this has been around since the beginning. And a restart of the client will get it out of the bad state

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
guozhangwang: if you just have

```
log.debug("FindCoordinator request failed due to {}", e)
```

Then e.toString would be called which would usually be `e.name(): e.getMessage()`.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
guozhangwang: That makes sense, we can `clearFindCoordinatorFuture` inside the hb thread as well.

ijuma:
> And a restart of the client will get it out of the bad state

This is a massive deal though, right?

ableegoldman:
@ijuma sorry, I missed your earlier response. It's definitely not a trivial bug, yes. The main reasons I didn't formally propose this as a blocker was that it's been around forever, and I'm not confident that the fix is low-risk. WDYT @bbejeck ?

ijuma:
Let's aim for 2.7.1. :) 2.7.0 is done at this point.

bbejeck:
> that it's been around forever, and I'm not confident that the fix is low-risk.

Sorry I didn't see this before.  

I'd agree that it is not a blocker for reasons outlined by @ableegoldman.   I'd like to see this get in, but it seems we'd like a little more time for this patch to soak.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -1342,10 +1351,17 @@ public void run() {
                         long now = time.milliseconds();
 
                         if (coordinatorUnknown()) {
-                            if (findCoordinatorFuture != null || lookupCoordinator().failed())
+                            if (findCoordinatorFuture != null || lookupCoordinator().failed()) {
```
guozhangwang: The nested condition is a bit awkward, how about this:

```
if (findCoordinatorFuture != null) {
    // if it has failed, clear it so that hb thread can try discover again in the next loop in case main thread is busy
    if (findCoordinatorFuture.failed()) {
        clearFindCoordinatorFuture();
    } 

    // backoff properly
    AbstractCoordinator.this.wait(rebalanceConfig.retryBackoffMs);
} else {
    lookupCoordinator();
}
```

guozhangwang (commented):
I made another pass and just have one minor comment around hb logic.

I've triggered the test again. If it passes @ableegoldman please feel free to merge after addressing it.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
ableegoldman: Sure yeah that's much better

ableegoldman:
Weird -- these changes seem to be causing the `SaslXConsumerTest` family of tests to hang. I'm not very (or at all) familiar with these tests so I haven't found anything yet but I'm actively looking into it

guozhangwang:
> Weird -- these changes seem to be causing the `SaslXConsumerTest` family of tests to hang. I'm not very (or at all) familiar with these tests so I haven't found anything yet but I'm actively looking into it

Hmm... I'm not familiar with SaslXConsumerTest either...

ableegoldman:
Ok I think I've gotten to the bottom of this hanging test, and pushed a fix. Tests seem to be passing reliably for me locally. Aiming to get this merged in the next day or so so let me know if you have any concerns around the latest @guozhangwang

ableegoldman:
Ugh, looks like the JDK 15 build still timed out. But I think this was probably environmental based on inspecting the output -- I also tracked down and verified that every run of the previously-hanging `#testCoordinatorFailover` test did complete (and pass) so that does seem to be fixed. Will retrigger the build to be safe

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -1359,10 +1368,15 @@ public void run() {
                         long now = time.milliseconds();
 
                         if (coordinatorUnknown()) {
-                            if (findCoordinatorFuture != null || lookupCoordinator().failed())
-                                // the immediate future check ensures that we backoff properly in the case that no
-                                // brokers are available to connect to.
+                            if (findCoordinatorFuture != null) {
```
guozhangwang, line 1371: I think the issue is spot-on! The logic here becomes a bit hard to understand for other readers now and I'd suggest update the cmment as:

"Clear the future so that after the backoff in the next iteration, if hb still sees coordinator unknown it will try re-discover the coordinator in case the main thread cannot"

Otherwise, LGTM.

ableegoldman: SG

ableegoldman:
A few tests failed, but no hanging this time:

```
StoreQueryIntegrationTest.shouldQuerySpecificStalePartitionStores -- known to be flaky
FetcherTest.testEarlierOffsetResetArrivesLate -- hit "TimeoutException: testEarlierOffsetResetArrivesLate() timed out after 10 seconds", I haven't seen this fail before, on this PR or on any other, so I believe it's unrelated. But I ran it 10 times locally to be sure and all passed
MirrorConnectorsIntegrationSSLTest.testReplication -- in Connect, seems to be unrelated
StoreQueryIntegrationTest.shouldQuerySpecificStalePartitionStores -- known flaky, looks environmental (slow startup)
```

ableegoldman:

guozhangwang:
Awesome!!

ableegoldman:
Cherrypicked to 2.6 as well

rqode:
When you experience this issue on 2.6.0 consumers is it enough to only upgrade the kafka client to 2.6.2 or does this fix require a server upgrade? Thanks

ableegoldman:
It should be sufficient to upgrade just the consumers, this is a client-side fix only

chenhongluo:
We had the same problem with the 0.10.2 client. The findCoordinatorFuture was not being cleared, which resulted in a ”coordinator not available“ error for every commitOffsetAsync call. We use consumer.assign() to consume. What I can't understand is that both lookupCoordinator() and clearFindCoordinatorFuture() use synchronized which means clearFindCoordinatorFuture() only can execute after lookupCoordinator() are completed. Now I want to reproduce this bug, is there any good idea? @ableegoldman @guozhangwang 

There are code:

```
protected synchronized RequestFuture<Void> lookupCoordinator() {
        if (findCoordinatorFuture == null) {
            // find a node to ask about the coordinator
            Node node = this.client.leastLoadedNode();
            if (node == null) {
                // TODO: If there are no brokers left, perhaps we should use the bootstrap set
                // from configuration?
                log.warn("No broker available to send GroupCoordinator request for group {}", groupId);
                return RequestFuture.noBrokersAvailable();
            } else {
                findCoordinatorFuture= sendGroupCoordinatorRequest(node);
            }
        }
        return findCoordinatorFuture;
    }
  private synchronized void clearFindCoordinatorFuture() {
        findCoordinatorFuture = null;
    }
```

PR #9675, opened 2020-12-02
KAFKA-10794 Replica leader election is too slow in the case of too many partitions

Montyleo:
There is more than 6000 topics and 300 brokers in my kafka cluster, and we frequently run kafka-preferred-replica-election.sh to rebalance our cluster. But the reblance process spendes too more time and cpu resource like the picture blow.

We find that the function:'controllerContext.allPartitions' is invoked too many times.
Thr jira link is https://issues.apache.org/jira/browse/KAFKA-10794

### DIFF
--- a/core/src/main/scala/kafka/controller/KafkaController.scala
+++ b/core/src/main/scala/kafka/controller/KafkaController.scala
@@ -2165,7 +2165,8 @@ class KafkaController(val config: KafkaConfig,
           case None => zkClient.getPreferredReplicaElection
         }
 
-        val (knownPartitions, unknownPartitions) = partitions.partition(tp => controllerContext.allPartitions.contains(tp))
+        val allPartitions = controllerContext.allPartitions
+        val (knownPartitions, unknownPartitions) = partitions.partition(tp => allPartitions.contains(tp))
         unknownPartitions.foreach { p =>
           info(s"Skipping replica leader election ($electionType) for partition $p by $electionTrigger since it doesn't exist.")
         }


chia7712 (approved):
@Montyleo nice finding. LGTM

chia7712 (approved):

Montyleo:
> @Montyleo nice finding. LGTM

Thanks

Montyleo:
@huxihx   Please help me review the code, thanks.

chia7712:
@Montyleo Is the failed test related to this PR?

Montyleo:
> @Montyleo Is the failed test related to this PR?
Hi, chia7712 
  Thanks for your reply. The  failed test is about SaslAuthenticator, not related to this PR, even not related the component:kafkacontroller.  It seems that the jdk 8 version is too low. I'll find the reason.

chia7712:
> I'll find the reason.

Is there a existent ticket? If not, could you file a jira to log it? Also, you can assign the ticket to yourself ( I have given the permission to you) if you have free cycle to trace it.

I will merge this PR tomorrow if no objection.

Montyleo:
> > I'll find the reason.
> 
> Is there a existent ticket? If not, could you file a jira to log it? Also, you can assign the ticket to yourself ( I have given the permission to you) if you have free cycle to trace it.
> 
> I will merge this PR tomorrow if no objection.

Ok，I have no objection. I have created a jira to log it, [https://issues.apache.org/jira/projects/KAFKA/issues/KAFKA-10797?filter=allissues]. I'll trace it in my local environment.

chia7712:
@Montyleo Thanks for your contribution!

lqjack:
@chia7712  does the patch can resolve the issue ? I find the only differences is that controllerContext.allPartitions can be invoked once or the number of partition times . please correct me if I am wrong. thanks.

chia7712:
@lqjack good question!

>  I find the only differences is that controllerContext.allPartitions can be invoked once or the number of partition times .

```controllerContext.allPartitions``` does not return a constant value. It create a new collection and the overhead could be high if there are a lot of partitions. This PR makes ```controllerContext.allPartitions``` be called only once to reduce the cost of getting "all partitions".

> does the patch can resolve the issue ?

@Montyleo It seems to me the optimization of this PR is good enough. However, it would be better to show the improvement on your env by this patch.

Montyleo:
> @chia7712 does the patch can resolve the issue ? I find the only differences is that controllerContext.allPartitions can be invoked once or the number of partition times . please correct me if I am wrong. thanks.

Hi，lqjack

Thanks for your question.
There is a saying that: quantitative change leads to qualitative change.
when  the function controllerContext.allPartitions was called too many time, the rebalance will become too slow.
I'll show you the effect after the PR published，1.3ms VS 35541ms

clipboard_image_1608279734070

image

PR #9676, opened 2020-12-02
KAFKA-10778: Fence appends after write failure

tombentley:


tombentley:
@hachikuji is this what you had in mind for what you described in [KAFKA-10778](https://issues.apache.org/jira/browse/KAFKA-10778)?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1219,6 +1219,9 @@ class Log(@volatile private var _dir: File,
               appendInfo.logAppendTime = duplicate.timestamp
               appendInfo.logStartOffset = logStartOffset
             case None =>
+              if (logDirFailureChannel.logDirIsFailed(parentDir)) {
```
hachikuji: Perhaps we could move this to somewhere near the top? I don't think we get much benefit by delaying the check since duplicates would be a rare case. We probably don't want to have to trust the producer state anyway after an append failure.

core/src/main/scala/kafka/server/LogDirFailureChannel.scala
```diff
@@ -49,6 +49,13 @@ class LogDirFailureChannel(logDirNum: Int) extends Logging {
       offlineLogDirQueue.add(logDir)
   }
 
+  /*
+   * Return whether the given log dir is offline.
+   */
+  def logDirIsFailed(logDir: String): Boolean = {
```
hachikuji: Maybe `isOffline` or `hasFailed`?

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -2818,6 +2818,21 @@ class LogTest {
...
 
+  @Test
+  def testAppendToLogInFailedLogDir(): Unit = {
+    val log = createLog(logDir, LogConfig())
+    log.appendAsLeader(TestUtils.singletonRecords(value = null), leaderEpoch = 0)
+    assertEquals(0, readLog(log, 0, 4096).records.records.iterator.next().offset)
+    log.logDirFailureChannel.maybeAddOfflineLogDir(logDir.getParent, "Simulating failed log dir", new IOException("Test failure"))
+    try {
+      log.appendAsLeader(TestUtils.singletonRecords(value = null), leaderEpoch = 0)
+      fail()
```
hachikuji: nit: we can use `assertThrows` or `intercept`

hachikuji (commented):
@tombentley Thanks for the patch. I was thinking of adding a flag, but I think this works too. Note that there are some test failures.

tombentley:
@hachikuji thanks for taking a look, I've fixed the test failures. Do we also need something on the read path, KAFKA-10778 suggested so?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1219,6 +1219,9 @@ class Log(@volatile private var _dir: File,
               appendInfo.logAppendTime = duplicate.timestamp
               appendInfo.logStartOffset = logStartOffset
             case None =>
+              if (logDirFailureChannel.logDirIsOffline(parentDir)) {
+                throw new KafkaStorageException(s"The log dir $parentDir has failed.");
```
guozhangwang: Maybe we can leave a more informative error message here? Sth like "... dir has failed due to a previous IO exception", just indicating it is not failed because of the current calling trace.

tombentley: @guozhangwang good point. I also changed the language to 'offline' rather than 'failed', since this is more consistent with other usage.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1106,6 +1106,13 @@ class Log(@volatile private var _dir: File,
         // they are valid, insert them in the log
         lock synchronized {
           checkIfMemoryMappedBufferClosed()
+
+          // check for offline log dir in case a retry following an IOException happens before the log dir
```
hachikuji: Hmm.. In case there is an IOException on an append, we will release the lock and fail the log dir in `maybeHandleIOException`. There is a window for another append to sneak by. It looks like it should be possible to pull `maybeHandleIOException` into the locked section here since `analyzeAndValidateRecords` does not do any IO.

core/src/main/scala/kafka/server/LogDirFailureChannel.scala
```diff
@@ -49,6 +49,13 @@ class LogDirFailureChannel(logDirNum: Int) extends Logging {
       offlineLogDirQueue.add(logDir)
   }
 
+  /*
+   * Return whether the given log dir is offline.
+   */
+  def logDirIsOffline(logDir: String): Boolean = {
+    offlineLogDirs.containsKey(logDir)
```
hachikuji: I am not sure if it is really necessary, but since offline dirs are a rare situation, I'm wondering if makes sense to optimize for the common case to avoid the lookup. For example, maybe we could leave `offlineLogDirs` uninitialized until the first log dir failure.

hachikuji (commented):
Thanks, left a couple comments. I think it is useful if we add the fencing on the read path as well to answer your question. Let me know if you can think of a reason not to.

tombentley:
@hachikuji @guozhangwang I decided to switch to a flag rather than implementing the uninitialised `offlineLogDirs` with an `AtomicReference` which would be contented by every thread operating on logs in the same parent directory. Please could you take another look?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1556,6 +1572,8 @@ class Log(@volatile private var _dir: File,
           done = fetchDataInfo != null || segmentEntry == null
         }
 
+        checkForLogDirFailure()
```
hachikuji: Seems more intuitive to move this check before the segment read. I don't think we can totally avoid race conditions with a failure in append since we don't have the lock here. Perhaps we could even move this check to `maybeHandleIOException` so that we handle all cases?

core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -2818,6 +2818,22 @@ class LogTest {
       new SimpleRecord(RecordBatch.NO_TIMESTAMP, "key".getBytes, "value".getBytes)), leaderEpoch = 0)
   }
 
+  @Test
+  def testAppendToOrReadFromLogInFailedLogDir(): Unit = {
+    val log = createLog(logDir, LogConfig())
+    log.appendAsLeader(TestUtils.singletonRecords(value = null), leaderEpoch = 0)
+    assertEquals(0, readLog(log, 0, 4096).records.records.iterator.next().offset)
+    try {
+      log.maybeHandleIOException("Simulating failed log dir") {
```
hachikuji: Another way to trigger an IO exception is to rename the log file. This trick is used in `testAppendToTransactionIndexFailure`. Then we don't need to expose `maybeHandleIOException` for testing.

hachikuji (commented):
Looks good. Left a couple small comments.

core/src/main/scala/kafka/log/Log.scala
tombentley: Yeah, I wondered about putting it in `maybeHandleIOException` too, and I've made this change. There are a few places where changing the order in which we do `maybeHandleIOException` with respect to `lock synchronized` would seem to avoid some possible races with no side effects other than increasing the duration the lock is held (for the `try` and the volatile read). I'm thinking specifically `maybeIncrementLogStartOffset`, `roll`, `delete`, `truncateFullyAndStartAt` and perhaps even `truncateTo`. But given that there will always exist some races maybe it's not worth it, WDYT?

core/src/test/scala/unit/kafka/log/LogTest.scala
tombentley: That only seems to work when writing to the transaction index (because the file opening is deferred so that it's only at the point where we try to append some transaction markers that the index file open is attempted and the path found to be a directory, which means that the exception propagates through `Log`). Renaming the segment file, on the other hand, doesn't propagate through `Log`, because the rename throws immediately and propagates directly from `FileRecords` to the test. Obviously when the exception doesn't propagate via `Log` the `logDirOffline` doesn't get set.

Making this change kind-of mixes transactional appends into a test for something with is orthogonal to transactional logs. In any case, I've done as you suggest, I just thought it worth mentioning.

tombentley:
Thanks for the review @hachikuji.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1093,19 +1097,25 @@ class Log(@volatile private var _dir: File,
...
+      var validRecords = trimInvalidBytes(records, appendInfo)
+
+      // they are valid, insert them in the log
+      lock synchronized {
+        maybeHandleIOException(s"Error while appending records to $topicPartition in dir ${dir.getParent}") {
           checkIfMemoryMappedBufferClosed()
+
+          // check for offline log dir in case a retry following an IOException happens before the log dir
+          // is taken offline, which would result in inconsistent producer state
+          checkForLogDirFailure()
```
hachikuji: Do we still need this since this check is in `maybeHandleIOException`?

hachikuji (commented):
LGTM overall. Just one more question.

tombentley:
Thanks for the review @hachikuji, now fixed.

hachikuji (approved):
LGTM

PR #9100, opened 2020-07-29
Add AlterISR RPC and use it for ISR modifications

mumrah:
# Background

On the broker side, ISR changes in the following cases:
* Follow catches up to the leader, we expand the ISR
* On a timer, we check for followers which have fallen out of sync, we shrink the ISR

Previously, we would synchronously update the partition state in ZK which includes the current leader+epoch as well as the ISR. The ZK version is used for concurrency control and is known to the broker (kept in LeaderAndIsr struct). In order to notify the controller of the ISR changes, the leader would periodically write to a special "isr_change_notification" znode which the controlled kept a ZK watch on. A list of modified partitions was written to this znode so the controller would only reload the necessary ISRs.

With [KIP-497](https://cwiki.apache.org/confluence/display/KAFKA/KIP-497%3A+Add+inter-broker+API+to+alter+ISR), we introduce an asynchronous ISR workflow that makes the controller the only component which can change the ISR state.

# TLDR

* ISR updates are now sent asynchronously to the controller
* ISRs are now only persisted by the controller 
* Leader can immediately assume a larger ISR during ISR expansion
* No more ZK watch mechanism for ISR propagation
* Gated behind IBP 2.7-IV2

# Broker changes

Now when the broker wants to modify the ISR, it will send an AlterIsr request to the controller. Since this request can include any number of partition's ISR modifications, we will want to allow batching of ISR updates. This is done by adding a small delay (50ms) before sending the request to the controller after an ISR change is requested. By adding this delay, other ISR changes can accumulate before getting sent off as a single AlterIsr request. 

Once the controller makes the update, it will return the updated ISR to the leader and send UpdateMetadata requests to the other interested brokers. As specified in the KIP, while a leader is waiting on the controller to respond it will assume the largest ISR (i.e., the union of the actual ISR and the pending ISR). For ISR expansions, this allows the leader to assume the new ISR right away. For ISR shrinks however, the leader must wait for confirmation from the controller that the ISR has been persisted (see the KIP for more details). We call this oversubscribed ISR the "maximal" ISR

A new `AlterIsrManager` class is added which accumulates ISR changes and uses a background `BrokerToControllerChannelManager` thread to send AlterIsr requests to the controller.

The use of AlterIsr is only enabled when a broker is configured with IBP version 2.7-IV2 or greater.

# Controller changes

[...]

PR #9677, opened 2020-12-02
KAFKA-10799 AlterIsr utilizes ReplicaManager ISR metrics

mumrah:
In #9100, we missed the inclusion of the high level ISR shrink/expand metrics that are managed by ReplicaManager. This PR adds a small abstraction that allows us to mark these metrics without bringing ReplicaManager in as a dependency of Partition.

### DIFF
--- a/core/src/main/scala/kafka/cluster/Partition.scala
+++ b/core/src/main/scala/kafka/cluster/Partition.scala
@@ -1429,15 +1429,15 @@ class Partition(val topicPartition: TopicPartition,
         case Left(error: Errors) =>
           isrChangeListener.markFailed()
           error match {
-          case Errors.UNKNOWN_TOPIC_OR_PARTITION =>
-            debug(s"Controller failed to update ISR to $proposedIsrState since it doesn't know about this topic or partition. Giving up.")
-          case Errors.FENCED_LEADER_EPOCH =>
-            debug(s"Controller failed to update ISR to $proposedIsrState since we sent an old leader epoch. Giving up.")
-          case Errors.INVALID_UPDATE_VERSION =>
-            debug(s"Controller failed to update ISR to $proposedIsrState due to invalid zk version. Giving up.")
-          case _ =>
-            warn(s"Controller failed to update ISR to $proposedIsrState due to unexpected $error. Retrying.")
-            sendAlterIsrRequest(proposedIsrState)
+            case Errors.UNKNOWN_TOPIC_OR_PARTITION =>
+              debug(s"Controller failed to update ISR to $proposedIsrState since it doesn't know about this topic or partition. Giving up.")
+            case Errors.FENCED_LEADER_EPOCH =>
+              debug(s"Controller failed to update ISR to $proposedIsrState since we sent an old leader epoch. Giving up.")
+            case Errors.INVALID_UPDATE_VERSION =>
+              debug(s"Controller failed to update ISR to $proposedIsrState due to invalid zk version. Giving up.")
+            case _ =>
+              warn(s"Controller failed to update ISR to $proposedIsrState due to unexpected $error. Retrying.")
+              sendAlterIsrRequest(proposedIsrState)
         }
         case Right(leaderAndIsr: LeaderAndIsr) =>
           // Success from controller, still need to check a few things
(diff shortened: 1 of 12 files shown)


core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -53,7 +59,7 @@ trait PartitionStateStore {
 
 class ZkPartitionStateStore(topicPartition: TopicPartition,
                             zkClient: KafkaZkClient,
-                            replicaManager: ReplicaManager) extends PartitionStateStore {
+                            isrChangeListener: IsrChangeListener) extends PartitionStateStore {
```
hachikuji: Maybe we can move this out of `ZkPartitionStateStore` since we need the dependence on `IsrChangeListener` in `Partition` anyway.

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -45,6 +45,12 @@ import org.apache.kafka.common.{IsolationLevel, TopicPartition}
 import scala.collection.{Map, Seq}
 import scala.jdk.CollectionConverters._
 
+trait IsrChangeListener {
```
hachikuji, line 48: Maybe `IsrChangeMetrics`?

hachikuji (commented):
Looks reasonable. Would it be possible to alter one of the existing test cases to verify that the meters are updated?

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -107,10 +106,24 @@ object Partition extends KafkaMetricsGroup {
   def apply(topicPartition: TopicPartition,
             time: Time,
             replicaManager: ReplicaManager): Partition = {
+
+    val isrChangeMetrics = new IsrChangeMetrics {
+      override def markExpand(): Unit = {
+        replicaManager.recordIsrChange(topicPartition)
```
hachikuji, line 112: Hmm.. I had forgotten about this. The purpose of this call is so that we can propagate the ISR change to the controller, but we shouldn't need to do this down the AlterIsr path. This also makes your initial name `IsrChangeListener` a bit more accurate 😅 .

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -1406,10 +1429,13 @@ class Partition(val topicPartition: TopicPartition,
...
+            isrChangeMetrics.markFailed()
           case Errors.FENCED_LEADER_EPOCH =>
             debug(s"Controller failed to update ISR to $proposedIsrState since we sent an old leader epoch. Giving up.")
+            isrChangeMetrics.markFailed()
           case Errors.INVALID_UPDATE_VERSION =>
             debug(s"Controller failed to update ISR to $proposedIsrState due to invalid zk version. Giving up.")
+            isrChangeMetrics.markFailed()
           case _ =>
             warn(s"Controller failed to update ISR to $proposedIsrState due to unexpected $error. Retrying.")
             sendAlterIsrRequest(proposedIsrState)
```
mumrah: Note that I'm not including the `markFailed` call here. Think we should include it here? Or possibly introduce a new retry metric?

hachikuji: Hmm, I think we should probably do it. If we hit some unexpected case which the broker can't get out of, we'd want the metric to show it.

hachikuji (commented):
LGTM except for the one case you called it.

core/src/main/scala/kafka/cluster/Partition.scala
mumrah: Sounds good

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -1403,7 +1426,9 @@ class Partition(val topicPartition: TopicPartition,
       }
 
       result match {
-        case Left(error: Errors) => error match {
+        case Left(error: Errors) =>
+          isrChangeListener.markFailed()
+          error match {
```
hachikuji, line 1431: nit: now the block below is misaligned

hachikuji (approved):
LGTM. Left one nit.

mumrah:
Failed tests are known flaky

https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-9677/8/
image

PR #9684, opened 2020-12-04
KAFKA-10764: Add support for returning topic IDs on create, supplying topic IDs for delete

jolshan:
Updated CreateTopicResponse, DeleteTopicsRequest/Response and added some new AdminClient methods and classes. Now the newly created topic ID will be returned in CreateTopicsResult and found in TopicAndMetadataConfig, and topics can be deleted by supplying topic IDs through deleteTopicsWithIds which will return DeleteTopicsWithIdsResult. 


core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1896,6 +1896,11 @@ class KafkaApis(val requestChannel: RequestChannel,
               .setTopicConfigErrorCode(Errors.NONE.code)
           }
         }
+        val topicIds = zkClient.getTopicIdsForTopics(results.asScala.map(result => result.name()).toSet)
```
jolshan: If going to ZK here is too slow, another option is to provide a callback to adminManager.createTopics which can be called after createTopicWithAssignment. This callback would add the topic ID to the result. The idea is that createTopicWithAssignment (and writeTopicPartitionAssignment) would return the topic ID to avoid an extra call to ZK. I wasn't sure which option was better.

dengziming: I think we can remove support for zk when using topic-id since KIP-500 will remove zk, also ping @rajinisivaram to have a look.

jolshan: Currently the KIP-516 work relies on ZK for storing the topic IDs. I think we will need to make some more changes when KIP-500 fully removes ZK

clients/src/main/java/org/apache/kafka/clients/admin/CreateTopicsResult.java
```diff
@@ -68,6 +69,19 @@ protected CreateTopicsResult(Map<String, KafkaFuture<TopicMetadataAndConfig>> fu
...
+     * If broker returned an error for topic configs, throw appropriate exception. For example,
+     * {@link org.apache.kafka.common.errors.TopicAuthorizationException} is thrown if user does not
+     * have permission to describe topic configs.
+     */
+    public KafkaFuture<Uuid> topicId(String topic) {
+        return futures.get(topic).thenApply(TopicMetadataAndConfig::topicId);
+    }
+    
     /**
      * Returns a future that provides number of partitions in the topic when the request completes.
```
jolshan, line 86: I noticed all these comments say "If broker version doesn't support **replication factor** in the response..." should each actually say whatever the method is returning (numPartitions, topicId, etc.)

core/src/main/scala/kafka/zk/KafkaZkClient.scala
```diff
@@ -621,10 +621,10 @@ class KafkaZkClient private[zk] (zooKeeperClient: ZooKeeperClient, isSecure: Boo
         case Code.NONODE => None
         case _ => throw getDataResponse.resultException.get
       }
-    }.map(_.get)
-      .map(topicIdAssignment => (topicIdAssignment.topic,
-        topicIdAssignment.topicId.getOrElse(
-          throw new IllegalStateException("Topic " + topicIdAssignment.topic + " does not have a topic ID."))))
+    }.filter(_.isDefined)
```
dengziming: nit: getDataResponses.map(xxx).filter(_.isDefined).map(_.get) can be replaced by getDataResponses.flatMap(xxx)

clients/src/main/resources/common/message/DeleteTopicsRequest.json
```diff
@@ -23,10 +23,17 @@
...
-  "validVersions": "0-5",
+  //
+  // Version 6 reorganizes topics, adds topic IDs and allows topic names to be null.
+  "validVersions": "0-6",
   "flexibleVersions": "4+",
   "fields": [
-    { "name": "TopicNames", "type": "[]string", "versions": "0+", "entityType": "topicName",
+    { "name": "Topics", "type": "[]DeleteTopicState", "versions": "6+", "about": "The name or topic ID of the topic",
+      "fields": [
+        {"name": "Name", "type": "string", "versions": "6+", "nullableVersions": "6+", "about": "The topic name"},
```
dengziming, line 33: Why is topic name nullable but topicId not? do you mean we will fill all delete requests with a topicId before sending a request?

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1623,6 +1625,32 @@ public DeleteTopicsResult deleteTopics(final Collection<String> topicNames,
...
 
+    @Override
+    public DeleteTopicsWithIdsResult deleteTopicsWithIds(final Collection<Uuid> topicIds,
+                                           final DeleteTopicsOptions options) {
+        final Map<Uuid, KafkaFutureImpl<Void>> topicFutures = new HashMap<>(topicIds.size());
+        final List<Uuid> validTopicIds = new ArrayList<>(topicIds.size());
+        for (Uuid topicId : topicIds) {
+            if (topicId.equals(Uuid.ZERO_UUID)) {
+                KafkaFutureImpl<Void> future = new KafkaFutureImpl<>();
+                future.completeExceptionally(new InvalidTopicException("The given topic ID '" +
```
dengziming: Should this be UnknownTopicIdException?

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1694,6 +1722,79 @@ void handleFailure(Throwable throwable) {
...
+                // Check for controller change
+                handleNotControllerError(abstractResponse);
+                // Handle server responses for particular topics.
+                final DeleteTopicsResponse response = (DeleteTopicsResponse) abstractResponse;
+                final List<Uuid> retryTopics = new ArrayList<>();
+                final Map<Uuid, ThrottlingQuotaExceededException> retryTopicQuotaExceededExceptions = new HashMap<>();
+                for (DeletableTopicResult result : response.data().responses()) {
+                    KafkaFutureImpl<Void> future = futures.get(result.topicId());
+                    if (future == null) {
+                        log.warn("Server response mentioned unknown topic {}", result.name());
```
dengziming: here we'd better use result.topicId() since we are using topicId.

clients/src/main/java/org/apache/kafka/common/requests/DeleteTopicsRequest.java
```diff
@@ -39,8 +45,25 @@ public Builder(DeleteTopicsRequestData data) {
 
         @Override
         public DeleteTopicsRequest build(short version) {
+            if (version >= 6 && data.topicNames().size() != 0) {
+                data.setTopics(groupByTopic(data.topicNames()));
+            } else if (version >= 6) {
+                for (DeleteTopicState topic : data.topics()) {
+                    if (topic.name().equals("")) {
```
dengziming: if topic.name() == null, a NEP will throw here.

core/src/main/scala/kafka/server/KafkaApis.scala
dengziming: Here could we use controllerContext since createTopicRep is handled at the active controller.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1981,29 +1986,39 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-          .setName(topic)
+          .setName(topic.name())
+          .setTopicId(topic.topicId())
           .setErrorCode(error.code))
       }
       sendResponseCallback(results)
     } else {
-      deleteTopicRequest.data.topicNames.forEach { topic =>
+      deleteTopicRequest.topics().forEach { topic =>
+        val name = if (topic.name() != null) topic.name()
```
dengziming: According to the kip, we will use topicId firstly and topic name will be used only if topicId is ZERO, but you just use the topic name even topicId is not ZERO here?

core/src/main/scala/kafka/zk/KafkaZkClient.scala
jolshan: This will change soon with the new PR I'll be opening tomorrow that gates topic IDs behind IBP 2.8.

clients/src/main/resources/common/message/DeleteTopicsRequest.json
jolshan: We will use the zero UUID to represent null as described in the KIP.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
jolshan: That makes sense.

clients/src/main/java/org/apache/kafka/common/requests/DeleteTopicsRequest.java
jolshan: Good point. I think I was assuming the default is the empty string, but someone could set the data field to null.

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: I'm setting both topic ID and name here for the result since we will return with both, but if you look at the code below, it handles topic ID errors first. From then on, I'm not sure if it makes a difference. I suppose I could check if the topic ID is zero rather than if the name is null, but right now the code paths either set topic ID and name to null or set topic name and topic ID to zero UUID. So for now, those do equivalent things.

core/src/main/scala/kafka/zk/KafkaZkClient.scala
jolshan: Decided to just open it now. https://github.com/apache/kafka/pull/9814/files

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: I'll switch to checking for zero UUID for now.

jolshan:
I'm planning on adding one more test, but that test is blocked on https://github.com/apache/kafka/pull/9814

jolshan:
retest this please

core/src/test/scala/unit/kafka/server/DeleteTopicsRequestTest.scala
```diff
@@ -132,6 +133,25 @@ class DeleteTopicsRequestTest extends BaseRequestTest {
     validateTopicIsDeleted(timeoutTopic)
   }
 
+  @Test
+  def testErrorDeleteTopicRequestsOnOldVersions(): Unit = {
```
jolshan: This is incorrect. Should be removed until 2.8 PR is added.

clients/src/main/java/org/apache/kafka/clients/admin/CreateTopicsResult.java
rajinisivaram: In the create result, we wouldn't see UnsupportedVersionException for topicId() just because topic ids are not enabled, would we?

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1625,6 +1628,32 @@ public DeleteTopicsResult deleteTopics(final Collection<String> topicNames,
         return new DeleteTopicsResult(new HashMap<>(topicFutures));
     }
 
+    @Override
+    public DeleteTopicsWithIdsResult deleteTopicsWithIds(final Collection<Uuid> topicIds,
+                                           final DeleteTopicsOptions options) {
```
rajinisivaram: nit: indentation

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1696,6 +1725,79 @@ void handleFailure(Throwable throwable) {
             }
         };
     }
+   
+    private Call getDeleteTopicsWithIdsCall(final DeleteTopicsOptions options,
+                                     final Map<Uuid, KafkaFutureImpl<Void>> futures,
```
rajinisivaram: nit: indentation

clients/src/main/java/org/apache/kafka/common/requests/DeleteTopicsRequest.java
```diff
@@ -36,8 +42,25 @@ public Builder(DeleteTopicsRequestData data) {
 
         @Override
         public DeleteTopicsRequest build(short version) {
+            if (version >= 6 && data.topicNames().size() != 0) {
```
rajinisivaram: nit: isEmpty() instead of .size()?

clients/src/main/resources/common/message/DeleteTopicsResponse.json
```diff
@@ -27,15 +27,20 @@
   //
   // Version 5 adds ErrorMessage in the response and may return a THROTTLING_QUOTA_EXCEEDED error
   // in the response if the topics deletion is throttled (KIP-599).
-  "validVersions": "0-5",
+  //
+  // Version 6 adds topic ID to responses. An UNSUPPORTED_VERSION error code will be returned when attempting to
+  // delete using topic IDs when IBP > 2.8. UNKNOWN_TOPIC_ID error code will be returned when IBP is at least 2.8, but
```
rajinisivaram: IBP < 2.8 for UNSUPPORTED_VERSION?

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
```diff
@@ -919,6 +973,36 @@ public void testDeleteTopicsRetryThrottlingExceptionWhenEnabled() throws Excepti
...
+            Uuid topicId1 = Uuid.randomUuid();
+            Uuid topicId2 = Uuid.randomUuid();
+            Uuid topicId3 = Uuid.randomUuid();
+            
+            env.kafkaClient().prepareResponse(
+                    expectDeleteTopicsRequestWithTopicIds(topicId1, topicId2, topicId3),
+                    prepareDeleteTopicsResponse(1000,
+                            deletableTopicResultWithId(topicId1, Errors.NONE),
+                            deletableTopicResultWithId(topicId2, Errors.THROTTLING_QUOTA_EXCEEDED),
+                            deletableTopicResultWithId(topicId3, Errors.TOPIC_ALREADY_EXISTS)));
```
rajinisivaram: Doesn't really matter for this test, but perhaps we could use `UNKNOWN_TOPIC_ID` (multiple tests below)?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1843,6 +1843,8 @@ class KafkaApis(val requestChannel: RequestChannel,
               .setNumPartitions(-1)
               .setReplicationFactor(-1)
               .setTopicConfigErrorCode(Errors.NONE.code)
+          } else {
+            result.setTopicId(controller.controllerContext.topicIds.getOrElse(result.name(), Uuid.ZERO_UUID))
```
rajinisivaram: Can't we populate topic id in AdminManager similar to other metadata in the result?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1930,29 +1932,43 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-          .setName(topic))
+          .setName(name)
+          .setTopicId(topic.topicId()))
       }
       val authorizedTopics = authHelper.filterByAuthorized(request.context, DELETE, TOPIC,
         results.asScala)(_.name)
       results.forEach { topic =>
-         if (!authorizedTopics.contains(topic.name))
+         val foundTopicId = !topic.topicId().equals(Uuid.ZERO_UUID) && topic.name() != null
+         val topicIdSpecified = !topic.topicId().equals(Uuid.ZERO_UUID)
```
rajinisivaram: We can move this one line above and use `topicIdSpecified`  in the line above

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1930,29 +1932,43 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-          .setName(topic)
+          .setName(topic.name())
+          .setTopicId(topic.topicId())
           .setErrorCode(error.code))
       }
       sendResponseCallback(results)
     } else {
-      deleteTopicRequest.data.topicNames.forEach { topic =>
+      deleteTopicRequest.topics().forEach { topic =>
+        val name = if (topic.topicId().equals(Uuid.ZERO_UUID)) topic.name()
```
rajinisivaram: We can use `topic.topicId == Uuid.ZERO_UUID` in Scala instead of `equals`. (a couple of places below too)

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1930,29 +1932,43 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+          else controller.controllerContext.topicNames.getOrElse(topic.topicId(), null)
```
rajinisivaram: What happens if both topic id and topic name are provided in the request?

core/src/test/scala/unit/kafka/server/BaseRequestTest.scala
```diff
@@ -21,6 +21,7 @@ import java.io.{DataInputStream, DataOutputStream}
 import java.net.Socket
 import java.nio.ByteBuffer
 import java.util.Properties
+
```
rajinisivaram: We could revert changes to this file since there aren't any real changes?

core/src/test/scala/unit/kafka/server/MetadataRequestIBPTest/TopicIdWithOldInterBrokerProtocolTest.scala
```diff
@@ -59,8 +61,56 @@ class TopicIdWithOldInterBrokerProtocolTest extends BaseRequestTest {
...
+    val timeout = 10000
+    createTopic("topic-7", 3, 2)
+    createTopic("topic-6", 1, 2)
+    val ids = Map("topic-7" -> Uuid.randomUuid(), "topic-6" -> Uuid.randomUuid())
+    val request = new DeleteTopicsRequest.Builder(
+      new DeleteTopicsRequestData()
+        .setTopics(Arrays.asList(new DeleteTopicState().setTopicId(ids("topic-7")),
+          new DeleteTopicState().setTopicId(ids("topic-6"))
+        )
+        ).setTimeoutMs(timeout)).build()
```
rajinisivaram: This could move to the previous line, so the identation doesn't look odd

rajinisivaram (commented):
@jolshan Thanks for the PR, looks good. Left some minor comments. I didn't see an authorization test for delete topics where topic id is specified, but we are authorizing topic name. Can we add a test in `AuthorizerIntegrationTest`?

clients/src/main/java/org/apache/kafka/clients/admin/CreateTopicsResult.java
jolshan: I think we could either do that or just return Uuid.ZERO_UUID

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: The way the requests are built, it is not possible to provide both. But if this somehow happened, we would want to use topic IDs as specified in the KIP.

clients/src/main/java/org/apache/kafka/clients/admin/CreateTopicsResult.java
rajinisivaram: Returning `Uuid.ZERO_UUID` should be fine. The comment just seems odd, I am not sure why we have it that way in the other methods.

core/src/main/scala/kafka/server/KafkaApis.scala
rajinisivaram: Java clients won't specify both, but the protocol allows it right? Perhaps we could throw InvalidRequestException?

jolshan: Good idea.

clients/src/main/java/org/apache/kafka/clients/admin/CreateTopicsResult.java
jolshan: I thought it was odd too.

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: Thinking on this more, I can simplify the line. Especially if I make changes with the code above.

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: I've added something like this to ZkAdminManager. Let me know if it makes sense.

core/src/test/scala/unit/kafka/server/TopicIdWithOldInterBrokerProtocolTest.scala
```diff
@@ -15,12 +15,13 @@
  * limitations under the License.
  */
 
-package kafka.server
+package unit.kafka.server
```
rajinisivaram: We use `kafka.server` for the other unit tests in this directory, so we should do the same here (in Scala, the package name doesn't need to match the directory structure).

rajinisivaram (approved):
@jolshan Thanks for the PR, LGTM. Left just one minor comment.

core/src/test/scala/unit/kafka/server/TopicIdWithOldInterBrokerProtocolTest.scala
jolshan: Got it.

rajinisivaram:
The failing test FetcherTest.testEarlierOffsetResetArrivesLate is a known flaky test (https://issues.apache.org/jira/browse/KAFKA-12245). Merging to trunk.

PR #9687, opened 2020-12-04
MINOR: Move lock method outside try block

g1geordie:
The `Lock.lock` in the try block may cause `Lock.unlock`  throw exception when it throw exception .
I think it's nice to move  outside although` ReentrantLock.lock` impl doesn't throw exception.

```
 class X {
   private final ReentrantLock lock = new ReentrantLock();
   // ...

   public void m() {
     lock.lock();  // block until condition holds
     try {
       // ... method body
     } finally {
       lock.unlock()
     }
   }
 }
```
pattern recommended by 
https://docs.oracle.com/javase/10/docs/api/java/util/concurrent/locks/ReentrantLock.html


g1geordie:
@chia7712   can you help me to take a look?

chia7712 (approved):
@g1geordie LGTM. thanks for you patch. I will merge it tomorrow if no objection.

PR #9688, opened 2020-12-04
KAFKA-10017: fix flaky EOS-beta upgrade test

mjsax:
Call for review @abbccdda @ableegoldman @guozhangwang 

This PR is for `trunk` and `2.7`. PR for `2.6` is slightly different: #9690

### DIFF
--- a/streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
@@ -922,7 +922,7 @@ public class EosBetaUpgradeIntegrationTest {
         properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.REQUEST_TIMEOUT_MS_CONFIG), (int) Duration.ofSeconds(5L).toMillis());
         properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG), (int) Duration.ofSeconds(5L).minusMillis(1L).toMillis());
         properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG), MAX_POLL_INTERVAL_MS);
-        properties.put(StreamsConfig.producerPrefix(ProducerConfig.TRANSACTION_TIMEOUT_CONFIG), (int) Duration.ofMinutes(5L).toMillis());
+        properties.put(StreamsConfig.producerPrefix(ProducerConfig.TRANSACTION_TIMEOUT_CONFIG), (int) Duration.ofMinutes(1L).toMillis());
         properties.put(StreamsConfig.producerPrefix(ProducerConfig.PARTITIONER_CLASS_CONFIG), KeyPartitioner.class);
         properties.put(StreamsConfig.CACHE_MAX_BYTES_BUFFERING_CONFIG, 0);
         properties.put(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getPath() + File.separator + appDir);
--- a/streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
@@ -1041,7 +1041,15 @@ public class EosBetaUpgradeIntegrationTest {
         addAllKeys(allKeys, expectedResult);
 
         for (final Long key : allKeys) {
-            assertThat(getAllRecordPerKey(key, result), equalTo(getAllRecordPerKey(key, expectedResult)));
+            try {
+                assertThat(getAllRecordPerKey(key, result), equalTo(getAllRecordPerKey(key, expectedResult)));
+            } catch (final AssertionError error) {
+                throw new AssertionError(
+                    "expected result: " + expectedResult.stream().map(KeyValue::toString).reduce("", (kv, str) -> str.isEmpty() ? kv : str + ", " + kv) +
+                    "\nreceived records: " + result.stream().map(KeyValue::toString).reduce("", (kv, str) -> str.isEmpty() ? kv : str + ", " + kv),
+                    error
+                );
+            }
         }
     }
 
--- a/streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
@@ -23,7 +23,6 @@ import org.apache.kafka.clients.producer.Producer;
 import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.IsolationLevel;
-import org.apache.kafka.common.errors.ProducerFencedException;
 import org.apache.kafka.common.serialization.ByteArraySerializer;
 import org.apache.kafka.common.serialization.LongDeserializer;
 import org.apache.kafka.common.serialization.LongSerializer;
@@ -42,10 +41,7 @@ import org.apache.kafka.streams.integration.utils.IntegrationTestUtils.StableAss
 import org.apache.kafka.streams.kstream.KStream;
 import org.apache.kafka.streams.kstream.Transformer;
 import org.apache.kafka.streams.kstream.TransformerSupplier;
-import org.apache.kafka.streams.processor.Cancellable;
 import org.apache.kafka.streams.processor.ProcessorContext;
-import org.apache.kafka.streams.processor.PunctuationType;
-import org.apache.kafka.streams.processor.Punctuator;
 import org.apache.kafka.streams.processor.internals.DefaultKafkaClientSupplier;
 import org.apache.kafka.streams.state.KeyValueIterator;
 import org.apache.kafka.streams.state.KeyValueStore;
@@ -83,7 +79,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import static org.apache.kafka.common.utils.Utils.mkSet;
-import static org.apache.kafka.streams.integration.utils.IntegrationTestUtils.getStore;
 import static org.apache.kafka.test.TestUtils.waitForCondition;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -105,8 +100,8 @@ public class EosBetaUpgradeIntegrationTest {
     public boolean injectError;
 
     private static final int NUM_BROKERS = 3;
-    private static final int MAX_POLL_INTERVAL_MS = 100 * 1000;
-    private static final int MAX_WAIT_TIME_MS = 60 * 1000;
+    private static final int MAX_POLL_INTERVAL_MS = (int) Duration.ofSeconds(100L).toMillis();
+    private static final long MAX_WAIT_TIME_MS = Duration.ofMinutes(1L).toMillis();
 
     private static final List<KeyValue<KafkaStreams.State, KafkaStreams.State>> CLOSE =
         Collections.unmodifiableList(
@@ -151,27 +146,6 @@ public class EosBetaUpgradeIntegrationTest {
     private final AtomicInteger commitCounterClient1 = new AtomicInteger(-1);
     private final AtomicInteger commitCounterClient2 = new AtomicInteger(-1);
     private final AtomicInteger commitRequested = new AtomicInteger(0);
-
-    // Note: this pattern only works when we just have a single instance running with a single thread
-    // If we want to extend the test or reuse this CommitPunctuator we should tighten it up
-    private final AtomicBoolean requestCommit = new AtomicBoolean(false);
-    private static class CommitPunctuator implements Punctuator {
-        final ProcessorContext context;
-        final AtomicBoolean requestCommit;
-
-        public CommitPunctuator(final ProcessorContext context, final AtomicBoolean requestCommit) {
-            this.context = context;
-            this.requestCommit = requestCommit;
-        }
-
-        @Override
-        public void punctuate(final long timestamp) {
-            if (requestCommit.get()) {
-                context.commit();
-                requestCommit.set(false);
-            }
-        }
-    }
 
     private Throwable uncaughtException;
 
@@ -319,16 +293,18 @@ public class EosBetaUpgradeIntegrationTest {
             //   p-2: 10 rec + C ---> 5 rec (pending)
             //   p-3: 10 rec + C ---> 5 rec (pending)
             // crash case: (we just assumes that we inject the error for p-0; in reality it might be a different partition)
+            //             (we don't crash right away and write one record less)
             //   p-0: 10 rec + C ---> 4 rec (pending)
             //   p-1: 10 rec + C ---> 5 rec (pending)
             //   p-2: 10 rec + C ---> 5 rec (pending)
             //   p-3: 10 rec + C ---> 5 rec (pending)
             final Set<Long> cleanKeys = mkSet(0L, 1L, 2L, 3L);
-            final Set<Long> keyFilterFirstClient = keysFromInstance(streams1Alpha);
-            final long potentiallyFirstFailingKey = keyFilterFirstClient.iterator().next();
-            cleanKeys.remove(potentiallyFirstFailingKey);
+            final Set<Long> keysFirstClientAlpha = keysFromInstance(streams1Alpha);
+            final long firstFailingKeyForCrashCase = keysFirstClientAlpha.iterator().next();
+            cleanKeys.remove(firstFailingKeyForCrashCase);
 
             final List<KeyValue<Long, Long>> uncommittedInputDataBeforeFirstUpgrade = new LinkedList<>();
+            final HashMap<Long, Long> uncommittedState = new HashMap<>(committedState);
             if (!injectError) {
                 uncommittedInputDataBeforeFirstUpgrade.addAll(
                     prepareData(10L, 15L, 0L, 1L, 2L, 3L)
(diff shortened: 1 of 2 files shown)


streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -105,8 +100,8 @@
     public boolean injectError;
 
     private static final int NUM_BROKERS = 3;
-    private static final int MAX_POLL_INTERVAL_MS = 100 * 1000;
-    private static final int MAX_WAIT_TIME_MS = 60 * 1000;
+    private static final int MAX_POLL_INTERVAL_MS = (int) Duration.ofSeconds(100L).toMillis();
+    private static final long MAX_WAIT_TIME_MS = Duration.ofMinutes(1L).toMillis();
```
mjsax, line 104: Side cleanup

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -152,27 +147,6 @@
     private final AtomicInteger commitCounterClient2 = new AtomicInteger(-1);
     private final AtomicInteger commitRequested = new AtomicInteger(0);
 
-    // Note: this pattern only works when we just have a single instance running with a single thread
-    // If we want to extend the test or reuse this CommitPunctuator we should tighten it up
-    private final AtomicBoolean requestCommit = new AtomicBoolean(false);
-    private static class CommitPunctuator implements Punctuator {
```
mjsax, line 158: This punctuator was an attempt to stabilize the test, but without success. Removing it as this should be a proper fix now.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -319,24 +293,26 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
             //   p-2: 10 rec + C ---> 5 rec (pending)
             //   p-3: 10 rec + C ---> 5 rec (pending)
             // crash case: (we just assumes that we inject the error for p-0; in reality it might be a different partition)
+            //             (we don't crash right away and write one record less)
```
mjsax, line 296: Added some more details/explanations and also renames a few variables below

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -440,15 +419,18 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
             waitForRunning(stateTransitions1);
             waitForRunning(stateTransitions2);
 
-            final Set<Long> committedKeys = mkSet(0L, 1L, 2L, 3L);
+            final Set<Long> newlyCommittedKeys;
```
mjsax, line 422: This is the first fix: ie how we compute those keys.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -457,33 +439,46 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
...
             commitCounterClient1.set(0);
 
             if (!injectError) {
-                final List<KeyValue<Long, Long>> committedInputDataDuringUpgrade =
-                    prepareData(15L, 20L, 0L, 1L, 2L, 3L);
-                writeInputData(committedInputDataDuringUpgrade);
+                final List<KeyValue<Long, Long>> finishSecondBatch = prepareData(15L, 20L, 0L, 1L, 2L, 3L);
+                writeInputData(finishSecondBatch);
 
+                final List<KeyValue<Long, Long>> committedInputDataDuringUpgrade = uncommittedInputDataBeforeFirstUpgrade
```
mjsax, line 461: This is the second fix: depending on task movement, we have different set of committed records.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -457,33 +439,46 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
...
+                    .stream()
+                    .filter(pair -> !keysFirstClientAlpha.contains(pair.key))
+                    .filter(pair -> !newlyCommittedKeys.contains(pair.key))
+                    .collect(Collectors.toList());
+                committedInputDataDuringUpgrade.addAll(
+                    finishSecondBatch
+                );
+
+                expectedUncommittedResult.addAll(
+                    computeExpectedResult(finishSecondBatch, uncommittedState)
```
mjsax, line 471: For this, we needed to preserve old `uncommittedState` further above.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -774,15 +769,18 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
             waitForRunning(stateTransitions1);
             waitForRunning(stateTransitions2);
 
-            committedKeys.addAll(mkSet(0L, 1L, 2L, 3L));
+            newlyCommittedKeys.clear();
```
mjsax, line 772: Similar fix as above: we compute those keys differently now.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -792,24 +790,35 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
...
-            final List<KeyValue<Long, Long>> committedInputDataAfterUpgrade =
+            final List<KeyValue<Long, Long>> finishLastBatch =
                 prepareData(35L, 40L, 0L, 1L, 2L, 3L);
-            writeInputData(committedInputDataAfterUpgrade);
+            writeInputData(finishLastBatch);
+
+            final Set<Long> uncommittedKeys = mkSet(0L, 1L, 2L, 3L);
+            uncommittedKeys.removeAll(keysSecondClientAlphaTwo);
+            uncommittedKeys.removeAll(newlyCommittedKeys);
+            final List<KeyValue<Long, Long>> committedInputDataDuringUpgrade = uncommittedInputDataBeforeSecondUpgrade
```
mjsax, line 812: Similar to above: we need to be more flexible (ie, depend on actual task movement)

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -916,11 +917,12 @@ public void close() {
...
         properties.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, Long.MAX_VALUE);
-        properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.METADATA_MAX_AGE_CONFIG), "1000");
+        properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.METADATA_MAX_AGE_CONFIG), Duration.ofSeconds(1L).toMillis());
         properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG), "earliest");
-        properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.REQUEST_TIMEOUT_MS_CONFIG), 5 * 1000);
-        properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG), 5 * 1000 - 1);
+        properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.REQUEST_TIMEOUT_MS_CONFIG), (int) Duration.ofSeconds(5L).toMillis());
+        properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG), (int) Duration.ofSeconds(5L).minusMillis(1L).toMillis());
         properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.MAX_POLL_INTERVAL_MS_CONFIG), MAX_POLL_INTERVAL_MS);
+        properties.put(StreamsConfig.producerPrefix(ProducerConfig.TRANSACTION_TIMEOUT_CONFIG), (int) Duration.ofMinutes(5L).toMillis());
```
mjsax: I also increase the TX timeout from the to low default of 10 seconds, to avoid broker side TX-abort during the test.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -1019,7 +1021,8 @@ private void verifyUncommitted(final List<KeyValue<Long, Long>> expectedResult)
                     )
                 ),
                 MULTI_PARTITION_OUTPUT_TOPIC,
-                numberOfRecords
+                numberOfRecords,
+                MAX_WAIT_TIME_MS
```
mjsax, line 1025: Increase wait time here, too.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -1074,19 +1077,27 @@ private void addAllKeys(final Set<Long> allKeys, final List<KeyValue<Long, Long>
     }
 
     private Set<Long> keysFromInstance(final KafkaStreams streams) throws Exception {
-        final ReadOnlyKeyValueStore<Long, Long> store = getStore(
```
mjsax, line 1077: This is another fix (we did see some error for getting the state stores, too).

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -1132,7 +1143,7 @@ public ErrorInjector(final Map<String, Object> configs) {
         }
 
         @Override
-        public void commitTransaction() throws ProducerFencedException {
+        public void commitTransaction() {
```
mjsax, line 1154: Side cleanup

streams/src/test/java/org/apache/kafka/streams/integration/utils/IntegrationTestUtils.java
```diff
@@ -1161,7 +1161,7 @@ public static void verifyKeyValueTimestamps(final Properties consumerConfig,
     }
 
     private static boolean continueConsuming(final int messagesConsumed, final int maxMessages) {
-        return maxMessages <= 0 || messagesConsumed < maxMessages;
+        return maxMessages > 0 && messagesConsumed < maxMessages;
```
mjsax, line 1164: We have cases when we pass in `0` and for this case, the old code did loop forever until the timeout hits and the test fails. Seems this logic was wrong from the beginning on an we should stop fetching if `maxMessages <= 0` instead of looping forever.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
ableegoldman: I'm guessing the root source of this all is a bad assumption that the assignment would be stable if a stable `CLIENT_ID` was used? I remember we discussed that back when you first wrote this test, I'm sorry for any misinformation I supplied based on my own assumption about how the CLIENT_ID would be used :/

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
ableegoldman: 5 minutes seems kind of long, the whole test should take only a few minutes and it has 11 phases.  Would 1 minute be more reasonable? Or do we actually need this timeout to cover more than one or two phases?

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -319,24 +293,26 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
...
             //   p-0: 10 rec + C ---> 4 rec (pending)
             //   p-1: 10 rec + C ---> 5 rec (pending)
             //   p-2: 10 rec + C ---> 5 rec (pending)
             //   p-3: 10 rec + C ---> 5 rec (pending)
             final Set<Long> cleanKeys = mkSet(0L, 1L, 2L, 3L);
-            final Set<Long> keyFilterFirstClient = keysFromInstance(streams1Alpha);
-            final long potentiallyFirstFailingKey = keyFilterFirstClient.iterator().next();
-            cleanKeys.remove(potentiallyFirstFailingKey);
+            final Set<Long> keysFirstClientAlpha = keysFromInstance(streams1Alpha);
+            final long firstFailingKeyForCrashCase = keysFirstClientAlpha.iterator().next();
```
ableegoldman, line 303: Thanks for cleaning up the variable names 🙂

ableegoldman (commented):
Nice work! Seems like the underlying problem here was just that the task assignments weren't as predictable as we thought? 

Had a few minor questions but overall makes sense, if I remember how this test works

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
mjsax: Yes, the test assumed a more stable task->thread mapping during the assignment. But it turns out, that task assignment may "flip" (not sure about details)

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
mjsax: Good catch -- I set to to 5 minutes during debugging (ie, setting breakpoints). 1 minutes should be enough.

> Or do we actually need this timeout to cover more than one or two phases?

Not sure what you mean by this?

ableegoldman: Was just thinking about how long a. transaction might possibly be open. 1 minute SGTM

ableegoldman (approved):
LGTM, glad we finally have this sorted out (and that it wasn't a real bug)

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -540,18 +535,18 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
...
             //         * the eos-alpha client should not pickup the pending offsets
             //         * verify uncommitted and committed result
             //     7c. restart the first client in eos-beta mode and wait until rebalance stabilizes
             //
             // crash case:
-            //   p-0: 10 rec + C + 4 rec + A + 5 rec + C + 5 rec + C ---> 10 rec + A + 10 rec + C
-            //   p-1: 10 rec + C + 5 rec + A + 5 rec + C + 5 rec + C ---> 10 rec + A + 10 rec + C
-            //   p-2: 10 rec + C + 5 rec + C + 5 rec + A + 5 rec + C ---> 10 rec + C
-            //   p-3: 10 rec + C + 5 rec + C + 5 rec + A + 5 rec + C ---> 10 rec + C
+            //   p-0: 10 rec + C   +   4 rec + A + 5 rec + C + 5 rec + C ---> 10 rec + A + 10 rec + C
```
guozhangwang, line 546: Are these changes intentional?

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -540,18 +535,18 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
             }
 
             // 7. only for crash case:
-            //     7a. restart the second client in eos-alpha mode and wait until rebalance stabilizes
+            //     7a. restart the failed second client in eos-alpha mode and wait until rebalance stabilizes
```
guozhangwang, line 538: nit: second failed client?

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
guozhangwang: Nice catch.

Reminds me though, why the second rebalance may not be deterministic in migrating tasks back? I thought our algorithm should produce deterministic results? cc @ableegoldman

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
guozhangwang: @ableegoldman is it related to the UUID randomness? If yes please ignore my other question above.

guozhangwang (approved):
LGTM.

BTW should we re-enable this test in the same PR?

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
ableegoldman: I think "failed second client" is correct. It's the 2nd client, which has failed, not the 2nd client to have failed (English is confusing 😣 )

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
ableegoldman: Yes, I think so

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
mjsax: Yes. I wanted to improve the readability of the comment -- the additional blanks separate the the main phases of the test (each main phase write 10 records per partition that should eventually be committed).

mjsax:
> BTW should we re-enable this test in the same PR?

The test is enabled... But the test failed on the 2.6 branch PR -- Seems there is still something going on.

mjsax:
Merged to `trunk` and cherry-picked to `2.7`.

PR #9693, opened 2020-12-04
KAFKA-10825 Refactor ZK ISR updates to use AlterIsrManager

mumrah:
In an effort to consolidate the ISR write paths, this PR adapts the ZK ISR update into the new AlterIsrManager trait. This will allow us minimize divergence in the ISR update code in Partition.scala. 

TODO:
* [x] Remove expandIsr and shrinkIsr from PartitionStateStore
* [x] Move ISR propagation thread to ZkIsrManager


core/src/main/scala/kafka/server/ZkIsrManager.scala
```diff
@@ -0,0 +1,51 @@
...
+    // No async processing is done, so nothing to do here
+  }
+
+  override def clearPending(topicPartition: TopicPartition): Unit = {
+    // Since we always immediately process ZK updates and never actually enqueue anything, there is nothing to
+    // clear here so this is a no-op
+  }
+
+  override def enqueue(alterIsrItem: AlterIsrItem): Boolean = {
+    debug(s"Writing new ISR to ZooKeeper")
```
mumrah: Not sure we need this, I just wanted something in the logs to indicate which implementation is being used

hachikuji: The logging seems fine, but maybe we can enrich it with details from `alterIsrItem`?

hachikuji (commented):
Looking good. Does this mean we can get rid of `PartitionStateStore`?

mumrah:
@hachikuji yes indeed. I just pushed a commit which does just that (removes PartitionStateStore)

I also pushed a change which relocates the ISR propagation logic to ZkIsrManager

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -35,8 +36,12 @@ import scala.collection.mutable.ListBuffer
...
- * Handles the sending of AlterIsr requests to the controller. Updating the ISR is an asynchronous operation,
- * so partitions will learn about updates through LeaderAndIsr messages sent from the controller
+ * Handles updating the ISR by sending AlterIsr requests to the controller (as of 2.7) or by updating ZK directly
+ * (prior to 2.7). Updating the ISR is an asynchronous operation, so partitions will learn about the result of their
+ * request through a callback.
+ *
+ * Note that ISR state changes can still be initiated by the controller and sent to the partitions via LeaderAndIsr
+ * requests.
  */
 trait AlterIsrManager {
```
mumrah, line 46: I wonder if we should rename this IsrManager now?

mumrah:
Closed in favor of #9713 (fixed branch name for JIRA)

PR #9694, opened 2020-12-04
MINOR: Change the input type of DescribeAclsResponse#aclsResources from Collection to Iterable to avoid creating useless collection

chris8099:
Modify DescribeAclsResponse, remove Collection and change to Iterable,
Modify KafkaApis, aclsResources method

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaApis.scala
+++ b/core/src/main/scala/kafka/server/KafkaApis.scala
@@ -2509,12 +2509,10 @@ class KafkaApis(val requestChannel: RequestChannel,
             .setThrottleTimeMs(requestThrottleMs)))
       case Some(auth) =>
         val filter = describeAclsRequest.filter
-        val returnedAcls = new util.HashSet[AclBinding]()
-        auth.acls(filter).forEach(returnedAcls.add)
         sendResponseMaybeThrottle(request, requestThrottleMs =>
           new DescribeAclsResponse(new DescribeAclsResponseData()
             .setThrottleTimeMs(requestThrottleMs)
-            .setResources(DescribeAclsResponse.aclsResources(returnedAcls))))
+            .setResources(DescribeAclsResponse.aclsResources(auth.acls(filter)))))
     }
   }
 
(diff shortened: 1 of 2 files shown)


chris8099:
@chia7712  PTAL thanks

chia7712:
@chris8099 Could you rebase code to include https://github.com/apache/kafka/commit/b9640a71c42ad117ed418209f87dc2962173469d?

chris8099:
@chia7712 PTAL thanks~

chia7712:
@chris8099 Could you revise the title? It seems to me the title is not clear.

chris8099:
@chia7712 Okay, please check again.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -2509,12 +2509,10 @@ class KafkaApis(val requestChannel: RequestChannel,
             .setThrottleTimeMs(requestThrottleMs)))
       case Some(auth) =>
         val filter = describeAclsRequest.filter
```
chia7712, line 2511: Could we avoid this local variable?

chia7712 (commented):
@chris8099 Thanks for your patch. one small question is left. Please take a look.

chia7712:
@chris8099 Could you rebase code to include https://github.com/apache/kafka/commit/8db3b1a09af0bad274e07161336994610d616b35? Also, please fix the conflicting.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -2509,12 +2509,10 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-        val returnedAcls = new util.HashSet[AclBinding]()
-        auth.acls(filter).forEach(returnedAcls.add)
```
ijuma, line 2513: Note that this means we no longer filter duplicates. We should make sure that doesn't cause issues.

chia7712: Good point! It seems to me the method ```aclsResources``` should de-duplicate as it tries to traverse all elements from input

chia7712:
the related code is gone

PR #9697, opened 2020-12-04
KAFKA-10810: Replace stream threads

wcarlson5:
StreamThreads can now be replaced in the streams uncaught exception handler

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -445,10 +445,8 @@ public class KafkaStreams implements AutoCloseable {
         }
         switch (action) {
             case REPLACE_THREAD:
-                log.warn("The global thread can not be replaced. Reverting to shutting down the client.");
-
                 if (globalStreamThread != null && Thread.currentThread().getName().equals(globalStreamThread.getName())) {
-                    log.warn("The global thread can not be replaced. Reverting to shutting down the client.");
+                    log.warn("The global thread cannot be replaced. Reverting to shutting down the client.");
                     log.error("Encountered the following exception during processing " +
                             "and the registered exception handler opted to " + action + "." +
                             " The streams client is going to shut down now. ", throwable);
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -445,7 +445,16 @@ public class KafkaStreams implements AutoCloseable {
         }
         switch (action) {
             case REPLACE_THREAD:
-                StreamThread deadThread = (StreamThread) threads.stream().filter(n -> n.getName().equals(Thread.currentThread().getName())).toArray()[0];
+                log.warn("The global thread can not be replaced. Reverting to shutting down the client.");
+
+                if (globalStreamThread != null && Thread.currentThread().getName().equals(globalStreamThread.getName())) {
+                    log.warn("The global thread can not be replaced. Reverting to shutting down the client.");
+                    log.error("Encountered the following exception during processing " +
+                            "and the registered exception handler opted to " + action + "." +
+                            " The streams client is going to shut down now. ", throwable);
+                    close(Duration.ZERO);
+                }
+                final StreamThread deadThread = (StreamThread) threads.stream().filter(n -> n.getName().equals(Thread.currentThread().getName())).toArray()[0];
                 threads.remove(deadThread);
                 addStreamThread();
                 deadThread.shutdown();
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -436,6 +436,26 @@ public class KafkaStreams implements AutoCloseable {
         }
     }
 
+    private void replaceThreadHelper(final Throwable throwable) {
+        if (globalStreamThread != null && Thread.currentThread().getName().equals(globalStreamThread.getName())) {
+            log.warn("The global thread cannot be replaced. Reverting to shutting down the client.");
+            log.error("Encountered the following exception during processing " +
+                    " The streams client is going to shut down now. ", throwable);
+            close(Duration.ZERO);
+        }
+        final StreamThread deadThread = (StreamThread) Thread.currentThread();
+        threads.remove(deadThread);
+        addStreamThread();
+        deadThread.shutdown();
+        if (throwable instanceof RuntimeException) {
+            throw (RuntimeException) throwable;
+        } else if (throwable instanceof Error) {
+            throw (Error) throwable;
+        } else {
+            throw new RuntimeException("Unexpected checked exception caught in the uncaught exception handler", throwable);
+        }
+    }
+
     private void handleStreamsUncaughtException(final Throwable throwable,
                                                 final StreamsUncaughtExceptionHandler streamsUncaughtExceptionHandler) {
         final StreamsUncaughtExceptionHandler.StreamThreadExceptionResponse action = streamsUncaughtExceptionHandler.handle(throwable);
@@ -445,24 +465,8 @@ public class KafkaStreams implements AutoCloseable {
         }
         switch (action) {
             case REPLACE_THREAD:
-                if (globalStreamThread != null && Thread.currentThread().getName().equals(globalStreamThread.getName())) {
-                    log.warn("The global thread cannot be replaced. Reverting to shutting down the client.");
-                    log.error("Encountered the following exception during processing " +
-                            "and the registered exception handler opted to " + action + "." +
-                            " The streams client is going to shut down now. ", throwable);
-                    close(Duration.ZERO);
-                }
-                final StreamThread deadThread = (StreamThread) Thread.currentThread();
-                threads.remove(deadThread);
-                addStreamThread();
-                deadThread.shutdown();
-                if (throwable instanceof RuntimeException) {
-                    throw (RuntimeException) throwable;
-                } else if (throwable instanceof Error) {
-                    throw (Error) throwable;
-                } else {
-                    throw new RuntimeException("Unexpected checked exception caught in the uncaught exception handler", throwable);
-                }
+                replaceThreadHelper(throwable);
+                break;
             case SHUTDOWN_CLIENT:
                 log.error("Encountered the following exception during processing " +
                         "and the registered exception handler opted to " + action + "." +
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -441,7 +441,7 @@ public class KafkaStreams implements AutoCloseable {
             log.warn("The global thread cannot be replaced. Reverting to shutting down the client.");
             log.error("Encountered the following exception during processing " +
                     " The streams client is going to shut down now. ", throwable);
-            close(Duration.ZERO);
+            closeToError();
         }
         final StreamThread deadThread = (StreamThread) Thread.currentThread();
         threads.remove(deadThread);
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -444,6 +444,18 @@ public class KafkaStreams implements AutoCloseable {
                     "The old handler will be ignored as long as a new handler is set.");
         }
         switch (action) {
+            case REPLACE_THREAD:
+                StreamThread deadThread = (StreamThread) threads.stream().filter(n -> n.getName().equals(Thread.currentThread().getName())).toArray()[0];
+                threads.remove(deadThread);
+                addStreamThread();
+                deadThread.shutdown();
+                if (throwable instanceof RuntimeException) {
+                    throw (RuntimeException) throwable;
+                } else if (throwable instanceof Error) {
+                    throw (Error) throwable;
+                } else {
+                    throw new RuntimeException("Unexpected checked exception caught in the uncaught exception handler", throwable);
+                }
             case SHUTDOWN_CLIENT:
                 log.error("Encountered the following exception during processing " +
                         "and the registered exception handler opted to " + action + "." +
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -436,12 +436,12 @@ public class KafkaStreams implements AutoCloseable {
         }
     }
 
-    private void replaceThreadHelper(final Throwable throwable) {
+    private void replaceStreamThread(final Throwable throwable) {
         if (globalStreamThread != null && Thread.currentThread().getName().equals(globalStreamThread.getName())) {
             log.warn("The global thread cannot be replaced. Reverting to shutting down the client.");
             log.error("Encountered the following exception during processing " +
                     " The streams client is going to shut down now. ", throwable);
-            closeToError();
+            close(Duration.ZERO);
         }
         final StreamThread deadThread = (StreamThread) Thread.currentThread();
         threads.remove(deadThread);
@@ -465,7 +465,7 @@ public class KafkaStreams implements AutoCloseable {
         }
         switch (action) {
             case REPLACE_THREAD:
-                replaceThreadHelper(throwable);
+                replaceStreamThread(throwable);
                 break;
             case SHUTDOWN_CLIENT:
                 log.error("Encountered the following exception during processing " +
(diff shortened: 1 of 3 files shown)


streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
```diff
@@ -145,6 +148,26 @@ public void shouldShutdownClient() throws InterruptedException {
...
+            AtomicInteger count = new AtomicInteger();
+            kafkaStreams.setUncaughtExceptionHandler(exception -> {
+                count.getAndIncrement();
+                return REPLACE_THREAD;
+            });
+
+            StreamsTestUtils.startKafkaStreamsAndWaitForRunningState(kafkaStreams);
+
+            produceMessages(0L, inputTopic, "A");
+            TestUtils.waitForCondition(() -> count.get() > 2, DEFAULT_TIMEOUT, "At least 3 threads have died");
```
wcarlson5, line 164: All of the original threads have died and been replaced and it is still running

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -444,6 +444,18 @@ private void handleStreamsUncaughtException(final Throwable throwable,
                     "The old handler will be ignored as long as a new handler is set.");
         }
         switch (action) {
+            case REPLACE_THREAD:
+                StreamThread deadThread = (StreamThread) threads.stream().filter(n -> n.getName().equals(Thread.currentThread().getName())).toArray()[0];
+                threads.remove(deadThread);
+                addStreamThread();
+                deadThread.shutdown();
+                if (throwable instanceof RuntimeException) {
```
wcarlson5: We need to throw the error or the task gets lost and we drop records

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -444,6 +444,18 @@ private void handleStreamsUncaughtException(final Throwable throwable,
                     "The old handler will be ignored as long as a new handler is set.");
         }
         switch (action) {
+            case REPLACE_THREAD:
+                StreamThread deadThread = (StreamThread) threads.stream().filter(n -> n.getName().equals(Thread.currentThread().getName())).toArray()[0];
+                threads.remove(deadThread);
```
wcarlson5: removing from the thread list does 2 things

1. keeps DEAD threads out of the list as kip-663 dictates
2. ensures the next thread has the same name

wcarlson5 (commented):
@cadonna part 6 is ready for review

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -444,6 +444,27 @@ private void handleStreamsUncaughtException(final Throwable throwable,
                     "The old handler will be ignored as long as a new handler is set.");
         }
         switch (action) {
+            case REPLACE_THREAD:
+                log.warn("The global thread can not be replaced. Reverting to shutting down the client.");
```
lct45: nit: "can not" -> "cannot", same below

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -444,6 +444,25 @@ private void handleStreamsUncaughtException(final Throwable throwable,
...
         switch (action) {
+            case REPLACE_THREAD:
+                if (globalStreamThread != null && Thread.currentThread().getName().equals(globalStreamThread.getName())) {
+                    log.warn("The global thread cannot be replaced. Reverting to shutting down the client.");
+                    log.error("Encountered the following exception during processing " +
+                            "and the registered exception handler opted to " + action + "." +
+                            " The streams client is going to shut down now. ", throwable);
+                    close(Duration.ZERO);
+                }
+                final StreamThread deadThread = (StreamThread) threads.stream().filter(n -> n.getName().equals(Thread.currentThread().getName())).toArray()[0];
```
vvcephei: Is this equivalent to `Thread.currentThead()`?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
vvcephei: To ensure #2 holds, do we need to have a mutex on thread creation? Or does it not matter?

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
```diff
@@ -202,6 +213,28 @@ private void testShutdownApplication(final int numThreads) throws InterruptedExc
...
+                    return SHUTDOWN_CLIENT;
+                }
+                return REPLACE_THREAD;
+            });
+            StreamsTestUtils.startKafkaStreamsAndWaitForRunningState(kafkaStreams);
+
+            produceMessages(0L, inputTopic, "A");
+            waitForApplicationState(Collections.singletonList(kafkaStreams), KafkaStreams.State.NOT_RUNNING, DEFAULT_DURATION);
+
+            assertThat(processorValueCollector.size(), equalTo(3));
```
vvcephei: It's not obvious to me how this verifies that the thread actually got replaced. Maybe an explanatory comment is in order?

vvcephei (commented):
Thanks, @wcarlson5 ! I had a few thoughts above.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: It won't matter. Ensures is too strong of a word. IF there has been other threads removed before this it may have a different name. However this ensures that the replacement thread will never have a thread index larger than the number of threads

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Yep, it is now :)

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
wcarlson5: good idea

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -444,6 +444,25 @@ private void handleStreamsUncaughtException(final Throwable throwable,
...
+                    log.warn("The global thread cannot be replaced. Reverting to shutting down the client.");
+                    log.error("Encountered the following exception during processing " +
+                            "and the registered exception handler opted to " + action + "." +
+                            " The streams client is going to shut down now. ", throwable);
+                    close(Duration.ZERO);
+                }
+                final StreamThread deadThread = (StreamThread) Thread.currentThread();
+                threads.remove(deadThread);
+                addStreamThread();
+                deadThread.shutdown();
```
cadonna: Do we need to shutdown the dead stream thread? `completeShutDown()` will be called anyways.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -444,6 +444,25 @@ private void handleStreamsUncaughtException(final Throwable throwable,
...
+                if (throwable instanceof RuntimeException) {
+                    throw (RuntimeException) throwable;
+                } else if (throwable instanceof Error) {
+                    throw (Error) throwable;
+                } else {
+                    throw new RuntimeException("Unexpected checked exception caught in the uncaught exception handler", throwable);
+                }
```
cadonna: I think it would be cleaner to extract this code to a separate method.

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
```diff
@@ -202,6 +213,29 @@ private void testShutdownApplication(final int numThreads) throws InterruptedExc
...
+            //because we only have 2 threads at the start and each record kills a thread we must have replaced threads
```
cadonna: Could you please be a bit clearer in the explanatory comment? BTW, we execute this test also once with just one stream thread so the 2 stream threads in the comment are not correct. Also, wouldn't it be better to explain the verification in the call to `assertThat()` instead of in a comment? You can pass a reason to the method.

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
```diff
@@ -202,6 +213,29 @@ private void testShutdownApplication(final int numThreads) throws InterruptedExc
...
+        }
+    }
 }
```
cadonna, line 279: A test is missing for a global stream thread that calls the uncaught exception handler.

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
```diff
@@ -202,6 +213,29 @@ private void testShutdownApplication(final int numThreads) throws InterruptedExc
...
+            kafkaStreams.setUncaughtExceptionHandler((t, e) -> fail("should not hit old handler"));
+
+            final AtomicInteger count = new AtomicInteger();
+            kafkaStreams.setUncaughtExceptionHandler(exception -> {
+                count.getAndIncrement();
+                if (count.get() > 2) {
+                    return SHUTDOWN_CLIENT;
+                }
+                return REPLACE_THREAD;
+            });
```
cadonna: I think it would be better to have a test that shows that a new thread that replaced a failed one, actually is able to process records. So, I would let the new thread process some records and then shutdown the client with a normal close.

Maybe similar applies to the shutdown tests. First let the client/application process some records and then throw an exception that shuts down the client/application. I guess, this last paragraph is something for a separate PR.

cadonna (commented):
Thanks for the PR @wcarlson5 !

Here my feedback!

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I don't think it matters, it just set the thread state earlier, but we can delete it

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: we can do that

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
wcarlson5: I edited the test to be based on the number of thread instead of hard coding. And gave a reason

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
wcarlson5: We can change the test so that we verify the replaced threads can process records. 

I am not sure that is necessary for the shutdown as testing if streams can process some records once started should be tested elsewhere, but in any case I think that the PR is not the place for this discussion

wcarlson5 (commented):
@cadonna here are some changes

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
wcarlson5: added

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -436,6 +436,26 @@ private void defaultStreamsUncaughtExceptionHandler(final Throwable throwable) {
         }
     }
 
+    private void replaceThreadHelper(final Throwable throwable) {
```
cadonna: Please rename to `replaceStreamThread()`.

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
```diff
@@ -155,16 +171,46 @@ public void shouldShutdownSingleThreadApplication() throws InterruptedException
...
+            ),
+            inputTopic,
+            Consumed.with(Serdes.String(), Serdes.String()),
+            () -> new ShutdownProcessor(processorValueCollector)
+        );
+        properties.put(StreamsConfig.NUM_STREAM_THREADS_CONFIG, 0);
+
+        try (final KafkaStreams kafkaStreams = new KafkaStreams(builder.build(), properties)) {
+            kafkaStreams.setUncaughtExceptionHandler((t, e) -> fail("should not hit old handler"));
+            kafkaStreams.setUncaughtExceptionHandler(exception -> SHUTDOWN_CLIENT);
```
cadonna: The point of this test should be that a global stream thread is not replaced but the client is shutdown instead. Hence, the uncaught exception handler should return `REPLACE_THREAD`, not `SHUTDOWN_CLIENT`.

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
```diff
@@ -155,16 +171,46 @@ public void shouldShutdownSingleThreadApplication() throws InterruptedException
         testShutdownApplication(1);
     }
 
+    @Test
+    public void testGlobalThreadException() throws InterruptedException {
```
cadonna: Please rename to something like `shouldShutDownClientIfGlobalStreamThreadWantsToReplaceThread()`.

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
```diff
@@ -27,6 +27,7 @@
      * Enumeration that describes the response from the exception handler.
      */
     enum StreamThreadExceptionResponse {
+        REPLACE_THREAD(0, "REPLACE_STREAM_THREAD"),
```
cadonna: I would rename it to `REPLACE_STREAM_THREAD`.

vvcephei: Oh, good catch.

Just a quick question: did we misname this option in the KIP? A StreamThread is a specific kind of thread. What I mean is that a GlobalStreamThread is _not_ a StreamThread. Perhaps `REPLACE_THREAD` and `SHUTDOWN_THREAD` would have been better, more general names, for these. If you agree, I think we can just amend the KIP and fix it in this PR.

wcarlson5: It should be the string is `REPLACE_THREAD`. Ill just fix that, its just left over from a period of time where we thought of calling it `REPLACE_STREAM_THREAD`

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
wcarlson5: I guess we can test that here at the same time

streams/src/main/java/org/apache/kafka/streams/errors/StreamsUncaughtExceptionHandler.java
cadonna: I actually would be in favor of calling the enum value `REPLACE_STREAM_THREAD`. A stream thread is a stream thread and a global stream thread is a global stream thread. I am aware that the KIP calls the enum value differently, but we also have a config that is called 'NUM_STREAM_THREADS_CONFIG' and we have also 'addStreamThread()' and `removeStreamThread()`. So I guess, the name to the outside of this is stream thread and not thread. We have also other threads in Kafka Streams like the state directory cleaner thread and the RocksDB metrics recording thread.

wcarlson5: What if we add an option to replace the global thread?

cadonna: Ah, now I got it! Sorry! Makes sense! In that case we can reuse `REPLACE_THREAD` also for the global stream thread. Forgot about that!

cadonna (approved):
Thanks @wcarlson5!

LGTM!

vvcephei (approved):
Thanks @wcarlson5 !

zydzjy:
what can i do with replace new thread?Can i skip the current offset in exception handler,cause the exception occured by the data with current offset?

wcarlson5:
@zydzjy I don't think you can do that yet, at least not with anything I have added here. The idea is to recover from transient expectations without having to restart the client.

PR #9698, opened 2020-12-04
KAFKA-10811: Correct the MirrorConnectorsIntegrationTest to correctly mask the exit procedures

rhauch:
Normally the `EmbeddedConnectCluster` class masks the `Exit` procedures using within the Connect worker. This normally works great when a single instance of the embedded cluster is used. However, the `MirrorConnectorsIntegrationTest` uses two `EmbeddedConnectCluster` instances, and when the first one is stopped it would reset the (static) exit procedures, and any problems during shutdown of the second embedded Connect cluster would cause the worker to shut down the JVM running the tests.

Instead, the `MirrorConnectorsIntegrationTest` class should mask the `Exit` procedures and instruct the `EmbeddedConnectClusters` instances (via the existing builder method) to not mask the procedures.

Ideally this should also be backported to `2.7`, `2.6`, and `2.5` branches.

### DIFF
--- a/connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
+++ b/connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
@@ -24,6 +24,7 @@ import org.apache.kafka.clients.consumer.OffsetAndMetadata;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.connect.util.clusters.EmbeddedConnectCluster;
 import org.apache.kafka.connect.util.clusters.EmbeddedKafkaCluster;
+import org.apache.kafka.connect.util.clusters.UngracefulShutdownException;
 import org.apache.kafka.test.IntegrationTest;
 import org.apache.kafka.common.utils.Exit;
 import org.junit.After;
@@ -44,13 +45,11 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import static org.apache.kafka.test.TestUtils.waitForCondition;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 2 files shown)


rhauch:
@C0urante as mentioned offline, here is my fix for the MM2 integration test frequently exiting the builds.

C0urante:
This is definitely an improvement not only to the flaky MM2 test but to the integration testing framework in general. ✅

As a follow-up item, could we track the unclean shutdowns for MM2 we've observed from these tests in a separate ticket? Given the flakiness that they're causing right now, it seems best to temporarily ignore them, but ideally it should be possible to run MM2 without seeing `ERROR`-level stack traces in logs on shutdown.

rhauch:
@C0urante is there a KAFKA issue for those stack traces? If not, please create one.

C0urante:
@rhauch it's not so much about the stack traces in these tests as it is about them presumably appearing when MM2 is run for real, and being caused by unclean worker shutdown. I've written up [KAFKA-10812](https://issues.apache.org/jira/browse/KAFKA-10812) to track the issue.

rhauch:
Rebased on top of `trunk` to incorporate an upstream fix for the StreamsThreadTest compile error.

mimaison (approved):
Thanks @rhauch, LGTM

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
```diff
@@ -75,14 +74,45 @@
...
     private static final int OFFSET_SYNC_DURATION_MS = 30_000;
 
-    private final AtomicBoolean exited = new AtomicBoolean(false);
+    private volatile boolean shuttingDown;
     private Map<String, String> mm2Props;
     private MirrorMakerConfig mm2Config;
     private EmbeddedConnectCluster primary;
     private EmbeddedConnectCluster backup;
 
+    private Exit.Procedure exitProcedure;
```
chia7712, line 83: It seems ```exitProcedure``` and ```haltProcedure``` can be local variable

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
```diff
@@ -194,20 +224,27 @@ private void waitUntilMirrorMakerIsRunning(EmbeddedConnectCluster connectCluster
...
+                backup.deleteConnector(x);
+            }
+            deleteAllTopics(primary.kafka());
+            deleteAllTopics(backup.kafka());
         } finally {
-            Exit.resetExitProcedure();
+            shuttingDown = true;
+            try {
+                try {
+                    primary.stop();
```
chia7712, line 240: How about using ```Utils.closeQuietly``` to replace this nested try-block?

chia7712 (approved):
@rhauch Thanks for this great fix!

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
rhauch: Perhaps, but the log messages for each is different, so IMO it's better to keep them separate.

connect/mirror/src/test/java/org/apache/kafka/connect/mirror/MirrorConnectorsIntegrationTest.java
rhauch: `EmbeddedConnectCluster` is not `AutoCloseable`, which means we can't use that. Since this PR is to fix broken builds, I'd like to minimize the additional changes, so I'll issue a followup PR.

PR #9700, opened 2020-12-05
KAFKA-10813: InvalidProducerEpoch should be caught and throw as TaskMigrated

abbccdda:
We should catch InvalidProducerEpoch and rethrow as TaskMigrated, similar to ProducerFenced.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
@@ -23,6 +23,7 @@ import org.apache.kafka.common.PartitionInfo;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.errors.AuthenticationException;
 import org.apache.kafka.common.errors.AuthorizationException;
+import org.apache.kafka.common.errors.InvalidProducerEpochException;
 import org.apache.kafka.common.errors.InvalidTopicException;
 import org.apache.kafka.common.errors.OffsetMetadataTooLarge;
 import org.apache.kafka.common.errors.OutOfOrderSequenceException;
@@ -199,7 +200,9 @@ public class RecordCollectorImpl implements RecordCollector {
         if (isFatalException(exception)) {
             errorMessage += "\nWritten offsets would not be recorded and no more records would be sent since this is a fatal error.";
             sendException.set(new StreamsException(errorMessage, exception));
-        } else if (exception instanceof ProducerFencedException || exception instanceof OutOfOrderSequenceException) {
+        } else if (exception instanceof ProducerFencedException ||
+                exception instanceof InvalidProducerEpochException ||
+                exception instanceof OutOfOrderSequenceException) {
             errorMessage += "\nWritten offsets would not be recorded and no more records would be sent since the producer is fenced, " +
                 "indicating the task may be migrated out";
             sendException.set(new TaskMigratedException(errorMessage, exception));
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsProducer.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsProducer.java
@@ -184,12 +184,12 @@ public class StreamsProducer {
         transactionInitialized = false;
     }
 
-    private void maybeBeginTransaction() throws ProducerFencedException {
+    private void maybeBeginTransaction() {
         if (eosEnabled() && !transactionInFlight) {
             try {
                 producer.beginTransaction();
                 transactionInFlight = true;
-            } catch (final ProducerFencedException error) {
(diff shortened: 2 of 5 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsProducer.java
```diff
@@ -184,12 +184,12 @@ public void resetProducer() {
...
     }
 
-    private void maybeBeginTransaction() throws ProducerFencedException {
+    private void maybeBeginTransaction() {
         if (eosEnabled() && !transactionInFlight) {
             try {
                 producer.beginTransaction();
                 transactionInFlight = true;
-            } catch (final ProducerFencedException error) {
+            } catch (final ProducerFencedException | InvalidProducerEpochException error) {
```
ableegoldman, line 192:
```suggestion
            } catch (final ProducerFencedException || InvalidProducerEpochException error) {
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
```diff
@@ -199,7 +200,9 @@ private void recordSendError(final String topic, final Exception exception, fina
         if (isFatalException(exception)) {
             errorMessage += "\nWritten offsets would not be recorded and no more records would be sent since this is a fatal error.";
             sendException.set(new StreamsException(errorMessage, exception));
-        } else if (exception instanceof ProducerFencedException || exception instanceof OutOfOrderSequenceException) {
+        } else if (exception instanceof ProducerFencedException ||
+                exception instanceof InvalidProducerEpochException ||
+                exception instanceof OutOfOrderSequenceException) {
```
ableegoldman, line 205: Why is that here, we catch these three exceptions -- ProducerFenced, InvalidProducerEpoch, and OutOfOrderSequence -- and wrap them as TaskMigratedException, while in `StreamsProducer#send`, we catch  ProducerFenced, InvalidProducerEpoch, and UnknownProducerId exceptions and wrap those as TaskMigrated?

ableegoldman: Just to clarify, I think my question/confusion is twofold:

1) Why is it OutOfOrderSequence in one place and UnknownProducerId in another?
2) What is the difference between these two code paths? Is it really possible for example for the ProducerFencedException to sometimes be thrown directly from `Producer#send`, and sometimes be passed along through the callback ( as in `streamsProducer.send(serializedRecord, (metadata, exception)` )

ableegoldman (commented):
Thanks for catching this. I had a few questions since I'm not that familiar with the EOS code and exception handling I guess. Would be good if @mjsax or @guozhangwang could take a look as well

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
abbccdda: Wasn't over thinking here since this is a blocker-fix, so we don't want to trigger any regression here by changing existing exception catching logic. Would be good to do this as a follow-up I guess.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamsProducer.java
abbccdda: This won't work I suppose?

ableegoldman: Oh lol sorry I was just comparing it with the other block. Obviously nvm

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
ableegoldman: Ok SG, agree we should keep things simple since this is a last minute blocker

ableegoldman (approved):
LGTM, ok to look into the question about the different exceptions as followup work, or as part of your work to clean up exception handling in the Producer. Thanks for the fix!

mjsax (approved):
LGTM

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
guozhangwang: I thought about this a bit, and I think both `UnknownProducerId` and `OutOfOrderSequence` could be possibly thrown from the caller or directly (though in the later case they would be wrapped as KafkaException).

I created two tickets, one for producer and one for streams to improve the general picture moving forward.

https://issues.apache.org/jira/browse/KAFKA-10829
https://issues.apache.org/jira/browse/KAFKA-10830

guozhangwang (approved):

abbccdda:
Cherry-picked to 2.7

PR #9703, opened 2020-12-05
KAFKA-10697 Remove ProduceResponse.responses

UnityLung:
issue : https://issues.apache.org/jira/browse/KAFKA-10697

Remove the old method Map <TopicPartition, PartitionResponse> responses() 
Use the new method public ProduceResponseData data() instead.


ijuma:
Thanks for the PR. I don't see `responses` being removed.

UnityLung:
@ijuma  Thanks for your reply.
I'm going to do it for remove "responses".

PR #9707, opened 2020-12-07
KAFKA-10790 Detect/Prevent Deadlock on Producer Network Thread

g1geordie:
KafkaProducer.flush method in callback will cause deadlock .

because flush method  wait the future complete
``` java
//  KafkaProducer
public void flush() {
   .....
        try {
            this.accumulator.awaitFlushCompletion();
        } catch (InterruptedException e) {
            throw new InterruptException("Flush interrupted.", e);
        }
}

//RecordAccumulator
    public void awaitFlushCompletion() throws InterruptedException {
        try {
            for (ProducerBatch batch : this.incomplete.copyAll())
                batch.produceFuture.await();
        } finally {
            this.flushesInProgress.decrementAndGet();
        }
    }
```

but  future complete after the callback .
```java
//ProducerBatch
 private void completeFutureAndFireCallbacks(long baseOffset, long logAppendTime, RuntimeException exception) {
     ...
        // execute callbacks
        for (Thunk thunk : thunks) {
            try {
                if (exception == null) {
                    RecordMetadata metadata = thunk.future.value();
                    if (thunk.callback != null)
                        thunk.callback.onCompletion(metadata, null);
                } else {
                    if (thunk.callback != null)
                        thunk.callback.onCompletion(null, exception);
                }
            } catch (Exception e) {
                log.error("Error executing user-provided callback on message for topic-partition '{}'", topicPartition, e);
            }
        }

        produceFuture.done();
 }
```

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
@@ -1083,6 +1083,10 @@ public class KafkaProducer<K, V> implements Producer<K, V> {
                     " configuration.");
     }
 
+    private boolean isInvokedFromCallback() {
+        return Thread.currentThread() == this.ioThread;
+    }
+
     /**
      * Invoking this method makes all buffered records immediately available to send (even if <code>linger.ms</code> is
      * greater than 0) and blocks on the completion of the requests associated with these records. The post-condition
@@ -1116,9 +1120,13 @@ public class KafkaProducer<K, V> implements Producer<K, V> {
      * </p>
      *
      * @throws InterruptException If the thread is interrupted while blocked
+     * @throws IllegalStateException If the thread is callback thread. Calling flush() from callback is not allowed since it makes deadlock.
      */
     @Override
     public void flush() {
+        if (isInvokedFromCallback())
+            throw new IllegalStateException("Calling flush() from callback is not allowed since it makes deadlock.");
+
         log.trace("Flushing accumulated records in producer.");
         this.accumulator.beginFlush();
         this.sender.wakeup();
@@ -1205,9 +1213,8 @@ public class KafkaProducer<K, V> implements Producer<K, V> {
 
         // this will keep track of the first encountered exception
         AtomicReference<Throwable> firstException = new AtomicReference<>();
-        boolean invokedFromCallback = Thread.currentThread() == this.ioThread;
         if (timeoutMs > 0) {
-            if (invokedFromCallback) {
+            if (isInvokedFromCallback()) {
                 log.warn("Overriding close timeout {} ms to 0 ms in order to prevent useless blocking due to self-join. " +
                         "This means you have incorrectly invoked close with a non-zero timeout from the producer call-back.",
                         timeoutMs);
@@ -1231,7 +1238,7 @@ public class KafkaProducer<K, V> implements Producer<K, V> {
                     "within timeout {} ms.", timeoutMs);
             this.sender.forceClose();
             // Only join the sender thread when not calling from callback.
-            if (!invokedFromCallback) {
+            if (!isInvokedFromCallback()) {
                 try {
                     this.ioThread.join();
                 } catch (InterruptedException e) {
--- a/clients/src/test/java/org/apache/kafka/clients/producer/KafkaProducerTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/producer/KafkaProducerTest.java
@@ -392,6 +392,44 @@ public class KafkaProducerTest {
         };
     }
 
+    @Test(timeout = 10000)
+    public void testFlushInCallbackNotCauseDeadlock() {
+        Cluster oneLeaderCluster = new Cluster(
+                "dummy",
+                nodes,
+                Collections.singletonList(new PartitionInfo(topic, 0, host1, null, null)),
+                Collections.emptySet(),
+                Collections.emptySet());
+
+        Map<String, Object> configs = new HashMap<>();
+        configs.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9999");
+        ProducerMetadata metadata = mock(ProducerMetadata.class);
+
+        when(metadata.fetch()).thenReturn(emptyCluster, oneLeaderCluster);
+
+        Time time = new MockTime();
+        MockClient mockClient = new MockClient(time, metadata);
+        mockClient.prepareResponse(null);
+
+        AtomicReference<Exception> exceptionThrow = new AtomicReference<>();
+        try (KafkaProducer<String, String> producer = kafkaProducer(configs,
+                new StringSerializer(), new StringSerializer(), metadata, mockClient, null, time)) {
+
+            ProducerRecord<String, String> record = new ProducerRecord<>(topic, "value");
+            producer.send(record, (metadata1, exception) -> {
+                try {
+                    producer.flush();
+                } catch (Exception e) {
+                    exceptionThrow.set(e);
+                }
+            });
+        }
+
+        Assert.assertNotNull(exceptionThrow.get());
+        Assert.assertEquals(IllegalStateException.class, exceptionThrow.get().getClass());
+
+    }
+
     @Test
     public void testMetadataFetch() throws InterruptedException {
         Map<String, Object> configs = new HashMap<>();


g1geordie:
@chia7712 Can you help me review?

clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
```diff
@@ -1115,6 +1115,10 @@ private void ensureValidRecordSize(int size) {
      */
     @Override
     public void flush() {
+        boolean invokedFromCallback = Thread.currentThread() == this.ioThread;
```
chia7712: As this check is used by ```close``` also, could you add a helper method to return boolean?

clients/src/test/java/org/apache/kafka/clients/producer/KafkaProducerTest.java
```diff
@@ -391,6 +391,23 @@ Sender newSender(LogContext logContext, KafkaClient kafkaClient, ProducerMetadat
...
+        Map<String, Object> configs = new HashMap<>();
+        configs.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9999");
+        ProducerMetadata metadata = mock(ProducerMetadata.class);
+
+        when(metadata.fetch()).thenReturn(emptyCluster, emptyCluster, emptyCluster, onePartitionCluster);
+
+        KafkaProducer<String, String> producer = producerWithOverrideNewSender(configs, metadata);
+        ProducerRecord<String, String> record = new ProducerRecord<>(topic, "value");
+
+        producer.send(record, (metadata1, exception) -> producer.flush());
```
chia7712: Could you add a flag after ```producer.flush()```? We should make sure ```producer.flush()``` fails.

g1geordie: Thank you  . I add a Assert.fail after flush

clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
```diff
@@ -1079,6 +1079,13 @@ private void ensureValidRecordSize(int size) {
                     " configuration.");
     }
 
+    /**
+     * @return is called in callback.
```
chia7712: I feel the method name is good to be document.

clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
```diff
@@ -1201,7 +1211,7 @@ private void close(Duration timeout, boolean swallowException) {
 
         // this will keep track of the first encountered exception
         AtomicReference<Throwable> firstException = new AtomicReference<>();
-        boolean invokedFromCallback = Thread.currentThread() == this.ioThread;
+        boolean invokedFromCallback = isInvokedFromCallback();
```
chia7712: this local variable is redundant

clients/src/test/java/org/apache/kafka/clients/producer/KafkaProducerTest.java
```diff
@@ -391,6 +391,27 @@ Sender newSender(LogContext logContext, KafkaClient kafkaClient, ProducerMetadat
...
+        ProducerMetadata metadata = mock(ProducerMetadata.class);
+
+        when(metadata.fetch()).thenReturn(emptyCluster, emptyCluster, emptyCluster, onePartitionCluster);
+
+        KafkaProducer<String, String> producer = producerWithOverrideNewSender(configs, metadata);
+        ProducerRecord<String, String> record = new ProducerRecord<>(topic, "value");
+
+        producer.send(record, (metadata1, exception) -> {
+            producer.flush();
+            Assert.fail("Flush method needs to throw exception.");
```
chia7712: I don't think this is valid since this callback is triggered by another thread. The assert failure can't be aware by test thread.

chia7712 (commented):
@g1geordie Thanks for your patch. Some comments are left. Please take a look.

clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
g1geordie: remove

clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
g1geordie: direct to method

clients/src/test/java/org/apache/kafka/clients/producer/KafkaProducerTest.java
g1geordie: move assert to main thread

g1geordie (commented):
thanks for all suggestion .

clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
```diff
@@ -1115,6 +1119,9 @@ private void ensureValidRecordSize(int size) {
      */
     @Override
     public void flush() {
+        if (isInvokedFromCallback())
+            throw new UnsupportedOperationException("Can't specify flush method in callback.");
```
chia7712: It seems to me ```IllegalStateException``` is more suitable for this case. Also, please update docs of ```flush()```

clients/src/test/java/org/apache/kafka/clients/producer/KafkaProducerTest.java
```diff
@@ -391,6 +391,31 @@ Sender newSender(LogContext logContext, KafkaClient kafkaClient, ProducerMetadat
...
+        Map<String, Object> configs = new HashMap<>();
+        configs.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9999");
+        ProducerMetadata metadata = mock(ProducerMetadata.class);
+
+        when(metadata.fetch()).thenReturn(emptyCluster, emptyCluster, emptyCluster, onePartitionCluster);
+
+        KafkaProducer<String, String> producer = producerWithOverrideNewSender(configs, metadata);
+        ProducerRecord<String, String> record = new ProducerRecord<>(topic, "value");
+
+        AtomicBoolean  exceptionThrow = new AtomicBoolean(false);
```
chia7712: Could you save the exception caused by ```producer.flush()``` and then check the error type in the test thread?

chia7712 (commented):
@g1geordie Thanks for your updating. Two minor comments are left.

clients/src/test/java/org/apache/kafka/clients/producer/KafkaProducerTest.java
```diff
@@ -391,6 +391,31 @@ Sender newSender(LogContext logContext, KafkaClient kafkaClient, ProducerMetadat
...
 
+    @Test(timeout = 10000)
+    public void testFlushInCallbackNotCauseDeadlock() throws InterruptedException {
+        Map<String, Object> configs = new HashMap<>();
+        configs.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:9999");
+        ProducerMetadata metadata = mock(ProducerMetadata.class);
+
+        when(metadata.fetch()).thenReturn(emptyCluster, emptyCluster, emptyCluster, onePartitionCluster);
+
+        KafkaProducer<String, String> producer = producerWithOverrideNewSender(configs, metadata);
```
chia7712: Could you use try block?

clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
```diff
@@ -1116,11 +1116,12 @@ private boolean isInvokedFromCallback() {
      * </p>
      *
      * @throws InterruptException If the thread is interrupted while blocked
+     * @throws IllegalStateException If flush method in callback
```
chia7712: ```If the thread is callback thread. Calling flush() from callback is not allowed since it makes deadlock.```

clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java
```diff
@@ -1116,11 +1116,12 @@ private boolean isInvokedFromCallback() {
...
      */
     @Override
     public void flush() {
         if (isInvokedFromCallback())
-            throw new UnsupportedOperationException("Can't specify flush method in callback.");
+            throw new IllegalStateException("Can't specify flush method in callback.");
```
chia7712: ```Calling flush() from callback is not allowed since it makes deadlock.```

g1geordie:
@chia7712  thanks again

clients/src/test/java/org/apache/kafka/clients/producer/KafkaProducerTest.java
```diff
@@ -131,6 +131,14 @@
             Collections.singletonList(new PartitionInfo(topic, 0, null, null, null)),
             Collections.emptySet(),
             Collections.emptySet());
+
+    private final Cluster oneLeaderCluster = new Cluster(
```
chia7712: Pardon me, why this change is required.

chia7712 (approved):

clients/src/test/java/org/apache/kafka/clients/producer/KafkaProducerTest.java
g1geordie: producer.close(Duration.of(0))  will force close without sending all message . 
producer.close()  will gracefully close after sending all message .

the callback method will be invoke after  one of above  does

when we want to  send message  , we need to have
-  A Cluster has a leader partition 
-  Mockclient  prepare a fake response

So I add a cluster with one leader

g1geordie: if the message is not sent  , the thread will have a infinite loop

```
#Sender 
while(  ... .  this.accumulator.hasUndrained() || this.client.inFlightRequestCount() > 0  ...)
```

chia7712: Thanks for your explanation. ```oneLeaderCluster``` is used by this test case only. Could you make it be a local variable?

g1geordie: you are so nice !

chia7712 (commented):
Do other methods which need to wait for IO thread have similar issue? For example, calling ```sendOffsetsToTransaction``` in callback always causes the timeout as the IO thread is locked by the "waiting".

Is it worth adding similar check for them? A explicit exception about "this is disallowed in callback" should be more suitable than ```TimeoutException```.

chia7712:
@hachikuji FYI

ijuma:
Have we considered whether we could do something similar to `close` instead of throwing an exception? Not saying that's a good idea, just want to make sure we thought through the options.

g1geordie:
@ijuma hello 

The close method want to prevent self-join but there are no deadlock  . 

The flush method  in callback has a deadlock.
The deadlock is because  `flush` wait the `sender` send all message .
but messages are done after callback complete .
**flush in callback  , callback will never complete .**

`flush`'s semanteme is send message . 
so wait all message send I think it's necessary .

Messages are done  after callback I think is option ,
Or we can make async to `callback`  ,then messages will done no matter `callback` is complete.

chia7712:
Personally, the callback is a bit weird to me. The callback mechanism makes “internal” thread to handle user-defined code. The user code to handle response/error is vary different from serialization so it seems to me that callback causes some problems.

1. one thread’s callback can hurt other threads when we share a prouder to different threads.
2. bad performance due to “single” thread
3. can’t isolate the internal thread from user
4. callback has some cryptic limits (for instance, this issue and #9842)

I prefer the Future that user can handle it by user-defined thread. Also, it  does not have limits like callback and user code can be isolated from internal thread. For another, we should consider CompletabeFuture as we are already in JDK 8.

chia7712:
close as @frankvicky is working on it

PR #9708, opened 2020-12-07
KAFKA-9126: KIP-689: StreamJoined changelog configuration

lct45:
Adds `withLoggingEnabled` and `withLoggingDisabled` for `StreamJoined` to give `StreamJoined` the same flexibility as `Materialized`

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/StreamJoined.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/StreamJoined.java
@@ -338,7 +338,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
             name,
             storeName,
             false,
-            null
+            new HashMap<>()
         );
     }
 
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/StreamJoined.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/StreamJoined.java
@@ -20,6 +20,9 @@ package org.apache.kafka.streams.kstream;
 import org.apache.kafka.common.serialization.Serde;
 import org.apache.kafka.streams.state.WindowBytesStoreSupplier;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Class used to configure the name of the join processor, the repartition topic name,
  * state stores or state store names in  Stream-Stream join.
@@ -36,8 +39,8 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
     protected final WindowBytesStoreSupplier otherStoreSupplier;
     protected final String name;
     protected final String storeName;
-    protected boolean loggingEnabled = true;
-    protected boolean cachingEnabled = true;
+    protected boolean loggingEnabled;
+    protected Map<String, String> topicConfig;
 
     protected StreamJoined(final StreamJoined<K, V1, V2> streamJoined) {
         this(streamJoined.keySerde,
@@ -48,7 +51,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
             streamJoined.name,
             streamJoined.storeName,
             streamJoined.loggingEnabled,
-            streamJoined.cachingEnabled);
+            streamJoined.topicConfig);
     }
 
     private StreamJoined(final Serde<K> keySerde,
@@ -59,7 +62,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
                          final String name,
                          final String storeName,
                          final boolean loggingEnabled,
-                         final boolean cachingEnabled) {
+                         final Map<String, String> topicConfig) {
         this.keySerde = keySerde;
         this.valueSerde = valueSerde;
         this.otherValueSerde = otherValueSerde;
@@ -68,7 +71,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
         this.name = name;
         this.storeName = storeName;
         this.loggingEnabled = loggingEnabled;
-        this.cachingEnabled = cachingEnabled;
+        this.topicConfig = topicConfig;
     }
 
     /**
@@ -94,7 +97,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
             null,
             null,
             true,
-            true
+            new HashMap<>()
         );
     }
 
@@ -121,7 +124,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
             null,
             storeName,
             true,
-            true
+            new HashMap<>()
         );
     }
 
@@ -150,7 +153,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
             null,
             null,
             true,
-            true
+            new HashMap<>()
         );
     }
 
@@ -170,7 +173,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
             name,
             storeName,
             loggingEnabled,
-            cachingEnabled
+            topicConfig
         );
     }
 
@@ -194,7 +197,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
             name,
             storeName,
             loggingEnabled,
-            cachingEnabled
+            topicConfig
         );
     }
 
@@ -213,7 +216,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
             name,
             storeName,
             loggingEnabled,
-            cachingEnabled
+            topicConfig
         );
     }
 
@@ -232,7 +235,7 @@ public class StreamJoined<K, V1, V2> implements NamedOperation<StreamJoined<K, V
             name,
             storeName,
             loggingEnabled,
-            cachingEnabled
+            topicConfig
         );
     }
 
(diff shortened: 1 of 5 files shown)


streams/src/main/java/org/apache/kafka/streams/kstream/StreamJoined.java
```diff
@@ -267,7 +296,49 @@ private StreamJoined(final Serde<K> keySerde,
...
+        return new StreamJoined<>(
+            keySerde,
+            valueSerde,
+            otherValueSerde,
+            thisStoreSupplier,
+            otherStoreSupplier,
+            name,
+            storeName,
+            false,
+            null
```
vvcephei: I think you meant this here:

```suggestion
            new HashMap<>()
```

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KStreamKStreamJoinTest.java
```diff
@@ -152,6 +155,52 @@ private void shouldLogAndMeterOnSkippedRecordsWithNullValue(final String builtIn
...
+            right,
+            (value1, value2) -> value1 + value2,
+            joinWindows,
+            streamJoined);
+
+        final Topology topology = builder.build();
+        final InternalTopologyBuilder internalTopologyBuilder = TopologyWrapper.getInternalTopologyBuilder(topology);
+
+        assertThat(internalTopologyBuilder.stateStores().get("store-this-join-store").loggingEnabled(), equalTo(true));
+        assertThat(internalTopologyBuilder.stateStores().get("store-other-join-store").loggingEnabled(), equalTo(true));
```
vvcephei, line 212: It seems like these would pass by default. Maybe we should check set some log configs and then check that they got propagated ?

vvcephei (commented):
Thanks @lct45 , this looks good to me overall, just a couple of small problems. How does the test coverage look? It seems like the `null` should have resulted in an NPE at some point.

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KStreamKStreamJoinTest.java
lct45: Yeah it does pass by default. I experimented with passing the logs in but I wasn't able to find a good way to confirm that the logs I was passing were getting set somewhere. Do you know where they're exposed for me to check? Materialized doesn't seem to check this either

lct45:
@vvcephei Thanks for the review! Overall the test coverage looks pretty good, IIUC the log config is never touched when logging is disabled so I don't think there's a NPE that we're missing

lct45:
@cadonna for review

streams/src/main/java/org/apache/kafka/streams/kstream/StreamJoined.java
```diff
@@ -36,6 +39,8 @@
     protected final WindowBytesStoreSupplier otherStoreSupplier;
     protected final String name;
     protected final String storeName;
+    protected boolean loggingEnabled;
+    protected Map<String, String> topicConfig;
```
cadonna: I guess both can be declared as `final`, right?

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KStreamKStreamJoinTest.java
```diff
@@ -152,6 +155,52 @@ private void shouldLogAndMeterOnSkippedRecordsWithNullValue(final String builtIn
         }
     }
 
+    @Test
+    public void shouldDisableLoggingOnStreamJoined() {
+
+        final JoinWindows joinWindows = JoinWindows.of(ofMillis(100)).grace(Duration.ofMillis(50));
+        final StreamJoined<String, Integer, Integer> streamJoined = StreamJoined.with(Serdes.String(), Serdes.Integer(), Serdes.Integer()).withStoreName("store").withLoggingDisabled();
```
cadonna: Suggestion:
```suggestion
        final StreamJoined<String, Integer, Integer> streamJoined = StreamJoined
            .with(Serdes.String(), Serdes.Integer(), Serdes.Integer())
            .withStoreName("store")
            .withLoggingDisabled();
```

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KStreamKStreamJoinTest.java
```diff
@@ -152,6 +155,52 @@ private void shouldLogAndMeterOnSkippedRecordsWithNullValue(final String builtIn
...
+
+        final StreamsBuilder builder = new StreamsBuilder();
+        final KStream<String, Integer> left = builder.stream("left", Consumed.with(Serdes.String(), Serdes.Integer()));
+        final KStream<String, Integer> right = builder.stream("right", Consumed.with(Serdes.String(), Serdes.Integer()));
+
+        left.join(
+            right,
+            (value1, value2) -> value1 + value2,
+            joinWindows,
+            streamJoined);
```
cadonna:
```suggestion
        left.join(
            right,
            (value1, value2) -> value1 + value2,
            joinWindows,
            streamJoined
        );
```

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KStreamKStreamJoinTest.java
```diff
@@ -152,6 +155,52 @@ private void shouldLogAndMeterOnSkippedRecordsWithNullValue(final String builtIn
...
+
+        assertThat(internalTopologyBuilder.stateStores().get("store-this-join-store").loggingEnabled(), equalTo(false));
+        assertThat(internalTopologyBuilder.stateStores().get("store-other-join-store").loggingEnabled(), equalTo(false));
+    }
+
+    @Test
+    public void shouldEnableLoggingWithCustomConfigOnStreamJoined() {
+
+        final JoinWindows joinWindows = JoinWindows.of(ofMillis(100)).grace(Duration.ofMillis(50));
+        final StreamJoined<String, Integer, Integer> streamJoined = StreamJoined.with(Serdes.String(), Serdes.Integer(), Serdes.Integer()).withStoreName("store").withLoggingEnabled(Collections.emptyMap());
```
cadonna:
```suggestion
        final StreamJoined<String, Integer, Integer> streamJoined = StreamJoined
            .with(Serdes.String(), Serdes.Integer(), Serdes.Integer())
            .withStoreName("store")
            .withLoggingEnabled(Collections.emptyMap());
```

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KStreamKStreamJoinTest.java
```diff
@@ -152,6 +155,52 @@ private void shouldLogAndMeterOnSkippedRecordsWithNullValue(final String builtIn
...
+
+        final StreamsBuilder builder = new StreamsBuilder();
+        final KStream<String, Integer> left = builder.stream("left", Consumed.with(Serdes.String(), Serdes.Integer()));
+        final KStream<String, Integer> right = builder.stream("right", Consumed.with(Serdes.String(), Serdes.Integer()));
+
+        left.join(
+            right,
+            (value1, value2) -> value1 + value2,
+            joinWindows,
+            streamJoined);
```
cadonna:
```suggestion
        left.join(
            right,
            (value1, value2) -> value1 + value2,
            joinWindows,
            streamJoined
        );
```

streams/src/test/java/org/apache/kafka/streams/kstream/internals/KStreamKStreamJoinTest.java
cadonna: Maybe there is an easier way, but I found the following:

Set the config to:

```
        final StreamJoined<String, Integer, Integer> streamJoined = StreamJoined
            .with(Serdes.String(), Serdes.Integer(), Serdes.Integer())
            .withStoreName("store")
            .withLoggingEnabled(Collections.singletonMap("test", "property"));
```
and then check it:

```
        internalTopologyBuilder.buildSubtopology(0);

        assertThat(internalTopologyBuilder.stateStores().get("store-this-join-store").loggingEnabled(), equalTo(true));
        assertThat(internalTopologyBuilder.stateStores().get("store-other-join-store").loggingEnabled(), equalTo(true));
        assertThat(internalTopologyBuilder.topicGroups().get(0).stateChangelogTopics.size(), equalTo(2));
        for (final InternalTopicConfig config : internalTopologyBuilder.topicGroups().get(0).stateChangelogTopics.values()) {
            assertThat(
                config.getProperties(Collections.emptyMap(), 0).get("test"),
                equalTo("property")
            );
        }
```

Without

```
assertThat(internalTopologyBuilder.topicGroups().get(0).stateChangelogTopics.size(), equalTo(2));
```

the test would pass without checking the config if `buildSubtopology()` is not called because no changelog topics would be registered in the topology. So it basically checks that `buildSubtopology()` is called.

cadonna (commented):
Thanks for the PR, @lct45 !

We do not need to change anything in scala because you added only instance methods, right?

Here my feedback.

lct45:
Thanks for the review and test update @cadonna , and yes the scala should still be good to go

vvcephei:
Test failure was unrelated:

cadonna (approved):
LGTM!

mjsax:
@lct45 -- Seems we missed to update the docs, ie, `streams/upgrade_guide.html` for this KIP. Can you do a follow up PR for it?

lct45:
@mjsax Ahh yeah I'll submit a PR, thanks for catching that

mjsax:
Thanks @lct45!

For reference: https://github.com/apache/kafka/pull/9951

PR #9709, opened 2020-12-07
MINOR: Configure reconnect backoff in `BrokerToControllerChannelManager`

hachikuji:
We should configure a reconnect backoff for controller connections to prevent tight reconnect loops when the controller cannot be reached. I have borrowed the same configuration we use in `TransactionMarkerChannelManager`.


abbccdda (approved):

chia7712 (approved):

chia7712:
```kafka.api.TransactionsTest.testBumpTransactionalEpoch``` is traced by #9291 - ping for reviews BTW :)

mumrah (approved):
I believe the constructor calls here were copied from ControllerChannelManager. Do we want to add a reconnect backoff max there as well?

As a side note, we probably want a better way of setting up these classes that always go together (ChannelBuilder, Selector, NetworkClient).

hachikuji:
@mumrah Yeah, I noticed that too. I think `ControllerChannelManager` uses more of a synchronous backoff approach, so I left it as is.

PR #9710, opened 2020-12-07
MINOR: Remove redundant default parameter values in call to LogSegment.open

kowshik:
Few call sites in `Log.scala` were passing redundant default values in their calls to `LogSegment.open(...)`. In this PR, I've cleaned these up. For reference, here is a link to the definition of `LogSegment.open`: https://github.com/apache/kafka/blob/d2521855b8eb8a0dbb6f94cd9bb5093276fb7db2/core/src/main/scala/kafka/log/LogSegment.scala#L660-L661 

**Test plan:**
Relying on existing unit and integration tests.

### DIFF
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -770,9 +770,7 @@ class Log(@volatile private var _dir: File,
             baseOffset = 0,
             config,
             time = time,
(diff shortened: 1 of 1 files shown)


kowshik:
cc @junrao @ijuma for review

junrao (approved):

PR #9712, opened 2020-12-08
KAFKA-10822 Force some stdout from system tests for Travis

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-10822

downgrade_test.py/upgrade_test.py does upgrade/downgrade for each tests. the upgrade/downgrade tasks take 10+ mins in Travis env so we ought to print something in order to avoid timeout caused by Travis.

### DIFF
--- a/tests/kafkatest/tests/core/downgrade_test.py
+++ b/tests/kafkatest/tests/core/downgrade_test.py
@@ -120,11 +120,11 @@ class TestDowngrade(EndToEndTest):
         self.setup_services(kafka_version, compression_types, security_protocol, static_membership)
         self.await_startup()
 
-        self.logger.info("First pass bounce - rolling upgrade")
+        print("First pass bounce - rolling upgrade", flush=True) # Force some stdout for Travis
         self.upgrade_from(kafka_version)
         self.run_validation()
 
-        self.logger.info("Second pass bounce - rolling downgrade")
+        print("Second pass bounce - rolling downgrade", flush=True) # Force some stdout for Travis
         self.downgrade_to(kafka_version)
         self.run_validation()
         assert self.kafka.check_protocol_errors(self)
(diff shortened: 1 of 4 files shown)


abbccdda (commented):
Looks like the build is still in error?

chia7712:
> Looks like the build is still in error?

Thanks for your reviews. The main purpose of this PR is to avoid timeout from downgrade_test.py/upgrade_test.py. I browse the test results and the timeout does not happen on downgrade_test.py/upgrade_test.py again. After fixing the timeout issue, I will go back to dead with the true failure of system tests.

chia7712:
The timeout of "no output" is caused by only ```replica_scale_test.py```.  will go back here if #9797 get merged.

tests/kafkatest/tests/core/downgrade_test.py
```diff
@@ -120,11 +120,11 @@ def test_upgrade_and_downgrade(self, version, compression_types, security_protoc
         self.setup_services(kafka_version, compression_types, security_protocol, static_membership)
         self.await_startup()
 
-        self.logger.info("First pass bounce - rolling upgrade")
+        print("First pass bounce - rolling upgrade", flush=True) # Force some stdout for Travis
```
ijuma, line 123: This seems weird. If we want to force stdout, we should configure the logger to do so, right?

chia7712: I tried to configure ```ducktape```  to force more output but ```ducktape``` supports only ```--debug```. It is too noisy to me.

Another way is to make a patch to ```ducktape``` for INFO level support. WDYT?

ijuma: That sounds good to me.

chia7712: https://github.com/confluentinc/ducktape/issues/260

PR #9713, opened 2020-12-08
KAFKA-10825 ZK ISR manager

mumrah:
In an effort to consolidate the ISR write paths, this PR adapts the ZK ISR update into the new AlterIsrManager trait. This will allow us minimize divergence in the ISR update code in Partition.scala.

This also removes PartitionStateStore and moves the ISR propagation code to the new ZK ISR manager class


core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -46,13 +51,36 @@ trait AlterIsrManager {
   def clearPending(topicPartition: TopicPartition): Unit
 }
 
-case class AlterIsrItem(topicPartition: TopicPartition, leaderAndIsr: LeaderAndIsr, callback: Either[Errors, LeaderAndIsr] => Unit)
+case class AlterIsrItem(topicPartition: TopicPartition,
+                        leaderAndIsr: LeaderAndIsr,
+                        callback: Either[Errors, LeaderAndIsr] => Unit,
+                        controllerEpoch: Int = -1) // controllerEpoch needed for Zk impl
```
hachikuji: We don't have to do it here, but I think the controller epoch is no longer needed. This was previously used as a sort of poor man's controller fencing, but now the controller has a stronger mechanism relying on conditional zk updates.

By the way, do we need the default value? Especially the fact that it is a sentinel seems dangerous.

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -51,40 +51,8 @@ trait IsrChangeListener {
...
-      leaderAndIsr, controllerEpoch)
-
-    if (updateSucceeded) {
-      Some(newVersion)
-    } else {
-      None
-    }
-  }
+trait TopicConfigProvider {
+  def get(): Properties
```
hachikuji: nit: I wonder if it might be better to use a verb like `fetch` which suggests some overhead involved

core/src/main/scala/kafka/server/ZkIsrManager.scala
```diff
@@ -0,0 +1,116 @@
...
+      period = isrChangeNotificationConfig.checkIntervalMs, unit = TimeUnit.MILLISECONDS)
+  }
+
+  override def clearPending(topicPartition: TopicPartition): Unit = {
+    // Since we always immediately process ZK updates and never actually enqueue anything, there is nothing to
+    // clear here so this is a no-op. Even if there are changes that have not been propagated, the write to ZK
+    // has already happened, so we may as well send the notification to the controller.
+  }
+
+  override def enqueue(alterIsrItem: AlterIsrItem): Boolean = {
```
hachikuji: The name "enqueue" suggests an asynchronous change. Wonder if there is another name we could use. Perhaps "submit" suggests less about the implementation?

core/src/main/scala/kafka/server/ZkIsrManager.scala
```diff
@@ -0,0 +1,116 @@
...
+      alterIsrItem.leaderAndIsr, alterIsrItem.controllerEpoch)
+
+    if (updateSucceeded) {
+      // Track which partitions need to be propagated to the controller
+      isrChangeSet synchronized {
+        isrChangeSet += alterIsrItem.topicPartition
+        lastIsrChangeMs.set(time.milliseconds())
+      }
+
+      // We actually need to apply the callback in another thread since Partition#sendAlterIsrRequest will write
```
hachikuji: Hmm.. It feels a tad brittle to rely on locking assumptions like this. Perhaps we could let `enqueue` return a different response in the case that the change was applied immediately? For example, we could return `Option[Int]` to indicate that new version if the change was applied. Note that we can probably discard the boolean return type and let `enqueue` raise an illegal state exception directly since that is what `Partition` is already doing.

core/src/main/scala/kafka/server/AlterIsrManager.scala
mumrah: That `-1` is left over from test refactoring. I'll remove

core/src/main/scala/kafka/server/ZkIsrManager.scala
mumrah: Ok, so in this case, we would need to call the callback from `sendAlterIsrRequest` for the ZK case. I think that would probably be ok. I agree that relying on the separate thread and the lock is a little wonky.

Something like this?
```scala
try {
  alterIsrManager.enqueue(alterIsrItem) match {
    case Some(newVersion) => 
      isrState = proposedIsrState
      handleAlterIsrResponse(proposedState.withZkVersion(newVersion))
    case None =>
      isrState = proposedIsrState
} catch {
   e: Throwable => 
    isrChangeListener.markFailed()
    throw e
}
```

mumrah: One issue is the invalid version case for ZK -- if we return None there then partition will think it's a call to DefaultAlterIsrManager... I guess we could throw from ZkIsrManager if the update fails. WDYT?

mumrah: I went with a slightly different approach. See the latest commits

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -1374,47 +1314,28 @@ class Partition(val topicPartition: TopicPartition,
...
-      isrChangeListener.markFailed()
-      throw new IllegalStateException(s"Failed to enqueue `AlterIsr` request with state " +
-        s"$newLeaderAndIsr for partition $topicPartition")
-    }
-
-    isrState = proposedIsrState
-    debug(s"Sent `AlterIsr` request to change state to $newLeaderAndIsr after transition to $proposedIsrState")
+    alterIsrManager.submit(alterIsrItem, (wasSubmitted: Boolean) => {
+      if (wasSubmitted) {
+        isrState = proposedIsrState
```
hachikuji: Could we solve the problem by moving this above the call to `submit`? Then the callback logic should work even if the change is made synchronously. In the case that the request fails to be submitted, then we can reset to the previous state. What do you think?

mumrah: Yea, this seems simpler than the callback thing. I've made this change in the latest commit

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -1374,46 +1314,29 @@ class Partition(val topicPartition: TopicPartition,
...
+    if (!alterIsrManager.submit(alterIsrItem)) {
+      // If the ISR manager did not accept our update, we need to revert back to previous state
+      isrState = oldState
       isrChangeListener.markFailed()
       throw new IllegalStateException(s"Failed to enqueue `AlterIsr` request with state " +
         s"$newLeaderAndIsr for partition $topicPartition")
     }
 
-    isrState = proposedIsrState
     debug(s"Sent `AlterIsr` request to change state to $newLeaderAndIsr after transition to $proposedIsrState")
```
hachikuji: Should we generalize this log message as well? Also the couple `IllegalStateException` messages above.

core/src/main/scala/kafka/server/ZkIsrManager.scala
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  override def clearPending(topicPartition: TopicPartition): Unit = {
+    // Since we always immediately process ZK updates and never actually enqueue anything, there is nothing to
+    // clear here so this is a no-op. Even if there are changes that have not been propagated, the write to ZK
+    // has already happened, so we may as well send the notification to the controller.
+  }
+
+  override def submit(alterIsrItem: AlterIsrItem): Boolean = {
+    debug(s"Writing new ISR " + alterIsrItem.leaderAndIsr.isr + " to ZooKeeper with version " +
```
hachikuji: nit: rewrite using `$` substitutions?

hachikuji (approved):
LGTM. Really nice cleanup! Left a couple small comments which don't need re-review.

PR #9714, opened 2020-12-08
MINOR: Remove connection id from Send and consolidate request/message utils

ijuma:
Connection id is now only present in `NetworkSend`, which is now
the class used by `Selector`/`NetworkClient`/`KafkaChannel` (which
works well since `NetworkReceive` is the class used for
received data).

The previous `NetworkSend` was also responsible for adding a size
prefix. This logic is already present in `SendBuilder`, but for the
minority of cases where `SendBuilder` is not used (including
a number of tests), we now have `ByteBufferSend.sizePrefixed()`.

With regards to the request/message utilities:
* Renamed `toByteBuffer`/`toBytes` in `MessageUtil` to
`toVersionPrefixedByteBuffer`/`toVersionPrefixedBytes` for clarity.
* Introduced new `MessageUtil.toByteBuffer` that does not include
the version as the prefix.
* Renamed `serializeBody` in `AbstractRequest/Response` to
`serialize` for symmetry with `parse`.
* Introduced `RequestTestUtils` and moved relevant methods from
`TestUtils`.
* Moved `serializeWithHeader` methods that were only used in
tests to `RequestTestUtils`.
* Deleted `MessageTestUtil`.

Finally, a couple of changes to simplify coding patterns:
* Added `flip()` and `buffer()` to `ByteBufferAccessor`.
* Added `MessageSizeAccumulator.sizeExcludingZeroCopy`.
* Used lambdas instead of `TestCondition`.
* Used `Arrays.copyOf` instead of `System.arraycopy` in `MessageUtil`.


clients/src/main/java/org/apache/kafka/common/requests/RequestUtils.java
```diff
@@ -76,36 +75,22 @@ public static boolean hasTransactionalRecords(ProduceRequest request) {
...
         Message apiMessage,
         short apiVersion
     ) {
-        ObjectSerializationCache serializationCache = new ObjectSerializationCache();
-        MessageSizeAccumulator messageSize = RequestUtils.size(serializationCache, header, headerVersion, apiMessage, apiVersion);
+        ObjectSerializationCache cache = new ObjectSerializationCache();
+
+        int headerSize = header.size(cache, headerVersion);
+        int messageSize = apiMessage.size(cache, apiVersion);
+        ByteBufferAccessor writable = new ByteBufferAccessor(ByteBuffer.allocate(4 + headerSize + messageSize));
```
chia7712: Why it need 4 byte if it exclude size prefix?

clients/src/main/java/org/apache/kafka/common/record/MultiRecordsSend.java
```diff
@@ -46,8 +45,7 @@
      * Construct a MultiRecordsSend for the given destination from a queue of Send objects. The queue will be
```
chia7712: This comment need to be updated.

clients/src/main/java/org/apache/kafka/common/network/Send.java
```diff
@@ -24,11 +24,6 @@
  */
 public interface Send {
```
chia7712, line 25: please remove "destination" from the docs

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -964,7 +963,7 @@ private[kafka] class Processor(val id: Int,
     // removed from the Selector after discarding any pending staged receives.
     // `openOrClosingChannel` can be None if the selector closed the connection because it was idle for too long
     if (openOrClosingChannel(connectionId).isDefined) {
-      selector.send(responseSend)
+      selector.send(new NetworkSend(response.request.context.connectionId, responseSend))
```
chia7712: line#955 already has ```connectionId```.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1134,11 +1133,11 @@ private[kafka] class Processor(val id: Int,
         // Try unmuting the channel. If there was no quota violation and the channel has not been throttled,
         // it will be unmuted immediately. If the channel has been throttled, it will unmuted only if the throttling
         // delay has already passed by now.
-        handleChannelMuteEvent(send.destination, ChannelMuteEvent.RESPONSE_SENT)
-        tryUnmuteChannel(send.destination)
+        handleChannelMuteEvent(send.destinationId(), ChannelMuteEvent.RESPONSE_SENT)
+        tryUnmuteChannel(send.destinationId())
```
chia7712: How about keeping "parameterless"? fewer changes and more clear.

clients/src/test/java/org/apache/kafka/common/requests/RequestTestUtils.java
```diff
@@ -0,0 +1,185 @@
...
+                                                      final Map<String, Integer> topicPartitionCounts) {
+        return metadataUpdateWith(clusterId, numNodes, topicErrors,
+                topicPartitionCounts, tp -> null, MetadataResponse.PartitionMetadata::new, ApiKeys.METADATA.latestVersion());
+    }
+
+    public static MetadataResponse metadataUpdateWith(final String clusterId,
+                                                      final int numNodes,
+                                                      final Map<String, Errors> topicErrors,
+                                                      final Map<String, Integer> topicPartitionCounts,
+                                                      final short responseVersion) {
```
chia7712, line 130: unused argument

clients/src/test/java/org/apache/kafka/common/requests/RequestTestUtils.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+
+public class RequestTestUtils {
+
+    public static ByteBuffer serializeRequestHeader(RequestHeader header) {
+        ObjectSerializationCache serializationCache = new ObjectSerializationCache();
+        ByteBuffer buffer = ByteBuffer.allocate(header.size(serializationCache));
+        header.write(buffer, serializationCache);
```
chia7712, line 42: ```RequestHeader#write``` has only 2 usages and both of them are in test scope. It should be fine to remove ```RequestHeader#write``` from production.

chia7712 (commented):
@ijuma Thanks for this great cleanup. Please take a look at following minor comments. Thanks!

clients/src/test/java/org/apache/kafka/common/requests/RequestTestUtils.java
ijuma: Looks like a test bug from the previous PR. Will fix so that it's used.

clients/src/test/java/org/apache/kafka/common/requests/RequestTestUtils.java
ijuma: It's about conceptual integrity for the class. It should provide a mechanism for serialization that doesn't require reaching into its internal structures.

core/src/main/scala/kafka/network/SocketServer.scala
ijuma: Looks like the IDE added this when I renamed it or something. Fixed.

clients/src/main/java/org/apache/kafka/common/requests/RequestUtils.java
ijuma: Great catch. This was a copy and paste bug and we had no tests verifying this. I have fixed it and added a test.

ijuma:
@chia7712 Thanks for the review, I'm particularly glad you caught the 4 wasted bytes. :) I pushed updates and left a comment in the one case I haven't addressed.

clients/src/test/java/org/apache/kafka/common/requests/RequestTestUtils.java
chia7712: fair enough

clients/src/main/java/org/apache/kafka/common/network/SizeDelimitedSend.java
```diff
@@ -14,18 +14,22 @@
...
-        ObjectSerializationCache cache = new ObjectSerializationCache();
-        int size = message.size(cache, version);
-        ByteBuffer bytes = ByteBuffer.allocate(size);
-        message.write(new ByteBufferAccessor(bytes), cache, version);
-        bytes.rewind();
-        return bytes;
+/**
+ * A size delimited Send that consists of a 4 byte network-ordered size N followed by N bytes of content.
+ */
+public class SizeDelimitedSend extends ByteBufferSend {
```
chia7712: How about making ```SizeDelimitedSend``` be a static method in ```ByteBufferSend```? For example:
```java
    public static Send withSizeDelimited(ByteBuffer buffer) {
        ByteBuffer sizeBuffer = ByteBuffer.allocate(4);
        sizeBuffer.putInt(0, buffer.remaining());
        return new ByteBufferSend(sizeBuffer, buffer);
    }
```

chia7712 (approved):
@ijuma overall +1 and thanks for this great refactor. please take a look at last comment :)

clients/src/main/java/org/apache/kafka/common/network/SizeDelimitedSend.java
ijuma: I had the same thought and initially did that, but it's an issue for many tests since they verify the properties of the ByteBuffer and it is not accessible if we do this.

chia7712: > but it's an issue for many tests since they verify the properties of the ByteBuffer and it is not accessible if we do this.

Can it be resolved if we return ```ByteBuffer``` rather than ```Send```?

ijuma: It has to be a `Send` too. I looked at this a bit closer and we can just return `ByteBufferSend` from the static factory method. The thing I attempted before was to combine at the `NetworkSend` level and that doesn't work since `NetworkSend` can work with `Send`s that are not backed by a `ByteBuffer`. But it actually works fine at `ByteBufferSend` level. Thanks for the suggestion, please check the update.

chia7712:
+ 1 to latest commit

hachikuji (commented):
Nice cleanup! I didn't look too carefully, but it seems like a step in the right direction. Pulling destination out of Send is something I've wanted to do for some time.

ijuma:
Tests passed, merging to trunk.

PR #9715, opened 2020-12-08
Upstream ApisUtils from kip-500

mumrah:
In the [KIP-500 development branch](https://github.com/confluentinc/kafka/tree/kip-500), we have a separate ControllerApis that shares a lot of functionality with KafkaApis. We introduced a utility class ApisUtils to pull out the common code. Some things were moved to RequestChannel as well.

We'd like to upstream this work now so we don't continue to diverge (since KafkaApis is a frequently modified class). There should be no logical changes in this PR, only shuffling code around.


mumrah:
@abbccdda can you take a look at the forwarding parts?

core/src/main/scala/kafka/server/ApisUtils.scala
```diff
@@ -0,0 +1,184 @@
...
+    val throttleTimeMs = response.throttleTimeMs()
+    quotas.request.throttle(request, throttleTimeMs, requestChannel.sendResponse)
+    requestChannel.sendResponse(request, Some(response), None)
+  }
+
+  // Throttle the channel if the request quota is enabled but has been violated. Regardless of throttling, send the
+  // response immediately.
+  def sendResponseMaybeThrottle(request: RequestChannel.Request,
+                                createResponse: Int => AbstractResponse,
+                                onComplete: Option[Send => Unit] = None): Unit = {
```
chia7712: ```onComplete``` is unused.

core/src/main/scala/kafka/server/ApisUtils.scala
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  /**
+   * Throttle the channel if the controller mutations quota or the request quota have been violated.
+   * Regardless of throttling, send the response immediately.
+   */
+  def sendResponseMaybeThrottleWithControllerQuota(controllerMutationQuota: ControllerMutationQuota,
+                                                   request: RequestChannel.Request,
+                                                   createResponse: Int => AbstractResponse,
+                                                   onComplete: Option[Send => Unit]): Unit = {
```
chia7712: ditto

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -98,6 +97,7 @@ import scala.annotation.nowarn
  * Logic to handle the various Kafka requests
  */
 class KafkaApis(val requestChannel: RequestChannel,
+                val apisUtils: ApisUtils,
```
chia7712: Instantiating a "utils" object is a bit weird to me. it seems to me we can make ApisUtils be a ```trait``` with self-type (this: KafkaApis) if the main purpose of this PR is to reduce the size of ```KafkaApis```. The benefit of using self-type is that 
1. we don't need to instantiate a "utils"
1. we can move some code from KafkaApis to ApisUtils
1. we don't need to change "autxxx" to "apisUtils.autxxx"

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -432,6 +432,43 @@ class RequestChannel(val queueSize: Int,
...
+          onComplete
+        )
+      case None =>
+        new RequestChannel.NoOpResponse(request)
+    }
+
+    sendResponse(response)
+  }
+
+  def sendErrorOrCloseConnection(request: RequestChannel.Request, error: Throwable, throttleMs: Int): Unit = {
```
chia7712: There is a another "sendErrorOrCloseConnection" in ```ApisUtils```

chia7712 (commented):
@mumrah Thanks for this refactor. Some minor comments are left. Please take a look.

core/src/main/scala/kafka/server/KafkaApis.scala
mumrah: Thanks, @chia7712, I like this suggestion. I'll see if I can figure this out (I wasn't too happy with the utility class needing to be instantiated either)

core/src/main/scala/kafka/server/KafkaRequestHandler.scala
```diff
@@ -34,6 +34,10 @@ trait ApiRequestHandler {
   def handle(request: RequestChannel.Request): Unit
 }
 
+trait BaseApis extends ApiRequestHandler {
```
mumrah: Added this trait to avoid polluting TestRaftRequestHandler with the ApisUtils dependencies

mumrah:
@chia7712 thanks for the feedback. I've incorporated your suggestions, can you take another look?

core/src/main/scala/kafka/server/KafkaRequestHandler.scala
chia7712: not sure whether ```BaseApis```  is required. ```KafkaApis``` can still extend ```ApiRequestHandler``` as you have added required variables to ```ApisUtils```
```scala
  val requestChannel: RequestChannel
  val quotas: QuotaManagers
  val time: Time
  val authorizer: Option[Authorizer]
```

core/src/main/scala/kafka/network/RequestChannel.scala
chia7712: just curious. Why this method is located in ```RequestChannel``` rather than ```ApiUtils```

core/src/main/scala/kafka/server/KafkaServer.scala
```diff
@@ -369,15 +369,17 @@ class KafkaServer(val config: KafkaConfig, time: Time = Time.SYSTEM, threadNameP
             KafkaServer.MIN_INCREMENTAL_FETCH_SESSION_EVICTION_MS))
 
         /* start processing requests */
-        dataPlaneRequestProcessor = new KafkaApis(socketServer.dataPlaneRequestChannel, replicaManager, adminManager, groupCoordinator, transactionCoordinator,
+        dataPlaneRequestProcessor = new KafkaApis(socketServer.dataPlaneRequestChannel,
```
chia7712: Is this change still valid?

core/src/main/scala/kafka/server/KafkaServer.scala
```diff
@@ -369,15 +369,17 @@ class KafkaServer(val config: KafkaConfig, time: Time = Time.SYSTEM, threadNameP
...
+          replicaManager, adminManager, groupCoordinator, transactionCoordinator,
           kafkaController, forwardingManager, zkClient, config.brokerId, config, metadataCache, metrics, authorizer, quotaManagers,
           fetchManager, brokerTopicStats, clusterId, time, tokenManager, brokerFeatures, featureCache)
 
         dataPlaneRequestHandlerPool = new KafkaRequestHandlerPool(config.brokerId, socketServer.dataPlaneRequestChannel, dataPlaneRequestProcessor, time,
           config.numIoThreads, s"${SocketServer.DataPlaneMetricPrefix}RequestHandlerAvgIdlePercent", SocketServer.DataPlaneThreadPrefix)
 
         socketServer.controlPlaneRequestChannelOpt.foreach { controlPlaneRequestChannel =>
-          controlPlaneRequestProcessor = new KafkaApis(controlPlaneRequestChannel, replicaManager, adminManager, groupCoordinator, transactionCoordinator,
+          controlPlaneRequestProcessor = new KafkaApis(controlPlaneRequestChannel,
```
chia7712: ditto

core/src/main/scala/kafka/server/KafkaApis.scala
ijuma: Is this an improvement? I am not sure. A couple of things:
1. The `Utils` suffix looks odd both in a trait as well as a class that is instantiated.
2. Using delegation allows for better separation of concerns than inheritance.

core/src/main/scala/kafka/server/KafkaRequestHandler.scala
chia7712: BTW, my thought was 

```scala
trait ApisUtils extends Logging {
  this: KafkaApis =>
```

chia7712 (commented):
@mumrah Thanks for updating. I'm thinking about how we correctly reduce the size of ```KafkaApis```. Yes, it and its test (```KafkaApisTest```) are too huge. They should be refactor and it would be better to have an overall plan instead of just moving some methods out of ```KafkaApis```.

core/src/main/scala/kafka/server/KafkaRequestHandler.scala
ijuma: Self types like this are an anti pattern, in my opinion. They were used years ago as a DI replacement (the cake pattern), but people have moved on since they tend to be hard to understand.

ijuma:
Note that the goal here is not to reduce `KafkaApis` footprint, it is to allow reuse from `ControllerApis`. The latter is required for the KIP-500 effort. I think it makes sense to go with a simple change here that extracts commonly used methods. If we want to discuss breaking up `KafkaApis`, that is a fair discussion, but can happen separately. Does that make sense @chia7712?

core/src/main/scala/kafka/server/KafkaApis.scala
mumrah: 1. Agreed. I don't really like the name, but can't think of a better one.

2. Originally, I had hoped to remove some dependencies from KafkaApis when pulling out these methods, but it turns out they are all needed for other things as well. In the end it just adds a new dependency (which is fine). Seeing how I had to mess around with the type hierarchy to not break or complicate TestRaftRequestHandler, I'm leaning towards the delegation approach we had before.

chia7712:
> I think it makes sense to go with a simple change here that extracts commonly used methods. If we want to discuss breaking up KafkaApis, that is a fair discussion, but can happen separately. Does that make sense

makes sense to me :)

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -412,8 +412,8 @@ class RequestChannel(val queueSize: Int,
   }
 
   def sendResponse(request: RequestChannel.Request,
-                   responseOpt: Option[AbstractResponse],
-                   onComplete: Option[Send => Unit]): Unit = {
+                           responseOpt: Option[AbstractResponse],
+                           onComplete: Option[Send => Unit]): Unit = {
```
ijuma: Is this intentional?

core/src/main/scala/kafka/server/ApisUtils.scala
```diff
@@ -109,7 +106,7 @@ trait ApisUtils extends Logging {
   // Throttle the channel if the request quota is enabled but has been violated. Regardless of throttling, send the
   // response immediately.
   def sendResponseMaybeThrottle(request: RequestChannel.Request,
-                                createResponse: Int => AbstractResponse): Unit = {
+                                        createResponse: Int => AbstractResponse): Unit = {
```
ijuma: Is this intentional?

core/src/main/scala/kafka/network/RequestChannel.scala
mumrah: I suppose I put it here since it doesn't need any of the other objects that are provided by ApisUtils, it just needs the RequestChannel.

core/src/main/scala/kafka/server/ApisUtils.scala
```diff
@@ -0,0 +1,204 @@
...
+import org.apache.kafka.common.utils.{LogContext, Time, Utils}
+import org.apache.kafka.server.authorizer.{Action, AuthorizationResult, Authorizer}
+
+import scala.jdk.CollectionConverters._
+
+/**
+ * Helper methods and helper class factories for request handlers. Provides common functionality around throttling,
+ * authorizations, and error handling
+ */
+object ApisUtils {
```
hachikuji: Might be just me, but the plural `Apis` reads awkwardly to me. How about `RequestHandlerUtils`?

hachikuji (commented):
It would be nice if we could make use of the new util class in `TestRaftRequestHandler`.

core/src/main/scala/kafka/server/ApisUtils.scala
mumrah: It is pretty awkward, but `ApiUtils` was taken. `RequestHandlerUtils` sounds good.

core/src/main/scala/kafka/server/RequestHandlerUtils.scala
```diff
@@ -0,0 +1,203 @@
...
+    updatedFollowers.foreach { partition =>
+      if (partition.topic == Topic.GROUP_METADATA_TOPIC_NAME)
+        groupCoordinator.onResignation(partition.partitionId)
+      else if (partition.topic == Topic.TRANSACTION_STATE_TOPIC_NAME)
+        txnCoordinator.onResignation(partition.partitionId, Some(partition.getLeaderEpoch))
+    }
+  }
+}
+
+class AuthHelper(val requestChannel: RequestChannel,
```
ijuma: I would move this to its own file. It seems unrelated to the other things in this class.

core/src/main/scala/kafka/server/RequestHandlerUtils.scala
```diff
@@ -0,0 +1,203 @@
...
+          .filter(_._1 == AuthorizationResult.ALLOWED)
+          .map(_._2).toSet
+      case None =>
+        supportedOps.toSet
+    }
+    Utils.to32BitField(authorizedOps.map(operation => operation.code.asInstanceOf[JByte]).asJava)
+  }
+}
+
+class ChannelHelper(val requestChannel: RequestChannel,
```
ijuma: Should this be `RequestHandlerHelper` then? Maybe we should ditch the `Utils` name and only have `RequestHandlerHelper` with the static methods in the companion.

core/src/main/scala/kafka/server/RequestHandlerHelper.scala
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+}
+
+
+
+class RequestHandlerHelper(val requestChannel: RequestChannel,
+                           val quotas: QuotaManagers,
+                           val time: Time,
+                           val logPrefix: String) extends Logging {
```
jsancio: Adding `val` to the constructor arguments makes them `public` members. Glancing at the code it doesn't look like this is needed.

core/src/main/scala/kafka/server/AuthHelper.scala
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.kafka.common.resource.ResourceType.CLUSTER
+import org.apache.kafka.common.resource.{PatternType, Resource, ResourcePattern, ResourceType}
+import org.apache.kafka.common.utils.Utils
+import org.apache.kafka.server.authorizer.{Action, AuthorizationResult, Authorizer}
+
+import scala.jdk.CollectionConverters._
+
+
+class AuthHelper(val requestChannel: RequestChannel,
+                 val authorizer: Option[Authorizer]) {
```
jsancio: Adding `val` to the constructor arguments makes the `public` members. Glancing at the code it doesn't look like this is needed.

jsancio (commented):
Thanks for the PR. Minor comments.

ijuma (approved):
LGTM, thanks.

ijuma (changes requested):
Actually, tests are failing and they need to be fixed before we can merge this.

mumrah:
The test failures were due to unhappy mocks in KafkaApisTest. It looks rather tedious to fix the mocks so I ended up moving two of the methods I had put into RequestChannel back into the helper class. They are now passing for me locally.

ijuma (approved):
LGTM

PR #9716, opened 2020-12-09
KAFKA-10826; Ensure raft io thread respects linger timeout

hachikuji:
When there are no pending operations, the raft IO thread can block indefinitely waiting for a network event. We rely on asynchronous wakeups in order to break the blocking wait in order to respond to a scheduled append. The current logic already does this, but only for the case when the linger time has been completed during the call to `scheduleAppend`. It is possible instead that after making one call to `scheduleAppend` to start the linger timer, the application does not do any additional appends. In this case, we still need the IO thread to wakeup when the linger timer expires. This patch fixes the problem by ensuring that the IO thread gets woken up after the first append which begins the linger timer.

### DIFF
--- a/raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
+++ b/raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
@@ -280,8 +280,8 @@ public class BatchAccumulator<T> implements Closeable {
     }
 
     /**
-     * Get the number of batches including the one that is currently being
-     * written to (if it exists).
+     * Get the number of completed batches which are ready to be drained.
+     * This does not include the batch that is currently being filled.
      */
     public int numCompletedBatches() {
         return completed.size();
--- a/raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
+++ b/raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
@@ -141,6 +141,7 @@ public class BatchAccumulator<T> implements Closeable {
             startNewBatch();
         } else if (!currentBatch.hasRoomFor(batchSize)) {
             completeCurrentBatch();
+            startNewBatch();
         }
         return currentBatch;
     }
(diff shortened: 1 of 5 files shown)


raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -271,22 +272,19 @@ public int epoch() {
         }
     }
 
+    public boolean isEmpty() {
+        // The linger timer begins running when we have pending batches.
+        // We use this to infer when the accumulator is empty to avoid the
```
guozhangwang, line 277: Nice one.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -271,22 +272,19 @@ public int epoch() {
...
-            int count = completed.size();
-            if (currentBatch != null) {
-                return count + 1;
-            } else {
-                return count;
-            }
-        } finally {
-            appendLock.unlock();
-        }
+    public int numCompletedBatches() {
```
guozhangwang, line 286: Does the javadoc above need updates?

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -477,6 +477,74 @@ public void testAccumulatorClearedAfterBecomingUnattached() throws Exception {
...
+        assertEquals(1L, context.log.endOffset().offset);
+
+        int epoch = context.currentEpoch();
+        assertEquals(1L, context.client.scheduleAppend(epoch, singletonList("a")));
+        assertTrue(context.channel.wakeupRequested());
+
+        context.client.poll();
+        assertEquals(OptionalLong.of(lingerMs), context.channel.lastReceiveTimeout());
+
+        context.time.sleep(25);
```
guozhangwang: nit: maybe we split 50 to 20/30 to avoid some reading difficulty? :)

guozhangwang (approved):
LGTM! Please feel free to merge afterwards.

PR #9718, opened 2020-12-09
KAFKA-10832: Fix Log to use the correct ProducerStateManager instance when updating producers

kowshik:
I have fixed what looked like a potential bug to me.

**Bug:**
The bug is that from within `Log.updateProducers(…)`, the code operates on the `producerStateManager` attribute of the `Log` instance instead of operating on an input parameter. Please see [this](https://github.com/apache/kafka/blob/1d84f543678c4c08800bc3ea18c04a9db8adf7e4/core/src/main/scala/kafka/log/Log.scala#L1464) LOC where it calls `producerStateManager.prepareUpdate` thus accessing the attribute from the `Log` object (see [this](https://github.com/apache/kafka/blob/1d84f543678c4c08800bc3ea18c04a9db8adf7e4/core/src/main/scala/kafka/log/Log.scala#L251)). This looks unusual particularly for `Log.loadProducersFromLog(...)` [path](https://github.com/apache/kafka/blob/1d84f543678c4c08800bc3ea18c04a9db8adf7e4/core/src/main/scala/kafka/log/Log.scala#L956). Here I believe we should be using the instance passed to the method, rather than the attribute from the `Log` instance. I have fixed the same in this PR.

I'm not sure (yet) if this bug has any drastic consequences (probably not), but it is still worthwhile making things consistent.

**Fix:**
The fix is to explicitly pass the `ProducerStateManager` into `Log.updateProducers` and move the following methods: `Log.loadProducersFromLog` and `Log.updateProducers` into the `Log` companion object.

**Tests:**
Rely on existing tests.

### DIFF
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -953,7 +953,9 @@ class Log(@volatile private var _dir: File,
     val completedTxns = ListBuffer.empty[CompletedTxn]
     records.batches.forEach { batch =>
       if (batch.hasProducerId) {
-        val maybeCompletedTxn = updateProducers(batch,
+        val maybeCompletedTxn = updateProducers(
+          producerStateManager,
+          batch,
           loadedProducers,
           firstOffsetMetadata = None,
           origin = AppendOrigin.Replication)
@@ -1349,7 +1351,7 @@ class Log(@volatile private var _dir: File,
         else
           None
 
-        val maybeCompletedTxn = updateProducers(batch, updatedProducers, firstOffsetMetadata, origin)
+        val maybeCompletedTxn = updateProducers(producerStateManager, batch, updatedProducers, firstOffsetMetadata, origin)
         maybeCompletedTxn.foreach(completedTxns += _)
       }
 
@@ -1456,7 +1458,8 @@ class Log(@volatile private var _dir: File,
       RecordConversionStats.EMPTY, sourceCodec, targetCodec, shallowMessageCount, validBytesCount, monotonic, lastOffsetOfFirstBatch)
   }
 
-  private def updateProducers(batch: RecordBatch,
+  private def updateProducers(producerStateManager: ProducerStateManager,
+                              batch: RecordBatch,
                               producers: mutable.Map[Long, ProducerAppendInfo],
                               firstOffsetMetadata: Option[LogOffsetMetadata],
                               origin: AppendOrigin): Option[CompletedTxn] = {
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -864,11 +864,83 @@ class Log(@volatile private var _dir: File,
     recoveryPoint
   }
 
+  // Rebuild producer state until lastOffset. This method may be called from the recovery code path, and thus must be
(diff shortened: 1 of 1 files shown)


kowshik:
cc @dhruvilshah3 @junrao @gardnervickers for review

ijuma:
Good catch. This looks a bit error prone. Should we be moving some of these methods to the companion object or something?

kowshik:
@ijuma Good idea. I can move `Log.rebuildProducerState` into the `Log` companion object, and consequently the following methods will move too: `Log.loadProducersFromLog` and `Log.updateProducers`.

kowshik:
@ijuma Thanks for the review! I've addressed your comments in 6d348bcc12f9e8b2ddf0552c089945b4eda67538.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2670,6 +2573,112 @@ object Log {
   private def isLogFile(file: File): Boolean =
     file.getPath.endsWith(LogFileSuffix)
 
+  // Rebuild producer state until lastOffset. This method may be called from the recovery code path, and thus must be
+  // free of all side-effects, i.e. it must not update any log-specific state.
+  private def rebuildProducerState(log: Log,
```
junrao: This seems too large as a companion method. I am also not sure that we should pass along Log to this method since it exposes all internal states in Log, which make it harder to track usage.

junrao (commented):
@kowshik : Thanks for the PR. Nice find. Just one comment below.

core/src/main/scala/kafka/log/Log.scala
kowshik: That is true. Do you feel the original approach in f0466b2420dc500112b4839678564996c0677d2e was better, where, we didn't move methods into a companion object? I'd be happy to revert to that.

junrao: We can probably keep the two other two simpler methods (without Log as the input) as the companion methods.

kowshik:
Thanks for the review, @junrao! I've addressed the comments in 5d798b7d884bd11e43884318ae3fc8d12e635c40.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2670,6 +2645,34 @@ object Log {
   private def isLogFile(file: File): Boolean =
     file.getPath.endsWith(LogFileSuffix)
 
+  private def loadProducersFromLog(producerStateManager: ProducerStateManager, records: Records): Unit = {
```
junrao: Perhaps this could be name loadProducersFromRecords?

junrao (commented):
@kowshik : Thanks for the updated PR. Just a minor comment below. Also, is the test failure related to this PR?

core/src/main/scala/kafka/log/Log.scala
kowshik: Done

kowshik:
Thanks for the review @junrao ! I've addressed the comment in ec00e9f. The JDK 15 test failure in `org.apache.kafka.streams.integration.PurgeRepartitionTopicIntegrationTest.shouldRestoreState` seems unrelated to this PR.

junrao (approved):

PR #9720, opened 2020-12-09
KAFKA-10555: Improve client state machine

wcarlson5:
Removes the transition to error for when there are no threads and makes Error terminal

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -291,8 +291,12 @@ public class KafkaStreams implements AutoCloseable {
             } else if (state == State.REBALANCING && newState == State.REBALANCING) {
                 // when the state is already in REBALANCING, it should not transit to REBALANCING again
                 return false;
-            } else if (state == State.ERROR && newState == State.ERROR) {
-                // when the state is already in ERROR, it should not transit to ERROR again
+            } else if (state == State.ERROR && (newState == State.PENDING_ERROR || newState == State.ERROR)) {
+                // when the state is already in ERROR, its transition to PENDING_ERROR or ERROR (due to consecutive close calls)
+                return false;
+            } else if (state == State.PENDING_ERROR && newState != State.ERROR) {
+                // when the state is already in PENDING_ERROR, all other transitions than ERROR (due to thread dying) will be
+                // refused but we do not throw exception here, to allow appropriate error handling
                 return false;
             } else if (!state.isValidTransition(newState)) {
                 throw new IllegalStateException("Stream-client " + clientId + ": Unexpected state transition from " + oldState + " to " + newState);
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -227,7 +227,8 @@ public class KafkaStreams implements AutoCloseable {
         RUNNING(1, 2, 3, 5),    // 2
         PENDING_SHUTDOWN(4),    // 3
         NOT_RUNNING,            // 4
-        ERROR(3);               // 5
+        PENDING_ERROR(6),       // 5
+        ERROR;               // 6
 
         private final Set<Integer> validTransitions = new HashSet<>();
 
@@ -469,7 +470,7 @@ public class KafkaStreams implements AutoCloseable {
                 log.error("Encountered the following exception during processing " +
                         "and the registered exception handler opted to " + action + "." +
                         " The streams client is going to shut down now. ", throwable);
-                close(Duration.ZERO);
+                closeToError();
                 break;
             case SHUTDOWN_APPLICATION:
                 if (throwable instanceof Error) {
@@ -482,7 +483,7 @@ public class KafkaStreams implements AutoCloseable {
                     log.error("Exception in global thread caused the application to attempt to shutdown." +
                             " This action will succeed only if there is at least one StreamThread running on this client." +
                             " Currently there are no running threads so will now close the client.");
-                    close(Duration.ZERO);
+                    closeToError();
                 } else {
                     processStreamThread(thread -> thread.sendShutdownRequest(AssignorError.SHUTDOWN_REQUESTED));
                     log.error("Encountered the following exception during processing " +
@@ -553,22 +554,6 @@ public class KafkaStreams implements AutoCloseable {
         }
 
         /**
-         * If all threads are dead set to ERROR
-         */
-        private void maybeSetError() {
-            // check if we have at least one thread running
-            for (final StreamThread.State state : threadState.values()) {
-                if (state != StreamThread.State.DEAD) {
-                    return;
-                }
-            }
-
-            if (setState(State.ERROR)) {
-                log.error("All stream threads have died. The instance will be in error state and should be closed.");
-            }
-        }
-
-        /**
          * If all threads are up, including the global thread, set to RUNNING
          */
         private void maybeSetRunning() {
@@ -603,8 +588,6 @@ public class KafkaStreams implements AutoCloseable {
                         setState(State.REBALANCING);
                     } else if (newState == StreamThread.State.RUNNING) {
                         maybeSetRunning();
-                    } else if (newState == StreamThread.State.DEAD) {
-                        maybeSetError();
                     }
                 } else if (thread instanceof GlobalStreamThread) {
                     // global stream thread has different invariants
@@ -614,9 +597,8 @@ public class KafkaStreams implements AutoCloseable {
                     if (newState == GlobalStreamThread.State.RUNNING) {
                         maybeSetRunning();
                     } else if (newState == GlobalStreamThread.State.DEAD) {
-                        if (setState(State.ERROR)) {
-                            log.error("Global thread has died. The instance will be in error state and should be closed.");
-                        }
+                        log.error("Global thread has died. The streams application or client will now close to ERROR.");
+                        closeToError();
                     }
                 }
             }
@@ -1204,11 +1186,21 @@ public class KafkaStreams implements AutoCloseable {
             metrics.close();
             if (!error) {
                 setState(State.NOT_RUNNING);
+            } else {
+                setState(State.ERROR);
             }
         }, "kafka-streams-close-thread");
     }
 
     private boolean close(final long timeoutMs) {
+        if (state == State.ERROR) {
+            log.info("Streams client is already in the terminal state ERROR, all resources are closed and the client has stopped.");
+            return false;
+        }
+        if (state == State.PENDING_ERROR) {
+            log.info("Streams client is in PENDING_ERROR, all resources are being closed and the client will be stopped.");
+            return false;
+        }
         if (!setState(State.PENDING_SHUTDOWN)) {
             // if transition failed, it means it was either in PENDING_SHUTDOWN
             // or NOT_RUNNING already; just check that all threads have been stopped
@@ -1230,7 +1222,7 @@ public class KafkaStreams implements AutoCloseable {
     }
 
     private void closeToError() {
-        if (!setState(State.ERROR)) {
+        if (!setState(State.PENDING_ERROR)) {
             log.info("Skipping shutdown since we are already in " + state());
         } else {
             final Thread shutdownThread = shutdownHelper(true);
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -444,7 +444,7 @@ public class KafkaStreams implements AutoCloseable {
             log.warn("The global thread cannot be replaced. Reverting to shutting down the client.");
             log.error("Encountered the following exception during processing " +
                     " The streams client is going to shut down now. ", throwable);
-            close(Duration.ZERO);
+            closeToError();
         }
         final StreamThread deadThread = (StreamThread) Thread.currentThread();
         threads.remove(deadThread);
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -202,14 +202,14 @@ public class KafkaStreams implements AutoCloseable {
      *         |              |                  |
      *         |              v                  |
      *         |       +------+-------+     +----+-------+
-     *         +-----&gt; | Pending      |&lt;--- | Error (5)  |
-     *                 | Shutdown (3) |     +------------+
-     *                 +------+-------+
-     *                        |
-     *                        v
-     *                 +------+-------+
-     *                 | Not          |
-     *                 | Running (4)  |
+     *         +-----&gt; | Pending      |     | Pending    |
+     *                 | Shutdown (3) |     | Error (5)  |
+     *                 +------+-------+     +-----+------+
+     *                        |                   |
+     *                        v                   v
+     *                 +------+-------+     +-----+--------+
+     *                 | Not          |     | Error (6)    |
+     *                 | Running (4)  |     +--------------+
      *                 +--------------+
      *
      *
@@ -217,9 +217,9 @@ public class KafkaStreams implements AutoCloseable {
      * Note the following:
      * - RUNNING state will transit to REBALANCING if any of its threads is in PARTITION_REVOKED or PARTITIONS_ASSIGNED state
      * - REBALANCING state will transit to RUNNING if all of its threads are in RUNNING state
-     * - Any state except NOT_RUNNING can go to PENDING_SHUTDOWN (whenever close is called)
+     * - Any state except NOT_RUNNING, PENDING_ERROR or ERROR can go to PENDING_SHUTDOWN (whenever close is called)
      * - Of special importance: If the global stream thread dies, or all stream threads die (or both) then
-     *   the instance will be in the ERROR state. The user will need to close it.
+     *   the instance will be in the ERROR state. The user will not need to close it.
      */
     public enum State {
         CREATED(1, 3),          // 0
@@ -228,7 +228,7 @@ public class KafkaStreams implements AutoCloseable {
         PENDING_SHUTDOWN(4),    // 3
         NOT_RUNNING,            // 4
         PENDING_ERROR(6),       // 5
-        ERROR;               // 6
+        ERROR;                  // 6
 
         private final Set<Integer> validTransitions = new HashSet<>();
 
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -1199,11 +1199,17 @@ public class KafkaStreams implements AutoCloseable {
     private boolean close(final long timeoutMs) {
         if (state == State.ERROR) {
             log.info("Streams client is already in the terminal state ERROR, all resources are closed and the client has stopped.");
-            return false;
+            return true;
         }
         if (state == State.PENDING_ERROR) {
             log.info("Streams client is in PENDING_ERROR, all resources are being closed and the client will be stopped.");
-            return false;
+            if (waitOnState(State.ERROR, timeoutMs)) {
+                log.info("Streams client stopped to ERROR completely");
+                return true;
+            } else {
+                log.info("Streams client cannot transition to ERROR completely within the timeout");
+                return false;
+            }
         }
         if (!setState(State.PENDING_SHUTDOWN)) {
             // if transition failed, it means it was either in PENDING_SHUTDOWN
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -295,6 +295,7 @@ public class StreamThread extends Thread {
     private java.util.function.Consumer<Throwable> streamsUncaughtExceptionHandler;
     private Runnable shutdownErrorHook;
     private AtomicInteger assignmentErrorCode;
+    private final ProcessingMode processingMode;
 
     public static StreamThread create(final InternalTopologyBuilder builder,
                                       final StreamsConfig config,
@@ -482,7 +483,7 @@ public class StreamThread extends Thread {
         this.shutdownErrorHook = shutdownErrorHook;
         this.streamsUncaughtExceptionHandler = streamsUncaughtExceptionHandler;
         this.cacheResizer = cacheResizer;
-
+        this.processingMode = processingMode(config);
 
         // The following sensors are created here but their references are not stored in this object, since within
         // this object they are not recorded. The sensors are created here so that the stream threads starts with all
@@ -539,8 +540,7 @@ public class StreamThread extends Thread {
         }
         boolean cleanRun = false;
         try {
-            runLoop();
-            cleanRun = true;
+            cleanRun = runLoop();
         } finally {
             completeShutdown(cleanRun);
         }
@@ -552,7 +552,7 @@ public class StreamThread extends Thread {
      * @throws IllegalStateException If store gets registered after initialized is already finished
      * @throws StreamsException      if the store's change log does not contain the partition
      */
-    void runLoop() {
+    boolean runLoop() {
         subscribeConsumer();
 
         // if the thread is still in the middle of a rebalance, we should keep polling
@@ -587,11 +587,18 @@ public class StreamThread extends Thread {
                 }
                 failedStreamThreadSensor.record();
                 this.streamsUncaughtExceptionHandler.accept(e);
+                if (processingMode == ProcessingMode.EXACTLY_ONCE_ALPHA || processingMode == ProcessingMode.EXACTLY_ONCE_BETA) {
+                    return false;
+                }
             } catch (final Throwable e) {
                 failedStreamThreadSensor.record();
                 this.streamsUncaughtExceptionHandler.accept(e);
+                if (processingMode == ProcessingMode.EXACTLY_ONCE_ALPHA || processingMode == ProcessingMode.EXACTLY_ONCE_BETA) {
+                    return false;
+                }
             }
         }
+        return true;
     }
 
     /**
--- a/streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
@@ -515,12 +515,18 @@ public class KafkaStreamsTest {
                 () -> globalStreamThread.state() == GlobalStreamThread.State.DEAD,
                 "Thread never stopped.");
             globalStreamThread.join();
-            assertEquals(streams.state(), KafkaStreams.State.PENDING_ERROR);
+            TestUtils.waitForCondition(
+                () -> streams.state() == KafkaStreams.State.PENDING_ERROR,
+                "Thread never stopped."
+            );
         } finally {
             streams.close();
         }
 
-        assertEquals(streams.state(), KafkaStreams.State.PENDING_ERROR);
+        TestUtils.waitForCondition(
+            () -> streams.state() == KafkaStreams.State.ERROR,
+            "Thread never stopped."
+        );
     }
 
     @Test
--- a/streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
@@ -429,70 +429,6 @@ public class KafkaStreamsTest {
     }
 
     @Test
-    public void stateShouldTransitToErrorIfAllThreadsDead() throws InterruptedException {
-        final KafkaStreams streams = new KafkaStreams(getBuilderWithSource().build(), props, supplier, time);
-        streams.setStateListener(streamsStateListener);
-
-        Assert.assertEquals(0, streamsStateListener.numChanges);
-        Assert.assertEquals(KafkaStreams.State.CREATED, streams.state());
-
-        streams.start();
-
-        TestUtils.waitForCondition(
-            () -> streamsStateListener.numChanges == 2,
-            "Streams never started.");
-        Assert.assertEquals(KafkaStreams.State.RUNNING, streams.state());
-
-        for (final StreamThread thread : streams.threads) {
-            threadStatelistenerCapture.getValue().onChange(
-                thread,
-                StreamThread.State.PARTITIONS_REVOKED,
-                StreamThread.State.RUNNING);
-        }
-
-        Assert.assertEquals(3, streamsStateListener.numChanges);
-        Assert.assertEquals(KafkaStreams.State.REBALANCING, streams.state());
-
-        threadStatelistenerCapture.getValue().onChange(
-            streams.threads.get(NUM_THREADS - 1),
-            StreamThread.State.PENDING_SHUTDOWN,
-            StreamThread.State.PARTITIONS_REVOKED);
-
-        threadStatelistenerCapture.getValue().onChange(
-            streams.threads.get(NUM_THREADS - 1),
-            StreamThread.State.DEAD,
-            StreamThread.State.PENDING_SHUTDOWN);
-
-        Assert.assertEquals(3, streamsStateListener.numChanges);
-        Assert.assertEquals(KafkaStreams.State.REBALANCING, streams.state());
-
-        for (final StreamThread thread : streams.threads) {
-            if (thread != streams.threads.get(NUM_THREADS - 1)) {
-                threadStatelistenerCapture.getValue().onChange(
-                    thread,
-                    StreamThread.State.PENDING_SHUTDOWN,
-                    StreamThread.State.PARTITIONS_REVOKED);
-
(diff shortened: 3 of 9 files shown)


streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -532,22 +537,6 @@ public void setGlobalStateRestoreListener(final StateRestoreListener globalState
             this.threadStatesLock = new Object();
         }
 
-        /**
-         * If all threads are dead set to ERROR
-         */
-        private void maybeSetError() {
```
wcarlson5, line 558: This will not be needed with the new error definition

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -594,9 +581,8 @@ public synchronized void onChange(final Thread thread,
                     if (newState == GlobalStreamThread.State.RUNNING) {
                         maybeSetRunning();
                     } else if (newState == GlobalStreamThread.State.DEAD) {
-                        if (setState(State.ERROR)) {
```
wcarlson5, line 617: we will be doing the same thing but closing the client for now. Maybe a replace globalThread will be added later

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1142,11 +1128,21 @@ private Thread shutdownHelper(final boolean error) {
...
             if (!error) {
                 setState(State.NOT_RUNNING);
+            } else {
+                setState(State.ERROR);
             }
         }, "kafka-streams-close-thread");
     }
 
     private boolean close(final long timeoutMs) {
+        if (state == State.ERROR) {
```
wcarlson5, line 1200: We want to make `close()` idempotent and not throw an exception but we will log a warning, but only for close so that is why these logs are not in the `setState()` method.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -546,7 +546,7 @@ public void run() {
      * @throws IllegalStateException If store gets registered after initialized is already finished
      * @throws StreamsException      if the store's change log does not contain the partition
      */
-    void runLoop() {
+    boolean runLoop() {
```
wcarlson5, line 555: This will let Streams shutdown uncleanly when in EOS mode

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -418,70 +418,6 @@ public void stateShouldTransitToRunningIfNonDeadThreadsBackToRunning() throws In
         Assert.assertEquals(KafkaStreams.State.NOT_RUNNING, streams.state());
     }
 
-    @Test
```
wcarlson5, line 431: This test is for the functionality we are removing

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -625,8 +561,7 @@ public void shouldNotAddThreadWhenError() {
         final KafkaStreams streams = new KafkaStreams(getBuilderWithSource().build(), props, supplier, time);
         final int oldSize = streams.threads.size();
         streams.start();
-        streamThreadOne.shutdown();
-        streamThreadTwo.shutdown();
+        globalStreamThread.shutdown();
```
wcarlson5, line 592: removing stream threads does not put the client in error anymore. The global does

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -524,9 +460,6 @@ public void testStateThreadClose() throws Exception {
                     "Thread never stopped.");
                 streams.threads.get(i).join();
             }
-            TestUtils.waitForCondition(
```
wcarlson5, line 537: There is no transition to Error here anymore

streams/src/test/java/org/apache/kafka/streams/integration/EOSUncleanShutdownIntegrationTest.java
```diff
@@ -99,7 +99,6 @@ public static void setupConfigsAndUtils() {
     }
 
     @Test
-    @SuppressWarnings("deprecation")
```
wcarlson5, line 102: remove deprecation

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -411,7 +405,8 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
                 errorInjectedClient1.set(false);
                 stateTransitions1.clear();
                 streams1Alpha.close();
-                waitForStateTransition(stateTransitions1, CLOSE_CRASHED);
+                final KafkaStreams finalStreams1Alpha = streams1Alpha;
+                waitForCondition(() -> finalStreams1Alpha.state() == State.ERROR, "Stream did not go to ERROR");
```
wcarlson5: Just wait for ERROR because you don't close crashed

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -526,12 +521,10 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
 
                 assignmentListener.waitForNextStableAssignment(MAX_WAIT_TIME_MS);
 
-                waitForStateTransition(stateTransitions2, CRASH);
-
                 commitErrorInjectedClient2.set(false);
                 stateTransitions2.clear();
                 streams2Alpha.close();
-                waitForStateTransition(stateTransitions2, CLOSE_CRASHED);
```
wcarlson5, line 541: No longer do you need to close crashed

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
```diff
@@ -174,8 +177,8 @@ private KafkaStreams buildStreamWithDirtyStateDir(final String stateDirPath,
     }
 
     @Test
-    @Deprecated
```
wcarlson5, line 177: remove deprecated handler

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
```diff
@@ -174,8 +177,8 @@ private KafkaStreams buildStreamWithDirtyStateDir(final String stateDirPath,
...
     public void shouldWipeOutStandbyStateDirectoryIfCheckpointIsMissing() throws Exception {
+        final long time = System.currentTimeMillis();
```
wcarlson5: There was a problem with the cleaner thread sometimes wiping out old segments because they were expired

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
```diff
@@ -139,7 +139,7 @@ public void shouldShutdownClient() throws InterruptedException {
             StreamsTestUtils.startKafkaStreamsAndWaitForRunningState(kafkaStreams);
 
             produceMessages(0L, inputTopic, "A");
-            waitForApplicationState(Collections.singletonList(kafkaStreams), KafkaStreams.State.NOT_RUNNING, DEFAULT_DURATION);
+            waitForApplicationState(Collections.singletonList(kafkaStreams), KafkaStreams.State.ERROR, DEFAULT_DURATION);
```
wcarlson5, line 148: When the client is shutdown it now goes to ERROR

wcarlson5 (commented):
@mjsax Can you give this a look? I hope that I haven't broken any of the EOS tests

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -226,7 +226,8 @@
         RUNNING(1, 2, 3, 5),    // 2
         PENDING_SHUTDOWN(4),    // 3
         NOT_RUNNING,            // 4
-        ERROR(3);               // 5
+        PENDING_ERROR(6),       // 5
+        ERROR;               // 6
```
mjsax: super nit: fix indention of comment

Also please update the state diagram in the comment above.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1165,11 +1151,21 @@ private Thread shutdownHelper(final boolean error) {
...
+            } else {
+                setState(State.ERROR);
             }
         }, "kafka-streams-close-thread");
     }
 
     private boolean close(final long timeoutMs) {
+        if (state == State.ERROR) {
+            log.info("Streams client is already in the terminal state ERROR, all resources are closed and the client has stopped.");
+            return false;
```
mjsax: Should we return `true` there? I understand that we are not in `NOT_RUNNING` state, but in the end we are in a terminal state and we did cleanup all resources. -- I guess, I am raising the question if we should relax the definition of this return value?

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
mjsax: Should we add a check that the thread count went down to zero? We could use the corresponding metrics.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -557,12 +490,12 @@ public void testStateGlobalThreadClose() throws Exception {
...
                 "Thread never stopped.");
             globalStreamThread.join();
-            assertEquals(streams.state(), KafkaStreams.State.ERROR);
+            assertEquals(streams.state(), KafkaStreams.State.PENDING_ERROR);
         } finally {
             streams.close();
         }
 
-        assertEquals(streams.state(), KafkaStreams.State.NOT_RUNNING);
+        assertEquals(streams.state(), KafkaStreams.State.PENDING_ERROR);
```
mjsax: If the global thread dies, should we not transit to `ERROR` at the end?

streams/src/test/java/org/apache/kafka/streams/integration/EOSUncleanShutdownIntegrationTest.java
mjsax: Why did we even set the exception handler that does not do anything? 🤔

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -388,7 +382,8 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
                 errorInjectedClient1.set(false);
                 stateTransitions1.clear();
                 streams1Alpha.close();
-                waitForStateTransition(stateTransitions1, CLOSE_CRASHED);
```
mjsax, line 402: Why not just change this from `CLOSE_CRASHED` to `CRASH` ?

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -113,14 +114,7 @@
...
             Collections.singletonList(
-                KeyValue.pair(KafkaStreams.State.RUNNING, KafkaStreams.State.ERROR)
-            )
-        );
-    private static final List<KeyValue<KafkaStreams.State, KafkaStreams.State>> CLOSE_CRASHED =
-        Collections.unmodifiableList(
-            Arrays.asList(
-                KeyValue.pair(KafkaStreams.State.ERROR, KafkaStreams.State.PENDING_SHUTDOWN),
-                KeyValue.pair(KafkaStreams.State.PENDING_SHUTDOWN, KafkaStreams.State.NOT_RUNNING)
+                KeyValue.pair(KafkaStreams.State.RUNNING, State.PENDING_ERROR)
```
mjsax: Should we extend it to have two state transitions, `RUNNING -> PENDING_ERROR` and `PENDING_ERROR -> ERROR` ?

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -521,12 +516,10 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
 
                 assignmentListener.waitForNextStableAssignment(MAX_WAIT_TIME_MS);
 
-                waitForStateTransition(stateTransitions2, CRASH);
```
mjsax, line 536: Why do we remove this? If the fail, we should transit to PENDING_ERROR state first, and only afterwards move forward with the test.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
mjsax: Why this change? client2 should transit from `PENDING_ERROR` to `ERROR` (not from `PENDING_SHUTDOWN` to `CLOSED`, and thus the new `CLOSE` check seem not to be right? Or do I miss something?

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -611,12 +604,11 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
                 verifyUncommitted(expectedUncommittedResult);
 
                 assignmentListener.waitForNextStableAssignment(MAX_WAIT_TIME_MS);
-                waitForStateTransition(stateTransitions1, CRASH);
```
mjsax, line 627: As above

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -611,12 +604,11 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
...
 
                 commitErrorInjectedClient1.set(false);
                 stateTransitions1.clear();
                 streams1Beta.close();
-                waitForStateTransition(stateTransitions1, CLOSE_CRASHED);
+                waitForStateTransition(stateTransitions1, CLOSE);
```
mjsax: As above

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -737,7 +729,7 @@ public void shouldUpgradeFromEosAlphaToEosBeta() throws Exception {
                 errorInjectedClient2.set(false);
                 stateTransitions2.clear();
                 streams2AlphaTwo.close();
-                waitForStateTransition(stateTransitions2, CLOSE_CRASHED);
+                waitForStateTransition(stateTransitions2, CLOSE);
```
mjsax: As above

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
mjsax: Not sure if I understand. Can you elaborate?

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
```diff
@@ -128,8 +128,8 @@ public void shouldShutdownThreadUsingOldHandler() throws InterruptedException {
             TestUtils.waitForCondition(() -> counter.get() == 1, "Handler was called 1st time");
             // should call the UncaughtExceptionHandler after rebalancing to another thread
             TestUtils.waitForCondition(() -> counter.get() == 2, DEFAULT_DURATION.toMillis(), "Handler was called 2nd time");
-            // the stream should now turn into ERROR state after 2 threads are dead
-            waitForApplicationState(Collections.singletonList(kafkaStreams), KafkaStreams.State.ERROR, DEFAULT_DURATION);
+            // there is no threads running but the client is still in running
+            waitForApplicationState(Collections.singletonList(kafkaStreams), KafkaStreams.State.RUNNING, DEFAULT_DURATION);
```
mjsax, line 132: Should we add a metric check to verify that the thread count is zero?

streams/src/test/java/org/apache/kafka/streams/integration/SuppressionIntegrationTest.java
```diff
@@ -526,6 +526,6 @@ private static void produceSynchronously(final String topic, final List<KeyValue
 
     private static void verifyErrorShutdown(final KafkaStreams driver) throws InterruptedException {
         waitForCondition(() -> !driver.state().isRunningOrRebalancing(), DEFAULT_TIMEOUT, "Streams didn't shut down.");
-        assertThat(driver.state(), is(KafkaStreams.State.PENDING_SHUTDOWN));
+        waitForCondition(() -> driver.state() == KafkaStreams.State.ERROR, "finish shutdown");
```
mjsax: Nit: the error message is weird -> `Streams didn't transit to ERROR state`.

mjsax (commented):
Made a pass over it.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: yay ascii art! good catch I forgot about that

When I build the docs locally it looks about right

image

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I don't think that we should return true. A user would then expect that that the state would then change to `NOT_RUNNING` and could be stuck waiting on that. Where if we return false in then they might retry and get stuck there.  

I don't know which is a better problem to have but I think that changing the meaning of this return value won't add any more clarity to the situation.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
wcarlson5: It will. We don't strictly need to test it here as that is tested elsewhere but we can add it for clarity

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
wcarlson5: thats probably a good idea

streams/src/test/java/org/apache/kafka/streams/integration/EOSUncleanShutdownIntegrationTest.java
wcarlson5: It made it so that the new default was not used until we updated the Error transition as we are doing in this PR. :)

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
wcarlson5: My thoughts where that we are calling close and so its is going to just be closed as the state transitions are cleared. This is the test that made me want you to review the PR I was worried that I might have broken the integrity of the tests as it is not behaving as I expected. However no other test was having this problem so I was not sure if I was understanding it correctly.

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
wcarlson5: My theory (based on logs) is that the cleaner thread was sometimes activating and removing the segment before it should so the record that makes the thread crash was not being consumed by the recovery thread sometimes. I just changed the time stamps so that the cleaner thread would not find them old and if it did activate it would not clean them.

streams/src/test/java/org/apache/kafka/streams/integration/StreamsUncaughtExceptionHandlerIntegrationTest.java
wcarlson5: We can but we should be able to see that with `counter` that the 2 threads already failed

wcarlson5 (commented):
@mjsax I believe that I got to all your comments except for the `shouldUpgradeFromEosAlphaToEosBeta ` tests one

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
mjsax: Fair point. Might be good to get input from somebody else in addition. \cc @cadonna

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
mjsax: I guess it would not hurt. Leave it up to you than.

streams/src/test/java/org/apache/kafka/streams/integration/EOSUncleanShutdownIntegrationTest.java
mjsax: Ah. Thanks for clarifying.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
mjsax: From my understanding/ according to the new state transition diagram, if there is an error we go to `PENDING_ERROR` and if we call `close()` we transit `PENDING_ERROR -> ERROR`?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -201,32 +201,33 @@
...
      *
      * </pre>
      * Note the following:
      * - RUNNING state will transit to REBALANCING if any of its threads is in PARTITION_REVOKED or PARTITIONS_ASSIGNED state
      * - REBALANCING state will transit to RUNNING if all of its threads are in RUNNING state
-     * - Any state except NOT_RUNNING can go to PENDING_SHUTDOWN (whenever close is called)
+     * - Any state except NOT_RUNNING, PENDING_ERROR or ERROR can go to PENDING_SHUTDOWN (whenever close is called)
      * - Of special importance: If the global stream thread dies, or all stream threads die (or both) then
-     *   the instance will be in the ERROR state. The user will need to close it.
+     *   the instance will be in the ERROR state. The user will not need to close it.
```
mjsax, line 222: I thought we always need to call close? If an error happens, we call the handler, and if the handler return shutdown, we transit to `PENDING_ERROR`. On `close()` we transit from `PENDING_ERROR -> ERROR`?

Or do I have some misconception?

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
mjsax: I see. You are talking about the broker side log cleaner thread -- not the local state directory cleaner thread. (windowed stores also use the concept of segments, not just topic-partitions).

However, it seems we want to extract this fix into a separate PR to be able to cherry-pick it to older branches? Seems to be unrelated to this PR.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: the handler will call close, but the user will not need to. The `PENDING_ERROR` state is indicating the resources are closing before the transition to `ERROR` after which no more work will be done. We made it so the user can call close on `PENDING_ERROR` or `ERROR` but it will only log a warning

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
wcarlson5: the call to `close()` will have no effect. The `PENDING_ERROR -> ERROR` transition is like the `PENDING_SHUTDOWN -> NOT_RUNNING ` where the transition from the `PENDING...` state is automatic

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
wcarlson5: Maybe we will. It doesn't seems to be a problem for the other branches but something with how the test changes in this PR exposed it. This happened in the handler int tests too. I think as long as the other tests don't change it won't be a problem and if those tests change we can fix it then.

But if you think it should be fixed in the other branches anyways I'll trust your judgement.

wcarlson5 (commented):
@mjsax I am still looking into the `shouldUpgradeFromEosAlphaToEosBeta` test. but not making a ton of progress I will try to carve out a little time tomorrow for it.

I think there was a bit of miss-understanding about what `PENDING_ERROR` means but I hope I cleared it up the comments below.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
mjsax: Thanks for clarifying!

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
mjsax: Thanks for clarifying.

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
mjsax: I don't understand why other branches would not be affected, or to what extend this PR would expose a new issue? Can you elaborate? -- If this PR introduces an issue that requires this change, I agree that we don't need to port it to other branches -- but I don't see the connection. What do I miss?

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
wcarlson5: Actually it could probably be just `PENDING_ERROR -> ERROR` because it could be in `RUNNING` or `REBALNCING` previously

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
wcarlson5: It looks like this issue is fixed by another PR. https://github.com/apache/kafka/pull/9733#issuecomment-761242084 So I am closing this now

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -113,14 +114,7 @@
...
             Collections.singletonList(
-                KeyValue.pair(KafkaStreams.State.RUNNING, KafkaStreams.State.ERROR)
-            )
-        );
-    private static final List<KeyValue<KafkaStreams.State, KafkaStreams.State>> CLOSE_CRASHED =
-        Collections.unmodifiableList(
-            Arrays.asList(
-                KeyValue.pair(KafkaStreams.State.ERROR, KafkaStreams.State.PENDING_SHUTDOWN),
-                KeyValue.pair(KafkaStreams.State.PENDING_SHUTDOWN, KafkaStreams.State.NOT_RUNNING)
+                KeyValue.pair(State.PENDING_ERROR, State.ERROR)
```
wcarlson5, line 117: Using just `PENDING_ERROR -> ERROR` because the transition to `PENDING_ERROR` can be from multiple sources. Also that transition is already tested so this check implies it

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
cadonna: As far as I can see, where the return value is used the javadoc says

```
true if all threads were successfully stopped, false if the timeout was reached.
```

Since all threads were successfully stopped, I would return `true`. We clearly document that `ERROR` is a terminal state, so I do not see why somebody should wait for `NOT_RUNNING` when the client is in `ERROR` and `close()` returns `true`.

wcarlson5: Alright I adjusted the close response to align with this.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -978,7 +971,7 @@ private void waitForRunning(final List<KeyValue<KafkaStreams.State, KafkaStreams
         waitForCondition(
             () -> !observed.isEmpty() && observed.get(observed.size() - 1).value.equals(State.RUNNING),
             MAX_WAIT_TIME_MS,
-            () -> "Client did not startup on time. Observers transitions: " + observed
+            () -> "Client did not have the expected state transition on time. Observers transitions: " + observed
```
mjsax: Why this change? We do wait for `RUNNING`?

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -993,6 +986,17 @@ private void waitForStateTransition(final List<KeyValue<KafkaStreams.State, Kafk
         );
     }
 
+    private void waitForStateTransitionContains(final List<KeyValue<KafkaStreams.State, KafkaStreams.State>> observed,
+                                        final List<KeyValue<KafkaStreams.State, KafkaStreams.State>> expected)
```
mjsax: nit: fix indention

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -993,6 +986,17 @@ private void waitForStateTransition(final List<KeyValue<KafkaStreams.State, Kafk
...
+            throws Exception {
+
+        waitForCondition(
+            () -> observed.containsAll(expected),
+            MAX_WAIT_TIME_MS,
+            () -> "Client did not have the expected state transition on time. Observers transitions: " + observed
```
mjsax, line 997: Can we add the expected transitions, too? Easier to debug if the test fails.

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
mjsax: The PR you liked seems to be unrelated to this test.

Still wondering if we should extract this change to a dedicated PR and cherry-pick to older branches? -- Or do we have a good explanation why older branches would not be affected?

mjsax (commented):
Just some final nits. Overall LGTM.

mjsax:
Btw: to what extend do we need to update the docs? We should at least add a section to `streams/upgrade_guide.html` to mention the change.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
wcarlson5: We do wait for running, I was thinking of bringing it to match with the other methods below but that doesn't make it anymore useful so I will revet it.

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
wcarlson5: sure that is fine

wcarlson5:
I extracted the https://github.com/apache/kafka/pull/9720/files#r562203226 fix to https://github.com/apache/kafka/pull/9948

wcarlson5:
@mjsax it looks like there is one unrelated failure `java.lang.RuntimeException: java.util.concurrent.ExecutionException: org.apache.kafka.common.errors.TopicExistsException: Topic 'primary.test-topic-2' already exists.
`

glasser:
Am I correct in believing this change was released in Kafka 2.8.0? I can't find it in [the release notes](https://archive.apache.org/dist/kafka/2.8.0/RELEASE_NOTES.html). We just had a minor outage because we didn't realize when upgrading to v2.8 that some error handling/shutdown code in our Kafka Streams app couldn't assume that it only had to kick in when the state transitioned to `NOT_RUNNING` but also now has to handle `ERROR` too.

PR #9727, opened 2020-12-10
[KAFKA-10417] Update Cogrouped processor to work with suppress() and joins

lct45:
Changes the cogrouped processor from `PassThrough` to `KTablePassThrough` to allow for sending old values. `KTablePassThrough` extends `KTableProcessorSupplier` instead of `ProcessorSupplier` to implement sending old values and the `view()` method.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
+++ b/streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
@@ -62,7 +62,8 @@ class CogroupedStreamAggregateBuilder<K, VOut> {
         boolean stateCreated = false;
         int counter = 0;
         for (final Entry<KGroupedStreamImpl<K, ?>, Aggregator<? super K, Object, VOut>> kGroupedStream : groupPatterns.entrySet()) {
-            final KStreamAggProcessorSupplier<K, K, ?, ?> parentProcessor = new KStreamAggregate<>(storeBuilder.name(), initializer, kGroupedStream.getValue());
+            final KStreamAggProcessorSupplier<K, K, ?, ?> parentProcessor =
+                new KStreamAggregate<>(storeBuilder.name(), initializer, kGroupedStream.getValue());
             parentProcessors.add(parentProcessor);
             final StatefulProcessorNode<K, ?> statefulProcessorNode = getStatefulProcessorNode(
                 named.suffixWithOrElseGet(
@@ -95,7 +96,12 @@ class CogroupedStreamAggregateBuilder<K, VOut> {
         boolean stateCreated = false;
         int counter = 0;
         for (final Entry<KGroupedStreamImpl<K, ?>, Aggregator<? super K, Object, VOut>> kGroupedStream : groupPatterns.entrySet()) {
-            final KStreamAggProcessorSupplier<K, K, ?, ?>  parentProcessor = (KStreamAggProcessorSupplier<K, K, ?, ?>) new KStreamWindowAggregate<K, K, VOut, W>(windows, storeBuilder.name(), initializer, kGroupedStream.getValue());
+            final KStreamAggProcessorSupplier<K, K, ?, ?>  parentProcessor =
+                (KStreamAggProcessorSupplier<K, K, ?, ?>) new KStreamWindowAggregate<K, K, VOut, W>(
+                    windows,
+                    storeBuilder.name(),
+                    initializer,
+                    kGroupedStream.getValue());
             parentProcessors.add(parentProcessor);
             final StatefulProcessorNode<K, ?> statefulProcessorNode = getStatefulProcessorNode(
                 named.suffixWithOrElseGet(
@@ -128,7 +134,13 @@ class CogroupedStreamAggregateBuilder<K, VOut> {
         boolean stateCreated = false;
         int counter = 0;
         for (final Entry<KGroupedStreamImpl<K, ?>, Aggregator<? super K, Object, VOut>> kGroupedStream : groupPatterns.entrySet()) {
-            final KStreamAggProcessorSupplier<K, K, ?, ?> parentProcessor = (KStreamAggProcessorSupplier<K, K, ?, ?>) new KStreamSessionWindowAggregate<K, K, VOut>(sessionWindows, storeBuilder.name(), initializer, kGroupedStream.getValue(), sessionMerger);
+            final KStreamAggProcessorSupplier<K, K, ?, ?> parentProcessor =
+                (KStreamAggProcessorSupplier<K, K, ?, ?>) new KStreamSessionWindowAggregate<K, K, VOut>(
+                    sessionWindows,
+                    storeBuilder.name(),
+                    initializer,
+                    kGroupedStream.getValue(),
+                    sessionMerger);
             parentProcessors.add(parentProcessor);
             final StatefulProcessorNode<K, ?> statefulProcessorNode = getStatefulProcessorNode(
                 named.suffixWithOrElseGet(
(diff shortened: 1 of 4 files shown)


lct45:
@wcarlson5 @vvcephei for initial review

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -201,9 +219,10 @@ private void processRepartitions(final Map<KGroupedStreamImpl<K, ?>, Aggregator<
             "-cogroup-merge",
             builder,
             CogroupedKStreamImpl.MERGE_NAME);
-        final ProcessorSupplier<K, VOut> passThrough = new PassThrough<>();
```
wcarlson5, line 204: I believe PassThrough is only used for the cogroup but now I think you can remove it completely.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -115,24 +124,28 @@
                                 final Merger<? super K, VOut> sessionMerger) {
         processRepartitions(groupPatterns, storeBuilder);
         final Collection<GraphNode> processors = new ArrayList<>();
+        final Collection<KStreamAggProcessorSupplier> parentProcessors = new ArrayList<>();
         boolean stateCreated = false;
         int counter = 0;
         for (final Entry<KGroupedStreamImpl<K, ?>, Aggregator<? super K, Object, VOut>> kGroupedStream : groupPatterns.entrySet()) {
+            final KStreamAggProcessorSupplier<K, K, ?, ?> parentProcessor = (KStreamAggProcessorSupplier<K, K, ?, ?>) new KStreamSessionWindowAggregate<K, K, VOut>(sessionWindows, storeBuilder.name(), initializer, kGroupedStream.getValue(), sessionMerger);
```
wcarlson5: I think these lines are too long. The ones in the other methods too

wcarlson5 (commented):
@lct45  Thanks for getting this fix done.

I only had a couple comments

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
lct45: It's still used for some of the kstream ksteam internals so I think it's gotta stay for now

wcarlson5: Ok. then LGTM

danielkman:
👍

streams/src/main/java/org/apache/kafka/streams/kstream/internals/KTablePassThrough.java
```diff
@@ -0,0 +1,88 @@
...
+        this.storeName = storeName;
+    }
+
+    @Override
+    public Processor<K, Change<V>> get() {
+        return new KTablePassThroughProcessor();
+    }
+
+    @Override
+    public boolean enableSendingOldValues(final boolean forceMaterialization) {
```
vvcephei, line 43: It looks like the parameter is ignored. Is that correct? If so, it might be nice to add an explanatory comment.

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
```diff
@@ -201,9 +236,10 @@ private void processRepartitions(final Map<KGroupedStreamImpl<K, ?>, Aggregator<
             "-cogroup-merge",
             builder,
             CogroupedKStreamImpl.MERGE_NAME);
-        final ProcessorSupplier<K, VOut> passThrough = new PassThrough<>();
+        final KTableProcessorSupplier<K, VOut, VOut> passThrough = new KTablePassThrough<>(parentProcessors, queryableName);
```
vvcephei: I noticed that `queryableName` is a different parameter than the one we actually build the parent processors with (`storeBuilder.name()`). It wouldn't surprise me if there's a subtle difference between them.

streams/src/test/java/org/apache/kafka/streams/kstream/internals/CogroupedKStreamImplTest.java
```diff
@@ -1202,6 +1204,46 @@ public void testCogroupWithThreeGroupedStreams() {
         }
     }
 
+    @Test
+    public void testCogroupWithKTableKTableInnerJoin() {
```
vvcephei, line 1208: Just checking: does this fail on trunk?

streams/src/test/java/org/apache/kafka/streams/kstream/internals/SuppressScenarioTest.java
```diff
@@ -803,6 +808,18 @@ public void shouldWorkBeforeJoinLeft() {
 
     }
 
+    @Test
+    public void shouldWorkWithCogrouped() {
```
vvcephei, line 812: And this one too?

vvcephei (commented):
Thanks, @lct45 !

Just a couple of small comments, otherwise it LGTM.

streams/src/test/java/org/apache/kafka/streams/kstream/internals/CogroupedKStreamImplTest.java
lct45: It does

streams/src/test/java/org/apache/kafka/streams/kstream/internals/SuppressScenarioTest.java
lct45: Also fails (:

streams/src/main/java/org/apache/kafka/streams/kstream/internals/CogroupedStreamAggregateBuilder.java
lct45: Looks like they're the same when cogroup is materialized, but if it's not the `storeBuilder.name()` is auto generated and `queryableName` is null. With that said, it seems sto me that it would be better to pass `storeBuilder.name()` to the `passThrough`, to be safe. WDYT?

vvcephei: Yes, that sounds like the right thing to do. Thanks!

streams/src/test/java/org/apache/kafka/streams/kstream/internals/CogroupedKStreamImplTest.java
vvcephei: Thank you!

streams/src/test/java/org/apache/kafka/streams/kstream/internals/SuppressScenarioTest.java
vvcephei: Thank you!

vvcephei (approved):
Thanks @lct45 !

I'll merge it after the tests are green.

vvcephei:
The test failures were unrelated:


Build / JDK 8 / kafka.network.SocketServerTest.testIdleConnection | 15 sec | 1
-- | -- | --
Build / JDK 11 / org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true] | 1 min 24 sec | 1
Build / JDK 15 / kafka.api.TransactionsBounceTest.testWithGroupMetadata | 24 sec | 1
Build / JDK 15 / kafka.network.ConnectionQuotasTest.testNonDefaultConnectionCountLimitAndRateLimit

PR #9728, opened 2020-12-10
MINOR: make sure all dir jobs are completed

chia7712:
revert the change introduced by #9680 and add a unit test to verify

### DIFF
--- a/core/src/main/scala/kafka/log/LogManager.scala
+++ b/core/src/main/scala/kafka/log/LogManager.scala
@@ -437,20 +437,6 @@ class LogManager(logDirs: Seq[File],
   }
 
   /**
-   * wait all jobs to complete
-   * @param jobs jobs
-   * @return true if all pass. Otherwise, false
-   */
-  private[log] def waitForAllToComplete(jobs: Seq[Future[_]]): Boolean = {
-    jobs.count(future => Try(future.get) match {
-      case Success(_) => false
-      case Failure(e) =>
-        warn(s"There was an error in one of the threads during LogManager shutdown: ${e.getCause}")
-        true
-    }) == 0
-  }
-
-  /**
    * Close all the logs
    */
   def shutdown(): Unit = {
@@ -494,7 +480,8 @@ class LogManager(logDirs: Seq[File],
 
     try {
       jobs.forKeyValue { (dir, dirJobs) =>
-        if (waitForAllToComplete(dirJobs)) {
+        if (waitForAllToComplete(dirJobs,
+          e => warn(s"There was an error in one of the threads during LogManager shutdown: ${e.getCause}"))) {
           val logs = logsInDir(localLogsByDir, dir)
 
           // update the last flush point
@@ -1172,6 +1159,21 @@ class LogManager(logDirs: Seq[File],
 }
 
 object LogManager {
+
+  /**
+   * wait all jobs to complete
+   * @param jobs jobs
+   * @param callback handle the exception caused by Future#get
+   * @return true if all pass. Otherwise, false
+   */
+  private[log] def waitForAllToComplete(jobs: Seq[Future[_]], callback: Throwable => Unit): Boolean = {
+    jobs.count(future => Try(future.get) match {
+      case Success(_) => false
+      case Failure(e) =>
+        callback(e)
+        true
+    }) == 0
+  }
 
   val RecoveryPointCheckpointFile = "recovery-point-offset-checkpoint"
   val LogStartOffsetCheckpointFile = "log-start-offset-checkpoint"
--- a/core/src/main/scala/kafka/log/LogManager.scala
+++ b/core/src/main/scala/kafka/log/LogManager.scala
@@ -35,6 +35,7 @@ import scala.jdk.CollectionConverters._
 import scala.collection._
 import scala.collection.mutable.ArrayBuffer
 import scala.util.{Failure, Success, Try}
+import kafka.utils.Implicits._
 
 /**
  * The entry point to the kafka log management subsystem. The log manager is responsible for log creation, retrieval, and cleaning.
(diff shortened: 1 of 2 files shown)


core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -435,6 +436,20 @@ class LogManager(logDirs: Seq[File],
       cleaner.startup()
   }
 
+  /**
+   * wait all jobs to complete
+   * @param jobs jobs
+   * @return true if all pass. Otherwise, false
+   */
+  private[log] def allPass(jobs: Seq[Future[_]]): Boolean = {
```
ijuma: The name makes this look generic, but it's not. The logging is very specific. I would change the name to be specific.

chia7712: How about "waitForAllToComplete"

ijuma: The log message talks explicitly about LogManager shutdown. Seems like you'd want it to say something about log manager shutdown. I would even set the type parameter to the type we expect from the shutdown method.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -478,17 +493,8 @@ class LogManager(logDirs: Seq[File],
...
-            case Success(_) => false
-            case Failure(e) =>
-              warn(s"There was an error in one of the threads during LogManager shutdown: ${e.getCause}")
-              true
-          }
-        }
-
-        if (!hasErrors) {
+      jobs.forKeyValue { (dir, dirJobs) =>
+        if (allPass(dirJobs)) {
```
kowshik: This refactoring is good.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
```diff
@@ -680,4 +680,15 @@ class LogManagerTest {
     time.sleep(logConfig.fileDeleteDelayMs + 1)
     verifyMetrics(1)
   }
+
+  @Test
```
kowshik, line 685: This test may not still catch the bug introduced in #9680. 
We can continue more discussion under this review comment in #9680: https://github.com/apache/kafka/pull/9680/files#r540476157.

We should still get this PR merged though, since it fixes the problem.

cc @ijuma

kowshik (approved):
LGTM. Thanks for the fix @chia7712 ! I've added a few comments. It will be useful to get this PR into master.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
chia7712: @kowshik Thanks for you explanation! It is indeed hard to give full test on the shutdown path.

> This test may not still catch the bug introduced in #9680.

you are right. I'm going to enrich the unit tests.

chia7712: For another, Your comment in ttps://github.com/apache/kafka/pull/9680/files#r540476157 is great to me. It would be better that we can make higher test coverage on that!

core/src/main/scala/kafka/log/LogManager.scala
chia7712: > I would even set the type parameter to the type we expect from the shutdown method.

There is no explicit type so I didn't define the type parameter.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
kowshik: Sounds good!

kowshik (approved):
LGTM, thanks for the updated PR!

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -435,6 +436,20 @@ class LogManager(logDirs: Seq[File],
       cleaner.startup()
   }
 
+  /**
+   * wait all jobs to complete
+   * @param jobs jobs
+   * @return true if all pass. Otherwise, false
+   */
+  private[log] def waitForAllToComplete(jobs: Seq[Future[_]]): Boolean = {
```
ijuma: I would add a callback here for the failures and do the logging in the callback. That way you can keep the name generic.

chia7712: good suggestion. will address it.

chia7712:
@ijuma Could you take a look?

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -1167,6 +1160,21 @@ class LogManager(logDirs: Seq[File],
 
 object LogManager {
 
+  /**
+   * wait all jobs to complete
+   * @param jobs jobs
+   * @param callback handle the exception caused by Future#get
```
ijuma: We should mention that this will be called for each future that throws an exception.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -1167,6 +1160,21 @@ class LogManager(logDirs: Seq[File],
 
 object LogManager {
 
+  /**
+   * wait all jobs to complete
```
ijuma: Nit: start with capital?

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
```diff
@@ -680,4 +681,31 @@ class LogManagerTest {
...
+  def testWaitForAllToComplete(): Unit = {
+    var invokedCount = 0
+    val success: Future[Boolean] = Mockito.mock(classOf[Future[Boolean]])
+    Mockito.when(success.get()).thenAnswer(_ => invokedCount += 1)
```
ijuma: Should we not return `true` in this answer?

ijuma (commented):
Looks good overall, just a few nits.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
chia7712: ```waitForAllToComplete``` does not use the return value so it does not cause casting error even though we don't return ```true```. However, it does look like a bug so I will add the return value.

ijuma (approved):
LGTM, thanks.

PR #9730, opened 2020-12-11
MINOR: fix error message

mjsax:
Got the accumulator lambda wrong in #9688 

Call for review @ableegoldman

### DIFF
--- a/streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
@@ -1045,8 +1045,8 @@ public class EosBetaUpgradeIntegrationTest {
                 assertThat(getAllRecordPerKey(key, result), equalTo(getAllRecordPerKey(key, expectedResult)));
             } catch (final AssertionError error) {
                 throw new AssertionError(
-                    "expected result: " + expectedResult.stream().map(KeyValue::toString).reduce("", (kv, str) -> str.isEmpty() ? kv : str + ", " + kv) +
-                    "\nreceived records: " + result.stream().map(KeyValue::toString).reduce("", (kv, str) -> str.isEmpty() ? kv : str + ", " + kv),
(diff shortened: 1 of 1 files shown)


streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
```diff
@@ -1045,8 +1045,8 @@ private void checkResultPerKey(final List<KeyValue<Long, Long>> result,
                 assertThat(getAllRecordPerKey(key, result), equalTo(getAllRecordPerKey(key, expectedResult)));
             } catch (final AssertionError error) {
                 throw new AssertionError(
-                    "expected result: " + expectedResult.stream().map(KeyValue::toString).reduce("", (kv, str) -> str.isEmpty() ? kv : str + ", " + kv) +
-                    "\nreceived records: " + result.stream().map(KeyValue::toString).reduce("", (kv, str) -> str.isEmpty() ? kv : str + ", " + kv),
+                    "expected result: " + expectedResult.stream().map(KeyValue::toString).reduce("", (str, kv) -> str.isEmpty() ? kv : str + ", " + kv) +
```
chia7712: How about using ```.collect(Collectors.joining(",")```?

ableegoldman (approved):

streams/src/test/java/org/apache/kafka/streams/integration/EosBetaUpgradeIntegrationTest.java
mjsax: Still a noob using the JavaStream API... Does this only use the separator if there are more than one element?

chia7712: > Does this only use the separator if there are more than one element?

yep

chia7712 (approved):
@mjsax +1 to latest commit :)

mjsax:
Merged to trunk and cherry-picked to `2.7` branch.

PR #9731, opened 2020-12-11
MINOR: Make Log.recordVersion private and other small cleanups

kowshik:
Couple small cleanups in this PR:
- I've converted `Log.recordVersion` to be private, since it is used only inside `Log` class. I've also changed one of the call sites to reuse the same method.
- In a few classes I've removed a redundant default parameter `fileAlreadyExists = false` in a call to `LogSegment.open`.

**Tests:**
Relying on existing unit & integration tests.

### DIFF
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -535,7 +535,7 @@ class Log(@volatile private var _dir: File,
   /** The name of this log */
   def name  = dir.getName()
 
-  def recordVersion: RecordVersion = config.messageFormatVersion.recordVersion
+  private def recordVersion: RecordVersion = config.messageFormatVersion.recordVersion
 
   private def initializeLeaderEpochCache(): Unit = lock synchronized {
     val leaderEpochFile = LeaderEpochCheckpointFile.newFile(dir)
@@ -870,7 +870,6 @@ class Log(@volatile private var _dir: File,
                                    reloadFromCleanShutdown: Boolean,
                                    producerStateManager: ProducerStateManager): Unit = lock synchronized {
     checkIfMemoryMappedBufferClosed()
-    val messageFormatVersion = config.messageFormatVersion.recordVersion.value
     val segments = logSegments
     val offsetsToSnapshot =
       if (segments.nonEmpty) {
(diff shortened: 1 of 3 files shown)


kowshik:
cc @junrao @ijuma @chia7712 for review

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -893,7 +892,7 @@ class Log(@volatile private var _dir: File,
     // offset (see below). The next time the log is reloaded, we will load producer state using this snapshot
     // (or later snapshots). Otherwise, if there is no snapshot file, then we have to rebuild producer state
     // from the first segment.
-    if (messageFormatVersion < RecordBatch.MAGIC_VALUE_V2 ||
+    if (recordVersion.value < RecordBatch.MAGIC_VALUE_V2 ||
```
chia7712, line 895: How about replacing ```messageFormatVersion``` by ```recordVersion.value``` for all usages in this method.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1970,7 +1969,6 @@ class Log(@volatile private var _dir: File,
           baseOffset = newOffset,
           config,
           time = time,
```
chia7712, line 1971: there is a same issue https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/log/LogCleaner.scala#L456

chia7712 (approved):
@kowshik thanks for this patch. some small comments are left. overall +1

core/src/main/scala/kafka/log/Log.scala
kowshik: This was already done. Please let me now if thats not the case.

core/src/main/scala/kafka/log/Log.scala
kowshik: Done.

kowshik:
Thanks for the review @chia7712 ! I've addressed the comments in 2d2b4a2e732b42397df62da141df30856e5c88db.

kowshik:
@chia7712 if the change looks good to you, please could you help merge it? :)

chia7712 (approved):
@kowshik LGTM

will commit it tomorrow if there is no objection.

PR #9732, opened 2020-12-11
KAFKA-10842; Use `InterBrokerSendThread` for raft's outbound network channel

hachikuji:
This patch contains the following improvements:

- Separate inbound/outbound request flows so that we can open the door for concurrent inbound request handling
- Rewrite `KafkaNetworkChannel` to use `InterBrokerSendThread` which fixes a number of bugs/shortcomings
- Get rid of a lot of boilerplate conversions in `KafkaNetworkChannel` 
- Improve validation of inbound responses in `KafkaRaftClient` by checking correlationId. This fixes a bug which could cause an out of order Fetch to be applied incorrectly.

### DIFF
--- a/core/src/main/scala/kafka/common/InterBrokerSendThread.scala
+++ b/core/src/main/scala/kafka/common/InterBrokerSendThread.scala
@@ -54,7 +54,11 @@ class InterBrokerSendThread(
   }
 
   def sendRequest(request: RequestAndCompletionHandler): Unit = {
-    inboundQueue.offer(request)
+    sendRequests(Seq(request))
+  }
+
+  def sendRequests(requests: Iterable[RequestAndCompletionHandler]): Unit = {
+    inboundQueue.addAll(requests.asJavaCollection)
     wakeup()
   }
 
--- a/core/src/main/scala/kafka/common/InterBrokerSendThread.scala
+++ b/core/src/main/scala/kafka/common/InterBrokerSendThread.scala
@@ -17,7 +17,6 @@
 package kafka.common
 
 import java.util.Map.Entry
-import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.{ArrayDeque, ArrayList, Collection, Collections, HashMap, Iterator}
 
 import kafka.utils.ShutdownableThread
@@ -33,7 +32,7 @@ import scala.jdk.CollectionConverters._
 /**
  *  Class for inter-broker send thread that utilize a non-blocking network client.
  */
-class InterBrokerSendThread(
+abstract class InterBrokerSendThread(
   name: String,
   networkClient: KafkaClient,
   requestTimeoutMs: Int,
@@ -41,8 +40,9 @@ class InterBrokerSendThread(
   isInterruptible: Boolean = true
 ) extends ShutdownableThread(name, isInterruptible) {
 
-  private val inboundQueue = new ConcurrentLinkedQueue[RequestAndCompletionHandler]()
   private val unsentRequests = new UnsentRequests
+
+  def generateRequests(): Iterable[RequestAndCompletionHandler]
 
   def hasUnsentRequests: Boolean = unsentRequests.iterator().hasNext
 
@@ -53,35 +53,25 @@ class InterBrokerSendThread(
     awaitShutdown()
   }
 
-  def sendRequest(request: RequestAndCompletionHandler): Unit = {
-    sendRequests(Seq(request))
-  }
-
-  def sendRequests(requests: Iterable[RequestAndCompletionHandler]): Unit = {
-    inboundQueue.addAll(requests.asJavaCollection)
-    wakeup()
-  }
-
-  private def drainInboundQueue(): Unit = {
-    while (!inboundQueue.isEmpty) {
-      val request = inboundQueue.poll()
-      val completionHandler = request.handler
+  private def drainGeneratedRequests(): Unit = {
+    generateRequests().foreach { request =>
       unsentRequests.put(request.destination,
         networkClient.newClientRequest(
           request.destination.idString,
           request.request,
-          time.milliseconds(),
+          request.creationTimeMs,
           true,
           requestTimeoutMs,
-          completionHandler))
+          request.handler
+        ))
     }
   }
 
-  override def doWork(): Unit = {
+  protected def pollOnce(maxTimeoutMs: Long): Unit = {
     try {
+      drainGeneratedRequests()
       var now = time.milliseconds()
-      drainInboundQueue()
-      val timeout = sendRequests(now)
+      val timeout = sendRequests(now, maxTimeoutMs)
       networkClient.poll(timeout, now)
       now = time.milliseconds()
       checkDisconnects(now)
@@ -99,8 +89,12 @@ class InterBrokerSendThread(
     }
   }
 
-  private def sendRequests(now: Long): Long = {
-    var pollTimeout = Long.MaxValue
+  override def doWork(): Unit = {
+    pollOnce(Long.MaxValue)
+  }
+
+  private def sendRequests(now: Long, maxTimeoutMs: Long): Long = {
+    var pollTimeout = maxTimeoutMs
     for (node <- unsentRequests.nodes.asScala) {
       val requestIterator = unsentRequests.requestIterator(node)
       while (requestIterator.hasNext) {
@@ -157,9 +151,12 @@ class InterBrokerSendThread(
   def wakeup(): Unit = networkClient.wakeup()
 }
 
-case class RequestAndCompletionHandler(destination: Node,
-                                       request: AbstractRequest.Builder[_ <: AbstractRequest],
-                                       handler: RequestCompletionHandler)
+case class RequestAndCompletionHandler(
+  creationTimeMs: Long,
+  destination: Node,
+  request: AbstractRequest.Builder[_ <: AbstractRequest],
+  handler: RequestCompletionHandler
+)
 
 private class UnsentRequests {
   private val unsent = new HashMap[Node, ArrayDeque[ClientRequest]]
--- a/core/src/main/scala/kafka/common/InterBrokerSendThread.scala
+++ b/core/src/main/scala/kafka/common/InterBrokerSendThread.scala
@@ -16,8 +16,9 @@
  */
 package kafka.common
 
-import java.util.{ArrayDeque, ArrayList, Collection, Collections, HashMap, Iterator}
 import java.util.Map.Entry
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.{ArrayDeque, ArrayList, Collection, Collections, HashMap, Iterator}
 
 import kafka.utils.ShutdownableThread
 import org.apache.kafka.clients.{ClientRequest, ClientResponse, KafkaClient, RequestCompletionHandler}
@@ -32,17 +33,18 @@ import scala.jdk.CollectionConverters._
 /**
  *  Class for inter-broker send thread that utilize a non-blocking network client.
  */
-abstract class InterBrokerSendThread(name: String,
-                                     networkClient: KafkaClient,
-                                     time: Time,
-                                     isInterruptible: Boolean = true)
-  extends ShutdownableThread(name, isInterruptible) {
+class InterBrokerSendThread(
+  name: String,
+  networkClient: KafkaClient,
+  requestTimeoutMs: Int,
+  time: Time,
+  isInterruptible: Boolean = true
+) extends ShutdownableThread(name, isInterruptible) {
 
-  def generateRequests(): Iterable[RequestAndCompletionHandler]
-  def requestTimeoutMs: Int
+  private val inboundQueue = new ConcurrentLinkedQueue[RequestAndCompletionHandler]()
   private val unsentRequests = new UnsentRequests
 
-  def hasUnsentRequests = unsentRequests.iterator().hasNext
+  def hasUnsentRequests: Boolean = unsentRequests.iterator().hasNext
 
   override def shutdown(): Unit = {
     initiateShutdown()
@@ -51,22 +53,30 @@ abstract class InterBrokerSendThread(name: String,
     awaitShutdown()
   }
 
-  override def doWork(): Unit = {
-    var now = time.milliseconds()
+  def sendRequest(request: RequestAndCompletionHandler): Unit = {
+    inboundQueue.offer(request)
+    wakeup()
+  }
 
-    generateRequests().foreach { request =>
+  private def drainInboundQueue(): Unit = {
+    while (!inboundQueue.isEmpty) {
+      val request = inboundQueue.poll()
       val completionHandler = request.handler
       unsentRequests.put(request.destination,
         networkClient.newClientRequest(
           request.destination.idString,
           request.request,
-          now,
+          time.milliseconds(),
           true,
           requestTimeoutMs,
           completionHandler))
     }
+  }
 
+  override def doWork(): Unit = {
     try {
+      var now = time.milliseconds()
+      drainInboundQueue()
       val timeout = sendRequests(now)
       networkClient.poll(timeout, now)
       now = time.milliseconds()
@@ -198,5 +208,5 @@ private class UnsentRequests {
       requests.iterator
   }
 
-  def nodes = unsent.keySet
+  def nodes: java.util.Set[Node] = unsent.keySet
 }
--- a/core/src/main/scala/kafka/coordinator/transaction/TransactionMarkerChannelManager.scala
+++ b/core/src/main/scala/kafka/coordinator/transaction/TransactionMarkerChannelManager.scala
@@ -127,11 +127,14 @@ class TxnMarkerQueue(@volatile var destination: Node) {
   def totalNumMarkers(txnTopicPartition: Int): Int = markersPerTxnTopicPartition.get(txnTopicPartition).fold(0)(_.size)
 }
 
-class TransactionMarkerChannelManager(config: KafkaConfig,
-                                      metadataCache: MetadataCache,
-                                      networkClient: NetworkClient,
-                                      txnStateManager: TransactionStateManager,
-                                      time: Time) extends InterBrokerSendThread("TxnMarkerSenderThread-" + config.brokerId, networkClient, time) with Logging with KafkaMetricsGroup {
+class TransactionMarkerChannelManager(
+  config: KafkaConfig,
+  metadataCache: MetadataCache,
+  networkClient: NetworkClient,
+  txnStateManager: TransactionStateManager,
+  time: Time
+) extends InterBrokerSendThread("TxnMarkerSenderThread-" + config.brokerId, networkClient, config.requestTimeoutMs, time)
+  with Logging with KafkaMetricsGroup {
 
   this.logIdent = "[Transaction Marker Channel Manager " + config.brokerId + "]: "
 
@@ -145,8 +148,6 @@ class TransactionMarkerChannelManager(config: KafkaConfig,
 
   private val transactionsWithPendingMarkers = new ConcurrentHashMap[String, PendingCompleteTxn]
 
-  override val requestTimeoutMs: Int = config.requestTimeoutMs
-
   val writeTxnMarkersRequestVersion: Short =
     if (config.interBrokerProtocolVersion >= KAFKA_2_8_IV0) 1
     else 0
@@ -154,7 +155,10 @@ class TransactionMarkerChannelManager(config: KafkaConfig,
   newGauge("UnknownDestinationQueueSize", () => markersQueueForUnknownBroker.totalNumMarkers)
   newGauge("LogAppendRetryQueueSize", () => txnLogAppendRetryQueue.size)
 
-  override def generateRequests() = drainQueuedTransactionMarkers()
+  override def doWork(): Unit = {
+    drainQueuedTransactionMarkers().foreach(super.sendRequest)
+    super.doWork()
+  }
 
   override def shutdown(): Unit = {
     super.shutdown()
--- a/core/src/main/scala/kafka/coordinator/transaction/TransactionMarkerChannelManager.scala
+++ b/core/src/main/scala/kafka/coordinator/transaction/TransactionMarkerChannelManager.scala
@@ -156,7 +156,7 @@ class TransactionMarkerChannelManager(
   newGauge("LogAppendRetryQueueSize", () => txnLogAppendRetryQueue.size)
 
   override def doWork(): Unit = {
-    drainQueuedTransactionMarkers().foreach(super.sendRequest)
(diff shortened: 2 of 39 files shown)


clients/src/main/java/org/apache/kafka/common/requests/AddOffsetsToTxnRequest.java
```diff
@@ -53,7 +53,7 @@ public AddOffsetsToTxnRequest(AddOffsetsToTxnRequestData data, short version) {
     }
 
     @Override
-    protected AddOffsetsToTxnRequestData data() {
+    public AddOffsetsToTxnRequestData data() {
```
ijuma: It seems like there are many classes where direct access to `data` is not actually needed. How do you feel about having a public method in `RequestUtils` that exposes `data` for the raft layer instead? Something like:

```java
public static ApiMessage requestData(AbstractRequest req)
public static ApiMessage responseData(AbstractResponse resp)
```

Or is it not worth it?

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
```diff
@@ -51,22 +53,30 @@ abstract class InterBrokerSendThread(name: String,
...
-    var now = time.milliseconds()
+  def sendRequest(request: RequestAndCompletionHandler): Unit = {
+    inboundQueue.offer(request)
+    wakeup()
+  }
 
-    generateRequests().foreach { request =>
+  private def drainInboundQueue(): Unit = {
+    while (!inboundQueue.isEmpty) {
+      val request = inboundQueue.poll()
```
mumrah: Similar to above comment, either check for null or use `remove()`

clients/src/main/java/org/apache/kafka/common/requests/AddOffsetsToTxnRequest.java
hachikuji: Yeah, not sure it's worth it, though I don't feel strongly. I think ultimately we're going to start relying more on the generated classes to avoid unnecessary conversions. We're now entering "phase 2" of the request overhaul which means we can start figuring out how to remove the AbstractRequest/AbstractResponse layer. I think it will take more smarts in the generated classes to make a dent here, but if we are agreed on the goal (?), then I do not think preserving the encapsulation here is worthwhile.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
```diff
@@ -32,17 +33,18 @@ import scala.jdk.CollectionConverters._
...
-  def requestTimeoutMs: Int
+class InterBrokerSendThread(
+  name: String,
+  networkClient: KafkaClient,
+  requestTimeoutMs: Int,
+  time: Time,
+  isInterruptible: Boolean = true
+) extends ShutdownableThread(name, isInterruptible) {
+
+  private val inboundQueue = new ConcurrentLinkedQueue[RequestAndCompletionHandler]()
```
mumrah: Just wondering, is there any reason why we might want a bounded queue here? I suspect not since we never expect too many requests to be enqueued at once.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
```diff
@@ -51,22 +53,30 @@ abstract class InterBrokerSendThread(name: String,
     awaitShutdown()
   }
 
-  override def doWork(): Unit = {
-    var now = time.milliseconds()
+  def sendRequest(request: RequestAndCompletionHandler): Unit = {
+    inboundQueue.offer(request)
```
mumrah: We should probably either check the result of `offer` or use `BlockingQueue#add` instead. Since we're using an unbounded queue and never expect this to fail, I would lean towards `add`

clients/src/main/java/org/apache/kafka/common/requests/AddOffsetsToTxnRequest.java
ijuma: I don't quite understand how we would handle versioning well if we only have data classes. Do you have thoughts on that?

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
```diff
@@ -51,22 +53,30 @@ abstract class InterBrokerSendThread(name: String,
...
+          time.milliseconds(),
           true,
           requestTimeoutMs,
           completionHandler))
     }
+  }
 
+  override def doWork(): Unit = {
     try {
+      var now = time.milliseconds()
```
mumrah, line 73: Previously, all the requests we gathered from `generateRequests` would have the same timestamp which was also passed to the subsequent call to NetworkClient#ready and send (via `sendRequests`). 

What's the reason for recomputing the timestamp for each request we create? 

Should we get a newer timestamp for the call to NetworkClient? 

Seems a little weird to create requests at t1, t2, etc and then call NetworkClient.send with t0. I wonder if this would have weird throttling side effects.

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
```diff
@@ -68,179 +51,93 @@ object KafkaNetworkChannel {
...
-    for (response <- responses.asScala) {
-      inboundMessages.add(buildInboundRaftResponse(response))
+      case None =>
+        completeFuture(errorResponse(request.data, Errors.BROKER_NOT_AVAILABLE))
     }
   }
 
-  private def drainInboundRequests(inboundMessages: util.List[RaftMessage]): Unit = {
-    undelivered.drainTo(inboundMessages)
+  def pollOnce(): Unit = {
```
mumrah: Is this just for testing?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
```diff
@@ -164,13 +164,11 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
...
                                       threadName: String)
-  extends InterBrokerSendThread(threadName, networkClient, time, isInterruptible = false) {
+  extends InterBrokerSendThread(threadName, networkClient, config.controllerSocketTimeoutMs, time, isInterruptible = false) {
 
   private var activeController: Option[Node] = None
 
-  override def requestTimeoutMs: Int = config.controllerSocketTimeoutMs
-
-  override def generateRequests(): Iterable[RequestAndCompletionHandler] = {
+  def generateRequests(): Iterable[RequestAndCompletionHandler] = {
```
mumrah: Not needed here, but I wonder if we should just make this return an `Option[RequestAndCompletionHandler]`

clients/src/main/java/org/apache/kafka/common/requests/AddOffsetsToTxnRequest.java
hachikuji: Support for optional fields would go a long way I think. I am not sure it will be possible to remove all intermediate representations, but perhaps they can be the exception and not the rule. Some version checks in `KafkaApis` are probably inevitable.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
```diff
@@ -51,22 +53,34 @@ abstract class InterBrokerSendThread(name: String,
...
 
-  override def doWork(): Unit = {
-    var now = time.milliseconds()
+  def sendRequest(request: RequestAndCompletionHandler): Unit = {
+    sendRequests(Seq(request))
+  }
 
-    generateRequests().foreach { request =>
+  def sendRequests(requests: Iterable[RequestAndCompletionHandler]): Unit = {
+    inboundQueue.addAll(requests.asJavaCollection)
```
mumrah: Just a quick note: you had `offer` here before which has different behavior. It probably won't matter though, since the queue is unbounded.

(I only noticed this because I had commented on the `offer` before your latest commit 😄 )

core/src/main/scala/kafka/server/BrokerToControllerChannelManager.scala
```diff
@@ -135,7 +117,16 @@ class BrokerToControllerChannelManagerImpl(metadataCache: kafka.server.MetadataC
...
+   * Send request to the controller.
+   *
+   * @param request         The request to be sent.
+   * @param callback        Request completion callback.
+   * @param retryDeadlineMs The retry deadline which will only be checked after receiving a response.
+   *                        This means that in the worst case, the total timeout would be twice of
+   *                        the configured timeout.
+   */
+  def sendRequest(request: AbstractRequest.Builder[_ <: AbstractRequest],
                            callback: ControllerRequestCompletionHandler,
```
mumrah: nit: indent misaligned

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
hachikuji: That's fair. It's probably fine to compute it just once for `sendRequests`.

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
hachikuji: Yeah, let me make that clearer.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
hachikuji: Yeah, I was going to change this to use `add` anyway to make the expectation clearer.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
hachikuji: Haha, yeah, exactly what I was thinking.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
hachikuji: The `BrokerToControllerChannelManager` is probably the only case where this might be a concern since requests can be forwarded form the client. We wouldn't want to block the request handler if the queue gets too large, but we might want to start failing inbound requests. I'd suggest we create a separate JIRA to think about this. Sound fair?

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
hachikuji: Can you clarify what the option would achieve over an empty collection?

raft/src/main/java/org/apache/kafka/raft/RaftMessageQueue.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * See {@link org.apache.kafka.raft.internals.BlockingMessageQueue}.
+ */
+public interface RaftMessageQueue {
+
+    /**
+     * Block for the arrival of a new message.
+     *
+     * @param timeoutMs timeout in milliseconds to wait for a new event
+     * @return the event or null if the timeout was reached
```
mumrah: nit: should add something like "timeout was reached or #wakeup() was called"

raft/src/main/java/org/apache/kafka/raft/internals/BlockingMessageQueue.java
```diff
@@ -0,0 +1,78 @@
...
+    public boolean isEmpty() {
+        return size.get() == 0;
+    }
+
+    @Override
+    public void wakeup() {
+        queue.add(Wakeup.INSTANCE);
+    }
+
+    public interface RaftEvent {
```
mumrah: Could you use a sentinel RaftMessage object here instead? Might simplify this class a bit. Not a big deal either way

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
mumrah: I was just noticing that we only ever return a Queue with a single item. Optional seems to fit that case more cleanly. However (as we discussed offline), I wonder why we have the code like this in the first place. Maybe @abbccdda has some insight?

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
mumrah: Ah, didn't see your earlier comment 👍

raft/src/main/java/org/apache/kafka/raft/internals/BlockingMessageQueue.java
hachikuji: That's a good idea.

core/src/main/scala/kafka/server/BrokerToControllerChannelManagerImpl.scala
hachikuji: Yeah, I agree that looks a bit odd. It looks like it was intended to be a loop.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
hachikuji: I ended up moving creation time to RequestAndCompletionHandler.

clients/src/main/java/org/apache/kafka/common/requests/AddOffsetsToTxnRequest.java
ijuma: I can see this:

1. AbstractRequest/AbstractResponse methods become part of the ApiMessage hierarchy.
2. FooRequest/FooResponse extends FooDataRequest/FooDataResponse (like Colin suggested before)

But I don't think you want to eliminate FooRequest/FooResponse in the example above. You don't need to perform conversions for the inner classes, but it's a place where you can normalize the representation. We do that for many of the existing request/response classes.

abbccdda:
Just to reply https://github.com/apache/kafka/pull/9732#discussion_r541192299 here, `generateRequests()` is part of the InterBrokerSender thread which I didn't write up. In the meantime, I think changing it to one element at a time makes sense for AlterIsr and forwarding cases for now, if we believe that would make the interface easier to be used, let's refactor it.

hachikuji:
@abbccdda To clarify, what we are asking is why the implementation of `generateRequests` in `BrokerToControllerChannelManager` builds a queue for a single item.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
mumrah: Yea, sounds good 👍

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
```diff
@@ -53,6 +54,41 @@ object KafkaNetworkChannel {
 
 }
 
+private[raft] class RaftSendThread(
```
mumrah, line 57: style nit/question: I think we have a mixture of argument indentation for Scala classes/methods. Do we have an established style convention for this? Normally I follow the opening paren when breaking out arguments into their own line (though I'm not sure that's correct)

core/src/main/scala/kafka/server/BrokerToControllerChannelManager.scala
```diff
@@ -149,29 +152,32 @@ case class BrokerToControllerQueueItem(request: AbstractRequest.Builder[_ <: Abs
...
-      val request = RequestAndCompletionHandler(
+  def enqueue(request: BrokerToControllerQueueItem): Unit = {
+    requestQueue.add(request)
+    if (activeController.isDefined) {
+      wakeup()
+    }
+  }
+
+  override def generateRequests(): Iterable[RequestAndCompletionHandler] = {
+    Option(requestQueue.poll()).map { queueItem =>
```
mumrah: 👍

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
```diff
@@ -32,17 +32,19 @@ import scala.jdk.CollectionConverters._
...
+) extends ShutdownableThread(name, isInterruptible) {
 
-  def generateRequests(): Iterable[RequestAndCompletionHandler]
-  def requestTimeoutMs: Int
   private val unsentRequests = new UnsentRequests
 
-  def hasUnsentRequests = unsentRequests.iterator().hasNext
+  def generateRequests(): Iterable[RequestAndCompletionHandler]
+
+  def hasUnsentRequests: Boolean = unsentRequests.iterator().hasNext
```
mumrah, line 47: Is it worth adding a `size` or `isEmpty` to `UnsentRequests`?

core/src/main/scala/kafka/coordinator/transaction/TransactionMarkerChannelManager.scala
```diff
@@ -127,11 +127,14 @@ class TxnMarkerQueue(@volatile var destination: Node) {
   def totalNumMarkers(txnTopicPartition: Int): Int = markersPerTxnTopicPartition.get(txnTopicPartition).fold(0)(_.size)
 }
 
-class TransactionMarkerChannelManager(config: KafkaConfig,
-                                      metadataCache: MetadataCache,
-                                      networkClient: NetworkClient,
-                                      txnStateManager: TransactionStateManager,
-                                      time: Time) extends InterBrokerSendThread("TxnMarkerSenderThread-" + config.brokerId, networkClient, time) with Logging with KafkaMetricsGroup {
+class TransactionMarkerChannelManager(
```
mumrah, line 130: nit: (related to style question elsewhere) if we want to change the style of these class definitions, can we do it as a separate PR? I always find it difficult when style changes are conflated with logical changes

core/src/main/scala/kafka/server/BrokerToControllerChannelManager.scala
```diff
@@ -221,7 +216,7 @@ class BrokerToControllerRequestThread(networkClient: KafkaClient,
       } else {
         // need to backoff to avoid tight loops
         debug("No controller defined in metadata cache, retrying after backoff")
-        backoff()
```
mumrah, line 224: Hmm, this seems strange, though maybe I'm missing something.

If we get here, `activeController` is not defined. If we then pollOnce it looks like the request produced by `generateRequests` will get an exception since the activeController Option is empty. 

Previously, the backoff method was just pausing the thread for some time.

raft/src/main/java/org/apache/kafka/raft/RaftMessageQueue.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    RaftMessage poll(long timeoutMs);
+
+    /**
+     * Offer a new message to the queue.
+     *
+     * @param message the message to deliver
+     * @throws IllegalStateException if the queue cannot accept the message
+     */
+    void offer(RaftMessage message);
```
mumrah: nit: Maybe name this `add` so it aligns with the java.util.Queue method?

raft/src/main/java/org/apache/kafka/raft/internals/BlockingMessageQueue.java
```diff
@@ -0,0 +1,76 @@
...
+import org.apache.kafka.raft.RaftMessage;
+import org.apache.kafka.raft.RaftMessageQueue;
+
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class BlockingMessageQueue implements RaftMessageQueue {
+    private static final RaftMessage WAKEUP_MESSAGE = new RaftMessage() {
```
mumrah, line 30: 👍

core/src/main/scala/kafka/server/BrokerToControllerChannelManager.scala
mumrah: Also, if we don't need `backoff` anymore, we can remove it since this was the only usage

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
hachikuji: Not really as far as I know. This is my favored style of late because it results in consistent alignment of parameters. I get really annoyed when I see stuff like this:
```
def someMethodWithAnArguablyOverVerboseName(foo: String,
                                            bar: Option[List[Int]]): (Option[String], List[String)

def concicseMethod(foo: String,
                   bar: String): Unit = {
```
No matter how you look at it, it seems hideous.

core/src/main/scala/kafka/coordinator/transaction/TransactionMarkerChannelManager.scala
hachikuji: That's fair. I actually held myself back. I tried to only touch the cases that I was modifying anyway, but let me know if there are others. This one was especially obnoxious because of the long parameter list to `InterBrokerSendThread`.

core/src/main/scala/kafka/server/BrokerToControllerChannelManager.scala
hachikuji: Good catch. Will fix. Let me check on test cases.

hachikuji: While addressing this issue, I realized the current timeout logic does not handle the case when the controller is not known. This will cause the requests to keep piling up until we find the controller.

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -34,25 +35,58 @@ import scala.collection.mutable
...
+  def apply(
+    config: KafkaConfig,
+    metadataCache: MetadataCache,
+    scheduler: KafkaScheduler,
+    time: Time,
+    metrics: Metrics,
+    threadNamePrefix: Option[String],
+    brokerEpochSupplier: () => Long
+  ): AlterIsrManager = {
+    val channelManager = new BrokerToControllerChannelManager(
```
mumrah, line 76: I guess we're not sharing these channel managers between anything. In that case, moving them into the classes that need them seems fine.

core/src/main/scala/kafka/server/BrokerToControllerChannelManager.scala
```diff
@@ -152,76 +151,89 @@ abstract class ControllerRequestCompletionHandler extends RequestCompletionHandl
...
-        handleResponse(topRequest)
-      )
+  def queueSize: Int = {
+    requestQueue.size
+  }
 
-      requestsToSend.enqueue(request)
+  override def generateRequests(): Iterable[RequestAndCompletionHandler] = {
+    val currentTimeMs = time.milliseconds()
+    val requestIter = requestQueue.iterator()
```
mumrah, line 187: Any reason to use Iterator here instead of queue methods (i.e., peek and remove). Is it to ensure a consistent view of the queue while we're going through it?

mumrah (approved):

core/src/main/scala/kafka/server/BrokerToControllerChannelManager.scala
hachikuji: Yeah, I just thought it was a little simpler.

PR #9735, opened 2020-12-11
KAFKA-10846: Grow buffer in FileSourceStreamTask only when needed

tombentley:
See [KAFKA-10846](https://issues.apache.org/jira/browse/KAFKA-10846)

### DIFF
--- a/connect/file/src/test/java/org/apache/kafka/connect/file/FileStreamSourceTaskTest.java
+++ b/connect/file/src/test/java/org/apache/kafka/connect/file/FileStreamSourceTaskTest.java
@@ -30,6 +30,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.file.Files;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


tombentley:
@C0urante @kkonstantine maybe one of you could take a look.

chia7712:
@tombentley Thanks for this patch. Could you offer a test?

tombentley:
Thanks for taking a look @chia7712. I've added a test, but had to expose the buffer size.

chia7712 (commented):
@tombentley Thanks for the test. Pardon me, which one covers the 'foundOneLine=true'?

tombentley:
@chia7712 I think `testNormalLifecycleWithResize()` covers that when a buffer resize is required, and `testNormalLifecycle()` in the case where a buffer resize is not required. Or did I miss a case?

chia7712:
>  I think testNormalLifecycleWithResize() covers that when a buffer resize is required, and testNormalLifecycle() in the case where a buffer resize is not required. Or did I miss a case?

Sorry that I neglect ```testNormalLifecycleWithResize```. It is a little hard to review code by iPhone. Will take another look later.

chia7712 (approved):
@tombentley Thanks for your patch. LGTM

connect/file/src/test/java/org/apache/kafka/connect/file/FileStreamSourceTaskTest.java
```diff
@@ -77,6 +78,17 @@ private void replay() {
 
     @Test
     public void testNormalLifecycle() throws InterruptedException, IOException {
+        normalLifecycle();
+    }
+
+    @Test
+    public void testNormalLifecycleWithResize() throws InterruptedException, IOException {
+        task = new FileStreamSourceTask(2);
```
chia7712: This new test case LGTM. However, it re-create the task. How about setting a small initial size of FileStreamSourceTask to make all test cases meet the resize?

chia7712: Or we make each test create task instead of creating task in ```setup```

tombentley:
@chia7712 I've implemented the first option.

chia7712:
@tombentley Thanks for this contribution.

tombentley:
@chia7712 thank _you_ for the reviews.

PR #9737, opened 2020-12-12
KAFKA-10828: Replacing endorsing with acknowledging for voters 

vamossagar12:
This PR replaces the terms endorsing with acknowledging for voters which have recognised the current leader.

### DIFF
--- a/raft/src/main/java/org/apache/kafka/raft/LeaderState.java
+++ b/raft/src/main/java/org/apache/kafka/raft/LeaderState.java
@@ -30,7 +30,7 @@ import java.util.stream.Collectors;
  * In the context of LeaderState, an acknowledged voter means one who has acknowledged the current leader by either
  * responding to a `BeginQuorumEpoch` request from the leader or by beginning to send `Fetch` requests.
  * More specifically, the set of unacknowledged voters are targets for BeginQuorumEpoch requests from the leader until
- * it doesn't receive an acknowledgement from them.
+ * they acknowledge the leader.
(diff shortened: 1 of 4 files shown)


raft/src/main/java/org/apache/kafka/raft/LeaderState.java
```diff
@@ -50,8 +50,8 @@ protected LeaderState(
         this.highWatermark = Optional.empty();
 
         for (int voterId : voters) {
-            boolean hasEndorsedLeader = voterId == localId;
-            this.voterReplicaStates.put(voterId, new VoterState(voterId, hasEndorsedLeader));
+            boolean hasAcknowledgedLeader = voterId == localId;
```
hachikuji, line 59: Could we add a brief comment in the class javadoc about what we mean by "acknowledged"? It would be helpful to mention specifically that the set of unacknowledged voters are targets for `BeginQuorumEpoch` requests from the leader.

hachikuji (commented):
Thanks for the patch. Just a minor suggestion to improve documentation.

raft/src/main/java/org/apache/kafka/raft/LeaderState.java
vamossagar12: done

vamossagar12:
hey @hachikuji , did you get a chance to review this one?

hachikuji (approved):
LGTM. Thanks for the patch!

PR #9739, opened 2020-12-13
KAFKA-10636 Bypass log validation for writes to raft log

feyman2016:


core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -68,7 +68,7 @@ class KafkaMetadataLog(
 
     val appendInfo = log.appendAsLeader(records.asInstanceOf[MemoryRecords],
       leaderEpoch = epoch,
-      origin = AppendOrigin.Coordinator)
+      origin = AppendOrigin.Replication)
```
feyman2016: I intended to add a new `AppendOrigin` called `Leader` for this case, but then think it implies the same as the Replication, so just reuses it.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1477,6 +1477,7 @@ private void appendBatch(
     ) {
         try {
             int epoch = state.epoch();
+            batch.data.batches().forEach(recordBatch -> recordBatch.setPartitionLeaderEpoch(epoch));
```
feyman2016: This is necessary for the sanity check in `Log.scala append()`
```
         if (batch.magic >= RecordBatch.MAGIC_VALUE_V2) {
              maybeAssignEpochStartOffset(batch.partitionLeaderEpoch, batch.baseOffset)
```

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1050,7 +1050,8 @@ class Log(@volatile private var _dir: File,
                      leaderEpoch: Int,
                      origin: AppendOrigin = AppendOrigin.Client,
                      interBrokerProtocolVersion: ApiVersion = ApiVersion.latestVersion): LogAppendInfo = {
-    append(records, origin, interBrokerProtocolVersion, assignOffsets = true, leaderEpoch, ignoreRecordSize = false)
+    val assignOffsets = if (origin == AppendOrigin.Replication) false else true
```
dengziming: This can be replaced by `val assignOffsets = if (origin != AppendOrigin.Replication)`

feyman2016: @dengziming Thanks, but it needs `else` for a complete `if` expression

core/src/test/scala/unit/kafka/log/LogCleanerTest.scala
```diff
@@ -599,12 +599,12 @@ class LogCleanerTest {
     val log = makeLog(config = LogConfig.fromProps(logConfig.originals, logProps))
 
     val appendFirstTransaction = appendTransactionalAsLeader(log, producerId, producerEpoch,
-      origin = AppendOrigin.Replication)
+      origin = AppendOrigin.Coordinator)
```
feyman2016: Since now for `AppendOrigin.Replication`, the `LogValidator` is bypassed when `Log.appendAsLeader` is called, which has an assumption that the records to be appended should have leader epoch ready in its underlying batches. And moreover, the transaction/group coordinator are supposed to be calling the `Log.appendAsLeader` with `origin = AppendOrigin.Coordinator`.

core/src/main/scala/kafka/log/Log.scala
dengziming: o, I mean `val assignOffsets = origin != AppendOrigin.Replication`

feyman2016:
@hachikuji Request for review, thanks!
BTW,  failed tests `kafka.server.ControllerMutationQuotaTest.testUnboundedDeleteTopicsRequest` and `org.apache.kafka.streams.integration.EosBetaUpgradeIntegrationTest.shouldUpgradeFromEosAlphaToEosBeta[true]` doesn't seem to be related, both have been executed successfully locally.

core/src/main/scala/kafka/log/Log.scala
feyman2016: Got it, will let @hachikuji review and fix it in one shot, thanks

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
hachikuji: I was going to ask about adding a new `AppendOrigin`. I agree that the behavior should be the same as for `Replication`, but it seems like it could lead to confusion. Maybe we could add an `AppendOrigin.RaftLeader`. Then we can add a simple comment which emphasizes that the Raft leader is responsible for assigning offsets. That would also allow us to revert the changes in `LogCleanerTest`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Hmm.. Do we need this? I thought we already set leader epoch here: https://github.com/apache/kafka/blob/trunk/raft/src/main/java/org/apache/kafka/raft/internals/BatchBuilder.java#L256.

core/src/main/scala/kafka/log/Log.scala
hachikuji: By the way, it seems we might want to rename `assignOffsets` since we also rely on this flag for record validation. It's a bit on the verbose side, but maybe `validateAndAssignOffsets` to go along with the similarly named method in `LogValidator`?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1050,7 +1050,8 @@ class Log(@volatile private var _dir: File,
                      leaderEpoch: Int,
                      origin: AppendOrigin = AppendOrigin.Client,
```
hachikuji, line 1079: It would be helpful to have a test case which verifies `appendAsLeader` with the new append origin.

hachikuji (commented):
Thanks @feyman2016 and sorry for the delay. Left a few comments.

feyman2016:
@hachikuji Thanks for the review, I will address the comments shortly~

core/src/main/scala/kafka/log/Log.scala
feyman2016: fixed

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
feyman2016: Added a `AppendOrigin.RaftLeader` and also reverted the changes in `LogCleanerTest `

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
feyman2016: @hachikuji I thought the same as you, re-check the code, the `PARTITION_LEADER_EPOCH_OFFSET` has been set in `DefaultRecordBatch#writeHeader`, but the problem is that the `BatchBuilder` is always constructed with the `epoch=RecordBatch.NO_PARTITION_LEADER_EPOCH`, so I updated it and reverted the L1480.

core/src/main/scala/kafka/log/Log.scala
feyman2016: Make sense, updated

core/src/test/scala/unit/kafka/log/LogCleanerTest.scala
feyman2016: Reverted since we have a newly added `AppendOrigin.RaftLeader`

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -19,7 +19,6 @@
 import org.apache.kafka.common.memory.MemoryPool;
 import org.apache.kafka.common.record.CompressionType;
 import org.apache.kafka.common.record.MemoryRecords;
-import org.apache.kafka.common.record.RecordBatch;
```
feyman2016, line 22: Fx checkstyle error

feyman2016:
@hachikuji @dengziming Addressed the comments, call for a second review, thanks a lot!

feyman2016:
@hachikuji @dengziming Call for a second review, the 7 failed tests run successfully locally, should be unrelated to this PR, thanks!

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -180,7 +179,7 @@ private void startNewBatch() {
                 nextOffset,
                 time.milliseconds(),
                 false,
-                RecordBatch.NO_PARTITION_LEADER_EPOCH,
+                epoch,
```
hachikuji, line 182: Good catch. Do we have any test cases in `BatchAccumulatorTest` which can be modified to catch this case?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1057,7 +1057,8 @@ class Log(@volatile private var _dir: File,
...
                      interBrokerProtocolVersion: ApiVersion = ApiVersion.latestVersion): LogAppendInfo = {
-    append(records, origin, interBrokerProtocolVersion, assignOffsets = true, leaderEpoch, ignoreRecordSize = false)
+    val validateAndAssignOffsets = origin != AppendOrigin.RaftLeader
+    append(records, origin, interBrokerProtocolVersion, validateAndAssignOffsets, leaderEpoch, ignoreRecordSize = false)
```
hachikuji, line 1082: It is a little odd here is that we have to pass through `leaderEpoch` even though we expect the Raft leader to have set it already. Perhaps we should be validating it in `analyzeAndValidateRecords`. We can verify for each batch that the leader epoch matches when the append origin is `RaftLeader`. Does that make sense?

hachikuji (commented):
Thanks for the updates. Just a couple small comments.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
feyman2016: Let me check

core/src/main/scala/kafka/log/Log.scala
feyman2016: Make sense to me, it is also a little bit odd to me, but I put it here because I think `assignOffsets`==true for `appendAsLeader` and ==false for `appendAsFollower`, which means normally `assignOffsets` is determined by the caller, the `RaftLeader` is just a special case for `appendAsLeader`, if we move the logic in `analyzeAndValidateRecords`, that means it need to determine whether to `assignOffsets` without caller info, does that doable?

hachikuji: The call to `analyzeAndValidateRecords` is done both for leader as well as follower appends. Basically we do a shallow iteration over the batches in order to collect some information and validate the CRC. My thought was to pass `leaderEpoch` into `analyzeAndValidateRecords` and add a basic check like this:
```java
records.batches.forEach { batch =>
  ...
  if (origin === RaftLeader && batch.partitionLeaderEpoch != leaderEpoch) {
    throw new InvalidRecordException("Append from Raft leader did not set the batch epoch correctly")
  }
}  
```

hachikuji:
@feyman2016 I can help getting this over the line if you don't have time. I think it's just a couple small changes after rebasing.

feyman2016:
@hachikuji Sorry for the delay, I'm on it now, will update it soon~

feyman2016:
KafkaMetadataLogTest failed because they didn't pass the epoch check when it comes from RaftLeader, will check and update tonight

feyman2016:
Fixed the KafkaMetadataLogTest, let's see if jenkins still complains~

hachikuji (approved):
LGTM. Thanks for the patch!

feyman2016:
@hachikuji Thanks for the review and help!

PR #9742, opened 2020-12-13
KAFKA-10768 Add a test for ByteBufferInputStream to ByteBufferLogInputStreamTest

bertber:
I made a test for ByteBufferInputStream in the ByteBufferLogInputStreamTest.
First, I add a ByteBuffer that it's not empty to the ByteBufferInputStream, in order to verify it.
After that, I try to use ByteBufferInputStream's read function and check return value whether it's correct.


core/src/main/scala/kafka/tools/StreamsResetter.java
```diff
@@ -489,19 +489,7 @@ public void resetOffsetsFromResetPlan(final Consumer<byte[], byte[]> client,
     private void resetByDuration(final Consumer<byte[], byte[]> client,
                                  final Set<TopicPartition> inputTopicPartitions,
                                  final Duration duration) {
-        final Instant now = Instant.now();
```
chia7712: Please don't include other commit code

chia7712 (commented):
@bertber Thanks for your patch. 

Please include the jira number in the title

chia7712 (commented):
please revert the deletion...

Also, the improvement is not addressed. Please take a look at origin PR

PR #9743, opened 2020-12-13
KAFKA-10841 LogReadResult should be able to converted to FetchPartiti…

g1geordie:
I put the conversion logic into LogReadResult named toFetchPartitionData.

### DIFF
--- a/core/src/main/scala/kafka/server/ReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/ReplicaManager.scala
@@ -111,6 +111,17 @@ case class LogReadResult(info: FetchDataInfo,
     case Some(e) => Errors.forException(e)
   }
 
+  def toFetchPartitionData( isReassignmentFetch : Boolean) : FetchPartitionData = FetchPartitionData (
+    this.error,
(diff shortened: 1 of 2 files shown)


g1geordie:
@chia7712  can you help me take a look ?

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -111,6 +111,17 @@ case class LogReadResult(info: FetchDataInfo,
     case Some(e) => Errors.forException(e)
   }
 
+  def toFetchPartitionData( isReassignmentFetch : Boolean) : FetchPartitionData = FetchPartitionData (
```
chia7712: Could you remove those space?

chia7712 (commented):
@g1geordie Thanks for your patch.

core/src/main/scala/kafka/server/ReplicaManager.scala
g1geordie: Thank you

ijuma (approved):
LGTM, thanks.

PR #9746, opened 2020-12-14
MINOR: Replace ApiVersion by auto-generated protocol

chia7712:
This PR includes following changes.

1. rename (auto-generated) ```ApiVersionsResponseKey``` to ```ApiVersion```
2. replace clients/src/main/java/org/apache/kafka/clients/ApiVersion.java by auto-generated ```ApiVersion```


clients/src/main/java/org/apache/kafka/clients/NodeApiVersions.java
```diff
@@ -37,18 +36,18 @@
 public class NodeApiVersions {
 
     // A map of the usable versions of each API, keyed by the ApiKeys instance
-    private final Map<ApiKeys, ApiVersion> supportedVersions = new EnumMap<>(ApiKeys.class);
+    private final Map<ApiKeys, ApiVersionsResponseKey> supportedVersions = new EnumMap<>(ApiKeys.class);
```
ijuma: Should we rename the generated class to have a more concise name? It seems unnecessarily verbose.

chia7712: How about renaming ```ApiVersionsResponseKey``` to ```ApiVersion```?

dajac: Using `ApiVersion` sounds reasonable to me.

ijuma: Sounds good if there are no name conflicts as a result of that change.

chia7712: I have renamed it in the latest commit.

chia7712:
@dajac @ijuma Could you please take a look?

dajac:
@chia7712 Sorry for the delay. I will take a look at it next week.

chia7712:
fix conflicting files

chia7712:
rewrite whole PR due to recent changes ...

clients/src/main/java/org/apache/kafka/clients/NodeApiVersions.java
```diff
@@ -39,18 +38,18 @@
 public class NodeApiVersions {
 
     // A map of the usable versions of each API, keyed by the ApiKeys instance
-    private final Map<ApiKeys, ApiVersion> supportedVersions = new EnumMap<>(ApiKeys.class);
+    private final Map<ApiKeys, ApiVersionsResponseData.ApiVersion> supportedVersions = new EnumMap<>(ApiKeys.class);
```
ijuma: Can we import `ApiVersion` directly? I don't think `ApiVersionsResponseData` adds any value here and other similar places.

clients/src/main/java/org/apache/kafka/clients/NodeApiVersions.java
```diff
@@ -60,18 +59,21 @@ public static NodeApiVersions create() {
...
                     exists = true;
                     break;
                 }
             }
             if (!exists) {
-                apiVersions.add(new ApiVersion(apiKey));
+                apiVersions.add(new ApiVersionsResponseData.ApiVersion()
+                        .setApiKey(apiKey.id)
+                        .setMinVersion(apiKey.oldestVersion())
+                        .setMaxVersion(apiKey.latestVersion()));
```
ijuma: It would be good to have a helper method somewhere that converts `ApiKeys` to `ApiVersion`.

clients/src/main/java/org/apache/kafka/common/requests/ApiVersionsResponse.java
```diff
@@ -220,4 +217,16 @@ private static FinalizedFeatureKeyCollection createFinalizedFeatureKeys(
 
         return converted;
     }
+
+    public static Optional<ApiVersionsResponseData.ApiVersion> intersect(ApiVersionsResponseData.ApiVersion thisVersion,
+                                                                         ApiVersionsResponseData.ApiVersion other) {
+        if (other == null) return Optional.empty();
```
ijuma: It's a bit weird that we only check if `other` is null. Do we need this check at all?

clients/src/test/java/org/apache/kafka/clients/NodeApiVersionsTest.java
```diff
@@ -56,12 +54,18 @@ public void testUnknownApiVersionsToString() {
...
+                versionList.add(new ApiVersionsResponseData.ApiVersion()
+                        .setApiKey(apiKey.id)
+                        .setMinVersion((short) 10000)
+                        .setMaxVersion((short) 10001));
             } else {
-                versionList.add(new ApiVersion(apiKey));
+                versionList.add(new ApiVersionsResponseData.ApiVersion()
+                        .setApiKey(apiKey.id)
+                        .setMinVersion(apiKey.oldestVersion())
+                        .setMaxVersion(apiKey.latestVersion()));
```
ijuma: We can use the helper that converts from ApiKey to ApiVersion here too.

ijuma (commented):
Thanks for the PR. Left a few minor comments.

ijuma:
Also, one file has conflicts that need to be resolved.

clients/src/main/java/org/apache/kafka/common/requests/ApiVersionsResponse.java
chia7712: From the usage of this method, only ```other``` can be null. So current check is enough.

ijuma: That's not a good way to design apis, right? This kind of thing should be symmetric. Why is `other` sometimes `null`?

core/src/test/scala/unit/kafka/server/AbstractApiVersionsRequestTest.scala
```diff
@@ -52,9 +52,9 @@ abstract class AbstractApiVersionsRequestTest extends BaseRequestTest {
     if (listenerName == controlPlaneListenerName) {
       expectedApis.add(ApiKeys.ENVELOPE)
     }
-    assertEquals("API keys in ApiVersionsResponse must match API keys supported by broker.",
-      expectedApis.size(), apiVersionsResponse.data.apiKeys().size())
-    for (expectedApiVersion: ApiVersionsResponseData.ApiVersion <- ApiVersionsResponse.DEFAULT_API_VERSIONS_RESPONSE.data.apiKeys().asScala) {
+    assertEquals(expectedApis.size(), apiVersionsResponse.data.apiKeys().size(),
+      "API keys in ApiVersionsResponse must match API keys supported by broker.")
+      for (expectedApiVersion: ApiVersionsResponseData.ApiVersion <- ApiVersionsResponse.DEFAULT_API_VERSIONS_RESPONSE.data.apiKeys().asScala) {
```
ijuma: Indent change here seems unintentional.

clients/src/main/java/org/apache/kafka/common/requests/ApiVersionsResponse.java
chia7712: > Why is other sometimes null?

The api key is supported by client-side but Controller doesn't.

> That's not a good way to design apis, right? This kind of thing should be symmetric. 

make sense. will address it and add unit tests.

ijuma (approved):
Thanks for the updates, LGTM.

PR #9747, opened 2020-12-14
KAFKA-10852: AlterIsr should not be throttled

ijuma:
Set it as a cluster action and update the handler in KafkaApis. We keep the throttleTimeMs field
since we intend to enable throttling in the future (especially relevant when we switch to the
built-in quorum mode).

### DIFF
--- a/clients/src/test/java/org/apache/kafka/common/protocol/ApiKeysTest.java
+++ b/clients/src/test/java/org/apache/kafka/common/protocol/ApiKeysTest.java
@@ -20,11 +20,12 @@ import org.apache.kafka.common.protocol.types.BoundField;
 import org.apache.kafka.common.protocol.types.Schema;
 import org.junit.Test;
 
-import java.util.Arrays;
-import java.util.List;
+import java.util.EnumSet;
+import java.util.Set;
 
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class ApiKeysTest {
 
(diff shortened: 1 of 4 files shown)


clients/src/test/java/org/apache/kafka/common/protocol/ApiKeysTest.java
```diff
@@ -55,11 +61,14 @@ public void schemaVersionOutOfRange() {
      */
     @Test
     public void testResponseThrottleTime() {
-        List<ApiKeys> authenticationKeys = Arrays.asList(ApiKeys.SASL_HANDSHAKE, ApiKeys.SASL_AUTHENTICATE);
+        Set<ApiKeys> authenticationKeys = EnumSet.of(ApiKeys.SASL_HANDSHAKE, ApiKeys.SASL_AUTHENTICATE);
+        // Newer protocol apis include throttle time ms even for cluster actions
+        Set<ApiKeys> clusterActionsWithThrottleTimeMs = EnumSet.of(ApiKeys.ALTER_ISR);
```
ijuma, line 66: We have to decide whether we want to do this or remove the throttle time ms from AlterIsr before 2.7.0.

ijuma: cc @bbejeck

mumrah (approved):
Thanks @ijuma! LGTM

clients/src/test/java/org/apache/kafka/common/protocol/ApiKeysTest.java
ijuma: We decided this is not a blocker.

ijuma:
@mumrah A couple of the test failures indicated that we had to change some code in KafkaApis too. Updated.

mumrah:
@ijuma, thanks! So, we'll keep the throttle field in the RPC, but we won't actually set the throttle. Seems reasonable 👍

ijuma:
Unrelated flaky test failures:

> Build / JDK 11 / kafka.server.ScramServerStartupTest.testAuthentications
> Build / JDK 15 / org.apache.kafka.streams.integration.StandbyTaskEOSIntegrationTest.shouldWipeOutStandbyStateDirectoryIfCheckpointIsMissing[exactly_once]
>

ijuma:
Merged to trunk and 2.7 branches.

PR #9748, opened 2020-12-14
MINOR: Simplify ApiKeys by relying on ApiMessageType

ijuma:
* The naming for `ListOffsets` was inconsistent, in some places it was `ListOffset` and in others
it was `ListOffsets`. Picked the latter since it was used in metrics and the protocol documentation
and made it consistent.
* Removed unused methods in ApiKeys.
* Deleted `CommonFields`.
* Added `lowestSupportedVersion` and `highestSupportedVersion` to `ApiMessageType`
* Removed tests in `MessageTest` that are no longer relevant.


generator/src/main/java/org/apache/kafka/message/ApiMessageTypeGenerator.java
```diff
@@ -169,7 +169,7 @@ private void generateEnumValues() {
     }
 
     private void generateInstanceVariables() {
-        buffer.printf("private final String name;%n");
+        buffer.printf("public final String name;%n");
```
chia7712, line 178: It is a bit weird that ```messageType.name``` and ```messageType.name()``` return different string.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -343,9 +174,15 @@ public Struct parseResponse(short version, ByteBuffer buffer) {
                 throw new IllegalStateException("Response schema for api " + name + " for version " + i + " is null");
         }
 
+        this.messageType = messageType;
+        this.id = id;
+        this.name = name;
```
chia7712: Could we fix the comment about "name" as it is used to be a part of metrics name (see https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/network/RequestChannel.scala#L239). The origin comment is stale.

clients/src/test/java/org/apache/kafka/common/message/MessageTest.java
```diff
@@ -952,40 +904,6 @@ public String toString() {
...
-                    entryB, entryB.type.isNullable() ? "nullable" : "non-nullable"));
-            }
-            if (entryA.type.isArray()) {
-                compareTypes(new NamedType(entryA.name, entryA.type.arrayElementType().get()),
-                             new NamedType(entryB.name, entryB.type.arrayElementType().get()));
-            }
-        }
-    }
-
     /**
```
chia7712: this method is never used.

chia7712 (commented):
@ijuma Thanks for this nice cleanup.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -289,49 +129,40 @@ public Struct parseResponse(short version, ByteBuffer buffer) {
...
         boolean clusterAction,
         byte minRequiredInterBrokerMagic,
-        boolean isEnabled,
-        Schema[] requestSchemas,
-        Schema[] responseSchemas,
-        boolean forwardable
+        boolean forwardable,
+        boolean isEnabled
     ) {
+        short id = messageType.apiKey();
```
chia7712: It seems to me this check should be moved to ```ApiMessageTypeGenerator``` as it is a generated code from json.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -289,49 +129,40 @@ public Struct parseResponse(short version, ByteBuffer buffer) {
...
-        this.id = (short) id;
-        this.name = name;
-        this.clusterAction = clusterAction;
-        this.minRequiredInterBrokerMagic = minRequiredInterBrokerMagic;
-        this.isEnabled = isEnabled;
 
+        Schema[] requestSchemas = messageType.requestSchemas();
+        Schema[] responseSchemas = messageType.responseSchemas();
+        String name = messageType.name;
         if (requestSchemas.length != responseSchemas.length)
```
chia7712: ditto

chia7712 (commented):
As the "id" is removed from the construction, is it possible to add new ApiKeys instances with incorrect order  in the future? the static field ```ID_TO_TYPE``` assume all ApiKeys instances are created by id order.

generator/src/main/java/org/apache/kafka/message/ApiMessageTypeGenerator.java
ijuma: Yeah, that's an existing problem with ApiKeys too. The enum method cannot be overridden. I think it should be tackled separately.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
ijuma: We can, but it doesn't seem particularly beneficial. It's important to generate code that is specific to each data class. Generic code is best written without generation.

We can move this code to a base class for ApiMessageType potentially, particularly if we decide to remove ApiKeys.

chia7712: >  it doesn't seem particularly beneficial

It seems to me the benefit is the error happens from runtime/test_runtime to build time (generate message code).

ijuma: Oh, I see what you mean. You're suggesting to validate during generation, not to include the validation in the generated code.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
ijuma: Good catch.

ijuma:
> As the "id" is removed from the construction, is it possible to add new ApiKeys instances with incorrect order in the future? the static field ID_TO_TYPE assume all ApiKeys instances are created by id order.

@chia7712 Can you please elaborate? I don't see the assumption you mention. The main assumption is that api keys are dense. I changed the code not to rely on the latter although it doesn't change the behavior.

chia7712:
> Can you please elaborate? I don't see the assumption you mention. The main assumption is that api keys are dense. I changed the code not to rely on the latter although it doesn't change the behavior.

I assumed the order of instancing ApiKeys is strict so removing the id may makes it hard to "see" where to put the new ApiKeys in the future. However, it seems the order of instancing ApiKeys is free so the worry is invalid.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
ijuma: I took a look at this and not clear how to validate this in the generator. I added unit tests to `ApiMessageTypeTest` instead and I think that's good enough for now. What do you think?

chia7712: That alternative looks good to me :)

ijuma:
Tests passed for JDK 11 and 15. JDK 8 had one flaky failure:

> kafka.api.ConsumerBounceTest.testClose failed

@chia7712 Do you have cycles to review this?

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -157,124 +38,76 @@
...
     static {
-        int maxKey = -1;
-        for (ApiKeys key : ApiKeys.values())
-            maxKey = Math.max(maxKey, key.id);
-        ApiKeys[] idToType = new ApiKeys[maxKey + 1];
         for (ApiKeys key : ApiKeys.values())
-            idToType[key.id] = key;
-        ID_TO_TYPE = idToType;
-        MAX_API_KEY = maxKey;
+            ID_TO_TYPE.put((int) key.id, key);
```
chia7712: Does it need to check duplicate id?

chia7712 (approved):
@ijuma +1 and thanks for this big cleanup!

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -361,69 +172,35 @@ private static boolean shouldRetainsBufferReference(Schema[] requestSchemas) {
     }
 
     public static ApiKeys forId(int id) {
```
chia7712, line 172: Is 'short' type more suitable?

chia7712: If the above comment is valid, it can be addressed in separate PR :)

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
ijuma: I think the generator checks for this. I will double check.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
ijuma: We used `int` here because `short` in Java is a second class citizen and casting is required in some cases. If we decide to change this, a separate PR (as you suggested) would be better.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
ijuma: Yes, the generator checks it:

> Caused by: java.lang.RuntimeException: Found more than one response with API key 3

There's also `ApiMessageTypeTest.testUniqueness` that verifies it after the generation.

I added a comment explaining that uniqueness is ensured.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -157,124 +38,78 @@
...
+    ALTER_USER_SCRAM_CREDENTIALS(ApiMessageType.ALTER_USER_SCRAM_CREDENTIALS, false, true),
+    VOTE(ApiMessageType.VOTE, true, RecordBatch.MAGIC_VALUE_V0, false, false),
+    BEGIN_QUORUM_EPOCH(ApiMessageType.BEGIN_QUORUM_EPOCH, true, RecordBatch.MAGIC_VALUE_V0, false, false),
+    END_QUORUM_EPOCH(ApiMessageType.END_QUORUM_EPOCH, true, RecordBatch.MAGIC_VALUE_V0, false, false),
+    DESCRIBE_QUORUM(ApiMessageType.DESCRIBE_QUORUM, true, RecordBatch.MAGIC_VALUE_V0, false, false),
+    ALTER_ISR(ApiMessageType.ALTER_ISR, true),
+    UPDATE_FEATURES(ApiMessageType.UPDATE_FEATURES, false, true),
+    ENVELOPE(ApiMessageType.ENVELOPE, true, RecordBatch.MAGIC_VALUE_V0, false, false);
+
+    private static final Map<Integer, ApiKeys> ID_TO_TYPE = new HashMap<>();
```
chia7712: It seems to me this initialization can be rewrite by following code.
```
    private static final Map<Integer, ApiKeys> ID_TO_TYPE = Arrays.stream(ApiKeys.values())
            .collect(Collectors.toMap(key -> (int) key.id, Function.identity()));
```

The benefit is that he static block can be removed.

ijuma: Updated.

chia7712 (approved):
+1 to latest commit.

ijuma:
JDK8 and 11 passed, 15 had one flaky unrelated failure:

> kafka.server.ClientQuotasRequestTest.testAlterIpQuotasRequest

PR #9749, opened 2020-12-14
Only schedule AlterIsr thread when we have an ISR change

mumrah:
Rather than scheduling every 50ms to check for unsent updates, we should schedule the propagation thread only after we receive ISR updates


core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -62,27 +61,38 @@ class AlterIsrManagerImpl(val controllerChannelManager: BrokerToControllerChanne
...
-    scheduler.schedule("send-alter-isr", propagateIsrChanges, 50, 50, TimeUnit.MILLISECONDS)
-  }
+  override def start(): Unit = { }
 
   override def enqueue(alterIsrItem: AlterIsrItem): Boolean = {
-    unsentIsrUpdates.putIfAbsent(alterIsrItem.topicPartition, alterIsrItem) == null
+    if (unsentIsrUpdates.putIfAbsent(alterIsrItem.topicPartition, alterIsrItem) == null) {
+      if (inflightRequest.compareAndSet(false, true)) {
+        // optimistically set the inflight flag even though we haven't sent the request yet
+        scheduler.schedule("send-alter-isr", propagateIsrChanges, 50, -1, TimeUnit.MILLISECONDS)
```
mumrah: The 50ms delay here acts like a simplistic linger time to allow ISR updates to batch. If we schedule the thread immediately after the first call to `enqueue`, we'll likely only send a single partition in the AlterIsr request. With the small delay, we give partitions a chance to accumulate for sending as one batch.

We might consider adding a real linger time with maximum delay if we find that we're sending inefficient batches in practice.

hachikuji: The linger attempt feels a tad clumsy. I wonder if it is necessary. With the constraint of only having one inflight request, I think we will end up batching effectively when the rate of requests goes up. Most importantly, we'll be able to batch effectively when the controller is slow to respond. Would this be complex to implement?

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -122,43 +121,47 @@ class DefaultAlterIsrManager(
...
   override def shutdown(): Unit = {
     controllerChannelManager.shutdown()
   }
 
   override def submit(alterIsrItem: AlterIsrItem): Boolean = {
-    unsentIsrUpdates.putIfAbsent(alterIsrItem.topicPartition, alterIsrItem) == null
+    if (unsentIsrUpdates.putIfAbsent(alterIsrItem.topicPartition, alterIsrItem) == null) {
+      if (inflightRequest.compareAndSet(false, true)) {
+        // optimistically set the inflight flag even though we haven't sent the request yet
+        scheduler.schedule("send-alter-isr", propagateIsrChanges, 1, -1, TimeUnit.MILLISECONDS)
```
hachikuji: If we are only waiting 1ms, would it be simpler to call `propagateIsrChanges` directly? Similarly after receiving a response with no error.

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -122,43 +121,47 @@ class DefaultAlterIsrManager(
...
   }
 
   override def shutdown(): Unit = {
     controllerChannelManager.shutdown()
   }
 
   override def submit(alterIsrItem: AlterIsrItem): Boolean = {
-    unsentIsrUpdates.putIfAbsent(alterIsrItem.topicPartition, alterIsrItem) == null
+    if (unsentIsrUpdates.putIfAbsent(alterIsrItem.topicPartition, alterIsrItem) == null) {
+      if (inflightRequest.compareAndSet(false, true)) {
```
hachikuji: This seems to introduce a race condition. Say we have an inflight request. Is the following sequence possible?

1. the response returns and the io thread calls `propagateIsrChanges` and sees an empty `unsentIsrUpdates`
2. request thread calls submit and inserts a new item in `unsentIsrUpdates`
3. request thread fails `compareAndSet` on `inflightRequest`
4. io thread clears `inflightRequest`

It seems like we might need a lock.

core/src/main/scala/kafka/server/AlterIsrManager.scala
mumrah: Yea I don't see any problem with that

core/src/main/scala/kafka/server/AlterIsrManager.scala
mumrah: You're right, there's a race between the isEmpty check and clearing the inflight flag. Good catch

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -116,63 +116,78 @@ class DefaultAlterIsrManager(
...
-  private def propagateIsrChanges(): Unit = {
-    if (!unsentIsrUpdates.isEmpty && inflightRequest.compareAndSet(false, true)) {
-      // Copy current unsent ISRs but don't remove from the map
-      val inflightAlterIsrItems = new ListBuffer[AlterIsrItem]()
-      unsentIsrUpdates.values().forEach(item => inflightAlterIsrItems.append(item))
-
-      val now = time.milliseconds()
-      lastIsrPropagationMs.set(now)
-      sendRequest(inflightAlterIsrItems.toSeq)
+  private def propagateIsrChanges(checkInflight: Boolean): Unit = inWriteLock(inflightLock) {
```
hachikuji: Do you think contention for this lock will be an issue? It seems unlikely to me. That makes me think we might be able to simplify the concurrency if we replace the read-write lock with a simple ReentrantLock. Then we could design this around two methods

```scala
def maybePropagateIsrChanges(): Unit = lock synchronized {
 if (!inflightRequest) {
   ... 
 }
}

def clearInFlightRequest(): Unit = lock synchronized {
  inFlightRequest = false
}
```
What do you think?

mumrah: I was thinking of the case when lots of partitions come up at once, like in a rolling restart scenario. However, not much happens in the submit call, so it's probably negligible. I tried this change out this morning and there's no noticeable difference. I'll polish it up and fix the tests.

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -116,63 +116,75 @@ class DefaultAlterIsrManager(
...
-            val body = response.responseBody().asInstanceOf[AlterIsrResponse]
-            handleAlterIsrResponse(body, message.brokerEpoch, inflightAlterIsrItems)
-          } finally {
-            clearInflightRequests()
+          debug(s"Received AlterIsr response $response")
+          val body = response.responseBody().asInstanceOf[AlterIsrResponse]
+          handleAlterIsrResponse(body, message.brokerEpoch, inflightAlterIsrItems) match {
+            case Errors.NONE =>
+              // In the normal case, check for pending updates to send immediately
+              clearInFlightRequest()
```
hachikuji: nit: shall we pull this out of the match since it is done regardless? It would also make the code a little more resilient since we got rid of the `try/catch` to put it before `handleAlterIsrResponse`.

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -116,63 +116,75 @@ class DefaultAlterIsrManager(
   private val unsentIsrUpdates: util.Map[TopicPartition, AlterIsrItem] = new ConcurrentHashMap[TopicPartition, AlterIsrItem]()
 
   // Used to allow only one in-flight request at a time
-  private val inflightRequest: AtomicBoolean = new AtomicBoolean(false)
+  @volatile
```
hachikuji: Do we still need this? It looks like all accesses are protected with the lock.

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -116,63 +116,75 @@ class DefaultAlterIsrManager(
...
         override def onComplete(response: ClientResponse): Unit = {
-          try {
-            debug(s"Received AlterIsr response $response")
-            val body = response.responseBody().asInstanceOf[AlterIsrResponse]
-            handleAlterIsrResponse(body, message.brokerEpoch, inflightAlterIsrItems)
-          } finally {
-            clearInflightRequests()
+          debug(s"Received AlterIsr response $response")
+          val body = response.responseBody().asInstanceOf[AlterIsrResponse]
+          handleAlterIsrResponse(body, message.brokerEpoch, inflightAlterIsrItems) match {
```
hachikuji: Just doublechecking our locking order. When we call `submit` from `Partition`, we first have the leader and ISR write lock and then we acquire the inflight lock added here. Now when we call `handleAlterIsrResponse`, we may need to reacquire the leader and ISR write lock, but that is ok, because do not need to hold the inflight lock when we do so. I think it might be worth adding some comments on the locking order somewhere in this class since the use of the leader and ISR lock is kind of hidden.

hachikuji (commented):
Thanks, left a few more comments. After looking at this again, I can't help but feel that the inflight lock is overkill. I guess what's messing is up is the fact that we need to check both `inflightRequest` as well as `unsentUpdates`. Looking at the initial ordering, we have the following:

- `submit`: first add to queue, then check inflight request
- `onResponse`: first check queue, then clear inflight request

The race condition we were trying to protect occurs because we might enqueue a new item between the two operations in `onResponse`. But maybe we just needed to change the order. In other words, what if we first cleared the inflight flag, then we check the queue?

```scala
def submit(item): Unit = {
  unsentItems.putIfAbsent(item)
  maybePropagate()
}

def maybePropagate(): Unit = {
  if (!unsentItems.isEmpty() && inflightRequest.set(false, true)) {
   sendRequest()
  }
}

def onResponse(response): Unit = {
  inflightRequest.set(false)
  maybePropagate()
}
```
So after enqueuing the item in `submit`, we have an opportunity to send immediately if nothing is inflight. On the other hand, if a request is inflight, then we are guaranteed that `onResponse` will see the addition to the queue because we know it checks the queue only after clearing the flag.

Does that work?

mumrah:
It seems I over-complicated this while trying to avoid clearing the inflight in the response handler when another request needs to go out. I think what you suggest does work. If the response handler just immediately sets the inflight flag to false it doesn't matter if another `submit` call races with checking the queue size since one of them will set the inflight flag true and the request will get submitted. 

I'll go ahead and try this out

core/src/main/scala/kafka/server/AlterIsrManager.scala
```diff
@@ -112,68 +111,74 @@ class DefaultAlterIsrManager(
...
-  private def sendRequest(inflightAlterIsrItems: Seq[AlterIsrItem]): Unit = {
-    val message = buildRequest(inflightAlterIsrItems)
-
-    def clearInflightRequests(): Unit = {
-      // Be sure to clear the in-flight flag to allow future AlterIsr requests
-      if (!inflightRequest.compareAndSet(true, false)) {
-        throw new IllegalStateException("AlterIsr response callback called when no requests were in flight")
-      }
+  private[server] def clearInFlightRequest(): Unit = {
+    if(!inflightRequest.compareAndSet(true, false)) {
```
hachikuji: nit: space after `if`

hachikuji (approved):
LGTM. Trivial nit

PR #9753, opened 2020-12-15
KAFKA-10855; Fix non-local return in `KafkaApis.handle`

hachikuji:
The non-local return when `maybeHandleInvalidEnvelope` returns true causes the default error handler to execute after a response has already been sent. This patch rewrites the check as a local return.


chia7712 (approved):
@hachikuji Nice catching. As 'return' in scala is error-prone, I prefer to use 'if-else' to replace 'return'

ijuma:
Yeah, I agree that using `return` in Scala is dangerous and should be avoided in most cases.

hachikuji:
I agree returns are dangerous in scala. I'm also not fond of the nesting that often results from avoiding them.

PR #9754, opened 2020-12-15
KAFKA-10856: Convert sticky assignor userData schemas to use generated protocol

dengziming:
Replace sticky assignor userData schemas to use generated protocol

Unit test

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java
@@ -25,11 +25,9 @@ import java.util.Optional;
 import java.util.Set;
 import org.apache.kafka.clients.consumer.internals.AbstractStickyAssignor;
 import org.apache.kafka.common.TopicPartition;
-import org.apache.kafka.common.protocol.types.ArrayOf;
-import org.apache.kafka.common.protocol.types.Field;
-import org.apache.kafka.common.protocol.types.Schema;
-import org.apache.kafka.common.protocol.types.Struct;
-import org.apache.kafka.common.protocol.types.Type;
+import org.apache.kafka.common.message.StickyAssignorUserData;
+import org.apache.kafka.common.protocol.ByteBufferAccessor;
+import org.apache.kafka.common.protocol.MessageUtil;
 import org.apache.kafka.common.utils.CollectionUtils;
 
 /**
@@ -175,22 +173,6 @@ import org.apache.kafka.common.utils.CollectionUtils;
  */
 public class StickyAssignor extends AbstractStickyAssignor {
 
-    // these schemas are used for preserving consumer's previously assigned partitions
-    // list and sending it as user data to the leader during a rebalance
-    static final String TOPIC_PARTITIONS_KEY_NAME = "previous_assignment";
-    static final String TOPIC_KEY_NAME = "topic";
-    static final String PARTITIONS_KEY_NAME = "partitions";
-    private static final String GENERATION_KEY_NAME = "generation";
-
-    static final Schema TOPIC_ASSIGNMENT = new Schema(
-        new Field(TOPIC_KEY_NAME, Type.STRING),
-        new Field(PARTITIONS_KEY_NAME, new ArrayOf(Type.INT32)));
-    static final Schema STICKY_ASSIGNOR_USER_DATA_V0 = new Schema(
-        new Field(TOPIC_PARTITIONS_KEY_NAME, new ArrayOf(TOPIC_ASSIGNMENT)));
-    private static final Schema STICKY_ASSIGNOR_USER_DATA_V1 = new Schema(
-        new Field(TOPIC_PARTITIONS_KEY_NAME, new ArrayOf(TOPIC_ASSIGNMENT)),
-        new Field(GENERATION_KEY_NAME, Type.INT32));
-
     private List<TopicPartition> memberAssignment = null;
     private int generation = DEFAULT_GENERATION; // consumer group generation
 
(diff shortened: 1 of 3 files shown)


dengziming:
Hi, @ableegoldman @guozhangwang PTAL.

clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java
```diff
@@ -222,51 +204,50 @@ protected MemberData memberData(Subscription subscription) {
...
-        Struct struct = new Struct(STICKY_ASSIGNOR_USER_DATA_V1);
-        List<Struct> topicAssignments = new ArrayList<>();
+        return serializeTopicPartitionAssignment(memberData, StickyAssignorUserData.HIGHEST_SUPPORTED_VERSION);
+    }
+
+    // visible for testing
+    static ByteBuffer serializeTopicPartitionAssignment(MemberData memberData, short version) {
+
+        List<StickyAssignorUserData.TopicPartition> topicAssignments = new ArrayList<>();
         for (Map.Entry<String, List<Integer>> topicEntry : CollectionUtils.groupPartitionsByTopic(memberData.partitions).entrySet()) {
```
chia7712: Why we need ```CollectionUtils.groupPartitionsByTopic(memberData.partitions)``` here ?

clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java
```diff
@@ -222,51 +204,50 @@ protected MemberData memberData(Subscription subscription) {
...
-            topicAssignment.set(PARTITIONS_KEY_NAME, topicEntry.getValue().toArray());
-            topicAssignments.add(topicAssignment);
+            StickyAssignorUserData.TopicPartition topicPartition = new StickyAssignorUserData.TopicPartition()
+                    .setTopic(topicEntry.getKey())
+                    .setPartitions(topicEntry.getValue());
+            topicAssignments.add(topicPartition);
+        }
+        StickyAssignorUserData data = new StickyAssignorUserData()
+                .setPreviousAssignment(topicAssignments);
+        if (version >= 1) {
```
chia7712: How about using ```"ignorable": true``` to eliminate this if-else?

clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java
```diff
@@ -222,51 +204,50 @@ protected MemberData memberData(Subscription subscription) {
...
+            memberData.generation.ifPresent(data::setGeneration);
         }
-        struct.set(TOPIC_PARTITIONS_KEY_NAME, topicAssignments.toArray());
-        if (memberData.generation.isPresent())
-            struct.set(GENERATION_KEY_NAME, memberData.generation.get());
-        ByteBuffer buffer = ByteBuffer.allocate(STICKY_ASSIGNOR_USER_DATA_V1.sizeOf(struct));
-        STICKY_ASSIGNOR_USER_DATA_V1.write(buffer, struct);
-        buffer.flip();
-        return buffer;
+        return MessageUtil.toVersionPrefixedByteBuffer(version, data);
```
chia7712: Does previous serialization have version field?

clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java
```diff
@@ -222,51 +204,50 @@ protected MemberData memberData(Subscription subscription) {
...
     }
 
-    private static MemberData deserializeTopicPartitionAssignment(ByteBuffer buffer) {
-        Struct struct;
-        ByteBuffer copy = buffer.duplicate();
+    private static MemberData deserializeTopicPartitionAssignment(ByteBuffer buffer, short version) {
+        StickyAssignorUserData data;
         try {
-            struct = STICKY_ASSIGNOR_USER_DATA_V1.read(buffer);
+            data = new StickyAssignorUserData(new ByteBufferAccessor(buffer), version);
```
chia7712: It seems to me previous serialization does not carry version field so we have to use this ugly code to handle different version.

chia7712 (commented):
@dengziming Thanks for your patch. Some major comments are left. Feel free to raise question to me if I make unclear comments.

clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java
dengziming: Because topicAssignments put all partitions of a topic in a field. I think the name `TopicPartition` is misleading and I changed it to `TopicPartitions`, and also the `TopicPartition` in `ConsumerProtocolAssignment.json` and `ConsumerProtocolSubscription.json` is also misleading, we could also create a new pr the alter them.

clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java
dengziming: Yes, the previous ser and de-ser carry no version so the code isn't graceful, I think we needn't keep this backtrack compatibility since this code is only executed at client.

chia7712: > I think we needn't keep this backtrack compatibility since this code is only executed at client.

It seems to me that is not allowed since the different (consumer) clients with different (kafka) versions should be able to work together. Please take a look at https://cwiki.apache.org/confluence/display/KAFKA/KIP-341%3A+Update+Sticky+Assignor%27s+User+Data+Protocol to see how Kafka consider backward/Forward compatibility.

dengziming: @chia7712 Thank you, I removed the version when serialize, and add a loop when deserialize, Do you think that's right.

dengziming: Do you think we need flexible version support here? @chia7712

PR #9756, opened 2020-12-15
KAFKA-10652: Adding size based linger semnatics to Raft metadata

vamossagar12:
This PR aims to add leader fsync occur with size based heuristic. Meaning, if we accumulate a configurable N bytes, then we should not wait for linger expiration and should just fsync immediately.

### DIFF
--- a/raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
+++ b/raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
@@ -81,6 +81,11 @@ public class RaftConfig {
         "wait for writes to accumulate before flushing them to disk.";
     public static final int DEFAULT_QUORUM_LINGER_MS = 25;
 
+    public static final String QUORUM_APPEND_MAX_UNFLUSHED_BYTES_CONFIG = QUORUM_PREFIX + "append.max.unflushed.bytes";
+    public static final String QUORUM_APPEND_MAX_UNFLUSHED_BYTES_DOC = "The maximum number of bytes that the leader " +
+        "will allow to be accumulated before appending to the topic partition.";
+    public static final int DEFAULT_QUORUM_APPEND_MAX_UNFLUSHED_BYTES = 1_024;
+
     public static final String QUORUM_REQUEST_TIMEOUT_MS_CONFIG = QUORUM_PREFIX +
         CommonClientConfigs.REQUEST_TIMEOUT_MS_CONFIG;
     public static final String QUORUM_REQUEST_TIMEOUT_MS_DOC = CommonClientConfigs.REQUEST_TIMEOUT_MS_DOC;
@@ -98,6 +103,7 @@ public class RaftConfig {
     private final int fetchTimeoutMs;
     private final int appendLingerMs;
     private final Map<Integer, AddressSpec> voterConnections;
+    private final int maxUnflushedBytes;
 
     public interface AddressSpec {
     }
@@ -144,7 +150,8 @@ public class RaftConfig {
             abstractConfig.getInt(QUORUM_ELECTION_TIMEOUT_MS_CONFIG),
             abstractConfig.getInt(QUORUM_ELECTION_BACKOFF_MAX_MS_CONFIG),
             abstractConfig.getInt(QUORUM_FETCH_TIMEOUT_MS_CONFIG),
-            abstractConfig.getInt(QUORUM_LINGER_MS_CONFIG));
+            abstractConfig.getInt(QUORUM_LINGER_MS_CONFIG),
+            abstractConfig.getInt(QUORUM_APPEND_MAX_UNFLUSHED_BYTES_CONFIG));
     }
 
     public RaftConfig(
@@ -154,7 +161,8 @@ public class RaftConfig {
         int electionTimeoutMs,
         int electionBackoffMaxMs,
         int fetchTimeoutMs,
-        int appendLingerMs
+        int appendLingerMs,
+        int maxUnflushedBytes
     ) {
         this.voterConnections = voterConnections;
         this.requestTimeoutMs = requestTimeoutMs;
@@ -163,6 +171,7 @@ public class RaftConfig {
         this.electionBackoffMaxMs = electionBackoffMaxMs;
         this.fetchTimeoutMs = fetchTimeoutMs;
         this.appendLingerMs = appendLingerMs;
+        this.maxUnflushedBytes = maxUnflushedBytes;
     }
 
     public int requestTimeoutMs() {
@@ -187,6 +196,10 @@ public class RaftConfig {
 
     public int appendLingerMs() {
         return appendLingerMs;
+    }
+
+    public int maxUnflushedBytes() {
+        return maxUnflushedBytes;
     }
 
     public Set<Integer> quorumVoterIds() {
--- a/raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
+++ b/raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
@@ -40,6 +40,7 @@ public class BatchAccumulator<T> implements Closeable {
     private final Time time;
     private final SimpleTimer lingerTimer;
     private final int lingerMs;
+    private final int maxUnflushedBytes;
     private final int maxBatchSize;
     private final CompressionType compressionType;
     private final MemoryPool memoryPool;
@@ -61,6 +62,7 @@ public class BatchAccumulator<T> implements Closeable {
         int epoch,
         long baseOffset,
         int lingerMs,
+        int maxUnflushedBytes,
         int maxBatchSize,
         MemoryPool memoryPool,
         Time time,
@@ -69,7 +71,8 @@ public class BatchAccumulator<T> implements Closeable {
     ) {
         this.epoch = epoch;
         this.lingerMs = lingerMs;
-        this.maxBatchSize = maxBatchSize;
+        this.maxUnflushedBytes = maxUnflushedBytes;
+        this.maxBatchSize = Math.min(maxBatchSize, maxUnflushedBytes);
         this.memoryPool = memoryPool;
         this.time = time;
         this.lingerTimer = new SimpleTimer();
@@ -129,6 +132,11 @@ public class BatchAccumulator<T> implements Closeable {
 
         appendLock.lock();
         try {
+            if (accumulatedMaxUnflushedBytes()) {
+                if (currentBatch != null && currentBatch.nonEmpty()) {
+                    completeCurrentBatch();
+                }
+            }
             maybeCompleteDrain();
 
             BatchBuilder<T> batch = null;
@@ -161,6 +169,20 @@ public class BatchAccumulator<T> implements Closeable {
         if (!lingerTimer.isRunning()) {
             lingerTimer.reset(time.milliseconds() + lingerMs);
         }
+    }
+
+    private boolean accumulatedMaxUnflushedBytes() {
+        int currentBatchSize = currentBatch == null || !currentBatch.nonEmpty() ? 0 : currentBatch.approximateSizeInBytes();
+        int completedSize = getCompletedSize();
+        return currentBatchSize + completedSize >= maxUnflushedBytes;
+    }
+
+    private int getCompletedSize() {
+        int completedSize = 0;
+        for (CompletedBatch completedBatch : completed) {
+            completedSize += completedBatch.sizeInBytes();
+        }
+        return completedSize;
     }
 
     private BatchBuilder<T> maybeAllocateBatch(
@@ -251,6 +273,9 @@ public class BatchAccumulator<T> implements Closeable {
         if (drainStatus == DrainStatus.FINISHED) {
             return 0;
         } else {
+            if (lingerTimer.remainingMs(currentTimeMs) > 0 && getCompletedSize() > maxUnflushedBytes)
+                return 0;
+
             return lingerTimer.remainingMs(currentTimeMs);
         }
     }
--- a/raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientSnapshotTest.java
+++ b/raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientSnapshotTest.java
@@ -1360,7 +1360,8 @@ final public class KafkaRaftClientSnapshotTest {
     private static SnapshotWriter<String> snapshotWriter(RaftClientTestContext context, RawSnapshotWriter snapshot) {
         return new SnapshotWriter<>(
             snapshot,
-            4 * 1024,
+            1024,
+            1024,
             MemoryPool.NONE,
             context.time,
             CompressionType.NONE,
--- a/raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
+++ b/raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
@@ -476,6 +476,7 @@ public class KafkaRaftClientTest {
             .thenReturn(buffer);
 
         RaftClientTestContext context = new RaftClientTestContext.Builder(localId, voters)
+            .withMaxUnflushedBytes(KafkaRaftClient.MAX_BATCH_SIZE_BYTES)
             .withAppendLingerMs(lingerMs)
             .withMemoryPool(memoryPool)
             .build();
@@ -505,6 +506,7 @@ public class KafkaRaftClientTest {
             .thenReturn(buffer);
 
         RaftClientTestContext context = new RaftClientTestContext.Builder(localId, voters)
+            .withMaxUnflushedBytes(KafkaRaftClient.MAX_BATCH_SIZE_BYTES)
             .withAppendLingerMs(lingerMs)
             .withMemoryPool(memoryPool)
             .build();
@@ -535,6 +537,7 @@ public class KafkaRaftClientTest {
             .thenReturn(buffer);
 
         RaftClientTestContext context = new RaftClientTestContext.Builder(localId, voters)
+            .withMaxUnflushedBytes(KafkaRaftClient.MAX_BATCH_SIZE_BYTES)
             .withAppendLingerMs(lingerMs)
             .withMemoryPool(memoryPool)
             .build();
@@ -612,6 +615,38 @@ public class KafkaRaftClientTest {
         assertEquals(OptionalLong.of(lingerMs), context.messageQueue.lastPollTimeoutMs());
 
         context.time.sleep(lingerMs);
+        assertEquals(2L, context.client.scheduleAppend(epoch, singletonList("b")));
+        assertTrue(context.messageQueue.wakeupRequested());
+
+        context.client.poll();
+        assertEquals(3L, context.log.endOffset().offset);
+    }
+
+    @Test
+    public void testChannelWokenUpIfMinFlushSizeReachedDuringAppend() throws Exception {
+        // This test verifies that the client will get woken up immediately
+        // if the minimum flush bytes has exceeded during an append
+
+        int localId = 0;
+        int otherNodeId = 1;
+        int maxUnflushedBytes = 120;
+        Set<Integer> voters = Utils.mkSet(localId, otherNodeId);
+
+        RaftClientTestContext context = new RaftClientTestContext.Builder(localId, voters)
+            .withMaxUnflushedBytes(maxUnflushedBytes)
+            .build();
+
+        context.becomeLeader();
+        assertEquals(OptionalInt.of(localId), context.currentLeader());
+        assertEquals(1L, context.log.endOffset().offset);
+
+        int epoch = context.currentEpoch();
+        assertEquals(1L, context.client.scheduleAppend(epoch, singletonList("a")));
+        assertTrue(context.messageQueue.wakeupRequested());
+
+        context.client.poll();
+        assertFalse(context.messageQueue.wakeupRequested());
+
         assertEquals(2L, context.client.scheduleAppend(epoch, singletonList("b")));
         assertTrue(context.messageQueue.wakeupRequested());
 
--- a/core/src/main/scala/kafka/server/KafkaConfig.scala
+++ b/core/src/main/scala/kafka/server/KafkaConfig.scala
@@ -279,6 +279,7 @@ object Defaults {
   val QuorumFetchTimeoutMs = RaftConfig.DEFAULT_QUORUM_FETCH_TIMEOUT_MS
   val QuorumElectionBackoffMs = RaftConfig.DEFAULT_QUORUM_ELECTION_BACKOFF_MAX_MS
   val QuorumLingerMs = RaftConfig.DEFAULT_QUORUM_LINGER_MS
+  val QuorumMaxUnflushedBytes = RaftConfig.DEFAULT_QUORUM_APPEND_MAX_UNFLUSHED_BYTES
   val QuorumRequestTimeoutMs = RaftConfig.DEFAULT_QUORUM_REQUEST_TIMEOUT_MS
   val QuorumRetryBackoffMs = RaftConfig.DEFAULT_QUORUM_RETRY_BACKOFF_MS
 }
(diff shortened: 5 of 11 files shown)


vamossagar12 (commented):
@hachikuji , while you review the rest of the PR whenever you get a chance, this line here: https://github.com/apache/kafka/pull/9756/files#diff-1da15c51e641ea46ea5c86201ab8f21cfee9e7c575102a39c7bae0d5ffd7de39R1531

I wanted to know your opinion on this. Right now, I return whatever value was set in timeUnitFlush. However, this may lead to fsyncs which happen very near in time.

One such scenario that I can think of:

Let's say at the instant when timeUnitFlush was calculated, it's value was 5 ms. Since it's not still <=0, the code also checks if we have accumulated more than min flush bytes and let's assume we have. So, the drain would happen and the logs would get flushed out. However, the timeUnitFlush is still 5ms which gets returned. If this value is less than `timeUntilSend` returned by `maybeSendRequests`, then the pollTimeout would be set to 5 ms and in the next call, there would be another flush that might happen which IMO kind of defeats the purpose of deferring the writes if this happens too frequently.
One option could be to use reset the timer, but again may lead to linger time never getting kicked in. 
Wanted to know if this is a valid scenario or am I over thinking it? Also, do you think in terms of implementation, something needs to change here instead?

vamossagar12:
@hachikuji , did you get a chance to review this one?

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -37,6 +37,7 @@
     private final Time time;
     private final SimpleTimer lingerTimer;
     private final int lingerMs;
+    private final int minFlushSize;
```
hachikuji: The main thing I'm wondering is if it makes sense to merge `minFlushSize` and `maxBatchSize` into a single configuration. I think that would simplify the implementation a bit since we could then check if `completed` is not empty to know whether to drain. It is clear that `minFlushSize` should be at least as large as `maxBatchSize`, but I'm not sure how useful it is for it to be larger. I was planning to keep `maxBatchSize` a static configuration, so I guess if we want `minFlushSize` to be configurable, then we need to allow for `minFlushSize` to be larger. What do you think?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -59,6 +60,11 @@
     private static final String QUORUM_LINGER_MS_DOC = "The duration in milliseconds that the leader will " +
         "wait for writes to accumulate before flushing them to disk.";
 
+    public static final String QUORUM_FLUSH_MIN_SIZE_BYTES_CONFIG = QUORUM_PREFIX + "flush.minSize.bytes";
```
hachikuji: If we decide to keep this, how about we call it `quorum.append.max.unflushed.bytes` or something like that? Basically it's the maximum number of bytes that the raft implementation is allowed to accumulate before forcing an fsync.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -211,6 +214,33 @@ public long timeUntilDrain(long currentTimeMs) {
...
+     *
+     * This means that if the thread responsible for appending is holding the lock and the linger time hasn't expired
+     * yet, then even though the batch size exceeds the min flush size, the records won't be drained as the lock
+     * couldn't be acquired. This also means that in subsequent run(s), this method should be able to acquire the lock
+     * and return true in the event the linger time hasn't expired yet.
+     *
+     * @return true if the append lock could be acquired and the accumulated bytes are greater than configured min flush
+     * bytes size, false otherwise.
+     */
+    public boolean batchSizeExceedsMinFlushSize() {
```
hachikuji: I think we need to look at completed batches in here as well, right? To simplify the implementation, I think we could do something like the following:

1. Inside `append`, while we are already holding the lock, we can check if the accumulated bytes (including `completed` and `currentBatch`) have reached `minFlushSize`. If so, we can call `completeCurrentBatch` to ensure that `completed` holds all the data that needs to be drained.
2. Inside `timeUntilDrain`, if the linger timer hasn't been reached, we can iterate `completed` and check if there are enough bytes to flush. Then we don't need to acquire the lock unless we need to drain.

Would that work?

hachikuji (commented):
@vamossagar12 Thanks for the patch and sorry for the review delay. I left a few comments.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
vamossagar12: The way I thought about it is to still keep it configurable and keep it within the bounds of maxBatchSize. With that, the users would have the option of being able to flush more frequently based on size if it suits them and not wait for time bound flushes. We should still ensure a lower bound for this config otherwise, the fsyncs can become too frequent which can have adverse effects. I agree, setting it higher than `maxBatchSize` also might not be too useful as that might potentially lead to delayed fsyncs if in case the linger time has also been set to a higher value.

Having said that, IMO it might be useful to keep minFlushSize configurable which gives the users more knobs to control the behaviour based upon their needs(even though more knobs sometimes can mean more confusion). Do you think it makes sense?

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
vamossagar12: Yeah completed batches should be considered. Also, with what you have suggested, it removes the need to hold the lock again in batchSizeExceedsMinFlushSize. Basically, we won't even need this method anymore.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
hachikuji: Yeah, I think there should be a configuration here; I am just a little unsure how it should affect the batch size. In the producer, for example, we just expose the batch size directly. We could do the same here, but I'm slightly inclined to keep the batch size decoupled from this configuration. Say for example, that we have a hard-coded 1MB max batch size. Then we could say that the effective batch size is equal to `min(1MB, maxUnflushedSize)`. A user _could_ then define `maxUnflushedSize` to be larger than 1MB and we would not have to deal with the implications of huge batches. Does that make sense?

@jsancio Any thoughts on this?

jsancio: I think it is important to agree on the current (`trunk`) behavior of `BatchAccumulator`. It is my understanding that the `BatchAccumulator` creates an in memory batch when `append` has been called enough times with enough data such that the sum of records sizes is greater than `maxBatchSize` (https://github.com/apache/kafka/pull/9756/files#diff-b7a2129b03764fbafab69c81e985d8ac6006d55f95307f0e21c70fb4750f61b5R145-R148) **OR** the first un-drained `append` is `lingerMs` old (https://github.com/apache/kafka/pull/9756/files#diff-b7a2129b03764fbafab69c81e985d8ac6006d55f95307f0e21c70fb4750f61b5R278).

I see two issues with the current implementation (`trunk`):

1. The draining thread will not call `drain`, even if `completed` is not empty, until `lingerMs` has expired (https://github.com/apache/kafka/pull/9756/files#diff-b7a2129b03764fbafab69c81e985d8ac6006d55f95307f0e21c70fb4750f61b5R254-R255).

Can this issue be address by checking if `completed` is non empty in `timeUntilDrain`?

2. The KafkaRaftClient doesn't expose a way to configure the `BatchAccumulator`'s `maxBatchSize`.

Can this issue be address by adding this configuration to `RaftConfig` and the second constructor to `KafkaRaftClient` (https://github.com/apache/kafka/pull/9756/files#diff-1da15c51e641ea46ea5c86201ab8f21cfee9e7c575102a39c7bae0d5ffd7de39R179)?

jsancio (commented):
Thanks for the PR. I have one high-level comment.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
jsancio: > Say for example, that we have a hard-coded 1MB max batch size. Then we could say that the effective batch size is equal to min(1MB, maxUnflushedSize)

Hmm. I am sure on the right behaviour here. You probably already discussed this and I missed it but can you elaborate on why we want to have a minimum for the maximum batch size?

Are you thinking that we need to have a maximum batch size that is at least as large as the smallest record that the Kafka Controller can generate?

hachikuji: Batching has a lot of implications in the system. The flush behavior is just one. It also impacts the segment size and the ability to do down-conversion efficiently since we have to read the whole batch into memory. Basically any time we need to do anything in memory with the batches, we have to allocate at least enough memory to hold the largest batch. The controller is designed to write small messages, so I do not think 1MB (say) would be much of a constraint. For example, we have avoided storing assignment state in a single message as we did with Zookeeper. We can reconsider it if ever needed, but I'd like to keep batches relatively small if possible.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
jsancio: I like the suggestion of using `quorum.append.max.unflushed.bytes`. I would mention in the documentation that this is a best effort. Based on how `KafakRaftClient` calls drain, it is possible for accumulator have more unflushed bytes.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
jsancio: > 1. Inside append, while we are already holding the lock, we can check if the accumulated bytes (including completed and currentBatch) have reached minFlushSize. If so, we can call completeCurrentBatch to ensure that completed holds all the data that needs to be drained.
> 2. Inside timeUntilDrain, if the linger timer hasn't been reached, we can iterate completed and check if there are enough bytes to flush. Then we don't need to acquire the lock unless we need to drain.

I like this suggestion.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
hachikuji: It sounds like we are basically on the same page here. I think we agree that there should be a configuration. As suggested in the other comment, I like `quorum.append.max.unflushed.bytes` to go along with `quorum.append.linger.ms`. Perhaps we could even use `quorum.append.max.linger.ms` to make it clear that it is the maximum amount of time that the server will hold onto a write before flushing.

In regard to the question of setting the batch size, I suggest we stick with the hard-coded value for now and set the effective value to be the minimum of the hard-coded value and `max.unflushed.bytes`. We can file a separate jira to consider whether this should be configurable separately or whether we can allow `max.unflushed.bytes` to specify the batch size precisely.

@vamossagar12 Does that sound reasonable to you?

vamossagar12: @hachikuji , @jsancio I agree to the points. I am slightly confused on the advantage of letting the config to be set higher than maxBatchSize(1MB) and then limiting the effective batch size to the minimum of `maxUnflushedBytes` and `maxBatchSize`. Won't it lead to confusion in terms of usage?
Infact, I have followed a similar approach in the PR in a different way. What I have done is restricting the value of this new config within a range of 0(which is wrong, need to change) to maxBatchSize - 1:


```            .define(QUORUM_FLUSH_MIN_SIZE_BYTES_CONFIG,
                ConfigDef.Type.INT,
                1024,
                between(0, (1024 * 1024) - 1),
                ConfigDef.Importance.MEDIUM,
                QUORUM_FLUSH_MIN_SIZE_BYTES_DOC);
```

I feel this way, what the config provides and what actually happens remains the same and effectively, this will be equivalent to `min(maxBatchsize, maxUnflushedBytes)`.

Also @hachikuji , wanted to understand how the newly proposed config; `quorum.append.max.linger.ms` would interplay with the existing `quorum.append.linger.ms` config.  As per my understanding, the moment `quorum.append.linger.ms` is crossed, the flush would start. This happens even in this new implementation irrespective of hitting `maxUnflushedBytes` or not. Are you suggesting that we still hold onto writes until we hit the `quorum.append.max.linger.ms`  thereby overriding `quorum.append.linger.ms`? 

Considering just these 2 configs, I am confused of how it will work but if we add the `maxUnflushedBytes`, then probably we will have to write some logic to hold onto writes between `quorum.append.linger.ms` and `quorum.append.max.linger.ms`. Maybe, we give further time between this time range to let the writes be accumulated. The moment we hit either of the 2 i.e `maxUnflushedBytes` or `quorum.append.max.linger.ms`, we flush the writes. This gives me the impression that we are giving more preference to `maxUnflushedBytes`. 

Let me know your thoughts plz.

hachikuji: > I am slightly confused on the advantage of letting the config to be set higher than maxBatchSize(1MB) and then limiting the effective batch size to the minimum of maxUnflushedBytes and maxBatchSize. Won't it lead to confusion in terms of usage?

Not sure if this is the cause of the confusion, but just to be clear, what I'm referring to as the batch size is just a limit to the size of a `DefaultRecordBatch`. Records in Kafka are grouped together into batches with a schema roughly like this:
```
Batch => Size Offset Crc ... [Record]
```
It is useful to keep batches from getting too large because we don't have a way to index into a batch without storing the whole decompressed contents in memory. 

However, we can let `maxUnflushedBytes` be as large as we want. That just means that we will collect multiple batches (i.e. multiple `CompletedBatch` instances) as the code already does. So if `maxUnflushedBytes` is set to 4MB, then we will end up collecting 4x1MB batches before we flush to disk rather than collecting one big 4MB batch. 

We could also require that `maxUnflushedBytes` be less than or equal to our desired max batch size, but if it's a similar level of effort to support the more general config, then that seems better.

> Also @hachikuji , wanted to understand how the newly proposed config; quorum.append.max.linger.ms would interplay with the existing quorum.append.linger.ms config. As per my understanding, the moment quorum.append.linger.ms is crossed, the flush would start. This happens even in this new implementation irrespective of hitting maxUnflushedBytes or not. Are you suggesting that we still hold onto writes until we hit the quorum.append.max.linger.ms thereby overriding quorum.append.linger.ms?

My expectation is that we flush after either of these are reached. So if the linger time is hit first, then we take whatever unflushed bytes we have even if they are smaller than `maxUnflushedBytes`. On the other hand, if `maxUnflushedBytes` is reached, then we ignore linger. I think this is consistent with the implementation I suggested here: https://github.com/apache/kafka/pull/9756#discussion_r552268919. Basically we override `timeUntilDrain` so that it returns 0 when `maxUnflushedBytes` is reached.

Hope that helps!

vamossagar12:
@hachikuji  @jsancio  i have made the changes. Plz review.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientSnapshotTest.java
```diff
@@ -1078,6 +1078,7 @@ private static FetchResponseData snapshotFetchResponse(
     private static SnapshotWriter<String> snapshotWriter(RaftClientTestContext context, RawSnapshotWriter snapshot) {
         return new SnapshotWriter<>(
             snapshot,
+            1024,
```
jsancio, line 1363: In the snapshot tests, we can set the `maxUnflushedBytes` to the same value as `maxBatchSize`.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -474,6 +474,7 @@ public void testAccumulatorClearedAfterBecomingFollower() throws Exception {
             .thenReturn(buffer);
 
         RaftClientTestContext context = new RaftClientTestContext.Builder(localId, voters)
+            .withMaxUnflushedBytes(KafkaRaftClient.MAX_BATCH_SIZE)
```
jsancio: This comment applies to a few places in this file.

Is there a reason why we override this value?

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -617,6 +620,38 @@ public void testChannelWokenUpIfLingerTimeoutReachedDuringAppend() throws Except
         assertEquals(3L, context.log.endOffset().offset);
     }
 
+    @Test
+    public void testChannelWokenUpIfMinFlushSizeReachedDuringAppend() throws Exception {
+        // This test verifies that the client will get woken up immediately
+        // if the linger timeout has expired during an append
```
jsancio: > "if the linger timeout..."

Did you mean minimum flush bytes?

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -617,6 +620,38 @@ public void testChannelWokenUpIfLingerTimeoutReachedDuringAppend() throws Except
...
+
+        int localId = 0;
+        int otherNodeId = 1;
+        int minFlushSizeInBytes = 120;
+        Set<Integer> voters = Utils.mkSet(localId, otherNodeId);
+
+        RaftClientTestContext context = new RaftClientTestContext.Builder(localId, voters)
+            .withMaxUnflushedBytes(minFlushSizeInBytes)
```
jsancio: The local variable is called `minFlushSizeInBytes` yet the `KafkaRaftClient` uses `maxUnflushBytes`. Did we agree to use `minUnflushedBytes`?

vamossagar12:
@hachikuji did you get a chance to review this?

hachikuji:
@vamossagar12 Thanks, I missed the update. I will review today.

vamossagar12:
@hachikuji , there were some merge conflicts which needed to be resolved. Can you plz review the PR whenever you get the chance ?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -76,6 +76,11 @@
         "wait for writes to accumulate before flushing them to disk.";
     public static final int DEFAULT_QUORUM_LINGER_MS = 25;
 
+    public static final String QUORUM_APPEND_MAX_UNFLUSHED_BYTES_CONFIG = QUORUM_PREFIX + "append.max.unflushed.bytes";
```
jsancio, line 84: @vamossagar12 and @hachikuji  How about we call this `append.linger.bytes`?

Excuse the back and forth on this name put let me explain.

In the current implementation this does determine when the bytes will be flush because the `KafkaRaftClient` flushes the `Log` every time it drains the `BatchAccumulator` and it appends to the log.

https://github.com/apache/kafka/blob/trunk/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java#L1871-L1879

In the past, we have talked about delaying the flush/fsync to the log. I believe that the invariant that we need to satisfied is that the leader cannot increase the high-watermark past the flushed offsets. Followers need to flush/fsync before sending a Fetch request since the leader assumes that the followers have safely replicated the offset in the Fetch request, 

If in the future we implement this logic or something similar, I think the name `append.max.unflushed.bytes` would not be accurate since it contains the word "unflushed". And the leader may decide to have more than `append.max.unflushed.bytes` bytes unflushed.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -76,6 +76,11 @@
         "wait for writes to accumulate before flushing them to disk.";
```
jsancio, line 81: Outside the scope of this PR but how about changing this description to:
> The duration in milliseconds that the leader will wait for writes to accumulate before appending to the topic partition.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -76,6 +76,11 @@
...
     public static final int DEFAULT_QUORUM_LINGER_MS = 25;
 
+    public static final String QUORUM_APPEND_MAX_UNFLUSHED_BYTES_CONFIG = QUORUM_PREFIX + "append.max.unflushed.bytes";
+    public static final String QUORUM_APPEND_MAX_UNFLUSHED_BYTES_DOC = "The maximum number of bytes that the leader " +
+        "will allow to be accumulated before flushing them to disk ";
```
jsancio:
```suggestion
    public static final String QUORUM_APPEND_MAX_UNFLUSHED_BYTES_DOC = "The maximum number of bytes that the leader " +
        "will allow to be accumulated before appending to the topic partition.";
```

jsancio (commented):
Excuse the delayed comments.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientSnapshotTest.java
vamossagar12: done

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
vamossagar12: yes.. changed

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
vamossagar12: yeah that was an oversight as well. Changed.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
vamossagar12: Yeah. The reason for that is that in this PR @hachikuji had suggested to change the logic for setting maxBatchSize in BatchAccumulator to the following way: 

`this.maxBatchSize = Math.min(maxBatchSize, maxUnflushedBytes);`

That is the reason I am setting some combinations of value to check if it behaves correctly.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
vamossagar12: I see what you are trying to say. Well, the premise of this ticket originally was to trigger fsyncs happen the moment a configured amount of bytes have been accumulated. Here is the original description for Jason in the ticket:

> In KAFKA-10601, we implemented linger semantics similar to the producer to let the leader accumulate a batch of writes before fsyncing them to disk. Currently the fsync is only based on the linger time, but it would be helpful to make it size-based as well. In other words, if we accumulate a configurable N bytes, then we should not wait for linger expiration and should just fsync immediately.

But as you pointed out, it is also due to the fact that in the current implementation batch append and fsync goes hand in hand. 

With the future implementation on deferring fsync, this might just affect the batch appends and considering that in mind, imo it makes sense to rename it to `append.linger.bytes` . It also matches with `append.linger.ms` .

BTW, on the fsync deferral track. i had created a draft PR where i have outlined my approach: https://github.com/apache/kafka/pull/10278

Request you or Jason to review this..

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
vamossagar12: I updated the doc for the maxUnflushedBytes config.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
vamossagar12: hey @jsancio .. sorry to bother again, but could you plz review this whenever you get the chance?

vamossagar12: hi @jsancio / @hachikuji  could you plz review the PR whenever you get the chance?

vamossagar12:
@jsancio could you plz review this PR whenever you get the chance?

PR #9758, opened 2020-12-16
MINOR: remove FetchResponse.AbortedTransaction and redundant construc…

chia7712:
## Changes
1. rename ```INVALID_HIGHWATERMARK```  to ```INVALID_HIGH_WATERMARK```
1. replace ```FetchResponse.AbortedTransaction``` by ```FetchResponseData.AbortedTransaction```
1. remove redundant constructors from ```FetchResponse.PartitionData```
1. rename ```recordSet``` to ```records``` 

## Performance Tests

loop **10** times and get average.

### case 1:  @parametrize(acks=1, topic=TOPIC_REP_ONE)

diff: **-0.3994016685 %**
- trunk: 64.847 MB/sec
- patch: 64.588 MB/sec
    
### case 2:  @parametrize(acks=-1, topic=TOPIC_REP_THREE)
    
diff: **+0.4917916785 %**
- trunk: 28.264 MB/sec
- patch: 28.403 MB/sec


core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -777,12 +775,12 @@ class KafkaApis(val requestChannel: RequestChannel,
     }
 
     def maybeConvertFetchedData(tp: TopicPartition,
```
chia7712, line 727: @ijuma This is the only case that we create ```LazyDownConversionRecords``` in production. Through this PR, this case can get rid of generic ```FetchResponse.PartitionData```. Hence, we can remove generic from ```FetchResponse.PartitionData``` after this PR goes in trunk.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
```diff
@@ -733,11 +733,11 @@ void resetOffsetIfNeeded(TopicPartition partition, OffsetResetStrategy requested
...
 
     private void resetOffsetsAsync(Map<TopicPartition, Long> partitionResetTimestamps) {
-        Map<Node, Map<TopicPartition, ListOffsetsPartition>> timestampsToSearchByNode =
+        Map<Node, Map<TopicPartition, ListOffsetsRequestData.ListOffsetsPartition>> timestampsToSearchByNode =
                 groupListOffsetRequests(partitionResetTimestamps, new HashSet<>());
-        for (Map.Entry<Node, Map<TopicPartition, ListOffsetsPartition>> entry : timestampsToSearchByNode.entrySet()) {
+        for (Map.Entry<Node, Map<TopicPartition, ListOffsetsRequestData.ListOffsetsPartition>> entry : timestampsToSearchByNode.entrySet()) {
             Node node = entry.getKey();
-            final Map<TopicPartition, ListOffsetsPartition> resetTimestamps = entry.getValue();
+            final Map<TopicPartition, ListOffsetsRequestData.ListOffsetsPartition> resetTimestamps = entry.getValue();
```
ijuma: These changes seem unrelated?

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -59,136 +60,140 @@
...
+    public static final long INVALID_HIGH_WATERMARK = -1L;
     public static final long INVALID_LAST_STABLE_OFFSET = -1L;
     public static final long INVALID_LOG_START_OFFSET = -1L;
     public static final int INVALID_PREFERRED_REPLICA_ID = -1;
 
     private final FetchResponseData data;
-    private final LinkedHashMap<TopicPartition, PartitionData<T>> responseDataMap;
+    private final LinkedHashMap<TopicPartition, FetchResponseData.FetchablePartitionResponse> partitionData;
+    // lazily instantiate this field
+    private LinkedHashMap<TopicPartition, PartitionData<T>> responseDataMap = null;
```
ijuma: Can we remove this altogether?

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -59,136 +60,140 @@
...
-            } else {
-                partitionResponse.setAbortedTransactions(null);
-            }
-            partitionResponse.setPreferredReadReplica(preferredReadReplica.orElse(INVALID_PREFERRED_REPLICA_ID));
-            partitionResponse.setRecordSet(records);
-            divergingEpoch.ifPresent(partitionResponse::setDivergingEpoch);
 
-            this.partitionResponse = partitionResponse;
-        }
+    public static final class PartitionData<T extends BaseRecords> {
```
ijuma: Do we need this class? What does it add over `FetchResponseData.FetchablePartitionRespons`?

ijuma (commented):
Thanks for the PR. A couple of initial comments.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: ```partitionData``` can be removed but ```responseDataMap``` is still a response type and it is used by production code. I did not refactor that to avoid big patch.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: > What does it add over FetchResponseData.FetchablePartitionRespons?

It offers methods to transfer "non-defined value" to ```Optional``` type.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
ijuma: I understand. It's a bit hard to understand the end state for this class as it stands. We've taken a few intermediate steps to this point. Can we just go to the end state now?

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
ijuma: It seems like we do it for a couple of cases. It doesn't seem worth it given that fetch is one of the hot paths. We could have static helper methods instead if helpful.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: You are right. Will address it in next commit.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: Nice suggestions. will address it.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -304,58 +115,12 @@ public int sessionId() {
...
-            topicResponse.partitionResponses().forEach(partitionResponse -> {
-                TopicPartition tp = new TopicPartition(topicResponse.topic(), partitionResponse.partition());
-                PartitionData<T> partitionData = new PartitionData<>(partitionResponse);
-                responseMap.put(tp, partitionData);
-            });
-        });
-        return responseMap;
-    }
-
-    private static <T extends BaseRecords> FetchResponseData toMessage(int throttleTimeMs, Errors error,
```
chia7712, line 331: I don't batch the partitions again in this PR as it create a new ```FetchResponseData```.

core/src/main/scala/kafka/server/FetchSession.scala
```diff
@@ -142,19 +141,20 @@ class CachedPartition(val topic: String,
       if (updateResponseData)
         localLogStartOffset = respData.logStartOffset
     }
-    if (respData.preferredReadReplica.isPresent) {
+    if (FetchResponse.preferredReadReplica(respData).isPresent) {
```
ijuma: Maybe we could have a isPreferredReplica method since there are at least two places where we just call isPresent on the result.

chia7712:
fix conflicting files

chia7712:
fix conflicting files again

chia7712:
@ijuma I add some perf results to description. In short, I don't observe significant regression.

ijuma:
Sorry for the delay on this one. I'll try to review in the next few days if no-one else beats me to it.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
```diff
@@ -290,7 +291,7 @@ public void onSuccess(ClientResponse resp) {
                             Set<TopicPartition> partitions = new HashSet<>(response.responseData().keySet());
                             FetchResponseMetricAggregator metricAggregator = new FetchResponseMetricAggregator(sensors, partitions);
 
-                            for (Map.Entry<TopicPartition, FetchResponse.PartitionData<Records>> entry : response.responseData().entrySet()) {
+                            for (Map.Entry<TopicPartition, FetchResponseData.FetchablePartitionResponse> entry : response.responseData().entrySet()) {
```
ijuma: `FetchablePartitionResponse` is a bit long and redundant. Could we find a shorter name for it now that `PartitionData` is gone?

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
```diff
@@ -1257,7 +1259,7 @@ private CompletedFetch initializeCompletedFetch(CompletedFetch nextCompletedFetc
 
                 log.trace("Preparing to read {} bytes of data for partition {} with offset {}",
                         partition.records().sizeInBytes(), tp, position);
-                Iterator<? extends RecordBatch> batches = partition.records().batches().iterator();
+                Iterator<? extends RecordBatch> batches = ((Records) partition.records()).batches().iterator();
```
ijuma: Could we encapsulate this cast in a utility method?

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/fetchsession/FetchSessionBenchmark.java
```diff
@@ -70,24 +70,26 @@ public void setUp() {
...
-                    null));
+            respMap.put(tp, new FetchResponseData.FetchablePartitionResponse()
+                            .setPartition(tp.partition())
+                            .setErrorCode(Errors.NONE.code())
+                            .setHighWatermark(0)
+                            .setLastStableOffset(0)
+                            .setLogStartOffset(0)
+                            .setAbortedTransactions(null)
+                            .setRecords(null)
+                            .setPreferredReadReplica(FetchResponse.INVALID_PREFERRED_REPLICA_ID));
```
ijuma: Maybe we can remove the defaults from this and every other place where we build `FetchablePartitionResponse`

core/src/test/scala/unit/kafka/server/LogOffsetTest.scala
```diff
@@ -127,7 +124,7 @@ class LogOffsetTest extends BaseRequestTest {
       Map(topicPartition -> new FetchRequest.PartitionData(consumerOffsets.head, FetchRequest.INVALID_LOG_START_OFFSET,
         300 * 1024, Optional.empty())).asJava).build()
     val fetchResponse = sendFetchRequest(fetchRequest)
-    assertFalse(fetchResponse.responseData.get(topicPartition).records.batches.iterator.hasNext)
+    assertFalse(fetchResponse.responseData.get(topicPartition).records.asInstanceOf[Records].batches.iterator.hasNext)
```
ijuma: Similar to elsewhere, it would be useful to have a utility method to avoid unsafe operations all over the code.

core/src/main/scala/kafka/server/ReplicaAlterLogDirsThread.scala
```diff
@@ -110,7 +113,7 @@ class ReplicaAlterLogDirsThread(name: String,
   // process fetched data
   override def processPartitionData(topicPartition: TopicPartition,
                                     fetchOffset: Long,
-                                    partitionData: PartitionData[Records]): Option[LogAppendInfo] = {
+                                    partitionData: FetchResponseData.FetchablePartitionResponse): Option[LogAppendInfo] = {
```
ijuma: Could this be `FetchData` too? Are there other places like it?

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -57,238 +55,51 @@
...
-    /**
-     * From version 3 or later, the entries in `responseData` should be in the same order as the entries in
-     * `FetchRequest.fetchData`.
-     *
-     * @param error             The top-level error code.
-     * @param responseData      The fetched data grouped by partition.
-     * @param throttleTimeMs    The time in milliseconds that the response was throttled
-     * @param sessionId         The fetch session id.
-     */
     public FetchResponse(Errors error,
```
ijuma: I think it would be better to make this a static factory method and keep the constructor for the case where we receive `FetchResponseData`.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -57,238 +55,51 @@
...
+                .setTopic(entry.getKey().topic())
+                .setPartitionResponses(Collections.singletonList(entry.getValue().setPartition(entry.getKey().partition()))))
+                .collect(Collectors.toList())));
     }
 
     public FetchResponse(FetchResponseData fetchResponseData) {
         super(ApiKeys.FETCH);
         this.data = fetchResponseData;
-        this.responseDataMap = toResponseDataMap(fetchResponseData);
+        this.responseData = new LinkedHashMap<>();
```
ijuma: This isn't needed when we return a fetch from the broker, right? If this is true, can we remove it from the fetch response and build it on the client when needed?

ijuma (commented):
Thanks for the PR. I took an initial pass and left some comments.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
chia7712: sure. I reuse the name `PartitionData`

clients/src/main/resources/common/message/FetchResponse.json
```diff
@@ -53,9 +53,9 @@
       "about": "The response topics.", "fields": [
       { "name": "Topic", "type": "string", "versions": "0+", "entityType": "topicName",
         "about": "The topic name." },
-      { "name": "PartitionResponses", "type": "[]FetchablePartitionResponse", "versions": "0+",
+      { "name": "Partitions", "type": "[]PartitionData", "versions": "0+",
         "about": "The topic partitions.", "fields": [
-        { "name": "Partition", "type": "int32", "versions": "0+",
+        { "name": "Index", "type": "int32", "versions": "0+",
```
ijuma: We typically call this `PartitionIndex` in other requests/responses. Is that right?

chia7712: > We typically call this PartitionIndex in other requests/responses. Is that right?

You are right!

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -57,238 +56,39 @@
...
+public class FetchResponse extends AbstractResponse {
+    public static final long INVALID_HIGH_WATERMARK = -1L;
     public static final long INVALID_LAST_STABLE_OFFSET = -1L;
     public static final long INVALID_LOG_START_OFFSET = -1L;
     public static final int INVALID_PREFERRED_REPLICA_ID = -1;
 
     private final FetchResponseData data;
-    private final LinkedHashMap<TopicPartition, PartitionData<T>> responseDataMap;
+    // we build responseData when needed.
+    private LinkedHashMap<TopicPartition, FetchResponseData.PartitionData> responseData = null;
```
ijuma: This is not thread-safe and requests are typically thread-safe. What's the thinking here?

ijuma: Looks like this method is used only `Fetcher.sendFetches` outside of tests, benchmaks, etc.. Also once in `errorCounts`, but we can change that code. Maybe we can remove this altogether. Thoughts?

ijuma: I think I would move the method to a test utility so that tests can use that instead.

chia7712: It is used by `FetchSessionHandler` also. I prefer to rewrite that code by another PR to avoid big patch :(

It seems to me using synchronization block can resolve the thread issue. Also, I will file a jira as follow-up. WDYT?

ijuma: Yeah, you can use a volatile field and synchronize on the assignment if still null. And then file a separate Jira to remove it from the class in a separate PR.

chia7712: jira: https://issues.apache.org/jira/browse/KAFKA-12385

> you can use a volatile field and synchronize on the assignment if still null.

updated

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -304,58 +108,12 @@ public int sessionId() {
     public Map<Errors, Integer> errorCounts() {
         Map<Errors, Integer> errorCounts = new HashMap<>();
         updateErrorCounts(errorCounts, error());
-        responseDataMap.values().forEach(response ->
-            updateErrorCounts(errorCounts, response.error())
-        );
+        responseData().values().forEach(response -> updateErrorCounts(errorCounts, Errors.forCode(response.errorCode())));
```
ijuma: Can we update this not to use `responseData`? Then we at least have the right behavior for the broker and we can fix the clients in the subsequent PR.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
ijuma: Can you clarify what you mean here?

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -365,17 +123,70 @@ public int sessionId() {
...
+
+    public static FetchResponseData.PartitionData partitionResponse(int partition, Errors error) {
+        return new FetchResponseData.PartitionData()
+                .setPartitionIndex(partition)
+                .setErrorCode(error.code())
+                .setHighWatermark(FetchResponse.INVALID_HIGH_WATERMARK)
+                .setLastStableOffset(FetchResponse.INVALID_LAST_STABLE_OFFSET)
+                .setLogStartOffset(FetchResponse.INVALID_LOG_START_OFFSET)
+                .setAbortedTransactions(null)
+                .setRecords(MemoryRecords.EMPTY);
```
ijuma: Aren't many of these set automatically by the generated classes?

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -365,17 +123,70 @@ public int sessionId() {
...
+    }
+
+    /**
+     * cast the BaseRecords of PartitionData to Records. This is used to eliminate duplicate code of type casting.
```
ijuma: Does this ever fail? If so, it would be good to explain under which conditions it can fail. Also "This is used to eliminate duplicate code of type casting." seems a bit redundant.

clients/src/test/java/org/apache/kafka/clients/FetchSessionHandlerTest.java
```diff
@@ -150,22 +150,23 @@ private static void assertListEquals(List<TopicPartition> expected, List<TopicPa
...
-            this.data = new FetchResponse.PartitionData<>(
-                Errors.NONE,
-                highWatermark,
-                lastStableOffset,
-                0,
-                null,
-                null);
+
+            this.data = new FetchResponseData.PartitionData()
+                        .setErrorCode(Errors.NONE.code())
```
ijuma: Nit: indenting seems excessive.

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
```diff
@@ -2377,14 +2378,19 @@ private ListOffsetsResponse listOffsetsResponse(Map<TopicPartition, Long> partit
...
-                    Errors.NONE, highWatermark, FetchResponse.INVALID_LAST_STABLE_OFFSET,
-                    logStartOffset, null, records));
+            tpResponses.put(partition,
+                    new FetchResponseData.PartitionData()
+                            .setErrorCode(Errors.NONE.code())
+                            .setHighWatermark(highWatermark)
+                            .setLastStableOffset(FetchResponse.INVALID_LAST_STABLE_OFFSET)
+                            .setLogStartOffset(logStartOffset)
+                            .setAbortedTransactions(null)
+                            .setRecords(records));
```
ijuma: Are some of these redundant? (eg `setAbortedTransactions(null)`)

clients/src/test/java/org/apache/kafka/clients/FetchSessionHandlerTest.java
```diff
@@ -150,22 +150,23 @@ private static void assertListEquals(List<TopicPartition> expected, List<TopicPa
...
+                        .setHighWatermark(highWatermark)
+                        .setLastStableOffset(lastStableOffset)
+                        .setLogStartOffset(0)
+                        .setAbortedTransactions(null)
+                        .setRecords(null);
```
ijuma: Are some of these redundant? (eg setAbortedTransactions(null))

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
```diff
@@ -1270,13 +1271,24 @@ public void testFetchPositionAfterException() {
...
+                .setLogStartOffset(FetchResponse.INVALID_LOG_START_OFFSET)
+                .setAbortedTransactions(null)
+                .setRecords(records));
+        partitions.put(tp0, new FetchResponseData.PartitionData()
+                .setErrorCode(Errors.OFFSET_OUT_OF_RANGE.code())
+                .setHighWatermark(100)
+                .setLastStableOffset(FetchResponse.INVALID_LAST_STABLE_OFFSET)
+                .setLogStartOffset(FetchResponse.INVALID_LOG_START_OFFSET)
+                .setAbortedTransactions(null)
+                .setRecords(MemoryRecords.EMPTY));
```
ijuma: Are some of these redundant? (eg `setAbortedTransactions(null)`). Other examples in the same file.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -364,7 +361,7 @@ abstract class AbstractFetcherThread(name: String,
                       }
                     }
                     if (isTruncationOnFetchSupported) {
-                      partitionData.divergingEpoch.ifPresent { divergingEpoch =>
+                     FetchResponse.divergingEpoch(partitionData).ifPresent { divergingEpoch =>
```
ijuma: Nit: indenting seems wrong.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -340,7 +336,8 @@ abstract class AbstractFetcherThread(name: String,
             // the current offset is the same as the offset requested.
             val fetchPartitionData = sessionPartitions.get(topicPartition)
             if (fetchPartitionData != null && fetchPartitionData.fetchOffset == currentFetchState.fetchOffset && currentFetchState.isReadyForFetch) {
-              partitionData.error match {
+              val partitionError = Errors.forCode(partitionData.errorCode)
```
ijuma: Not clear why we need this val. Seems like we can introduce a variable in the `case _` instead.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -734,7 +730,7 @@ abstract class AbstractFetcherThread(name: String,
     Option(partitionStates.stateValue(topicPartition))
   }
 
-  protected def toMemoryRecords(records: Records): MemoryRecords = {
+  protected def toMemoryRecords(records: BaseRecords): MemoryRecords = {
```
ijuma: Do we have to update the matching inside the method to handle other potential records types? Or do we want to avoid changing this method signature instead?

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/fetchsession/FetchSessionBenchmark.java
```diff
@@ -70,24 +70,25 @@ public void setUp() {
...
-                    null,
-                    null));
+            respMap.put(tp, new FetchResponseData.PartitionData()
+                            .setPartitionIndex(tp.partition())
+                            .setErrorCode(Errors.NONE.code())
+                            .setHighWatermark(0)
+                            .setLastStableOffset(0)
+                            .setLogStartOffset(0)
+                            .setAbortedTransactions(null)
+                            .setRecords(null));
```
ijuma: We can remove some redundant set calls?

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/fetcher/ReplicaFetcherThreadBenchmark.java
```diff
@@ -174,8 +173,14 @@ public int sizeInBytes() {
...
                 }
             };
-            initialFetched.put(tp, new FetchResponse.PartitionData<>(Errors.NONE, 0, 0, 0,
-                    new LinkedList<>(), fetched));
+            initialFetched.put(tp, new FetchResponseData.PartitionData()
+                    .setPartitionIndex(tp.partition())
+                    .setErrorCode(Errors.NONE.code())
+                    .setHighWatermark(0)
+                    .setLastStableOffset(0)
+                    .setLogStartOffset(0)
```
ijuma, line 179: We can remove some redundant setters?

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/common/FetchResponseBenchmark.java
```diff
@@ -78,19 +78,25 @@ public void setup() {
...
             String topic = UUID.randomUUID().toString();
             for (int partitionId = 0; partitionId < partitionCount; partitionId++) {
-                FetchResponse.PartitionData<MemoryRecords> partitionData = new FetchResponse.PartitionData<>(
-                    Errors.NONE, 0, 0, 0, Optional.empty(), Collections.emptyList(), records);
+                FetchResponseData.PartitionData partitionData = new FetchResponseData.PartitionData()
+                                .setPartitionIndex(partitionId)
+                                .setErrorCode(Errors.NONE.code())
+                                .setHighWatermark(0)
+                                .setLastStableOffset(0)
+                                .setLogStartOffset(0)
```
ijuma, line 83: We can remove some redundant setters?

core/src/test/scala/unit/kafka/server/ReplicaFetcherThreadTest.scala
```diff
@@ -963,9 +966,13 @@ class ReplicaFetcherThreadTest {
...
     val records = MemoryRecords.withRecords(CompressionType.NONE,
       new SimpleRecord(1000, "foo".getBytes(StandardCharsets.UTF_8)))
-
-    val partitionData: thread.FetchData = new FetchResponse.PartitionData[Records](
-      Errors.NONE, 0, 0, 0, Optional.empty(), Collections.emptyList(), records)
+    val partitionData: thread.FetchData = new FetchResponseData.PartitionData()
+        .setErrorCode(Errors.NONE.code)
+        .setHighWatermark(0)
+        .setLastStableOffset(0)
+        .setLogStartOffset(0)
```
ijuma: We can remove some redundant setters?

core/src/test/scala/unit/kafka/server/FetchSessionTest.scala
```diff
@@ -155,13 +155,28 @@ class FetchSessionTest {
...
-    response.put(tp2, new FetchResponse.PartitionData(
-      Errors.NONE, 5, 5, 5, null, null))
+    val response = new util.LinkedHashMap[TopicPartition, FetchResponseData.PartitionData]
+    response.put(tp0, new FetchResponseData.PartitionData()
+      .setErrorCode(Errors.NONE.code)
+      .setHighWatermark(100)
+      .setLastStableOffset(100)
+      .setLogStartOffset(100)
+      .setAbortedTransactions(null)
+      .setRecords(null))
```
ijuma: We can remove some redundant setters?

core/src/main/scala/kafka/tools/ReplicaVerificationTool.scala
```diff
@@ -403,25 +401,24 @@ private class ReplicaFetcher(name: String, sourceBroker: Node, topicPartitions:
...
         if (!isRunning)
           throw t
     }
 
     if (fetchResponse != null) {
-      fetchResponse.responseData.forEach { (tp, partitionData) =>
-        replicaBuffer.addFetchedData(tp, sourceBroker.id, partitionData)
-      }
+      fetchResponse.data.responses().forEach(topicResponse =>
+        topicResponse.partitions().forEach(partitionResponse =>
```
ijuma: Nit: remove `()` twice.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -761,76 +754,80 @@ class KafkaApis(val requestChannel: RequestChannel,
...
-                  new LazyDownConversionRecords(tp, unconvertedRecords, magic, fetchContext.getFetchOffset(tp).get, time))
+                new FetchResponseData.PartitionData()
+                  .setPartitionIndex(tp.partition)
+                  .setErrorCode(maybeDownConvertStorageError(Errors.forCode(partitionData.errorCode)).code)
+                  .setHighWatermark(partitionData.highWatermark)
+                  .setLastStableOffset(partitionData.lastStableOffset)
+                  .setLogStartOffset(partitionData.logStartOffset)
+                  .setAbortedTransactions(partitionData.abortedTransactions)
+                  .setRecords(new LazyDownConversionRecords(tp, unconvertedRecords, magic, fetchContext.getFetchOffset(tp).get, time))
+                  .setPreferredReadReplica(partitionData.preferredReadReplica())
```
ijuma, line 774: Do we have to copy like this or can we mutate the response?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -761,76 +754,80 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+            new FetchResponseData.PartitionData()
+              .setPartitionIndex(tp.partition)
+              .setErrorCode(maybeDownConvertStorageError(Errors.forCode(partitionData.errorCode)).code)
+              .setHighWatermark(partitionData.highWatermark)
+              .setLastStableOffset(partitionData.lastStableOffset)
+              .setLogStartOffset(partitionData.logStartOffset)
+              .setAbortedTransactions(partitionData.abortedTransactions)
+              .setRecords(unconvertedRecords)
+              .setPreferredReadReplica(partitionData.preferredReadReplica)
+              .setDivergingEpoch(partitionData.divergingEpoch)
```
ijuma, line 791: Similar question, is the copy required?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -761,76 +754,80 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+        partitions.put(tp, new FetchResponseData.PartitionData()
+            .setPartitionIndex(tp.partition)
+            .setErrorCode(maybeDownConvertStorageError(data.error).code)
+            .setHighWatermark(data.highWatermark)
+            .setLastStableOffset(lastStableOffset)
+            .setLogStartOffset(data.logStartOffset)
+            .setAbortedTransactions(abortedTransactions)
+            .setRecords(data.records)
+            .setPreferredReadReplica(data.preferredReadReplica.getOrElse(FetchResponse.INVALID_PREFERRED_REPLICA_ID))
+            .setDivergingEpoch(data.divergingEpoch.getOrElse(new FetchResponseData.EpochEndOffset)))
```
ijuma: Similar question, is the copy required?

ijuma (commented):
Thanks for the updates. A few more questions/comments below.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: the data from KRPC always use `MemoryRecords` so it should never fail if the data is from KRPC. I will add more comments for this case.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: you are right. Except for `HighWatermark`, other args have default value. remove duplicate assignment!

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
chia7712: Good catch! updated

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
chia7712: This change can be reverted as we have helper method `FetchResponse#records` now.

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/fetcher/ReplicaFetcherThreadBenchmark.java
chia7712: I will remove redundant setter `setHighWatermark(0)`. Other values are different from default value in KRPC.

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/common/FetchResponseBenchmark.java
chia7712: updated

core/src/test/scala/unit/kafka/server/FetchSessionTest.scala
chia7712: updated

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/fetchsession/FetchSessionBenchmark.java
chia7712: updated

core/src/main/scala/kafka/server/KafkaApis.scala
chia7712: It created copy before so I don't change the behavior in this PR. We can investigate it in separate PR :)

core/src/main/scala/kafka/server/KafkaApis.scala
chia7712: same to https://github.com/apache/kafka/pull/9758#discussion_r584224693

core/src/main/scala/kafka/server/KafkaApis.scala
chia7712: same to #9758 (comment)

core/src/main/scala/kafka/server/KafkaApis.scala
chia7712: https://issues.apache.org/jira/browse/KAFKA-12387

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: Oh, I planed to remove all usages of this method from production (i.e `KafkaApis` should generate batched response directly). However, it can produce a big patch so I will keep this method in next commit.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -791,15 +791,18 @@ public void produceRequestGetErrorResponseTest() {
...
-                Errors.NONE, 1000000, FetchResponse.INVALID_LAST_STABLE_OFFSET,
-                0L, Optional.empty(), Collections.emptyList(), records));
-
-        FetchResponse<MemoryRecords> v0Response = new FetchResponse<>(Errors.NONE, responseData, 0, INVALID_SESSION_ID);
-        FetchResponse<MemoryRecords> v1Response = new FetchResponse<>(Errors.NONE, responseData, 10, INVALID_SESSION_ID);
+        responseData.put(new TopicPartition("test", 0),
+                new FetchResponseData.PartitionData()
+                        .setHighWatermark(1000000)
+                        .setLogStartOffset(0)
+                        .setAbortedTransactions(Collections.emptyList())
```
ijuma: Aborted transactions is empty by default.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -808,22 +811,32 @@ public void fetchResponseVersionTest() {
...
-
-        FetchResponse<MemoryRecords> response = new FetchResponse<>(Errors.NONE, responseData, 10, INVALID_SESSION_ID);
-        FetchResponse<MemoryRecords> deserialized = FetchResponse.parse(response.serialize((short) 4), (short) 4);
+        responseData.put(new TopicPartition("bar", 0),
+                new FetchResponseData.PartitionData()
+                        .setHighWatermark(1000000)
+                        .setAbortedTransactions(abortedTransactions)
+                        .setRecords(records));
+        responseData.put(new TopicPartition("bar", 1),
+                new FetchResponseData.PartitionData()
```
ijuma, line 839: Do we need to set the partition id here? There are a few other cases in this file that are similar.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -1146,38 +1159,47 @@ private FetchRequest createFetchRequest(int version) {
...
-            1000000, FetchResponse.INVALID_LAST_STABLE_OFFSET, 0L, Optional.empty(), abortedTransactions, MemoryRecords.EMPTY));
-        return new FetchResponse<>(Errors.NONE, responseData, 25, sessionId);
-    }
-
-    private FetchResponse<MemoryRecords> createFetchResponse(boolean includeAborted) {
-        LinkedHashMap<TopicPartition, FetchResponse.PartitionData<MemoryRecords>> responseData = new LinkedHashMap<>();
+        responseData.put(new TopicPartition("test", 0), new FetchResponseData.PartitionData()
+                        .setHighWatermark(1000000)
+                        .setLogStartOffset(0)
+                        .setAbortedTransactions(Collections.emptyList())
```
ijuma: No need to set aborted transactions.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -1146,38 +1159,47 @@ private FetchRequest createFetchRequest(int version) {
...
+        return FetchResponse.of(Errors.NONE, 25, sessionId, responseData);
+    }
+
+    private FetchResponse createFetchResponse(boolean includeAborted) {
+        LinkedHashMap<TopicPartition, FetchResponseData.PartitionData> responseData = new LinkedHashMap<>();
         MemoryRecords records = MemoryRecords.withRecords(CompressionType.NONE, new SimpleRecord("blah".getBytes()));
+        responseData.put(new TopicPartition("test", 0), new FetchResponseData.PartitionData()
+                        .setHighWatermark(1000000)
+                        .setLogStartOffset(0)
+                        .setAbortedTransactions(Collections.emptyList())
```
ijuma: Aborted transactions is empty by default.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
```diff
@@ -416,9 +412,8 @@ abstract class AbstractFetcherThread(name: String,
                        "expected to persist.")
                   partitionsWithError += topicPartition
 
-                case _ =>
-                  error(s"Error for partition $topicPartition at offset ${currentFetchState.fetchOffset}",
-                    partitionData.error.exception)
+                case partitionError: Errors =>
```
ijuma: We should not have `: Errors` here as it introduces a type test. What we want is for this to be a catch all.

core/src/test/scala/unit/kafka/server/AbstractFetcherThreadTest.scala
```diff
@@ -1144,8 +1143,14 @@ class AbstractFetcherThreadTest {
...
         }
 
-        (partition, new FetchData(error, leaderState.highWatermark, leaderState.highWatermark, leaderState.logStartOffset,
-          Optional.empty[Integer], List.empty.asJava, divergingEpoch.asJava, records))
+        (partition, new FetchResponseData.PartitionData()
+          .setErrorCode(error.code)
+          .setHighWatermark(leaderState.highWatermark)
+          .setLastStableOffset(leaderState.highWatermark)
+          .setLogStartOffset(leaderState.logStartOffset)
+          .setAbortedTransactions(Collections.emptyList())
```
ijuma: Redundant.

core/src/test/scala/unit/kafka/server/AbstractFetcherThreadTest.scala
```diff
@@ -1144,8 +1143,14 @@ class AbstractFetcherThreadTest {
...
+          .setRecords(records)
+          .setDivergingEpoch(divergingEpoch.getOrElse(new FetchResponseData.EpochEndOffset)))
```
ijuma: The previous code did `asJava`, why did we change it?

core/src/test/scala/unit/kafka/server/ReplicaFetcherThreadTest.scala
```diff
@@ -963,9 +963,11 @@ class ReplicaFetcherThreadTest {
 
     val records = MemoryRecords.withRecords(CompressionType.NONE,
       new SimpleRecord(1000, "foo".getBytes(StandardCharsets.UTF_8)))
-
-    val partitionData: thread.FetchData = new FetchResponse.PartitionData[Records](
-      Errors.NONE, 0, 0, 0, Optional.empty(), Collections.emptyList(), records)
+    val partitionData: thread.FetchData = new FetchResponseData.PartitionData()
+        .setLastStableOffset(0)
+        .setLogStartOffset(0)
+        .setAbortedTransactions(Collections.emptyList())
```
ijuma: Redundant.

core/src/test/scala/unit/kafka/server/ReplicaFetcherThreadTest.scala
```diff
@@ -531,10 +529,12 @@ class ReplicaFetcherThreadTest {
...
 
-    def partitionData(divergingEpoch: FetchResponseData.EpochEndOffset): FetchResponse.PartitionData[Records] = {
-      new FetchResponse.PartitionData[Records](
-        Errors.NONE, 0, 0, 0, Optional.empty(), Collections.emptyList(),
-        Optional.of(divergingEpoch), MemoryRecords.EMPTY)
+    def partitionData(divergingEpoch: FetchResponseData.EpochEndOffset): FetchResponseData.PartitionData = {
+      new FetchResponseData.PartitionData()
+          .setLastStableOffset(0)
+          .setLogStartOffset(0)
+          .setAbortedTransactions(Collections.emptyList())
```
ijuma: Redundant.

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/common/FetchResponseBenchmark.java
```diff
@@ -78,19 +78,23 @@ public void setup() {
         for (int topicIdx = 0; topicIdx < topicCount; topicIdx++) {
             String topic = UUID.randomUUID().toString();
             for (int partitionId = 0; partitionId < partitionCount; partitionId++) {
-                FetchResponse.PartitionData<MemoryRecords> partitionData = new FetchResponse.PartitionData<>(
-                    Errors.NONE, 0, 0, 0, Optional.empty(), Collections.emptyList(), records);
+                FetchResponseData.PartitionData partitionData = new FetchResponseData.PartitionData()
+                                .setPartitionIndex(partitionId)
+                                .setLastStableOffset(0)
+                                .setLogStartOffset(0)
+                                .setAbortedTransactions(Collections.emptyList())
```
ijuma: Redundant.

jmh-benchmarks/src/main/java/org/apache/kafka/jmh/fetcher/ReplicaFetcherThreadBenchmark.java
```diff
@@ -174,8 +173,12 @@ public int sizeInBytes() {
                     return null;
                 }
             };
-            initialFetched.put(tp, new FetchResponse.PartitionData<>(Errors.NONE, 0, 0, 0,
-                    new LinkedList<>(), fetched));
+            initialFetched.put(tp, new FetchResponseData.PartitionData()
+                    .setPartitionIndex(tp.partition())
+                    .setLastStableOffset(0)
+                    .setLogStartOffset(0)
+                    .setAbortedTransactions(Collections.emptyList())
```
ijuma: Redundant.

core/src/test/scala/unit/kafka/server/FetchSessionTest.scala
```diff
@@ -679,17 +760,24 @@ class FetchSessionTest {
...
 
     // Partitions that meet other conditions should be returned regardless of whether
     // divergingEpoch is set or not.
-    respData.put(tp1, new FetchResponse.PartitionData(Errors.NONE,
-      110, 110, 0, Optional.empty(), Collections.emptyList(), Optional.empty(), null))
+    respData.put(tp1, new FetchResponseData.PartitionData()
+        .setHighWatermark(110)
+        .setLastStableOffset(110)
+        .setLogStartOffset(0)
+        .setAbortedTransactions(Collections.emptyList()))
```
ijuma: Redundant.

core/src/test/scala/unit/kafka/server/FetchSessionTest.scala
```diff
@@ -679,17 +760,24 @@ class FetchSessionTest {
     assertEquals(Collections.singleton(tp2), resp2.responseData.keySet)
 
     // All partitions with divergent epoch should be returned.
-    respData.put(tp1, new FetchResponse.PartitionData(Errors.NONE,
-      105, 105, 0, Optional.empty(), Collections.emptyList(), divergingEpoch, null))
+    respData.put(tp1, new FetchResponseData.PartitionData()
+          .setHighWatermark(105)
+          .setLastStableOffset(105)
+          .setLogStartOffset(0)
+          .setAbortedTransactions(Collections.emptyList())
```
ijuma: Redundant.

core/src/test/scala/unit/kafka/server/FetchSessionTest.scala
```diff
@@ -658,12 +732,19 @@ class FetchSessionTest {
...
+        .setHighWatermark(105)
+        .setLastStableOffset(105)
+        .setLogStartOffset(0)
+        .setAbortedTransactions(Collections.emptyList()))
+    val divergingEpoch = new FetchResponseData.EpochEndOffset().setEpoch(3).setEndOffset(90)
+    respData.put(tp2, new FetchResponseData.PartitionData()
+        .setHighWatermark(105)
+        .setLastStableOffset(105)
+        .setLogStartOffset(0)
+        .setAbortedTransactions(Collections.emptyList())
```
ijuma: Redundant.

core/src/test/scala/unit/kafka/server/FetchSessionTest.scala
```diff
@@ -658,12 +732,19 @@ class FetchSessionTest {
...
-      105, 105, 0, Optional.empty(), Collections.emptyList(), Optional.empty(), null))
-    val divergingEpoch = Optional.of(new FetchResponseData.EpochEndOffset().setEpoch(3).setEndOffset(90))
-    respData.put(tp2, new FetchResponse.PartitionData(Errors.NONE,
-      105, 105, 0, Optional.empty(), Collections.emptyList(), divergingEpoch, null))
+    val respData = new util.LinkedHashMap[TopicPartition, FetchResponseData.PartitionData]
+    respData.put(tp1, new FetchResponseData.PartitionData()
+        .setHighWatermark(105)
+        .setLastStableOffset(105)
+        .setLogStartOffset(0)
+        .setAbortedTransactions(Collections.emptyList()))
```
ijuma: Redundant.

core/src/test/scala/unit/kafka/server/FetchSessionTest.scala
```diff
@@ -534,15 +588,21 @@ class FetchSessionTest {
...
-      Errors.NONE, 10, 10, 10, null, null))
+    val session2RespData = new util.LinkedHashMap[TopicPartition, FetchResponseData.PartitionData]
+    session2RespData.put(new TopicPartition("foo", 0),
+      new FetchResponseData.PartitionData()
+        .setHighWatermark(100)
+        .setLastStableOffset(100)
+        .setLogStartOffset(100))
+    session2RespData.put(new TopicPartition("foo", 1),
+      new FetchResponseData.PartitionData()
+        .setHighWatermark(10)
```
ijuma, line 623: Do we need to set the partition id here and other cases?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -761,79 +754,84 @@ class KafkaApis(val requestChannel: RequestChannel,
...
```
ijuma: Seems that we could set the diverging offset only if set and leave the default otherwise.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -365,17 +126,92 @@ public int sessionId() {
...
+    }
+
+    /**
+     * cast the BaseRecords of PartitionData to Records. KRPC converts the byte array to MemoryRecords so this method
+     * never fail if the data is from KRPC.
+     *
+     * @param partition partition data
+     * @return Records or empty record if the records in PartitionData is null.
+     */
+    public static Records records(FetchResponseData.PartitionData partition) {
```
ijuma: There is one place in this PR that we check for null when computing the records size, maybe we can use this utility function there.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -365,17 +126,92 @@ public int sessionId() {
...
+    public static FetchResponseData.PartitionData partitionResponse(int partition, Errors error) {
+        return new FetchResponseData.PartitionData()
+            .setPartitionIndex(partition)
+            .setErrorCode(error.code())
+            .setHighWatermark(FetchResponse.INVALID_HIGH_WATERMARK);
+    }
+
+    /**
+     * cast the BaseRecords of PartitionData to Records. KRPC converts the byte array to MemoryRecords so this method
+     * never fail if the data is from KRPC.
```
ijuma: Suggestion:

```text
Returns `partition.records` as `Records` (instead of `BaseRecords`). If `records` is `null`, returns `MemoryRecords.EMPTY`.

If this response was deserialized after a fetch, this method should never fail. An example where this would fail is a down-converted response (e.g. LazyDownConversionRecords) on the broker (before it's serialized and sent on the wire).
```
```

File: clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java [hunk 54]
```diff
@@ -365,17 +126,92 @@ public int sessionId() {
...
+     *
+     * @param partition partition data
+     * @return Records or empty record if the records in PartitionData is null.
+     */
+    public static Records records(FetchResponseData.PartitionData partition) {
+        return partition.records() == null ? MemoryRecords.EMPTY : (Records) partition.records();
```
ijuma: Instead of casting blindly, can we include a reasonable error message if the cast fails?

ijuma (commented):
Thanks for the updates. A few more comments below.

core/src/main/scala/kafka/server/AbstractFetcherThread.scala
chia7712: you are right. fixed

core/src/test/scala/unit/kafka/server/AbstractFetcherThreadTest.scala
chia7712: the type `FetchablePartitionResponse` is gone and the replacement `FetchResponseData.PartitionData` (generated data) can't accept `Optional` type.

core/src/test/scala/unit/kafka/server/FetchSessionTest.scala
chia7712: good catch. will fix all similar issues in next commit.

core/src/main/scala/kafka/server/KafkaApis.scala
chia7712: make sense. fixed

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: good one. will copy that

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: will copy that

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: good point. will copy that

chia7712:
@ijuma Thanks for all your great comments. I have updated this PR. Please take a look.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -365,17 +126,98 @@ public int sessionId() {
...
+     * fail is a down-converted response (e.g. LazyDownConversionRecords) on the broker (before it's serialized and
+     * sent on the wire).
+     *
+     * @param partition partition data
+     * @return Records or empty record if the records in PartitionData is null.
+     */
+    public static Records records(FetchResponseData.PartitionData partition) {
+        if (partition.records() == null) return MemoryRecords.EMPTY;
+        else if (partition.records() instanceof Records) return (Records) partition.records();
+        else throw new IllegalStateException("the record type is " + partition.records().getClass().getSimpleName() +
```
ijuma: No `else` needed since we used `return` for both other cases. For the exception, I think we can just throw `ClassCastException` since `IllegalStateException` doesn't fit very well for this case. I would also make the message a bit more generic to avoid it going stale when we add more `Records` subtypes. For example:

```java
"The record type is " + partition.records().getClass().getSimpleName() + ", which is not a subtype of " +
Records.class.getSimpleName() + ". This method is only safe to call if the `FetchResponse` was
deserialized from bytes."

File: clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java [hunk 56]
```diff
@@ -791,15 +791,17 @@ public void produceRequestGetErrorResponseTest() {
...
 
         MemoryRecords records = MemoryRecords.readableRecords(ByteBuffer.allocate(10));
-        responseData.put(new TopicPartition("test", 0), new FetchResponse.PartitionData<>(
-                Errors.NONE, 1000000, FetchResponse.INVALID_LAST_STABLE_OFFSET,
-                0L, Optional.empty(), Collections.emptyList(), records));
-
-        FetchResponse<MemoryRecords> v0Response = new FetchResponse<>(Errors.NONE, responseData, 0, INVALID_SESSION_ID);
-        FetchResponse<MemoryRecords> v1Response = new FetchResponse<>(Errors.NONE, responseData, 10, INVALID_SESSION_ID);
+        responseData.put(new TopicPartition("test", 0),
+                new FetchResponseData.PartitionData()
```
ijuma, line 809: Set partition id.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -1146,38 +1160,45 @@ private FetchRequest createFetchRequest(int version) {
...
-            new FetchResponse.AbortedTransaction(234L, 999L));
-        responseData.put(new TopicPartition("test", 1), new FetchResponse.PartitionData<>(Errors.NONE,
-            1000000, FetchResponse.INVALID_LAST_STABLE_OFFSET, 0L, Optional.empty(), abortedTransactions, MemoryRecords.EMPTY));
-        return new FetchResponse<>(Errors.NONE, responseData, 25, sessionId);
-    }
-
-    private FetchResponse<MemoryRecords> createFetchResponse(boolean includeAborted) {
-        LinkedHashMap<TopicPartition, FetchResponse.PartitionData<MemoryRecords>> responseData = new LinkedHashMap<>();
+        responseData.put(new TopicPartition("test", 0), new FetchResponseData.PartitionData()
+                        .setHighWatermark(1000000)
```
ijuma, line 1184: Set partition id.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -1146,38 +1160,45 @@ private FetchRequest createFetchRequest(int version) {
...
+                        .setLogStartOffset(0)
+                        .setRecords(records));
+        List<FetchResponseData.AbortedTransaction> abortedTransactions = Collections.singletonList(
+            new FetchResponseData.AbortedTransaction().setProducerId(234L).setFirstOffset(999L));
+        responseData.put(new TopicPartition("test", 1), new FetchResponseData.PartitionData()
+                        .setHighWatermark(1000000)
```
ijuma, line 1191: Set partition id.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -1146,38 +1160,45 @@ private FetchRequest createFetchRequest(int version) {
...
+                        .setLogStartOffset(0)
+                        .setAbortedTransactions(abortedTransactions));
+        return FetchResponse.of(Errors.NONE, 25, sessionId, responseData);
+    }
+
+    private FetchResponse createFetchResponse(boolean includeAborted) {
+        LinkedHashMap<TopicPartition, FetchResponseData.PartitionData> responseData = new LinkedHashMap<>();
         MemoryRecords records = MemoryRecords.withRecords(CompressionType.NONE, new SimpleRecord("blah".getBytes()));
+        responseData.put(new TopicPartition("test", 0), new FetchResponseData.PartitionData()
+                        .setHighWatermark(1000000)
```
ijuma, line 1202: Set partition id.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -1146,38 +1160,45 @@ private FetchRequest createFetchRequest(int version) {
...
+        List<FetchResponseData.AbortedTransaction> abortedTransactions = Collections.emptyList();
         if (includeAborted) {
             abortedTransactions = Collections.singletonList(
-                    new FetchResponse.AbortedTransaction(234L, 999L));
+                    new FetchResponseData.AbortedTransaction().setProducerId(234L).setFirstOffset(999L));
         }
-        responseData.put(new TopicPartition("test", 1), new FetchResponse.PartitionData<>(Errors.NONE,
-                1000000, FetchResponse.INVALID_LAST_STABLE_OFFSET, 0L, Optional.empty(), abortedTransactions, MemoryRecords.EMPTY));
+        responseData.put(new TopicPartition("test", 1), new FetchResponseData.PartitionData()
+                        .setHighWatermark(1000000)
```
ijuma, line 1213: Set partition id.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -761,79 +754,85 @@ class KafkaApis(val requestChannel: RequestChannel,
...
 
         // Prepare fetch response from converted data
-        val response = new FetchResponse(unconvertedFetchResponse.error, convertedData, throttleTimeMs,
-          unconvertedFetchResponse.sessionId)
+        val response = FetchResponse.of(unconvertedFetchResponse.error, throttleTimeMs, unconvertedFetchResponse.sessionId, convertedData)
         // record the bytes out metrics only when the response is being sent
         response.responseData.forEach { (tp, data) =>
-          brokerTopicStats.updateBytesOut(tp.topic, fetchRequest.isFromFollower, reassigningPartitions.contains(tp), data.records.sizeInBytes)
+          brokerTopicStats.updateBytesOut(tp.topic, fetchRequest.isFromFollower,
+            reassigningPartitions.contains(tp), if (data.records  == null) 0 else data.records.sizeInBytes)
```
ijuma: We can use the utility method you added to avoid the null check.

ijuma (commented):
Thanks for the updates. I think these are the last comments and then we should be good. :)

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
ijuma: One more thing, let's call this `recordsOrFail` to make it clear that the operation is not necessarily safe.

chia7712: Will address those nice comments

core/src/main/scala/kafka/server/KafkaApis.scala
chia7712: The converted record is `LazyDownConversionRecords` so casting it to `Records` can produce error.

ijuma: Ah, I thought the `size` method was in `FetchResponse`. Maybe we should add one (`recordsSize` or something)? That would always be safe to call.

chia7712: > That would always be safe to call.

make sense. Will copy that

chia7712:
@ijuma thanks for your comments. I have addressed them in latest commit (https://github.com/apache/kafka/pull/9758/commits/55b35ab5764d8129e23f386229eb350603b7b0aa). Please take a look.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -365,17 +126,105 @@ public int sessionId() {
...
      */
-    public static <T extends BaseRecords> int sizeOf(short version,
-                                                     Iterator<Map.Entry<TopicPartition, PartitionData<T>>> partIterator) {
+    public static int sizeOf(short version,
+                             Iterator<Map.Entry<TopicPartition, FetchResponseData.PartitionData>> partIterator) {
         // Since the throttleTimeMs and metadata field sizes are constant and fixed, we can
         // use arbitrary values here without affecting the result.
-        FetchResponseData data = toMessage(0, Errors.NONE, partIterator, INVALID_SESSION_ID);
+        LinkedHashMap<TopicPartition, FetchResponseData.PartitionData> data = new LinkedHashMap<>();
+        partIterator.forEachRemaining(entry -> data.put(entry.getKey(), entry.getValue()));
```
ijuma: Is this an additional copy compared to previous behavior?

ijuma (approved):
I left one comment, the rest LGTM.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: nice question. will revert to previous code :)

chia7712:
```
Build / JDK 11 / org.apache.kafka.connect.integration.BlockingConnectorTest.testWorkerRestartWithBlockInConnectorStop
```

unrelated flaky

ijuma:
Great to see this merged. @chia7712 Will you follow up with the items we split into separate JIRAs? I think the 3 key ones are:
1. Remove unnecessary copying, if possible, in KafkaApis
2. Remove the lazily populated map in FetchResponse
3. I think you had an idea for removing `FetchResponse.toMessage`

chia7712:
> Remove unnecessary copying, if possible, in KafkaApis

https://issues.apache.org/jira/browse/KAFKA-12385

> Remove the lazily populated map in FetchResponse

https://issues.apache.org/jira/browse/KAFKA-12387

> I think you had an idea for removing FetchResponse.toMessage

ugh, I missed this one. opened: https://issues.apache.org/jira/browse/KAFKA-12410

jolshan:
@chia7712  I noticed that all of these JIRAs are listed as open, but they reference this PR. Are there further changes to be made? Asking because I'm working on similar code for https://github.com/apache/kafka/pull/9944 and want to minimize merge conflicts as best as possible.

chia7712:
>  I noticed that all of these JIRAs are listed as open, but they reference this PR. Are there further changes to be made? Asking because I'm working on similar code for #9944 and want to minimize merge conflicts as best as possible.

thanks for this response. Only #10269 is in progress. I'm sorry that this PR brings a bunch of conflicting files to #9944. I'd like to merge #9944 before other minor refactor-related patches.

jolshan:
@chia7712 sounds good. So is the plan to merge https://github.com/apache/kafka/pull/10269 first or after https://github.com/apache/kafka/pull/9944? Also thanks for making these changes. I noticed a lot of this when working on https://github.com/apache/kafka/pull/9944

chia7712:
> So is the plan to merge #10269 first or after #9944? 

yep. #10269 is in progress (tweak performance). I will take a look at #9944 :)

ijuma:
Does the refactoring simplify #9944 or not?

jolshan:
@ijuma on first glance, I'm not sure if it simplifies. But it will introduce changes. And these are likely changes we want to have eventually.

jolshan:
So should we go forward with https://github.com/apache/kafka/pull/10269 first still?

chia7712:
> So should we go forward with #10269 first still?

It needs some time to complete performance tests. It seems to me we can fix conflicting files in #9944 to get reviews and then push it before merging #10269. BTW, don't worry about causing potential conflicting files to #10269 :)

jolshan:
@chia7712 I see. Thanks for the response. I'm not sure if https://github.com/apache/kafka/pull/9944 will be ready after fixing conflicts. There were still a few critical tests failing. I can try to fix those but still figuring out the timeline. I can go ahead and start with the conflicts though.

PR #9762, opened 2020-12-16
KAFKA-10861; Fix race condition in flaky test `testFencingOnSendOffsets`

hachikuji:
I wasn't able to reproduce the failure locally, but it looks like there is a race condition with the sending of the records in the first producer. The test case assumes that these records have been completed before the call to `sendOffsetsToTransaction`, but they very well might not be. It is even possible for the writes from the second producer to arrive first which would then result in the test failure that we are seeing. The solution is to force the send with `flush()`.


abbccdda (approved):
LGTM

guozhangwang:
LGTM

PR #9764, opened 2020-12-17
MINOR: Eliminate KafkaScheduler#scheduleOnce in favor of KafkaScheduler#schedule

kowshik:
In this PR, I've eliminated `KafkaScheduler#scheduleOnce` API. It had only 1 call site, and instead we can just use `KafkaScheduler#schedule`.

**Test plan:**
Rely on existing tests.

### DIFF
--- a/core/src/main/scala/kafka/utils/KafkaScheduler.scala
+++ b/core/src/main/scala/kafka/utils/KafkaScheduler.scala
@@ -99,10 +99,6 @@ class KafkaScheduler(val threads: Int,
     }
   }
 
-  def scheduleOnce(name: String, fun: () => Unit, delay: Long = 0L, unit: TimeUnit = TimeUnit.MILLISECONDS): Unit = {
-    schedule(name, fun, delay = delay, period = -1L, unit = unit)
-  }
-
   def schedule(name: String, fun: () => Unit, delay: Long, period: Long, unit: TimeUnit): ScheduledFuture[_] = {
     debug("Scheduling task %s with initial delay %d ms and period %d ms."
         .format(name, TimeUnit.MILLISECONDS.convert(delay, unit), TimeUnit.MILLISECONDS.convert(period, unit)))
--- a/core/src/main/scala/kafka/utils/KafkaScheduler.scala
+++ b/core/src/main/scala/kafka/utils/KafkaScheduler.scala
@@ -99,8 +99,8 @@ class KafkaScheduler(val threads: Int,
     }
   }
 
-  def scheduleOnce(name: String, fun: () => Unit): Unit = {
-    schedule(name, fun, delay = 0L, period = -1L, unit = TimeUnit.MILLISECONDS)
+  def scheduleOnce(name: String, fun: () => Unit, delay: Long = 0L, unit: TimeUnit = TimeUnit.MILLISECONDS): Unit = {
+    schedule(name, fun, delay = delay, period = -1L, unit = unit)
   }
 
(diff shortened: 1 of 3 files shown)


kowshik:
cc @junrao @chia7712 @hachikuji for review

core/src/main/scala/kafka/utils/KafkaScheduler.scala
```diff
@@ -99,8 +99,8 @@ class KafkaScheduler(val threads: Int,
     }
   }
 
-  def scheduleOnce(name: String, fun: () => Unit): Unit = {
-    schedule(name, fun, delay = 0L, period = -1L, unit = TimeUnit.MILLISECONDS)
+  def scheduleOnce(name: String, fun: () => Unit, delay: Long = 0L, unit: TimeUnit = TimeUnit.MILLISECONDS): Unit = {
```
chia7712: It seems to me we should remove ```scheduleOnce``` as it is equal to parent ```schedule`` (with default value)

see https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/utils/KafkaScheduler.scala#L56

core/src/main/scala/kafka/zookeeper/ZooKeeperClient.scala
```diff
@@ -430,10 +430,10 @@ class ZooKeeperClient(connectString: String,
 
   // Visibility for testing
   private[zookeeper] def scheduleReinitialize(name: String, message: String, delayMs: Long): Unit = {
-    reinitializeScheduler.schedule(name, () => {
+    reinitializeScheduler.scheduleOnce(name, () => {
```
chia7712: How about calling ```schedule```? ```period``` and ```unit``` are equal to default value.

chia7712 (commented):
@kowshik Thanks for your patch. Some minor comments are left. Please take a look.

core/src/main/scala/kafka/utils/KafkaScheduler.scala
kowshik: Good point. Done.

core/src/main/scala/kafka/zookeeper/ZooKeeperClient.scala
kowshik: Good point. Done.

kowshik:
@chia7712: Thanks for the review! I've addressed your comments in b21762a.

chia7712:
just another thought. the ```period``` can change the behavior of ```schedule```. It seems to me we should NOT give the default value to it as it is error-prone. How about moving the ```scheduleOnce``` to parent and follow your first commit to add ```delay``` and ```unit``` to ```scheduleOnce```? WDYT

kowshik:
@chia7712: In general, either approach seems better than what we have in the code right now. In the original approach I had in the PR, we made the existing code better and improved readability. Then, in the approach you proposed we eliminated an API. So actually I don't have a strong preference for either of these, but if I were to pick one I'd go for my original approach that improved `scheduleOnce` with additional args since it helps readability.

Let me know.

chia7712:
@kowshik Thanks for your response. You are right and the ```scheduleOnce``` is more readable. Also, I'd like to have following improvement.

1. remove default value of ```period``` (The callers should call ```scheduleOnce``` instead of setting -1 to period)
2. move scheduleOnce from ```KafkaScheduler``` to ```Scheduler``` ( it can have default implementation by calling ```schedule```)

PR #9765, opened 2020-12-18
KAFKA-10763: Fix incomplete cooperative rebalances preventing connector/task revocations

gharris1727:
[Jira](https://issues.apache.org/jira/browse/KAFKA-10763)
When two cooperative rebalances take place soon after one another, a prior rebalance may not complete before the next rebalance is started.
Under Eager rebalancing, no tasks would have been started, so the subsequent onRevoked call is intentionally skipped whenever rebalanceResolved was false.
Under Cooperative rebalancing, the same logic causes the DistributedHerder to skip stopping all of the connector/task revocations which occur in the second rebalance.
The DistributedHerder still removes the revoked connectors/tasks from its assignment, so that the DistributedHerder and Worker have different knowledge of running connectors/tasks.
This causes the connector/task instances that would have been stopped to disappear from the rebalance protocol, and left running until their workers are halted, or they fail.
Connectors/Tasks which were then reassigned to other workers by the rebalance protocol would be duplicated, and run concurrently with zombie connectors/tasks.
Connectors/Tasks which were reassigned back to the same worker would encounter exceptions in Worker, indicating that the connector/task existed and was already running.

* Add a test for revoking and then reassigning a connector under normal circumstances
* Add a test for revoking and then reassigning a connector following an incomplete cooperative rebalance
* Change expectRebalance to make assignment fields mutable before passing them into the DistributedHerder
* Only skip revocation for the Eager protocol, and never skip revocation for cooperative/sessioned protocols

This should be backported to all branches with Incremental Cooperative Rebalancing, 2.3+ if possible. This is a serious consistency bug, and the fix should probably be applied widely.

Signed-off-by: Greg Harris <gregh@confluent.io>

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
@@ -93,6 +93,7 @@ import java.util.stream.Collectors;
 import static org.apache.kafka.connect.runtime.WorkerConfig.TOPIC_TRACKING_ENABLE_CONFIG;
 import static org.apache.kafka.connect.runtime.distributed.ConnectProtocol.CONNECT_PROTOCOL_V0;
 import static org.apache.kafka.connect.runtime.distributed.ConnectProtocolCompatibility.EAGER;
+import static org.apache.kafka.connect.runtime.distributed.IncrementalCooperativeConnectProtocol.CONNECT_PROTOCOL_V1;
 import static org.apache.kafka.connect.runtime.distributed.IncrementalCooperativeConnectProtocol.CONNECT_PROTOCOL_V2;
 
 /**
@@ -1740,7 +1741,7 @@ public class DistributedHerder extends AbstractHerder implements Runnable {
             // Note that since we don't reset the assignment, we don't revoke leadership here. During a rebalance,
             // it is still important to have a leader that can write configs, offsets, etc.
 
-            if (rebalanceResolved) {
+            if (rebalanceResolved || currentProtocolVersion >= CONNECT_PROTOCOL_V1) {
                 List<Callable<Void>> callables = new ArrayList<>();
                 for (final String connectorName : connectors) {
                     callables.add(getConnectorStoppingCallable(connectorName));
--- a/connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
+++ b/connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
@@ -567,6 +567,112 @@ public class DistributedHerderTest {
     }
 
     @Test
+    public void testRevoke() throws TimeoutException {
+        revokeAndReassign(false);
+    }
+
+    @Test
+    public void testIncompleteRebalanceBeforeRevoke() throws TimeoutException {
+        revokeAndReassign(true);
+    }
+
+    public void revokeAndReassign(boolean incompleteRebalance) throws TimeoutException {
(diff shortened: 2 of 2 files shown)


gharris1727:
Note: as-is, this PR does not reproduce the exact exceptions noted in the JIRA, as the Worker is mocked out for this test. I believe that it's sufficient to show that without this fix, the DistributedHerder::onRevoked does not stop the tasks. This assumes that if we don't stop the task in the Worker (and the task does not fail), that a subsequent task-start would cause the Worker to throw the exception found in the JIRA.

cc @C0urante @ncliang PTAL, thanks!

connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
```diff
@@ -566,6 +566,112 @@ public Boolean answer() throws Throwable {
...
+    public void testRevoke() throws TimeoutException {
+        revokeAndReassign(false);
+    }
+
+    @Test
+    public void testIncompleteRebalanceBeforeRevoke() throws TimeoutException {
+        revokeAndReassign(true);
+    }
+
+    public void revokeAndReassign(boolean incompleteRebalance) throws TimeoutException {
```
ncliang, line 579: It's hard to tell if this actually reproduces the issue or not due to the heavy mocking required. Is there a more direct way to reproduce? Maybe in `RebalanceSourceConnectorsIntegrationTest` or similar? Even if the IT ends up being flaky, having that repro would boost confidence in this fix.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
```diff
@@ -1740,7 +1741,7 @@ public void onRevoked(String leader, Collection<String> connectors, Collection<C
             // Note that since we don't reset the assignment, we don't revoke leadership here. During a rebalance,
             // it is still important to have a leader that can write configs, offsets, etc.
 
-            if (rebalanceResolved) {
+            if (rebalanceResolved || currentProtocolVersion >= CONNECT_PROTOCOL_V1) {
```
ncliang, line 1744: Maybe add a comment explaining why the additional check is needed.

ncliang (approved):
Great work @gharris1727 ! Just a question about reproing the issue in IT, otherwise LGTM.

connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
gharris1727: Yeah, this test was a bit difficult to wrap my head around at first, but I think it's the best way to target this section of the code. I don't believe that adding a new flakey test is prudent, and making a non-flakey test with less mocks might end up to be harder to follow than this mocked test.

I think what _would_ be a good test to add would be a variant which replaced this contrived reading-config-offset-topic-failure with a [genuine WakeupException thrown from the end of tick](https://github.com/apache/kafka/blob/ac7b5d3389fddf46bc53ab656de1fa7e2562efdb/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L433-L443), which I believe is the true root cause of this issue most of the time. This is not easy with the boilerplate in this test as-is, and requires a little bit of refactoring to set up the rebalance during that block.

kkonstantine (approved):
Thanks for the fix @gharris1727 
LGTM

kkonstantine:
jdk 8 and jdk 15 tests green. A single failure in jdk 11 is on a streams, possibly flaky, test.

PR #9766, opened 2020-12-18
KAFKA-10864: Convert end txn marker schema to use auto-generated protocol

dengziming:
1. Convert end txn marker schema to use auto-generated protocol `EndTxnMarker`
2. substitute `CURRENT_END_TXN_MARKER_VALUE_SIZE` with an `endTnxMarkerValueSize` method since the size is accumulated from `EndTxnMarker`.
3. add buffer to `EndTransactionMarker` to avoid twice compute from `serializeValue` and `endTnxMarkerValueSize`.
4. flexibleVersions is set to none.

A unit test to verify serialize and deserialize

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/record/EndTransactionMarker.java
+++ b/clients/src/main/java/org/apache/kafka/common/record/EndTransactionMarker.java
@@ -16,14 +16,9 @@
  */
 package org.apache.kafka.common.record;
 
-import org.apache.kafka.common.InvalidRecordException;
-import org.apache.kafka.common.protocol.types.Field;
-import org.apache.kafka.common.protocol.types.Schema;
-import org.apache.kafka.common.protocol.types.Struct;
-import org.apache.kafka.common.protocol.types.Type;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.apache.kafka.common.message.EndTxnMarker;
+import org.apache.kafka.common.protocol.ByteBufferAccessor;
+import org.apache.kafka.common.protocol.MessageUtil;
 
 import java.nio.ByteBuffer;
 
@@ -33,25 +28,17 @@ import java.nio.ByteBuffer;
  * value embeds information useful for write validation (for now, just the coordinator epoch).
  */
 public class EndTransactionMarker {
-    private static final Logger log = LoggerFactory.getLogger(EndTransactionMarker.class);
-
-    private static final short CURRENT_END_TXN_MARKER_VERSION = 0;
-    private static final Schema END_TXN_MARKER_SCHEMA_VERSION_V0 = new Schema(
-            new Field("version", Type.INT16),
-            new Field("coordinator_epoch", Type.INT32));
-    static final int CURRENT_END_TXN_MARKER_VALUE_SIZE = 6;
-    static final int CURRENT_END_TXN_SCHEMA_RECORD_SIZE = DefaultRecord.sizeInBytes(0, 0L,
-            ControlRecordType.CURRENT_CONTROL_RECORD_KEY_SIZE,
-            EndTransactionMarker.CURRENT_END_TXN_MARKER_VALUE_SIZE,
-            Record.EMPTY_HEADERS);
 
     private final ControlRecordType type;
     private final int coordinatorEpoch;
+    private final ByteBuffer buffer;
 
     public EndTransactionMarker(ControlRecordType type, int coordinatorEpoch) {
         ensureTransactionMarkerControlType(type);
         this.type = type;
         this.coordinatorEpoch = coordinatorEpoch;
+        EndTxnMarker marker = new EndTxnMarker().setCoordinatorEpoch(coordinatorEpoch);
+        this.buffer = MessageUtil.toVersionPrefixedByteBuffer(EndTxnMarker.HIGHEST_SUPPORTED_VERSION, marker);
     }
 
     public int coordinatorEpoch() {
@@ -62,19 +49,8 @@ public class EndTransactionMarker {
         return type;
     }
 
-    private Struct buildRecordValue() {
-        Struct struct = new Struct(END_TXN_MARKER_SCHEMA_VERSION_V0);
-        struct.set("version", CURRENT_END_TXN_MARKER_VERSION);
-        struct.set("coordinator_epoch", coordinatorEpoch);
-        return struct;
-    }
-
     public ByteBuffer serializeValue() {
-        Struct valueStruct = buildRecordValue();
-        ByteBuffer value = ByteBuffer.allocate(valueStruct.sizeOf());
-        valueStruct.writeTo(value);
-        value.flip();
-        return value;
+        return buffer.duplicate();
     }
 
     @Override
@@ -95,7 +71,7 @@ public class EndTransactionMarker {
 
     private static void ensureTransactionMarkerControlType(ControlRecordType type) {
         if (type != ControlRecordType.COMMIT && type != ControlRecordType.ABORT)
-            throw new IllegalArgumentException("Invalid control record type for end transaction marker" + type);
+            throw new IllegalArgumentException("Invalid control record type for end transaction marker " + type);
     }
 
     public static EndTransactionMarker deserialize(Record record) {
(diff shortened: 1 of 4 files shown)


clients/src/main/java/org/apache/kafka/common/record/MemoryRecords.java
```diff
@@ -618,8 +618,7 @@ public static MemoryRecords withEndTransactionMarker(long timestamp, long produc
     public static MemoryRecords withEndTransactionMarker(long initialOffset, long timestamp, int partitionLeaderEpoch,
                                                          long producerId, short producerEpoch,
                                                          EndTransactionMarker marker) {
-        int endTxnMarkerBatchSize = DefaultRecordBatch.RECORD_BATCH_OVERHEAD +
-                EndTransactionMarker.CURRENT_END_TXN_SCHEMA_RECORD_SIZE;
+        int endTxnMarkerBatchSize = DefaultRecordBatch.RECORD_BATCH_OVERHEAD + marker.endTnxMarkerValueSize();
```
abbccdda: s/endTnxMarkerValueSize/endTxnMarkerValueSize

clients/src/main/resources/common/message/EndTxnMarker.json
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+{
+  "type": "data",
+  "name": "EndTxnMarker",
+  "validVersions": "0",
+  "fields": [
+    { "name": "CoordinatorEpoch", "type": "int32", "versions": "0+"}
```
abbccdda: Add a comment for the field. Also do you think we need flexible version support here?

abbccdda (commented):
Thanks for the PR, one high level question I have is that whether the new auto generated protocol format is exactly the same as old hard-coded schema format, since this field is used inter-broker. Could you write a test to prove that they could translate between each other?

clients/src/main/resources/common/message/EndTxnMarker.json
dengziming: Firstly I tried to use flexible version here, but it's not the same as the old hard-coded schema format. Could we bump the version and support flexible version from version 1 ?

dengziming:
> Thanks for the PR, one high level question I have is that whether the new auto generated protocol format is exactly the same as old hard-coded schema format, since this field is used inter-broker. Could you write a test to prove that they could translate between each other?

Thank you, I tested that the 2 schemas could translate between each other locally, now I add the code to `EndTransactionMarkerTest`.

dengziming:
ping @abbccdda , Also ping @chia7712 to have a look.

chia7712:
@dengziming could you please fix the conflicts?

dengziming:
Hi, @chia7712 , Thanks for noticing this PR, I have rebased on trunk and the test looks fine.

We have a controversy here that in the past the length of `EndTransactionMarker` is hard-coded to 6, however it is dynamically acquired from it's serialized ByteBuffer, so I stored buffer of a `EndTransactionMarker` when it's created, this will add some state but reduce one calculation.

Another alternative is to change `MemoryRecords.writeEndTransactionalMarker` signature to directly receive (xxx..., ByteBuffer markerBuffer, ControlRecordType markerType) as the below screenshot, but this involves more refactor, WDYT.

image

clients/src/main/java/org/apache/kafka/common/record/EndTransactionMarker.java
```diff
@@ -95,32 +71,28 @@ public int hashCode() {
...
+    // Visible for testing
     static EndTransactionMarker deserializeValue(ControlRecordType type, ByteBuffer value) {
         ensureTransactionMarkerControlType(type);
 
-        if (value.remaining() < CURRENT_END_TXN_MARKER_VALUE_SIZE)
-            throw new InvalidRecordException("Invalid value size found for end transaction marker. Must have " +
-                    "at least " + CURRENT_END_TXN_MARKER_VALUE_SIZE + " bytes, but found only " + value.remaining());
-
-        short version = value.getShort(0);
-        if (version < 0)
```
chia7712, line 114: Should we keep the version check to protect us from reading unknown version in the future?

chia7712 (commented):
@dengziming sorry for delay review :(

one small comment is left. PTAL

clients/src/main/java/org/apache/kafka/common/record/EndTransactionMarker.java
dengziming: Thank you, I thought it was trivial and deleted this, in fact this check makes sense as you suggested, I have changed it back.

clients/src/test/java/org/apache/kafka/common/record/EndTransactionMarkerTest.java
```diff
@@ -17,16 +17,30 @@
...
+import java.util.Arrays;
+import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 public class EndTransactionMarkerTest {
 
+    // Old hard-coded schema, used to validate old hard-coded schema format is exactly the same as new auto generated protocol format
+    private Schema v0Schema = new Schema(
```
chia7712: could you please add `final`?

clients/src/main/java/org/apache/kafka/common/record/EndTransactionMarker.java
```diff
@@ -95,32 +76,35 @@ public int hashCode() {
...
             log.debug("Received end transaction marker value version {}. Parsing as version {}", version,
-                    CURRENT_END_TXN_MARKER_VERSION);
+                    EndTxnMarker.HIGHEST_SUPPORTED_VERSION);
+        EndTxnMarker marker = new EndTxnMarker(new ByteBufferAccessor(value), version);
+        return new EndTransactionMarker(type, marker.coordinatorEpoch());
+    }
 
-        int coordinatorEpoch = value.getInt(2);
-        return new EndTransactionMarker(type, coordinatorEpoch);
+    public int endTxnMarkerValueSize() {
```
chia7712, line 103: Could you please add unit test for this method?

chia7712 (commented):
@dengziming thanks for this patch!

clients/src/main/java/org/apache/kafka/common/record/EndTransactionMarker.java
dengziming: Yes, it's good to test it since it's public.

clients/src/test/java/org/apache/kafka/common/record/EndTransactionMarkerTest.java
dengziming: done.

clients/src/main/java/org/apache/kafka/common/record/EndTransactionMarker.java
```diff
@@ -95,32 +76,35 @@ public int hashCode() {
...
+        if (version < EndTxnMarker.LOWEST_SUPPORTED_VERSION)
             throw new InvalidRecordException("Invalid version found for end transaction marker: " + version +
                     ". May indicate data corruption");
 
-        if (version > CURRENT_END_TXN_MARKER_VERSION)
+        if (version > EndTxnMarker.HIGHEST_SUPPORTED_VERSION)
             log.debug("Received end transaction marker value version {}. Parsing as version {}", version,
-                    CURRENT_END_TXN_MARKER_VERSION);
+                    EndTxnMarker.HIGHEST_SUPPORTED_VERSION);
+        EndTxnMarker marker = new EndTxnMarker(new ByteBufferAccessor(value), version);
```
chia7712: According to the debug message, we should use `EndTxnMarker.HIGHEST_SUPPORTED_VERSION` instead of `version` to parse it as `EndTxnMarker.HIGHEST_SUPPORTED_VERSION`, right?

chia7712 (commented):
@dengziming thanks for this patch, overall LGTM and one last comment is left

clients/src/main/java/org/apache/kafka/common/record/EndTransactionMarker.java
dengziming: Oh, you are so careful about this, though it seems obscure, I agree with you that we should indeed use EndTxnMarker.HIGHEST_SUPPORTED_VERSION.

chia7712 (approved):

PR #9769, opened 2020-12-19
KAFKA-10774; Support Describe topic using topic IDs

dengziming:
1. Add topicNames in `MetadataCache` and alter `KafkaApis.handleTopicMetadataRequest` according to the KIP
2. Add method `describeTopics(TopicCollection)` in KafkaAdminClient, similar to `describeTopics`
3. Change `DescribeTopicsResult` to support TopicId

Test locally, here is some result:
New server + new Client 
```
kafka-topics.sh --describe --zookeeper localhost:2181 --topic-id a-Paxi-LSka9sIVy9UFp_Q
Exception in thread "main" java.lang.IllegalArgumentException: --topic-id can used only with --bootstrap-server

kafka-topics.sh --describe --bootstrap-server localhost:9092 --topic-id a-Paxi-LSka9sIVy9UFp_Q
Topic: old-version-topic	TopicId: a-Paxi-LSka9sIVy9UFp_Q	PartitionCount: 2	ReplicationFactor: 1	Configs: segment.bytes=1073741824
	Topic: old-version-topic	Partition: 0	Leader: 0	Replicas: 0	Isr: 0
	Topic: old-version-topic	Partition: 1	Leader: 0	Replicas: 0	Isr: 0
```

Old Server + new Client
```
kafka-topics.sh --describe --bootstrap-server localhost:9092 --topic-id a-Paxi-LSka9sIVy9UFp_Q
Error while executing topic command : TopicId a-Paxi-LSka9sIVy9UFp_Q not found.
```

New server + old client
```
kafka-topics.sh --describe --bootstrap-server localhost:9092 --topic-id a-Paxi-LSka9sIVy9UFp_Q
Exception in thread "main" joptsimple.UnrecognizedOptionException: topic-id is not a recognized option
```


clients/src/main/java/org/apache/kafka/common/protocol/Errors.java
```diff
@@ -343,7 +344,8 @@
     INVALID_UPDATE_VERSION(95, "The given update version was invalid.", InvalidUpdateVersionException::new),
     FEATURE_UPDATE_FAILED(96, "Unable to update finalized features due to an unexpected server error.", FeatureUpdateFailedException::new),
     PRINCIPAL_DESERIALIZATION_FAILURE(97, "Request principal deserialization failed during forwarding. " +
-         "This indicates an internal error on the broker cluster security setup.", PrincipalDeserializationException::new);
+         "This indicates an internal error on the broker cluster security setup.", PrincipalDeserializationException::new),
+    UNKNOWN_TOPIC_ID(98, "The topic ID dose not exist", UnknownTopicIdException::new);
```
jolshan: nit: typo "dose" -> "does" I also had a slightly different message in my PR that implements this error. Either one is fine. https://github.com/apache/kafka/pull/9684/files

dengziming: Thank you, I think your message may be more appropriate.

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
```diff
@@ -31,29 +31,29 @@
  * The API of this class is evolving, see {@link Admin} for details.
  */
 @InterfaceStability.Evolving
-public class DescribeTopicsResult {
-    private final Map<String, KafkaFuture<TopicDescription>> futures;
+public class DescribeTopicsResult<T> {
```
jolshan: I think this parameterization is a pretty good idea, and I can add it to my delete topics PR. But if we are going to change the public API, we should update the KIP and potentially update the mailing list with the changes.

jolshan: I'm also curious what implications there are for changing the original DescribeTopicsResult when using old clients.

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
```diff
@@ -287,7 +288,23 @@ default DescribeTopicsResult describeTopics(Collection<String> topicNames) {
...
+     * default options. See the overload for more details.
+     *
+     * @param topicIds The topicIds of the topics to describe.
+     * @return The DescribeTopicsResult.
+     */
+    default DescribeTopicsResult<Uuid> describeTopicsWithIds(Collection<Uuid> topicIds) {
+        return describeTopicsWithIds(topicIds, new DescribeTopicsOptions());
+    }
+
+    DescribeTopicsResult<Uuid> describeTopicsWithIds(Collection<Uuid> topicIds, DescribeTopicsOptions options);
```
jolshan: nit: javadoc comment missing

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
dengziming: I first tried a new class `DescribeTopicsResultWithIds` and find it's almost the same as `DescribeTopicsResult`, so I just use parameterization. We only need to change `DescribeTopicsResult` to `DescribeTopicsResult<String>` when using old clients.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1238,8 +1239,21 @@ class KafkaApis(val requestChannel: RequestChannel,
     val metadataRequest = request.body[MetadataRequest]
     val requestVersion = request.header.apiVersion
 
+    val topicIds = metadataRequest.topicIds.asScala.toSet.filterNot(_ == Uuid.ZERO_UUID)
```
jolshan, line 1206: Hey, with the new addition of gating topic IDs behind IBP 2.8, can we do a check of the interbroker protocol version in this method and if it is not at least 2.8, return with UNSUPPORTED_VERSION error. There is a method in the 2.8 PR to easily get this information: `KafkaConfig.usesTopicId`

Here's the other PR for reference: https://github.com/apache/kafka/pull/9814
Here's the KIP with the updated error behavior: https://cwiki.apache.org/confluence/display/KAFKA/KIP-516%3A+Topic+Identifiers#KIP516:TopicIdentifiers-AdminandKafkaAdminClient

I will update the delete topics logic to follow this behavior as well.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1816,7 +1823,103 @@ void handleFailure(Throwable throwable) {
...
+        return new DescribeTopicsResult<>(new HashMap<>(topicFutures));
+    }
+
+    @Override
+    public DescribeTopicsResult<Uuid> describeTopicsWithIds(Collection<Uuid> topicIds, DescribeTopicsOptions options) {
+
+        final Map<Uuid, KafkaFutureImpl<TopicDescription>> topicFutures = new HashMap<>(topicIds.size());
+        final List<Uuid> topicIdsList = new ArrayList<>();
+        for (Uuid topicId : topicIds) {
+            if (topicIdIsUnrepresentable(topicId)) {
```
jolshan, line 1835: We might want to do the IBP check here to and complete exceptionally with UnsupportedVersion error with a message about topic IDs only in IBP 2.8 and above.

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
rajinisivaram: We need to keep the Admin API backwards compatible. An application that was written using the 2.7.0 should not break if it is compiled with a 2.8.0 clients jar. You can always add an internal class with shared code to avoid duplication, but the public API itself needs to remain compatible.

clients/src/main/java/org/apache/kafka/clients/admin/TopicDescription.java
```diff
@@ -62,25 +62,25 @@ public int hashCode() {
...
      *
      * @param name The topic name
+     * @param topicId the topic id
      * @param internal Whether the topic is internal to Kafka
      * @param partitions A list of partitions where the index represents the partition id and the element contains
      *                   leadership and replica information for that partition.
      * @param authorizedOperations authorized operations for this topic, or null if this is not known.
      */
-    public TopicDescription(String name, boolean internal, List<TopicPartitionInfo> partitions,
+    public TopicDescription(String name, Uuid topicId, boolean internal, List<TopicPartitionInfo> partitions,
```
rajinisivaram: We seemed to have removed a constructor?

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1816,7 +1823,103 @@ void handleFailure(Throwable throwable) {
...
+            private Node leader(PartitionInfo partitionInfo) {
+                if (partitionInfo.leader() == null || partitionInfo.leader().id() == Node.noNode().id())
+                    return null;
+                return partitionInfo.leader();
+            }
+
+            @Override
+            boolean handleUnsupportedVersionException(UnsupportedVersionException exception) {
+                if (supportsDisablingTopicCreation) {
+                    supportsDisablingTopicCreation = false;
```
rajinisivaram: Couldn't unsupported version mean that topic ids are not supported (rather than `disablingTopicCreation`)? In any case, you can't create topics with topic ids right?

clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
```diff
@@ -91,7 +91,7 @@ public int throttleTimeMs() {
     public Map<String, Errors> errors() {
         Map<String, Errors> errors = new HashMap<>();
         for (MetadataResponseTopic metadata : data.topics()) {
-            if (metadata.errorCode() != Errors.NONE.code())
+            if (metadata.errorCode() != Errors.NONE.code() && metadata.name() != null)
```
rajinisivaram: What happens if `name` is null - won't we quietly ignore? Would it be better to throw an exception since we don't want anyone using this with topic ids?

clients/src/test/java/org/apache/kafka/clients/admin/MockAdminClient.java
```diff
@@ -206,6 +209,8 @@ synchronized public void addTopic(boolean internal,
                 logDirs.add(brokerLogDirs.get(partition.leader().id()).get(0));
             }
         }
+        Uuid topicId = Uuid.randomUuid();
+        topicIds.put(name, topicId);
```
rajinisivaram: Do we want the option to test with and without topic ids?

core/src/main/scala/kafka/admin/TopicCommand.scala
```diff
@@ -742,10 +760,16 @@ object TopicCommand extends Logging {
...
         throw new IllegalArgumentException("Only one of --bootstrap-server or --zookeeper must be specified")
 
+      if (has(topicIdOpt) && has(zkConnectOpt)) {
+        throw new IllegalArgumentException("--topic-id can used only with --bootstrap-server")
+      }
+
       if (!has(bootstrapServerOpt))
         CommandLineUtils.checkRequiredArgs(parser, options, zkConnectOpt)
-      if(has(describeOpt) && has(ifExistsOpt))
-        CommandLineUtils.checkRequiredArgs(parser, options, topicOpt)
```
rajinisivaram, line 570: We have removed this check?

core/src/main/scala/kafka/admin/TopicCommand.scala
```diff
@@ -313,42 +313,54 @@ object TopicCommand extends Logging {
...
-            val hasNonDefault = config.entries().asScala.exists(!_.isDefault)
-            if (!opts.reportOverriddenConfigs || hasNonDefault) {
-              val numPartitions = td.partitions().size
-              val firstPartition = td.partitions.iterator.next()
-              val reassignment = reassignments.get(new TopicPartition(td.name, firstPartition.partition))
-              val topicDesc = TopicDescription(topicName, topicId, numPartitions, getReplicationFactor(firstPartition, reassignment), config, markedForDeletion = false)
-              topicDesc.printDescription()
-            }
+      if (topicId.isEmpty)
+        ensureTopicExists(topics, opts.topic, !opts.ifExists)
```
rajinisivaram, line 313: `topics` may also be empty?

rajinisivaram (commented):
@dengziming Thanks for the PR, left some comments. The main thing to watch out for is compatibility of the public API.

core/src/main/scala/kafka/server/KafkaApis.scala
dengziming: OK, I will add these check after 9814 merged.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
jolshan: For the second question, yes. Topics can not be created with metadata requests using topic IDs.

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
dengziming: Thank you for your reminder, I have rollbacked all the public API change.

clients/src/main/java/org/apache/kafka/clients/admin/TopicDescription.java
dengziming: This is also my mistake, I tried to add a constructor but remove one, I have rollbacked the removing.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
dengziming: Sorry, just removing those code and using default implement.

clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
dengziming: Firstly I want to leave those check to callers, now I think it's better to throw an exception, So I add another method errorsByTopicId.

core/src/main/scala/kafka/admin/TopicCommand.scala
dengziming: Yes, when we use `--describe`, `--topic` is not mandatory. I add a `if(has(topicOpt) == has(topicIdOpt))` to ensure only one of `--topic` and `--topic-id` is used.

core/src/main/scala/kafka/admin/TopicCommand.scala
dengziming: Here, if topic is provided, we ensure topic exists since the given topic maybe a regular expr. should we also check the topicId exists? then we should add a listTopicId method `listTopicIds` in the AdminClient

clients/src/test/java/org/apache/kafka/clients/admin/MockAdminClient.java
dengziming: Currently, MockAdminClient is mainly used in connect module and stream module, we could add another ticket to do this.

dengziming: Yes, I add a method with a `usesTopicId` parameter.

rajinisivaram:
@dengziming https://github.com/apache/kafka/pull/9814 has been merged, so this needs rebasing and the check for IBP. Thanks.

jolshan:
@dengziming I realized that for this, it really depends on the IBP of the controller. (That is, we need UpdateMetadata to send topic IDs to all the brokers). So maybe instead of checking IBP it would make sense to check if the MetadataCache does not have any topic IDs. What do you think?

dengziming:
> https://github.com/apache/kafka/pull/9769#issuecomment-764014519

@rajinisivaram @jolshan I also realized this, we add topicId in `MetadataCache` from IBP=2.8.1, so instead of using `config.usesTopicId` it would make sense to check `config.interBrokerProtocolVersion >= KAFKA_2_8_IV1`

dengziming:
@rajinisivaram I rebased the code and added a test in `PlaintextAdminIntegrationTest`, PTAL.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1822,6 +1829,88 @@ void handleFailure(Throwable throwable) {
...
+                    for (PartitionInfo partitionInfo : partitionInfos) {
+                        TopicPartitionInfo topicPartitionInfo = new TopicPartitionInfo(
+                                partitionInfo.partition(), leader(partitionInfo), Arrays.asList(partitionInfo.replicas()),
+                                Arrays.asList(partitionInfo.inSyncReplicas()));
+                        partitions.add(topicPartitionInfo);
+                    }
+                    partitions.sort(Comparator.comparingInt(TopicPartitionInfo::partition));
+                    TopicDescription topicDescription = new TopicDescription(topicName, isInternal, partitions,
+                            validAclOperations(response.topicAuthorizedOperations(topicName).get()), topicId);
+                    future.complete(topicDescription);
```
rajinisivaram, line 2041: Could we move the code shared between `describeTopics` and `describeTopicsWithIds` to a common private method?

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -1822,6 +1829,88 @@ void handleFailure(Throwable throwable) {
...
+                }
+            }
+
+            private Node leader(PartitionInfo partitionInfo) {
```
rajinisivaram: This could be shared with `describeTopics` as well?

clients/src/main/java/org/apache/kafka/clients/admin/TopicDescription.java
```diff
@@ -65,6 +65,19 @@ public TopicDescription(String name, boolean internal, List<TopicPartitionInfo>
...
+    /**
+     * Create an instance with the specified parameters.
+     *
+     * @param name The topic name
+     * @param internal Whether the topic is internal to Kafka
+     * @param partitions A list of partitions where the index represents the partition id and the element contains
+     *                   leadership and replica information for that partition.
+     * @param topicId the topic id
+     */
+    public TopicDescription(String name, boolean internal, List<TopicPartitionInfo> partitions, Uuid topicId) {
```
rajinisivaram: Do we really need this constructor in the public class? We could just use the one below?

clients/src/main/java/org/apache/kafka/clients/admin/TopicDescription.java
```diff
@@ -79,6 +92,16 @@ public TopicDescription(String name, boolean internal, List<TopicPartitionInfo>
...
     }
 
+    /**
+     * Create an instance with the specified parameters.
+     *
+     * @param name The topic name
+     * @param internal Whether the topic is internal to Kafka
+     * @param partitions A list of partitions where the index represents the partition id and the element contains
+     *                   leadership and replica information for that partition.
+     * @param authorizedOperations authorized operations for this topic, or null if this is not known.
```
rajinisivaram: Looks like default is empty set rather than null?

clients/src/main/java/org/apache/kafka/common/errors/UnknownTopicIdException.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.common.errors;
+
+public class UnknownTopicIdException extends ApiException {
+    private static final long serialVersionUID = 1L;
+    public UnknownTopicIdException() {
+        super();
```
rajinisivaram: `super()` is implied, so we can just leave the method empty?

clients/src/main/java/org/apache/kafka/common/protocol/Errors.java
```diff
@@ -350,7 +351,8 @@
     POSITION_OUT_OF_RANGE(
         99,
         "Requested position is not greater than or equal to zero, and less than the size of the snapshot.",
-        PositionOutOfRangeException::new);
+        PositionOutOfRangeException::new),
+    UNKNOWN_TOPIC_ID(100, "The topic ID dose not exist", UnknownTopicIdException::new);
```
rajinisivaram: typo: `does`

clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
```diff
@@ -91,12 +91,31 @@ public int throttleTimeMs() {
     public Map<String, Errors> errors() {
         Map<String, Errors> errors = new HashMap<>();
         for (MetadataResponseTopic metadata : data.topics()) {
+            if (metadata.name() == null) {
+                throw new NullPointerException("Use errorsByTopicId() when manage topic using topic id");
```
rajinisivaram: We should never get here through the public APIs right? We could use IllegalStateException.

clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
```diff
@@ -91,12 +91,31 @@ public int throttleTimeMs() {
...
 
+    /**
+     * Get a map of the topicIds which had metadata errors
+     * @return the map
+     */
+    public Map<Uuid, Errors> errorsByTopicId() {
+        Map<Uuid, Errors> errors = new HashMap<>();
+        for (MetadataResponseTopic metadata : data.topics()) {
+            if (metadata.topicId() == Uuid.ZERO_UUID) {
+                throw new NullPointerException("Use errors() when manage topic using topic name");
```
rajinisivaram: As before, IllegalStateException?

core/src/main/scala/kafka/admin/TopicCommand.scala
```diff
@@ -313,42 +313,57 @@ object TopicCommand extends Logging {
     }
 
     override def describeTopic(opts: TopicCommandOptions): Unit = {
-      val topics = getTopics(opts.topic, opts.excludeInternalTopics)
-      ensureTopicExists(topics, opts.topic, !opts.ifExists)
+      val topicId = opts.topicId.map(Uuid.fromString).filter(_ != Uuid.ZERO_UUID)
+      // if topicId is provided and not zero, will use topicId regardless of topic name
```
rajinisivaram: we would fail if both are provided?

core/src/main/scala/kafka/admin/TopicCommand.scala
rajinisivaram: Since listTopics uses MetadataRequest, we just need to add topic ids from the MetadataResponse to TopicListing?

core/src/main/scala/kafka/admin/TopicCommand.scala
```diff
@@ -635,6 +650,11 @@ object TopicCommand extends Logging {
                          .withRequiredArg
                          .describedAs("topic")
                          .ofType(classOf[String])
+    private val topicIdOpt = parser.accepts("topic-id", "The topic-id to describe." +
+      "This is used only with --bootstrap-server option for manage topic.")
```
rajinisivaram: for manage topic => `for describing topics`?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1190,8 +1192,31 @@ class KafkaApis(val requestChannel: RequestChannel,
     val metadataRequest = request.body[MetadataRequest]
     val requestVersion = request.header.apiVersion
 
+    // Check if topicId is presented firstly.
+    val topicIds = metadataRequest.topicIds.asScala.toSet.filterNot(_ == Uuid.ZERO_UUID)
+    val supportedVersionTopicIds = if (config.interBrokerProtocolVersion >= KAFKA_2_8_IV1) topicIds else Set.empty[Uuid]
+
+    val unSupportedVersionTopicIds = topicIds.diff(supportedVersionTopicIds)
```
rajinisivaram: `unSupported` => `unsupported` since it is one word (below as well)

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1190,8 +1192,31 @@ class KafkaApis(val requestChannel: RequestChannel,
     val metadataRequest = request.body[MetadataRequest]
     val requestVersion = request.header.apiVersion
 
+    // Check if topicId is presented firstly.
+    val topicIds = metadataRequest.topicIds.asScala.toSet.filterNot(_ == Uuid.ZERO_UUID)
+    val supportedVersionTopicIds = if (config.interBrokerProtocolVersion >= KAFKA_2_8_IV1) topicIds else Set.empty[Uuid]
```
rajinisivaram: @jolshan Shouldn't we use `KAFKA_2_8_IV1` for `KafkaConfig.usesTopicId` and use that here?

rajinisivaram (commented):
@dengziming Thanks for the updates. Left some comments/questions. I ran one of the tests and there were NullPointerExceptions, so those needs fixing as well.

core/src/main/scala/kafka/admin/TopicCommand.scala
dengziming: that's what I thought at the beginning, now I think it's better to use topicId and print a warning message if both are provided.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
dengziming: Thank you, I added a getTopicDescriptionFromCluster to do this.

dengziming:
Hello, @rajinisivaram , I resolved the comments/questions and fixed the NullPointerExceptions, call for a third review. 
1. Add topicId to`TopicListing ` and add `TopicCommand.getTopicIds`
2. move the code shared between describeTopics and describeTopicsWithIds to a common private method `getTopicDescriptionFromCluster`.

core/src/main/scala/kafka/server/KafkaApis.scala
rajinisivaram: Had a chat offline with Justine. Since `KafkaConfig.usesTopicId` is used only to decide whether to generate topic ids, 2_8_IV0 makes sense for that, and the check for KAFKA_2_8_IV1 here is the right one. Justine also mentioned:

> if we are in the process of bumping to 2_8_IV1, we could have a scenario where the controller is not yet bumped (and not sending topic IDs) but the broker serving the metadata is. In that case I think we would just get unknown topic ID errors and maybe that is ok. If the reverse is occurring, we could also have unsupported version exception when there are actually IDs.


Can we add tests for these two cases?

dengziming: Hello, I tried to add 2 test in `MetadataRequestBetweenDifferentIbpTest` to verify these 2 cases, this test will take probably 10 seconds since we need to constantly restart brokers to change the controller.

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
```diff
@@ -0,0 +1,117 @@
...
+
+import scala.collection.{Map, Seq}
+
+class MetadataRequestBetweenDifferentIbpTest extends BaseRequestTest {
+
+  override def brokerCount: Int = 3
+  override def generateConfigs: Seq[KafkaConfig] = {
+    Seq(
+      createConfig(0, KAFKA_2_8_IV0),
+      createConfig(1, KAFKA_2_8_IV1)
```
rajinisivaram: Don't we need the config for the third broker?

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
```diff
@@ -0,0 +1,117 @@
...
+    restartDeadBrokers()
+
+    // Send request to a broker whose version=KAFKA_2_8_IV1
+    val resp2 = sendMetadataRequest(new MetadataRequest(requestData(topic, topicId), 10.toShort), brokerSocketServer(1))
+    assertEquals(Errors.UNKNOWN_TOPIC_ID, resp2.topicMetadata.iterator().next().error())
+  }
+
+  private def ensureControllerIn(brokerIds: Seq[Int]): Unit = {
+    while (!brokerIds.contains(controllerSocketServer.config.brokerId)) {
+      killBroker(controllerSocketServer.config.brokerId)
```
rajinisivaram: Instead of restarting brokers, we could delete the controller node in ZK to force controller election. It doesn't guarantee that the expected broker id becomes the next controller, but it may still get there faster than broker restarts

rajinisivaram (commented):
@dengziming Thanks for the updates and tests, looks good. Left just a couple of minor comments. @jolshan Can you also review to make sure I haven't missed anything? Thanks.

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
dengziming: Firstly I add 2 brokers to ensure faster election after killing the broker, now I added a broker since we need not kill broker to trigger election.

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
dengziming: Good catch, updated.

rajinisivaram (approved):
@dengziming Thanks for the updates, LGTM. Will merge after @jolshan has a chance to review.

core/src/main/scala/kafka/admin/TopicCommand.scala
jolshan: I don't have an issue with this, but just curious why this was chosen over failing.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1190,8 +1192,31 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+
+    val unsupportedVersionTopicIds = topicIds.diff(supportedVersionTopicIds)
```
jolshan: I'm curious if this diff is needed. It seems that the above check will result in all topicIds being left out or all of them being included.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1213,8 +1238,7 @@ class KafkaApis(val requestChannel: RequestChannel,
     }
 
     val unauthorizedForCreateTopicMetadata = unauthorizedForCreateTopics.map(topic =>
-      metadataResponseTopic(Errors.TOPIC_AUTHORIZATION_FAILED, topic, isInternal(topic), util.Collections.emptyList()))
-
+      metadataResponseTopic(Errors.TOPIC_AUTHORIZATION_FAILED, topic, metadataCache.getTopicId(topic), isInternal(topic), util.Collections.emptyList()))
```
jolshan, line 1241: Does this potentially leak the topic name when we are using topic IDs?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1223,7 +1247,7 @@ class KafkaApis(val requestChannel: RequestChannel,
         Set.empty[MetadataResponseTopic]
       else
         unauthorizedForDescribeTopics.map(topic =>
-          metadataResponseTopic(Errors.TOPIC_AUTHORIZATION_FAILED, topic, false, util.Collections.emptyList()))
+          metadataResponseTopic(Errors.TOPIC_AUTHORIZATION_FAILED, topic, metadataCache.getTopicId(topic), false, util.Collections.emptyList()))
```
jolshan: Does this potentially acknowledge a topic exists if we give a valid id?

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.kafka.common.Uuid
+import org.apache.kafka.common.message.MetadataRequestData
+import org.apache.kafka.common.protocol.Errors
+import org.apache.kafka.common.requests.{MetadataRequest, MetadataResponse}
+import org.junit.jupiter.api.Assertions._
+import org.junit.jupiter.api.Test
+
+import scala.collection.{Map, Seq}
+
+class MetadataRequestBetweenDifferentIbpTest extends BaseRequestTest {
```
jolshan, line 33: This test is super cool and will be very useful when testing other request types like fetch.

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
```diff
@@ -0,0 +1,125 @@
...
+    // Make the broker whose version=KAFKA_2_8_IV0 controller
+    ensureControllerIn(Seq(0))
+
+    // Restart the broker whose version=KAFKA_2_8_IV1, and the controller will send metadata request to it
+    killBroker(1)
+    restartDeadBrokers()
+
+    // Send request to a broker whose version=KAFKA_2_8_IV1 and restarted just now
+    val resp2 = sendMetadataRequest(new MetadataRequest(requestData(topic, topicId), 10.toShort), brokerSocketServer(1))
+    assertEquals(Errors.UNKNOWN_TOPIC_ID, resp2.topicMetadata.iterator().next().error())
```
jolshan, line 70: This scenario is a bit of a corner case and maybe not a huge deal either way, but does it make more sense to respond with UNKNOWN_TOPIC_ID or UNSUPPORTED_VERSION here? Maybe either is fine since we can expect UNKNOWN_TOPIC_ID or UNSUPPORTED_VERSION during an upgrade like this anyway.

jolshan:
Thanks for the updates and all the hard work! I added a few comments. Mostly just questions for clarification.

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: (and if we give ID when using name, we acknowledge it exists)

jolshan: I think this usage is fine since we will only get here if use the topic name version of the request and the topic does not exist.

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: We should check some of the other error responses to make sure we use name/id appropriately

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
dengziming: Yes, both ERROR is ok here.

core/src/main/scala/kafka/server/KafkaApis.scala
dengziming: I checked the error responses and found that we should hide TopicId when error=`TOPIC_AUTHORIZATION_FAILED`, and we can respond topicId on other errors.

clients/src/main/java/org/apache/kafka/clients/admin/TopicListing.java
```diff
@@ -17,24 +17,36 @@
...
      */
-    public TopicListing(String name, boolean internal) {
+    public TopicListing(String name, Uuid topicId, boolean internal) {
+        this.topicId = topicId;
         this.name = name;
         this.internal = internal;
     }
 
+    /**
+     * The name of the topic.
```
satishd: typo: should have been "The id ~~name~~ of the topic"

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: Thanks for fixing this!

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1194,9 +1194,13 @@ class KafkaApis(val requestChannel: RequestChannel,
...
     val topicIds = metadataRequest.topicIds.asScala.toSet.filterNot(_ == Uuid.ZERO_UUID)
-    val supportedVersionTopicIds = if (config.interBrokerProtocolVersion >= KAFKA_2_8_IV1) topicIds else Set.empty[Uuid]
 
-    val unsupportedVersionTopicIds = topicIds.diff(supportedVersionTopicIds)
+    val (supportedVersionTopicIds, unsupportedVersionTopicIds) = if (config.interBrokerProtocolVersion >= KAFKA_2_8_IV1)
+      (topicIds, Set.empty[Uuid])
+    else
+      (Set.empty[Uuid], topicIds)
+
+    // val unsupportedVersionTopicIds = topicIds.diff(supportedVersionTopicIds)
```
jolshan: nit: remove commented out line.

core/src/main/scala/kafka/admin/TopicCommand.scala
```diff
@@ -313,42 +314,59 @@ object TopicCommand extends Logging {
     }
 
     override def describeTopic(opts: TopicCommandOptions): Unit = {
-      val topics = getTopics(opts.topic, opts.excludeInternalTopics)
-      ensureTopicExists(topics, opts.topic, !opts.ifExists)
+      // if topicId is provided and not zero, will use topicId regardless of topic name
+      val useTopicId = opts.topicId.map(Uuid.fromString).nonEmpty
```
satishd: Based on the above comment, does not it need to handle `UUID.ZERO_UUID` too?

satishd (commented):
thanks @dengziming for the PR. Oveall LGTM,  left minor comments.

core/src/main/scala/kafka/admin/TopicCommand.scala
jolshan: good point!

dengziming: Thank you, I was careless about this case, I add the following fix:
1. When --topic and --topic-id are both specified, I will print a warning message that saying that "topicId will be used if both are specified and topicId is not zero"
2. Filter ZERO UUID before handling in TopicCommand.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1223,7 +1251,7 @@ class KafkaApis(val requestChannel: RequestChannel,
         Set.empty[MetadataResponseTopic]
       else
         unauthorizedForDescribeTopics.map(topic =>
-          metadataResponseTopic(Errors.TOPIC_AUTHORIZATION_FAILED, topic, false, util.Collections.emptyList()))
+          metadataResponseTopic(Errors.TOPIC_AUTHORIZATION_FAILED, topic, Uuid.ZERO_UUID, false, util.Collections.emptyList()))
```
jolshan, line 1256: So will we never reach this code path when using topic IDs? I think we are using topics to decide authorization. So in the case where we use ids and the name exists, then we will expose the name and return a zero ID? Might be useful to create an authorizer integration test with topic IDs to ensure correctness.

rajinisivaram: Good point. If not authorized for describe when using topic ids, we need to make sure we don't return the topic or information about existence of a topic - i.e we can't return TOPIC_AUTHORIZATION_FAILED. Perhaps UNKNOWN_TOPIC_ID would be more suitable.

satishd (approved):
thanks @dengziming for addressing the review comments, LGTM.

core/src/main/scala/kafka/server/KafkaApis.scala
dengziming: @jolshan , Thank you, I misunderstood your comments before, I added the following changes:
1. If we are using topicId in metadataReq, we set topicName=null and error=UNKNOWN_TOPIC_ID on unauthorized error
2. If we are using topicName in metadataReq, we set topicId=ZERO and error=TOPIC_AUTHORIZATION_FAILED on unauthorized error.
3. Add a unit test for the above 2 cases.
@rajinisivaram @jolshan , pleas take a look.

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: Looks good now. We will only get to this point with knownTopicNames, so the metadataCache will also have the topic ID. 👍

jolshan:
Thanks for the update. The test case is good. In general, LGTM

core/src/main/scala/kafka/server/KafkaApis.scala
jolshan: LGTM

jolshan (approved):

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -1879,6 +1879,109 @@ class KafkaApisTest {
...
+
+    // 4. Send TopicMetadataReq using topicId
+    val capturedMetadataByTopicIdResp = expectNoThrottling()
+    EasyMock.replay(clientRequestQuotaManager, requestChannel, authorizer)
+
+    val metadataReqByTopicId = new MetadataRequest.Builder(util.Arrays.asList(authorizedTopicId, unauthorizedTopicId)).build()
+    createKafkaApis(authorizer = Some(authorizer)).handleTopicMetadataRequest(buildRequest(metadataReqByTopicId, plaintextListener))
+    val metadataByTopicIdResp = readResponse(metadataReqByTopicId, capturedMetadataByTopicIdResp).asInstanceOf[MetadataResponse]
+
+    val metadataByTopicId = metadataByTopicIdResp.data().topics().asScala.groupBy(_.topicId()).view.mapValues(_.head).toMap
```
rajinisivaram: @dengziming Thanks for the test! This line and the line declaring `metadataByTopicName` below are failing the scala 2.12 build, can we rewrite without `mapValues` so that the Java8/Scala2.12 build passes? Thanks.

dengziming: Thank you, Done.

rajinisivaram:
@dengziming Thanks for the update. Looks like there is an issue with the tests in the last commit (see test failures in the PR build).

dengziming:
@rajinisivaram I triggered the QA again but still has some failures, it seems the failure is unrelated to this pr since other PRs also have these failures.

rajinisivaram:
There are quite a few failures in the last PR build. Failures don't look related to this PR, rerunning the build.

rajinisivaram:
@dengziming Sorry missed your note. Will wait for the current build to complete in case it was a timing issue due to slow machines. Agree that the issues don't look related to this PR, if the tests still fail, I will verify locally and merge. Thanks.

rajinisivaram:
@dengziming The new PR build completed and we are still seeing NullPointerException in the JDK 8 build for the new test. The results are here: https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-9769/16/#showFailuresLink

dengziming:
@rajinisivaram Thank you, I think the problem comes from `EasyMock.expect(authorizer.authorize(anyObject[RequestContext], EasyMock.eq(expectedActions.asJava)))`.
below is what I expecet:
```
 val expectedActions = Seq(
      new Action(AclOperation.DESCRIBE, new ResourcePattern(ResourceType.TOPIC, unauthorizedTopic, PatternType.LITERAL), 1, true, true),
      new Action(AclOperation.DESCRIBE, new ResourcePattern(ResourceType.TOPIC, authorizedTopic, PatternType.LITERAL), 1, true, true)
    )
```
but maybe the order of the real params is reversed, so make some change.

rajinisivaram:
@dengziming Even the last PR build for this one seemed to have issues. I merged https://github.com/apache/kafka/pull/9684 first since that build looked good. But now this one needs a rebase. I can do the rebase if it is already too late for you.

rajinisivaram:
@dengziming I am so sorry, I didn't notice that you had updated the branch a few minutes ago, so force-pushed my version. Do you mind force-pushing your branch once more please?

dengziming:
The 2 failures in https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-9769/22/#showFailuresLink seem to be related to Mockito, maybe #10006 will fix it. @rajinisivaram PTAL.

clients/src/main/java/org/apache/kafka/clients/admin/TopicListing.java
```diff
@@ -17,24 +17,36 @@
...
+    public TopicListing(String name, Uuid topicId, boolean internal) {
+        this.topicId = topicId;
         this.name = name;
         this.internal = internal;
     }
 
+    /**
+     * The id of the topic.
+     */
+    public Uuid topicId() {
```
chia7712, line 60: Could you add topic id to ```toString```?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1191,8 +1192,35 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+    // Check if topicId is presented firstly.
+    val topicIds = metadataRequest.topicIds.asScala.toSet.filterNot(_ == Uuid.ZERO_UUID)
+    val useTopicId = topicIds.nonEmpty
+
+    val (supportedVersionTopicIds, unsupportedVersionTopicIds) = if (config.interBrokerProtocolVersion >= KAFKA_2_8_IV1)
+      (topicIds, Set.empty[Uuid])
+    else
+      (Set.empty[Uuid], topicIds)
+
+    val unsupportedVersionTopicMetadata = if (unsupportedVersionTopicIds.isEmpty)
```
chia7712:
```scala
    val unsupportedVersionTopicMetadata = unsupportedVersionTopicIds.map(topicId =>
        metadataResponseTopic(Errors.UNSUPPORTED_VERSION, null, topicId, false, util.Collections.emptyList())).toSeq
```

Is above code more simple?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1191,8 +1192,35 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+      Seq.empty[MetadataResponseTopic]
+    else
+      unsupportedVersionTopicIds.map(topicId =>
+        metadataResponseTopic(Errors.UNSUPPORTED_VERSION, null, topicId, false, util.Collections.emptyList())).toSeq
+
+    // Only get topicIds and topicNames when supporting topicId
+    val unknownTopicIds = supportedVersionTopicIds.filter(metadataCache.getTopicName(_).isEmpty)
+    val knownTopicNames = supportedVersionTopicIds.flatMap(metadataCache.getTopicName)
+
+    val unknownTopicIdsTopicMetadata = if (unknownTopicIds.isEmpty)
```
chia7712:
```scala
    val unknownTopicIdsTopicMetadata = unknownTopicIds.map(topicId =>
        metadataResponseTopic(Errors.UNKNOWN_TOPIC_ID, null, topicId, false, util.Collections.emptyList())).toSeq
```

Is above code more simple?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1191,8 +1192,35 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+      Seq.empty[MetadataResponseTopic]
+    else
+      unknownTopicIds.map(topicId =>
+        metadataResponseTopic(Errors.UNKNOWN_TOPIC_ID, null, topicId, false, util.Collections.emptyList())).toSeq
+
     val topics = if (metadataRequest.isAllTopics)
       metadataCache.getAllTopics()
+    else if (useTopicId)
```
chia7712, line 1218: Should we reject the requests carrying both topic id and topic name?

chia7712 (commented):
@dengziming Thanks for this nice patch. Please take a look at following minor questions :)

Does ```MetadataRequest#getErrorResponse``` need to fill topic id?
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/MetadataRequest.java#L122

rajinisivaram:
@dengziming Have you tried any system tests with this PR? @jolshan ran a system tests build on this branch twice and found quite a few tests had failed. It is not clear whether the failures are due to the PR, but we haven't been seeing the same failures on other builds. If you have time to investigate, you can try running the DEV_BRANCH version of `tests/kafkatest/sanity_checks/test_performance_services.py` or the various compatibility tests in `tests/kafkatest/tests/client/client_compatibility_produce_consume_test.py`.

core/src/main/scala/kafka/server/KafkaApis.scala
dengziming: According to KIP-516, we just ignore topic name if a none zero topic id is provided.

core/src/main/scala/kafka/server/KafkaApis.scala
dengziming: Thanks, Done!

dengziming:
@rajinisivaram I haven't run any system tests  and I will try to run some system tests later.

dengziming:
@rajinisivaram I haven't run any system tests and I will try to run some system tests later. Should I run the 2 python script locally and analyze the logs?

rajinisivaram:
@dengziming Yes, if you are able to run those two tests locally, that will be good. The failures looked similar with consumer failing to consume, so I guess may be only one issue to figure out (it is not 100% clear if the issue is with this PR).

dengziming:
@rajinisivaram  I run `TC_PATHS="tests/kafkatest/tests/client/client_compatibility_produce_consume_test.py" bash tests/docker/run_tests.sh` locally on the `trunk branch` version, and I also got a `ducktape.errors.TimeoutError: Consumer failed to consume messages for 60s` error. did that mean the problem is introduced by other PRs? This is weird, I havn't found out the reason for the error.

rajinisivaram:
@dengziming Thanks for looking into this. Looks like a timing issue. @jolshan had run two system test runs on this PR last week and both had these consumer issues. At the time trunk looked good. But last night's trunk run also had very similar set of failures. So there is clearly something going on that is not related to this PR. Will take a look today.

rajinisivaram:
Looks like a bug was introduced by a different PR, opened https://issues.apache.org/jira/browse/KAFKA-12268 to fix this.

jolshan:
@dengziming I don't know if you saw the mailing list discussion on this KIP, but the summary is that I talked to Ismael offline and we thought it would be good to hold off on the Admin Client APIs until after the 2.8 release is cut. I mentioned in that thread about splitting this PR into two (one PR with Admin client changes, one with everything else), but on further thought, I think it makes sense to wait until after 2.8 is cut.

dengziming:
@rajinisivaram @jolshan , Thank you for your effort, I pay attention to the 2.8 release and update this pr.

dengziming:
@rajinisivaram @jolshan , Currently we have startd merge pr for 3.0, so it's time to take a look at this again, also #9684 since part of it has been reverted, thank you.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1165,16 +1189,23 @@ class KafkaApis(val requestChannel: RequestChannel,
...
     // do not disclose the existence of topics unauthorized for Describe, so we've not even checked if they exist or not
     val unauthorizedForDescribeTopicMetadata =
       // In case of all topics, don't include topics unauthorized for Describe
       if ((requestVersion == 0 && (metadataRequest.topics == null || metadataRequest.topics.isEmpty)) || metadataRequest.isAllTopics)
         Set.empty[MetadataResponseTopic]
-      else
+      else if (useTopicId) {
+        // We should not return information about existence of a topic on unauthorized error, so we return an UNKNOWN_TOPIC_ID
+        unauthorizedForDescribeTopics.map(topic =>
+          metadataResponseTopic(Errors.UNKNOWN_TOPIC_ID, null, metadataCache.getTopicId(topic), false, util.Collections.emptyList()))
```
jolshan: In the time since we last looked at this, we decided that it is ok to acknowledge existence of the topic ID and return TOPIC_AUTHORIZATION_FAILED. This is so the client knows it is not a retriable error. Please see https://issues.apache.org/jira/browse/KAFKA-12394 for more details.

dengziming:
Check KAFKA-12701 and bump the version of the MetadataRequest/MetadataResponse, @rajinisivaram @jolshan PTAL.

jolshan:
Hi @dengziming thanks for the update. There are a few more minor changes to the KIP.
One is that after some discussion with people more familiar with clients, it makes sense to remove the UNSUPPORTED_VERSION error and only return UNKNOWN_TOPIC_ID when the topic ID is not found on the broker. (See: https://issues.apache.org/jira/browse/KAFKA-12976)

jolshan:
Hi again @dengziming. There was one more change to the API (should be the last one this time). It includes adding a new class that I implement in this PR: https://github.com/apache/kafka/pull/10892
Please check out the KIP for more details on the changes to DescribeTopicsResult and the new apis for describing with topic IDs. Thanks for your diligence and patience with this one.

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
```diff
@@ -32,28 +34,87 @@
...
-        return futures;
+        return nameFutures;
     }
 
     /**
      * Return a future which succeeds only if all the topic descriptions succeed.
      */
     public KafkaFuture<Map<String, TopicDescription>> all() {
-        return KafkaFuture.allOf(futures.values().toArray(new KafkaFuture[0])).
+        return all(nameFutures);
```
jolshan, line 104: We should deprecate this one too I believe.

clients/src/main/java/org/apache/kafka/common/requests/MetadataRequest.java
```diff
@@ -65,6 +65,20 @@ public Builder(List<String> topics, boolean allowAutoTopicCreation) {
             this(topics, allowAutoTopicCreation, ApiKeys.METADATA.oldestVersion(),  ApiKeys.METADATA.latestVersion());
         }
 
+        public Builder(List<Uuid> topicIds) {
+            super(ApiKeys.METADATA, ApiKeys.METADATA.oldestVersion(), ApiKeys.METADATA.latestVersion());
+            MetadataRequestData data = new MetadataRequestData();
+            if (topicIds == null)
+                data.setTopics(null);
+            else {
+                topicIds.forEach(topicId -> data.topics().add(new MetadataRequestTopic().setTopicId(topicId).setName("")));
```
jolshan: Is there a reason this is setting the name to empty and not null? (or no name set at all?)

clients/src/main/resources/common/message/MetadataResponse.json
```diff
@@ -65,7 +66,7 @@
       "about": "Each topic in the response.", "fields": [
       { "name": "ErrorCode", "type": "int16", "versions": "0+",
         "about": "The topic error, or 0 if there was no error." },
-      { "name": "Name", "type": "string", "versions": "0+", "mapKey": true, "entityType": "topicName",
+      { "name": "Name", "type": "string", "versions": "0+", "mapKey": true, "entityType": "topicName", "nullableVersions": "10+",
```
jolshan: can we set a previous version to be nullable? Or does this need to be 12+

core/src/main/scala/kafka/api/ApiVersion.scala
```diff
@@ -116,7 +116,9 @@ object ApiVersion {
     // Introduce AllocateProducerIds (KIP-730)
     KAFKA_3_0_IV0,
     // Introduce ListOffsets V7 which supports listing offsets by max timestamp (KIP-734)
-    KAFKA_3_0_IV1
+    KAFKA_3_0_IV1,
+    // Introduced topic IDs to MetadataRequest
```
jolshan: Why are we bumping IBP? Metadata is not an inter-broker protocol, so I think bumping the metadata protocol suffices.

dengziming: You are right, we no longer need to check UNSUPPORTED_VERSION_ERROR, so it's unnecessary to bump IBP.

clients/src/main/resources/common/message/MetadataResponse.json
dengziming: Yes, this should be 12+

clients/src/main/java/org/apache/kafka/common/requests/MetadataRequest.java
dengziming: This is old code since we do not support null names firstly, now we can remove this useless code.

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
dengziming: Done, and also change all usages to use `allTopicNames()`.

jolshan:
Looks like some deprecated methods should be replaced.
```
 /home/jenkins/jenkins-agent/workspace/Kafka_kafka-pr_PR-9769/connect/mirror/src/main/java/org/apache/kafka/connect/mirror/MirrorSourceConnector.java:402: warning: 

[deprecation] all() in org.apache.kafka.clients.admin.DescribeTopicsResult has been deprecated
 ```

**Comment by dengziming** [comment]
Replace deprecated methods and rebase on trunk.

File: clients/src/main/java/org/apache/kafka/clients/admin/Admin.java [hunk 45]
```diff
@@ -303,7 +303,33 @@ default DescribeTopicsResult describeTopics(Collection<String> topicNames) {
...
+     * @return The DescribeTopicsResult.
+     */
+    default DescribeTopicsResult describeTopics(TopicCollection topics) {
+        return describeTopics(topics, new DescribeTopicsOptions());
+    }
+
+    /**
+     * describe a batch of topics.
+     *
+     * When using topic IDs, this operation is supported by brokers with version 3.0.0 or higher.
```
jolshan: Since we are past the feature freeze date, I think this will need to be 3.1 now. Same above.

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
```diff
@@ -0,0 +1,103 @@
...
+    val resp1 = sendMetadataRequest(new MetadataRequest(requestData(topic, Uuid.ZERO_UUID), 12.toShort), controllerSocketServer)
+    val topicId = resp1.topicMetadata.iterator().next().topicId()
+
+    // We could still get topic metadata by topicId
+   val topicMetadata = sendMetadataRequest(new MetadataRequest(requestData(null, topicId), 12.toShort), controllerSocketServer)
+      .topicMetadata.iterator().next()
+    assertEquals(topicId, topicMetadata.topicId())
+    assertEquals(topic, topicMetadata.topic())
+
+    // Make the broker whose version=KAFKA_3_0_IV0 controller
```
jolshan: I'm not sure this comment is correct since we don't use this version. 
Just to confirm my understanding, what is this test showing? First we can get description when controller is high enough IBP (2_8_IV1+) and then we can't when controller is lower than that?

dengziming: This class is created to test the 2 cases here: https://github.com/apache/kafka/pull/9769#discussion_r562511974 , and now we only need to test the `UNKNOWN_TOPIC_ID` case since we do not need `UNSUPPORTED_VERSION` now.

And this comment is wrong, I will fix it.

dengziming:
@rajinisivaram Are you interested in taking this across finish line? I think this has something that is very close to a first version, we can try to land that and then improvise since this has been worked for a long time, thank you!

jolshan:
@dengziming -- @rajinisivaram can help in about 2 weeks. I will try to help review so it will be ready when she gets to it.

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
```diff
@@ -42,6 +34,14 @@
 import org.apache.kafka.common.quota.ClientQuotaFilter;
 import org.apache.kafka.common.requests.LeaveGroupResponse;
 
+import java.time.Duration;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
```
jolshan, line 40: nit: is there a reason we moved these imports?

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
```diff
@@ -303,7 +303,33 @@ default DescribeTopicsResult describeTopics(Collection<String> topicNames) {
...
+     *
+     * @param topics The topics to describe.
+     * @return The DescribeTopicsResult.
+     */
+    default DescribeTopicsResult describeTopics(TopicCollection topics) {
+        return describeTopics(topics, new DescribeTopicsOptions());
+    }
+
+    /**
+     * describe a batch of topics.
```
jolshan: Can we adjust this javadoc to be `Describe some topics in the cluster.` like the previous API?

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
```diff
@@ -32,28 +34,87 @@
...
+     */
+    public KafkaFuture<Map<Uuid, TopicDescription>> allTopicIds() {
+        return all(topicIdFutures);
+    }
+
+    /**
+     * Return a future which succeeds only if all the topic descriptions succeed.
+     */
+    private static <T> KafkaFuture<Map<T, TopicDescription>> all(Map<T, KafkaFuture<TopicDescription>> futures) {
+
```
jolshan: nit: extra newline here

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
```diff
@@ -0,0 +1,102 @@
...
+
+import scala.collection.{Map, Seq}
+
+class MetadataRequestBetweenDifferentIbpTest extends BaseRequestTest {
+
+  override def brokerCount: Int = 3
+  override def generateConfigs: Seq[KafkaConfig] = {
+    Seq(
+      createConfig(0, KAFKA_2_8_IV0),
+      createConfig(1, KAFKA_3_0_IV1),
```
jolshan: I think this was discussed briefly before, but is there a reason KAFKA_3_0_IV1 was chosen? Should we just use the most recent IBP? (meaning, not even specify in the properties -- just pick up the default?)

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
dengziming: This is invoked by the IDE spotless plugin, but I think this is reasonable since we are going to add import order checkstyle here:https://github.com/apache/kafka/pull/10428/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R615, and java.* will be put last.

core/src/test/scala/integration/kafka/server/MetadataRequestBetweenDifferentIbpTest.scala
dengziming: In fact, when I first change this to KAFKA_3_1_IV0, this test became a flaky test, and now I fixed it.

jolshan: Ah ok. There were some issues with KAFKA_3_1_IV0. I fixed one of the issues that led to flaky tests. Hopefully that helped this too.

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
```diff
@@ -32,28 +34,86 @@
  */
 @InterfaceStability.Evolving
 public class DescribeTopicsResult {
-    private final Map<String, KafkaFuture<TopicDescription>> futures;
+    private final Map<Uuid, KafkaFuture<TopicDescription>> topicIdFutures;
+    private final Map<String, KafkaFuture<TopicDescription>> nameFutures;
 
-    protected DescribeTopicsResult(Map<String, KafkaFuture<TopicDescription>> futures) {
```
rajinisivaram, line 37: Since this is a protected method in a public API, we should probably keep the method and deprecate. It can just invoke the method below.

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
```diff
@@ -32,28 +34,86 @@
...
-        this.futures = futures;
+    protected DescribeTopicsResult(Map<Uuid, KafkaFuture<TopicDescription>> topicIdFutures, Map<String, KafkaFuture<TopicDescription>> nameFutures) {
```
rajinisivaram, line 46: Make this package-private?

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
```diff
@@ -32,28 +34,86 @@
...
+     */
+    public Map<String, KafkaFuture<TopicDescription>> topicNameValues() {
+        return nameFutures;
     }
 
     /**
      * Return a map from topic names to futures which can be used to check the status of
      * individual topics.
      */
+    @Deprecated
```
rajinisivaram, line 89: Can we add ` @deprecated Since 3.1. Use xxx` in the javadoc above?

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
```diff
@@ -32,28 +34,86 @@
...
     public Map<String, KafkaFuture<TopicDescription>> values() {
-        return futures;
+        return nameFutures;
     }
 
     /**
      * Return a future which succeeds only if all the topic descriptions succeed.
      */
+    @Deprecated
```
rajinisivaram, line 102: Can we add ` @deprecated Since 3.1. Use xxx` in the javadoc above?

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
```diff
@@ -32,28 +34,86 @@
...
     public KafkaFuture<Map<String, TopicDescription>> all() {
-        return KafkaFuture.allOf(futures.values().toArray(new KafkaFuture[0])).
+        return all(nameFutures);
+    }
+
+    /**
+     * Return a future which succeeds only if all the topic descriptions succeed.
```
rajinisivaram: Also mention that this returns by topic name if the request used topic names. otherwise returns null.

clients/src/main/java/org/apache/kafka/clients/admin/TopicListing.java
```diff
@@ -17,24 +17,36 @@
...
 
     /**
      * Create an instance with the specified parameters.
      *
      * @param name The topic name
+     * @param topicId The topic id.
      * @param internal Whether the topic is internal to Kafka
      */
-    public TopicListing(String name, boolean internal) {
+    public TopicListing(String name, Uuid topicId, boolean internal) {
```
rajinisivaram, line 51: Since this is part of the public API, we should keep the original constructor and deprecate.

clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
```diff
@@ -91,12 +91,31 @@ public int throttleTimeMs() {
     public Map<String, Errors> errors() {
         Map<String, Errors> errors = new HashMap<>();
         for (MetadataResponseTopic metadata : data.topics()) {
+            if (metadata.name() == null) {
+                throw new IllegalStateException("Use errorsByTopicId() when manage topic using topic id");
```
rajinisivaram: `when manage` => `when managing`

clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java
```diff
@@ -91,12 +91,31 @@ public int throttleTimeMs() {
...
 
+    /**
+     * Get a map of the topicIds which had metadata errors
+     * @return the map
+     */
+    public Map<Uuid, Errors> errorsByTopicId() {
+        Map<Uuid, Errors> errors = new HashMap<>();
+        for (MetadataResponseTopic metadata : data.topics()) {
+            if (metadata.topicId() == Uuid.ZERO_UUID) {
+                throw new IllegalStateException("Use errors() when manage topic using topic name");
```
rajinisivaram: As before, `when manage` =>`when managing`

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1222,16 +1238,23 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+      metadataResponseTopic(Errors.TOPIC_AUTHORIZATION_FAILED, topic, Uuid.ZERO_UUID, isInternal(topic), util.Collections.emptyList()))
 
     // do not disclose the existence of topics unauthorized for Describe, so we've not even checked if they exist or not
     val unauthorizedForDescribeTopicMetadata =
       // In case of all topics, don't include topics unauthorized for Describe
       if ((requestVersion == 0 && (metadataRequest.topics == null || metadataRequest.topics.isEmpty)) || metadataRequest.isAllTopics)
         Set.empty[MetadataResponseTopic]
-      else
+      else if (useTopicId) {
+        // Topic IDs are not considered sensitive information, so we TOPIC_AUTHORIZATION_FAILED is OK
```
rajinisivaram: `we` => returning?

core/src/main/scala/kafka/admin/TopicCommand.scala
```diff
@@ -568,6 +619,10 @@ object TopicCommand extends Logging {
         throw new IllegalArgumentException("--bootstrap-server must be specified")
       if (has(describeOpt) && has(ifExistsOpt))
         CommandLineUtils.checkRequiredArgs(parser, options, topicOpt)
+      if(has(describeOpt) && has(ifExistsOpt)) {
+        if (has(topicOpt) == has(topicIdOpt))
+          println("Only topic id will be used when both --topic and --topic-id are specified and topicId is not UUID.ZERO_UUID")
```
rajinisivaram: Maybe just say `ZERO_UUID` (Or change case to `Uuid`

core/src/main/scala/kafka/admin/TopicCommand.scala
```diff
@@ -568,6 +619,10 @@ object TopicCommand extends Logging {
         throw new IllegalArgumentException("--bootstrap-server must be specified")
       if (has(describeOpt) && has(ifExistsOpt))
         CommandLineUtils.checkRequiredArgs(parser, options, topicOpt)
+      if(has(describeOpt) && has(ifExistsOpt)) {
+        if (has(topicOpt) == has(topicIdOpt))
```
rajinisivaram: Should this be `&&` since the error below refers to both being specified? And should there an error for both not being specified?

rajinisivaram (commented):
@dengziming Thanks for the updates, looks good. Left some minor comments, I think it is ready to go after that.

clients/src/main/java/org/apache/kafka/clients/admin/DescribeTopicsResult.java
dengziming: This is used in `InternalTopicManagerTest` so it should be protected, I just add a "VisibleForTesting" comment to it.

core/src/main/scala/kafka/admin/TopicCommand.scala
dengziming: Thank you for this reminding, I also fixed a bug when using --topic-id and --if-exists.

rajinisivaram (approved):
@dengziming Thanks for the updates, LGTM. Sorry about the delay in getting this in and thanks for your patience! Merging to trunk.

PR #9776, opened 2020-12-22
KAFKA-10878 Check failed message in ProtocolSerializationTest

g1geordie:
use assert to check failed message 

-  testReadWhenOptionalDataMissingAtTheEndIsNotTolerated
-  testReadWithMissingNonOptionalExtraDataAtTheEnd

### DIFF
--- a/clients/src/test/java/org/apache/kafka/common/protocol/types/ProtocolSerializationTest.java
+++ b/clients/src/test/java/org/apache/kafka/common/protocol/types/ProtocolSerializationTest.java
@@ -23,6 +23,8 @@ import org.junit.Test;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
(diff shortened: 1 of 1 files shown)


g1geordie:
@chia7712  can you help my take a look?

chia7712 (approved):

clients/src/test/java/org/apache/kafka/common/protocol/types/ProtocolSerializationTest.java
```diff
@@ -414,7 +416,7 @@ public void testReadWhenOptionalDataMissingAtTheEndIsNotTolerated() {
         oldFormat.writeTo(buffer);
         buffer.flip();
         SchemaException e = assertThrows(SchemaException.class, () -> newSchema.read(buffer));
-        e.getMessage().contains("Error reading field 'field2': java.nio.BufferUnderflowException");
+        assertThat(e.getMessage(), containsString("Error reading field 'field2': java.nio.BufferUnderflowException"));
```
chia7712: Is this error message from java standard library? If so, does it got different on different location (i.e different language)?

g1geordie: The SchemaException is throw by KAFKA. 
it's message is  `fix string + field info   +  e.msg || e.className` 
the lastest string here is className   

but It's seems to me that only check the prefix is more better

chia7712: You are right. Checking ```Error reading field 'field2'``` should be enough for this test case.

PR #9778, opened 2020-12-22
KAFKA-10874 Fix flaky ClientQuotasRequestTest.testAlterIpQuotasRequest

g1geordie:
The error  occcur when dynamic ip config  update after borker's config  get .  
solve this by waiting dynamic ip config update.

### DIFF
--- a/core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
+++ b/core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
@@ -29,6 +29,8 @@ import org.junit.Test
 import java.util
 import java.util.concurrent.{ExecutionException, TimeUnit}
 
+import kafka.utils.TestUtils
+
 import scala.jdk.CollectionConverters._
 
 class ClientQuotasRequestTest extends BaseRequestTest {
@@ -200,7 +202,7 @@ class ClientQuotasRequestTest extends BaseRequestTest {
     val defaultEntityFilter = ClientQuotaFilterComponent.ofDefaultEntity(ClientQuotaEntity.IP)
     val allIpEntityFilter = ClientQuotaFilterComponent.ofEntityType(ClientQuotaEntity.IP)
 
-    def verifyIpQuotas(entityFilter: ClientQuotaFilterComponent, expectedMatches: Map[ClientQuotaEntity, Double]): Unit = {
+    def verifyIpQuotas(entityFilter: ClientQuotaFilterComponent, expectedMatches: Map[ClientQuotaEntity, Double]): Unit = TestUtils.retry(10000L){
       val result = describeClientQuotas(ClientQuotaFilter.containsOnly(List(entityFilter).asJava))
       assertEquals(expectedMatches.keySet, result.asScala.keySet)
       result.asScala.foreach { case (entity, props) =>
--- a/core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
+++ b/core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
@@ -215,6 +215,7 @@ class ClientQuotasRequestTest extends BaseRequestTest {
         else
           InetAddress.getByName(entityName)
         TestUtils.retry(10000L) {
+          //retry until Broker update prop from Zookeeper
           assertEquals(expectedMatches(entity), servers.head.socketServer.connectionQuotas.connectionRateForIp(entityIp), 0.01)
         }
       }
--- a/core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
+++ b/core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
@@ -216,7 +216,7 @@ class ClientQuotasRequestTest extends BaseRequestTest {
           InetAddress.getByName(entityName)
         TestUtils.waitUntilTrue(
           () => expectedMatches(entity) - servers.head.socketServer.connectionQuotas.connectionRateForIp(entityIp) < 0.01
-          ,"Broker didn't update prop from Zookeeper")
+          ,"Broker didn't update quotas from Zookeeper")
       }
     }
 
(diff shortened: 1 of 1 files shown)


g1geordie:
@chia7712  can you help my take a look?

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
```diff
@@ -200,7 +202,7 @@ class ClientQuotasRequestTest extends BaseRequestTest {
     val defaultEntityFilter = ClientQuotaFilterComponent.ofDefaultEntity(ClientQuotaEntity.IP)
     val allIpEntityFilter = ClientQuotaFilterComponent.ofEntityType(ClientQuotaEntity.IP)
 
-    def verifyIpQuotas(entityFilter: ClientQuotaFilterComponent, expectedMatches: Map[ClientQuotaEntity, Double]): Unit = {
+    def verifyIpQuotas(entityFilter: ClientQuotaFilterComponent, expectedMatches: Map[ClientQuotaEntity, Double]): Unit = TestUtils.retry(10000L){
```
chia7712: Is it overkill to retry whole method?

chia7712 (commented):
@g1geordie Thanks for your patch. A minor comment is left.

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
g1geordie: yes , I will change to only retry `assertEquals(expectedMatches(entity), servers.head.socketServer.connectionQuotas.connectionRateForIp(entityIp), 0.01)`

g1geordie:
@chia7712  thank for you comment . please help me take a look again .

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
```diff
@@ -212,7 +214,9 @@ class ClientQuotasRequestTest extends BaseRequestTest {
           InetAddress.getByName(unknownHost)
         else
           InetAddress.getByName(entityName)
-        assertEquals(expectedMatches(entity), servers.head.socketServer.connectionQuotas.connectionRateForIp(entityIp), 0.01)
+        TestUtils.retry(10000L) {
```
chia7712: Could you add comment for this retry?

g1geordie: Yes sir

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
```diff
@@ -212,7 +214,10 @@ class ClientQuotasRequestTest extends BaseRequestTest {
...
```
chia7712: It seems to me ```waitUntilTrue``` is more suitable since the ```exception``` is not necessary in this test case.

g1geordie: Thank you

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
```diff
@@ -212,7 +214,9 @@ class ClientQuotasRequestTest extends BaseRequestTest {
           InetAddress.getByName(unknownHost)
         else
           InetAddress.getByName(entityName)
-        assertEquals(expectedMatches(entity), servers.head.socketServer.connectionQuotas.connectionRateForIp(entityIp), 0.01)
+        TestUtils.waitUntilTrue(
+          () => expectedMatches(entity) - servers.head.socketServer.connectionQuotas.connectionRateForIp(entityIp) < 0.01
+          ,"Broker didn't update prop from Zookeeper")
```
chia7712: Could we rename "prop" to "quotas"?

g1geordie: sure!!

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
```diff
@@ -212,7 +214,9 @@ class ClientQuotasRequestTest extends BaseRequestTest {
           InetAddress.getByName(unknownHost)
         else
           InetAddress.getByName(entityName)
-        assertEquals(expectedMatches(entity), servers.head.socketServer.connectionQuotas.connectionRateForIp(entityIp), 0.01)
+        TestUtils.waitUntilTrue(
+          () => expectedMatches(entity) - servers.head.socketServer.connectionQuotas.connectionRateForIp(entityIp) < 0.01
+          ,"Broker didn't update quotas from Zookeeper")
```
dajac: * We usually put the `,` at the end of the previous line when we break a line.
* Could we add the expected value and the current value in the message? We could say something like `Expected connection quota $expected for $entity, got $actual.`

dajac (commented):
Thanks for the PR. Left a small comment.

core/src/test/scala/unit/kafka/server/ClientQuotasRequestTest.scala
g1geordie: Thank you ! Both are very helpful . let me fix it .

g1geordie: @dajac 
hello~
I remain the message `Broker didn't update quotas from Zookeeper` which clearly indicate the error.
And add  `expect` and `got` to error message

Can you help me take a look again

dajac: Thanks for the update. I think that we should not mention ZK here as ZK is going away. How about `Connection quota of $entity is not $expected but $actual`?

g1geordie: Thank you  ! I have fixed it .

dajac (approved):
LGTM

PR #9781, opened 2020-12-22
MINOR: Use top-level error in `UpdateFeaturesRequest.getErrorResponse`

hachikuji:
The current `getErrorResponse` sets all of the feature errors, but does not set a top-level error. It seems like the whole point of having the top-level error is so that it could be used in cases like this.


clients/src/main/java/org/apache/kafka/common/requests/UpdateFeaturesRequest.java
```diff
@@ -57,18 +55,11 @@ public UpdateFeaturesRequest(UpdateFeaturesRequestData data, short version) {
...
-                .setErrorCode(apiError.error().code())
-                .setErrorMessage(apiError.message());
-            results.add(result);
-        }
-        final UpdateFeaturesResponseData responseData = new UpdateFeaturesResponseData()
-            .setThrottleTimeMs(throttleTimeMs)
-            .setResults(results);
-        return new UpdateFeaturesResponse(responseData);
+        return UpdateFeaturesResponse.createWithErrors(
+            apiError,
```
kowshik: You could inline `ApiError.fromThrowable(e)` here, and eliminate the `apiError` local variable.

kowshik (approved):
@hachikuji: LGTM with small comment inline. Thanks for the patch!

clients/src/main/java/org/apache/kafka/common/requests/UpdateFeaturesRequest.java
```diff
@@ -56,19 +54,11 @@ public UpdateFeaturesRequest(UpdateFeaturesRequestData data, short version) {
...
-                .setErrorMessage(apiError.message());
-            results.add(result);
-        }
-        final UpdateFeaturesResponseData responseData = new UpdateFeaturesResponseData()
-            .setThrottleTimeMs(throttleTimeMs)
-            .setResults(results);
-        return new UpdateFeaturesResponse(responseData);
+        return UpdateFeaturesResponse.createWithErrors(
+            ApiError.fromThrowable(e),
+            Collections.emptyMap(),
```
chia7712, line 59: If we want set only top-level error, we have to make ```UpdateFeaturesResponse#errors``` use top-level error as well. Otherwise, the top-level error can't be propagated correctly. 


see https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/UpdateFeaturesResponse.java#L48

hachikuji: @chia7712 Good catch. Pushed an update with some test cases.

clients/src/main/java/org/apache/kafka/common/requests/UpdateFeaturesResponse.java
```diff
@@ -45,16 +44,18 @@ public UpdateFeaturesResponse(UpdateFeaturesResponseData data) {
...
+    public ApiError topLevelError() {
+        return new ApiError(Errors.forCode(data.errorCode()), data.errorMessage());
     }
 
     @Override
     public Map<Errors, Integer> errorCounts() {
-        return apiErrorCounts(errors());
+        Map<Errors, Integer> errorCounts = new HashMap<>();
+        updateErrorCounts(errorCounts, Errors.forCode(data.errorCode()));
+        for (UpdatableFeatureResult result : data.results().valuesSet()) {
```
chia7712: Is ```valuesSet``` necessary?

chia7712 (commented):
This PR avoids top-level error (when calling ```getErrorResponse```) being propagated to sub data ```UpdatableFeatureResult``` (so the sub data is empty). However, this behavior is not consistent to other protocols (for example: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/AlterPartitionReassignmentsRequest.java#L71)

Also, some response handlers depend on the sub-data to complete futures (see https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L3599 and https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L3725). 

In short, there are two concerns from me.
1. should we make them consistent? propagate top-level to all sub data or not?
2. is it related to compatible issue? for example, server-side does NOT propagate top-level error to sub data but client-side handle response according to sub data.

hachikuji:
@chia7712 Since this is a new api, I think we are free to choose its error semantics. I agree that there is some inconsistency between different APIs, but we are often stuck with that because a top-level error code didn't come until later (e.g. see `FetchResponse`). The benefit of having a top-level error is that we do not need to propagate it. See `AlterIsr` for a similar example. Also note the handling logic in `KafkaApis.handleUpdateFeatures` which already relies on this approach: https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/KafkaApis.scala#L3222.

chia7712 (approved):
@hachikuji Nice patch and thanks for your explanation.

PR #9520, opened 2020-10-28
MINOR: replace test "expected" parameter by assertThrows

chia7712:
`assertThrows` can make checks in testing more flexible and accurate so I think it is worth substituting `assertThrows` for "expected" parameter.

This PR includes following changes.

1. ```@Test(expected = Exception.class)``` is replaced by ```assertThrows```
1. remove reference to ```org.scalatest.Assertions``` 
1. change the magic code from 1 to 2 for ```testAppendAtInvalidOffset``` to test ZSTD
1. rename ```testMaybeAddPartitionToTransactionXXXX``` to ```testNotReadyForSendXXX```
1. increase maxBlockMs from 1s to 3s to avoid unexpected timeout from ```TransactionsTest#testTimeout``` 

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9785, opened 2020-12-24
KAFKA-10887 Migrate log4j-appender module to JUnit 5

g1geordie:
Migrate log4j-appender module to JUnit 5

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -241,7 +241,7 @@ subprojects {
     }
   }
 
-  def shouldUseJUnit5 = ["tools", "raft"].contains(it.project.name)
+  def shouldUseJUnit5 = ["tools", "raft", "log4j-appender"].contains(it.project.name)
   def testLoggingEvents = ["passed", "skipped", "failed"]
   def testShowStandardStreams = false
   def testExceptionFormat = 'full'
@@ -1782,8 +1782,8 @@ project(':log4j-appender') {
     compile libs.slf4jlog4j
 
     testCompile project(':clients').sourceSets.test.output
-    testCompile libs.junitJupiterApi
-    testCompile libs.junitVintageEngine
+    testCompile libs.junitJupiter
+    testCompile libs.hamcrest
     testCompile libs.easymock
   }
 
--- a/log4j-appender/src/test/java/org/apache/kafka/log4jappender/KafkaLog4jAppenderTest.java
+++ b/log4j-appender/src/test/java/org/apache/kafka/log4jappender/KafkaLog4jAppenderTest.java
@@ -16,11 +16,9 @@
  */
 package org.apache.kafka.log4jappender;
 
-import static org.hamcrest.CoreMatchers.equalTo;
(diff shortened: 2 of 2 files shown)


g1geordie:
@chia7712  please help me take a look

build.gradle
```diff
@@ -1782,8 +1782,8 @@ project(':log4j-appender') {
     compile libs.slf4jlog4j
 
     testCompile project(':clients').sourceSets.test.output
-    testCompile libs.junitJupiterApi
-    testCompile libs.junitVintageEngine
+    testCompile libs.junitJupiter
+    testCompile libs.hamcrest
```
chia7712, line 1786: @ijuma Should we remove hamcrest? Removing it causes a bigger patch. The benefit of removing it is that our test assert get more consistent.

chia7712: see https://github.com/apache/kafka/pull/9520#issuecomment-757426244 so keeping ```libs.hamcrest``` is fine.

chia7712 (approved):
@g1geordie LGTM. Thanks for your patch.

log4j-appender/src/test/java/org/apache/kafka/log4jappender/KafkaLog4jAppenderTest.java
```diff
@@ -130,55 +116,51 @@ public void testLog4jAppends() {
...
+    public void testSyncSendWithoutIgnoringExceptionsShouldNotThrowException() {
         Properties props = getLog4jConfig(true);
         props.put("log4j.appender.KAFKA.IgnoreExceptions", "false");
         PropertyConfigurator.configure(props);
 
         MockKafkaLog4jAppender mockKafkaLog4jAppender = getMockKafkaLog4jAppender();
         replaceProducerWithMocked(mockKafkaLog4jAppender, true);
 
-        logger.error(getMessage(0));
+        assertDoesNotThrow(() -> logger.error(getMessage(0)));
```
ijuma, line 143: What is the reason for having `assertDoesNotThrow` here and below? The test will fail if an exception is thrown, so seems like unnecessary noise.

chia7712: I considered it is a kind of highlight. However, it is unnecessary in this test case as it is the final line. @g1geordie WDYT? Should we file a PR to remove this unnecessary noise?

build.gradle
g1geordie: Thank you .
You are so strong that you can track a lot of issues simultaneously !

log4j-appender/src/test/java/org/apache/kafka/log4jappender/KafkaLog4jAppenderTest.java
g1geordie: As you think .  I just point out where we do the check .
I will file a MINOR to change that    . Thank you

PR #9789, opened 2020-12-25
KAFKA-10891:The control plane needs to force the validation of requests from the controller

wenbingshen:
…ect the control plane to be safe and stable.

At present,data requests and control requests use different endpoints,these endpoints are all registered with ZooKeeper.This results in the client business still having access to the metadata that controls the endpoint,this returns the control endpoint back to the data endpoint.One option is to turn on the firewall to disable client access to the control port.But does Kafka need to do something on its own to handle data request access control endpoints to deny those requests.I modified part of the code to deny access control endpoints for data requests.This is the first time I've submitted code to the community,I hope to get some advice, if possible, I will continue to work on this.

### DIFF
--- a/core/src/main/scala/kafka/network/SocketServer.scala
+++ b/core/src/main/scala/kafka/network/SocketServer.scala
@@ -990,17 +990,10 @@ private[kafka] class Processor(val id: Int,
     }
   }
 
-  protected def isControlRequest(header: RequestHeader): Boolean = {
-    if (isControlPlane) {
-      header.apiKey() match {
-        case ApiKeys.LEADER_AND_ISR => true
-        case ApiKeys.STOP_REPLICA => true
-        case ApiKeys.UPDATE_METADATA => true
-        case ApiKeys.CONTROLLED_SHUTDOWN => true
-        case _ => false
-      }
-    }else {
-      true
+  private def isControlRequest(header: RequestHeader): Boolean = {
+    header.apiKey() match {
+      case ApiKeys.LEADER_AND_ISR | ApiKeys.STOP_REPLICA | ApiKeys.UPDATE_METADATA | ApiKeys.CONTROLLED_SHUTDOWN => true
+      case _ => false
     }
   }
 
@@ -1010,10 +1003,10 @@ private[kafka] class Processor(val id: Int,
         openOrClosingChannel(receive.source) match {
           case Some(channel) =>
             val header = parseRequestHeader(receive.payload)
-            if (!isControlRequest(header)) {
-              info(s"Current plane is control-plan, disconnecting non controller channel: $channel : $header")
+            if (isControlPlane && !isControlRequest(header)) {
+              info(s"Current plane is control plane, disconnecting non controller channel: $channel : $header")
               close(channel.id)
-            }else{
+            } else {
               if (header.apiKey == ApiKeys.SASL_HANDSHAKE && channel.maybeBeginServerReauthentication(receive,
                 () => time.nanoseconds()))
                 trace(s"Begin re-authentication: $channel")
--- a/core/src/main/scala/kafka/network/SocketServer.scala
+++ b/core/src/main/scala/kafka/network/SocketServer.scala
@@ -812,6 +812,7 @@ private[kafka] class Processor(val id: Int,
   private val newConnections = new ArrayBlockingQueue[SocketChannel](connectionQueueSize)
   private val inflightResponses = mutable.Map[String, RequestChannel.Response]()
   private val responseQueue = new LinkedBlockingDeque[RequestChannel.Response]()
+  private val forceControllerRequests = config.controlPlaneForceControllerRequestsEnable
 
   private[kafka] val metricTags = mutable.LinkedHashMap(
     ListenerMetricTag -> listenerName.value,
@@ -1003,7 +1004,7 @@ private[kafka] class Processor(val id: Int,
         openOrClosingChannel(receive.source) match {
           case Some(channel) =>
             val header = parseRequestHeader(receive.payload)
-            if (isControlPlane && !isControlRequest(header)) {
+            if (isControlPlane && forceControllerRequests && !isControlRequest(header)) {
               info(s"Current plane is control plane, disconnecting non controller channel: $channel : $header")
               close(channel.id)
             } else {
--- a/core/src/main/scala/kafka/network/SocketServer.scala
+++ b/core/src/main/scala/kafka/network/SocketServer.scala
@@ -260,7 +260,7 @@ class SocketServer(val config: KafkaConfig,
       connectionQuotas.addListener(config, endpoint.listenerName)
       val controlPlaneAcceptor = createAcceptor(endpoint, ControlPlaneMetricPrefix)
       val controlPlaneProcessor = newProcessor(nextProcessorId, controlPlaneRequestChannelOpt.get,
-        connectionQuotas, endpoint.listenerName, endpoint.securityProtocol, memoryPool, isPrivilegedListener = true)
+        connectionQuotas, endpoint.listenerName, endpoint.securityProtocol, memoryPool, isPrivilegedListener = true, isControlPlane = true)
       controlPlaneAcceptorOpt = Some(controlPlaneAcceptor)
       controlPlaneProcessorOpt = Some(controlPlaneProcessor)
       val listenerProcessors = new ArrayBuffer[Processor]()
@@ -287,7 +287,7 @@ class SocketServer(val config: KafkaConfig,
 
     for (_ <- 0 until newProcessorsPerListener) {
       val processor = newProcessor(nextProcessorId, dataPlaneRequestChannel, connectionQuotas,
-        listenerName, securityProtocol, memoryPool, isPrivilegedListener)
+        listenerName, securityProtocol, memoryPool, isPrivilegedListener, false)
       listenerProcessors += processor
       dataPlaneRequestChannel.addProcessor(processor)
       nextProcessorId += 1
@@ -413,7 +413,7 @@ class SocketServer(val config: KafkaConfig,
 
   // `protected` for test usage
   protected[network] def newProcessor(id: Int, requestChannel: RequestChannel, connectionQuotas: ConnectionQuotas, listenerName: ListenerName,
-                                      securityProtocol: SecurityProtocol, memoryPool: MemoryPool, isPrivilegedListener: Boolean): Processor = {
+                                      securityProtocol: SecurityProtocol, memoryPool: MemoryPool, isPrivilegedListener: Boolean, isControlPlane: Boolean): Processor = {
     new Processor(id,
       time,
       config.socketRequestMaxBytes,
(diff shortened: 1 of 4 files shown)


wenbingshen:
@MayureshGharat Can you take a look at this?

wenbingshen:
@showuon This is my first time to submit pr to the community.Could you give me some suggestions?Thank you very much!

showuon:
@wenbingshen , thanks for the PR. Some comments left, but most importantly, we don't know which JIRA ticket you're fixing. Please check the [contributing code change](https://cwiki.apache.org/confluence/display/KAFKA/Contributing+Code+Changes) page. Thank you.
> The PR title should usually be of the form KAFKA-xxxx: Title, where KAFKA-xxxx is the relevant JIRA id and Title may be the JIRA's title or a more specific title describing the PR itself. For trivial cases where a JIRA is not required (see JIRA section for more details) MINOR: or HOTFIX: can be used as the PR title prefix.

Also, please add tests. Thanks.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -988,60 +990,79 @@ private[kafka] class Processor(val id: Int,
...
-                  req = parseForwardedPrincipal(req, channel.principalSerde.asScala) match {
-                    case Some(forwardedPrincipal) =>
-                      buildForwardedRequestContext(req, forwardedPrincipal)
-
-                    case None =>
-                      val envelopeResponse = new EnvelopeResponse(Errors.PRINCIPAL_DESERIALIZATION_FAILURE)
-                      sendEnvelopeResponse(req, envelopeResponse)
-                      null
+            if (!isControlRequest(header)) {
+              info(s"Current plane is control-plan, disconnecting non controller channel: $channel : $header")
```
showuon: typo: `control-plane`

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -988,60 +990,79 @@ private[kafka] class Processor(val id: Int,
...
+              close(channel.id)
+            }else{
```
showuon: formatting is not correct. `} else {`

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -988,60 +990,79 @@ private[kafka] class Processor(val id: Int,
...
+  protected def isControlRequest(header: RequestHeader): Boolean = {
+    if (isControlPlane) {
+      header.apiKey() match {
+        case ApiKeys.LEADER_AND_ISR => true
+        case ApiKeys.STOP_REPLICA => true
+        case ApiKeys.UPDATE_METADATA => true
+        case ApiKeys.CONTROLLED_SHUTDOWN => true
+        case _ => false
+      }
+    }else {
```
showuon: formatting is not correct. `} else {`

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -988,60 +990,79 @@ private[kafka] class Processor(val id: Int,
     }
   }
 
+  protected def isControlRequest(header: RequestHeader): Boolean = {
```
showuon: could we use `private` here?

core/src/main/scala/kafka/network/SocketServer.scala
showuon: Not sure if this is better=:
```scala
  header.apiKey() match {
    case ApiKeys.LEADER_AND_ISR | ApiKeys.STOP_REPLICA | ... | ... => true
    case _ => false
  }
```

wenbingshen:
@showuon  Thank you very much for your suggestions and comments,In order to meet the requirement of submitting PR, do I need to submit a PR again?I'm going to create a new jIRA ID for this PR and add test cases.

showuon:
> @showuon Thank you very much for your suggestions and comments,In order to meet the requirement of submitting PR, do I need to submit a PR again?I'm going to create a new jIRA ID for this PR and add test cases.

No, just update in this PR directly.

wenbingshen:
@showuon Hello,I made some changes and added a test case.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -988,60 +994,72 @@ private[kafka] class Processor(val id: Int,
...
-                  req = parseForwardedPrincipal(req, channel.principalSerde.asScala) match {
-                    case Some(forwardedPrincipal) =>
-                      buildForwardedRequestContext(req, forwardedPrincipal)
-
-                    case None =>
-                      val envelopeResponse = new EnvelopeResponse(Errors.PRINCIPAL_DESERIALIZATION_FAILURE)
-                      sendEnvelopeResponse(req, envelopeResponse)
-                      null
+            if (isControlPlane && !isControlRequest(header)) {
+              info(s"Current plane is control plan, disconnecting non controller channel: $channel : $header")
```
showuon: typo: control plane

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -61,17 +61,17 @@ import scala.util.control.ControlThrowable
...
+ * Acceptor has N Processor threads that each have their own selector and read requests from sockets
+ * M Handler threads that handle requests and produce responses back to the processor threads for writing.
  *  - control-plane :
  *    - Handles requests from controller. This is optional and can be configured by specifying "control.plane.listener.name".
- *      If not configured, the controller requests are handled by the data-plane.
+ * If not configured, the controller requests are handled by the data-plane.
  *    - The threading model is
- *      1 Acceptor thread that handles new connections
- *      Acceptor has 1 Processor thread that has its own selector and read requests from the socket.
- *      1 Handler thread that handles requests and produce responses back to the processor thread for writing.
```
showuon: I guess this change is completed by fomatter tool? You might need to manually revert these changes, and some empty line changes that doesn't relate to your fix.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1089,9 +1107,9 @@ private[kafka] class Processor(val id: Int,
   }
 
   private def buildForwardedRequestContext(
-    envelopeRequest: RequestChannel.Request,
-    forwardedPrincipal: KafkaPrincipal
-  ): RequestChannel.Request = {
+                                            envelopeRequest: RequestChannel.Request,
+                                            forwardedPrincipal: KafkaPrincipal
+                                          ): RequestChannel.Request = {
```
showuon: Please revert this unnecessary change. Thanks.

core/src/main/scala/kafka/network/SocketServer.scala
```diff
@@ -1331,24 +1352,34 @@ object ConnectionQuotas {
 
   private case class ListenerQuotaEntity(listenerName: String) extends ConnectionQuotaEntity {
     override def sensorName: String = s"$ConnectionRateSensorName-$listenerName"
+
     override def sensorExpiration: Long = Long.MaxValue
+
```
showuon: Are these empty lines needed ?

core/src/test/scala/unit/kafka/network/SocketServerTest.scala
```diff
@@ -1352,7 +1369,7 @@ class SocketServerTest {
 
   private def verifySendFailureAfterRemoteClose(makeClosing: Boolean): Unit = {
     props ++= sslServerProps
-    withTestableServer (testWithServer = { testableServer =>
+    withTestableServer(testWithServer = { testableServer =>
```
showuon: I don't think this is necessary, and other places.

wenbingshen:
@showuon  Thank you,Luke,I did accidentally use the formatting tool.I submitted the latest code.Thank you very much for your help.

wenbingshen:
@chia7712 Can you take a look at this PR?Thank you.

chia7712:
@wenbingshen Thanks for your patch. This is an interesting issue. Please take a look at following questions.

1. Is there a case that client sends controller-unrelated request to controller plane ? I guess that users define "bootstrap servers" as controller listener name can reproduce this issue?
1. How to keep the compatibility if our kafka users do have this cases (This PR can break such behavior)? 
1. Does ```inter.broker.listener.name``` have similar issue?

wenbingshen:
@chia7712  Thanks for your comments and questions.I think we just need to completely separate control requests from data requests, and provide the user with an optional configuration to control whether the control request is forced to validate, which can be disabled by default.There is no way to do this now, when the user needs to fully protect control from possible client requests.This patch can do that.On the [KIP-291](https://cwiki.apache.org/confluence/display/KAFKA/KIP-291%3A+Separating+controller+connections+and+requests+from+the+data+plane) basis,we can ignore inter.broker.listener.name.

chia7712:
@wenbingshen Thanks for your response. 

> I think we just need to completely separate control requests from data requests, and provide the user with an optional configuration to control whether the control request is forced to validate, which can be disabled by default.There is no way to do this now, when the user needs to fully protect control from possible client requests.

I feel this needs a KIP as it proposed a new config and it is related to compatibility.

for another, was my comment valid? 
> Is there a case that client sends controller-unrelated request to controller plane ? I guess that users define "bootstrap servers" as controller listener name can reproduce this issue?

I'm curious whether there are other ways which can produce this issue.

wenbingshen:
@chia7712 Thanks for your comment.I'm sorry that I forgot to reply to the first question. During my use, I found no other way to cause this problem except to define the configuration "bootstrap.servers" as the control plane.I haven't created KIP yet. If possible, could you give me some guidance?Thanks.

> @wenbingshen Thanks for your response.
> 
> > I think we just need to completely separate control requests from data requests, and provide the user with an optional configuration to control whether the control request is forced to validate, which can be disabled by default.There is no way to do this now, when the user needs to fully protect control from possible client requests.
> 
> I feel this needs a KIP as it proposed a new config and it is related to compatibility.
> 
> for another, was my comment valid?
> 
> > Is there a case that client sends controller-unrelated request to controller plane ? I guess that users define "bootstrap servers" as controller listener name can reproduce this issue?
> 
> I'm curious whether there are other ways which can produce this issue.

chia7712:
@wenbingshen see https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Improvement+Proposals for guidance. In short, you have to complete following tasks.

1. require KIP permission (post to https://lists.apache.org/list.html?dev@kafka.apache.org)
1. write KIP
1. post KIP to https://lists.apache.org/list.html?dev@kafka.apache.org to get discussion and votes

wenbingshen:
@chia7712  Thanks for your guidance. I will apply for KIP permissions and learn how to create KIP.

> @wenbingshen see https://cwiki.apache.org/confluence/display/KAFKA/Kafka+Improvement+Proposals for guidance. In short, you have to complete following tasks.
> 
> 1. require KIP permission (post to https://lists.apache.org/list.html?dev@kafka.apache.org)
> 2. write KIP
> 3. post KIP to https://lists.apache.org/list.html?dev@kafka.apache.org to get discussion and votes

PR #9790, opened 2020-12-27
Some parameters will be overwritten which was configured in consumer.config.

SweetMojito:
Some parameters will be overwritten which was configured in consumer.config where running "ConsumerPerformance.scala"
linked see https://issues.apache.org/jira/browse/KAFKA-10043

comment to ping reviewers.*


SweetMojito:
> Could you update the code (according to the reviews) and then push the changes to be a new commit. Also, please rebase your code (by trunk).
@chia7712  Can you take a look at this?

chia7712:
Could you add test for this issue?

SweetMojito:
> Could you add test for this issue?

Do you mean to add unit tests？But i did not found unit tests for this class `kafka.tools.ConsumerPerformance.scala`

chia7712:
> But i did not found unit tests for this class kafka.tools.ConsumerPerformance.scala

It would be great if you could be the first person to write the tests for  ```ConsumerPerformance``` :)

SweetMojito:
> > But i did not found unit tests for this class kafka.tools.ConsumerPerformance.scala
> 
> It would be great if you could be the first person to write the tests for `ConsumerPerformance` :)

I had committed the unit test in `ConsumerPerformanceTest`, What should I do to relate to that?

SweetMojito:
> > > But i did not found unit tests for this class kafka.tools.ConsumerPerformance.scala
> > 
> > 
> > It would be great if you could be the first person to write the tests for `ConsumerPerformance` :)
> 
> I had committed the unit test in `ConsumerPerformanceTest`, What should I do to relate to that?

I found that it was automatically linked，but all checks had failed :( .
what should I do next :)

chia7712:
> what should I do next :)

please take a look at the report and fix the error. https://github.com/apache/kafka/pull/9790/checks?check_run_id=1616222522

SweetMojito:
> > what should I do next :)
> 
> please take a look at the report and fix the error. https://github.com/apache/kafka/pull/9790/checks?check_run_id=1616222522

I fixed the error, but there were many errors in the second build. It seemed that the errors had nothing to do with what I changed. Please help me to have a look :)

chia7712:
@itantiger Thanks for your patch. Could you revise the title?

SweetMojito:
> @itantiger Thanks for your patch. Could you revise the title?

Also thank you very much for your guidance :) .

SweetMojito:
> @itantiger Thanks for your patch. Could you revise the title?

Finished.
When will this patch merge to trunk?

core/src/test/scala/unit/kafka/tools/ConsumerPerformanceTest.scala
```diff
@@ -96,6 +97,28 @@ class ConsumerPerformanceTest {
...
     assertEquals(10, config.numMessages)
   }
+  
+  @Test
+  def testConfigWithRecognizedOptionOverride(): Unit = {
+    val propsFile = TestUtils.tempFile()
+    val propsStream = Files.newOutputStream(propsFile.toPath)
+    propsStream.write("group.id=test_group_id\n".getBytes())
+    propsStream.write("client.id=test_client_id".getBytes())
+    propsStream.close()
```
chia7712, line 107: Could you test ```--topic``` to make sure it is not overrided by command-line arguments.

chia7712 (commented):
@itantiger Thanks for your patch. one small comment is left. Please take a look :)

core/src/test/scala/unit/kafka/tools/ConsumerPerformanceTest.scala
SweetMojito: `--topic` is a `REQUIRED` argument in command-line, I thank Its value is determined by the command line.

SweetMojito: So, if config file has the param `topic=test`, It will not take effect.

chia7712: Or could we add ```group.id``` to command-line arguments to make sure it (from file) is NOT override by command-line args?

SweetMojito: Okey, I finished the local test and submitted the code :)

SweetMojito: @chia7712  Please take a look :) Is there anything else that needs to be optimized？

chia7712 (approved):
@itantiger Thanks for your great patch. LGTM. Will merge it tomorrow :)

chia7712:
@itantiger The priority of loading props is a bit chaos in this case. It seems to me the property of loading properties is shown below.
1. command-lind
1. config file (consumer.config)
1. default value

The current priority is shown below.
1. command-lind
1. default value
1. config file (consumer.config)

The priority from this PR is shown below.
1. config file (consumer.config)
1. command-lind
1. default value


Of course, current priority has issue as properties from config file always get overwritten by default value. We need to fix it. However, I'd like to check the priority again. WDYT? Should config file has higher priority than command-lind?

SweetMojito:
> 1. command-lind
> 2. config file (consumer.config)
> 3. default value
I think this priority is more better, and this PR is to do this.

chia7712:
> I think this priority is more better, and this PR is to do this.

We load the props from config file first. https://github.com/apache/kafka/blob/afa5423356d3d2a2135a51200573b45d097f6d60/core/src/main/scala/kafka/tools/ConsumerPerformance.scala#L275 and then call ```props.putIfAbsent``` to set args from command-line only if the prop is nonexistent in config file. Hence, the priority of config file is higher than command-line, right?

SweetMojito:
> and then call `props.putIfAbsent` to set args from command-line only if the prop is nonexistent in config file. Hence, the priority of config file is higher than command-line, right?
You're right. I've ignored it. It should be modified here. :)

SweetMojito:
> > and then call `props.putIfAbsent` to set args from command-line only if the prop is nonexistent in config file. Hence, the priority of config file is higher than command-line, right?
> > You're right. I've ignored it. It should be modified here. :)

SweetMojito:
Current priority, some config params would be overwritten by default value, Do you think it needs to be revised?

chia7712:
> Do you think it needs to be revised?

I guess you are talking about "title". yes, the title need to be revised.

itantiger:
> I guess you are talking about "title". yes, the title need to be revised.
In addition, I think the trunk also needs to be modified.
Some config params would be overwritten by default value, WDYT?

SweetMojito:
> I guess you are talking about "title". yes, the title need to be revised.

In addition, I think the trunk also needs to be modified.
Some config params would be overwritten by default value, WDYT?

chia7712:
> I think the trunk also needs to be modified.

trunk?  pardon me that I can't catch your point.

> Some config params would be overwritten by default value, WDYT?

yep, the default value CAN overwrite the configs from either command-line or config file.

itantiger:
> >trunk? pardon me that I can't catch your point.
> That's my problem. hah
>> yep, the default value CAN overwrite the configs from either command-line or config file.
> Let me optimize it again and commit. Do you think it's necessary?

SweetMojito:
>>trunk? pardon me that I can't catch your point.
>
>That's my problem. hah
>
>>yep, the default value CAN overwrite the configs from either command-line or config file.
>
>Let me optimize it again and commit. Do you think it's necessary?

chia7712:
> Let me optimize it again and commit. Do you think it's necessary?

I'd like to address the following priority rule in this PR :)

1. command-lind
2. config file (consumer.config)
3. default value

SweetMojito:
> I'd like to address the following priority rule in this PR :)
> 
> 1. command-lind
> 2. config file (consumer.config)
> 3. default value

Of course, I also think this is more reasonable

PR #9797, opened 2020-12-30
KAFKA-10893 Increase target_messages_per_sec of ReplicaScaleTest to r…

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-10893

The expected size is 3400000 and the throttle is 10000 msg/sec so the expected minimal run time of one consumer/producer is about 6 mins. Hence, the test case is always timeout on travis (6 + 6 > 10 mins)


mumrah (approved):
Thanks @chia7712. That stdout timeout thing is annoying, but its better than nothing I suppose.

LGTM

chia7712:
@mumrah Thanks for your reviews. Merge it to trunk

PR #9799, opened 2020-12-30
QuotaRecord support in broker

mumrah:
This PR adds support for handling QuotaRecord metadata in the broker. It also includes a new QuotaCache class for efficient lookups of the quotas.


mumrah:
Wrong remote/base branch

PR #9800, opened 2020-12-30
KAFKA-9274: Fix commit-TimeoutException handling for EOS

mjsax:
If EOS is enabled and the TX commit fails with a timeout,
we should not process more messages (what is ok for non-EOS)
because we don't really know the status of the TX.
If the commit was indeed successful, we won't have an open TX
and calling send() would fail with a fatal error.

Instead, we should retry the (idempotent) commit of the TX,
and start a new TX afterwards.

Call for review @vvcephei @abbccdda @guozhangwang @hachikuji @bob-barrett

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
@@ -103,7 +103,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
     private long idleStartTimeMs;
     private boolean commitNeeded = false;
     private boolean commitRequested = false;
-    private boolean hasPendingTransaction = false;
+    private boolean hasPendingTxCommit = false;
 
     public StreamTask(final TaskId id,
                       final Set<TopicPartition> partitions,
@@ -379,7 +379,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
                     // TODO: this should be removed after we decouple caching with emitting
                     stateMgr.flushCache();
                     recordCollector.flush();
-                    hasPendingTransaction = eosEnabled;
+                    hasPendingTxCommit = eosEnabled;
 
                     log.debug("Prepared {} task for committing", state());
                     return committableOffsetsAndMetadata();
@@ -478,7 +478,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
 
         commitRequested = false;
         commitNeeded = false;
-        hasPendingTransaction = false;
+        hasPendingTxCommit = false;
     }
 
     private Map<TopicPartition, Long> extractPartitionTimes() {
@@ -636,7 +636,9 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
             return false;
         }
 
-        if (hasPendingTransaction) {
+        if (hasPendingTxCommit) {
+            // if the task has a pending TX commit, we should just retry the commit but not process any records
+            // thus, the task is not processable, even if there is available data in the record queue
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -633,6 +636,10 @@ public boolean isProcessable(final long wallClockTime) {
             return false;
         }
 
+        if (hasPendingTransaction) {
```
abbccdda: Could we add a comment here to explain why a task is not processable under pending transaction?

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -408,9 +417,74 @@ public void shouldProcessInOrder() {
...
+            getConsumerRecord(partition1, 20),
+            getConsumerRecord(partition1, 30)
+        ));
+
+        assertTrue(task.process(time.milliseconds()));
+        task.prepareCommit();
+        assertFalse(task.process(time.milliseconds()));
+        task.postCommit(false);
+        assertTrue(task.process(time.milliseconds()));
+        assertTrue(task.process(time.milliseconds()));
```
abbccdda, line 458: Why do we need to assert twice?

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
```diff
@@ -408,9 +417,74 @@ public void shouldProcessInOrder() {
         assertEquals(3, source2.numReceived);
     }
 
+    @Test
+    public void shouldProcessRecordsAfterPrepareCommitWhenEosDisabled() {
+        task = createStatelessTask(createConfig(), StreamsConfig.METRICS_LATEST);
```
abbccdda, line 422: nit: we may extract the common test logic to be reused by all 3 process record tests.

abbccdda (commented):
One high level question, which part of the change is triggering a retry of TX commit? I assume `isProcessable` will return false once we are already within a transaction, how could we break out of this check?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
mjsax: Sure. Just realizing that the variable name is not ideal -- it actually about an "pending TX _commit_"...

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
mjsax: We need to call `process()` twice here. We have 3 input records, and want to confirm that all three are processed. If we don't call `process()` two times here, the assert in L460 below would not return `false`.

mjsax:
> which part of the change is triggering a retry of TX commit?

The main processing loop does, ie, we don't really need any change to get a retry. In each loop we iterate over all tasks, and if the tasks flag "needCommit" is set to `true` it would retry to commit the task (note, that we don't reset the flag to `false` if we hit a timeout) when commit-interval-ms passed. I though it's ok to just "pause" processing for a full commit interval for this case -- we could of course also set `commitRequested=true` to retry the commit directly in the next iteration instead of waiting for the next commit-interval to pass. Thoughts?

> I assume isProcessable will return false once we are already within a transaction

No. Before this change, `isProcessable()` returns `true` if there is available data in our internal buffer. It does not know anything about TX.

> how could we break out of this check

Not sure if I can follow?

streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamTaskTest.java
mjsax: Thinking about it, the common test logic is so small, it seems to make the test hard to read if we extract it?

abbccdda (approved):
LGTM, thanks for the patch!

vvcephei (approved):
LGTM! Thanks, @mjsax

PR #9802, opened 2020-12-30
KAFKA-10894; Ensure PartitionInfo replicas are not null in client quota callback

hachikuji:
Previously offline replicas were included as `null` in the array of replicas in `PartitionInfo` when populated by the `MetadataCache` for the purpose of the client quota callback. This patch instead initializes empty non-null nodes, which is consistent with how `PartitionInfo` is constructed by the clients in `MetadataResponse`.

### DIFF
--- a/core/src/main/scala/kafka/server/MetadataCache.scala
+++ b/core/src/main/scala/kafka/server/MetadataCache.scala
@@ -266,8 +266,16 @@ class MetadataCache(brokerId: Int) extends Logging {
 
   def getClusterMetadata(clusterId: String, listenerName: ListenerName): Cluster = {
     val snapshot = metadataSnapshot
-    val nodes = snapshot.aliveNodes.map { case (id, nodes) => (id, nodes.get(listenerName).orNull) }
-    def node(id: Integer): Node = nodes.get(id.toLong).orNull
+    val nodes = snapshot.aliveNodes.flatMap { case (id, nodesByListener) =>
+      nodesByListener.get(listenerName).map { node =>
+        id -> node
+      }
+    }
+
+    def node(id: Integer): Node = {
+      nodes.getOrElse(id.toLong, new Node(id, "", -1))
+    }
+
     val partitions = getAllPartitions(snapshot)
       .filter { case (_, state) => state.leader != LeaderAndIsr.LeaderDuringDelete }
       .map { case (tp, state) =>
@@ -278,7 +286,7 @@ class MetadataCache(brokerId: Int) extends Logging {
       }
     val unauthorizedTopics = Collections.emptySet[String]
     val internalTopics = getAllTopics(snapshot).filter(Topic.isInternal).asJava
-    new Cluster(clusterId, nodes.values.filter(_ != null).toBuffer.asJava,
+    new Cluster(clusterId, nodes.values.toBuffer.asJava,
       partitions.toBuffer.asJava,
       unauthorizedTopics, internalTopics,
       snapshot.controllerId.map(id => node(id)).orNull)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/kafka/server/MetadataCache.scala
```diff
@@ -266,8 +266,16 @@ class MetadataCache(brokerId: Int) extends Logging {
...
-    val nodes = snapshot.aliveNodes.map { case (id, nodes) => (id, nodes.get(listenerName).orNull) }
-    def node(id: Integer): Node = nodes.get(id.toLong).orNull
+    val nodes = snapshot.aliveNodes.flatMap { case (id, nodesByListener) =>
+      nodesByListener.get(listenerName).map { node =>
+        id -> node
+      }
+    }
+
+    def node(id: Integer): Node = {
+      nodes.getOrElse(id.toLong, new Node(id, "", -1))
```
chia7712, line 276: How about using a helper method to create ```new Node(id, "", -1)``` ? It can be used by ```MetadataResponse``` also.

chia7712 (commented):
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/MetadataResponse.java#L150 

Is this null (```metadata.leaderId.map(nodesById::get).orElse(null)```) should be replaced also?

core/src/main/scala/kafka/server/MetadataCache.scala
hachikuji: That's not a bad idea. I think `Node` is public, though we haven't always been strict about requiring KIPs for new methods. A couple options are to add a new constructor or a static factory. We could also add a utility somewhere, but that feels a little clumsy. Thoughts?

hachikuji:
@chia7712 Good question about `leader()`. I looked into usage and it seems we do not have consistent `isEmpty` checks, so I'm a little concerned about compatibility if we change that. To be honest, I do not like the use of the empty sentinel. It seems like a dangerous practice and I'm not too eager to expand its usage. However, given the choice between returning a list of replica nodes which contains null values (what is done today for this callback) and making use of the empty sentinel in the same way  that it is already used in the client logic, the latter seems preferable. Does that make sense?

core/src/main/scala/kafka/server/MetadataCache.scala
chia7712: >  I think Node is public, though we haven't always been strict about requiring KIPs for new methods. A couple options are to add a new constructor or a static factory. We could also add a utility somewhere, but that feels a little clumsy. Thoughts?

You are right.

For another, is this a kind of behavior change to ```ClientQuotaCallback``` (if custom ```ClientQuotaCallback``` does depend on the ```null``` value)? If so, how we keep the compatibility or should we offer a helper method to enable ```ClientQuotaCallback``` to validate the "new Node(id, "", -1)"?

hachikuji: I think what we are implementing here is what is documented in `PartitionInfo`:
```java
    /**
     * The complete set of replicas for this partition regardless of whether they are alive or up-to-date
     */
    public Node[] replicas() {
        return replicas;
    }
```
Even the internal logic in `PartitionInfo` assumes non-null values in the array (I caught this issue because of an NPE in the `toString`). We probably can't 100% guarantee that all quota callbacks have appropriate `isEmpty` checks, but my guess is that usage of this plugin is rare in any case and reliance on the replica set is less likely than reliance on the leader. All in all, I think we're better fixing the bug.

chia7712: > my guess is that usage of this plugin is rare in any case and reliance on the replica set is less likely than reliance on the leader. All in all, I think we're better fixing the bug.

make sense to me. Thanks for your explanation!

chia7712 (approved):
LGTM

PR #9803, opened 2020-12-30
MINOR: Don't include stack trace for StateChangeFailedException

hachikuji:
The stack trace for `StateChangeFailedException` is useless since the exception is just used to propagate an error message to the state change logger. This patch eliminates the stack trace which should help to clean up logging a little bit. 

### DIFF
--- a/core/src/main/scala/kafka/common/StateChangeFailedException.scala
+++ b/core/src/main/scala/kafka/common/StateChangeFailedException.scala
@@ -17,7 +17,6 @@
 
 package kafka.common
 
-class StateChangeFailedException(message: String, cause: Throwable) extends RuntimeException(message, cause) {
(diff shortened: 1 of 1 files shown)


hachikuji:
@ijuma @chia7712 Thoughts on this PR? I was getting tired of seeing these useless traces in logging...

chia7712:
> The stack trace for StateChangeFailedException is useless since the exception is just used to propagate an error message to the state change logger.

https://github.com/apache/kafka/blob/afa5423356d3d2a2135a51200573b45d097f6d60/core/src/main/scala/kafka/controller/KafkaController.scala#L1208 Does this case also propagate error message only? 

For another, how about having a variety of ```logFailedStateChange``` which accepts error message (string) rather than ```Exception``` type?

PR #9807, opened 2021-01-01
KAFKA-10779; Reassignment tool sets throttles incorrectly when overriding a reassignment

dengziming:
1. The addingReplicas is included in the AR during reassignment, so remove addingReplicas from AR  in `calculateCurrentMoveMap`
2.  The addingReplicas is also included in MetadataResponse, so only use MetadataResponse data when there is no existing reassignment of this partition
3. Alter test code data
4. Improve some code style

### DIFF
--- a/core/src/main/scala/kafka/admin/ReassignPartitionsCommand.scala
+++ b/core/src/main/scala/kafka/admin/ReassignPartitionsCommand.scala
@@ -1237,16 +1237,15 @@ object ReassignPartitionsCommand extends Logging {
     val moveMap = new mutable.HashMap[String, mutable.Map[Int, PartitionMove]]()
     // Add the current reassignments to the move map.
     currentReassignments.foreach { case (part, reassignment) =>
-      val partMoves = moveMap.getOrElseUpdate(part.topic, new mutable.HashMap[Int, PartitionMove])
+      val allReplicas = reassignment.replicas().asScala.map(Int.unbox)
+      val addingReplicas = reassignment.addingReplicas.asScala.map(Int.unbox)
 
       // The addingReplicas is included in the replicas during reassignment
-      val sources = mutable.Set[Int]() ++ reassignment.replicas().asScala.map(Int.unbox)
-        .diff(reassignment.addingReplicas.asScala.map(Int.unbox))
+      val sources = mutable.Set[Int]() ++ allReplicas.diff(addingReplicas)
+      val destinations = mutable.Set[Int]() ++ addingReplicas
 
-      val destinations = mutable.Set[Int]() ++ reassignment.addingReplicas.asScala.map(Int.unbox)
-
-      partMoves.put(part.partition,
-        PartitionMove(sources, destinations))
+      val partMoves = moveMap.getOrElseUpdate(part.topic, new mutable.HashMap[Int, PartitionMove])
+      partMoves.put(part.partition, PartitionMove(sources, destinations))
     }
     moveMap
   }
--- a/core/src/main/scala/kafka/admin/ReassignPartitionsCommand.scala
+++ b/core/src/main/scala/kafka/admin/ReassignPartitionsCommand.scala
@@ -597,17 +597,17 @@ object ReassignPartitionsCommand extends Logging {
                                 : String = {
     val bld = new mutable.ArrayBuffer[String]
     states.keySet.toBuffer.sortWith(compareTopicPartitionReplicas).foreach {
-      case replica =>
+      replica =>
         val state = states(replica)
         state match {
-          case MissingLogDirMoveState(targetLogDir) =>
+          case MissingLogDirMoveState(_) =>
             bld.append(s"Partition ${replica.topic()}-${replica.partition()} is not found " +
               s"in any live log dir on broker ${replica.brokerId()}. There is likely an " +
               s"offline log directory on the broker.")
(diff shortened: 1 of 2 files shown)


dengziming:
Hi, @hachikuji ,PTAL.

core/src/main/scala/kafka/admin/ReassignPartitionsCommand.scala
```diff
@@ -1250,38 +1252,36 @@ object ReassignPartitionsCommand extends Logging {
...
    * @param currentReassignments    The currently active reassignments.
-   * @param proposedReassignments   The proposed reassignments (destinations replicas only).
+   * @param proposedParts           The proposed location of the partitions (destinations replicas only).
    * @param currentParts            The current location of the partitions that we are
    *                                proposing to move.
    * @return                        A map from topic name to partition map.
    *                                The partition map is keyed on partition index and contains
    *                                the movements for that partition.
    */
   def calculateProposedMoveMap(currentReassignments: Map[TopicPartition, PartitionReassignment],
```
hachikuji, line 1264: It might be just me, but the logic in this method is as clear as mud to me. At a high level, we are just trying to identify the sources and the destinations to fill a `PartitionMove`. The destinations are the adding replicas (AR) and the sources are the current replicas (CR) without the adding replicas. The current logic below first calls `calculateCurrentMoveMap` which does the following:

1. Add each replica to both sources and destinations
2. Add all AR to destinations 
3. Remove all removing replicas (RR) from destinations

I think this computes the destinations correctly (even if it does so in an odd way), but sources will include all replicas, which is wrong. You have fixed this by removing AR from sources explicitly, but I think we can simplify a bit more. Maybe something like this

```scala
move.destinations = reassignment.addingReplicas
move.sources = reassignment.replicas - reassignment.addingReplicas
```

After computing the current move map, the current logic tries to account for the proposed reassignments. It does the following:

1. Overwrite current move with an empty move and set destinations to the target replicas (TR)
2. Iterate through all moves and check if sources is empty, which means we have overwritten the move in step 1. If so, then we add all of CR to sources.
3. Now we do one more pass over the moves and remove all of the destinations from sources.

I have a hard time making sense of this logic. The main problem is that it assumes that TR are only destinations, but that is wrong. For example, if we reassign [1, 2] to [1, 3], then 1 should be a source, not a destination even though it is present in TR. I think the correct logic should be the following:

1. for each proposed assignment TR, check if there is already a reassignment in progress:
2. if there is a reassignment in progress, then we need to account for the current AR. We can compute sources as CR - AR, which is what we already did when calculating the current move map
3. if there is no reassignment in progress, then we set sources to CR
4. regardless of whether there is a reassignment in progress, we set destinations to TR - sources

Something like this?
```scala
proposedParts.foreach {
  case (part, targetReplicas) =>
    val partMoves = moveMap.getOrElseUpdate(part.topic, new mutable.HashMap[Int, PartitionMove])
    val sources = partMoves.get(part) match {
      case Some(move) => move.sources
      case None => currentParts.get(part)
    }
    val destinations = targetReplicas - sources
    partMoves.put(part.partition, PartitionMove(sources, destinations))
}
```

Check that over and see if it makes sense to you. Basically I think the current logic is unnecessarily complicated and probably wrong in multiple ways.

dengziming: Thank you @hachikuji , I also spend much time making sense of this logic since it's unnecessarily complicated and I tried to solve the issue with the least code change, which made it even much more untidy. Now I have rewritten the code according to your suggestions and have simplified it tremendously, PTAL.

core/src/main/scala/kafka/admin/ReassignPartitionsCommand.scala
```diff
@@ -1229,19 +1229,24 @@ object ReassignPartitionsCommand extends Logging {
...
-        move.destinations += replica
-      }
-      reassignment.addingReplicas.forEach(move.destinations += _)
-      reassignment.removingReplicas.forEach(move.destinations -= _)
       val partMoves = moveMap.getOrElseUpdate(part.topic, new mutable.HashMap[Int, PartitionMove])
-      partMoves.put(part.partition, move)
+
+      // The addingReplicas is included in the replicas during reassignment
+      val sources = mutable.Set[Int]() ++ reassignment.replicas().asScala.map(Int.unbox)
+        .diff(reassignment.addingReplicas.asScala.map(Int.unbox))
```
hachikuji: nit: not a big deal for a tool, but we may as well extract a variable for `reassignment.addingReplicas.asScala.map(Int.unbox)` since we use it twice

hachikuji (commented):
Left a minor nitpick, but looks good otherwise. Only question I had is whether you think current test coverage is good enough or if there are gaps we should address. For example, should we have a case where we override the reassignment with the same reassignment that is in progress?

dengziming:
@hachikuji , Thank you for your reminder, currently, the method only test the basic logic, maybe it's uneasy to cover all edge cases of reassignment. I added a few cases considering the relationship between old and new reassignments, you can also give some suggestions.

hachikuji (approved):
LGTM. Thanks for the fix!

See here for the latest build: https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-9807/7/

PR #9808, opened 2021-01-01
MINOR: rename @returns to @return

dengziming:
As the title


chia7712 (approved):

PR #9023, opened 2020-07-14
KAFKA-10272: Add IBM i support to "stop" scripts

ThePrez:
Details are documented in jira issue [KAFKA-10272](https://issues.apache.org/jira/projects/KAFKA/issues/KAFKA-10272).

Turns out a similar issue was fixed for IBM z/OS earlier this year, as part of PR #7913.

### Committer Checklist (excluded from commit message)
- [x] Verify design and implementation 
- [x] ~~Verify test coverage and CI build status~~  N/A
- [x] ~~Verify documentation (including upgrade notes)~~ N/A

PR #9810, opened 2021-01-01
MINOR: Tweak IBM i platform support in "stop" scripts

ThePrez:
IBM i support to the "stop" scripts was added in a previous PR (https://github.com/apache/kafka/pull/9023), but it misses cases where certain service managers dispatch the zookeeper/kafka jobs with "nohup" or system batch job capabilities. We need a tweak to the `ps` invocation for this to work properly in those scenarios. 

Due to the minor nature of the problem and the trivial nature of the fix, I did not create a JIRA issue, but I will happily do so if that's the recommended best route. 

- [x] ~~Verify test coverage and CI build status~~ (self tested)
- [x] ~~Verify documentation (including upgrade notes)~~ n/a


ThePrez:
cc @earleye2018 for awareness only

mimaison (approved):
Thanks for the PR

PR #9811, opened 2021-01-01
MINOR: rename produce to produces 

wenbingshen:
Fixed a syntax error where the predicate needed to use the singular.


abbccdda (approved):
LGTM

PR #9813, opened 2021-01-02
MINOR: gradle wrapper should handle directories with spaces

gwenshap:
If attempting to build the project from a directory with spaces in its name and gradle-wrapper.jar is missing, the script will fail to download a new one because the "if" condition will break because $APP_HOME will resolve to multiple strings. Protecting $APP_HOME with quotes fixes the issue. Tested by deleting gradle.wrapper and trying to build.

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -83,7 +83,7 @@ esac
 
 # Loop in case we encounter an error.
 for attempt in 1 2 3; do
-  if [ ! -e $APP_HOME/gradle/wrapper/gradle-wrapper.jar ]; then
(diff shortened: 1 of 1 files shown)


chia7712 (approved):
@gwenshap Thanks for your patch. LGTM

BTW, ```run_tests.sh``` has similar issue.

ijuma:
Hmm, we updated an auto-generated file instead of the code that generates the file. I attempted to fix the latter here https://github.com/apache/kafka/pull/9849. @gwenshap please take a look and see if that fixes your issue too.

PR #9814, opened 2021-01-02
KAFKA-10869: Gate topic IDs behind IBP 2.8

jolshan:
Topics processed by the controller and topics newly created will only be given topic IDs if the inter-broker protocol version on the controller is greater than 2.8. This PR also adds a kafka config to specify whether the IBP is greater or equal to 2.8. System tests have been modified to include topic ID checks for upgrade/downgrade tests.

This PR also adds a new integration test file for requests/responses that are not gated by IBP (ex: metadata) This file can be expanded to check topic IDs are handled correctly when the request/response version uses topic IDs but the brokers are not able to handle them.


core/src/main/scala/kafka/zk/AdminZkClient.scala
```diff
@@ -153,16 +157,17 @@ class AdminZkClient(zkClient: KafkaZkClient) extends Logging {
     LogConfig.validate(config)
   }
 
-  private def writeTopicPartitionAssignment(topic: String, replicaAssignment: Map[Int, ReplicaAssignment], isUpdate: Boolean): Unit = {
+  private def writeTopicPartitionAssignment(topic: String, replicaAssignment: Map[Int, ReplicaAssignment],
+                                            isUpdate: Boolean, usesTopicId: Boolean = false): Unit = {
```
dengziming, line 161: Why do you set the default value to false? I don't think we should give it a default value since the caller can decide it by caller.

core/src/main/scala/kafka/zk/AdminZkClient.scala
```diff
@@ -90,11 +92,13 @@ class AdminZkClient(zkClient: KafkaZkClient) extends Logging {
...
    * @param partitionReplicaAssignment The assignments of the topic
    * @param validate Boolean indicating if parameters must be validated or not (true by default)
+   * @param usesTopicId Boolean indicating whether the topic ID will be created
    */
   def createTopicWithAssignment(topic: String,
                                 config: Properties,
                                 partitionReplicaAssignment: Map[Int, Seq[Int]],
-                                validate: Boolean = true): Unit = {
+                                validate: Boolean = true,
+                                usesTopicId: Boolean = false): Unit = {
```
dengziming, line 101: So if we use `TopicCommand` with `ZookeeperTopicService` to create a topic, we will always not set a topicId?

core/src/main/scala/kafka/zk/AdminZkClient.scala
jolshan: Some places that eventually call this do not have access to IBP. The controller will eventually pick up and assign topic IDs that need them.

core/src/main/scala/kafka/zk/AdminZkClient.scala
jolshan: Initially yes. The controller will process and add topic IDs to these topics if IBP is at least 2.8 in processTopicChange() using the processTopicIds method.

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
```diff
@@ -884,6 +884,37 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
...
+    val controller = getController().kafkaController
+    val tp1 = new TopicPartition("t1", 0)
+    val assignment1 = Map(tp1.partition -> Seq(0))
+
+    // Before adding the topic, an attempt to get the ID should result in None.
+    assertEquals(None, controller.controllerContext.topicIds.get("t1"))
+
+    TestUtils.createTopic(zkClient, tp1.topic(), assignment1, servers)
+
+    // Test that the first topic has its ID added correctly
```
rajinisivaram: Comment should say: no topic ID (or just remove the comment)?

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
```diff
@@ -884,6 +884,37 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
...
+    waitForPartitionState(tp1, firstControllerEpoch, 0, LeaderAndIsr.initialLeaderEpoch,
+      "failed to get expected partition state upon topic creation")
+    assertEquals(None, controller.controllerContext.topicIds.get("t1"))
+
+    val tp2 = new TopicPartition("t2", 0)
+    val assignment2 = Map(tp2.partition -> Seq(0))
+    TestUtils.createTopic(zkClient, tp2.topic(), assignment2, servers)
+
+    // Test that the second topic has its ID added correctly
```
rajinisivaram: As above, update or remove comment?

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
```diff
@@ -884,6 +884,37 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
...
+    waitForPartitionState(tp2, firstControllerEpoch, 0, LeaderAndIsr.initialLeaderEpoch,
+      "failed to get expected partition state upon topic creation")
+    assertEquals(controller.controllerContext.topicIds.get("t2"), None)
```
rajinisivaram: Make None the first argument?

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
```diff
@@ -911,6 +942,32 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
...
+    val tp = new TopicPartition("t", 0)
+    val assignment = Map(tp.partition -> ReplicaAssignment(Seq(0), List(), List()))
+    val adminZkClient = new AdminZkClient(zkClient)
+
+    servers = makeServers(1, interBrokerProtocolVersion = Some(KAFKA_2_7_IV0))
+    adminZkClient.createTopic(tp.topic, 1, 1)
+    waitForPartitionState(tp, firstControllerEpoch, 0, LeaderAndIsr.initialLeaderEpoch,
+      "failed to get expected partition state upon topic creation")
+    val topicIdAfterCreate = zkClient.getTopicIdsForTopics(Set(tp.topic())).get(tp.topic())
+    assertTrue(topicIdAfterCreate.isEmpty)
```
rajinisivaram: assertEquals(None, ...)?

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
```diff
@@ -950,6 +1027,33 @@ class ControllerIntegrationTest extends ZooKeeperTestHarness {
...
+    assertEquals(None, emptyTopicId)
+
+    servers(controllerId).shutdown()
+    servers(controllerId).awaitShutdown()
+    servers = makeServers(1)
+    TestUtils.waitUntilTrue(() => zkClient.getControllerId.isDefined, "failed to elect a controller")
+    val controller2 = getController().kafkaController
+    assertNotEquals(emptyTopicId, controller2.controllerContext.topicIds.get("t"))
+    val topicId = controller2.controllerContext.topicIds.get("t").get
+    assertEquals("t", controller2.controllerContext.topicNames(topicId))
```
rajinisivaram, line 1056: Should we also check the topic id in ZooKeeper?

tests/kafkatest/version.py
```diff
@@ -185,4 +189,10 @@ def get_version(node=None):
 
 # 2.7.x versions
 V_2_7_0 = KafkaVersion("2.7.0")
-LATEST_2_7 = V_2_7_0
+V_2_7_1 = KafkaVersion("2.7.1")
+V_2_7_2 = KafkaVersion("2.7.2")
+LATEST_2_7 = V_2_7_2
```
rajinisivaram: We are currently on 2.7.0, so don't think we need 2.7.1 and 2.7.2

rajinisivaram (commented):
@jolshan Thanks for the PR, looks good. Left some minor comments.

tests/kafkatest/version.py
jolshan: Oops. I thought these were related to IBPs. Will fix that.

core/src/test/scala/unit/kafka/controller/ControllerIntegrationTest.scala
jolshan: Good catch.

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1767,6 +1767,10 @@ class KafkaConfig(val props: java.util.Map[_, _], doLog: Boolean, dynamicConfigO
     }
   }
 
+  def usesTopicId: Boolean = {
```
lbradstreet: nit: the braces are unnecessary here

rajinisivaram (approved):
@jolshan Thanks for the updates, LGTM. Can you confirm that the upgrade and downgrade system tests passed with the PR before I merge?

rajinisivaram:
@jolshan Thanks for running the system tests, merging to trunk.

PR #9815, opened 2021-01-02
MINOR: Fixed two properties in EOS example plus some thread cleanup

gwenshap:
Changes:
* Passing the latches into the consumer/producer wrappers seemed a bit unclean. I removed the latches and used futures to wait on producer/consumer execution.
* Consumer that is part of the transaction should not auto-commit
* Transactional producers don't need to set idempotency explicitly.

### DIFF
--- a/examples/src/main/java/kafka/examples/Consumer.java
+++ b/examples/src/main/java/kafka/examples/Consumer.java
@@ -16,7 +16,6 @@
  */
 package kafka.examples;
 
-import kafka.utils.ShutdownableThread;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
@@ -26,7 +25,6 @@ import java.time.Duration;
 import java.util.Collections;
 import java.util.Optional;
 import java.util.Properties;
-import java.util.concurrent.CountDownLatch;
 
 public class Consumer implements Runnable {
     private final KafkaConsumer<Integer, String> consumer;
--- a/examples/src/main/java/kafka/examples/Consumer.java
+++ b/examples/src/main/java/kafka/examples/Consumer.java
@@ -28,21 +28,18 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.CountDownLatch;
 
-public class Consumer extends ShutdownableThread {
+public class Consumer implements Runnable {
     private final KafkaConsumer<Integer, String> consumer;
     private final String topic;
     private final String groupId;
     private final int numMessageToConsume;
     private int messageRemaining;
-    private final CountDownLatch latch;
 
     public Consumer(final String topic,
                     final String groupId,
                     final Optional<String> instanceId,
                     final boolean readCommitted,
-                    final int numMessageToConsume,
-                    final CountDownLatch latch) {
-        super("KafkaConsumerExample", false);
+                    final int numMessageToConsume) {
         this.groupId = groupId;
         Properties props = new Properties();
         props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, KafkaProperties.KAFKA_SERVER_URL + ":" + KafkaProperties.KAFKA_SERVER_PORT);
@@ -62,7 +59,6 @@ public class Consumer extends ShutdownableThread {
         this.topic = topic;
         this.numMessageToConsume = numMessageToConsume;
         this.messageRemaining = numMessageToConsume;
-        this.latch = latch;
     }
 
     KafkaConsumer<Integer, String> get() {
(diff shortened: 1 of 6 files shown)


examples/src/main/java/kafka/examples/Consumer.java
```diff
@@ -26,33 +25,34 @@
...
         Properties props = new Properties();
         props.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, KafkaProperties.KAFKA_SERVER_URL + ":" + KafkaProperties.KAFKA_SERVER_PORT);
         props.put(ConsumerConfig.GROUP_ID_CONFIG, groupId);
         instanceId.ifPresent(id -> props.put(ConsumerConfig.GROUP_INSTANCE_ID_CONFIG, id));
-        props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "true");
-        props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, "1000");
         props.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, "30000");
         props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.IntegerDeserializer");
         props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.StringDeserializer");
+        // if consuming as part of exactly-once processor, committing will be done by the producer
```
abbccdda, line 50: I'm not sure this is required here, in EOS consumer auto commit will not be successful right?

examples/src/main/java/kafka/examples/Consumer.java
```diff
@@ -62,34 +62,24 @@ public Consumer(final String topic,
...
     }
 
     @Override
-    public void doWork() {
+    public void run() {
         consumer.subscribe(Collections.singletonList(this.topic));
-        ConsumerRecords<Integer, String> records = consumer.poll(Duration.ofSeconds(1));
-        for (ConsumerRecord<Integer, String> record : records) {
-            System.out.println(groupId + " received message : from partition " + record.partition() + ", (" + record.key() + ", " + record.value() + ") at offset " + record.offset());
+        while (!Thread.currentThread().isInterrupted() &&  messageRemaining > 0) {
```
abbccdda, line 74: So previously this example only runs once?

examples/src/main/java/kafka/examples/KafkaExactlyOnceDemo.java
```diff
@@ -88,40 +91,50 @@ public static void main(String[] args) throws InterruptedException, ExecutionExc
...
+            Exit.exit(1);
         }
 
-        CountDownLatch consumeLatch = new CountDownLatch(1);
-
-        /* Stage 4: consume all processed messages to verify exactly once */
-        Consumer consumerThread = new Consumer(OUTPUT_TOPIC, "Verify-consumer", Optional.empty(), true, numRecords, consumeLatch);
-        consumerThread.start();
+        /* Stage 4: consume all processed messages to verify exactly once.
+        Consumer uses read committed to guarantee that uncommitted events will not be included in verification
```
abbccdda, line 124: s/Consumer/The verifying consumer

examples/src/main/java/kafka/examples/KafkaExactlyOnceDemo.java
```diff
@@ -88,40 +91,50 @@ public static void main(String[] args) throws InterruptedException, ExecutionExc
...
+        but the consumer is not part of the transaction itself
```
abbccdda, line 125: while the verifying consumer itself is not part of the transaction

abbccdda (commented):
Overall looks good, just some minor comments

examples/src/main/java/kafka/examples/Consumer.java
gwenshap: @abbccdda thank you! Is this true since KIP-447? or even earlier?

examples/src/main/java/kafka/examples/Consumer.java
gwenshap: I believe so...

examples/src/main/java/kafka/examples/Consumer.java
gwenshap: I am having trouble finding the code that prevents EOS consumer from auto-committing. Pointers will help :)

hachikuji: The change makes sense to me. I don't think anything would stop the auto-commits from going through. Even if there was such a mechanism, it seems better to explicitly disable it.

examples/src/main/java/kafka/examples/Consumer.java
```diff
@@ -26,33 +25,34 @@
...
-        props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "true");
-        props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, "1000");
         props.put(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, "30000");
         props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.IntegerDeserializer");
         props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, "org.apache.kafka.common.serialization.StringDeserializer");
+        // if consuming as part of exactly-once processor, committing will be done by the producer
+        if (transactional) {
+            props.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false");
+        } else {
+            props.put(ConsumerConfig.AUTO_COMMIT_INTERVAL_MS_CONFIG, "1000");
```
hachikuji, line 54: nit: since we are setting auto commit interval, perhaps we should set enable auto commit explicitly rather than rely on the default

examples/src/main/java/kafka/examples/Consumer.java
hachikuji: `doWork` is just one iteration. `ShutdownableThread` has the loop. I'm ok with the change, but we probably will need to copy over some of the shutdown logic.

PR #9816, opened 2021-01-03
KAFKA-10761: Kafka Raft update log start offset

jsancio:
Adds support for nonzero log start offsets.

Changes to `Log`:
1. Add a new "reason" for increasing the log start offset. This is used by `KafkaMetadataLog` when a snapshot is generated.
2. `LogAppendInfo` should return if it was rolled because of an records append. A log is rolled when a new segment is created. This is used by `KafkaMetadataLog` to in some cases delete the created segment based on the log start offset.

Changes to `KafkaMetadataLog`:
1. Update both append functions to delete old segments based on the log start offset whenever the log is rolled.
2. Update `lastFetchedEpoch` to return the epoch of the latest snapshot whenever the log is empty.
3. Add a function that empties the log whenever the latest snapshot is greater than the replicated log. This is used when first loading the `KafkaMetadataLog` and whenever the `KafkaRaftClient` downloads a snapshot from the leader.

Changes to `KafkaRaftClient`:
1. Improve `validateFetchOffsetAndEpoch` so that it can handle fetch offset and last fetched epoch that are smaller than the log start offset. This is in addition to the existing code that check for a diverging log. This is used by the raft client to determine if the Fetch response should include a diverging epoch or a snapshot id. 
2. When a follower finishes fetching a snapshot from the leader fully truncate the local log.
3. When polling the current state the raft client should check if the state machine has generated a new snapshot and update the log start offset accordingly.

comment to ping reviewers.*


core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -113,6 +140,20 @@ class KafkaMetadataLog(
     log.truncateTo(offset)
   }
 
+  override def truncateFullyToLatestSnapshot(): Boolean = {
```
jsancio:
```
	at kafka.log.Log.updateHighWatermarkMetadata(Log.scala:463)
	at kafka.log.Log.updateHighWatermark(Log.scala:392)
	at kafka.log.Log.updateLogStartOffset(Log.scala:817)
	at kafka.log.Log.$anonfun$truncateFullyAndStartAt$2(Log.scala:2133)
	at kafka.log.Log.truncateFullyAndStartAt(Log.scala:2362)
	at kafka.raft.KafkaMetadataLog.$anonfun$truncateFullyToLatestSnapshot$1(KafkaMetadataLog.scala:149)
	at java.base/java.util.Optional.ifPresent(Optional.java:183)
	at kafka.raft.KafkaMetadataLog.truncateFullyToLatestSnapshot(KafkaMetadataLog.scala:146)
```

Note that truncate fully will update the high watermark when this function returns true. The current signature doesn't have a way to communicate this back to the `KafkaRaftClient`. I think we have two choices:

1. Add a `highWatermark` method to the `ReplicatedLog` interface which the `KafkaRaftClient` will call after this function returns `true`.
2. Change this signature to something like `truncateFullyToLatestSnapshot(): Optional[Long]` where it will return `Optional.of(hightWatermark)` instead of `true`.

I am leaning toward option 1. What do you think @hachikuji ?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +188,100 @@ class KafkaMetadataLog(
...
-      Optional.of(FileRawSnapshotReader.open(log.dir.toPath, snapshotId))
+      if (snapshotIds.contains(snapshotId)) {
+        Optional.of(FileRawSnapshotReader.open(log.dir.toPath, snapshotId))
+      } else {
+        Optional.empty()
+      }
     } catch {
-      case e: NoSuchFileException => Optional.empty()
+      case e: NoSuchFileException =>
+        snapshotIds.remove(snapshotId)
```
jsancio: Remove this because of concurrency issues. This should not happened unless there is a bug in the code or the snapshot was deleted outside of the Kafka process.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
hachikuji: The fact that we have the high watermark in two places has always been unsatisfying, but I have not prioritized coming up with a solution. I have been inclined to say that `ReplicatedLog` owns the high watermark, so option 1 is consistent with that. I might be tempted to leave that for a separate PR though. In this case, since we are requesting a full truncation, it seems we could just set the high watermark to the snapshot offset?

jsancio: > The fact that we have the high watermark in two places has always been unsatisfying, but I have not prioritized coming up with a solution.

Yeah. I thought about it a bit when thinking about this PR. High watermark is both a property of the local log. For example, locally the high watermark cannot be less than the log start offset or greater than the log end offset.

But it is also a replicated and consensus value based on the log end offset of all the voters. For example it is technically possible for a leader to send a Fetch response to a voter or observer where the high watermark is greater than the follower's log end offset.

I agree that it would be interesting to simplify some of the code by not tracking the high watermark in two places. I'll file an issue for this.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
```diff
@@ -175,6 +180,19 @@ default OptionalLong truncateToEndOffset(OffsetAndEpoch endOffset) {
...
 
-    default void close() {}
+    // TODO: write documentation
+    Optional<OffsetAndEpoch> latestSnapshotId();
+
+    /**
+     * TODO: document this method
+     * TODO document that this method is thread safe
+     */
+    void snapshotFrozen(OffsetAndEpoch snapshotId);
```
jsancio: This is an implementation detail. We should be able to remove this. The problem is that the `RawSnapshotWriter` implementation lives in the `raft` module while the `ReplicatedLog` implementation lives in the `core` module. To remove this we need to move `FileRawSnapshotWriter` implementation to `core`.

cc @hachikuji

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -2015,6 +2109,11 @@ private long pollUnattachedAsObserver(UnattachedState state, long currentTimeMs)
     }
 
     private long pollCurrentState(long currentTimeMs) throws IOException {
+        // TODO: File a jira for this: Okay for now but the logic should be a little different.
+        // Leader should only advance the log start offset if all of followers fetched past it or there is a timeout
+        // While followers should only call maybeUpdateLogStartOffset if the leader's logStartOffset is greater than ours
+        log.maybeUpdateLogStartOffset();
```
jsancio: This implementation is okay for now but this simple implementation should be improved by https://issues.apache.org/jira/browse/KAFKA-12155

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -2247,7 +2346,18 @@ public synchronized OptionalLong nextExpectedOffset() {
...
-                return OptionalLong.of(lastAckedOffset);
+                return OptionalLong.of(lastAckedEndOffset);
+            }
+        }
+
+        /**
+         * This API is used when the Listener needs to be notified of a new Snapshot. This happens
+         * when the context last acked end offset is less that then log start offset.
+         */
+        public void fireHandleSnapshot(long logStartOffset) {
```
jsancio, line 2310: This is a temporary signature. This will change when we implement the API for loading snapshots. Tracking is work in this issue: https://issues.apache.org/jira/browse/KAFKA-12154

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -272,26 +272,29 @@ private void onUpdateLeaderHighWatermark(
...
     }
 
-    private void maybeFireHandleCommit(long highWatermark) {
-        maybeFireHandleCommit(listenerContexts, highWatermark);
+    private void onHighWatermarkUpdated(long highWatermark) {
+        onHighWatermarkUpdated(listenerContexts, highWatermark);
     }
 
-    private void maybeFireHandleCommit(List<ListenerContext> listenerContexts, long highWatermark) {
+    private void onHighWatermarkUpdated(List<ListenerContext> listenerContexts, long highWatermark) {
```
hachikuji: This name seems misleading. It can be called from `pollListeners` when a listener is catching up which does not require the high watermark to be updated. Perhaps we could call this `updateListenerProgress` or something like that instead.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -2216,7 +2316,7 @@ public void complete() {
         // These fields are visible to both the Raft IO thread and the listener
         // and are protected through synchronization on this `ListenerContext` instance
         private BatchReader<T> lastSent = null;
-        private long lastAckedOffset = 0;
+        private long lastAckedEndOffset = 0;
```
hachikuji: Can you clarify the reason for the name change?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1326,6 +1408,17 @@ private boolean handleFetchSnapshotResponse(
             // Finished fetching the snapshot.
             snapshot.freeze();
             state.setFetchingSnapshot(Optional.empty());
+
+            if (log.truncateFullyToLatestSnapshot()) {
+                updateFollowerHighWatermark(state, OptionalLong.of(log.highWatermark()));
+            } else {
+                logger.error(
+                    "Full log trunctation expected but didn't happend. Snapshot of {}, log end offset {}, last fetched {}",
```
hachikuji: nit: `happend`

What happens next after this error? Will the node resume fetching? I'm wondering if there are valid situations this could happen, or if we should treat it as a fatal error.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -970,40 +996,98 @@ private FetchResponseData tryCompleteFetchRequest(
...
-            return Optional.empty();
+            // TODO: Investiage this. Can the diverging offset be less than log start offset? If so, then we might as well
+            // avoid a round trip and return the snapshot id instead.
+            return ValidatedFetchOffsetAndEpoch.diverging(endOffsetAndEpoch);
+        }
+
+        return ValidatedFetchOffsetAndEpoch.valid(new OffsetAndEpoch(fetchOffset, lastFetchedEpoch));
+    }
+
+    private final static class ValidatedFetchOffsetAndEpoch {
```
hachikuji: nit: since this class is getting large, perhaps we can pull this out? Or maybe at least move it to the end so it's out of the way.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -29,13 +32,15 @@ import org.apache.kafka.snapshot.FileRawSnapshotReader
 import org.apache.kafka.snapshot.FileRawSnapshotWriter
 import org.apache.kafka.snapshot.RawSnapshotReader
 import org.apache.kafka.snapshot.RawSnapshotWriter
+import org.apache.kafka.snapshot.Snapshots
 
 import scala.compat.java8.OptionConverters._
 
-class KafkaMetadataLog(
+final class KafkaMetadataLog private (
```
hachikuji, line 45: Does the log recovery logic need to be updated at all for the presence of snapshots?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -914,7 +934,13 @@ private FetchResponseData buildEmptyFetchResponse(
                 Errors.INVALID_REQUEST, Optional.empty()));
         }
 
-        FetchResponseData response = tryCompleteFetchRequest(request.replicaId(), fetchPartition, currentTimeMs);
+        FetchResponseData response;
+        try {
```
hachikuji: nit: maybe we can move this `try/catch` into `tryCompleteFetchRequest` since both callers are doing the same thing

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -970,40 +996,98 @@ private FetchResponseData tryCompleteFetchRequest(
...
+                logger.warn(
+                    "Replica sent a zero fetch offset ({}) but the last fetched epoch ({}) was not zero",
+                    fetchOffset,
+                    lastFetchedEpoch
+                );
+            }
+            return ValidatedFetchOffsetAndEpoch.valid(new OffsetAndEpoch(fetchOffset, lastFetchedEpoch));
+        }
+
+        if (fetchOffset < log.startOffset() || fetchOffset == log.startOffset()) {
```
hachikuji: Isn't this just `fetchOffset <= log.startOffset()`?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -970,40 +996,98 @@ private FetchResponseData tryCompleteFetchRequest(
...
+            OffsetAndEpoch latestSnapshotId = log.latestSnapshotId().orElseThrow(() -> {
+                return new IllegalStateException(
+                    String.format(
+                        "The log start offset (%s) was greater than zero but no snapshot was found",
+                        log.startOffset()
+                    )
+                );
+            });
+
+            if (fetchOffset < log.startOffset() || lastFetchedEpoch != latestSnapshotId.epoch) {
```
hachikuji: So if the fetch offset matches the start offset, but we have a more recent snapshot, then the second check is likely to fail. Is the intent to have the replica fetch the latest snapshot rather than catch up from the log?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -970,40 +996,98 @@ private FetchResponseData tryCompleteFetchRequest(
...
+            }
         }
 
         OffsetAndEpoch endOffsetAndEpoch = log.endOffsetForEpoch(lastFetchedEpoch)
             .orElse(new OffsetAndEpoch(-1L, -1));
         if (endOffsetAndEpoch.epoch != lastFetchedEpoch || endOffsetAndEpoch.offset < fetchOffset) {
-            return Optional.of(endOffsetAndEpoch);
-        } else {
-            return Optional.empty();
+            // TODO: Investiage this. Can the diverging offset be less than log start offset? If so, then we might as well
```
hachikuji: If `lastFetchedEpoch` is lower than the epoch of the first data in the log, then I believe `endOffsetForEpoch` will return `lastFetchedEpoch` with the end offset set to the current log start offset. In this case, I think we probably would prefer to return the latest snapshot.

You can let me know if it makes sense, but I think the way I'd try to write the order of the checks in this method is the following:

1. First call `endOffsetForEpoch` with `lastFetchedEpoch`
2. If the end offset is undefined, then we need to read from the latest snapshot.
3. If the end offset is defined and is equal to the start offset, then check the epoch of the first snapshot. If they do not match, then we need to read from latest snapshot
4. Next check `endOffsetAndEpoch.epoch != lastFetchedEpoch || endOffsetAndEpoch.offset < fetchOffset` as we do today. If this is true, then we return the diverging epoch.
5. We have a valid fetch.

hachikuji (commented):
Thanks, left a few initial comments.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: This PR changes the recovery logic so that the log is `truncateFullyToLatestSnapshot()` (https://github.com/apache/kafka/pull/9816/files#diff-b332f85b04775c821226b6f704e91d51f9647f29ba73dace65b99cf36f6b9ceaR298) if the `latestSnapshotId` is greater than `new OffsetAndEpoch(endOffset().offset, log.lastFetchedEpoch)`.

This can happen when a follower crashed after downloading a snapshot from the leader but before calling `truncateFullyToLatestSnapshot()`. Or https://github.com/apache/kafka/pull/9816/files#diff-b332f85b04775c821226b6f704e91d51f9647f29ba73dace65b99cf36f6b9ceaR155

I think this is correct because `KafkaRaftClient` guarantees that any offset less then high watermark has been flushed and after https://issues.apache.org/jira/browse/KAFKA-10800 any snapshot generated by the state machine will be less than or equal to the high-watermark.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Good question. I am not sure. I think, since we pause regular fetch handling, if this happens then we have:

1. A bug in the Leader or the Follower:
  a. There is a bug in the leader if the returned snapshot id is less that the last fetched epoch and log end offset. 
  b. There is a bug in the follower if it sent a fetch offset and last fetched offset that doesn't match the local log.
2. The leader lost data at the end of the log.

What do you think? I am thinking we should throw an fatal exception like `IllegalStateException`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: I am always confused if an given offset is inclusive or exclusive. I was thinking that by adding "end offset" it would make it clear that the offset is exclusive. I can write a comment instead if you prefer.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Yeah. I think this code was wrong. I changed the `ReplicatedLog` interface to have two snapshots `latestSnapshotId` and `startSnapshotId`. We need this for a couple of reasons. One of the reasons is that `latestSnapshotId` is changed concurrently by the state machine. While `startSnapshotId` is changed by the polling thread.

Another reason is that increasing the log start offset will be delayed based on what is describe in https://issues.apache.org/jira/browse/KAFKA-12155.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Thanks for the suggestions. I implemented something slightly different to better match the types. When reading the code it is important to keep in mind that when `logStartOffset` is zero then `startSnapshotId` is `Optional.empty()` otherwise
`logStartOffset == startSnapshotId.offset <= latestSnashotId.offset <= highWatermark <= logEndOffset`.

When reading this code if anything is not clear please let me know and I'll document it in the code.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Yes. This code has changed a bit. Please take a look again.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Updated the code to throw an exception instead of logging an error.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Done. Moved it to `internals` package.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1314,11 +1318,24 @@ class Log(@volatile private var _dir: File,
 
   /**
    * Increment the log start offset if the provided offset is larger.
+   *
+   * If the log start offset changed, then this method:
+   *
+   * 1. Records the new log start offset.
+   * 2. Updates the high watermark if it is less than the new log start offset
```
hachikuji: nit: Some additional documentation may be helpful here, but just describing the logic in the code does not seem too useful. Comments like these tend to get out of sync with the code over time.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -29,15 +32,22 @@ import org.apache.kafka.snapshot.FileRawSnapshotReader
...
   log: Log,
+  snapshotIds: ConcurrentSkipListSet[raft.OffsetAndEpoch],
```
hachikuji: It would be useful to add some comments about concurrency somewhere. I assume we are using a concurrent collection because we want to allow snapshots to be created outside of the raft IO thread. Are there any other reasons?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -106,6 +106,7 @@ case class LogAppendInfo(var firstOffset: Option[Long],
                          var logAppendTime: Long,
                          var logStartOffset: Long,
                          var recordConversionStats: RecordConversionStats,
+                         var rolled: Boolean,
```
hachikuji: Painful to me to have another parameter here. I wonder if we could make the type of `firstOffset` be `Option[LogOffsetMetadata]`. Then we could infer `rolled` from the segment position. If we keep the extra parameter, we need to update the doc above.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -29,15 +32,22 @@ import org.apache.kafka.snapshot.FileRawSnapshotReader
...
   topicPartition: TopicPartition,
-  maxFetchSizeInBytes: Int = 1024 * 1024
+  maxFetchSizeInBytes: Int
 ) extends ReplicatedLog {
 
+  private[this] var startSnapshotId = snapshotIds
```
hachikuji: nit: isn't `[this]` redundant?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -79,13 +94,30 @@ class KafkaMetadataLog(
...
+        val logEndOffset = endOffset().offset
+        if (snapshotId.offset == logEndOffset) {
+          snapshotId.epoch
+        } else {
+          throw new KafkaException(
+            s"Log doesn't have a last fetch epoch and there is a snapshot ($snapshotId). " +
+            s"Expected the snapshot's end offset to match the logs end offset ($logEndOffset)"
+          )
+        }
+      } orElse(0)
```
hachikuji: nit: it's a little unusual to leave off the `.` in an expression like this

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -79,13 +94,30 @@ class KafkaMetadataLog(
...
       throw new KafkaException("Append failed unexpectedly")
     }, appendInfo.lastOffset)
   }
 
   override def lastFetchedEpoch: Int = {
-    log.latestEpoch.getOrElse(0)
+    log.latestEpoch.getOrElse {
+      latestSnapshotId.map { snapshotId =>
+        val logEndOffset = endOffset().offset
+        if (snapshotId.offset == logEndOffset) {
```
hachikuji: I think the invariant we are trying to enforce is that we always have a snapshot at the log start offset, so it's a little surprising to see the check for the end offset here. I think this is handling the case when the log is empty, so could we just as well use the log start offset?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -113,6 +145,22 @@ class KafkaMetadataLog(
     log.truncateTo(offset)
   }
 
+  override def truncateFullyToLatestSnapshot(): Boolean = {
+    // Truncate the log fully if the latest snapshot is greater than the log end offset
+    var truncated = false
```
hachikuji: nit: we could probably convert to scala option and rewrite the check below as `exists`

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -113,6 +145,22 @@ class KafkaMetadataLog(
...
+    latestSnapshotId.ifPresent { snapshotId =>
+      if (snapshotId.epoch > log.latestEpoch.getOrElse(0) ||
```
hachikuji: nit: use `val` for `log.latestEpoch.getOrElse(0)` to avoid duplication?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +199,106 @@ class KafkaMetadataLog(
...
+    try {
+      Optional.of(snapshotIds.last)
+    } catch {
+      case _: NoSuchElementException =>
+        Optional.empty()
+    }
+  }
+
+  override def startSnapshotId(): Optional[raft.OffsetAndEpoch] = {
+    startSnapshotId;
```
hachikuji: nit: remove semicolon

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -113,6 +145,22 @@ class KafkaMetadataLog(
...
   }
 
+  override def truncateFullyToLatestSnapshot(): Boolean = {
+    // Truncate the log fully if the latest snapshot is greater than the log end offset
+    var truncated = false
+    latestSnapshotId.ifPresent { snapshotId =>
+      if (snapshotId.epoch > log.latestEpoch.getOrElse(0) ||
+        (snapshotId.epoch == log.latestEpoch.getOrElse(0) && snapshotId.offset > endOffset().offset)) {
+
+        log.truncateFullyAndStartAt(snapshotId.offset)
```
hachikuji, line 181: Do we need to delete older snapshots?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -131,6 +179,10 @@ class KafkaMetadataLog(
     }
   }
 
+  override def highWatermark: Long = {
```
hachikuji: It might be useful to expose this as `LogOffsetMetadata`. That would be more consistent with `updateHighWatermark`.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +199,106 @@ class KafkaMetadataLog(
   }
 
   override def createSnapshot(snapshotId: raft.OffsetAndEpoch): RawSnapshotWriter = {
-    FileRawSnapshotWriter.create(log.dir.toPath, snapshotId)
+    // TODO: Talk to Jason about truncation past the high-watermark since it can lead to truncation past snapshots.
```
hachikuji: Can you clarify the question? Do you mean truncation to an offset below the high watermark?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +199,106 @@ class KafkaMetadataLog(
...
+      case e: NoSuchFileException =>
+        Optional.empty()
     }
   }
 
+  override def latestSnapshotId(): Optional[raft.OffsetAndEpoch] = {
+    try {
+      Optional.of(snapshotIds.last)
+    } catch {
+      case _: NoSuchElementException =>
```
hachikuji, line 265: Man, this is annoying. There is also a `pollLast` which returns null if the set is empty.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +199,106 @@ class KafkaMetadataLog(
...
+        Optional.empty()
+    }
+  }
+
+  override def startSnapshotId(): Optional[raft.OffsetAndEpoch] = {
```
hachikuji: nit: how about `oldestSnapshotId` to go along with `latestSnapshotId`?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +199,106 @@ class KafkaMetadataLog(
...
+    startSnapshotId;
+  }
+
+  override def snapshotFrozen(snapshotId: raft.OffsetAndEpoch): Unit = {
```
hachikuji: Maybe `onSnapshotFrozen`?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +199,106 @@ class KafkaMetadataLog(
...
+    snapshotIds.add(snapshotId)
+  }
+
+  override def updateLogStart(logStartSnapshotId: raft.OffsetAndEpoch): Boolean = {
```
hachikuji: Do we need to validate that there really is a snapshot associated with this id?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +199,106 @@ class KafkaMetadataLog(
...
+
+    updated
+  }
+
   override def close(): Unit = {
     log.close()
   }
 }
+
+object KafkaMetadataLog {
```
hachikuji, line 299: Kind of a general note, but I think there are some logging gaps in here. I think we should tend toward the verbose side initially since snapshot events will be relatively infrequent.

hachikuji (commented):
Posted some comments on the logic in `KafkaMetadataLog`. Mostly small stuff.

core/src/main/scala/kafka/log/Log.scala
jsancio: Yeah, I agree. I originally decided to add a comment because it was educational to me when I discovered that this call updated a few other components. I changed the documentation to be a bit more high-level and less likely to change.

core/src/main/scala/kafka/log/Log.scala
jsancio: I implemented this suggestion instead of using `rolled`. We discuss this offline and is it documented in the code. The caveat in the implementation is that the segment's base offset and relative offset in the segment is unknown when the appended records are duplicates.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Done.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Yes. Up to this point we don't delete any snapshot. I was thinking of doing this in a future PR.

https://issues.apache.org/jira/browse/KAFKA-12205

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: If you look at the if statement below this comment...
```scala
if (latest.epoch > snapshotId.epoch || latest.offset > snapshotId.offset) {
```
That naked offset comparison is only valid if the snapshot are always less that the high-watermark and we never truncate past the high-watermark.

I checked this after writing the comment. Out side of data loss we don't allow followers to truncate past the high watermark.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: I agree. Unfortunately, `pollLast` removes the element from the set. I don't think we have a choice but to do this.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Sounds good. In this PR it is not technically the oldest snapshot but it will be when we implement deleting snapshot in https://issues.apache.org/jira/browse/KAFKA-12205

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Yes.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Give me a sec to reply to this.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Done.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: I made this explicit by checking both log start offset and log end offset.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Done.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Hehe. Done.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Done.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Done. Used Scala's `match`.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Done.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -16,28 +16,41 @@
...
 import org.apache.kafka.snapshot.RawSnapshotWriter
+import org.apache.kafka.snapshot.Snapshots
 
 import scala.compat.java8.OptionConverters._
 
-class KafkaMetadataLog(
+final class KafkaMetadataLog private (
   log: Log,
+  // This object needs to be thread-safe because the polling thread in the KafkaRaftClient implementation
+  // and other threads will access this object. This object is used to efficiently notify the polling thread
```
hachikuji: What other threads?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Scala generates different code for `private[this]` vs `private` when used on a field.

Scala visibility model is different from the JVM visibility model. Because of this when using `private` on a field, Scala generates access methods for the field. When using `private[this]` Scala doesn't generate these access methods. For example:

```scala
final class PrivateThisTest {
  private[this] var privateThis = Some(10);
  private var justPrivate = Some(20);
}
```
generates (`javap -p ...`) the following class:
```java
public final class PrivateThisTest {
  private scala.Some<java.lang.Object> privateThis;
  private scala.Some<java.lang.Object> justPrivate;
  private scala.Some<java.lang.Object> justPrivate();
  private void justPrivate_$eq(scala.Some<java.lang.Object>);
  public PrivateThisTest();
}
```

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -370,7 +375,9 @@ class Log(@volatile private var _dir: File,
       throw new KafkaStorageException(s"The memory mapped buffer for log of $topicPartition is already closed")
   }
 
-  def highWatermark: Long = highWatermarkMetadata.messageOffset
+  def highWatermark: Long = _highWatermarkMetadata.messageOffset
+
+  def highWatermarkMetadata: LogOffsetMetadata = _highWatermarkMetadata
```
hachikuji: Could we instead either expose `fetchHighWatermarkMetadata` or make use of `fetchOffsetSnapshot` in `KafkaMetadataLog`?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -69,23 +82,55 @@ class KafkaMetadataLog(
     val appendInfo = log.appendAsLeader(records.asInstanceOf[MemoryRecords],
       leaderEpoch = epoch,
       origin = AppendOrigin.Coordinator)
-    new LogAppendInfo(appendInfo.firstOffset.getOrElse {
-      throw new KafkaException("Append failed unexpectedly")
-    }, appendInfo.lastOffset)
+
+    if (appendInfo.firstOffset.exists(_.relativePositionInSegment == 0)) {
```
hachikuji: Maybe add a little helper? This code looks the same as in `appendAsFollower`.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -69,23 +82,55 @@ class KafkaMetadataLog(
...
+      // Assume that a new segment was created if the relative position is 0
+      log.deleteOldSegments()
+    }
+
+    new LogAppendInfo(
+      appendInfo.firstOffset.map(_.messageOffset).getOrElse {
```
hachikuji: nit: could probably consolidate the `exists` and `map` here into a single `match`

```scala
appendInfo.firstOffset match {
  case None => 
    throw new KafkaException

  case Some(offsetMetadata=>
    if (offsetMetadata.relativePositionInSegment == 0) {
      log.deleteOldSegments()
    }
    appendInfo.lastOffset
}
```

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -69,23 +82,55 @@ class KafkaMetadataLog(
...
+    log.latestEpoch.getOrElse {
+      latestSnapshotId.map { snapshotId =>
+        val logEndOffset = endOffset().offset
+        if (snapshotId.offset == startOffset && snapshotId.offset == logEndOffset) {
+          // Return the epoch of the snapshot when the log is empty
+          snapshotId.epoch
+        } else {
+          throw new KafkaException(
+            s"Log doesn't have a last fetch epoch and there is a snapshot ($snapshotId). " +
+            s"Expected the snapshot's end offset to match the logs end offset ($logEndOffset) " +
```
hachikuji: nit: `log's`

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -69,23 +82,55 @@ class KafkaMetadataLog(
...
-    log.latestEpoch.getOrElse(0)
+    log.latestEpoch.getOrElse {
+      latestSnapshotId.map { snapshotId =>
+        val logEndOffset = endOffset().offset
+        if (snapshotId.offset == startOffset && snapshotId.offset == logEndOffset) {
+          // Return the epoch of the snapshot when the log is empty
+          snapshotId.epoch
+        } else {
+          throw new KafkaException(
+            s"Log doesn't have a last fetch epoch and there is a snapshot ($snapshotId). " +
```
hachikuji, line 128: I guess it's probably better to throw here. I was trying to think how we could end up here, but the only way I came up with is an invalid state transition which left the start or end offset inconsistent with the segment data. Sadly we have had a number of those bugs in the past. I debated whether we should just delete the snapshot, but I'm not sure that helps if we are left with inconsistent start/end offsets.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -113,6 +158,22 @@ class KafkaMetadataLog(
     log.truncateTo(offset)
   }
 
+  override def truncateFullyToLatestSnapshot(): Boolean = {
```
hachikuji: nit: the method name suggests that the truncation occurs unconditionally

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +221,102 @@ class KafkaMetadataLog(
   }
 
   override def createSnapshot(snapshotId: raft.OffsetAndEpoch): RawSnapshotWriter = {
-    FileRawSnapshotWriter.create(log.dir.toPath, snapshotId)
+    // Do not let the state machine create snapshots older than the latest snapshot
+    latestSnapshotId().ifPresent { latest =>
+      if (latest.epoch > snapshotId.epoch || latest.offset > snapshotId.offset) {
+        // Since snapshots are less than the high-watermark absolute offset comparison is okay.
```
hachikuji, line 238: Is it useful here to ensure that `snapshotId` is indeed lower than the high watermark?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +221,102 @@ class KafkaMetadataLog(
...
+
+  override def oldestSnapshotId(): Optional[raft.OffsetAndEpoch] = {
+    oldestSnapshotId
+  }
+
+  override def onSnapshotFrozen(snapshotId: raft.OffsetAndEpoch): Unit = {
+    snapshotIds.add(snapshotId)
+  }
+
+  override def updateLogStart(logStartSnapshotId: raft.OffsetAndEpoch): Boolean = {
```
hachikuji: One thing I was considering is whether we even want to expose this at all. In other words, who should be responsible for cleaning up old snapshots? I am tempted to say that that should be the responsibility of the Log implementation, but perhaps there are reasons it should not be?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -16,28 +16,41 @@
...
+  // This object needs to be thread-safe because the polling thread in the KafkaRaftClient implementation
+  // and other threads will access this object. This object is used to efficiently notify the polling thread
+  // when snapshots are created.
+  snapshotIds: ConcurrentSkipListSet[raft.OffsetAndEpoch],
   topicPartition: TopicPartition,
-  maxFetchSizeInBytes: Int = 1024 * 1024
+  maxFetchSizeInBytes: Int
 ) extends ReplicatedLog {
 
+  private[this] var oldestSnapshotId = snapshotIds
```
hachikuji, line 58: Do we need to keep this as a var or could we access it from `snapshotIds` when needed (as we do for `latestSnapshotId`)?

core/src/test/scala/kafka/raft/KafkaMetadataLogTest.scala
```diff
@@ -0,0 +1,333 @@
...
+      assertEquals(0, snapshot.sizeInBytes())
+    }
+  }
+
+  @Test
+  def testReadMissingSnapshot(): Unit = {
+    val topicPartition = new TopicPartition("cluster-metadata", 0)
+    val log = buildMetadataLog(tempDir, mockTime, topicPartition)
+
+    assertFalse(log.readSnapshot(new OffsetAndEpoch(10, 0)).isPresent)
```
hachikuji: nit: doesn't matter too much for deterministic tests like this, but I think this is a nicer pattern:
```scala
assertEquals(Optional.empty(), log.readSnapshot(new OffsetAndEpoch(10, 0)))
```
Then the assertion message tells you what the value was.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -871,15 +875,32 @@ private FetchResponseData buildFetchResponse(
...
 
     private FetchResponseData buildEmptyFetchResponse(
         Errors error,
         Optional<LogOffsetMetadata> highWatermark
     ) {
-        return buildFetchResponse(error, MemoryRecords.EMPTY, Optional.empty(), highWatermark);
+        return buildFetchResponse(
+            error,
+            MemoryRecords.EMPTY,
+            ValidatedFetchOffsetAndEpoch.valid(new OffsetAndEpoch(-1, -1)),
```
hachikuji: Not sure it's worth creating another type, but it is a little surprising to see `valid(new OffsetAndEpoch(-1, -1))`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -963,49 +979,111 @@ private FetchResponseData tryCompleteFetchRequest(
...
 
-        OffsetAndEpoch endOffsetAndEpoch = log.endOffsetForEpoch(lastFetchedEpoch)
-            .orElse(new OffsetAndEpoch(-1L, -1));
-        if (endOffsetAndEpoch.epoch != lastFetchedEpoch || endOffsetAndEpoch.offset < fetchOffset) {
-            return Optional.of(endOffsetAndEpoch);
+
+        Optional<OffsetAndEpoch> endOffsetAndEpochOpt = log
+            .endOffsetForEpoch(lastFetchedEpoch)
+            .flatMap(endOffsetAndEpoch -> {
+                if (endOffsetAndEpoch.epoch == lastFetchedEpoch && endOffsetAndEpoch.offset == log.startOffset()) {
```
hachikuji: So basically we are trying to detect when the fetch is outside the range of the log, which means we will need to send a snapshot to the follower. Is that right?

One thought we don't have to do here is perhaps we should change the semantics of `endOffsetForEpoch` as exposed by `ReplicatedLog`. Maybe we should only return the end offset of an epoch when we know it with certainty.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -963,49 +979,111 @@ private FetchResponseData tryCompleteFetchRequest(
...
+            OffsetAndEpoch oldestSnapshotId = log.oldestSnapshotId().orElseThrow(() -> {
+                return new IllegalStateException(
+                    String.format(
+                        "The log start offset (%s) was greater than zero but start snapshot was not found",
+                        log.startOffset()
+                    )
+                );
+            });
+
+            if (fetchOffset == log.startOffset() && lastFetchedEpoch == oldestSnapshotId.epoch) {
```
hachikuji: Just checking, but I think we would only hit this case if the snapshot epoch does not match the epoch of the first entry in the log. Is that right? This might be another case that we should try to handle inside `KafkaMetadataLog. endOffsetForEpoch `.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: I don't feel too strongly about it, but the new name is a little confusing to me. Why would the client only be acking end offsets? Especially confusing when I see this: `lastAckedEndOffset = logStartOffset` 🙂 .  I think we probably need a comment here regardless.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
```diff
@@ -106,6 +120,14 @@
      */
     void updateHighWatermark(LogOffsetMetadata offsetMetadata);
 
+    /**
+     * Updates the log start offset if necessary.
+     *
+     * The replicated log's start offset can be increased when there is a snapshot greater than the
+     * current log start offset.
+     */
+    boolean updateLogStart(OffsetAndEpoch logStartSnapshotId);
```
hachikuji: Perhaps we can make this mirror `truncateFullyToLatestSnapshot`. Perhaps `deleteToLatestSnapshot()` or something like that.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientSnapshotTest.java
```diff
@@ -43,7 +44,59 @@
...
+        context.pollUntilResponse();
+        FetchResponseData.FetchablePartitionResponse partitionResponse = context.assertSentFetchResponse();
+        assertEquals(Errors.NONE, Errors.forCode(partitionResponse.errorCode()));
+        assertEquals(epoch, partitionResponse.currentLeader().leaderEpoch());
+        assertEquals(localId, partitionResponse.currentLeader().leaderId());
+        assertEquals(snapshotId.epoch, partitionResponse.snapshotId().epoch());
+        assertEquals(snapshotId.offset, partitionResponse.snapshotId().endOffset());
+    }
+
+    // TODO: Add a few more fetch request tests checking some of the error conditions and edge conditions
```
hachikuji: Reminder here

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Note that the latest implementation changed a bit but this comment should still apply.

Yeah. I think we can't trust the `endOffsetAndEpoch` returned by `kafka.log.Log` when the epoch is smaller than any known epoch in the log. We know this is the case if `endOffsetForEpoch` returns `new OffsetAndEpoch(log.startOffset(), lastFetchedEpoch)`. If `endOffsetForEpoch` returns this value we overwrite it to the `oldestSnapshotId`. In other words if the leader has the following state:

oldestSnapshotId: endOffset = 10, epoch = 2
logStartOffset: 10
leaderEpochCache: (epoch = 4, startOffset = 10), ...

If the follower sends the Fetch request for offset 11 and last fetched offset 3, then the leader should return a diverging epoch of `OffsetAndEpoch(offset=10, epoch=2)`. If we don't do this overwrite the leader would return `OffsetAndEpoch(offset=10, epoch=3)`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Sorry @hachikuji but I changed this code again. I was having a hard time reading my implementation. Yes, I think you are right,we should be able to move this code into `KafkaMetadataLog`. Let me do that.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Fixed.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: I re-wrote the comment to be more helpful.

core/src/main/scala/kafka/log/Log.scala
jsancio: Yeah. Good point. I changed it to use `fetchOffsetSnapshot`.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Thanks for the suggestion. Fixed.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Done.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
jsancio: Done. Named it `deleteToNewOldestSnapshotId`.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientSnapshotTest.java
jsancio: Yep. Replaced the TODO with a few more tests.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: How about just calling it `nextOffset`?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Sorry for the delay reply. I missed this comment for some reason.

At the moment we need to keep this variable/reference. In this PR, the `oldestSnapshotId` can be anywhere in the `snapshotIds` set. I think that when we implement https://issues.apache.org/jira/browse/KAFKA-12205 can remove this variable.

https://issues.apache.org/jira/browse/KAFKA-12205 tracks the work needed to delete any snapshot that is less than the log start offset.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: I don't think we can delete the snapshot if `startOffset` is greater than 0.

I think we need multiple bugs to throw this exception. Like you said the leader epoch cache would need to be inconsistent with respect to the log itself. Or we don't have a snapshot at the log start offset.

I think at this point the best we can do is throw an exception. I think that deleting the snapshots and/or log could result in data loss. Maybe in the future we can do a `truncateFullyAtLastestSnapshot` but I would like to have a concrete reason for doing this.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: I can prefix it with maybe. E.g. `maybeTruncateFullyToLatestSnapshot`.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Yes but only when it is the state machine that is creating the snapshot. When the raft client is creating the snapshot because it is downloading it from the leader we don't want to validate this.

We'll add this check to the `RaftClient::createSnapshot` API when we implement https://issues.apache.org/jira/browse/KAFKA-10800

core/src/test/scala/kafka/raft/KafkaMetadataLogTest.scala
jsancio: Fixed.

I think the reason that I didn't do this was because `RawSnapshotReader` doesn't implement in an interesting equality. It doesn't also override `toString` but probably it should. Let me create a Jira to add `toString` to some of the classes I added.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Yeah. I can hide this by adding an overloaded signature as `ValidatedFetchOffsetAndEpoch valid()`. The Raft Client implementation ignores this value when it is valid when it is creating a Fetch response.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Yeah. I think it depends on the threading model. Right now I am relying on the polling thread to trigger the deletes. Note that the delete still happen async in the scheduler passed to the `kafka.log.Log`.

Having said that, when we implement https://issues.apache.org/jira/browse/KAFKA-12155 I think will have to give the Raft Client event more control as to how the log start offset gets incremented. For example, that Jira proposes looking at the `nextOffset` (acked offset) of the state machines and the `fetch offset` of the follower to increase the log start offset to minimize state machine invalidation.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -1910,7 +1931,7 @@ class Log(@volatile private var _dir: File,
         in the header.
       */
       appendInfo.firstOffset match {
-        case Some(firstOffset) => roll(Some(firstOffset))
+        case Some(firstOffset) => roll(Some(firstOffset.messageOffset))
```
hachikuji: nit: might be able to simplify this a little bit. Maybe something like this:
```scala
val rollOffset = appendInfo.firstOffset.map(_.messageOffset)
  .getOrElse(maxOffsetInMessages - Integer.MAX_VALUE))
roll(Some(rollOffset))
```

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2118,7 +2139,7 @@ class Log(@volatile private var _dir: File,
    *
    *  @param newOffset The new offset to start the log with
    */
-  private[log] def truncateFullyAndStartAt(newOffset: Long): Unit = {
+  def truncateFullyAndStartAt(newOffset: Long): Unit = {
```
hachikuji, line 2144: It is a little unfortunate to expose the truncation APIs from `Log`. For logs which are managed by `LogManager`, truncation calls are expected to go through `LogManager`. That makes me think it might be useful to have a distinction between the low-level `Log` object and that which is managed by `LogManager`. For example, maybe we could use `ManagedLog` or something like that. Not something to address here, just food for thought in the way of future improvement.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
hachikuji: Hmm... It might be helpful to document this fact somewhere since it seems non-obvious. It surprised me anyway.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -66,32 +78,67 @@ class KafkaMetadataLog(
...
+  private def handleAndConvertLogAppendInfo(appendInfo: kafka.log.LogAppendInfo): LogAppendInfo = {
+    appendInfo.firstOffset match {
+      case Some(firstOffset) =>
+        if (firstOffset.relativePositionInSegment == 0) {
+          // Assume that a new segment was created if the relative position is 0
+          log.deleteOldSegments()
+        }
+        new LogAppendInfo(firstOffset.messageOffset, appendInfo.lastOffset)
+      case None =>
+        throw new KafkaException(s"Append failed unexpectedly: $appendInfo")
```
hachikuji: The `LogAppendInfo` contains an `errorMessage` field. Perhaps we can include it here?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -66,32 +78,67 @@ class KafkaMetadataLog(
...
+          )
+        }
+      }.orElse(0)
+    }
   }
 
   override def endOffsetForEpoch(leaderEpoch: Int): Optional[raft.OffsetAndEpoch] = {
     val endOffsetOpt = log.endOffsetForEpoch(leaderEpoch).map { offsetAndEpoch =>
-      new raft.OffsetAndEpoch(offsetAndEpoch.offset, offsetAndEpoch.leaderEpoch)
+      if (oldestSnapshotId.isPresent() &&
```
hachikuji: nit: might consider converting to scala Option and changing this to a `match`. A little easier on the eyes

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -66,32 +78,67 @@ class KafkaMetadataLog(
...
+        offsetAndEpoch.offset == oldestSnapshotId.get().offset &&
+        offsetAndEpoch.leaderEpoch == leaderEpoch) {
+
+        // The leaderEpoch is smaller thant the smallest epoch on the log. Overide the diverging
+        // epoch to the oldest snapshot which should be the snapshot at the log start offset
+        val snapshotId = oldestSnapshotId().get();
```
hachikuji: nit: remove semicolon here and below. also some unnecessary parenthesis in here

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -66,32 +78,67 @@ class KafkaMetadataLog(
...
+      }.orElse(0)
+    }
   }
 
   override def endOffsetForEpoch(leaderEpoch: Int): Optional[raft.OffsetAndEpoch] = {
     val endOffsetOpt = log.endOffsetForEpoch(leaderEpoch).map { offsetAndEpoch =>
-      new raft.OffsetAndEpoch(offsetAndEpoch.offset, offsetAndEpoch.leaderEpoch)
+      if (oldestSnapshotId.isPresent() &&
+        offsetAndEpoch.offset == oldestSnapshotId.get().offset &&
+        offsetAndEpoch.leaderEpoch == leaderEpoch) {
```
hachikuji: Always tough to think my way through this logic. So here we are handling the case where the client has requested the end offset for an epoch which is smaller than the entries remaining in the log. The current epoch cache implementation handles this by returning the requested epoch with an end offset equal to the log start offset. So we detect the case here by checking that the returned epoch matches the requested epoch and the end offset matches  the offset corresponding to the oldest snapshot, which should be the same as the log start offset. Right so far?

So there are still a couple cases that are possible after verifying this:

1. The requested epoch matches the snapshot epoch. We know then that the end offset of this epoch must be equal to the snapshot offset. This is a case where there is a new epoch which begins on the first offset in the log.
2. The requested epoch is less than the snapshot epoch. The follower might have been offline for a long time and is requesting an ancient epoch which we no longer have any knowledge of.

I'm convinced about the handling for the first case, but less sure about the second one. The guarantee that this API provides is that it will return the largest epoch which is less than or equal to the requested epoch and its last offset. But for this case, the snapshot epoch is larger than the request epoch. In fact, we have no way of satisfying this API because we have lost the old data.

Thinking in terms of what we want to happen might be helpful. We want to detect this as an out of range fetch which causes the follower to fetch from the latest snapshot. What if we return `Optional.empty` since we cannot determine the end offset for the requested epoch? Currently we do the following in `KafkaRaftClient`:

```java
       OffsetAndEpoch endOffsetAndEpoch = log.endOffsetForEpoch(lastFetchedEpoch).orElseThrow(() -> {
          return new IllegalStateException(
                String.format(
                    "Expected to find an end offset for epoch %s since it must be less than the current epoch %s",
                    lastFetchedEpoch,
                    quorum.epoch()
                )
            );
        });
```

There are two cases we want to handle: the requested epoch is less than any known epoch we have and the requested epoch is larger than any known epoch. The latter case might be possible if there was an "unclean" election of some kind. However, we can handle both cases the same. We can send the latest snapshot id to the follower, which will cause it to truncate. The follower may or may not see this as a loss of committed data and fail, but that is ok. At least we will get a clear message about the failure.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -131,6 +194,19 @@ class KafkaMetadataLog(
     }
   }
 
+  override def highWatermark: LogOffsetMetadata = {
+    val LogOffsetSnapshot(_, _, hwm, _) = log.fetchOffsetSnapshot
+    val segmentPosition: Optional[OffsetMetadata] = if (hwm.segmentBaseOffset != Log.UnknownOffset &&
```
hachikuji: Maybe use `hwm.messageOffsetOnly`?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -147,18 +223,102 @@ class KafkaMetadataLog(
...
+
+  override def oldestSnapshotId(): Optional[raft.OffsetAndEpoch] = {
+    oldestSnapshotId
+  }
+
+  override def onSnapshotFrozen(snapshotId: raft.OffsetAndEpoch): Unit = {
+    snapshotIds.add(snapshotId)
+  }
+
+  override def deleteToNewOldestSnapshotId(logStartSnapshotId: raft.OffsetAndEpoch): Boolean = {
```
hachikuji: nit: the name is a tad awkward. How about `deleteUpToSnapshot` or `deleteBeforeSnapshot`?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -113,6 +160,22 @@ class KafkaMetadataLog(
     log.truncateTo(offset)
   }
 
+  override def maybeTruncateFullyToLatestSnapshot(): Boolean = {
```
hachikuji: Having second thoughts about the name here (I know I suggested the `maybe`). Since there is a returned value, perhaps it is clear enough already and we can try a more concise name. How about just `truncateToLatestSnapshot`?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -983,48 +999,88 @@ private FetchResponseData tryCompleteFetchRequest(
...
+    private ValidatedFetchOffsetAndEpoch validateFetchOffsetAndEpoch(long fetchOffset, int lastFetchedEpoch) {
+        if (log.startOffset() == 0 && fetchOffset == 0) {
+            if (lastFetchedEpoch != 0) {
+                logger.warn(
+                    "Replica sent a zero fetch offset ({}) but the last fetched epoch ({}) was not zero",
+                    fetchOffset,
+                    lastFetchedEpoch
+                );
+            }
+            return ValidatedFetchOffsetAndEpoch.valid(new OffsetAndEpoch(fetchOffset, lastFetchedEpoch));
```
hachikuji: nit: should we use 0 for `fetchOffset` and `lastFetchedEpoch` here?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -983,48 +999,88 @@ private FetchResponseData tryCompleteFetchRequest(
...
+                    lastFetchedEpoch,
+                    quorum.epoch()
+                )
+            );
+        });
+
+        if (log.oldestSnapshotId().isPresent() &&
+            ((fetchOffset < log.startOffset()) ||
+             (fetchOffset == log.startOffset() && lastFetchedEpoch != log.oldestSnapshotId().get().epoch) ||
+             (lastFetchedEpoch < log.oldestSnapshotId().get().epoch))) {
```
hachikuji: Related to my other comment, but I think this check can go away if we change `endOffsetForEpoch` to only return the end offset and epoch when they can be definitively determined.

The other condition where we need to return a snapshot is when the returned epoch `endOffsetAndEpoch.epoch` equals the oldest snapshot epoch and the fetch offset is less than the oldest snapshot offset, which I think is equivalent to what is here. So I guess I'm thinking we can write this like this: 
```

OffsetAndEpoch endOffsetAndEpoch = log.endOffsetForEpoch(lastFetchedEpoch)
if (!endOffsetAndEpoch.isPresent) {
  // return latest snapshot
}

OffsetAndEpoch oldestSnapshotId = log.oldestSnapshotId().getOrElse(new OffsetAndEpoch(0, 0));
if (endOffsetAndEpoch.epoch == oldestSnapshotId.epoch && fetchOffset < oldestSnapshotId.offset) {
  // return latest snapshot
else if (endOffsetAndEpoch.epoch != lastFetchedEpoch || endOffsetAndEpoch.offset < fetchOffset) {
  // return diverging epoch
} else {
  // valid fetch
}
```

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: First, thanks a lot for thinking through this code and provide such detail comment. This code is important to get right.

> the requested epoch is larger than any known epoch.

For this case I decided to throw an exception because the Fetch request handling code already checks for this condition and returns an error Fetch response. The leader returns an error Fetch response when this is invariant is violated: `lastFetchedEpoch <= currentLeaderEpoch == quorum.epoch`. In other words based on the current implementation, I think it is a bug if `endOffsetForEpoch` returns `Optional.empty()`.

1. https://github.com/apache/kafka/blob/cbe435b34acb1a4563bc7c1f06895d2b52be2672/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java#L951-L954
2. https://github.com/apache/kafka/blob/cbe435b34acb1a4563bc7c1f06895d2b52be2672/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java#L1618-L1621

> the requested epoch is less than any known epoch we have

When thinking though this case I convinced myself. That the leader can determine if it should send a snapshot simply by comparing "fetch offset" and "last fetched epoch" against the `oldestSnapshotId`. The `oldestSnapshotId` is the snapshot with an end offset equal to the log start offset.

> The current epoch cache implementation handles this by returning the requested epoch with an end offset equal to the log start offset. So we detect the case here by checking that the returned epoch matches the requested epoch and the end offset matches the offset corresponding to the oldest snapshot, which should be the same as the log start offset. Right so far?

Correct. My comment here assumes that the fetch offset is between the log start offset and log end offset, and that sending a snapshot is not required. When thinking through the code in `ReplicatedLog::endOffsetForEpoch`, I always interpreted as trying to find the largest offset at which it the follower could have diverged. I would argue that the lower bound for this would be the `oldestSnapshotId` hence the modification if the leader cache couldn't find the end offset for the given epoch.

> However, we can handle both cases the same. We can send the latest snapshot id to the follower, which will cause it to truncate.

Regarding the "last fetched epoch" > `quorum.epoch()` case:
Even though this cannot happen based on my earlier comment, why not return `ValidatedFetchOffsetAndEpoch.diverging(logEndOffset, quorum.epoch())`? This is much easier for me to reason about and I think it will have the same effect on the follower. The follower will truncate to the end offset where the largest local epoch <= the leader epoch?

I think that sending `largestSnapshotId` will not cause the follower to `maybeTruncateFullyToLatestSnapshot` because `largestSnapshotId.epoch < log.latestEpoch` on the follower.

I need to think about it some more but I think the reality is that `KafkaRaftClient` doesn't handle the case very well when quorum.epoch() is not strictly increasing after a leader change.

### Proposed Changes

I think we can make this code much easier to understand if we remove `ReplicatedLog::endOffsetForEpoch` and instead add a method to `ReplicatedLog` with the following signature `ValidatedFetchOffsetAndEpoch validateOffsetAndEpoch(long offset, int epoch)`. We can move the code in `KafkaRaftClient::validateFetchOffsetAndEpoch` to this new method. I'll also rename `ValidatedFetchOffsetAndEpoch` to `ValidOffsetAndEpoch` but let me know if you have a better name. :)

```java
    /**
       * Validate the given offset and epoch against the log and oldest snapshot.
       *
       * Returns the largest valid offset and epoch given `offset` and `epoch` as the upper bound.
       * This can result in three possible values returned:
       *
       * 1. ValidatedOffsetAndEpoch.valid if the given offset and epoch is valid in the log.
       *
       * 2. ValidatedOffsetAndEpoch.diverging if the given offset and epoch is not valid; and the
       * largest valid offset and epoch is in the log.
       *
       * 3. ValidatedOffsetAndEpoch.snapshot if the given offset and epoch is not valid; and the largest
       * valid offset and epoch is less than the oldest snapshot.
       *
       * @param offset the offset to validate
       * @param epoch the epoch of the record at offset - 1
       * @return the largest valid offset and epoch
       */
    ValidOffsetAndEpoch validateOffsetAndEpoch(long offset, int epoch);
```

I am liking this design because it would also allow us to move some of the tests from `KafkaRaftClientSnapshotTest` to `KafkaMetadataLogTest`. Tests against `KafkaMetadataLogTest` are much easier to write and read. This design will also require the code to handle all possible values for `offset` and `epoch` since it is in `ReplicatedLog` and we cannot assume a specific implementation of `RaftClient`.

What do you think?

hachikuji: I'm ok with the suggestion about `validateFetchOffsetAndEpoch`. If we do that, my only ask is that we document at this level the cases that are possible and how they should be handled by the implementation. I don't think there are so many of them. Basically what happens when last fetched epoch/offset is out of range, what exactly indicates divergence, etc.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Hmmm. The case that I have in mind is this one:

```
oldestSnapshotId: OffsetAndEpoch(10, 2)
leaderEpochCache: (epoch: 4, startOffset: 10), (epoch: 5, starOffset: 20), ...
```

If the follower fetches `(fetchOffset: ..., lastFetchdEpoch: 3)`, then the leader should return `diverging(10, 2)`.

If the follower fetches `(fetchOffset: 15, lastFetchdEpoch: 2)`, then the leader should return `diverging(10, 2)`.

I think we can only do this if we have the check against the `oldestSnapshotId`.

core/src/main/scala/kafka/log/Log.scala
jsancio: Done.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Picked `deleteBeforeSnapshot`.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
jsancio: Done.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java

hachikuji (approved):
LGTM. Thanks for all the iterations.

hachikuji:
The failures look like known flakes. Will merge to trunk.

PR #3053, opened 2017-05-15
KAFKA-5239: Producer buffer pool allocates memory inside a lock

smccauliff:
Move byte buffer allocation out of lock.
Add unit test for restoring count when OOM is thrown from byte buffer allocation.

PR #9817, opened 2021-01-03
KAFKA-10899: Producer's BufferPool closing check

unlizhao:
According to the idea of #7967, it made BufferPool support closing.
Now, there are two places in the BufferPool#allocate() method to judge the 'closed' flag. One is when 'lock' is acquired, and the other is when 'condition' is awakened.

However, if the memory is allocated outside the 'lock' code block after 'freeup',
Therefore, it is possible for another thread to modify the 'closed' flag, causing this part of memory to be allocated in vain.

So two modifications have been made,
One is to add the 'volatile' modifier before 'closed',
One is to determine whether to 'closed' before allocating memory.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/producer/internals/BufferPool.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/internals/BufferPool.java
@@ -56,7 +56,7 @@ public class BufferPool {
     private final Metrics metrics;
     private final Time time;
     private final Sensor waitTime;
-    private boolean closed;
+    private volatile boolean closed;
 
     /**
      * Create a new buffer pool
@@ -212,10 +212,15 @@ public class BufferPool {
     /**
      * Allocate a buffer.  If buffer allocation fails (e.g. because of OOM) then return the size count back to
      * available memory and signal the next waiter if it exists.
+     *
+     * if producer is closing, will throw close-exception
      */
     private ByteBuffer safeAllocateByteBuffer(int size) {
         boolean error = true;
         try {
+            if (this.closed) {
+                throw new KafkaException("Producer closed while allocating memory");
+            }
             ByteBuffer buffer = allocateByteBuffer(size);
             error = false;
             return buffer;


unlizhao:
hi，@chia7712 @rajinisivaram PTAL，thanks

clients/src/main/java/org/apache/kafka/clients/producer/internals/BufferPool.java
```diff
@@ -212,10 +212,15 @@ protected void recordWaitTime(long timeNs) {
     /**
      * Allocate a buffer.  If buffer allocation fails (e.g. because of OOM) then return the size count back to
      * available memory and signal the next waiter if it exists.
+     *
+     * if producer is closing, will throw close-exception
      */
     private ByteBuffer safeAllocateByteBuffer(int size) {
         boolean error = true;
         try {
+            if (this.closed) {
```
chia7712, line 221: It is not holding the lock so this pool can be closed anytime. How we know the "best" place to check the close flag?

chia7712 (commented):
@unlizhao Thanks for your patch. one comment is left. Please take a look.

clients/src/main/java/org/apache/kafka/clients/producer/internals/BufferPool.java
unlizhao: @chia7712 thanks for your reviewed. This is what I thought when I wrote this code 

In the PR of #3053, for efficiency, the operation of allocating memory is moved out of the locked code block.
Subsequent modifications should be based on this consensus, and do some time-consuming operations without locking.

So what we can do is to try our best to check whether the memory can be allocated before allocating, and reduce the operation of allocating memory, just like the 'fail-fast' mechanism.

In this scenario, both `BufferPool#allocate()` and `BufferPool#close() `compete for locks.
Here we can think that lock blocking is more likely to occur in concurrent cases.

Finally, even if the memory is allocated in the `close` state, there is a plan to release the memory. Therefore, this COMMIT is like an optimization rather than bug-fix

PR #9469, opened 2020-10-21
MINOR: replace FetchRequest.TopicAndPartitionData by Map.Entry

chia7712:
This PR includes two changes.

1. replace FetchRequest.TopicAndPartitionData by Map.Entry (to reduce duplicate code)
1. move ```batchByTopic``` to ```FetchResponse``` 

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9818, opened 2021-01-03
MINOR: refactor FetchResponse#toMessage to avoid creating a lot of un…

chia7712:
Instead of creating intermediate collections to batch data, we directly check the last topic in auto-generated data to collect partitions having same topic.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
+++ b/clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
@@ -331,28 +331,32 @@ public class FetchResponse<T extends BaseRecords> extends AbstractResponse {
     private static <T extends BaseRecords> FetchResponseData toMessage(int throttleTimeMs, Errors error,
                                                                        Iterator<Map.Entry<TopicPartition, PartitionData<T>>> partIterator,
                                                                        int sessionId) {
-        FetchResponseData message = new FetchResponseData();
-        message.setThrottleTimeMs(throttleTimeMs);
-        message.setErrorCode(error.code());
-        message.setSessionId(sessionId);
-
         List<FetchResponseData.FetchableTopicResponse> topicResponseList = new ArrayList<>();
-        List<FetchRequest.TopicAndPartitionData<PartitionData<T>>> topicsData =
-                FetchRequest.TopicAndPartitionData.batchByTopic(partIterator);
-        topicsData.forEach(partitionDataTopicAndPartitionData -> {
-            List<FetchResponseData.FetchablePartitionResponse> partitionResponses = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


chia7712:
@ijuma This PR is from #9469 and it brings the better refactor :)

ijuma:
Thanks for the PR. Have you checked that we get the same batching this way? The behavior won't be the same unless the iterator order is correct.

chia7712:
@ijuma Thanks for your quick response!

> Have you checked that we get the same batching this way?

There is a test to verify the batch order (https://github.com/apache/kafka/blob/trunk/core/src/test/scala/unit/kafka/server/FetchRequestTest.scala#L83).

>  The behavior won't be the same unless the iterator order is correct.

There are two rules of batching partition data.

1. the order must be kept. the tp order from input (iterator) must be equal to tp order in ```FetchResponseData```.
2. batch the partitions only if we can keep the their order in ```FetchResponseData```. For example: we should group t0-p0, t0-p1 and t1-p2 to two ```FetchableTopicResponse```s since the t0-p0 and t0-p1 can be merged. By contrast, t0-p0, t1-p1 and t0-p2 ought to generate three ```FetchableTopicResponse```s.

It seems to me the reason of generating the intermediate collections before is that reusing the data in ```Struct/Schema``` is not convenient.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
```diff
@@ -331,28 +331,32 @@ public int sessionId() {
...
-                partitionData.partitionResponse.setPartition(partitionId);
+        partIterator.forEachRemaining(entry -> {
+            PartitionData<T> partitionData = entry.getValue();
+            // Since PartitionData alone doesn't know the partition ID, we set it here
+            partitionData.partitionResponse.setPartition(entry.getKey().partition());
+            // We have to keep the order of input topic-partition. Hence, we batch the partitions only if the last
+            // batch is in the same topic group.
+            if (!topicResponseList.isEmpty() && topicResponseList.get(topicResponseList.size() - 1)
+                    .topic().equals(entry.getKey().topic())) {
+                topicResponseList.get(topicResponseList.size() - 1)
```
ijuma: Maybe we could have a variable called `previousTopic` to make this code more readable.

ijuma (approved):
On closer inspection, I now remember that `batchByTopic` already relies on the ordering, so the new implementation is equivalent. LGTM overall. A nit below that should be easy to resolve.

clients/src/main/java/org/apache/kafka/common/requests/FetchResponse.java
chia7712: Nice suggestion. will address it later.

ijuma (approved):

PR #4574, opened 2018-02-15
MINOR: Free sends in MultiSend as they complete

hachikuji:
Currently we hold onto all Records references in a multi-partition
fetch response until the full response has completed. This can be
a problem when the records have been down-converted since they
will be occupying a potentially large chunk of memory. This patch
changes the behavior in MultiSend so that once a Send is completed,
we no longer keep a reference to it, which will allow the Records
objects to be freed sooner.

I have added a simple unit test to verify that sends are removed as the
MultiSend progresses.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9819, opened 2021-01-03
KAFKA-10694: Implement zero copy for FetchSnapshot

dengziming:
1. `RawSnapshotWriter` and `RawSnapshotReader` interact with `BaseRecords` instead of `ByteBuffer`
2. Use `fileRecords.slice` when read a snapshot
3. No big change of writing a snapshot

1. Change code in unit test of `RawSnapshotWriter` and `RawSnapshotReader` 
2. Add unit test in `SendBuilderTest` and `FileRawSnapshotTest`
3. Add unit test for `FileRegion` and `MemoryRegion`


hachikuji:
@dengziming Thanks for the patch. I have one high level comment. The approach we've taken here is to implement a new zero-copy "region" type. One alternative is to use the existing "records" type since it already supports "zero-copy" and the snapshot data is expected to be records data anyway. I'd like to understand the drawbacks of this approach a bit better. I think having two zero-copy types is a bit undesirable. An alternative is that we can let the "region" type subsume "records," but I suspect that will be tricky because of the down-conversion logic.

@jsancio also suggested that we might be able to represent the "region" type using `type=bytes zeroCopy=true`. I think that would require us to introduce an interface of some kind so that we can represent both a ByteBuffer and FileChannel. I'm not sure what that would look like, but it's also worth considering.

Anyway, just want to make sure we have covered all of the options before getting too far with this.

clients/src/main/java/org/apache/kafka/common/record/FileRegion.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.kafka.common.network.TransferableChannel;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+
+/**
+ * Represent a file with a start offset and length, we use it to send large snapshot file
+ */
+public class FileRegion implements BaseRegion {
```
ijuma: It seems like there is similarity with how we handle records. Why do we have to duplicate this logic? cc @jsancio

dengziming: We discussed here: https://github.com/apache/kafka/pull/9553#discussion_r549053981

ijuma: I see. Not sure I'm convinced, but I'll wait to hear from @jsancio and @hachikuji.

jsancio: I agree that it is very similar to how `FileRecords` works with the difference that the file region described doesn't have to be a valid record batch. It would be interesting to see if we can unify these abstractions.

dengziming: I replaced FileRegion with FileRecords locally and it worked smoothly even the start position isn't a valid record batch, which means the requirement is a soft requirement.
So we could just use FileRecords to implement zero-copy regardless of the requirement that position should be a record batch, or we could get the max position which is a valid record batch and also smaller than request position using `OffsetIndex` to satisfy the soft requirement.

hachikuji:
@dengziming @jsancio I'll give my take what I think we should do here, but let's try to agree before we go any further. Basically I am not really convinced that we need a new type, at least not in the near term. The snapshot data will be record data, so I cannot see a strong argument not to use the existing "records" type which already handles zero-copy. 

The one difference between `FetchSnapshot` and `Fetch` is that the latter is aligned on offsets while the former can indicate an arbitrary position in a file. I assume that is the main reason we have diverged here. However, I do not think this presents a major hurdle. On the server side, I think we have everything we need. `FileRawSnapshotReader` already has a `FileRecords` object and we can return a slice at an arbitrary position using `slice(int position, int size)`. On the client side, when we parse a snapshot response, `Readable` will give us back a reference to a `MemoryRecords` which will not necessarily be aligned. However, `MemoryRecords` gives us direct access to the underlying buffer, so the misalignment does not cause any problems.

Longer term, I think we can make an effort to replace the "records" type and just rely on `type=bytes, zeroCopy=true`. For this to work, we need a type which generalizes both a `ByteBuffer` and a `FileChannel` slice. For example, we could have something like this:

```java
interface ZeroCopyBuffer {
  ByteBuffer read(int length);
  Send toSend();
}
```

When parsing a request or response received over the network, it will be in memory anyway, so we need some hook to get access to the ByteBuffer. On the other hand, when transmitting a request or response, we let the implementation build the Send itself. This gives us a way to work with `RecordsSend`. There's probably more to it than that, but that's probably enough to get started. Then we would change the generated code to rely on the new type and let `SendBuilder` invoke `ZeroCopyBuffer.toSend` as appropriate.

Thoughts?

jsancio:
Replying to https://github.com/apache/kafka/pull/9819#issuecomment-761119104

Thanks @hachikuji for thinking about this and for giving a detail reply. For the long term solution, I am trying to understand how we transition from `type: "records"` to `type: "bytes", zeroCopy="true"`.

Outside of what you have enumerate, how would changing the type of a field work after the version bump? Would the generated message class need to support access to this field as both `BaseRecords` and `ZeroCopyBuffer`? Or would the generated class would just have `ZeroCopyBuffer`? In other words, what is the expectation of 3rd party clients with regards to the generated classes?

hachikuji:
@jsancio It's a good question. Up to now, we have not enforced strict compatibility for the JSON schema representation. As long as the on-the-wire bytes are the same, we have allowed some flexibility for the generator. It was only in the past couple months that we finally got all of the APIs converted, so the flexibility was necessary. Personally I'd like to retain this loose compatibility going forward, but I guess it will depend on what users want. If there is a strong demand to keep the schema definitions compatible, then we will need to come up with some way to evolve them. However, for now, I don't think we should consider JSON type compatibility to be a blocker.

ijuma:
The json schema is indeed not public api today. A KIP will be required before we make it so.

jsancio:
The proposal in https://github.com/apache/kafka/pull/9819#issuecomment-761119104 sounds good to me. I'll update the associated Jira soon.

dengziming:
> [comment](https://github.com/apache/kafka/pull/9819#issuecomment-761119104)

I have tried locally to replace the FileRegion with FileRecords, and it worked fine, which means that "aligned on offsets" is just a soft requirement of `FetchRequest`,  the `MemoryRecords` from `FetchResponse` is not necessarily be aligned, so it's viable to use `FileRecords` and `MemoryRecords` to implement zero-copy, the only additional work is to support `BaseRecords.writeTo(FileChannel)`

dengziming:
@hachikuji @jsancio @ijuma ,Hello, This pr is ready for a second review, Thank you.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1220,34 +1216,35 @@ private FetchSnapshotResponseData handleFetchSnapshotRequest(
...
                 maxSnapshotSize = Integer.MAX_VALUE;
             }
 
-            ByteBuffer buffer = ByteBuffer.allocate(Math.min(data.maxBytes(), maxSnapshotSize));
-            snapshot.read(buffer, partitionSnapshot.position());
-            buffer.flip();
+            try {
+                maxSnapshotPosition = Math.toIntExact(partitionSnapshot.position());
+            } catch (ArithmeticException e) {
+                maxSnapshotPosition = Integer.MAX_VALUE;
```
jsancio: Resetting the position to `Integer.MAX_VALUE` doesn't seem safe. I think the best we can do is let the exception bubble up.

Related to this, shouldn't the position be a long in `RawSnapshotReader::read`?

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
```diff
@@ -53,13 +56,20 @@ public long sizeInBytes() throws IOException {
...
         }
-
+        ByteBuffer buffer;
+        if (records instanceof MemoryRecords) {
+            buffer = ((MemoryRecords) records).buffer();
+        } else {
+            buffer = ByteBuffer.allocate(records.sizeInBytes());
+            ((FileRecords) records).channel().read(buffer);
+            buffer.flip();
+        }
```
jsancio, line 72: What do you think about keeping this signature as `void append(ByterBuffer buffer)`, having a static function that knows how to convert a `BaseRecords` to `ByteBuffer` and moving this responsibility to the `KafkaRaftClient`?

If I am not mistake, as of right now it is only the `KafkaRaftClient` that needs to deal with this mismatch.

clients/src/main/java/org/apache/kafka/common/record/BaseRegion.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.kafka.common.network.TransferableChannel;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+
+/**
+ * Base interface for accessing raft snapshot which could be file region or an in-memory buffers.
+ */
+public interface BaseRegion {
```
jsancio: I don't think this interface is used or needed. Are you planning to remove it?

jsancio (commented):
Thanks for the change. Partial review. I wanted to give you some high-level comments early.

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
dengziming: In fact, I change the signature to keep consistent with `FileRawSnapshotReader`, I'm also a little wobbly to make these changes.

jsancio: > I change the signature to keep consistent with FileRawSnapshotReader

Okay. I think this is something that I struggled with when creating the original APIs. I am okay with "inconsistent" APIs since `RawSnapshot{Reader,Writer}` are internal interfaces to the raft client and are not exposed to the state machine (controller).

I think this "inconsistency" will go away when we implement the long term solution.

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
```diff
@@ -1093,9 +1093,11 @@ public static final void readFully(InputStream inputStream, ByteBuffer destinati
         destinationBuffer.position(destinationBuffer.position() + totalBytesRead);
     }
 
-    public static void writeFully(FileChannel channel, ByteBuffer sourceBuffer) throws IOException {
+    public static int writeFully(FileChannel channel, ByteBuffer sourceBuffer) throws IOException {
+        int size = 0;
```
hachikuji: Hmm.. Not sure we need to compute this. Wouldn't it be the same as `sourceBuffer.remaining()`?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: I agree we should probably throw this. Snapshot size limits are an interesting point which I hadn't thought about. Currently `FileRecords` does not support files which are larger than Int.MaxValue. That gives us a 2GB limit. My feeling is that is probably good enough initially, but perhaps that adds some fuel for the effort to generalize the zero-copy support.

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
```diff
@@ -53,13 +56,20 @@ public long sizeInBytes() throws IOException {
...
             throw new IllegalStateException(
-                String.format("Append is not supported. Snapshot is already frozen: id = %s; temp path = %s", snapshotId, tempSnapshotPath)
+                    String.format("Append is not supported. Snapshot is already frozen: id = %s; temp path = %s", snapshotId, tempSnapshotPath)
             );
         }
-
+        ByteBuffer buffer;
+        if (records instanceof MemoryRecords) {
+            buffer = ((MemoryRecords) records).buffer();
+        } else {
```
hachikuji: I feel more inclined to raise an exception if we get a `BaseRecords` type that is not `MemoryRecords`. If we really get an unexpected file in here, then we need to reconsider the IO model instead of hiding a big copy. We could even make the expectation explicit in the parameter type even if it is not 100% symmetric with `RawSnapshotReader`.

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
dengziming: Done, make parameter type to MemoryRecords.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
dengziming: Thank you, done!

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
dengziming: In fact, this change was brought when using `BaseRegion`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1337,7 +1334,7 @@ private boolean handleFetchSnapshotResponse(
             throw new IllegalStateException(String.format("Received fetch snapshot response with an invalid position. Expected %s; Received %s", snapshot.sizeInBytes(), partitionSnapshot.position()));
         }
 
-        snapshot.append(partitionSnapshot.bytes());
+        snapshot.append((MemoryRecords) partitionSnapshot.bytes());
```
hachikuji: Can we add an `instanceof` check? We can raise an `IllegalStateException` with a nice message if it is not `MemoryRecords`.

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotReader.java
```diff
@@ -51,7 +51,7 @@ public long sizeInBytes() {
     }
 
     public BaseRecords read(long position, int size) throws IOException {
-        return fileRecords.slice(Math.toIntExact(position), size);
+        return fileRecords.slice((int) position, size);
```
hachikuji: Using `Math.toIntExact` seemed better.

clients/src/main/resources/common/message/FetchSnapshotResponse.json
```diff
@@ -51,8 +51,7 @@
           "about": "The total size of the snapshot." },
         { "name": "Position", "type": "int64", "versions": "0+",
           "about": "The starting byte position within the snapshot included in the Bytes field." },
-        { "name": "Bytes", "type": "bytes", "versions": "0+", "zeroCopy": true,
-          "about": "Snapshot data." }
+        { "name": "Bytes", "type": "records", "versions": "0+", "about": "Snapshot data." }
```
hachikuji: I think it's worth adding a comment which explains that records data may not be aligned on an offset boundary. Perhaps we could even change the field name to "UnalignedRecords"?

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
```diff
@@ -53,13 +54,13 @@ public long sizeInBytes() throws IOException {
     }
 
     @Override
-    public void append(ByteBuffer buffer) throws IOException {
+    public void append(MemoryRecords records) throws IOException {
```
hachikuji: I am considering how to improve this API in the interim to make the alignment expectation clearer. In fact, the `BaseRecords` interface is really minimal and makes no assumptions about alignment. Suppose we introduce two new types:

```java
// Represents a record set which is not necessarily offset-aligned
interface UnalignedRecords extends BaseRecords;

class UnalignedMemoryRecords extends UnalignedRecords {
  ByteBuffer buffer;
}
class UnalignedFileRecords extends UnalignedRecords {
  FileChannel channel;
}
```

I think these could be very simple implementations and they would improve the type safety of these APIs. Then we can use `UnalignedRecords` in the in the `RawSnapshotWriter` and `RawSnapshotReader` interfaces which restores the symmetry. 

What do you think?

dengziming: This will be more clear and got 3 little problems when I tried:
1. `Readable.readRecords` still returns a `MemoryRecords` and we need to convert it to `UnalignedMemoryRecords`, this is simple
2. We need to implement `UnalignedFileRecords.toSend` which need to return RecordSend, but this is unreasonable, I added a `FileChannelSend` similar to `ByteBufferSend` to solve this problem
3. We should still use `UnalignedMemoryRecords` in `FileRawSnapshotWriter` so we don't restore symmetry

@hachikuji @jsancio ,I think my implementation is not the perfect one, please take a look to see whether we could make some improvement, thank you.

hachikuji:
@dengziming Thanks for the updates. I think we're on the right track. I thought of a few potential simplifications. Take a look here: https://github.com/hachikuji/kafka/commit/01cc3aaa147064e60624f8b759ede54b95242a57.

dengziming:
@hachikuji ,Thank you for your suggestions, I found `DefaultRecordsSend` similar to `UnalignedRecordsSend ` so I added a generic type to`DefaultRecordsSend`, and also added a parent interface `TransferableRecords` to `Records` and `UnalignedRecords`, please take a look.

clients/src/main/java/org/apache/kafka/common/record/SliceRange.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.common.record;
+
+
+public class SliceRange {
```
hachikuji: If this is only used in `FileRecords`, perhaps we can move it there and make it `private static`.

clients/src/main/java/org/apache/kafka/common/record/FileRecords.java
```diff
@@ -134,6 +134,16 @@ public void readInto(ByteBuffer buffer, int position) throws IOException {
      * @return A sliced wrapper on this message set limited based on the given position and size
      */
     public FileRecords slice(int position, int size) throws IOException {
+        SliceRange range = sliceRange(position, size);
+        return new FileRecords(file, channel, range.start, range.end, true);
+    }
+
+    public UnalignedFileRecords sliceUnaligned(int position, int size) {
```
hachikuji, line 153: Can we add method documentation? I think the key thing to note is that this method is reserved for cases where the position is not necessarily on an offset boundary. It is worth mentioning the `FetchSnapshot` use case explicitly.

clients/src/main/java/org/apache/kafka/common/record/FileRecords.java
```diff
@@ -144,11 +154,12 @@ public FileRecords slice(int position, int size) throws IOException {
...
-        int end = this.start + position + size;
+        int sliceStart = this.start + position;
+        int sliceEnd = sliceStart + size;
         // handle integer overflow or if end is beyond the end of the file
-        if (end < 0 || end > start + currentSizeInBytes)
-            end = start + currentSizeInBytes;
-        return new FileRecords(file, channel, this.start + position, end, true);
+        if (sliceEnd < 0 || sliceEnd > start + currentSizeInBytes)
+            sliceEnd = start + currentSizeInBytes;
+        return new SliceRange(sliceStart, sliceEnd);
```
hachikuji: I was thinking how we can simplify this a little bit and avoid the annoying object. I think the main logic here is validating the input and computing the size of the available bytes to read. 

```java
private int availableBytes(int position, int maxLength) {
   ...
   int limit = this.start + position + size;
   if (limit < 0 || limit > start + currentSizeInBytes)
     limit = start + currentSizeInBytes;
   return limit - position;
}
```
Then the caller can do something like this:
```
  int availableBytes = availableBytes(position, size);
  return new UnalignedFileRecords(channel, this.start + position, availableBytes);
```
What do you think? Any better?

clients/src/main/java/org/apache/kafka/common/record/TransferableRecords.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+package org.apache.kafka.common.record;
+
+import org.apache.kafka.common.network.TransferableChannel;
+
+import java.io.IOException;
+
+/**
+ * Represents a record set which can be transfer to a channel
```
hachikuji: nit: 
> Represents a record set which can be transferred to a channel.

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
```diff
@@ -1098,6 +1099,23 @@ public static void writeFully(FileChannel channel, ByteBuffer sourceBuffer) thro
             channel.write(sourceBuffer);
     }
 
+    public static long tryWriteTo(TransferableChannel destChannel,
```
hachikuji, line 1114: nit: add documentation?

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
```diff
@@ -53,13 +54,13 @@ public long sizeInBytes() throws IOException {
     }
 
     @Override
-    public void append(ByteBuffer buffer) throws IOException {
+    public void append(UnalignedMemoryRecords records) throws IOException {
```
hachikuji, line 57: It might be helpful to have an `append(MemoryRecords)`. That saves an annoying conversion when we are generating the snapshot instead of copying it from the leader.

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
```diff
@@ -1098,6 +1099,23 @@ public static void writeFully(FileChannel channel, ByteBuffer sourceBuffer) thro
...
+                                  long position,
+                                  int length,
+                                  ByteBuffer sourceBuffer) throws IOException {
+        if (position > Integer.MAX_VALUE)
```
hachikuji: Is this a restriction of this method itself or are we just carrying over the limitation from `FileRecords`?

clients/src/main/resources/common/message/FetchSnapshotResponse.json
```diff
@@ -51,8 +51,8 @@
           "about": "The total size of the snapshot." },
         { "name": "Position", "type": "int64", "versions": "0+",
           "about": "The starting byte position within the snapshot included in the Bytes field." },
-        { "name": "Bytes", "type": "bytes", "versions": "0+", "zeroCopy": true,
-          "about": "Snapshot data." }
+        { "name": "Bytes", "type": "records", "versions": "0+",
```
hachikuji: I had a previous comment about changing the name here from "Bytes" to "UnalignedRecords." What do you think about it?

clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
```diff
@@ -137,6 +138,9 @@ public void writeRecords(BaseRecords records) {
         if (records instanceof MemoryRecords) {
             flushPendingBuffer();
             addBuffer(((MemoryRecords) records).buffer());
+        } else if (records instanceof UnalignedMemoryRecords) {
```
hachikuji, line 141: Are there any tests we can update to cover this?

clients/src/main/java/org/apache/kafka/common/record/UnalignedFileRecords.java
```diff
@@ -0,0 +1,50 @@
...
+        this.size = size;
+    }
+
+    @Override
+    public int sizeInBytes() {
+        return size;
+    }
+
+    @Override
+    public long writeTo(TransferableChannel destChannel, long previouslyWritten, int remaining) throws IOException {
```
hachikuji, line 45: Do we have test coverage for this?

hachikuji (commented):
Thanks, this is looking good. Left a few more comment, but I think we're almost ready to wrap this up.

clients/src/main/java/org/apache/kafka/common/protocol/SendBuilder.java
dengziming: Add a testZeroCopyUnalignedRecords

clients/src/main/java/org/apache/kafka/common/record/UnalignedFileRecords.java
dengziming: Add UnalignedFileRecordsTest

clients/src/main/resources/common/message/FetchSnapshotResponse.json
dengziming: I have a question here, should we bump the version when changing the name?

raft/src/main/java/org/apache/kafka/snapshot/FileRawSnapshotWriter.java
dengziming: Keep 2 methods here, one for MemoryRecords and one for UnalignedMemoryRecords.

clients/src/main/resources/common/message/FetchSnapshotResponse.json
hachikuji: We do not need to. The name is not part of the serialization. It only affects the generated code.

clients/src/test/java/org/apache/kafka/common/record/UnalignedFileRecordsTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @BeforeEach
+    public void setup() throws IOException {
+        this.fileRecords = createFileRecords(values);
+    }
+
+    @Test
+    public void testWriteTo() throws IOException {
+
+        org.apache.kafka.common.requests.ByteBufferChannel channel = new org.apache.kafka.common.requests.ByteBufferChannel(fileRecords.sizeInBytes());
```
hachikuji, line 53: nit: why not import `ByteBufferChannel`?

clients/src/test/java/org/apache/kafka/common/record/UnalignedFileRecordsTest.java
```diff
@@ -0,0 +1,73 @@
...
+        int size = fileRecords.sizeInBytes();
+
+        UnalignedFileRecords records1 = fileRecords.sliceUnaligned(0, size / 2);
+        UnalignedFileRecords records2 = fileRecords.sliceUnaligned(size / 2, size - size / 2);
```
hachikuji, line 57: nit: `size - size / 2` looks a little funny. Is that the same as `size / 2 + 1`?

raft/src/main/java/org/apache/kafka/snapshot/RawSnapshotWriter.java
```diff
@@ -29,39 +31,52 @@
...
+     * Fully appends the memory record set to the snapshot, the difference with {@link RawSnapshotWriter#append(MemoryRecords)}
+     * is that the record set are fetched from leader by FetchSnapshotRequest, so the records are unaligned.
+     *
+     * If the method returns without an exception the given records was fully writing the
+     * snapshot.
+     *
+     * @param records the region to append
+     * @throws IOException for any IO error during append
+     */
+
```
hachikuji: nit: remove newline

raft/src/main/java/org/apache/kafka/snapshot/SnapshotWriter.java
```diff
@@ -146,8 +149,9 @@ public void close() throws IOException {
 
     private void appendBatches(List<CompletedBatch<T>> batches) throws IOException {
         try {
-            for (CompletedBatch batch : batches) {
-                snapshot.append(batch.data.buffer());
+            for (CompletedBatch<T> batch : batches) {
+                ByteBuffer buffer = batch.data.buffer();
+                snapshot.append(new UnalignedMemoryRecords(buffer));
```
hachikuji: Could we use the overload `snapshot.append(batch.data)`?

raft/src/test/java/org/apache/kafka/raft/MockLog.java
```diff
@@ -522,12 +524,21 @@ public long sizeInBytes() {
         }
 
         @Override
-        public void append(ByteBuffer buffer) {
+        public void append(UnalignedMemoryRecords records) {
             if (frozen) {
                 throw new RuntimeException("Snapshot is already frozen " + snapshotId);
             }
+            data.write(records.buffer());
+
```
hachikuji: nit: remove newline

clients/src/main/java/org/apache/kafka/common/record/FileRecords.java
```diff
@@ -134,6 +134,27 @@ public void readInto(ByteBuffer buffer, int position) throws IOException {
      * @return A sliced wrapper on this message set limited based on the given position and size
      */
     public FileRecords slice(int position, int size) throws IOException {
+        int availableBytes = availableBytes(position, size);
+        return new FileRecords(file, channel, this.start + position, this.start + position + availableBytes, true);
```
hachikuji: nit: since we have `start + position` here a couple times, perhaps we may as well use a field:
```
int startPosition = this.start + position;
```

clients/src/main/java/org/apache/kafka/common/record/FileRecords.java
```diff
@@ -144,11 +165,11 @@ public FileRecords slice(int position, int size) throws IOException {
         if (size < 0)
             throw new IllegalArgumentException("Invalid size: " + size + " in read from " + this);
 
-        int end = this.start + position + size;
-        // handle integer overflow or if end is beyond the end of the file
-        if (end < 0 || end > start + currentSizeInBytes)
-            end = start + currentSizeInBytes;
-        return new FileRecords(file, channel, this.start + position, end, true);
+        int limit = this.start + position + size;
```
hachikuji: nit: my bad for suggesting `limit`. Can we use `end` instead as in the original code?

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
```diff
@@ -1098,6 +1099,29 @@ public static void writeFully(FileChannel channel, ByteBuffer sourceBuffer) thro
             channel.write(sourceBuffer);
     }
 
+    /**
+     * Trying to write data in source buffer to a {@link TransferableChannel}, we may need to call this method multiple
+     * times since this method doesn't ensure data in source buffer can be fully written to dest channel.
```
hachikuji: nit: 
> ... since this method doesn't ensure the data in the source buffer can be fully written to the destination channel.

clients/src/main/java/org/apache/kafka/common/record/FileRecords.java
```diff
@@ -134,6 +134,27 @@ public void readInto(ByteBuffer buffer, int position) throws IOException {
...
+        int availableBytes = availableBytes(position, size);
+        return new FileRecords(file, channel, this.start + position, this.start + position + availableBytes, true);
+    }
+
+    /**
+     * Return a slice of records from this instance, the difference with {@link FileRecords#slice(int, int)} is
+     * that the position is not necessarily on an offset boundary.
+     *
+     * This method is reserved for cases where we do not necessarily need to read an entire record batch such as
+     * fetching snapshot in for raft.
```
hachikuji: nit:
> This method is reserved for cases where offset alignment is not necessary, such as in the replication of raft snapshots.

clients/src/test/java/org/apache/kafka/common/record/UnalignedFileRecordsTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private byte[][] values = new byte[][] {
+            "foo".getBytes(),
+            "bar".getBytes()
+    };
+    private FileRecords fileRecords;
+
+    @BeforeEach
+    public void setup() throws IOException {
+        this.fileRecords = createFileRecords(values);
```
hachikuji, line 42: nit: should we add an `@After` so that we can close the file?

hachikuji (commented):
Thanks, should be the last round of reviews (at least from me).

clients/src/main/java/org/apache/kafka/common/record/UnalignedFileRecords.java
jsancio: Hmm. The semantic between the in-memory implementation vs the file channel implementation is slightly different. The in-memory version throws an exception if there are not enough bytes to send `length`. While the file channel version just sends less bytes if there is not enough to send `remaining`.

I suspect that we want to throw in both cases else the calling code won't know why `writeTo` sent less bytes. It could loop forever if it is attempting to send up to `remaining` bytes.

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
```diff
@@ -1098,6 +1099,29 @@ public static void writeFully(FileChannel channel, ByteBuffer sourceBuffer) thro
...
+     * @throws IOException If an I/O error occurs
+     */
+    public static long tryWriteTo(TransferableChannel destChannel,
+                                  int position,
+                                  int length,
+                                  ByteBuffer sourceBuffer) throws IOException {
+
+        ByteBuffer dup = sourceBuffer.duplicate();
+        dup.position(position);
+        dup.limit(position + length);
```
jsancio, line 1121: Isn't this `position()` and `limit()` modification the same as `ByteBuffer.slice`?

Having said that I find it strange that we want to set the absolute position and limit. Shouldn't this be set relative to the current position? For example think about `sourceBuffer.position() > position`.

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
```diff
@@ -1098,6 +1099,29 @@ public static void writeFully(FileChannel channel, ByteBuffer sourceBuffer) thro
...
+     * @param length The max size of bytes can be written
+     * @param sourceBuffer The source buffer
+     *
+     * @return The length of the actual written data
+     * @throws IOException If an I/O error occurs
+     */
+    public static long tryWriteTo(TransferableChannel destChannel,
+                                  int position,
+                                  int length,
+                                  ByteBuffer sourceBuffer) throws IOException {
```
jsancio, line 1117: How about removing the IO from this method with a signature like:
```java
public static ByteBuffer relativeSlice(ByteBuffer buffer, int position, int length);
```
The caller of this method can then `long written = destChannel.write(Utils.relativeSlice(buffer, position, length))`.

clients/src/test/java/org/apache/kafka/common/record/UnalignedFileRecordsTest.java
jsancio: `4 - 4/2 = 2` vs `4/2 + 1 = 3`.

raft/src/test/java/org/apache/kafka/snapshot/FileRawSnapshotTest.java
```diff
@@ -107,6 +112,48 @@ public void testWriteReadSnapshot() throws IOException {
...
+
+        ByteBuffer buffer1 = expectedBuffer.duplicate();
+        buffer1.position(0);
+        buffer1.limit(expectedBuffer.limit() / 2);
+        ByteBuffer buffer2 = expectedBuffer.duplicate();
+        buffer2.position(expectedBuffer.limit() / 2);
+        buffer2.limit(expectedBuffer.limit());
+
+        try (FileRawSnapshotWriter snapshot = FileRawSnapshotWriter.create(tempDir, offsetAndEpoch)) {
+
```
jsancio: Extra newline.

raft/src/test/java/org/apache/kafka/snapshot/FileRawSnapshotTest.java
```diff
@@ -107,6 +112,48 @@ public void testWriteReadSnapshot() throws IOException {
...
+            snapshot.append(new UnalignedMemoryRecords(buffer1));
+            snapshot.append(new UnalignedMemoryRecords(buffer2));
+            snapshot.freeze();
+        }
+
+        try (FileRawSnapshotReader snapshot = FileRawSnapshotReader.open(tempDir, offsetAndEpoch)) {
+
```
jsancio: Extra newline.

jsancio (commented):
@dengziming thanks for the changes. @hachikuji thanks for the reviews.

clients/src/test/java/org/apache/kafka/common/record/UnalignedFileRecordsTest.java
dengziming: ImportControl doesn't allow us to import ByteBufferChannel here, this is similar to `NonOverflowingByteBufferChannel` in your pr #4574

clients/src/test/java/org/apache/kafka/common/record/UnalignedFileRecordsTest.java
dengziming: Size may be an odd number or even number.

clients/src/main/java/org/apache/kafka/common/utils/Utils.java
dengziming: There are many occurrences of `position`+`limit` in Kafka project, I think this is because the developers of Kafka like to handle the ByteBuffer by ourselves.

clients/src/main/java/org/apache/kafka/common/record/UnalignedFileRecords.java
dengziming: @jsancio , I think you are right, I also find that the implementation of `MemoryRecords` and `FileRecords` is also different, I will spend some time checking this.

hachikuji (approved):
LGTM. Thanks for the great contribution!

@jsancio If you have any remaining concerns, let's address them in follow-up patches.

dengziming:
@hachikuji @jsancio Thank you for your great patience!
@jsancio your concerns seems similar to https://github.com/apache/kafka/pull/9970#discussion_r564747239 , we can discuss it there.

PR #8200, opened 2020-03-02
KAFKA-5876: IQ should throw different exceptions for different errors(part 1)

vitojeng:
> [KIP-216: IQ should throw different exceptions for different errors](https://cwiki.apache.org/confluence/display/KAFKA/KIP-216%3A+IQ+should+throw+different+exceptions+for+different+errors)

KAFKA-5876's PR break into multiple parts. This PR is part 1: the new exceptions introduced in KIP-216

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9821, opened 2021-01-04
KAFKA-5876: Apply UnknownStateStoreException for Interactive Queries

vitojeng:
follow-up #8200

KAFKA-5876's PR break into multiple parts, this PR is part 2 - apply UnknownStateStoreException

### DIFF
--- a/docs/streams/upgrade-guide.html
+++ b/docs/streams/upgrade-guide.html
@@ -94,6 +94,9 @@
 
     <h3><a id="streams_api_changes_300" href="#streams_api_changes_300">Streams API changes in 3.0.0</a></h3>
     <p>
+        We applyed UnknownStateStoreException to <code>KafkaStreams#store()</code>: If the specified store name does not exist in the topology, UnknownStateStoreException will be thrown instead InvalidStateStoreException(<a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-216%3A+IQ+should+throw+different+exceptions+for+different+errors">KIP-216</a>).
+    </p>
+    <p>
         We removed the default implementation of <code>RocksDBConfigSetter#close()</code>.
     </p>
     <p>
--- a/docs/streams/upgrade-guide.html
+++ b/docs/streams/upgrade-guide.html
@@ -94,7 +94,7 @@
 
     <h3><a id="streams_api_changes_300" href="#streams_api_changes_300">Streams API changes in 3.0.0</a></h3>
     <p>
-        We applyed UnknownStateStoreException to <code>KafkaStreams#store()</code>: If the specified store name does not exist in the topology, UnknownStateStoreException will be thrown instead InvalidStateStoreException(<a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-216%3A+IQ+should+throw+different+exceptions+for+different+errors">KIP-216</a>).
+        A new exception may be thrown from <code>KafkaStreams#store()</code>. If the specified store name does not exist in the topology, an UnknownStateStoreException will be thrown instead of the former InvalidStateStoreException. See <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-216%3A+IQ+should+throw+different+exceptions+for+different+errors">KIP-216</a> for more information.
     </p>
     <p>
         We removed the default implementation of <code>RocksDBConfigSetter#close()</code>.
--- a/streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
@@ -36,6 +36,7 @@ import org.apache.kafka.common.utils.MockTime;
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.streams.errors.StreamsUncaughtExceptionHandler;
 import org.apache.kafka.streams.errors.TopologyException;
+import org.apache.kafka.streams.errors.UnknownStateStoreException;
 import org.apache.kafka.streams.internals.metrics.ClientMetrics;
 import org.apache.kafka.streams.kstream.Materialized;
 import org.apache.kafka.streams.processor.StateRestoreListener;
@@ -91,6 +92,7 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
+import static org.apache.kafka.streams.state.QueryableStoreTypes.keyValueStore;
 import static org.apache.kafka.streams.integration.utils.IntegrationTestUtils.safeUniqueTestName;
 import static org.easymock.EasyMock.anyInt;
 import static org.easymock.EasyMock.anyLong;
@@ -743,6 +745,13 @@ public class KafkaStreamsTest {
     public void shouldNotGetQueryMetadataWithPartitionerWhenNotRunningOrRebalancing() {
         final KafkaStreams streams = new KafkaStreams(getBuilderWithSource().build(), props, supplier, time);
         assertThrows(IllegalStateException.class, () -> streams.queryMetadataForKey("store", "key", (topic, key, value, numPartitions) -> 0));
+    }
+
+    @Test(expected = UnknownStateStoreException.class)
+    public void shouldThrowUnknownStateStoreExceptionWhenStoreNotExist() throws Exception {
+        final KafkaStreams streams = new KafkaStreams(getBuilderWithSource().build(), props, supplier, time);
+        streams.start();
+        streams.store(StoreQueryParameters.fromNameAndType("unknown-store", keyValueStore()));
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


vitojeng:
@mjsax Sorry for the long pause, please take a look. :)

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -752,6 +754,13 @@ public void shouldNotGetQueryMetadataWithPartitionerWhenNotRunningOrRebalancing(
         streams.queryMetadataForKey("store", "key", (topic, key, value, numPartitions) -> 0);
     }
 
+    @Test(expected = UnknownStateStoreException.class)
```
mjsax: I would prefer not to use this annotation but instead use `assertThrows()` in the test. I know that other tests also use the annotation but those are older tests, and we try to move off this old pattern in favor of using `assertThrows`.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -752,6 +754,13 @@ public void shouldNotGetQueryMetadataWithPartitionerWhenNotRunningOrRebalancing(
...
+    public void shouldThrowUnknownStateStoreExceptionWhenStoreNotExist() throws Exception {
+        final KafkaStreams streams = new KafkaStreams(getBuilderWithSource().build(), props, supplier, time);
```
mjsax: We should use a try-with-resources clause to make sure `close()` is called.

Seems, other tests in this class have a similar issue. Would be good to fix all test accordingly (if you want you can also to a separate PR for it).

mjsax (commented):
Glad to have you back!

I cannot remember the details of the KIP but should we throw this new exception also for `allMetadataForStore` and `queryMetadataForKey`?

\cc @vvcephei

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
vitojeng: Got it, will do.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
vitojeng: > We should use a try-with-resources clause to make sure `close()` is called.

Got it

> Seems, other tests in this class have a similar issue. Would be good to fix all test accordingly (if you want you can also to a separate PR for it).

Ok. I would like to create another PR for it.

vitojeng:
> I cannot remember the details of the KIP but should we throw this new exception also for `allMetadataForStore` and `queryMetadataForKey`?

@mjsax In the KIP-216, we talk about all sub types of InvalidStateStoreException. That means this KIP only focus on the existing code that throws InvalidStateStoreException, IMHO. Seems there not have any InvalidStateStoreException thrown in the `allMetadataForStore` and `queryMetadataForKey`. So we didn't discuss whether we should throw this exception(UnknownStateStoreException) in the past.

It looks like it will break the API for the new exception. If the answer is 'Yes', should we need another KIP?

ableegoldman:
Hey @mjsax and @vitojeng , are you still interested in this? I haven't been following this KIP myself but I think the IQ exceptions are a valuable addition to Streams, and it would be nice to have this KIP finally pushed across the finish line 🙂

vitojeng:
> Hey @mjsax and @vitojeng , are you still interested in this? I haven't been following this KIP myself but I think the IQ exceptions are a valuable addition to Streams, and it would be nice to have this KIP finally pushed across the finish line 🙂

@ableegoldman Yes. Personally, I hope to complete this KIP's work. 😄

ableegoldman:
Awesome. I can help get this current PR reviewed and merged from here, and probably find you someone else to review the next PR(s) since I'm pretty busy 🙂

Now, regarding your last question on this PR: am I reading it correctly that we just don't ever throw `InvalidStateStoreException` from `allMetadataForStore` or `queryMetadataForKey` at the moment, therefore it doesn't make sense to throw UnknownStateStoreException from these methods as part of this KIP?

Personally, I think it's ok to just throw whatever exception makes sense from wherever in the code it makes sense to do so. You can send a quick update note to the KIP thread to say that you're making this amendment, and if anyone has a concern they can respond there.

For this specific case: it certainly seems to make sense that `allMetadataForStore` would throw an UnknownStateStoreException if a store name was passed in that's, well, unknown. I do see that we don't currently throw anything like that, and it looks like we would just return nothing if an invalid store was passed in, but that was probably just an oversight -- I don't see why it should behave any differently from `KafkaStreams#store`. Same goes for `queryMetadataForKey`

vitojeng:
> Awesome. I can help get this current PR reviewed and merged from here, and probably find you someone else to review the next PR(s) since I'm pretty busy 🙂

Great, thanks.

 > Now, regarding your last question on this PR: am I reading it correctly that we just don't ever throw `InvalidStateStoreException` from `allMetadataForStore` or `queryMetadataForKey` at the moment, therefore it doesn't make sense to throw UnknownStateStoreException from these methods as part of this KIP?
>
> Personally, I think it's ok to just throw whatever exception makes sense from wherever in the code it makes sense to do so. You can send a quick update note to the KIP thread to say that you're making this amendment, and if anyone has a concern they can respond there.

@ableegoldman I just feel that I may need to point out this(will break the API for the new exception).
Your suggestion is great. I'm happy to update the KIP and update the PR for this.

cadonna:
@vitojeng @mjsax @ableegoldman 
My 2 euro cent: 
Looking at the code for `allMetadataForStore` and `queryMetadataForKey` it seems like the case when a state store is not unknown is not handled by throwing an exception but by returning `null` or by returning `UNKNOWN_HOST`. Since it is documented (almost) that way in the javadocs and the relevant KIPs ([KIP-67](https://cwiki.apache.org/confluence/display/KAFKA/KIP-67%3A+Queryable+state+for+Kafka+Streams) and [KIP-535](https://cwiki.apache.org/confluence/display/KAFKA/KIP-535%3A+Allow+state+stores+to+serve+stale+reads+during+rebalance)) do not mention an exception, I think we need a proper update to the existing KIP or a new dedicated KIP for it (both with voting), because it does not seem to be an oversight in the implementation, but rather an oversight in the design of the API. If it were an oversight in then implementation you could argue that users relied on buggy code and a quick update on the KIP thread would be fine. However, with an oversight in the API design, I think a  KIP with additions of new methods and deprecations of old methods is needed. If it is an update of the current KIP or a new one in order to not block the existing one is less important (although I am in favor of the latter). WDYT?

ableegoldman:
Fine with me. It does seem odd that we'd be inconsistent, but I don't think we need to solve everything with this one KIP/PR.  Let's just file a ticket for this so we don't forget.

(FWIW I think it was neither an oversight nor an intentional design, at least in the case of `queryMetadataForKey`, since this KIP-535 was started and completed while this KIP was still in progress, and the author most likely wasn't aware that we had plans to improve the IQ exceptions.)

I think in that case, this PR covers everything it needs to. @vitojeng Can you just add a quick note about this new exception to the streams upgrade guide under the section for API changes in 3.0?

vitojeng:
@cadonna , @ableegoldman 

Thanks for great discussion.
I'll rebase trunk and update the streams upgrade guide.

docs/streams/upgrade-guide.html
```diff
@@ -93,6 +93,9 @@ <h1>Upgrade Guide and API Changes</h1>
     </p>
 
     <h3><a id="streams_api_changes_300" href="#streams_api_changes_300">Streams API changes in 3.0.0</a></h3>
+    <p>
+        We applyed UnknownStateStoreException to <code>KafkaStreams#store()</code>: If the specified store name does not exist in the topology, UnknownStateStoreException will be thrown instead InvalidStateStoreException(<a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-216%3A+IQ+should+throw+different+exceptions+for+different+errors">KIP-216</a>).
```
ableegoldman:
```suggestion
        A new exception may be thrown from <code>KafkaStreams#store()</code>. If the specified store name does not exist in the topology, an UnknownStateStoreException will be thrown instead of the former InvalidStateStoreException. See <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-216%3A+IQ+should+throw+different+exceptions+for+different+errors">KIP-216</a> for more information.
```

vitojeng: Thanks @ableegoldman !

ableegoldman (approved):
Alright, LGTM. @vitojeng please ping me when the PR build is done and I'll merge

vitojeng:
@ableegoldman 
The PR build is done, please talk a look.

ableegoldman:
Just two unrelated test failures in the build: 
`connect.mirror.integration.MirrorConnectorsIntegrationSSLTest.testReplication()`
`kafka.integration.MetricsDuringTopicCreationDeletionTest.testMetricsDuringTopicCreateDelete()`

ableegoldman:
Merged to trunk, ready for the next PR 🙂

vitojeng:
> Merged to trunk, ready for the next PR 🙂

Will do, thanks @ableegoldman !
:)

PR #9822, opened 2021-01-05
KAFKA-10860: Fix NPE in JmxTool, add tests

tombentley:
Fixes [KAFKA-10860](https://issues.apache.org/jira/browse/KAFKA-10860) and adds a unit test.

The existing code was not easily testable, so refactored to allow the main logic to be executed without invoking `System.exit` or equivalent. The tests added here cover most of the non-connection options, but have to use `--one-time`. I just use the current JVM's MBean server for these tests. Testing the connection code would require spinning up a JVM to connect to, or using the Attach API (which is prevented by https://bugs.openjdk.java.net/browse/JDK-8180425).

Although `--wait` was described as not supported in the presence of a pattern but this wasn't validated previously. Supporting that actually makes the code simpler by avoiding the need to special-case the presence of a pattern. The semantics are that we wait until each of the given patterns selects at least one `ObjectName`, which is consistent with the behaviour for a non-patterned `--wait`.

### DIFF
--- a/core/src/main/scala/kafka/tools/JmxTool.scala
+++ b/core/src/main/scala/kafka/tools/JmxTool.scala
@@ -18,13 +18,14 @@
  */
 package kafka.tools
 
-import java.util.{Date, Objects}
+import java.{lang, util}
+import java.util.{Date}
 import java.text.SimpleDateFormat
+
 import javax.management._
 import javax.management.remote._
 import javax.rmi.ssl.SslRMIClientSocketFactory
-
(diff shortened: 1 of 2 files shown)


tombentley:
@chia7712 @dajac would one of you be able to review this? Unfortunately the `JmxTool` needed significant refactoring to make it more testable, so it's not trivial.

tombentley:
@chia7712 @dajac any chance of a review here, or can you suggest someone who can?

PR #9823, opened 2021-01-05
KAFKA-10903:Optimize producerBatch order performance

huangyiminghappy:
please. see https://issues.apache.org/jira/browse/KAFKA-10903,
as we konw,if we need sort  the producerBatch by sequence,now we use the new batch compare with the first batch in deque, and if the first batch in deque is less than the new batch,we will loop the deque and let the new batch insert to the right position.and if the new batch is more than the last producerBatch in the deque,we just need add the new batchProducer to the last position in the deque,most of time the deque not exists the mixed producerBatch(has sequence and not has the sequence),so i think this optimize make sense,thank you review


ijuma:
Thanks for the PR. Did you see a performance issue as a result of this? Do you have some numbers to share?

huangyiminghappy:
> Thanks for the PR. Did you see a performance issue as a result of this? Do you have some numbers to share?

thank you reply, Sometimes we don't need data to support it. may be we can use common sense to judge,When we insert data into an ordered deque, we will determine whether it is inside or outside the deque,and may better than always loop the deque

guozhangwang:
@huangyiminghappy it seems that your assumption is, for most cases, there will just be one inflight request, and hence the batch to be enqueued would most likely need to be add to the last of the queue since very likely none of the batches in `deque` would have a sequence number, is that right?

Logically I think I agree with your reasoning here, and this optimization makes sense. But I'm wondering how much it is actually optimizing in practice -- looping over the `deque` should usually be very quick since the `deque` should be relatively short, like having single-digit number of batches.

Do you observe in your production environment, that the number of batches in `deque` could be very large and hence looping over it to the end could be very costly?

PR #9825, opened 2021-01-05
KAFKA-10904: There is a misleading log when the replica fetcher thread handles offsets that are out of range

wenbingshen:
There is ambiguity in the replica fetcher thread's log. When the fetcher thread is handling with offset out of range, it needs to try to truncate the log. When the end offset of the follower replica is greater than the log start offset of the leader replica and smaller than the end offset of the leader replica, the follower replica will maintain its own fetch offset.However, such cases are processed together with cases where the follower replica's end offset is smaller than the leader replica's start offset, resulting in ambiguities in the log, where the follower replica's fetch offset is reported to reset to the leader replica's start offset.In fact, it still maintains its own fetch offset, so this WARN log is misleading to the user.It is more accurate to print this WARN log only when follower replica really need to truncate the fetch offset to the leader replica's log start offset.

comment to ping reviewers.*

### DIFF
--- a/core/src/main/scala/kafka/server/AbstractFetcherThread.scala
+++ b/core/src/main/scala/kafka/server/AbstractFetcherThread.scala
@@ -675,12 +675,18 @@ abstract class AbstractFetcherThread(name: String,
        * and the current leader's log start offset.
        */
       val leaderStartOffset = fetchEarliestOffsetFromLeader(topicPartition, currentLeaderEpoch)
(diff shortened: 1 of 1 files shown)


wenbingshen:
@chia7712  Can you take a look at this pr?Thanks.

chia7712:
fixed by https://github.com/apache/kafka/commit/88320a81404df313bbdf4443e0598e4fd1b62615

PR #9826, opened 2021-01-05
KAFKA-10816: Initialize REST endpoints only after the herder has started

tombentley:
Do this by making DistributedHerder#start() block until the herder has
really started, thus preventing the initialization of the REST resources until
that point.

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
@@ -78,7 +78,9 @@ import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingDeque;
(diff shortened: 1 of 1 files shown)


tombentley:
@rhauch any chance you could take a look at this, since you opened [KAFKA-10816](https://issues.apache.org/jira/browse/KAFKA-10816)?

tombentley:
@rhauch please could you review? Or @kkonstantine perhaps?

tombentley:
@kkonstantine were you going to review this? It would be much appreciated if you could.

tombentley:
@rhauch @kkonstantine @C0urante PTAL

C0urante:
Thanks for taking this on, Tom!

This seems fairly involved and I'm wondering if there's a simpler solution. The changes here cause `DistributedHerder::start` to block on the completion of `DistributedHerder::startServices`, which is invoked on a separate thread. If we want to retain this blocking behavior, could we move the invocation of `DistributedHerder::startServices` into the `DistributedHerder::start` method? I'm thinking something like this:

```java
public class DistributedHerder extends AbstractHerder {
  @Override
  public void start() {
    log.info("Herder starting");
    startServices();
    log.info("Herder started");

    this.herderExecutor.submit(this);
  }
}
```

Also, I haven't looked at standalone mode yet, but I'm hoping that any solution we arrive at can work for both it and distributed mode in order to avoid gotchas due to unexpected differences in behavior between the two.

C0urante:
@tombentley any plans to revisit this?

C0urante:
I've taken a slightly different approach in https://github.com/apache/kafka/pull/10520, which modifies the root resource to be viable as both a startup and a readiness probe. We can consider it either instead of or in addition to this change, as they serve slightly different purposes and should be mutually compatible.

smallYellowCat:
StandaloneHerder class also need this change.

C0urante:
Better late than never! https://cwiki.apache.org/confluence/display/KAFKA/KIP-1017%3A+Health+check+endpoint+for+Kafka+Connect

PR #9832, opened 2021-01-06
KAFKA-12152; Idempotent Producer does not reset the sequence number of partitions without in-flight batches

dajac:
When a `OutOfOrderSequenceException` error is received by an idempotent producer for a partition, the producer bumps its epoch, adjusts the sequence number and the epoch of the in-flight batches of the partitions affected by the `OutOfOrderSequenceException` error. This happens in `TransactionManager#bumpIdempotentProducerEpoch`.

The remaining partitions are treated separately. When the last in-flight batch of a given partition is completed, the sequence number is reset. This happens in `TransactionManager#handleCompletedBatch`.

However, when a given partition does not have in-flight batches when the producer epoch is bumped, its sequence number is not reset. It results in having subsequent producer request to use the new producer epoch with the old sequence number and to be rejected by the broker.

This PR adds logic to reset the sequences of those partitions.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/producer/internals/RecordAccumulator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/internals/RecordAccumulator.java
@@ -508,15 +508,11 @@ public final class RecordAccumulator {
                 return true;
 
             if (!first.hasSequence()) {
-                if (transactionManager.hasInflightBatches(tp)) {
+                if (transactionManager.hasInflightBatches(tp) && transactionManager.hasStaleProducerIdAndEpoch(tp)) {
                     // Don't drain any new batches while the partition has in-flight batches with a different epoch
                     // and/or producer ID. Otherwise, a batch with a new epoch and sequence number
                     // 0 could be written before earlier batches complete, which would cause out of sequence errors
-                    ProducerBatch firstInFlightBatch = transactionManager.nextBatchBySequence(tp);
-
-                    if (firstInFlightBatch != null && transactionManager.producerIdOrEpochNotMatch(firstInFlightBatch)) {
-                        return true;
-                    }
+                    return true;
                 }
 
                 if (transactionManager.hasUnresolvedSequence(first.topicPartition))
@@ -582,6 +578,12 @@ public final class RecordAccumulator {
                         transactionManager != null ? transactionManager.producerIdAndEpoch() : null;
                     ProducerBatch batch = deque.pollFirst();
                     if (producerIdAndEpoch != null && !batch.hasSequence()) {
+                        // The the producer id/epoch of the partition do not match the latest one
+                        // of the producer, we update it and reset the sequence. This should be
+                        // only done when all its in-flight batches have completed. This is guarantee
+                        // in `shouldStopDrainBatchesForPartition`.
+                        transactionManager.maybeUpdateProducerIdAndEpoch(batch.topicPartition);
+
                         // If the batch already has an assigned sequence, then we should not change the producer id and
                         // sequence number, since this may introduce duplicates. In particular, the previous attempt
                         // may actually have been accepted, and if we change the producer id and sequence here, this
--- a/clients/src/main/java/org/apache/kafka/clients/producer/internals/RecordAccumulator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/internals/RecordAccumulator.java
@@ -578,7 +578,7 @@ public final class RecordAccumulator {
                         transactionManager != null ? transactionManager.producerIdAndEpoch() : null;
                     ProducerBatch batch = deque.pollFirst();
                     if (producerIdAndEpoch != null && !batch.hasSequence()) {
-                        // The the producer id/epoch of the partition do not match the latest one
+                        // If the the producer id/epoch of the partition do not match the latest one
                         // of the producer, we update it and reset the sequence. This should be
                         // only done when all its in-flight batches have completed. This is guarantee
                         // in `shouldStopDrainBatchesForPartition`.
--- a/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
@@ -119,6 +119,10 @@ public class TransactionManager {
             return topicPartitions.containsKey(topicPartition);
         }
 
+        private Set<TopicPartition> partitions() {
+            return topicPartitions.keySet();
+        }
+
         private void reset() {
             topicPartitions.clear();
         }
@@ -567,8 +571,18 @@ public class TransactionManager {
             this.topicPartitionBookkeeper.startSequencesAtBeginning(topicPartition, this.producerIdAndEpoch);
             this.partitionsWithUnresolvedSequences.remove(topicPartition);
         }
-
         this.partitionsToRewriteSequences.clear();
+
+        // When the epoch is bumped, reset the sequences for the partitions(s) that do not
+        // have in-flight batches. Sequences of the in-flight batches that did not triggered
+        // the epoch bump are treated when their last in-flight batch is completed.
+        for (TopicPartition topicPartition : this.topicPartitionBookkeeper.partitions()) {
+            TopicPartitionEntry topicPartitionEntry = this.topicPartitionBookkeeper.getPartition(topicPartition);
+            if (topicPartitionEntry.inflightBatchesBySequence.isEmpty()) {
+                this.topicPartitionBookkeeper.startSequencesAtBeginning(topicPartition, this.producerIdAndEpoch);
+            }
+        }
+
         epochBumpRequired = false;
     }
 
--- a/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
+++ b/clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
@@ -119,10 +119,6 @@ public class TransactionManager {
             return topicPartitions.containsKey(topicPartition);
         }
 
-        private Set<TopicPartition> partitions() {
-            return topicPartitions.keySet();
-        }
-
         private void reset() {
             topicPartitions.clear();
         }
@@ -150,12 +146,16 @@ public class TransactionManager {
                 inFlightBatch.resetProducerState(newProducerIdAndEpoch, sequence.value, inFlightBatch.isTransactional());
                 sequence.value += inFlightBatch.recordCount;
             });
+            topicPartitionEntry.producerIdAndEpoch = newProducerIdAndEpoch;
             topicPartitionEntry.nextSequence = sequence.value;
             topicPartitionEntry.lastAckedSequence = NO_LAST_ACKED_SEQUENCE_NUMBER;
         }
     }
 
     private static class TopicPartitionEntry {
+
+        // The producer id/epoch being used for a given partition.
+        private ProducerIdAndEpoch producerIdAndEpoch;
 
         // The base sequence of the next batch bound for a given partition.
         private int nextSequence;
@@ -175,6 +175,7 @@ public class TransactionManager {
         private long lastAckedOffset;
 
         TopicPartitionEntry() {
+            this.producerIdAndEpoch = ProducerIdAndEpoch.NONE;
             this.nextSequence = 0;
             this.lastAckedSequence = NO_LAST_ACKED_SEQUENCE_NUMBER;
             this.lastAckedOffset = ProduceResponse.INVALID_OFFSET;
@@ -517,8 +518,23 @@ public class TransactionManager {
     }
 
     boolean producerIdOrEpochNotMatch(ProducerBatch batch) {
-        ProducerIdAndEpoch idAndEpoch = this.producerIdAndEpoch;
+        // Verifies that the producer id/epoch of the batch matches the current one
+        // of the partition.
+        ProducerIdAndEpoch idAndEpoch = topicPartitionBookkeeper.getPartition(batch.topicPartition).producerIdAndEpoch;
         return idAndEpoch.producerId != batch.producerId() || idAndEpoch.epoch != batch.producerEpoch();
+    }
+
+    synchronized public void maybeUpdateProducerIdAndEpoch(TopicPartition topicPartition) {
+        if (!isTransactional())
+            topicPartitionBookkeeper.addPartition(topicPartition);
+
+        if (hasStaleProducerIdAndEpoch(topicPartition) && !hasInflightBatches(topicPartition)) {
+            // If the batch was on a different ID and/or epoch (due to an epoch bump) and all its in-flight batches
+            // have completed, reset the partition sequence so that the next batch (with the new epoch) starts from 0
+            topicPartitionBookkeeper.startSequencesAtBeginning(topicPartition, this.producerIdAndEpoch);
+            log.debug("ProducerId of partition {} set to {} with epoch {}. Reinitialize sequence at beginning.",
+                      topicPartition, producerIdAndEpoch.producerId, producerIdAndEpoch.epoch);
+        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


dajac:
@hachikuji Could you take a look at this one please? I am not familiar with this part of the code so there may be a better way to address the issue. Let me know what you think.

dajac:
`org.apache.kafka.clients.producer.internals.TransactionManagerTest.testProducerIdReset` failed due to the changes. I will update it later once I get the confirmation that the approach and my understanding are correct.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -567,8 +571,18 @@ private void bumpIdempotentProducerEpoch() {
             this.topicPartitionBookkeeper.startSequencesAtBeginning(topicPartition, this.producerIdAndEpoch);
             this.partitionsWithUnresolvedSequences.remove(topicPartition);
         }
-
         this.partitionsToRewriteSequences.clear();
+
+        // When the epoch is bumped, reset the sequences for the partitions(s) that do not
+        // have in-flight batches. Sequences of the in-flight batches that did not triggered
+        // the epoch bump are treated when their last in-flight batch is completed.
+        for (TopicPartition topicPartition : this.topicPartitionBookkeeper.partitions()) {
```
hachikuji: Apologies for the lengthy comment, but it always takes me some time to wrap my head around the many cases that are possible. I'm trying to convince myself whether it is necessary to be selective or if we can reset the sequences for all partitions blindly. Suppose that we are producing to 3 topic partitions and we have the following local state:

```
epoch: 5
------------
partition: 0
last acked seq: 10
next seq: 11
inflight seqs: []
------------
partition: 1
last acked seq: 20
next seq: 23
inflight seqs: [21, 22]
------------
partition: 2
last acked seq: 30
next seq: 33
inflight seqs: [31, 32]
```

So we have one partition (0) with no inflights and two partitions (1 and 2) which have two batches in-flight each. Suppose that the request with seq 31 for partition 2 returns with an out of sequence error, which causes a call to `requestEpochBumpForPartition`. This in turn causes a call to `bumpIdempotentProducerEpoch`. I think it's clear that we want to reset the sequence number for partition 0 as you are doing here. So we end up with the following state for partition 0:
```
partition: 0
last acked seq: -1
next seq: 0
inflight seqs: []
```

For partition 2, the current logic assumes that seq 32 is doomed to fail since 31 has already failed, so we will reset the sequence numbers. In `TransactionManager.canRetry`, we have logic to check that the sequence number has been reset after an out of sequence error and to allow the retry. This leaves us with the following state:
```
partition: 2
last acked seq: -1
next seq: 2
inflight seqs: [0, 1]
```

The challenge for partition 1 is that we do not want to alter the sequence numbers of inflight requests unless we have to since that could introduce duplicates if they ultimately need to be retried. As far as I can tell, the following cases are possible:

Case 1: inflight sequence numbers 21 and 22 return successfully

We have the following check in `handleCompletedBatch`:
```java
        if (producerIdOrEpochNotMatch(batch) && !hasInflightBatches(batch.topicPartition)) {
            // If the batch was on a different ID and/or epoch (due to an epoch bump) and all its in-flight batches
            // have completed, reset the partition sequence so that the next batch (with the new epoch) starts from 0
            topicPartitionBookkeeper.startSequencesAtBeginning(batch.topicPartition, this.producerIdAndEpoch);
        }
```
So basically we wait until all of the batches have completed, and then reset the next sequence. This leaves us with this state:
```
partition: 1
last acked seq: -1
next seq: 0
inflight seqs: []
```

Case 2: one of the inflight batches fails

This case is harder to track down because of the logic in `canRetry`, but suppose that we eventually reach `handleFailedBatch`. I think we will hit the following case:
```java
        if (producerIdOrEpochNotMatch(batch)) {
            log.debug("Ignoring failed batch {} with producer id {}, epoch {}, and sequence number {} " +
                            "since the producerId has been reset internally", batch, batch.producerId(),
                    batch.producerEpoch(), batch.baseSequence(), exception);
            return;
        }
```
Because the epoch from the batch does not match, we just discard it. Unless I am missing something, that leaves us in the following state:
```
partition: 1
last acked seq: 20
next seq: 23
inflight seqs: []
```
Then the next produce request to this partition is likely to get an out of sequence error because of a similar problem to the one you found. (Please check this over and see if you reach the same conclusion.)


So coming back to the original question, I am trying to think how to simplify the logic here and fix some of the unhandled cases. It's very difficult to reason about the state when we allow inflight requests from different epochs, but I don't see a way around it. Here is an idea how we can improve this logic:

1. I think we need to track the producer epoch and producer id inside each `TopicPartitionEntry`. Our lives will be a lot easier if we can ensure that `nextSequence`, `lastAckedSequence`, and `inflightBatchesBySequence` are all consistently aligned on the same producerId/epoch and we have an easy way to know what the producerId/epoch are. 
2. As the logic does now, we want to drain inflight requests before we begin using the bumped producerId/epoch. This is used to enforce the invariant in 1.
3. If we hit an out of sequence error, we can basically check whether the producerId/epoch in `TopicPartitionEntry` match the latest one in use by the producer. If they do, then we request a bump and reset sequence numbers. If not, then we can use the latest producerId/epoch and reset sequence numbers.
4. When we attempt to increment the sequence number when draining batches in `RecordAccumulator.drainBatchesForOneNode`, we can check whether the producerId/epoch in `TopicPartitionEntry` matches the latest value. If it does, then we proceed normally. If it does not, then we check inflight requests. If there are still some remaining, then we cannot drain. If there are no inflights, then we can update producer state.

This probably needs a bit of refinement, but hopefully the core idea is clear. Basically we add producerId/epoch to `TopicPartitionEntry` and we keep the state consistently aligned. Then we lazily update to the latest state once all the inflights have cleared.

hachikuji (commented):
@dajac Thanks for the investigation and the patch. I think you have uncovered a weakness in the current bookkeeping that might call for deeper consideration. I left a comment to that effect.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
dajac: Thanks for your comment. I do agree with your conclusion about point 2. I have managed to reproduce this in a unit test as well.

Regarding your suggestion to store the producer id/epoch per partition, I like it. I actually thought about this as well but went conservative for my first draft. I have update the PR to reflect your suggestion. Please take a look and let me know what you think.

I think that there are still room for improvement in the PR. Especially, I have to review the unit tests. Some might not be entirely relevant with the changes done in this PR or some might need to be refactored a bit more. I am not sure.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -598,6 +618,16 @@ synchronized Integer sequenceNumber(TopicPartition topicPartition) {
         return topicPartitionBookkeeper.getPartition(topicPartition).nextSequence;
     }
 
+    /**
+     * Returns the current producer id/epoch of the given TopicPartition.
+     */
+    synchronized ProducerIdAndEpoch producerIdAndEpoch(TopicPartition topicPartition) {
+        if (!isTransactional())
+            topicPartitionBookkeeper.addPartition(topicPartition);
```
dajac: I wonder if we could create the partition state once for all instead of having to do this here and in other places. I need to find where we could do so.

clients/src/main/java/org/apache/kafka/clients/producer/internals/RecordAccumulator.java
```diff
@@ -582,6 +578,12 @@ private boolean shouldStopDrainBatchesForPartition(ProducerBatch first, TopicPar
                         transactionManager != null ? transactionManager.producerIdAndEpoch() : null;
                     ProducerBatch batch = deque.pollFirst();
                     if (producerIdAndEpoch != null && !batch.hasSequence()) {
+                        // The the producer id/epoch of the partition do not match the latest one
```
hachikuji: nit: "The the" -> "If the"?

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -513,10 +518,25 @@ ProducerIdAndEpoch producerIdAndEpoch() {
     }
 
     boolean producerIdOrEpochNotMatch(ProducerBatch batch) {
-        ProducerIdAndEpoch idAndEpoch = this.producerIdAndEpoch;
+        // Verifies that the producer id/epoch of the batch matches the current one
+        // of the partition.
+        ProducerIdAndEpoch idAndEpoch = topicPartitionBookkeeper.getPartition(batch.topicPartition).producerIdAndEpoch;
```
hachikuji: It looks like `producerIdOrEpochNotMatch` has one caller in `handleFailedBatch`. Now that we are keeping the producerId/epoch consistent with the inflight batches, I guess this check is more of a state validation than an essential check. I'm wondering if we can either remove the check and rely on our internal invariant or change it from logging a debug message to raising an IllegalState error.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
hachikuji: I agree that it convolutes the code to need to check the map all over the place. I guess we could add a `getOrCreate` method or something like that.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
```diff
@@ -513,10 +518,25 @@ ProducerIdAndEpoch producerIdAndEpoch() {
...
         return idAndEpoch.producerId != batch.producerId() || idAndEpoch.epoch != batch.producerEpoch();
     }
 
+    synchronized public void maybeUpdateProducerIdAndEpoch(TopicPartition topicPartition) {
+        if (!isTransactional())
+            topicPartitionBookkeeper.addPartition(topicPartition);
+
+        if (hasStaleProducerIdAndEpoch(topicPartition) && !hasInflightBatches(topicPartition)) {
```
hachikuji, line 530: One option I was discussing with @bob-barrett is whether to continue using the old epoch. The benefit is that we don't need to drain all partitions because of a sequence error on a single partition. I'm not sure if there are any major downsides other than possibly making debugging more confusing. I guess for the transactional use case, it might be better to stick with the latest epoch when possible.

hachikuji (commented):
Thanks for the updates. I left some small comments, but I feel like we're on the right track.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
dajac: Good point. I think that we can just remove it. It does not seem to bring much.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
dajac: Interesting point. My understanding is that we can't do this for the transactional use case so it might be better to stick with the latest epoch whenever possible as you said. However with [KIP-679](https://cwiki.apache.org/confluence/display/KAFKA/KIP-679%3A+Producer+will+enable+the+strongest+delivery+guarantee+by+default) on the way, this may be a nice optimization for the idempotent producer as, I suppose, an out of sequence for one partition would not affect the latency/throughput of the other partitions at all. If out of sequence errors are frequent, this may be worth. Do we believe that they are?

I suggest to tackle this in a separate PR anyway and keep this PR focused on fixing the identified issue.

dajac:
@hachikuji Thanks for your feedback. I have updated the PR to address your points.

clients/src/main/java/org/apache/kafka/clients/producer/internals/TransactionManager.java
bob-barrett: Yeah, I agree that this can be an optimization in a future PR

clients/src/test/java/org/apache/kafka/clients/producer/internals/SenderTest.java
```diff
@@ -934,6 +935,223 @@ public void testEpochBumpOnOutOfOrderSequenceForNextBatch() throws Exception {
...
+        sender.runOnce();
+
+        // Partition 0 - Last ack is updated
+        assertPartitionState(transactionManager, tp0, producerId, (short) 1, 1, OptionalInt.of(0));
+
+        // Partition 1 - Send second batch
+        appendToAccumulator(tp1);
+        sender.runOnce();
+
+        // Partition 1 - Epoch is bump, sequence is reset and incremented
```
bob-barrett: nit: `bump` -> `bumped`

bob-barrett (commented):
Thanks for the patch David. I think everything looks good to me.

One thing I was trying to think through was whether the new logic for tracking the epochs will have any problems with transactional epoch bumps, because when we handle an `InitProducerId` response that bumps the epoch, we clear all the `TopicPartitionBookkeeper` partitions away. It looks like it's not a problem; we stop draining if a partition isn't in the transaction, and anyway every time we mutate the partition state as part of increasing the epoch, we check that it exists first.

dajac:
@bob-barrett Thanks for your feedback.

hachikuji:
> One thing I was trying to think through was whether the new logic for tracking the epochs will have any problems with transactional epoch bumps, because when we handle an InitProducerId response that bumps the epoch, we clear all the TopicPartitionBookkeeper partitions away. It looks like it's not a problem; we stop draining if a partition isn't in the transaction, and anyway every time we mutate the partition state as part of increasing the epoch, we check that it exists first.

@bob-barrett Good point. Also, the `InitProducerIdRequest` will only be sent once all unsent or inflight batches have been cleared from `RecordAccumulator`.

clients/src/test/java/org/apache/kafka/clients/producer/internals/SenderTest.java
```diff
@@ -934,6 +935,223 @@ public void testEpochBumpOnOutOfOrderSequenceForNextBatch() throws Exception {
...
+
+        // Partition 0 - Successful Response
+        sendIdempotentProducerResponse(1, 0, tp0, Errors.NONE, 1, -1);
+        sender.runOnce();
+
+        // Partition 0 - Last ack is updated
+        assertPartitionState(transactionManager, tp0, producerId, (short) 1, 1, OptionalInt.of(0));
+
+        // Partition 1 - Failed response with NOT_LEADER_OR_FOLLOWER
+        sendIdempotentProducerResponse(0, 1, tp1, Errors.NOT_LEADER_OR_FOLLOWER, -1, -1);
```
hachikuji, line 1113: Is it worthwhile to assert that `hasStaleProducerIdAndEpoch` returns true for this partition until the inflights are resolved and we have sent the next batch?

hachikuji (approved):
LGTM. Left a minor suggestion on the test case.

clients/src/test/java/org/apache/kafka/clients/producer/internals/SenderTest.java
dajac: That makes sense. I've added those assertions.

dajac:
Rebased the PR to ensure that it has all recent the junit changes.

dajac:
Failed tests in the last build are not related to the changes in this PR:
* Build / JDK 8 / org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationSSLTest.testReplication()
* Build / JDK 8 / org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationTest.testReplication()
* Build / JDK 8 / kafka.integration.MetricsDuringTopicCreationDeletionTest.testMetricsDuringTopicCreateDelete()
* Build / JDK 11 / kafka.api.CustomQuotaCallbackTest.testCustomQuotaCallback()

I will go ahead and merge the PR in trunk and 2.7.

PR #9835, opened 2021-01-06
MINOR: Add 'task container' class to KafkaStreams TaskManager

mjsax:
Kafka Streams' TaskManager is a central class that grew quite big. This
PR breaks out a new 'task container' class to descope what TaskManager
does. In follow up PRs, we plan to move more methods from TaskManager
to the new 'Tasks.java' class and also improve task-type type safety.

Call for review @ableegoldman @vvcephei @guozhangwang

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
@@ -626,7 +626,7 @@ public class TaskManager {
         // so we consider all tasks that are either owned or on disk. This includes stateless tasks, which should
         // just have an empty changelogOffsets map.
         for (final TaskId id : union(HashSet::new, lockedTaskDirectories, tasks.tasksPerId().keySet())) {
-            final Task task = tasks.task(id);
+            final Task task = tasks.owned(id) ? tasks.task(id) : null;
             // Closed and uninitialized tasks don't have any offsets so we should read directly from the checkpoint
             if (task != null && task.state() != State.CREATED && task.state() != State.CLOSED) {
                 final Map<TopicPartition, Long> changelogOffsets = task.changelogOffsets();
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
@@ -37,6 +37,7 @@ import org.apache.kafka.streams.errors.TaskMigratedException;
 import org.apache.kafka.streams.errors.TaskTimeoutExceptions;
 import org.apache.kafka.streams.processor.TaskId;
 import org.apache.kafka.streams.processor.internals.Task.State;
+import org.apache.kafka.streams.processor.internals.metrics.StreamsMetricsImpl;
 import org.apache.kafka.streams.state.internals.OffsetCheckpoint;
 import org.slf4j.Logger;
 
@@ -54,7 +55,6 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicReference;
@@ -75,16 +75,11 @@ public class TaskManager {
     private final ChangelogReader changelogReader;
     private final UUID processId;
     private final String logPrefix;
-    private final ActiveTaskCreator activeTaskCreator;
-    private final StandbyTaskCreator standbyTaskCreator;
     private final InternalTopologyBuilder builder;
     private final Admin adminClient;
     private final StateDirectory stateDirectory;
     private final StreamThread.ProcessingMode processingMode;
-
-    private final Map<TaskId, Task> tasks = new TreeMap<>();
-    // materializing this relationship because the lookup is on the hot path
-    private final Map<TopicPartition, Task> partitionToTask = new HashMap<>();
+    private final Tasks tasks;
 
     private Consumer<byte[], byte[]> mainConsumer;
 
@@ -100,6 +95,7 @@ public class TaskManager {
                 final ChangelogReader changelogReader,
                 final UUID processId,
                 final String logPrefix,
+                final StreamsMetricsImpl streamsMetrics,
                 final ActiveTaskCreator activeTaskCreator,
                 final StandbyTaskCreator standbyTaskCreator,
                 final InternalTopologyBuilder builder,
@@ -110,12 +106,11 @@ public class TaskManager {
         this.changelogReader = changelogReader;
         this.processId = processId;
         this.logPrefix = logPrefix;
-        this.activeTaskCreator = activeTaskCreator;
-        this.standbyTaskCreator = standbyTaskCreator;
         this.builder = builder;
         this.adminClient = adminClient;
         this.stateDirectory = stateDirectory;
         this.processingMode = processingMode;
+        this.tasks = new Tasks(logPrefix, builder,  streamsMetrics, activeTaskCreator, standbyTaskCreator);
 
         final LogContext logContext = new LogContext(logPrefix);
         log = logContext.logger(getClass());
@@ -123,6 +118,7 @@ public class TaskManager {
 
     void setMainConsumer(final Consumer<byte[], byte[]> mainConsumer) {
         this.mainConsumer = mainConsumer;
+        tasks.setMainConsumer(mainConsumer);
     }
 
     public UUID processId() {
@@ -155,13 +151,13 @@ public class TaskManager {
         rebalanceInProgress = false;
     }
 
-    void handleCorruption(final Map<TaskId, Collection<TopicPartition>> tasksWithChangelogs) throws TaskMigratedException {
+    void handleCorruption(final Map<TaskId, Collection<TopicPartition>> tasksWithChangelogs) {
         final Map<Task, Collection<TopicPartition>> corruptedStandbyTasks = new HashMap<>();
         final Map<Task, Collection<TopicPartition>> corruptedActiveTasks = new HashMap<>();
 
         for (final Map.Entry<TaskId, Collection<TopicPartition>> taskEntry : tasksWithChangelogs.entrySet()) {
             final TaskId taskId = taskEntry.getKey();
-            final Task task = tasks.get(taskId);
+            final Task task = tasks.task(taskId);
             if (task.isActive()) {
                 corruptedActiveTasks.put(task, taskEntry.getValue());
             } else {
@@ -212,7 +208,7 @@ public class TaskManager {
 
             // For active tasks pause their input partitions so we won't poll any more records
             // for this task until it has been re-initialized;
-            // Note, closeDirty already clears the partitiongroup for the task.
+            // Note, closeDirty already clears the partition-group for the task.
             if (task.isActive()) {
                 final Set<TopicPartition> currentAssignment = mainConsumer.assignment();
                 final Set<TopicPartition> taskInputPartitions = task.inputPartitions();
@@ -273,12 +269,12 @@ public class TaskManager {
         final Set<Task> tasksToCloseDirty = new TreeSet<>(byId);
 
         // first rectify all existing tasks
-        for (final Task task : tasks.values()) {
+        for (final Task task : tasks.allTasks()) {
             if (activeTasks.containsKey(task.id()) && task.isActive()) {
-                updateInputPartitionsAndResume(task, activeTasks.get(task.id()));
+                tasks.updateInputPartitionsAndResume(task, activeTasks.get(task.id()));
                 activeTasksToCreate.remove(task.id());
             } else if (standbyTasks.containsKey(task.id()) && !task.isActive()) {
-                updateInputPartitionsAndResume(task, standbyTasks.get(task.id()));
+                tasks.updateInputPartitionsAndResume(task, standbyTasks.get(task.id()));
                 standbyTasksToCreate.remove(task.id());
             } else if (activeTasks.containsKey(task.id()) || standbyTasks.containsKey(task.id())) {
                 // check for tasks that were owned previously but have changed active/standby status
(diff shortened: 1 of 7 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -75,16 +75,11 @@
...
-    private final StandbyTaskCreator standbyTaskCreator;
     private final InternalTopologyBuilder builder;
     private final Admin adminClient;
     private final StateDirectory stateDirectory;
     private final StreamThread.ProcessingMode processingMode;
-
-    private final Map<TaskId, Task> tasks = new TreeMap<>();
-    // materializing this relationship because the lookup is on the hot path
-    private final Map<TopicPartition, Task> partitionToTask = new HashMap<>();
+    private final Tasks tasks;
```
mjsax, line 82: The new `Tasks` class is the container -- we move the task-creators and task-maps into this container class.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -155,13 +151,13 @@ void handleRebalanceComplete() {
...
 
-    void handleCorruption(final Map<TaskId, Collection<TopicPartition>> tasksWithChangelogs) throws TaskMigratedException {
+    void handleCorruption(final Map<TaskId, Collection<TopicPartition>> tasksWithChangelogs) {
         final Map<Task, Collection<TopicPartition>> corruptedStandbyTasks = new HashMap<>();
         final Map<Task, Collection<TopicPartition>> corruptedActiveTasks = new HashMap<>();
 
         for (final Map.Entry<TaskId, Collection<TopicPartition>> taskEntry : tasksWithChangelogs.entrySet()) {
             final TaskId taskId = taskEntry.getKey();
-            final Task task = tasks.get(taskId);
+            final Task task = tasks.task(taskId);
```
mjsax, line 163: The old `tasks` variable is the `Map`; the now one is the `Tasks` container class.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -377,8 +370,9 @@ private void handleCloseAndRecycle(final Set<Task> tasksToRecycle,
         for (final Task task : tasksToCloseClean) {
             try {
                 completeTaskCloseClean(task);
-                cleanUpTaskProducer(task, taskCloseExceptions);
-                tasks.remove(task.id());
+                if (task.isActive()) {
```
mjsax, line 376: The old `cleanUpTaskProducer` contained this check, while the new `cleanUpTaskProducerAndRemoveTask` does not --  maybe we could move the check into the new method, but was not sure about it.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -390,71 +384,28 @@ private void handleCloseAndRecycle(final Set<Task> tasksToRecycle,
...
+        for (final Task oldTask : tasksToRecycle) {
             final Task newTask;
             try {
-                if (task.isActive()) {
-                    final Set<TopicPartition> partitions = standbyTasksToCreate.remove(task.id());
-                    newTask = standbyTaskCreator.createStandbyTaskFromActive((StreamTask) task, partitions);
-                    cleanUpTaskProducer(task, taskCloseExceptions);
+                if (oldTask.isActive()) {
+                    final Set<TopicPartition> partitions = standbyTasksToCreate.remove(oldTask.id());
+                    tasks.convertActiveToStandby((StreamTask) oldTask, partitions, taskCloseExceptions);
```
mjsax, line 395: Renamed old `createStandbyTaskFromActive` to `convertActiveToStandby` (same below for reverse conversion)

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -390,71 +384,28 @@ private void handleCloseAndRecycle(final Set<Task> tasksToRecycle,
...
 
         // for tasks that cannot be cleanly closed or recycled, close them dirty
         for (final Task task : tasksToCloseDirty) {
             closeTaskDirty(task);
-            cleanUpTaskProducer(task, taskCloseExceptions);
-            tasks.remove(task.id());
-        }
-    }
-
-    private void cleanUpTaskProducer(final Task task,
```
mjsax, line 422: Moved those method to the new `Tasks` container class to descope what `TaskManager` does -- will move more methods in a follow up PR. (Just tried to keep this PR small).

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -804,45 +749,42 @@ private void closeTaskDirty(final Task task) {
...
-            partitionToTask.remove(inputPartition);
-        }
-    }
-
     void shutdown(final boolean clean) {
         final AtomicReference<RuntimeException> firstException = new AtomicReference<>(null);
 
         final Set<Task> tasksToCloseDirty = new HashSet<>();
+        // TODO: change type to `StreamTask`
+        final Set<Task> activeTasks = new TreeSet<>(Comparator.comparing(Task::id));
```
mjsax, line 769: We need to get a deep-copy of all active tasks here, because `tryCloseCleanAllActiveTasks` would remove tasks from the task container but we need them to iterate over them later to close the task producers.

streams/src/main/java/org/apache/kafka/streams/processor/internals/Tasks.java
```diff
@@ -0,0 +1,295 @@
...
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+
+class Tasks {
+    private final Logger log;
+    private final InternalTopologyBuilder builder;
+    private final StreamsMetricsImpl streamsMetrics;
+
+    private final Map<TaskId, Task> allTasksPerId = new TreeMap<>();
```
mjsax, line 41: We replace the old `Map<TaskId, Task>` with redundant data structures and read-only views. This will allow us to improve type safety later. We also don't need code to extract the views on demand, but the collections will be kept in sync automatically for us.

To keep this PR small, I did not implement the changes to get actual type safety yet.

streams/src/test/java/org/apache/kafka/streams/processor/internals/TaskManagerTest.java
```diff
@@ -336,6 +340,7 @@ public void shouldComputeOffsetSumForUnassignedTaskWeCanLock() throws Exception
 
         replay(stateDirectory);
         taskManager.handleRebalanceStart(singleton("topic"));
+        taskManager.addTask(new StateMachineTask(taskId00, Collections.emptySet(), true));
```
mjsax: We have stricter check if a tasks exists now, so we need to add a task to avoid that `getTaskOffsetSums` fails with "unknown task exception" (similar below)

streams/src/test/java/org/apache/kafka/streams/processor/internals/TaskManagerTest.java
```diff
@@ -1514,8 +1521,6 @@ public void shouldOnlyCommitRevokedStandbyTaskAndPropagatePrepareCommitException
 
     @Test
     public void shouldSuspendAllRevokedActiveTasksAndPropagateSuspendException() {
-        setUpTaskManager(StreamThread.ProcessingMode.EXACTLY_ONCE_ALPHA);
```
mjsax, line 1517: Not sure why this test sets up the TM for EOS? Does anyone know?

streams/src/test/java/org/apache/kafka/streams/processor/internals/TaskManagerTest.java
```diff
@@ -1495,8 +1502,8 @@ public void shouldOnlyCommitRevokedStandbyTaskAndPropagatePrepareCommitException
         };
         task01.setCommitNeeded();
 
-        taskManager.tasks().put(taskId00, task00);
```
mjsax, line 1498: We get a read-only view from `tasks()` now, what protects us from accidentally messing with the task container.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -155,13 +151,13 @@ void handleRebalanceComplete() {
         rebalanceInProgress = false;
     }
 
-    void handleCorruption(final Map<TaskId, Collection<TopicPartition>> tasksWithChangelogs) throws TaskMigratedException {
```
ableegoldman, line 158: Why remove this?

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
ableegoldman: IIUC the direction you want to head in (this is not the final PR, right? Lots of TODOs) is to improve type safety and ultimately have most of the methods in `Tasks` accept specific StreamTask/StandbyTask, vs just a generic "Task". Is that correct? In that case it feels correct to leave it here.
But if the ultimate goal is to make the TaskManager _more_ task-type agnostic, then maybe it should be in Tasks. How would you define the division of labor between TaskManager and Tasks?

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
ableegoldman: Much better name :)

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
ableegoldman: Would it be cleaner to just close/remove the task producer inside `tryCloseCleanAllActiveTasks`? Could we just close/remove the task producers before closing clean or do we specifically need to close them first? I don't remember...

streams/src/test/java/org/apache/kafka/streams/processor/internals/TaskManagerTest.java
ableegoldman: Doesn't this invalidate the test though? We're specifically testing whether we can compute offset sums for tasks that we _don't_ own

streams/src/test/java/org/apache/kafka/streams/processor/internals/TaskManagerTest.java
ableegoldman: cc @abbccdda

ableegoldman (commented):
Awesome. Can't wait for the next PR

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
mjsax: My understanding is, that we don't want to declare unchecked exception, but only checked exceptions (for which is mandatory). (Not my personal style, but most people seem to prefer it this way.)

If we think it's important, we should rather add a JavaDocs comment.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
mjsax: Your understanding is correct. TaskManager should know if it's dealing with an active or standby -- in fact, I want to remove some method like `addRecords()` from `Task` interface and move it to `StreamTask`, too.

And yes, this is just a first PR to keep the scope small.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
mjsax: Me neither. :) I actually wanted to look into this in more detail in a follow up PR and just added the deep-copy to get this PR into a "correct" state to make the tests pass. If you insist, I can try to update this PR for a proper cleanup, but I am not sure what the scope will be...

streams/src/test/java/org/apache/kafka/streams/processor/internals/TaskManagerTest.java
mjsax: Oh. Good point. I guess I did not pay enough attention to what this test actually verifies... _urgs_

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
ableegoldman: No worries, let's keep the scope small for now. Just wanted to raise the question

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
ableegoldman: I guess maybe we've had some warring styles at play for a while now. Let's just add a JavaDocs comment then

ableegoldman (approved):
LGTM, thanks for the cleanup

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
mjsax: Very fair question. Glad that we can follow up on it later.

mjsax:
Thanks @ableegoldman -- Added the JavaDocs. Will merge after Jenkins passed.

streams/src/main/java/org/apache/kafka/streams/processor/internals/Tasks.java
```diff
@@ -0,0 +1,295 @@
...
+
+    void convertActiveToStandby(final StreamTask activeTask,
+                                final Set<TopicPartition> partitions,
+                                final Map<TaskId, RuntimeException> taskCloseExceptions) {
+        if (activeTasksPerId.remove(activeTask.id()) == null) {
+            throw new IllegalStateException("Attempted to convert unknown active task to standby task: " + activeTask.id());
+        }
+        activeTasksPerPartition.entrySet().stream()
+            .filter(e -> e.getValue().id().equals(activeTask.id()))
+            .forEach(e -> activeTasksPerPartition.remove(e.getKey()));
```
ableegoldman, line 133: @mjsax this seems to be throwing a ConcurrentModificationException

mjsax: Yes... https://github.com/apache/kafka/pull/9940

PR #9836, opened 2021-01-06
KAFKA-10866: Add metadata to ConsumerRecords

vvcephei:
Expose fetched metadata via the ConsumerRecords
object as described in KIP-695.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1118,6 +1118,7 @@ project(':clients') {
     testCompile libs.bcpkix
     testCompile libs.junitJupiter
     testCompile libs.mockitoCore
+    testCompile libs.hamcrest
 
     testRuntime libs.slf4jlog4j
     testRuntime libs.jacksonDatabind
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -100,7 +100,10 @@
               files="RequestResponseTest.java|FetcherTest.java|KafkaAdminClientTest.java"/>
 
     <suppress checks="NPathComplexity"
-              files="MemoryRecordsTest|MetricsTest|RequestResponseTest|TestSslUtils|AclAuthorizerBenchmark"/>
+              files="MemoryRecordsTest|MetricsTest|RequestResponseTest|TestSslUtils|AclAuthorizerBenchmark|MockConsumer"/>
+
+    <suppress checks="CyclomaticComplexity"
+              files="MockConsumer"/>
 
     <suppress checks="(WhitespaceAround|LocalVariableName|ImportControl|AvoidStarImport)"
               files="Murmur3Test.java"/>
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
@@ -46,13 +46,13 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
 
         private final long receivedTimestamp;
         private final Long position;
-        private final long beginningOffset;
-        private final long endOffset;
+        private final Long beginningOffset;
+        private final Long endOffset;
 
         public Metadata(final long receivedTimestamp,
                         final Long position,
-                        final long beginningOffset,
-                        final long endOffset) {
+                        final Long beginningOffset,
+                        final Long endOffset) {
             this.receivedTimestamp = receivedTimestamp;
             this.position = position;
             this.beginningOffset = beginningOffset;
@@ -76,14 +76,14 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
         /**
          * @return The lag between the next position to fetch and the current end of the partition
          */
-        public long lag() {
-            return endOffset - position;
+        public Long lag() {
+            return endOffset == null || position == null ? null : endOffset - position;
         }
 
         /**
          * @return The current first offset in the partition.
          */
-        public long beginningOffset() {
+        public Long beginningOffset() {
             return beginningOffset;
         }
 
@@ -93,8 +93,18 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
          * replicated offset plus one. Under "read_committed," this is the minimum of the last successfully
          * replicated offset plus one or the smallest offset of any open transaction.
          */
-        public long endOffset() {
+        public Long endOffset() {
             return endOffset;
+        }
+
+        @Override
+        public String toString() {
+            return "Metadata{" +
+                "receivedTimestamp=" + receivedTimestamp +
+                ", position=" + position +
+                ", beginningOffset=" + beginningOffset +
+                ", endOffset=" + endOffset +
+                '}';
         }
     }
 
@@ -217,7 +227,7 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
     }
 
     public boolean isEmpty() {
-        return records.isEmpty();
+        return records.isEmpty() && metadata.isEmpty();
     }
 
     @SuppressWarnings("unchecked")
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
@@ -46,16 +46,13 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
 
         private final long receivedTimestamp;
         private final Long position;
-        private final Long beginningOffset;
         private final Long endOffset;
 
         public Metadata(final long receivedTimestamp,
                         final Long position,
-                        final Long beginningOffset,
                         final Long endOffset) {
             this.receivedTimestamp = receivedTimestamp;
             this.position = position;
-            this.beginningOffset = beginningOffset;
             this.endOffset = endOffset;
         }
 
@@ -82,13 +79,6 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
         }
 
         /**
-         * @return The current first offset in the partition, or if the beginning offset is not known.
-         */
-        public Long beginningOffset() {
-            return beginningOffset;
-        }
-
-        /**
          * @return The current last offset in the partition. The determination of the "last" offset
          * depends on the Consumer's isolation level. Under "read_uncommitted," this is the last successfully
          * replicated offset plus one. Under "read_committed," this is the minimum of the last successfully
@@ -104,7 +94,6 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
             return "Metadata{" +
                 "receivedTimestamp=" + receivedTimestamp +
                 ", position=" + position +
-                ", beginningOffset=" + beginningOffset +
                 ", endOffset=" + endOffset +
                 '}';
         }
@@ -118,7 +107,6 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
                 new Metadata(
                     entry.getValue().receivedTimestamp(),
                     entry.getValue().position() == null ? null : entry.getValue().position().offset,
-                    entry.getValue().beginningOffset(),
                     entry.getValue().endOffset()
                 )
             );
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
@@ -50,7 +50,7 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
         private final long endOffset;
 
         public Metadata(final long receivedTimestamp,
-                        final long position,
+                        final Long position,
                         final long beginningOffset,
                         final long endOffset) {
             this.receivedTimestamp = receivedTimestamp;
@@ -67,9 +67,9 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
         }
 
         /**
-         * @return The next position the consumer will fetch
+         * @return The next position the consumer will fetch, or null if the consumer has no position.
          */
-        public long position() {
+        public Long position() {
             return position;
         }
 
@@ -105,7 +105,7 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
                 entry.getKey(),
                 new Metadata(
                     entry.getValue().receivedTimestamp(),
-                    entry.getValue().position().offset,
+                    entry.getValue().position() == null ? null : entry.getValue().position().offset,
                     entry.getValue().beginningOffset(),
                     entry.getValue().endOffset()
                 )
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
@@ -45,7 +45,7 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
     public static final class Metadata {
 
         private final long receivedTimestamp;
-        private final long position;
+        private final Long position;
         private final long beginningOffset;
         private final long endOffset;
 
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
@@ -74,14 +74,15 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
         }
 
         /**
-         * @return The lag between the next position to fetch and the current end of the partition
+         * @return The lag between the next position to fetch and the current end of the partition, or
+         * null if the end offset is not known or there is no position.
          */
         public Long lag() {
             return endOffset == null || position == null ? null : endOffset - position;
         }
 
         /**
-         * @return The current first offset in the partition.
+         * @return The current first offset in the partition, or if the beginning offset is not known.
          */
         public Long beginningOffset() {
             return beginningOffset;
@@ -91,7 +92,8 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
          * @return The current last offset in the partition. The determination of the "last" offset
          * depends on the Consumer's isolation level. Under "read_uncommitted," this is the last successfully
          * replicated offset plus one. Under "read_committed," this is the minimum of the last successfully
-         * replicated offset plus one or the smallest offset of any open transaction.
+         * replicated offset plus one or the smallest offset of any open transaction. Null if the end offset
+         * is not known.
          */
         public Long endOffset() {
             return endOffset;
@@ -227,7 +229,7 @@ public class ConsumerRecords<K, V> implements Iterable<ConsumerRecord<K, V>> {
     }
 
     public boolean isEmpty() {
-        return records.isEmpty() && metadata.isEmpty();
+        return records.isEmpty();
     }
 
     @SuppressWarnings("unchecked")
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
@@ -16,11 +16,13 @@
  */
 package org.apache.kafka.clients.consumer;
 
+import org.apache.kafka.clients.consumer.internals.FetchedRecords;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.utils.AbstractIterator;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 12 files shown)


build.gradle
```diff
@@ -1070,6 +1070,7 @@ project(':clients') {
     testCompile libs.junitJupiterApi
     testCompile libs.junitVintageEngine
     testCompile libs.mockitoCore
+    testCompile libs.hamcrest
```
vvcephei, line 1121: To get access to more Matchers

checkstyle/suppressions.xml
```diff
@@ -100,7 +100,10 @@
               files="RequestResponseTest.java|FetcherTest.java|KafkaAdminClientTest.java"/>
 
     <suppress checks="NPathComplexity"
-              files="MemoryRecordsTest|MetricsTest|RequestResponseTest|TestSslUtils|AclAuthorizerBenchmark"/>
+              files="MemoryRecordsTest|MetricsTest|RequestResponseTest|TestSslUtils|AclAuthorizerBenchmark|MockConsumer"/>
+
+    <suppress checks="CyclomaticComplexity"
+              files="MockConsumer"/>
```
vvcephei, line 106: I've added a couple more branches to MockConsumer, which pushed it over the line.

clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
```diff
@@ -32,14 +34,99 @@
...
-    public static final ConsumerRecords<Object, Object> EMPTY = new ConsumerRecords<>(Collections.EMPTY_MAP);
+    public static final ConsumerRecords<Object, Object> EMPTY = new ConsumerRecords<>(
+        Collections.emptyMap(),
+        Collections.emptyMap()
+    );
 
     private final Map<TopicPartition, List<ConsumerRecord<K, V>>> records;
+    private final Map<TopicPartition, Metadata> metadata;
+
+    public static final class Metadata {
```
vvcephei, line 45: I made this a static inner class, since "metadata" is such an abstract concept. This way, the scope is clear.

clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerRecords.java
```diff
@@ -32,14 +34,99 @@
...
+         * depends on the Consumer's isolation level. Under "read_uncommitted," this is the last successfully
+         * replicated offset plus one. Under "read_committed," this is the minimum of the last successfully
+         * replicated offset plus one or the smallest offset of any open transaction.
+         */
+        public long endOffset() {
+            return endOffset;
+        }
+    }
+
+    private static <K, V> Map<TopicPartition, Metadata> extractMetadata(final FetchedRecords<K, V> fetchedRecords) {
```
vvcephei, line 102: This is where we translate from the internal result container to the public one. It could be moved to a utility class, but this seemed fine.

clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
```diff
@@ -1268,12 +1269,12 @@ boolean updateAssignmentMetadataIfNeeded(final Timer timer, final boolean waitFo
     /**
      * @throws KafkaException if the rebalance callback throws exception
      */
-    private Map<TopicPartition, List<ConsumerRecord<K, V>>> pollForFetches(Timer timer) {
+    private FetchedRecords<K, V> pollForFetches(Timer timer) {
```
vvcephei, line 1272: Added an internal struct so that the Fetcher can also return the desired metadata.

clients/src/main/java/org/apache/kafka/clients/consumer/MockConsumer.java
```diff
@@ -218,7 +218,22 @@ public synchronized void unsubscribe() {
...
+        final Map<TopicPartition, ConsumerRecords.Metadata> metadata = new HashMap<>();
+        for (final TopicPartition partition : subscriptions.assignedPartitions()) {
+            if (subscriptions.hasValidPosition(partition) && beginningOffsets.containsKey(partition) && endOffsets.containsKey(partition)) {
+                final SubscriptionState.FetchPosition position = subscriptions.position(partition);
+                final long offset = position.offset;
+                final long beginningOffset = beginningOffsets.get(partition);
+                final long endOffset = endOffsets.get(partition);
+                metadata.put(
+                    partition,
+                    new ConsumerRecords.Metadata(System.currentTimeMillis(), offset, beginningOffset, endOffset)
```
vvcephei: Just making the MockConsumer return its metadata as well. There are some extra checks in here because the mock doesn't guarantee a valid position or beginning/end offsets while calling poll, whereas the real consumer does.

clients/src/main/java/org/apache/kafka/clients/consumer/MockConsumer.java
```diff
@@ -229,6 +244,7 @@ public synchronized void addRecord(ConsumerRecord<K, V> record) {
             throw new IllegalStateException("Cannot add records for a partition that is not assigned to the consumer");
         List<ConsumerRecord<K, V>> recs = this.records.computeIfAbsent(tp, k -> new ArrayList<>());
         recs.add(record);
+        endOffsets.compute(tp, (ignore, offset) -> offset == null ? record.offset() : Math.max(offset, record.offset()));
```
vvcephei, line 246: Just adding this computation in for convenience.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/FetchedRecords.java
```diff
@@ -0,0 +1,100 @@
...
+            return endOffset;
+        }
+    }
+
+    public FetchedRecords() {
+        records = new HashMap<>();
+        metadata = new HashMap<>();
+    }
+
+    public void addRecords(final TopicPartition topicPartition, final List<ConsumerRecord<K, V>> records) {
```
vvcephei, line 72: moved this logic from the Fetcher, since it seemed natural to internalize it here.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
```diff
@@ -319,7 +319,7 @@ public void onSuccess(ClientResponse resp) {
                                     short responseVersion = resp.requestHeader().apiVersion();
 
                                     completedFetches.add(new CompletedFetch(partition, partitionData,
-                                            metricAggregator, batches, fetchOffset, responseVersion));
+                                            metricAggregator, batches, fetchOffset, responseVersion, resp.receivedTimeMs()));
```
vvcephei, line 321: Capture the received timestamp.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
```diff
@@ -637,20 +637,42 @@ private ListOffsetResult fetchOffsetsByTimes(Map<TopicPartition, Long> timestamp
...
+
+                        if (fetchMetadata == null
+                            || !fetchMetadata.position().offsetEpoch.isPresent()
+                            || fetchPosition.offsetEpoch.isPresent()
+                            && fetchMetadata.position().offsetEpoch.get() <= fetchPosition.offsetEpoch.get()) {
+
+                            fetched.addMetadata(
+                                partition,
+                                new FetchedRecords.FetchMetadata(receivedTimestamp, fetchPosition, startOffset, endOffset)
+                            );
```
vvcephei: First we capture the metadata

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
```diff
@@ -637,20 +637,42 @@ private ListOffsetResult fetchOffsetsByTimes(Map<TopicPartition, Long> timestamp
...
                         }
+                    }
+
+                    if (!records.isEmpty()) {
+                        fetched.addRecords(partition, records);
                         recordsRemaining -= records.size();
                     }
```
vvcephei: Then, we get the records (note the main logic is now internal to FetchedRecords).

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
```diff
@@ -596,7 +603,7 @@ public void testFetchProgressWithMissingPartitionPosition() {
         initMetadata(client, Collections.singletonMap(topic, 2));
 
         KafkaConsumer<String, String> consumer = newConsumerNoAutoCommit(time, client, subscription, metadata);
-        consumer.assign(Arrays.asList(tp0, tp1));
+        consumer.assign(asList(tp0, tp1));
```
vvcephei, line 610: A side effect of adding static imports.

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
```diff
@@ -1045,6 +1052,7 @@ public void fetchResponseWithUnexpectedPartitionIsIgnored() {
 
         ConsumerRecords<String, String> records = consumer.poll(Duration.ZERO);
         assertEquals(0, records.count());
+        assertThat(records.metadata(), equalTo(emptyMap()));
```
vvcephei, line 1059: I dropped in a couple of extra assertions where it seemed appropriate. There are also some new tests farther down to really exercise the new code.

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
```diff
@@ -2374,10 +2504,20 @@ private ListOffsetsResponse listOffsetsResponse(Map<TopicPartition, Long> partit
     }
 
     private static class FetchInfo {
+        long logFirstOffset;
+        long logLastOffset;
```
vvcephei, line 2508: I considered inferring these from the sequence of other mock interactions, but it seemed more sensible to just have a way to specify them explicitly.

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
```diff
@@ -524,7 +524,7 @@ public void testParseCorruptedRecord() throws Exception {
         consumerClient.poll(time.timer(0));
 
         // the first fetchedRecords() should return the first valid message
-        assertEquals(1, fetcher.fetchedRecords().get(tp0).size());
+        assertEquals(1, fetcher.fetchedRecords().records().get(tp0).size());
```
vvcephei, line 527: Just a consequence of moving the records inside of a response struct.

vvcephei (commented):
Hey @guozhangwang , do you have time to review this?

I'll send a follow-on PR for the Streams code.

By the way, I pushed this branch to the Apache repo so that I can concurrently send the follow-on PR based on it.

guozhangwang:
@vvcephei I'm starting to review this PR now (sorry for the late delay!). Could you rebase the PR a bit?

clients/src/main/java/org/apache/kafka/clients/consumer/MockConsumer.java
```diff
@@ -218,7 +218,21 @@ public synchronized void unsubscribe() {
         }
 
         toClear.forEach(p -> this.records.remove(p));
-        return new ConsumerRecords<>(results);
+
+        final Map<TopicPartition, ConsumerRecords.Metadata> metadata = new HashMap<>();
+        for (final TopicPartition partition : subscriptions.assignedPartitions()) {
+            if (subscriptions.hasValidPosition(partition) && beginningOffsets.containsKey(partition) && endOffsets.containsKey(partition)) {
```
guozhangwang: Why also check `beginningOffsets` here?

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
```diff
@@ -637,20 +636,32 @@ private ListOffsetResult fetchOffsetsByTimes(Map<TopicPartition, Long> timestamp
...
+                        final long receivedTimestamp = nextInLineFetch.receivedTimestamp;
+                        final Long beginningOffset = subscriptions.logStartOffset(partition);
+                        final Long endOffset = subscriptions.logEndOffset(partition, isolationLevel);
+                        final FetchPosition fetchPosition = subscriptions.position(partition);
+
+                        final FetchedRecords.FetchMetadata fetchMetadata = fetched.metadata().get(partition);
+                        if (fetchMetadata == null
+                            || !fetchMetadata.position().offsetEpoch.isPresent()
+                            || fetchPosition.offsetEpoch.isPresent()
+                            && fetchMetadata.position().offsetEpoch.get() <= fetchPosition.offsetEpoch.get()) {
```
guozhangwang: Interesting, why we do not want to update the metadata if epoch is stale?

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
```diff
@@ -637,20 +636,32 @@ private ListOffsetResult fetchOffsetsByTimes(Map<TopicPartition, Long> timestamp
...
-                            // this case shouldn't usually happen because we only send one fetch at a time per partition,
-                            // but it might conceivably happen in some rare cases (such as partition leader changes).
-                            // we have to copy to a new list because the old one may be immutable
-                            List<ConsumerRecord<K, V>> newRecords = new ArrayList<>(records.size() + currentRecords.size());
-                            newRecords.addAll(currentRecords);
-                            newRecords.addAll(records);
-                            fetched.put(partition, newRecords);
+                    TopicPartition partition = nextInLineFetch.partition;
+
+                    if (subscriptions.isAssigned(partition)) {
```
guozhangwang, line 643: Could it ever happen that this condition failed except mocking tests?

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
guozhangwang: Also, would `<=` accept `null` from `fetchMetadata.position().offsetEpoch.get()` on left hand side?

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
```diff
@@ -2041,6 +2052,118 @@ public void testInvalidGroupMetadata() throws InterruptedException {
...
+        final ConsumerRecords.Metadata actualMetadata = records.metadata().get(tp0);
+        assertEquals(100L, (long) actualMetadata.endOffset());
+        assertEquals(55L, (long) actualMetadata.position());
+        assertEquals(45L, (long) actualMetadata.lag());
+        consumer.close(Duration.ZERO);
+    }
+
+
+    @Test
+    public void testPollMetadataWithExtraPartitions() {
```
guozhangwang, line 2088: Does the test cover 1) stale epoch, 2) no prev value, cases?

guozhangwang (commented):
Mostly minor comments, otherwise LGTM!

clients/src/main/java/org/apache/kafka/clients/consumer/MockConsumer.java
vvcephei: Ah, this is from before I removed it.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
vvcephei: I copied this check from fetchRecords, which says "this can happen when a rebalance happened before fetched records are returned to the consumer's poll call". I.e., it seems like it can actually happen, but a comment is called for, at least. I'll add it.

clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java
vvcephei: Ah, good catch. It looks like this was also leftover from a previous version.

I used to directly populate the returned metadata from the fetch response, but now I'm just populating the returned metadata from the subscription state, which `initializeCompletedFetch` has already updated.

The benefit is that we don't have to worry about cases like this, since they've already been checked.

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
vvcephei: Good question. Thanks to your other comment, I've removed the stale epoch check.

By the "no prev value" case, are you referring to what happens when we get a fetch response for a partition for the first time? This is actually _all_ we're testing here.

vvcephei:
Huh, I can't get the PlaintextConsumerTest to fail locally...

clients/src/test/java/org/apache/kafka/clients/consumer/KafkaConsumerTest.java
guozhangwang: Cool, thanks!

guozhangwang:
Made another pass on the latest commit and it LGTM. Also triggered the test again.

Once it passed we can merge as-is.

vvcephei:
Thanks, @guozhangwang ! 

I ran the PlaintextConsumerTest a bunch more times, and also searched through the trunk build logs. I think this PR did make the test more flaky, and I suspect the reason is that the consumer long-poll will now return "early" if we get a metadata-only fetch response. I've adjusted the test to account for this, and we'll see how the build does now.

guozhangwang:
> I ran the PlaintextConsumerTest a bunch more times, and also searched through the trunk build logs. I think this PR did make the test more flaky, and I suspect the reason is that the consumer long-poll will now return "early" if we get a metadata-only fetch response. I've adjusted the test to account for this, and we'll see how the build does now.

Got it. How about this:

In `KafkaConsumer.java` line 

```
if (!records.isEmpty()) {
                    // before returning the fetched records, we can send off the next round of fetches
                    // and avoid block waiting for their responses to enable pipelining while the user
                    // is handling the fetched records.
                    //
                    // NOTE: since the consumed position has already been updated, we must not allow
                    // wakeups or any other errors to be triggered prior to returning the fetched records.
                    if (fetcher.sendFetches() > 0 || client.hasPendingRequests()) {
                        client.transmitSends();
                    }

                    return this.interceptors.onConsume(new ConsumerRecords<>(records));
                }
```

We change the condition to `records.records().isEmpty`?

vvcephei:
Thanks @guozhangwang ,

I considered that, but I think for this work, we actually do want to return the metadata if we have fetched some. Since the semantics of `poll(timeout)` are that it will wait _up to_ the timeout to return and that it may or may not return any records, it seems this PR's behavior is just fine.

All that is wrong is that this particular test expects to get records back after a single poll. Among our integration tests, this is a pretty unique expectation, so I felt good about relaxing it. Does that seem ok to you?

guozhangwang:
Hey @vvcephei I was concerned about not the tests actually, it just rang to me if we should be paranoid about any side-effects for a tighter consumer loop in user code like:

```
while (running()) {
    records = consumer.poll();
    if (!records.isEmpty()) // process them
}
```

Before this loop is iterated say N times every second, and now it could be M times every second where M >> N but we are still returning the same number of records in every sec. If user's process logic does not have, e.g., `if (!records.isEmpty())` or they have some computational logic for every `records` variable returned no matter if it is empty, then users may see their CPU increases surprisingly.

But I also realized that the above change I made would effectively "kill" the purpose of this approach. So probably we do not have a better solution ATM without even larger API changes, and we'd just bite it and see if there's any surprises.

vvcephei:
Thanks @guozhangwang . I share that concern.

It's a little different than what you pointed out, but to call attention to it: I actually didn't change the definition of `ConsumerRecords#isEmpty`. It is still just `return records.isEmpty()`, so it is only true when there are records. The specific rationale is not to break logic similar to what you quoted.

To your point, it's true that `M > N`, but I doubt that `M >> N`. The rate at which we return is still bounded by the `latency * concurrency` of fetch request-response lifecycles _that return some records or metadata_. As long as there's no new information to return, the Consumer's long-poll will still be effective.

And to the final point, I really doubt that any real code could have an expectation that they can just call `poll` and _always_ get nonzero records back. It just doesn't seem like a reasonable expectation from a user's perspective.

guozhangwang:
Yup, I think we are on the same page for:

1) `ConsumerRecords#isEmpty`: I was not pointing that it may not return true even if only metadata changed, I meant that if users do not have the `records.isEmpty` condition before the processing logic, then they may waste some more cycles now.
2) Users should not expect that `consumer.poll` always returns non-zero records: that is also true. Like I said, I'm just thinking if it is a paranoid thought for a rare case.

For `M v.s. N`, I think you're right that, even with different consumer's min.byte and max.wait configurations, the difference should be small if any. So all clear on my side.

vvcephei:
That last build was green, but it was a few days ago. Just rebased and pushed again to make sure I don't break the build when I merge this.

vvcephei:
Most of those failures were known flaky tests, but one was an EasyMock error. I'm not able to repro it locally after a rebase, though. Rebased, pushed, and trying one more time to get a clean build.

vvcephei:
Flaky test failures:

```
    Build / JDK 11 / org.apache.kafka.clients.consumer.internals.FetcherTest.testEarlierOffsetResetArrivesLate()
    Build / JDK 11 / org.apache.kafka.clients.producer.KafkaProducerTest.testHeadersWithExtendedClasses()
    Build / JDK 15 / kafka.integration.MetricsDuringTopicCreationDeletionTest.testMetricsDuringTopicCreateDelete()
```

The most concerning one is the FetcherTest, but it's also failing on trunk.

build.gradle
ijuma: We actually removed these from `clients` when we migrated to JUnit 5...

ijuma:
Have we verified that this doesn't cause perf regressions? It seems to have caused system test failures: https://issues.apache.org/jira/browse/KAFKA-12268

build.gradle
vvcephei: Ah, my apologies. I'll send a new PR to back this out.

vvcephei:
Thanks @ijuma ; We did run the Streams system tests and benchmarks and verified there was no perf regression. I should have run the Client system tests too; sorry about that. I'll follow up on https://github.com/apache/kafka/pull/10022

ijuma:
Thanks @vvcephei !

PR #9837, opened 2021-01-06
KAFKA-9566: Improve DeserializationExceptionHandler JavaDocs

mjsax:
Call for review @vvcephei


vvcephei (approved):
Thanks for clarifying this, @mjsax

chia7712:
Is there a incorrect jira number? I can't get the relation between this PR and KAFKA-6566.

chia7712 (approved):
+1 to this nice docs.

BTW, how about introducing a new interface without disallowed methods? for example, a read-only ```ProcessorContext```. It requires extra effort but it should be more useful than JavaDocs.

mjsax:
@chia7712 -- I miss-typed the ticket number... Fixed now.

The ticket actually discussed the exact same idea as you propose. We just wanted to get a "quick" doc-improvement out fast, as the change would require a KIP what takes longer.

PR #9838, opened 2021-01-06
KAFKA-12153; Update producer state before updating start/end offsets after truncation

hachikuji:
When we truncate the log, the first unstable offset might become valid. On the other hand, the logic in `updateHighWatermarkMetadata` assumes that the first stable offset remains at a valid position. Since this method can be reached through either `updateLogStartOffset` or `updateLogEndOffset` in the truncation paths, we need to ensure that the first unstable offset first reflects the truncated state.

### DIFF
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -2106,8 +2106,8 @@ class Log(@volatile private var _dir: File,
             // in case the first unstable offset is invalidated by the truncation.
             loadProducerState(targetOffset, reloadFromCleanShutdown = false)
 
-            updateLogStartOffset(math.min(targetOffset, this.logStartOffset))
             updateLogEndOffset(targetOffset)
+            updateLogStartOffset(math.min(targetOffset, this.logStartOffset))
           }
           true
         }
@@ -2139,8 +2139,8 @@ class Log(@volatile private var _dir: File,
         producerStateManager.truncateFullyAndStartAt(newOffset)
         maybeIncrementFirstUnstableOffset()
 
-        updateLogStartOffset(newOffset)
         updateLogEndOffset(newOffset)
+        updateLogStartOffset(newOffset)
       }
     }
   }
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -383,28 +383,28 @@ class Log(@volatile private var _dir: File,
    * @return the updated high watermark offset
    */
   def updateHighWatermark(hw: Long): Long = {
-    val newHighWatermark = if (hw < logStartOffset)
-      logStartOffset
-    else if (hw > logEndOffset)
-      logEndOffset
-    else
-      hw
-    updateHighWatermarkMetadata(LogOffsetMetadata(newHighWatermark))
-    newHighWatermark
+    updateHighWatermark(LogOffsetMetadata(hw))
   }
 
-  def updateHighWatermarkOffsetMetadata(hw: LogOffsetMetadata): Long = {
-    val newHighWatermark = if (hw.messageOffset < logStartOffset) {
-      updateHighWatermarkMetadata(LogOffsetMetadata(logStartOffset))
-      logStartOffset
-    } else if (hw.messageOffset > logEndOffset) {
-      updateHighWatermarkMetadata(logEndOffsetMetadata)
-      logEndOffset
+  /**
+   * Update high watermark with offset metadata. The new high watermark will be lower
+   * bounded by the log start offset and upper bounded by the log end offset.
+   *
+   * @param highWatermarkMetadata the suggested high watermark with offset metadata
+   * @return the updated high watermark offset
+   */
+  def updateHighWatermark(highWatermarkMetadata: LogOffsetMetadata): Long = {
+    val endOffsetMetadata = logEndOffsetMetadata
+    val newHighWatermarkMetadata = if (highWatermarkMetadata.messageOffset < logStartOffset) {
+      LogOffsetMetadata(logStartOffset)
+    } else if (highWatermarkMetadata.messageOffset >= endOffsetMetadata.messageOffset) {
+      endOffsetMetadata
     } else {
-      updateHighWatermarkMetadata(hw)
-      hw.messageOffset
+      highWatermarkMetadata
     }
-    newHighWatermark
+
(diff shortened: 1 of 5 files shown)


core/src/test/scala/unit/kafka/log/LogTest.scala
```diff
@@ -190,6 +190,59 @@ class LogTest {
...
+  def testTruncateBelowFirstUnstableOffset(): Unit = {
+    testTruncateBelowFirstUnstableOffset(_.truncateTo)
+  }
+
+  @Test
+  def testTruncateFullyAndStartBelowFirstUnstableOffset(): Unit = {
+    testTruncateBelowFirstUnstableOffset(_.truncateFullyAndStartAt)
+  }
+
+  def testTruncateBelowFirstUnstableOffset(
```
junrao: Could this be private?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2124,13 +2132,15 @@ class Log(@volatile private var _dir: File,
...
 
-        producerStateManager.truncate()
-        producerStateManager.updateMapEndOffset(newOffset)
+        // Clear producer state prior to adjusting log start and end offsets
+        // in case the first unstable offset is invalidated by the truncation.
+        producerStateManager.truncateFullyAndStartAt(newOffset)
         maybeIncrementFirstUnstableOffset()
+
         updateLogStartOffset(newOffset)
+        updateLogEndOffset(newOffset)
```
junrao: This is an existing issue. updateLogEndOffset() makes the assumption that newOffset is nextOffset since it uses nextOffsetMetadata directly. This can be error prone in the future. While you are at this, perhaps we could remove the input param from updateLogEndOffset()?

junrao (commented):
@hachikuji : Thanks for the PR. Just a couple of minor comments.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2096,10 +2100,14 @@ class Log(@volatile private var _dir: File,
             val deletable = logSegments.filter(segment => segment.baseOffset > targetOffset)
             removeAndDeleteSegments(deletable, asyncDelete = true, LogTruncation)
             activeSegment.truncateTo(targetOffset)
-            updateLogEndOffset(targetOffset)
-            updateLogStartOffset(math.min(targetOffset, this.logStartOffset))
             leaderEpochCache.foreach(_.truncateFromEnd(targetOffset))
+
+            // Load producer state prior to adjusting log start and end offsets
+            // in case the first unstable offset is invalidated by the truncation.
             loadProducerState(targetOffset, reloadFromCleanShutdown = false)
```
chia7712: In the case of requiring truncation below the high watermark, ```maybeIncrementFirstUnstableOffset``` gets called two times (by ```loadProducerState``` and ```updateLogEndOffset```). Is first call unnecessary?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2124,12 +2132,14 @@ class Log(@volatile private var _dir: File,
...
 
-        producerStateManager.truncate()
-        producerStateManager.updateMapEndOffset(newOffset)
+        // Clear producer state prior to adjusting log start and end offsets
+        // in case the first unstable offset is invalidated by the truncation.
+        producerStateManager.truncateFullyAndStartAt(newOffset)
         maybeIncrementFirstUnstableOffset()
+
+        updateLogEndOffset(newOffset)
         updateLogStartOffset(newOffset)
```
hachikuji: One note here: in the initial patch, I reordered `updateLogStartOffset` before `updateLogEndOffset`. This caused the  test case `ReassignPartitionsIntegrationTest.testHighWaterMarkAfterPartitionReassignment` to fail. The problem is the logic in `updateHighWatermark`, which is called by `updateLogStartOffset`, assumes that the log end offset has already been set correctly. The result of reversing the order was that the high watermark was left at a value lower than the log start offset. 

Frankly the test case is a bit contrived. It calls `truncateFullyAndStartAt` on the leader before starting a reassignment. I don't think there are any valid operations that could have the same effect. At the same time, it feels brittle to rely on tricky ordering assumptions like this, so I'm wondering whether we can improve this. One idea for the case of `truncateFullyAndStartAt` is to bypass `updateLogEndOffset` and `updateLogStartOffset` and update the values directly. Any thoughts?

core/src/main/scala/kafka/log/Log.scala
junrao: Actually, this is not an issue on second look.

core/src/main/scala/kafka/log/Log.scala
junrao: Maybe we can have a new method that updates the producerState, logEndOffset, high watermark, log start offset and last stable offset together? Sth like the following.

```
    rebuildProducerState()
    logStartOffset = offset
    nextOffsetMetadata = LogOffsetMetadata(offset, activeSegment.baseOffset, activeSegment.size)

    // Update the high watermark metadata if the offset changes
    if (highWatermark != offset) {
      updateHighWatermarkMetadata(nextOffsetMetadata)
    }

    if (this.recoveryPoint != offset) {
      this.recoveryPoint = offset
    }

```

This avoids dependencies and also avoids some the the duplication of the logic that Chia-Ping mentioned.

hachikuji: @junrao Thanks. I tried something like this in the latest commit. Let me know what you think.

core/src/main/scala/kafka/log/Log.scala
hachikuji: Thanks, I think this is addressed in the latest commit.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -2124,17 +2130,28 @@ class Log(@volatile private var _dir: File,
...
+          startOffset = newOffset,
+          endOffset = newOffset
+        )
       }
     }
   }
 
+  private def completeTruncation(
+    startOffset: Long,
+    endOffset: Long,
```
junrao: Do we need the trailing comma?

junrao (approved):
@hachikuji : Thanks for the updated PR. LGTM other than the minor comment below.

core/src/main/scala/kafka/log/Log.scala
hachikuji: Thanks, added by mistake. I will remove it.

PR #9839, opened 2021-01-07
MINOR: Factor `RaftManager` out of `TestRaftServer`

hachikuji:
This patch factors out a `RaftManager` class from `TestRaftServer` which will be needed when we integrate this layer into the server. This class encapsulates the logic to build `KafkaRaftClient` as well as its IO thread.

### DIFF
--- a/clients/src/test/java/org/apache/kafka/clients/MockClient.java
+++ b/clients/src/test/java/org/apache/kafka/clients/MockClient.java
@@ -263,8 +263,11 @@ public class MockClient implements KafkaClient {
             if (remainingBlockingWakeups <= 0)
                 return;
 
-            while (numBlockingWakeups == remainingBlockingWakeups)
-                wait();
+            TestUtils.waitForCondition(() -> {
+                if (numBlockingWakeups == remainingBlockingWakeups)
+                    MockClient.this.wait(500);
+                return numBlockingWakeups < remainingBlockingWakeups;
+            }, 5000, "Failed to receive expected wakeup");
         } catch (InterruptedException e) {
             throw new InterruptException(e);
         }
--- a/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
+++ b/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
@@ -87,6 +87,7 @@ private[raft] class RaftSendThread(
 
   def sendRequest(request: RequestAndCompletionHandler): Unit = {
     queue.add(request)
+    wakeup()
   }
 
 }
--- a/core/src/main/scala/kafka/raft/RaftManager.scala
+++ b/core/src/main/scala/kafka/raft/RaftManager.scala
@@ -1,0 +1,279 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package kafka.raft
+
+import java.io.File
+import java.nio.file.Files
+import java.util.Random
+import java.util.concurrent.CompletableFuture
+
+import kafka.log.{Log, LogConfig, LogManager}
+import kafka.raft.KafkaRaftManager.RaftIoThread
+import kafka.server.{BrokerTopicStats, KafkaConfig, KafkaServer, LogDirFailureChannel}
+import kafka.utils.timer.SystemTimer
+import kafka.utils.{KafkaScheduler, Logging, ShutdownableThread}
+import org.apache.kafka.clients.{ApiVersions, ClientDnsLookup, ManualMetadataUpdater, NetworkClient}
+import org.apache.kafka.common.TopicPartition
+import org.apache.kafka.common.metrics.Metrics
+import org.apache.kafka.common.network.{ChannelBuilders, NetworkReceive, Selectable, Selector}
+import org.apache.kafka.common.protocol.ApiMessage
+import org.apache.kafka.common.requests.RequestHeader
+import org.apache.kafka.common.security.JaasContext
+import org.apache.kafka.common.utils.{LogContext, Time}
+import org.apache.kafka.raft.{FileBasedStateStore, KafkaRaftClient, QuorumState, RaftClient, RaftConfig, RaftRequest, RecordSerde}
+
+import scala.jdk.CollectionConverters._
+
+object KafkaRaftManager {
+  class RaftIoThread(
+    client: KafkaRaftClient[_]
+  ) extends ShutdownableThread(
+    name = "raft-io-thread",
+    isInterruptible = false
+  ) {
+    override def doWork(): Unit = {
+      client.poll()
+    }
+
+    override def initiateShutdown(): Boolean = {
+      if (super.initiateShutdown()) {
+        client.shutdown(5000).whenComplete { (_, exception) =>
+          if (exception != null) {
+            error("Graceful shutdown of RaftClient failed", exception)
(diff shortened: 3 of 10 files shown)


core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
```diff
@@ -87,6 +87,7 @@ private[raft] class RaftSendThread(
 
   def sendRequest(request: RequestAndCompletionHandler): Unit = {
     queue.add(request)
+    wakeup()
```
chia7712, line 90: Is this a bug fix?

core/src/main/scala/kafka/raft/RaftManager.scala
```diff
@@ -0,0 +1,279 @@
...
+  private val raftClient = buildRaftClient()
+  private val raftIoThread = new RaftIoThread(raftClient)
+
+  def startup(): Unit = {
+    netChannel.start()
+    raftClient.initialize()
+    raftIoThread.start()
+  }
+
+  def shutdown(): Unit = {
```
chia7712, line 126: Should we shutdown ```raftClient``` in this method?

core/src/main/scala/kafka/raft/RaftManager.scala
```diff
@@ -0,0 +1,279 @@
...
+    listener: RaftClient.Listener[T]
+  ): Unit = {
+    raftClient.register(listener)
+  }
+
+  override def scheduleAppend(
+    epoch: Int,
+    records: Seq[T]
+  ): Option[Long] = {
+    val offset: java.lang.Long = raftClient.scheduleAppend(epoch, records.asJava)
```
chia7712, line 144: (this is unrelated to this PR.)

How about using ```Optional``` to replace the ```null```?

core/src/main/scala/kafka/raft/RaftManager.scala
```diff
@@ -0,0 +1,279 @@
...
+          }
+        }
+        true
+      } else {
+        false
+      }
+    }
+
+    override def isRunning: Boolean = {
+      client.isRunning
```
chia7712: Should it consider the state of "this thread also"? for example:
```scala 
  super.isRunning && client.isRunning
```

core/src/main/scala/kafka/tools/TestRaftServer.scala
```diff
@@ -131,162 +102,29 @@ class TestRaftServer(
...
 
-    socketServer.startProcessingRequests(Map.empty)
-    ioThread = new RaftIoThread(raftClient)
-    ioThread.start()
     workloadGenerator.start()
+    raftManager.startup()
+    socketServer.startProcessingRequests(Map.empty)
   }
 
   def shutdown(): Unit = {
```
chia7712, line 110: Is it need to shutdown ```raftManager```?

chia7712 (commented):
@hachikuji Thanks for nice refactor. Please take a look at following comments.

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
hachikuji: Yes indeed, thanks for noticing. I guess your point is that we should have a test case. Let me see about writing one.

core/src/main/scala/kafka/raft/RaftManager.scala
hachikuji: Yeah, it's a good idea. I used the `null` as a placeholder until I thought of a better idea, but `Optional` is probably just as good as a placeholder and safer.

core/src/main/scala/kafka/raft/RaftManager.scala
hachikuji: See the logic in `RaftIoThread.initiateShutdown`. Note that the thread continues polling until graceful shutdown has either completed or timed out.

core/src/main/scala/kafka/raft/RaftManager.scala
hachikuji: I think not. `super.isRunning` will return true as soon as `initiateShutdown` is called, but I want the thread to keep polling until the client itself says it is time to shutdown.

hachikuji: I've added a test case which hopefully makes the expected behavior clearer. Let me know if it seems reasonable.

clients/src/test/java/org/apache/kafka/clients/MockClient.java
```diff
@@ -263,8 +263,11 @@ private synchronized void maybeAwaitWakeup() {
             if (remainingBlockingWakeups <= 0)
                 return;
 
-            while (numBlockingWakeups == remainingBlockingWakeups)
-                wait();
+            TestUtils.waitForCondition(() -> {
+                if (numBlockingWakeups == remainingBlockingWakeups)
+                    MockClient.this.wait(500);
```
chia7712, line 268: ```TestUtils.waitForCondition``` already call ```sleep``` so is this ```wait``` necessary?

core/src/main/scala/kafka/raft/RaftManager.scala
```diff
@@ -0,0 +1,279 @@
...
+
+object KafkaRaftManager {
+  class RaftIoThread(
+    client: KafkaRaftClient[_]
+  ) extends ShutdownableThread(
+    name = "raft-io-thread",
+    isInterruptible = false
+  ) {
+    override def doWork(): Unit = {
+      client.poll()
```
chia7712, line 49: just curious. If ```poll``` throws exception, this I/O thread get died and ```isRunning``` still returns ```true``` (since raft client is running). How ```KafkaRaftManager``` deal with that scenario?

chia7712 (approved):
@hachikuji Thanks for nice refactor and test cases. just left two trivial comments. +1 to merge.

core/src/main/scala/kafka/raft/RaftManager.scala
hachikuji: I started to implement this, but found it was taking me a bit afield. If you don't mind, let's address this in https://issues.apache.org/jira/browse/KAFKA-12158.

clients/src/test/java/org/apache/kafka/clients/MockClient.java
hachikuji: I was thinking the wait would allow for faster notification.

core/src/main/scala/kafka/raft/RaftManager.scala
hachikuji: Good point. I changed `isRunning` so that it also checks if the thread has failed. Added a test case as well.

core/src/main/scala/kafka/raft/RaftManager.scala
```diff
@@ -0,0 +1,279 @@
...
+          }
+        }
+        true
+      } else {
+        false
+      }
+    }
+
+    override def isRunning: Boolean = {
+      client.isRunning && !isThreadFailed
```
chia7712, line 68: There is another issue. After ```client.poll()``` gets failed, the ```isThreadFailed``` becomes ```true``` and so ```isRunning``` returns ```false```. It means ```initiateShutdown``` does not call ```client.shutdown``` even though we call ```shutdown``` explicitly.

chia7712 (commented):
@hachikuji Thanks for updating code. There is another question. Please take a look.

core/src/main/scala/kafka/raft/RaftManager.scala
hachikuji: It's a good question and I was debating it earlier today. It's worth keeping in mind that `KafkaRaftClient` doesn't really own the resources that would need closing. Sockets and files are owned by `RaftManager`, so they will still get closed properly. The one exception is de-registration of metrics in `KafkaRaftMetrics.close`. 

The main purpose of `shutdown` is to attempt a graceful departure from the quorum. But after we after we have hit an unexpected exception and are left in an undetermined state, do we still want to try the graceful shutdown? I was leaning toward 'no'. 

Perhaps what this suggests is that we should have a separate `close` method which can handle cleanup of any resources that `KafkaRaftClient` owns today (and might own in the future). If we hit an uncaught exception, then we can forego the call to `shutdown`, but we can ensure that `close` always gets called. How does that sound?

chia7712: > we can forego the call to shutdown, but we can ensure that close always gets called. How does that sound?

sounds good to me :)

hachikuji:
@chia7712 Thanks for reviews. Take one more look at the updates when you have time.

chia7712 (approved):
@hachikuji Thanks for this great refactor. LGTM

PR #9847, opened 2021-01-08
KAFKA-10703: Better handling and doc for config defaults of topics

tombentley:
Default configs are not supported for topic resources, so document this
and prevent the creation of such ConfigResources.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/config/ConfigResource.java
+++ b/clients/src/main/java/org/apache/kafka/common/config/ConfigResource.java
@@ -59,13 +59,19 @@ public final class ConfigResource {
 
     /**
      * Create an instance of this class with the provided parameters.
+     * An empty name is used to refer to the default resource of a resource type.
+     * Default configs are not supported for topic resources.
      *
      * @param type a non-null resource type
      * @param name a non-null resource name
+     * @throws IllegalArgumentException if an empty name is passed for a topic resource.
      */
     public ConfigResource(Type type, String name) {
         Objects.requireNonNull(type, "type should not be null");
         Objects.requireNonNull(name, "name should not be null");
+        if (type == Type.TOPIC && name.isEmpty()) {
+            throw new IllegalArgumentException("Default configs are not supported for topic resources");
+        }
         this.type = type;
         this.name = name;
     }
--- a/clients/src/main/java/org/apache/kafka/common/config/ConfigResource.java
+++ b/clients/src/main/java/org/apache/kafka/common/config/ConfigResource.java
@@ -60,18 +60,13 @@ public final class ConfigResource {
     /**
      * Create an instance of this class with the provided parameters.
      * An empty name is used to refer to the default resource of a resource type.
-     * Default configs are not supported for topic resources.
      *
      * @param type a non-null resource type
      * @param name a non-null resource name
-     * @throws IllegalArgumentException if an empty name is passed for a topic resource.
      */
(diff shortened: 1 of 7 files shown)


tombentley:
@cmccabe grateful if you could take a look at this.

Strictly speaking, making throw ConfigResource is not entirely backwards compatible, but invalid instances could never have been successfully used before. The alternative would be to allow creation of such instances and fail on the relevant calls to the Admin client, which is more complicated for no obvious benefit imho.

tombentley:
@chia7712 maybe you could review this, if @cmccabe is busy?

chia7712:
> The alternative would be to allow creation of such instances and fail on the relevant calls to the Admin client, which is more complicated for no obvious benefit imho.

I prefer this solution. There is a similar issue (empty topic) in other requests (https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L1430) and the exception is propagated to ```Future``` instead of throwing directly.

tombentley:
@chia7712 thanks for taking a look. It's true that I should have fixed this on the broker, so it works for non-Java clients too. It turns out that the three RPCs are remarkably inconsistent in their error codes:

* `DescribeConfigs` code INVALID_TOPIC_EXCEPTION and message `Topic name is illegal, it can't be empty` which could be more explicit (it could be mistaken as a bug, rather than default topic configs intentionally not being supported)
* `AlterConfigs` code `UNKNOWN_TOPIC_OR_PARTITION` and message `The topic '' does not exist.`, which again could be more explicit
* `IncrementalAlterConfigs` code `INVALID_REQUEST` and message `Path must not end with / character`, which is downright confusing and what the JIRA describes.

To maintain compatibility I've just changed the error messages to be consistent, but left the error codes unchanged. I can open a KIP to bump the RPC versions to make these APIs use consistent error codes too. Other inconsistencies will exist between `DescribeConfigs` (which does a `Topic.validate()` then a metadata lookup) and `AlterConfigs` (the other way around) and `IncrementalAlterConfigs` (which ultimately follows the `AlterConfigs` path, but not before the ZK interaction using an unvalidated topic name).

tombentley:
Rebased for conflict.

@chia7712 could you review? And what did you think about the value of a KIP to make the error codes consistent?

tombentley:
@chia7712 any chance of a 2nd review here?

tombentley:
@chia7712, @dajac, @ijuma please could someone review this?

tombentley:
Rebased for conflict. 

@rondagostino I don't quite understand the plan for how topic configs are going to work with KIP-500. The `ConfigHelper` seems to provide an abstraction for `DescribeConfigs` RPC, but it seems that currently `AlterConfigs` and `IncrementalAlterConfigs` require ZK. Presumably an abstraction for these is in the works, and that would also need to handle default topic configs in a consistent way to what I'm trying to do here. I need to ensure that the ZKless path also has test coverage.

rondagostino:
@tombentley `Alter*` RPCs will be forwarded to the Raft-based controller and won't be handled in the broker anymore.  The changes occur in the controller and are written to the metadata log, and the brokers get the changes as they read that log.  The brokers apply the changes to their `CachedConfigRepository` instance, and then `DescribeConfigs` emits the current config state as recorded there.

tombentley:
Thanks @rondagostino.

@chia7712 @cmccabe would one of you be able to take a look?

PR #9850, opened 2021-01-08
KAFKA-12168; Move envelope request parsing out of SocketServer

hachikuji:
Prior to this patch, envelope handling was a shared responsibility between `SocketServer` and `KafkaApis`.  The former was responsible for parsing and validation, while the latter was responsible for authorization. This patch consolidates logic in `KafkaApis` so that envelope requests follow the normal request flow.

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaApis.scala
+++ b/core/src/main/scala/kafka/server/KafkaApis.scala
@@ -3212,13 +3212,10 @@ class KafkaApis(val requestChannel: RequestChannel,
   def handleEnvelope(request: RequestChannel.Request): Unit = {
     val envelope = request.body[EnvelopeRequest]
 
-    if (!config.metadataQuorumEnabled) {
-      // If forwarding is not yet enabled, we treat the request as unparsable and close the connection
+    if (!config.metadataQuorumEnabled || !request.context.fromPrivilegedListener) {
+      // If forwarding is not yet enabled or this request has been received on an invalid endpoint,
+      // then we treat the request as unparsable and close the connection.
       closeConnection(request, Collections.emptyMap())
-      return
-    } else if (!request.context.fromPrivilegedListener) {
-      sendErrorResponseMaybeThrottle(request, new ClusterAuthorizationException(
-        s"Invalid envelope request on unprivileged listener ${request.context.listenerName}"))
       return
     } else if (!authorize(request.context, CLUSTER_ACTION, CLUSTER, CLUSTER_NAME)) {
       sendErrorResponseMaybeThrottle(request, new ClusterAuthorizationException(
--- a/core/src/main/scala/kafka/server/KafkaApis.scala
+++ b/core/src/main/scala/kafka/server/KafkaApis.scala
@@ -19,16 +19,18 @@ package kafka.server
 
 import java.lang.{Byte => JByte}
 import java.lang.{Long => JLong}
+import java.net.{InetAddress, UnknownHostException}
 import java.nio.ByteBuffer
 import java.util
 import java.util.{Collections, Optional}
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
+
 import kafka.admin.{AdminUtils, RackAwareMode}
 import kafka.api.ElectLeadersRequestOps
 import kafka.api.{ApiVersion, KAFKA_0_11_0_IV0, KAFKA_2_3_IV0}
 import kafka.cluster.Partition
-import kafka.common.OffsetAndMetadata
+import kafka.common.{KafkaException, OffsetAndMetadata}
 import kafka.controller.{KafkaController, ReplicaAssignment}
 import kafka.coordinator.group.{GroupCoordinator, JoinGroupResult, LeaveGroupResult, SyncGroupResult}
 import kafka.coordinator.transaction.{InitProducerIdResult, TransactionCoordinator}
@@ -66,7 +68,7 @@ import org.apache.kafka.common.message.OffsetForLeaderEpochRequestData.OffsetFor
 import org.apache.kafka.common.message.OffsetForLeaderEpochResponseData
 import org.apache.kafka.common.message.OffsetForLeaderEpochResponseData.{EpochEndOffset, OffsetForLeaderTopicResult, OffsetForLeaderTopicResultCollection}
 import org.apache.kafka.common.metrics.Metrics
-import org.apache.kafka.common.network.{ListenerName, Send}
+import org.apache.kafka.common.network.{ClientInformation, ListenerName, Send}
 import org.apache.kafka.common.protocol.{ApiKeys, Errors}
 import org.apache.kafka.common.record._
 import org.apache.kafka.common.replica.ClientMetadata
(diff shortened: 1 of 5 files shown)


hachikuji:
cc @abbccdda

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -186,77 +160,70 @@ class KafkaApis(val requestChannel: RequestChannel,
       trace(s"Handling request:${request.requestDesc(true)} from connection ${request.context.connectionId};" +
         s"securityProtocol:${request.context.securityProtocol},principal:${request.context.principal}")
 
-      val handled = request.envelope.exists { envelope =>
```
chia7712, line 189: We don't need to check ```envelope``` for each request type now?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3242,6 +3209,133 @@ class KafkaApis(val requestChannel: RequestChannel,
...
 
+  def handleEnvelope(request: RequestChannel.Request): Unit = {
+    val envelope = request.body[EnvelopeRequest]
+
+    if (!config.metadataQuorumEnabled) {
+      // If forwarding is not yet enabled, we treat the request as unparsable and close the connection
+      closeConnection(request, Collections.emptyMap())
+      return
+    } else if (!request.context.fromPrivilegedListener) {
+      sendErrorResponseMaybeThrottle(request, new ClusterAuthorizationException(
```
chia7712: not sure whether ```ClusterAuthorizationException ``` suits this case. How about adding a (new) specify exception for this case?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3242,6 +3209,133 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+  ): Option[KafkaPrincipal] = {
+    val envelope = envelopeRequest.body[EnvelopeRequest]
+    try {
+      val principalSerde = envelopeRequest.context.principalSerde.asScala
+      principalSerde.map { serde =>
+        serde.deserialize(envelope.requestPrincipal)
+      }
+    } catch {
+      case e: Exception =>
+        warn(s"Failed to deserialize principal from envelope request $envelope", e)
```
chia7712: We can covert the exception to ```throw new PrincipalDeserializationException("Failed to parse client principal from envelope")``` here to avoid redundant try-catch (another is in handleEnvelope) and unnecessary ```Option``` wrap.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3242,6 +3209,133 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+    }
+  }
+
+  private def parseForwardedRequestHeader(
+    buffer: ByteBuffer
+  ): Option[RequestHeader] = {
+    try {
+      Some(RequestHeader.parse(buffer))
+    } catch {
+      case e: InvalidRequestException =>
```
chia7712, line 3306: ditto

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3242,6 +3209,133 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+        processor = envelope.processor,
+        context = forwardedContext,
+        startTimeNanos = envelope.startTimeNanos,
+        envelope.memoryPool,
+        buffer,
+        requestChannel.metrics,
+        Some(envelope)
+      ))
+    } catch {
+      case e: InvalidRequestException =>
```
chia7712, line 3291: ditto

chia7712 (commented):
@hachikuji Thanks for your patch. Left some comments. Please take a look.

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: We do all the same validation in `handleEnvelope` now. Previously the parsing would get done in SocketServer, which means that when the embedded request is first handled at this layer, the envelope still needs authorization. Now we will not even create the embedded request object unless the envelope has been authorized.

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: Makes sense. Initially I was doing a match on the result and returning directly, but it was verbose and awkward.

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: Thanks, makes sense and simplifies the code a bit.

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: To be honest, I'm more inclined to get rid of this requirement. It seems a little arbitrary that we require the inter-broker listener for envelope requests only. If a principal has CLUSTER_ACTION, I think the damage they can do already is virtually unlimited. 

@abbccdda What do you think? Do we have a strong reason for this?

hachikuji: Perhaps I can restore the original behavior and just close the connection in order to unblock the PR. Then we can consider this question separately.

chia7712: > Perhaps I can restore the original behavior and just close the connection in order to unblock the PR. Then we can consider this question separately.

+1

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3242,6 +3209,125 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+        // We use UNSUPPORTED_VERSION if the embedded request cannot be parsed.
+        // The purpose is to disambiguate structural errors in the envelope request
+        // itself, such as an invalid client address.
+        throw new UnsupportedVersionException("Failed to parse request header from envelope", e)
+    }
+  }
+
+  private def parseForwardedPrincipal(
+    envelopeRequest: RequestChannel.Request
```
chia7712: It can accept ```EnvelopeRequest``` instead of ```RequestChannel.Request``` as caller has a ```EnvelopeRequest``` already.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3242,6 +3209,125 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+        buffer,
+        requestChannel.metrics,
+        Some(envelope)
+      )
+    } catch {
+      case e: InvalidRequestException =>
+        // We use UNSUPPORTED_VERSION if the embedded request cannot be parsed.
+        // The purpose is to disambiguate structural errors in the envelope request
+        // itself, such as an invalid client address.
+        throw new UnsupportedVersionException(s"Failed to parse forwarded request " +
```
chia7712, line 3295: How about passing root cause to this new exception?

dajac (approved):
Nice refactor. LGTM.

core/src/main/scala/kafka/server/KafkaApis.scala
dajac: IIRC, the intend was to distinguish forwarded requests from non-forwarded requests in a prior version of the KIP when we were not using an envelope. I don't think that this is really necessary since we use an envelope now. At least, I can't think of a strong case for it.

chia7712 (approved):
overall + 1. Just left some trivial comments.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3242,6 +3209,125 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+        request.context.listenerName,
+        request.context.securityProtocol,
+        ClientInformation.EMPTY,
+        request.context.fromPrivilegedListener
+      )
+
+      val forwardedRequest = parseForwardedRequest(request, forwardedContext, forwardedRequestBuffer)
+      handle(forwardedRequest)
+    } catch {
+      case e: KafkaException =>
```
chia7712: Is this cache redundant? The ```handle```, which processing ```EnvelopeRequest```, can handle the exception.

hachikuji: I think I added the `try/catch` initially because I thought the logging in `handleError` was too verbose for some of these cases. However, parsing errors should be rare, and if they are happening, it probably implies misconfiguration. So perhaps the logging should be more verbose. I will go ahead and remove this.

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: Thanks I modified the arguments so that we pass down the envelope request context and the principal bytes.

PR #9856, opened 2021-01-11
KAFKA-12171 Migrate streams:test-utils module to JUnit 5

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-12171

1. replace ```org.junit.Assert``` by ```org.junit.jupiter.api.Assertions``` 
1. replace ```org.junit``` by ```org.junit.jupiter.api``` 
1. replace ```org.junit.runners.Parameterized``` by ```org.junit.jupiter.params.ParameterizedTest```
1. replace ```org.junit.runners.Parameterized.Parameters``` by ```org.junit.jupiter.params.provider.{Arguments, MethodSource}```
1. replace ```Before``` by ```BeforeEach```
1. replace ```After``` by ```AfterEach``` 

### DIFF
--- a/streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
+++ b/streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
@@ -59,9 +59,6 @@ import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
 import java.io.File;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.lang.reflect.Modifier;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
@@ -79,7 +76,6 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static org.apache.kafka.common.utils.Utils.mkEntry;
@@ -99,16 +95,6 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
-/**
- * the test case in TopologyTestDriverTest must obey following rules.
- * 1) the method name must starts with "should"
- * 2) the method modifier must be private
- * 3) the method can't take any arguments
- * 
- * Junit 5 does not support to add "parameters" at class level. Hence, we call test method dynamically to avoid adding
- * ParameterizedTest and MethodSource to each test method. We all hate duplicate code :)
- * see {@link TopologyTestDriverTest#test} for more details
- */
 public class TopologyTestDriverTest {
     private final static String SOURCE_TOPIC_1 = "source-topic-1";
     private final static String SOURCE_TOPIC_2 = "source-topic-2";
@@ -145,15 +131,17 @@ public class TopologyTestDriverTest {
     private final LongDeserializer longDeserializer = new LongDeserializer();
 
     public static Stream<Arguments> parameters() {
-        final List<Arguments> params = new ArrayList<>();
-        Arrays.stream(TopologyTestDriverTest.class.getDeclaredMethods())
-                .filter(m -> m.getParameterCount() == 0
-                        && Modifier.isPrivate(m.getModifiers())
-                        && m.getName().startsWith("should"))
-                .map(Method::getName)
-                .collect(Collectors.toSet()).forEach(methodName -> Arrays.asList(true, false).forEach(enableEof ->
-                params.add(Arguments.of(enableEof, methodName))));
-        return params.stream();
+        return Stream.of(true, false).map(Arguments::of);
+    }
+
+    private void setProperties(final boolean eosEnabled) {
+        config = mkProperties(mkMap(
+                mkEntry(StreamsConfig.APPLICATION_ID_CONFIG, "test-TopologyTestDriver"),
+                mkEntry(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getAbsolutePath())
+        ));
+        if (eosEnabled) {
+            config.setProperty(StreamsConfig.PROCESSING_GUARANTEE_CONFIG, StreamsConfig.EXACTLY_ONCE);
+        }
     }
 
     private final static class TTDTestRecord {
@@ -454,33 +442,25 @@ public class TopologyTestDriverTest {
         return builder.build(config);
     }
 
-    @ParameterizedTest(name = "eosEnabled = {0}, methodName = {1}")
-    @MethodSource(value = "parameters")
-    public void test(final boolean eosEnabled, final String methodName) throws InvocationTargetException, IllegalAccessException {
-        config = mkProperties(mkMap(
-            mkEntry(StreamsConfig.APPLICATION_ID_CONFIG, "test-TopologyTestDriver"),
-            mkEntry(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getAbsolutePath())
-        ));
-        if (eosEnabled) config.setProperty(StreamsConfig.PROCESSING_GUARANTEE_CONFIG, StreamsConfig.EXACTLY_ONCE);
-
-        // run specify test
-        Arrays.stream(TopologyTestDriverTest.class.getDeclaredMethods())
-                .filter(m -> m.getName().equals(methodName))
-                .findAny()
-                .get()
-                .invoke(this);
-    }
-
-    private void shouldNotRequireParameters() {
+    @ParameterizedTest(name = "Eos enabled = {0}")
+    @MethodSource("parameters")
+    public void shouldNotRequireParameters(final boolean eosEnabled) {
+        setProperties(eosEnabled);
         new TopologyTestDriver(setupSingleProcessorTopology(), new Properties());
     }
 
-    private void shouldInitProcessor() {
+    @ParameterizedTest(name = "Eos enabled = {0}")
+    @MethodSource("parameters")
+    public void shouldInitProcessor(final boolean eosEnabled) {
+        setProperties(eosEnabled);
         testDriver = new TopologyTestDriver(setupSingleProcessorTopology(), config);
         assertTrue(mockProcessors.get(0).initialized);
     }
 
-    private void shouldCloseProcessor() {
+    @ParameterizedTest(name = "Eos enabled = {0}")
+    @MethodSource("parameters")
+    public void shouldCloseProcessor(final boolean eosEnabled) {
+        setProperties(eosEnabled);
         testDriver = new TopologyTestDriver(setupSingleProcessorTopology(), config);
 
         testDriver.close();
@@ -489,7 +469,10 @@ public class TopologyTestDriverTest {
         testDriver = null;
     }
 
-    private void shouldThrowForUnknownTopic() {
+    @ParameterizedTest(name = "Eos enabled = {0}")
+    @MethodSource("parameters")
+    public void shouldThrowForUnknownTopic(final boolean eosEnabled) {
+        setProperties(eosEnabled);
         testDriver = new TopologyTestDriver(new Topology());
         assertThrows(
             IllegalArgumentException.class,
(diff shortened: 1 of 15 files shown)


streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
```diff
@@ -125,25 +124,20 @@
...
+        return Stream.of(true, false).map(Arguments::of);
     }
 
-    public TopologyTestDriverTest(final boolean eosEnabled) {
+    private void setProperties(final boolean eosEnabled) {
         config = mkProperties(mkMap(
-            mkEntry(StreamsConfig.APPLICATION_ID_CONFIG, "test-TopologyTestDriver"),
-            mkEntry(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getAbsolutePath())
+                mkEntry(StreamsConfig.APPLICATION_ID_CONFIG, "test-TopologyTestDriver"),
+                mkEntry(StreamsConfig.STATE_DIR_CONFIG, TestUtils.tempDirectory().getAbsolutePath())
```
ijuma: Indentation change unnecessary.

streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
```diff
@@ -1832,8 +1916,10 @@ public void process(final Record<String, String> record) {
         }
     }
 
-    @Test
-    public void shouldRespectTaskIdling() {
+    @ParameterizedTest(name = "Eos enabled = {0}")
+    @MethodSource("parameters")
```
ijuma: There is a huge amount of duplication in this parameterized test given the lack of class based parameterization in JUnit 5. I wonder if there's a better way of converting this test. Thoughts?

ijuma (commented):
Thanks for the PR. Most of it is fine, but `TopologyTestDriverTest` doesn't look great, see below.

streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
chia7712: see https://github.com/apache/kafka/pull/9856/commits/3a3c215b9081974b642e8ab5d8cf474e797c7e05

ijuma: The downside of this non standard naming approach is that one may end up not running some tests and it would be pretty hard to spot. What if we used `EnumSource` and had an enum `EosEnabled` with two values: `True` and `False`?

chia7712: > What if we used EnumSource and had an enum EosEnabled with two values: True and False?

I assumed that we don't like duplication like ```@ParameterizedTest``` and ```XXXSource``` :)

Will apply ```EnumSource```.

ijuma: I couldn't find a way to avoid duplication and not make it brittle. If you can think of one, that would be even better.

chia7712: I feel your idea (```EnumSource```) is the best one before Junit 5 supports to add parameter at class level.

ijuma: Also, I was hoping that by using `EnumSource` with an appropriately named enum, you can avoid the string in `ParameterizedTest`. In that case, the duplicated bits are:
1. The `EnumSource` line.
2. The enum parameter in each test.

Both are strongly typed.

chia7712: > In that case, the only duplication is the EnumSource line, which is strongly typed.

oh, I use ```ValueSource``` with boolean type instead. It is strongly typed as well.

ijuma: But `ValueSource` won't make it clear what `true`/`false` mean, right? That's the main advantage of using the `enum`, we can make it clear what it means.

chia7712: > But ValueSource won't make it clear what true/false mean, right? That's the main advantage of using the enum, we can make it clear what it means.

you are right. will copy that!

streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
```diff
@@ -448,29 +441,37 @@ private Topology setupTopologyWithInternalTopic(final String firstTableName,
         return builder.build(config);
     }
 
-    @Test
-    public void shouldNotRequireParameters() {
+    @ParameterizedTest(name = "{displayName} - {arguments}")
```
ijuma: What happens if the default `name` is used?

chia7712: ```@ParameterizedTest```
```
TopologyTestDriverTest > [1] ENABLED PASSED

TopologyTestDriverTest > [2] DISABLED PASSED

TopologyTestDriverTest > [1] ENABLED PASSED

TopologyTestDriverTest > [2] DISABLED PASSED
```

```@ParameterizedTest(name = "{displayName} - {arguments}")```
```
TopologyTestDriverTest > shouldCloseProcessor(EosMode) - ENABLED PASSED

TopologyTestDriverTest > shouldCloseProcessor(EosMode) - DISABLED PASSED

TopologyTestDriverTest > shouldFeedStoreFromGlobalKTable(EosMode) - ENABLED PASSED

TopologyTestDriverTest > shouldFeedStoreFromGlobalKTable(EosMode) - DISABLED PASSED
```

Comment (ijuma) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: Seems like we should maybe name the enum values: `EOS_ENABLED` and `EOS_DISABLED`, then the output will be good automatically. What do you think? Btw, I am being a nitpicky here because once we have a good pattern, we can apply it in other places.

Comment (chia7712) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: > EOS_ENABLED and EOS_DISABLED

make sense. will copy that.

Comment (chia7712) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]:
```
TopologyTestDriverTest > shouldCloseProcessor(EosMode) - EOS_ENABLED PASSED

TopologyTestDriverTest > shouldCloseProcessor(EosMode) - EOS_DISABLED PASSED

TopologyTestDriverTest > shouldFeedStoreFromGlobalKTable(EosMode) - EOS_ENABLED PASSED

TopologyTestDriverTest > shouldFeedStoreFromGlobalKTable(EosMode) - EOS_DISABLED PASSED
```

Comment (ijuma) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: I think you can remove the `name` part of `ParamterizedTest` now.

Comment (chia7712) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: copy that!

Comment (chia7712) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: we can't omit the ```name```.

oracle docs:
`It is permissible to omit the element name and equals sign (=) in a single-element annotation whose element name is value`

Comment (ijuma) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: You definitely can, I've done it in many tests in a local branch.

Comment (chia7712) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: I rewrite it to ```@ParameterizedTest("{displayName} - {arguments}")``` and then I get following error.

```
/home/chia7712/kafka/streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java:1918: error: cannot find symbol
    @ParameterizedTest("{displayName} - {arguments}")
                       ^
  symbol:   method value()
  location: @interface ParameterizedTest

```

Comment (ijuma) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: I'm suggesting `@ParameterizedTest`, that's it.

Comment (chia7712) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: The output of ```@ParameterizedTest``` is not readable to me so I prefer ```@ParameterizedTest(name = "{displayName} - {arguments}")```. 
```
TopologyTestDriverTest > [1] EOS_ENABLED PASSED

TopologyTestDriverTest > [2] EOS_DISABLED PASSED

TopologyTestDriverTest > [1] EOS_ENABLED PASSED

TopologyTestDriverTest > [2] EOS_DISABLED PASSED

TopologyTestDriverTest > [1] EOS_ENABLED PASSED
```

Comment (ijuma) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: Interesting, I've been testing this with IntelliJ and it seems to handle it differently.

Comment (ijuma) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: I looked into the reasoning for this choice and it's explained here:

https://github.com/junit-team/junit5/issues/1496#issuecomment-404178095

Seems like we are working around the intended design. It would be better to adjust the Gradle config to output the test display name too. (not sure if it's supported atm)

Comment (chia7712) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: Overriding display name by gradle config may cause trouble when the display name from test case is well-designed. Another way is that we can create custom parameterized annotation by junit 5 extension to offer readable display name by default. However, it seems to me that can be addressed in another PR.

I'm fine to use ```@ParameterizedTest``` as junit reports shows both method name and argument name.

Comment (ijuma) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: It looks like a bug to me, personally. Check the Gradle test output html:

Screenshot from 2021-01-12 13-00-01

IMO, the text output should output the method name or display name too.

Comment (chia7712) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: It seems to me there are three ways to deal with the repeated magic string (```name = "{displayName} - {arguments}```) in ```ParameterizedTest```

1. write custom ```TestTemplateInvocationContextProvider``` (junit 5 extension). However, the current implementation (```ParameterizedTestExtension```) is a heavy implementation and so it could bring some works to us.
1. junit 5 has to support ```ParameterizedTest``` at class level: https://github.com/junit-team/junit5/issues/878
1. file a PR to junit 5 to make ```ParameterizedTest``` have "long display name" by default , and so we can write following code ```@ParameterizedTest(name = ParameterizedTest.LONG_DISPLAY_NAME)```

At any rate, ```@ParameterizedTest(name = "{displayName} - {arguments}")``` is workaround to us currently and we can improve it in the future If there is any more simple solution for junit 5. 

@ijuma WDYT?

Comment (chia7712) [reply streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java]: I follow your solution (#9874) to remove the ```ParameterizedTest```. see https://github.com/apache/kafka/pull/9856/commits/7480d2692301424e2d9b72e33c1c9b7c34d3c376

File: streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java [hunk 4]
```diff
@@ -890,8 +875,10 @@ public void shouldProcessConsumerRecordList() {
         assertThat(record, equalTo(expectedResult));
     }
 
+
```
ijuma: Is this intentional?

streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
```diff
@@ -890,8 +875,10 @@ public void shouldProcessConsumerRecordList() {
         assertThat(record, equalTo(expectedResult));
     }
 
+    
     @Test
     public void shouldForwardRecordsFromSubtopologyToSubtopology() {
+
```
ijuma: Is this intentional?

streams/test-utils/src/test/java/org/apache/kafka/streams/TopologyTestDriverTest.java
```diff
@@ -1562,14 +1549,14 @@ public void process(final Record<String, Long> record) {
...
-            Assert.assertEquals(1L, testDriver.getKeyValueStore("storeProcessorStore").get("a"));
+            assertEquals(1L, testDriver.getKeyValueStore("storeProcessorStore").get("a"));
         }
 
 
         try (final TopologyTestDriver testDriver = new TopologyTestDriver(topology, config)) {
             assertNull(
-                "Closing the prior test driver should have cleaned up this store and value.",
-                testDriver.getKeyValueStore("storeProcessorStore").get("a")
+                testDriver.getKeyValueStore("storeProcessorStore").get("a"),
```
ijuma: This can probably go to the previous line now.

ijuma (approved):
LGTM, a few nits below, no re-review required.

PR #9858, opened 2021-01-11
KAFKA-12173 Migrate streams:streams-scala module to JUnit 5

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-12173

1. replace ```org.junit.Assert``` by ```org.junit.jupiter.api.Assertions``` 
1. replace ```org.junit``` by ```org.junit.jupiter.api```
1. replace ```Before``` by ```BeforeEach```
1. replace ```After``` by ```AfterEach``` 
1. remove  ```ExternalResource``` from all scala modules
1. add explicit ```AfterClass```/```BeforeClass``` to stop/start ```EmbeddedKafkaCluster```

Noted that this PR does not migrate stream module to junit 5 so it does not introduce callback of junit 5 to deal with beforeAll/afterAll. The next PR of migrating stream module can replace explicit beforeAll/afterAll by junit 5 extension. Or we can keep the beforeAll/afterAll if it make code more readable.


ijuma:
For the Stream modules, we should get one of the committers that is more involved in that part. @mjsax and @vvcephei are two possible candidates.

mjsax:
This PR is _huge_ and contains a lot of reformatting. Is this required? Very tedious to review. Why are there so many Java class changes -- this PR does not migrate Java tests to JUnit5? Could this PR be split into multiple smaller ones to make it easier to review?

vvcephei (approved):
Woah, thanks for all this work, @chia7712 !

I agree it's a little on the large side for a PR, but it looks like it was necessary, since changing EmbeddedKafkaCluster dragged in all the integration tests.

I'm assuming we had to ditch Scalatest stuff like Flatspec because the JUnitRunner isn't compatible with JUnit5? It's fine with me either way. We were writing those Scala tests in exactly the same way we write the Java ones, might as well use the same testing framework, too.

All in all, these changes look good to me. Please feel free to merge whenever you're satisfied with the test results, and thanks again, @chia7712 !

chia7712:
@mjsax Sorry for lazy response. I did not notice your feedback until @vvcephei touch this PR :(

> This PR is huge and contains a lot of reformatting. Is this required? Very tedious to review. Why are there so many Java class changes -- this PR does not migrate Java tests to JUnit5? Could this PR be split into multiple smaller ones to make it easier to review?

We are migrate code to Junit 5 by module in order to avoid huge patch (although this patch is still big ...). The root cause of making this tedious patch is that `stream-scale` depends on `stream` module (`EmbeddedKafkaCluster`, to be exact). Unfortunately, the Junit APIs used by `EmbeddedKafkaCluster` belong to Junit 4 and those APIs are not compatible to Junit 5. That is why this patch has to reformat a lot of java code of streams module.

chia7712:
@vvcephei Thanks for your feedback. I will fix conflicting files.

chia7712:
@ijuma @mjsax I'd like to merge this PR and then I can migrate `EmbeddedKafkaCluster`(streams module) to junit 5 in order to remove duplicate code. It would be helpful if you could give another look. thanks!

ijuma:
@chia7712 If @vvcephei reviewed it, I think you can go ahead and merge.

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -75,9 +76,19 @@
...
     public static final EmbeddedKafkaCluster CLUSTER = new EmbeddedKafkaCluster(1);
 
+    @BeforeClass
+    public static void startCluster() throws IOException {
+        CLUSTER.start();
+    }
+
+    @AfterClass
+    public static void closeCluster() {
+        CLUSTER.stop();
```
ableegoldman, line 88: Hey @chia7712 , sorry for never bothering to take a look at this PR until just now, but I had a question about this. I know it was necessary to remove the `ExternalResource` feature that used to be responsible for calling `start()` and `stop()` for us in the integration tests since `@ClassRule` was removed in JUnit5, but that was really quite a loss since this now leaves us vulnerable to
1) resource leaks due to forgetting to clean up the EmbeddedKafkaCluster in an integration test, or doing so in an incorrect way (eg such that a test failure might skip the cleanup stage, a mistake that we've certainly encountered in our tests in the past)
2) breaking compatibility of integration tests across older branches, so that if we ever need to backport a fix that includes an integration test -- which many will/should do -- we can easily break the build of older branches by accident. See for example [#11257](https://github.com/apache/kafka/pull/11257): aka the reason I started digging into this 🙂 . Even if we remember to fix this during the backport, it's just an extra hassle.

Now I'm certainly not an expert in all things JUnit, but a cursory glance suggests we can replicate the old behavior in which the EmbeddedKafkaCluster is automatically started/stopped without the need for this `@Before/AfterClass` boilerplate code in every integration test. I believe it's possible to do so using the `@Extension/ExtendWith` annotations. Can we try to port the EmbeddedKafkaCluster back to an automated lifecycle with these so we can clean up the Streams integration tests?

cc @ijuma @vvcephei who may be more familiar with these constructs and how/when/why to use them

ableegoldman: cc also @showuon -- would you be interested in taking a shot at this?

showuon: I'll check it! :)

showuon: https://issues.apache.org/jira/browse/KAFKA-13230 is created for this issue. Need more discussion to make sure what we really want.

PR #9861, opened 2021-01-11
MINOR: Modify unnecessary access specifiers

highluck:


highluck:
@chia7712 
May I ask for a review of this PR?

highluck:
Ping
@chia7712
May I ask for a review of this PR?

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
```diff
@@ -104,7 +104,7 @@
     private final RocksDBMetricsRecorder metricsRecorder;
 
     // visible for testing
```
chia7712: Is this comment still valid?

chia7712 (commented):
@highluck thanks for your patch. one comment is left.

streams/src/main/java/org/apache/kafka/streams/state/internals/RocksDBStore.java
highluck: Thanks for the comment.
That part doesn't seem necessary.
And I think you can clear that field.

highluck:
@chia7712
Thank you for the comment Code was modified

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredTimestampedKeyValueStore.java
```diff
@@ -96,7 +96,7 @@ public boolean putIfDifferentValues(final K key,
     public static class RawAndDeserializedValue<ValueType> {
```
chia7712: It seems this class is used by testing only. Could you change it from public to package-private?

chia7712 (commented):
@highluck thanks for your patch. one minor comment is left.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredTimestampedKeyValueStore.java
highluck: @chia7712 The code has been modified!

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredTimestampedKeyValueStore.java
```diff
@@ -58,7 +58,7 @@
     }
 
 
-    public RawAndDeserializedValue<V> getWithBinary(final K key) {
+    RawAndDeserializedValue<V> getWithBinary(final K key) {
```
chia7712: This change may be overkill as this is not an inner class.

streams/src/main/java/org/apache/kafka/streams/state/internals/MeteredTimestampedKeyValueStore.java
```diff
@@ -70,9 +70,9 @@
         }
     }
 
-    public boolean putIfDifferentValues(final K key,
-                                        final ValueAndTimestamp<V> newValue,
-                                        final byte[] oldSerializedValue) {
+    boolean putIfDifferentValues(final K key,
```
chia7712: ditto

chia7712 (commented):
@highluck thanks for updating code. a couple of comments. please take a look.

highluck:
@chia7712 
Thank you for the comment Code was modified

PR #9862, opened 2021-01-11
MINOR: Fix error message in SnapshotWriter.java

tang7526:
before
```
            String message = String.format(
                "Append not supported. Snapshot is already frozen: id = {}.",
                snapshot.snapshotId()
            );
```

after
```
            String message = String.format(
                "Append not supported. Snapshot is already frozen: id = '%s'.",
                snapshot.snapshotId()
            );
```

### DIFF
--- a/raft/src/main/java/org/apache/kafka/snapshot/SnapshotWriter.java
+++ b/raft/src/main/java/org/apache/kafka/snapshot/SnapshotWriter.java
@@ -107,8 +107,8 @@ final public class SnapshotWriter<T> implements Closeable {
     public void append(List<T> records) throws IOException {
         if (snapshot.isFrozen()) {
             String message = String.format(
-                "Append not supported. Snapshot is already frozen: id = {%s}.",
-                snapshot.snapshotId().toString()
+                "Append not supported. Snapshot is already frozen: id = '%s'.",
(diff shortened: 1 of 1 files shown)


raft/src/main/java/org/apache/kafka/snapshot/SnapshotWriter.java
```diff
@@ -107,7 +107,7 @@ public boolean isFrozen() {
     public void append(List<T> records) throws IOException {
         if (snapshot.isFrozen()) {
             String message = String.format(
-                "Append not supported. Snapshot is already frozen: id = {}.",
+                "Append not supported. Snapshot is already frozen: id = {%s}.",
```
chia7712: Is ```'%s'``` more suitable (if you want to highlight the object string) ?

chia7712 (commented):
@tang7526 Thanks for your patch.

raft/src/main/java/org/apache/kafka/snapshot/SnapshotWriter.java
tang7526: oh, I didn't notice that it's a object.

chia7712 (approved):
@tang7526 Thanks for your patch. LGTM

PR #9866, opened 2021-01-11
KAFKA-12180: Implement the KIP-631 message generator changes

cmccabe:
* Implement the uint16 type
* Implement MetadataRecordType and MetadataJsonConverters

### DIFF
--- a/generator/src/main/java/org/apache/kafka/message/MessageSpecType.java
+++ b/generator/src/main/java/org/apache/kafka/message/MessageSpecType.java
@@ -30,5 +30,8 @@ public enum MessageSpecType {
     HEADER,
 
     @JsonProperty("data")
-    DATA;
+    DATA,
+
+    @JsonProperty("metadata")
+    METADATA;
 }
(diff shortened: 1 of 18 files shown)


generator/src/main/java/org/apache/kafka/message/MessageSpecType.java
```diff
@@ -30,5 +30,8 @@
     HEADER,
 
     @JsonProperty("data")
-    DATA;
+    DATA,
+
+    @JsonProperty("metadata")
```
hachikuji, line 35: Might be helpful to have a short description here (or somewhere else if you can think of a better location) which explains that this type is reserved for the controller record types.

hachikuji (approved):
Thanks, LGTM. Minor suggestion to document the metadata spec type.

generator/src/main/java/org/apache/kafka/message/MessageSpecType.java
cmccabe: I added a comment

cmccabe:
merged, thanks!

PR #9867, opened 2021-01-11
KAFKA-10835: Replace Runnable and Callable overrides with lambdas in Connect

levzem:
Signed-off-by: Lev Zemlyanov <lev@confluent.io>

comment to ping reviewers.*

change interfaces to lambdas

unit as this is purely stylistic change

### DIFF
--- a/connect/api/src/test/java/org/apache/kafka/connect/data/ConnectSchemaTest.java
+++ b/connect/api/src/test/java/org/apache/kafka/connect/data/ConnectSchemaTest.java
@@ -69,7 +69,8 @@ public class ConnectSchemaTest {
 
     @Test
     public void testFieldsOnlyValidForStructs() {
-        assertThrows(DataException.class, Schema.INT8_SCHEMA::fields);
+        assertThrows(DataException.class,
+            Schema.INT8_SCHEMA::fields);
     }
 
     @Test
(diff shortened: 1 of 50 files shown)


kkonstantine (commented):
That's a great reduction of boiler plate.

Seeing the changes were quite a few I have to say we might need separate PRs for backporting to previous release branches. But let's see after we merge this one. 
The failures were not relevant (jdk 11 passed, 8 and 15 failed on kafka.api and a streams test). I triggered another run.

ijuma:
@kkonstantine We should not backport changes like this to older branches IMO. We should stick to important bug fixes for older branches.

kkonstantine:
@ijuma what are trying to achieve here with backporting is to have a similar basis for other future bugfixes that _will have to_ be applied to both the new and the old branches. Otherwise the conflicts will be a lot and possibly frequent.  
The purpose of these changes is to have very low risk. Wdyt?

ijuma:
I understood the intent, but I don't think it's worth it, personally. Backports are not meant to be frequent, so I don't see why there would be frequent conflicts. For clients and core, we never backport readability improvements and it has generally not been an issue.

kkonstantine:
I'm a bit surprise to hear that this doesn't come up with cherry-picks. In Connect we backport any bugfix, usually 2-3 releases back. I believe it would be a challenge to cherry-pick such a fix if it didn't map to similar lines. @rhauch had recommended similar cleanups and I think he also shared the view that backporting would be better than not.

kkonstantine (approved):
LGTM

Few test failures that are not relevant. Tests are all passing locally. 

Merging only on `trunk` due to the extent of the changes and to eliminate risk on patch versions. 

Thanks @levzem !

PR #9871, opened 2021-01-12
KAFKA-12161; Support raft observers with optional id

hachikuji:
We would like to be able to use `KafkaRaftClient` for tooling/debugging use cases. For this, we need the localId to be optional so that the client can be used more like a consumer. This is already supported in the `Fetch` protocol by setting `replicaId=-1`, which the Raft implementation checks for. We just need to alter `QuorumState` so that the `localId` is optional. The main benefit of doing this is that it saves tools the need to generate an arbitrary id (which might cause conflicts given limited Int32 space) and it lets the leader avoid any local state for these observers (such as `ReplicaState` inside `LeaderState`).

### DIFF
--- a/raft/src/main/java/org/apache/kafka/raft/QuorumState.java
+++ b/raft/src/main/java/org/apache/kafka/raft/QuorumState.java
@@ -73,7 +73,7 @@ import java.util.stream.Collectors;
  *
  */
 public class QuorumState {
-    public final int localId;
+    private final OptionalInt localId;
     private final Time time;
     private final Logger log;
     private final QuorumStateStore store;
@@ -84,7 +84,7 @@ public class QuorumState {
 
     private volatile EpochState state;
 
-    public QuorumState(int localId,
+    public QuorumState(OptionalInt localId,
                        Set<Integer> voters,
                        int electionTimeoutMs,
                        int fetchTimeoutMs,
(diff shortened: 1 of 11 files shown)


raft/src/main/java/org/apache/kafka/raft/QuorumState.java
```diff
@@ -196,7 +203,15 @@ public void initialize(OffsetAndEpoch logEndOffsetAndEpoch) throws IOException,
     }
 
     public Set<Integer> remoteVoters() {
-        return voters.stream().filter(voterId -> voterId != localId).collect(Collectors.toSet());
+        return voters.stream().filter(voterId -> voterId != localIdOrNil()).collect(Collectors.toSet());
+    }
+
+    public int localIdOrNil() {
```
ijuma: `localIdOrSentinel` would be more accurate, I think.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
```diff
@@ -2182,6 +2182,46 @@ public void testHandleCommitCallbackFiresInCandidateState() throws Exception {
         assertEquals(OptionalInt.empty(), secondListener.currentClaimedEpoch());
     }
 
+    @Test
+    public void testObserverFetchWithNoLocalId() throws Exception {
+        // When no `localId` is defined, the client will behave as an observer.
```
abbccdda, line 2187: Wondering whether we could copy this comment to the raft client localId field.

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
```diff
@@ -125,9 +125,16 @@ public void initialize(OffsetAndEpoch logEndOffsetAndEpoch) throws IOException,
...
-                ". Check if the quorum configuration is up to date, " +
-                "or wipe out the local state file if necessary");
+                + " is different from the voter set read from the state file: " + election.voters()
+                + ". Check if the quorum configuration is up to date, "
+                + "or wipe out the local state file if necessary");
+        } else if (election.hasVoted() && !isVoter()) {
+            String localIdDescription = localId.isPresent() ?
+                localId.getAsInt() + " is not a voter" :
+                "is undefined";
+            throw new IllegalStateException("Initialized quorum state " + election
```
abbccdda, line 135: So a node could not transit from follower to observer though restarts?

hachikuji: I thought I would start out strict and relax later if needed. Worst case, the user would just remove or edit the `quorum-state` file.

raft/src/test/java/org/apache/kafka/raft/KafkaRaftClientTest.java
hachikuji: Makes sense. I will move it.

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
abbccdda: Fair enough, we could relax later when we experiment out the static quorum changes.

abbccdda (approved):

raft/src/main/java/org/apache/kafka/raft/QuorumState.java
hachikuji: Fair enough. I guess "nil" suggests zero.

ijuma: Yeah, that would be the closest for primitive types.

PR #9872, opened 2021-01-12
KAFKA-10759: ARM support for Kafka

xiao-penglei:
Add the stage of arm ci in pipeline of jenkins.

comment to ping reviewers.*

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -160,5 +160,23 @@ pipeline {
         }
       }
     }
+    stage("Arm Build") {
+      agent { label 'arm4' }
+      options {
+        timeout(time: 8, unit: 'HOURS')
+        timestamps()
+      }
+      environment {
+        SCALA_VERSION=2.12
+      }
+      steps {
+        catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
+          setupGradle()
+          doValidation()
+          doTest()
+          tryStreamsArchetype()
+        }
+      }
+    }
   }
 }


ijuma:
Do we have to run the full test suite or would it be sufficient to run the unit tests only?

xiao-penglei:
I think it is better that running a full test suite under jdk8 branch first.

Ismael Juma <notifications@github.com> 于2021年1月15日周五 下午9:54写道：

guyuqi:
It's awesome to see Kafka would add Arm support to CI matrix.

liusheng:
Thanks for bringing this up, hope more and more Big-data projects can run on ARM platform :)

Jenkinsfile
```diff
@@ -160,5 +160,23 @@ pipeline {
         }
       }
     }
+    stage("Arm Build") {
+      agent { label 'arm4' }
+      options {
+        timeout(time: 8, unit: 'HOURS')
+        timestamps()
+      }
+      environment {
```
junrao, line 169: Do we need to specify the jdk version since scala 2.12 only works on jdk 8?

junrao (commented):
@xiao-penglei : Thanks for the PR. Did the "Arm Build" get triggered in jenkins? I can't seem to find it.

Jenkinsfile
xiao-penglei: @junrao Yeah, I have installed the openjdk-8u252-b09 that the version is same as jdk_1.8_latest of apache infra. The reason is that I could not find appropriate version of jdk that can run on arm platform from the jenkins software of apache infra. [infra jdk matrix](https://cwiki.apache.org/confluence/display/INFRA/JDK+Installation+Matrix) . I have also installed the maven-3.6.3 that version is same as maven_3_latest of apache infra. I will take time to ensure the versions of jdk and maven are consistent with those of apache infra. In most cases, the arm4 node is just for kafka arm ci.

xiao-penglei:
> @xiao-penglei : Thanks for the PR. Did the "Arm Build" get triggered in jenkins? I can't seem to find it.

@junrao When you create a Jenkins pipline, there are two configuration methods of Jenkinsfile source. One is to use the Jenkinsfile of repository, and the other way is to use the static Jenkinsfile. I do not know the configuration of the jenkis pipline of Kafka because I do not have permission. The "Arm build" can be found only after the Jenkinsfile used in the jenkins pipline is modified.  The Jenkinsfile modification that I submitted was tested locally. Now we need modify the Jenkinsfile used in the jenkins pipline.  Maybe @ijuma can help us. Thank you.

junrao:
@xiao-penglei : Thanks. Is the PR ready to be committed then?

@mumrah, since you helped set up the jenkins pipeline recently, could you advise on the next step needed? Thanks.

ijuma:
What's the underlying ARM CI capacity? How long does it take to run? I don't think we want to make the PR process slower. As I suggested before, let's start by running `unitTest` for the ARM build. If that works well, we can consider extending it to be the full test suite.

mumrah:
I believe we have Jenkins configured so it will only take Jenkinsfile changes in a PR if they are from a committer. I'll open a PR shortly to see if this works.

@ijuma looks like a few different ARM images are provided by Apache Infra: https://cwiki.apache.org/confluence/display/INFRA/ci-builds.apache.org

mumrah:
Ok, the ARM build compiles at least https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-9992/1/console

I'll try running the unit test suite next.

xiao-penglei:
> @xiao-penglei : Thanks. Is the PR ready to be committed then?
> 
> @mumrah, since you helped set up the jenkins pipeline recently, could you advise on the next step needed? Thanks.

@junrao I prefer the PR submitted by @mumrah .[KAFKA-10759 Add ARM build stage #9992](https://github.com/apache/kafka/pull/9992).  Thanks

xiao-penglei:
> What's the underlying ARM CI capacity? How long does it take to run? I don't think we want to make the PR process slower. As I suggested before, let's start by running `unitTest` for the ARM build. If that works well, we can consider extending it to be the full test suite.

@ijuma The arm4 node has 8 cpus and 16G memory. I run the full test suite locally and compare the running time of of build stage of jdk8 on x86 and arm. Almost no difference. If we parallelize the arm build stage with the original stage, the entire CI will not slow down. Again, I agree with you very much that let's start by running unitTest for the ARM build. If that works well, we can consider extending it to be the full test suite. Thanks

mumrah:
This was merged to trunk as https://github.com/apache/kafka/commit/f40a82e05440e579c69363dc62c2dd9c8520eb02

PR #9875, opened 2021-01-12
MINOR: tighten condition for logging StreamThread summary to avoid spam under low traffic

ableegoldman:
Noticed these messages flooding the logs on several integration tests. Since it doesn't really provide any useful information, we may as well skip the logging when there are no new records to be processed.

Should be cherrypicked back to the 2.7 branch

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -749,9 +749,11 @@ public class StreamThread extends Thread {
             // multiple iterations with reasonably large max.num.records and hence is less vulnerable to outliers
             taskManager.recordTaskProcessRatio(totalProcessLatency, now);
 
-            log.info("Processed {} total records, ran {} punctuators, and committed {} total tasks " +
-                        "for active tasks {} and standby tasks {}",
-                     totalProcessed, totalPunctuated, totalCommitted, taskManager.activeTaskIds(), taskManager.standbyTaskIds());
+            // Don't log summary if no new records were processed to avoid spamming logs for low-traffic topics
+            if (totalProcessed > 0 || totalPunctuated > 0 || totalCommitted > 0) {
+                log.info("Processed {} total records, ran {} punctuators, and committed {} total tasks",
+                         totalProcessed, totalPunctuated, totalCommitted);
+            }
         }
 
         now = time.milliseconds();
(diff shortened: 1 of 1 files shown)


ableegoldman:
call for a quick review from any of @cadonna @lct45 @wcarlson5

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -749,9 +749,11 @@ void runOnce() {
             // multiple iterations with reasonably large max.num.records and hence is less vulnerable to outliers
             taskManager.recordTaskProcessRatio(totalProcessLatency, now);
 
-            log.info("Processed {} total records, ran {} punctuators, and committed {} total tasks " +
-                        "for active tasks {} and standby tasks {}",
-                     totalProcessed, totalPunctuated, totalCommitted, taskManager.activeTaskIds(), taskManager.standbyTaskIds());
+            // Don't log summary if no new records were processed to avoid spamming logs for low-traffic topics
+            if (totalProcessed > 0 || totalPunctuated > 0 || totalCommitted > 0) {
+                log.info("Processed {} total records, ran {} punctuators, and committed {} total tasks",
```
wcarlson5, line 754: Maybe we do this else where but can we log at DEBUG if there is nothing done and there is tasks assigned to the thread? It might be good to be able to confirm that its not getting stuck here or its actually not polling anything.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -824,7 +826,9 @@ private long pollPhase() {
         final long pollLatency = advanceNowAndComputeLatency();
 
         final int numRecords = records.count();
-        log.info("Main Consumer poll completed in {} ms and fetched {} records", pollLatency, numRecords);
+        if (numRecords > 0) {
```
wcarlson5, line 829: The above probably doesn't need to be applied here as well, i don't think

wcarlson5 (commented):
This seems like a good change. I just have one minor concern

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
lct45: I have a similar question here, other than that LGTM

lct45 (approved):
Assuming we log error elsewhere when no progress is being made, LGTM. Thanks @ableegoldman

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
ableegoldman: Good thought, but don't worry, we already do. John added some pretty thorough logging at the DEBUG level after an escalation a while back

ableegoldman:

PR #9878, opened 2021-01-13
KAFKA-6987: Add KafkaFuture.toCompletionStage()

tombentley:
See [KAFKA-6987](https://issues.apache.org/jira/browse/KAFKA-6987) and [KIP-707](https://cwiki.apache.org/confluence/display/KAFKA/KIP-707%3A+The+future+of+KafkaFuture)

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
+++ b/clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
@@ -101,7 +101,7 @@ public abstract class KafkaFuture<T> implements Future<T> {
         KafkaFuture<Void> allOfFuture = new KafkaFutureImpl<>();
         AllOfAdapter<Object> allOfWaiter = new AllOfAdapter<>(futures.length, allOfFuture);
         for (KafkaFuture<?> future : futures) {
-            future.addWaiter(allOfWaiter);
+            future.whenComplete(allOfWaiter);
         }
         return allOfFuture;
     }
@@ -145,7 +145,6 @@ public abstract class KafkaFuture<T> implements Future<T> {
      */
     public abstract KafkaFuture<T> whenComplete(BiConsumer<? super T, ? super Throwable> action);
 
-    protected abstract void addWaiter(BiConsumer<? super T, ? super Throwable> action);
     /**
      * If not already completed, sets the value returned by get() and related methods to the given
      * value.
--- a/clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
+++ b/clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
@@ -16,7 +16,6 @@
  */
 package org.apache.kafka.common;
 
-import org.apache.kafka.common.annotation.InterfaceStability;
 import org.apache.kafka.common.internals.KafkaFutureImpl;
 
 import java.util.Arrays;
@@ -39,7 +38,6 @@ import java.util.concurrent.TimeoutException;
  * {@code KafkaFuture} will fail with an {@code ExecutionException}, whereas a {@code CompletionStage} fails
  * with a {@code CompletionException}.
  */
-@InterfaceStability.Evolving
 public abstract class KafkaFuture<T> implements Future<T> {
     /**
      * A function which takes objects of type A and returns objects of type B.
@@ -196,7 +194,7 @@ public abstract class KafkaFuture<T> implements Future<T> {
      * Returns the result value (or throws any encountered exception) if completed, else returns
      * the given valueIfAbsent.
      */
-    public abstract T getNow(T valueIfAbsent) throws InterruptedException, ExecutionException;
+    public abstract T getNow(T valueIfAbsent) throws ExecutionException;
 
     /**
      * Returns true if this CompletableFuture was cancelled before it completed normally.
--- a/clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
+++ b/clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
@@ -19,22 +19,32 @@ package org.apache.kafka.common;
 import org.apache.kafka.common.annotation.InterfaceStability;
 import org.apache.kafka.common.internals.KafkaFutureImpl;
 
+import java.util.Arrays;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 /**
- * A flexible future which supports call chaining and other asynchronous programming patterns. This will
- * eventually become a thin shim on top of Java 8's CompletableFuture.
+ * A flexible future which supports call chaining and other asynchronous programming patterns.
  *
- * The API for this class is still evolving and we may break compatibility in minor releases, if necessary.
+ * <h3>Relation to {@code CompletionStage}</h3>
+ * <p>It is possible to obtain a {@code CompletionStage} from a
+ * {@code KafkaFuture} instance by calling {@link #toCompletionStage()}.
+ * If converting {@link KafkaFuture#whenComplete(BiConsumer)} or {@link KafkaFuture#thenApply(BaseFunction)} to
+ * {@link CompletableFuture#whenComplete(java.util.function.BiConsumer)} or
+ * {@link CompletableFuture#thenApply(java.util.function.Function)} be aware that the returned
+ * {@code KafkaFuture} will fail with an {@code ExecutionException}, whereas a {@code CompletionStage} fails
+ * with a {@code CompletionException}.
  */
 @InterfaceStability.Evolving
 public abstract class KafkaFuture<T> implements Future<T> {
     /**
      * A function which takes objects of type A and returns objects of type B.
      */
+    @FunctionalInterface
     public interface BaseFunction<A, B> {
         B apply(A a);
     }
@@ -42,52 +52,24 @@ public abstract class KafkaFuture<T> implements Future<T> {
     /**
      * A function which takes objects of type A and returns objects of type B.
      *
-     * Prefer the functional interface {@link BaseFunction} over the class {@link Function}.  This class is here for
-     * backwards compatibility reasons and might be deprecated/removed in a future release.
+     * @deprecated Since Kafka 3.0. Use the {@link BaseFunction} functional interface.
      */
+    @Deprecated
     public static abstract class Function<A, B> implements BaseFunction<A, B> { }
 
     /**
      * A consumer of two different types of object.
      */
+    @FunctionalInterface
     public interface BiConsumer<A, B> {
         void accept(A a, B b);
-    }
-
-    private static class AllOfAdapter<R> implements BiConsumer<R, Throwable> {
-        private int remainingResponses;
-        private KafkaFuture<?> future;
-
-        public AllOfAdapter(int remainingResponses, KafkaFuture<?> future) {
-            this.remainingResponses = remainingResponses;
-            this.future = future;
-            maybeComplete();
-        }
-
-        @Override
-        public synchronized void accept(R newValue, Throwable exception) {
-            if (remainingResponses <= 0)
-                return;
-            if (exception != null) {
-                remainingResponses = 0;
-                future.completeExceptionally(exception);
-            } else {
-                remainingResponses--;
-                maybeComplete();
-            }
-        }
-
-        private void maybeComplete() {
-            if (remainingResponses <= 0)
-                future.complete(null);
-        }
     }
 
     /** 
      * Returns a new KafkaFuture that is already completed with the given value.
      */
     public static <U> KafkaFuture<U> completedFuture(U value) {
-        KafkaFuture<U> future = new KafkaFutureImpl<U>();
+        KafkaFuture<U> future = new KafkaFutureImpl<>();
         future.complete(value);
         return future;
     }
@@ -98,13 +80,44 @@ public abstract class KafkaFuture<T> implements Future<T> {
      * an exception, which one gets returned is arbitrarily chosen.
      */
     public static KafkaFuture<Void> allOf(KafkaFuture<?>... futures) {
-        KafkaFuture<Void> allOfFuture = new KafkaFutureImpl<>();
-        AllOfAdapter<Object> allOfWaiter = new AllOfAdapter<>(futures.length, allOfFuture);
-        for (KafkaFuture<?> future : futures) {
-            future.whenComplete(allOfWaiter);
-        }
-        return allOfFuture;
+        KafkaFutureImpl<Void> result = new KafkaFutureImpl<>();
+        CompletableFuture.allOf(Arrays.stream(futures)
+                .map(kafkaFuture -> {
+                    // Safe since KafkaFuture's only subclass is KafkaFuture for which toCompletionStage()
+                    // always return a CF.
+                    return (CompletableFuture<?>) kafkaFuture.toCompletionStage();
+                })
+                .toArray(CompletableFuture[]::new)).whenComplete((value, ex) -> {
+                    if (ex == null) {
+                        result.complete(value);
+                    } else {
+                        // Have to unwrap the CompletionException which allOf() introduced
+                        result.completeExceptionally(ex.getCause());
+                    }
+                });
+
+        return result;
     }
+
+    /**
+     * Gets a {@code CompletionStage} with the same completion properties as this {@code KafkaFuture}.
+     * The returned instance will complete when this future completes and in the same way
+     * (with the same result or exception).
+     *
+     * <p>Calling {@code toCompletableFuture()} on the returned instance will yield a {@code CompletableFuture},
+     * but invocation of the completion methods ({@code complete()} and other methods in the {@code complete*()}
+     * and {@code obtrude*()} families) on that {@code CompletableFuture} instance will result in
+     * {@code UnsupportedOperationException} being thrown. Unlike a "minimal" {@code CompletableFuture},
+     * the {@code get*()} and other methods of {@code CompletableFuture} that are not inherited from
+     * {@code CompletionStage} will work normally.
+     *
+     * <p>If you want to block on the completion of a KafkaFuture you should use
+     * {@link #get()}, {@link #get(long, TimeUnit)} or {@link #getNow(Object)}, rather then calling
+     * {@code .toCompletionStage().toCompletableFuture().get()} etc.
+     *
+     * @since Kafka 3.0
+     */
+    public abstract CompletionStage<T> toCompletionStage();
 
     /**
      * Returns a new KafkaFuture that, when this future completes normally, is executed with this
--- a/clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
+++ b/clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
@@ -194,7 +194,7 @@ public abstract class KafkaFuture<T> implements Future<T> {
      * Returns the result value (or throws any encountered exception) if completed, else returns
      * the given valueIfAbsent.
      */
-    public abstract T getNow(T valueIfAbsent) throws ExecutionException;
+    public abstract T getNow(T valueIfAbsent) throws InterruptedException, ExecutionException;
 
     /**
      * Returns true if this CompletableFuture was cancelled before it completed normally.
--- a/clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
+++ b/clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.common.internals;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+/**
+ * This internal class exists because CompletableFuture exposes complete(), completeExceptionally() and
+ * other methods which would allow erroneous completion by user code of a KafkaFuture returned from a
+ * Kafka API to a client application.
+ * @param <T> The type of the future value.
+ */
+public class KafkaCompletableFuture<T> extends CompletableFuture<T> {
+
+    public boolean kafkaComplete(T value) {
+        return super.complete(value);
+    }
+
+    public boolean kafkaCompleteExceptionally(Throwable throwable) {
+        return super.completeExceptionally(throwable);
+    }
+
+    @Override
+    public boolean complete(T value) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean completeExceptionally(Throwable ex) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void obtrudeValue(T value) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void obtrudeException(Throwable ex) {
+        throw new UnsupportedOperationException();
+    }
+
+    //@Override // enable once Kafka no longer supports Java 8
+    public <U> CompletableFuture<U> newIncompleteFuture() {
+        return new KafkaCompletableFuture<>();
+    }
+
+    //@Override // enable once Kafka no longer supports Java 8
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier, Executor executor) {
+        throw new UnsupportedOperationException();
+    }
+
+    //@Override // enable once Kafka no longer supports Java 8
+    public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier) {
+        throw new UnsupportedOperationException();
+    }
+
+    //@Override // enable once Kafka no longer supports Java 8
+    public CompletableFuture<T> completeOnTimeout(T value, long timeout, TimeUnit unit) {
+        throw new UnsupportedOperationException();
+    }
+}
--- a/clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
+++ b/clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
@@ -29,32 +29,44 @@ import java.util.function.Supplier;
  */
 public class KafkaCompletableFuture<T> extends CompletableFuture<T> {
 
+    /**
+     * Completes this future normally. For internal use by the Kafka clients, not by user code.
+     * @param value the result value
+     * @return {@code true} if this invocation caused this CompletableFuture
+     * to transition to a completed state, else {@code false}
+     */
     boolean kafkaComplete(T value) {
         return super.complete(value);
     }
 
+    /**
+     * Completes this future exceptionally. For internal use by the Kafka clients, not by user code.
+     * @param throwable the exception.
+     * @return {@code true} if this invocation caused this CompletableFuture
+     * to transition to a completed state, else {@code false}
+     */
     boolean kafkaCompleteExceptionally(Throwable throwable) {
         return super.completeExceptionally(throwable);
     }
 
     @Override
     public boolean complete(T value) {
-        throw new UnsupportedOperationException();
+        throw erroneousCompletionException();
     }
 
     @Override
     public boolean completeExceptionally(Throwable ex) {
-        throw new UnsupportedOperationException();
+        throw erroneousCompletionException();
     }
 
     @Override
     public void obtrudeValue(T value) {
-        throw new UnsupportedOperationException();
+        throw erroneousCompletionException();
     }
 
     @Override
     public void obtrudeException(Throwable ex) {
-        throw new UnsupportedOperationException();
+        throw erroneousCompletionException();
     }
 
     //@Override // enable once Kafka no longer supports Java 8
@@ -64,16 +76,20 @@ public class KafkaCompletableFuture<T> extends CompletableFuture<T> {
 
     //@Override // enable once Kafka no longer supports Java 8
     public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier, Executor executor) {
-        throw new UnsupportedOperationException();
+        throw erroneousCompletionException();
     }
 
     //@Override // enable once Kafka no longer supports Java 8
     public CompletableFuture<T> completeAsync(Supplier<? extends T> supplier) {
-        throw new UnsupportedOperationException();
+        throw erroneousCompletionException();
     }
 
     //@Override // enable once Kafka no longer supports Java 8
     public CompletableFuture<T> completeOnTimeout(T value, long timeout, TimeUnit unit) {
-        throw new UnsupportedOperationException();
+        throw erroneousCompletionException();
+    }
+
+    private UnsupportedOperationException erroneousCompletionException() {
+        return new UnsupportedOperationException("User code should not complete futures returned from Kafka clients");
     }
 }
--- a/clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
+++ b/clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
@@ -29,11 +29,11 @@ import java.util.function.Supplier;
  */
 public class KafkaCompletableFuture<T> extends CompletableFuture<T> {
 
-    public boolean kafkaComplete(T value) {
+    boolean kafkaComplete(T value) {
         return super.complete(value);
     }
 
-    public boolean kafkaCompleteExceptionally(Throwable throwable) {
+    boolean kafkaCompleteExceptionally(Throwable throwable) {
         return super.completeExceptionally(throwable);
     }
 
--- a/clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
+++ b/clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
@@ -16,9 +16,9 @@
  */
 package org.apache.kafka.common.internals;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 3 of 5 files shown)


tombentley:
@cmccabe @ijuma could one of you take a look at this? Or @dajac perhaps?

Thus far I've stuck strictly to reimplementing without breaking the existing `KafkaFuture` API. The logical next step would be to provide an accessor to the `CompletableFuture` as mentioned [previously](https://github.com/apache/kafka/pull/5131#issuecomment-395136834).

tombentley:
Rebased for conflict. @chia7712 maybe you could take a look if others are too busy?

chia7712:
I like this idea. We make some efforts to convert ```KafkaFuture``` to java ```CompletableFuture``` in our projects. It would be better to support ```CompletableFuture``` officially. 

However, this change can make a great impact on all producer users (bugs/performance/compatibility) so I feel it needs a KIP. (There is an KIP https://cwiki.apache.org/confluence/display/KAFKA/KIP-XXX%3A+Return+CompletableFuture+from+KafkaProducer.send)

tombentley:
@chia7712 to be clear, are you saying this reimplementation needs a KIP? Imho this PR as it stands should be OK without one, but of course any change to the API would need one. I was hoping to get this merged first. I'm happy to open a KIP for the API change though. I think everyone agrees that better interop between `KafkaFuture` and `CompletableFuture` would be a good thing.

chia7712:
oh, I misunderstood the purpose of this PR. The idea I mentioned is to migrate the return type from ```Future``` to ```CompletableFuture```.

What the benefit of reimplementing the ```KafkaFutureImpl``` by ```CompletableFuture``` if we eventually change the return type to ```CompletableFuture```?

dajac:
@tombentley Thanks for the PR. I wonder if `KafkaFuture` could implement `CompletionStage` and thus could be directly used with Java Future. Did we consider this?

chia7712:
```KafkaFuture``` is marked as ```Evolving``` so maybe we can open KIP to make it compatible to ```CompletionStage```?

**Comment by tombentley** [comment]
@dajac there are at least two reasons why that's problematic:

1. As described by @andrasbeni in his [previous PR](https://github.com/apache/kafka/pull/5131#issuecomment-397024119)
    > The problem with implementing `CompletionStage` is `whenComplete`. Right now user can write
`new KafkaFutureImpl<String>().thenApply(String::length)`. The lambda expression will be compiled to a `KafkaFuture.BiConsumer`. If we implement `CompletionStage`, there will be another `whenComplete` in the same class that takes `java.util.function.BiConsumer`, another SAM interface, as its parameter. That's why compiler will report ambiguity and will fail to compile above code.

2. The exception handling for some methods are just not compatible. For example, the contract of `CompletionStage` says the return value of `CompletionStage#thenApply` (and `whenComplete`) should fail with `CompletionException`:

    >  In all other cases, if a stage's computation terminates abruptly
    >  with an (unchecked) exception or error, then all dependent stages
    >  requiring its completion complete exceptionally as well, with a
    >  {@link CompletionException} holding the exception as its cause.

    But `KafkaFuture`'s `thenApply()` returns a `KafkaFuture` which will fail with `ExecutionException` (because `CompletionException` wasn't in Java 8 which Kafka supported at the time).

(FWIW there's also a mismatch with `getNow` as it works in `CompleteableFuture` and `KafkaFutures` though that's not a method declared on `CompletionStage`).

So trying to make `KafkaFuture` implement `CompleteableFuture` risks breaking users error handling code, possibly in quite subtle ways. Changes to make APIs return `CompletionStage` or `CompletableFuture` rather than `KafkaFuture` are neither source nor binary compatible and users would could have to change their error handling. So I think the only completely _safe_ option is to expose a `toCompletionStage()` method which returns the underlying `CompletableFuture`.

**Comment by chia7712** [comment]
> So I think the only completely safe option is to expose a toCompletionStage() method which returns the underlying CompletableFuture.

This is a good reason of reimplementing the ```KafkaFutureImpl``` by ```CompletableFuture``` :)

**Comment by tombentley** [comment]
@dajac, @chia7712 I realise now that I was wrong about 2. It's not a problem because `KafkaFuture.whenComplete` and `thenApply` are using `KafkaFuture.BiConsumer` and `KafkaFuture.Function`, not the ones from `java.util.function`, and so they're distinct from the ones inherited from `CompletionStage`. It would mean subtly different behaviour between the overloaded methods, which isn't ideal, but it is possible. 

That leaves 1, which is a source incompatibility, but not a binary incompatibility. Perhaps, with the `@InterfaceStability.Evolving` that's enough to justify implementing `CompletionStage`, wdyt?

**Comment by ijuma** [comment]
`@InterfaceStability.Evolving` has been there for so many releases when it comes to `Admin/AdminClient` that I don't think we can rely on it.

With regards to the method overloads, that would mean that lambdas won't work without casting, right?

**Comment by tombentley** [comment]
Casting would be required, yes.

**Comment by ijuma** [comment]
I think it would be worth agreeing on the end state. Do we want to have `KafkaFuture` long-term? If not, how do we go about removing its usage? A somewhat extreme approach would be to introduce a new interface `Admin2` (but with a better name) and use `CompletionStage` for all the methods. And eventually deprecate `Admin` and finally remove it.

**Comment by dajac** [comment]
@ijuma I was actually thinking about the same. The new or the old could likely be a thin facade of the other. Another approach would be to duplicate all the methods in the *Result classes but that does not sound good to me.

I think that it would be nice if we could eventually remove `KafkaFuture` and use `CompletionStage` instead. That would be nice from a user of the admin api perspective at least. That would also be inline with the producer api KIP-706 lands in the future.

**Comment by tombentley** [comment]
Note that `CompletionStage` on its own does not define `get`, `getNow`, which are vital methods for people wanting to use the client in a sync fashion, so I think the comparison is between `Admin2` API which used `CompletableFuture` and `Admin` using a `KafkaFuture` that implemented `CompletionStage`. I suppose you could use `toCompletableFuture`, but then it's less of a drop-in replacement.

A `KafkaFuture` that implemented `CompletionStage` adds very little additional value over `CompletableFuture` in the long term, _except_ the value of backwards compatibility. Anything like `Admin2`:

* would require quite a lot of work by the project (e.g. doubled test surface), albeit not a burden to be carried forever, and maybe the thin facade idea means it wouldn't really be so bad
* incurs a cost to users of discovering and switching to the new API, which offers nothing they couldn't do with a `KafkaFuture` that implemented `CompletionStage`.

My gut feeling is that if the only reason for an `Admin2` was to get rid of `KafkaFuture` then there doesn't seem to be enough of a benefit to justify it. If there were other things which could be cleaned up at the same time then perhaps that would shift the balance.

**Comment by ijuma** [comment]
It's a good thing that `CompletionStage` doesn't expose those methods directly, in my opinion. As you said, you can call `toCompletedFuture` to get them if you really want them.

**Comment by ijuma** [comment]
> My gut feeling is that if the only reason for an Admin2 was to get rid of KafkaFuture then there doesn't seem to be enough of a benefit to justify it. If there were other things which could be cleaned up at the same time then perhaps that would shift the balance.

I think the challenge is that there is no good way to retrofit `KafkaFuture`. I think the only viable approach I've seen so far is to add the `toCompletableFuture`/`toCompletionStage` methods.

**Comment by tombentley** [comment]
> It's a good thing that CompletionStage doesn't expose those methods directly, in my opinion. As you said, you can call toCompletedFuture to get them if you really want them.

It makes perfect sense for the use cases targeted by that API, where random blocking in an async pipeline is to be avoided. But many people who use the Admin client want to do so in a synchronous style. They already have to invoke `KafkaFuture.get` to do that, having to do `toCompletableFuture().get()` makes the API less discoverable and just slightly more painful to those who want sync.  They likely wouldn't see a switch to plain CompletionStage as an improvement. 

> I think the challenge is that there is no good way to retrofit KafkaFuture. I think the only viable approach I've seen so far is to add the toCompletableFuture/toCompletionStage methods.

It seems to be the only completely backwards compatible thing we can do.

**Comment by ijuma** [comment]
Another relevant item for consideration is that project Loom indicates that the Java libraries may try to promote async without async APIs.

**Comment by tombentley** [comment]
@ijuma I'm not quite sure what you're proposing. That some `Admin2` might not even need to use any kind of Future/CompletionStage API? I don't think that stacks up as an argument for not trying to do something to do something about/improve KafkaFuture today:

* Loom may or may not ultimately be successful in persuading people to do sync programming in virtual threads. Until that's been proven, having an Admin client that supports async is valuable.
* when will virtual threads actually make it into the JDK? My best guess is it won't be production ready for around 18 months in the best case (maybe you have better info than me though?) Even if it was here today people would still be using older JREs for years to come, and some of those people would want to use an Admin client async and use CompletionStage.

Or do I misunderstand the point you're trying to make?

**Comment by ijuma** [comment]
@tombentley I didn't suggest anything. :) Maybe Loom means we should avoid Admin2 and go with an incremental path first (maybe something like what you're suggesting here).

**Comment by ijuma** [comment]
Also worth thinking about what we would do for new admin methods.

**Comment by tombentley** [comment]
Incremental first would be my preference. It doesn't preclude doing something else in the future.

If we improved `KafkaFuture` then I would argue that new admin methods should still use it, for consistency. Allowing a mix of KafkaFuture and CompletionStage just makes the API inconsistent, makes Admin2 feel more necessary than it needs to be, and doesn't add any additional functionality.

**Comment by tombentley** [comment]
CI faliures are unrelated. @ijuma and/or @chia7712 could one of you review this please?

**Comment by tombentley** [comment]
@dajac @ijuma please could you review this, it's the implementation for KIP-707.

**Comment by tombentley** [comment]
Test failure is unrelated. @dajac @cmccabe @mimaison please could one of you review?

File: clients/src/test/java/org/apache/kafka/common/KafkaFutureTest.java [hunk 1]
```diff
@@ -17,68 +17,261 @@
...
+    @Test
+    public void testCompleteFuturesExceptionally() throws Exception {
         KafkaFutureImpl<Integer> futureFail = new KafkaFutureImpl<>();
-        futureFail.completeExceptionally(new RuntimeException("We require more vespene gas"));
-        ExecutionException e = assertThrows(ExecutionException.class, futureFail::get);
-        assertEquals(RuntimeException.class, e.getCause().getClass());
-        assertEquals("We require more vespene gas", e.getCause().getMessage());
+        assertTrue(futureFail.completeExceptionally(new RuntimeException("We require more vespene gas")));
+        assertIsFailed(futureFail);
+        assertFalse(futureFail.completeExceptionally(new RuntimeException("We require more minerals")));
```
mimaison, line 146: Good to see we're staying on theme!

tombentley:
@dajac @chia7712 (and let me try @rajinisivaram on the off-chance), I'd like to get this into Kafka 3.0, but this hasn't seen any review yet despite repeated requests. Any chance we could squeeze it in?

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -27,217 +28,107 @@
...
-            exception = newException;
-            done = true;
-            oldWaiters = waiters;
-            waiters = null;
-        }
-        for (BiConsumer<? super T, ? super Throwable> waiter : oldWaiters) {
-            waiter.accept(null, newException);
-        }
-        return true;
+        // CF#get() always wraps the _cause_ of a CompletionException in EE (which KF does not)
```
kkonstantine: Not sure what EE is here.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -27,217 +28,107 @@
...
+            }
+        });
+        return new KafkaFutureImpl<>(true, toKafkaCompletableFuture(completableFuture));
     }
 
-    public <R> void copyWith(KafkaFuture<R> future, BaseFunction<R, T> function) {
-        KafkaFutureImpl<R> futureImpl = (KafkaFutureImpl<R>) future;
-        futureImpl.addWaiter(new Applicant<>(function, this));
+    private <U> KafkaCompletableFuture<U> toKafkaCompletableFuture(CompletableFuture<U> tCompletableFuture) {
+        KafkaCompletableFuture<U> res;
```
kkonstantine: the fact that two variables are used here must be a remainder of an earlier iteration. 
```suggestion
        KafkaCompletableFuture<U> result;
```
can be declared here at the top level and we can remove the inner declaration.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -27,217 +28,107 @@
...
     /**
      * Returns a new KafkaFuture that, when this future completes normally, is executed with this
      * futures's result as the argument to the supplied function.
      */
     @Override
     public <R> KafkaFuture<R> thenApply(BaseFunction<T, R> function) {
-        KafkaFutureImpl<R> future = new KafkaFutureImpl<>();
-        addWaiter(new Applicant<>(function, future));
-        return future;
+        CompletableFuture<R> completableFuture = this.completableFuture.thenApply((java.util.function.Function<? super T, ? extends R>) value -> {
```
kkonstantine: Not a big fan of variable shadowing, which below you avoid by calling the local variable `tCompletableFuture`. Probably good idea to apply the naming here to and remove `this.`

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -267,50 +173,77 @@ public T get() throws InterruptedException, ExecutionException {
...
     }
 
     /**
      * Returns true if this CompletableFuture was cancelled before it completed normally.
      */
     @Override
     public synchronized boolean isCancelled() {
-        return exception instanceof CancellationException;
+        if (isDependant) {
+            Throwable exception;
```
kkonstantine: I believe just returning from the catch branch is fine. No need for this variable. (you return from any path)

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -267,50 +173,77 @@ public T get() throws InterruptedException, ExecutionException {
...
     public synchronized boolean isDone() {
-        return done;
+        return completableFuture.isDone();
     }
 
     @Override
     public String toString() {
-        return String.format("KafkaFuture{value=%s,exception=%s,done=%b}", value, exception, done);
+        T value = null;
+        Throwable ex = null;
```
kkonstantine: the naming used above seems better here
```suggestion
        Throwable exception = null;
```

kkonstantine (commented):
Took a pass and left a few minor comments. 
Looks good to me overall and I could approve if we address asap. Not sure if others have any reservations w/r/t this implementation. If not, it'd be nice to include in 3.0 given that KIP-707 has been adopted. The test coverage seems sufficient.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: Expanded all the class abbreviations.

clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
```diff
@@ -98,14 +80,45 @@ private void maybeComplete() {
...
-        AllOfAdapter<Object> allOfWaiter = new AllOfAdapter<>(futures.length, allOfFuture);
-        for (KafkaFuture<?> future : futures) {
-            future.addWaiter(allOfWaiter);
-        }
-        return allOfFuture;
+        KafkaFutureImpl<Void> result = new KafkaFutureImpl<>();
+        CompletableFuture.allOf(Arrays.stream(futures)
+                .map(kafkaFuture -> {
+                    // Safe since KafkaFuture's only subclass is KafkaFuture for which toCompletionStage()
+                    // always return a CF.
```
dajac, line 85: nit:  I would replace `CF` by `CompletableFuture`.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: Not a fan of the name `tCompletableFuture` either, so I used `appliedFuture` here and made the following method that used `tCompletableFuture` static to avoid shadowing there.

dajac: nit: Do we really need to specify `(java.util.function.Function<? super T, ? extends R>)` here?

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -27,217 +28,107 @@
...
     public <R> KafkaFuture<R> thenApply(BaseFunction<T, R> function) {
-        KafkaFutureImpl<R> future = new KafkaFutureImpl<>();
-        addWaiter(new Applicant<>(function, future));
-        return future;
+        CompletableFuture<R> completableFuture = this.completableFuture.thenApply((java.util.function.Function<? super T, ? extends R>) value -> {
+            try {
+                return function.apply(value);
+            } catch (Throwable t) {
+                if (t instanceof CompletionException) {
+                    throw new CompletionException(t);
```
dajac, line 68: I am not sure that I follow why we wrap it here. Could you elaborate?

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -27,217 +28,107 @@
...
+        return new KafkaFutureImpl<>(true, toKafkaCompletableFuture(completableFuture));
     }
 
-    public <R> void copyWith(KafkaFuture<R> future, BaseFunction<R, T> function) {
-        KafkaFutureImpl<R> futureImpl = (KafkaFutureImpl<R>) future;
-        futureImpl.addWaiter(new Applicant<>(function, this));
+    private <U> KafkaCompletableFuture<U> toKafkaCompletableFuture(CompletableFuture<U> tCompletableFuture) {
+        KafkaCompletableFuture<U> res;
+        if (tCompletableFuture instanceof KafkaCompletableFuture) {
+            res = (KafkaCompletableFuture<U>) tCompletableFuture;
```
dajac: nit: We could omit `res` and return directly in the two places below.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
dajac: nit: I would rather use the full name instead of using acronyms.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -247,17 +138,32 @@ public boolean completeExceptionally(Throwable newException) {
...
-        return completeExceptionally(new CancellationException()) || exception instanceof CancellationException;
+        return completableFuture.cancel(mayInterruptIfRunning);
+    }
+
+    // We need to deal with differences between KF's historic API and the API of CF:
+    // CF#get() does not wrap CancellationException in ExecutionException (nor does KF).
+    // CF#get() always wraps the _cause_ of a CompletionException in ExecutionException (which KF does not).
+    //
+    // The semantics for KafkaFuture are that all exceptional completions of the future (via #completeExceptionally() or exceptions from dependants)
+    // manifest as ExecutionException, as observed via both get() and getNow().
```
dajac: nit: Could we turn this into a javadoc or perhaps move it into the method?

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -247,17 +138,32 @@ public boolean completeExceptionally(Throwable newException) {
...
     @Override
     public T get() throws InterruptedException, ExecutionException {
-        SingleWaiter<T> waiter = new SingleWaiter<>();
-        addWaiter(waiter);
-        return waiter.await();
+        try {
+            return completableFuture.get();
+        } catch (ExecutionException e) {
+            maybeRewrapAndThrow(e.getCause());
+            throw e;
```
dajac: nit: Would it make sense to move `throw e` into `maybeRewrapAndThrow` to let `maybeRewrapAndThrow` throw in both cases? More generally, I wonder if we could handle all the case in `maybeRewrapAndThrow` and use it everywhere.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -267,50 +173,77 @@ public T get() throws InterruptedException, ExecutionException {
...
+        if (isDependant) {
+            Throwable exception;
+            try {
+                completableFuture.getNow(null);
+                return false;
+            } catch (Exception e) {
+                exception = e;
+            }
+            return exception instanceof CompletionException
+                    && exception.getCause() instanceof CancellationException;
```
dajac: I am curious here. Would it be possible to let the `CancellationException` propagate from the upstream future to this one when the upstream future is cancelled?

dajac (commented):
@tombentley I am really sorry for the delay on this one. I finally had the time to look at it. Overall, it looks pretty good to me. I left few comments and questions.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: The lambda passed to `whenComplete` needs to complete the `KafkaCompletableFuture`, so needs a captured variable, but I'll declare `result` in the `else` branch only.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: I agree it's a bit messy, but I don't think that works. The call site in `getNow()` differs from the others:

1. It's catching `CompletionException`, not `ExecutionException`
2. In the case there `maybeRewrapAndThrow` didn't throw it needs to wrap the cause in an `ExceptionException`, unlike the other call sites, to conform with the historical API of KafkaFuture. 

The first would require `maybeRewrapAndThrow()` to have an `Exception`-typed parameter, which, being a checked exception, means we'd need either to declare `throws Exception`, or we'd need to type cast within `maybeRewrapAndThrow()` and be left with an `else` case to deal with. 

The second is difficult to abstract over nicely for the other call sites: We'd need to pass a boolean to indicate whether the `ExceptionException` was needed.

dajac: How about defining two helper methods, one for each cases?
* `private void maybeRewrapAndThrow(ExecutionException exception)`; and
* `private void maybeRewrapAndThrow(CompletionException exception)`

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: The historical API of `KafkaFuture` on cancellation of an upstream future (e.g. `thenApply()`) was for both futures to return true from `isDone()`, `isCancelled()` and `isCompletedExceptionally()`, and for both `getNow` and `get()` to throw a `CancellationException`. 

If the new implementation of `isCancelled()` just does `return completableFuture.isCancelled();` then for a dependent future `isCancelled()` would return false because it's exception is a `CompletionException` wrapping a `CancellationException` and `CompleteableFuture#isCancelled()` just checks for the exception being `CancellationException`. So we need to do something different for cancellation of dependent futures. 

Because `KafkaFutureImpl.thenApply()` doesn't itself keep track of waiters any more it's not possible to do anything clever in the implementation of `cancel()`, so we have to do something in `isCancelled()`.

So to (hopefully) answer your question: It does propagate already, but the APIs differ and what we're doing here is to maintain the existing semantics of `KafkaFuture`.

dajac: Oh.. I see. Thanks for the clarification. It would be great if you could add a comment which explains this because this is not obvious.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: Again it comes down to maintaining the historical behaviour of `thenApply` when the supplied `Function` threw `CompletionException`.  Basically `CompletableFuture` does some unwrapping of `CompletionException` in dependent futures which is incompatible with what `KafkaFuture` did. We can't prevent the unwrapping, but since such wrapping only goes one level deep in the cause chain we can restore the `KafkaFuture` behaviour by inserting a bogus `CompletionException` for `CompletableFuture` to remove. 

I've added a comment and a test for this case.

tombentley:
@kkonstantine @dajac thanks for the reviews, I really appreciate it. I think I've addressed all the comments and hopefully answered your questions, if you'd like to take another look?

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: Added. Let me know whether it makes sense to you. Also: this is covered by the test which passes with the old and new implementations.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
dajac: @tombentley What do you think about this? Would it help to consolidate the logic?

tombentley: It doesn't seem to help with other places where we catch and rethrow because they catch unchecked exceptions `Error` and `RuntimeException` whose only supertype is `Throwable`, so we end up in a painful place due to Java's lack of abstraction for "unchecked exception".

For the existing call sites the `void` return type you proposed doesn't work with javac's definite return analysis so I have to declare `private <T> T rewrapAndThrow(CompletionException exception)` and `private <T> T rewrapAndThrow(ExecutionException exception)`. The method always throws, so the return type of `T` is just there to appease `javac` via a `return rewrapAndThrow(...)` at the call site. 

I can do this is you want, but I feel it's making the code less clear. The `rewrapAndThrow(CompletionException exception)` only has a single caller, so it's of little benefit.

I guess the other alternative would be to inline the method and just comment each site as to what's going on.

dajac: I don't like the fact that we throw in two places... Could we at least make the name of `maybeRewrapAndThrow` a bit more explicit? It is only about `CancellationException` in the end so we could name it `maybeThrowCancellationException` or something like this. Moreover, the method does not really "rewrap" anything, right? It just checks the type and throw it.

tombentley: There don't seem to be any good options here.

We could throw from a single method if you're willing to have the method like this

```java
    private <T, E extends Exception> T throwUnwrappingCancellationException(E exception) throws E {
        if (exception.getCause() instanceof CancellationException) {
            throw (CancellationException) exception.getCause();
        }
        throw exception;
    }
```

and the call sites to look like this:

```java
return throwUnwrappingCancellationException(e);
```

 The `return <methodThatAlwaysThrows>` is nasty, and I can't think of a good method name. Personally I don't think this is better than naming the current method `maybeThrowCancellationException` as you suggest, and living with the throw in two places, so I've gone with that.

kkonstantine (approved):
Thanks @tombentley 
LGTM
It'd be best to have @dajac's approval too.

clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+/**
+ * This internal class exists because CompletableFuture exposes complete(), completeExceptionally() and
+ * other methods which would allow erroneous completion by user code of a KafkaFuture returned from a
+ * Kafka API to a client application.
+ * @param <T> The type of the future value.
+ */
+public class KafkaCompletableFuture<T> extends CompletableFuture<T> {
```
chia7712, line 30: Why not throwing exception when users call `toCompletableFuture`? If users want to use blocking method (ex. `get()`), they ought to call `KafkaFuture#get` instead. What is the benefit if we expose `toCompletableFuture`?

clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
```diff
@@ -0,0 +1,79 @@
...
+
+    boolean kafkaComplete(T value) {
```
chia7712, line 38: why it needs prefix "kafka"?

clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
```diff
@@ -19,75 +19,57 @@
...
+ * <h3>Relation to {@code CompletionStage}</h3>
+ * <p>It is possible to obtain a {@code CompletionStage} from a
+ * {@code KafkaFuture} instance by calling {@link #toCompletionStage()}.
+ * If converting {@link KafkaFuture#whenComplete(BiConsumer)} or {@link KafkaFuture#thenApply(BaseFunction)} to
+ * {@link CompletableFuture#whenComplete(java.util.function.BiConsumer)} or
+ * {@link CompletableFuture#thenApply(java.util.function.Function)} be aware that the returned
+ * {@code KafkaFuture} will fail with an {@code ExecutionException}, whereas a {@code CompletionStage} fails
+ * with a {@code CompletionException}.
  */
 @InterfaceStability.Evolving
```
chia7712: Should we remove this annotation.

KIP-707 has following description.

```
Remove the @InterfaceStability.Evolving annotation on KafkaFuture to reflect the reality that changing this class incompatibly would cause of too much user code to break.
```

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -267,50 +180,82 @@ public T get() throws InterruptedException, ExecutionException {
...
     /**
      * Returns true if this CompletableFuture was cancelled before it completed normally.
      */
     @Override
     public synchronized boolean isCancelled() {
-        return exception instanceof CancellationException;
+        if (isDependant) {
+            // Having isCancelled() for a dependent future just return
+            // CompletableFuture.isCancelled() would break the historical KafkaFuture behaviour because
+            // CompleteableFuture#isCancelled() just checks for the exception being CancellationException
```
chia7712: typo: CompleteableFuture -> CompletableFuture

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -267,50 +180,82 @@ public T get() throws InterruptedException, ExecutionException {
...
+            // whereas it will be a CompletionException wrapping a CancellationException
+            // due needing to compensate for CompleteableFuture's CompletionException unwrapping
```
chia7712: typo: CompleteableFuture -> CompletableFuture

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -267,50 +180,82 @@ public T get() throws InterruptedException, ExecutionException {
...
      * the given valueIfAbsent.
      */
     @Override
-    public synchronized T getNow(T valueIfAbsent) throws InterruptedException, ExecutionException {
-        if (exception != null)
-            wrapAndThrow(exception);
-        if (done)
-            return value;
-        return valueIfAbsent;
+    public synchronized T getNow(T valueIfAbsent) throws ExecutionException {
```
chia7712: Does it need to remove `InterruptedException` from other get methods also? For another, why we don't remove `InterruptedException` from `KafkaFuture#getNow`?

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -247,17 +142,35 @@ public boolean completeExceptionally(Throwable newException) {
      */
     @Override
     public synchronized boolean cancel(boolean mayInterruptIfRunning) {
```
chia7712: Do we still need `synchronized`? It seems to me `CompletableFuture` can deal with thread issue.

clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
```diff
@@ -0,0 +1,79 @@
...
+        return super.complete(value);
+    }
+
+    boolean kafkaCompleteExceptionally(Throwable throwable) {
+        return super.completeExceptionally(throwable);
+    }
+
+    @Override
+    public boolean complete(T value) {
+        throw new UnsupportedOperationException();
```
chia7712: How about adding some comments to this exception?

chia7712 (commented):
@tombentley thanks for this good patch. This issue is hard but important. a couple of comments below.

clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
tombentley: The KIP describes it like this:

> Adding toCompletionStage() is sufficient because CompletionStage itself exposes toCompletableFuture(), so anyone who needs an actual CompletableFuture (e.g. for interoperating with 3rd party APIs that require one) can get one. However, CompletableFuture exposes methods for future completion which should not be called by users (only the Admin client should be completing the returned futures), so calling these will be prevented. It is expected that users wanting to block on the completion of the KafkaFuture would use kafkaFuture.get(), rather than calling kafkaFuture.toCompletionStage().toCompletableFuture().get(), so the need to access the CompleteableFuture should be rare.

Essentially we want to allow access to a `CompletableFuture` to allow interop with 3rd part APIs which might require it (e.g. for using the blocking `get` methods) , but not allow user code to complete those futures. The `CompletableFuture` API unhelpfully adds _both_ the blocking `get` methods and the `complete` methods.

clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
tombentley: The idea is that all Kafka code knows it's using a `KafkaCompletableFuture` and knows to complete it via the `kafka*()` methods. User code which erroneously tries to complete via `complete()` gets the exception thrown.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: > Does it need to remove InterruptedException from other get methods also?

`get()` and `get(long, TimeUnit)` can still throw `InterruptedException` due to the blocking behaviour (the `completableFuture.get` that we delegate to throws `InterruptedException`).

> For another, why we don't remove InterruptedException from KafkaFuture#getNow?

Good spot!  `KafkaFuture#getNow` should never have declared that and it's annoying for callers to have to deal with it. But this wasn't described in the KIP and wouldn't be a source compatible change (existing code with a `catch (InterruptedException)` would get a compile error). 

I'd be happy to remove it and notify the VOTE thread, WDYT @kkonstantine @dajac ?

clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
chia7712: > allow interop with 3rd part APIs which might require it

fair enough :)

clients/src/main/java/org/apache/kafka/common/internals/KafkaCompletableFuture.java
chia7712: > The idea is that all Kafka code knows it's using a KafkaCompletableFuture and knows to complete it via the kafka*() methods.

that is a good point. Could you add this comment?

tombentley: Done

tombentley:
@chia7712 thanks for the review!

tombentley:
@dajac @chia7712 I think I've addressed all your comments.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -27,217 +28,111 @@
...
-            action.accept(null, exception);
-        } else if (done) {
-            action.accept(value, null);
-        } else {
-            waiters.add(action);
-        }
-    }
 
     @Override
     public synchronized boolean complete(T newValue) {
```
dajac: Could we also remove `synchronized` here?

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -267,50 +180,82 @@ public T get() throws InterruptedException, ExecutionException {
...
+            throw e;
+        }
     }
 
     /**
      * Returns the result value (or throws any encountered exception) if completed, else returns
      * the given valueIfAbsent.
      */
     @Override
-    public synchronized T getNow(T valueIfAbsent) throws InterruptedException, ExecutionException {
```
dajac, line 280: Could we also remove `synchronized` here?

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
dajac: Good question. AK 3.0 is a good opportunity to do such breaking changes so I would be in favour of doing it. Let's see what other think.

dajac (approved):
LGTM, thanks for the patch. It would be great to have +1 from @chia7712 as well.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
chia7712: > But this wasn't described in the KIP and wouldn't be a source compatible change (existing code with a catch (InterruptedException)

As it can cause compatible change, we should deprecate it and then add a new method (for example: get(T value)) to replace it. This can be discussed in another issue :)

clients/src/test/java/org/apache/kafka/common/KafkaFutureTest.java
```diff
@@ -17,68 +17,265 @@
...
+            future.get();
+            fail("Expected an exception");
+        } catch (CancellationException e) {
+            assertEquals(CancellationException.class, e.getClass());
+            assertEquals(expectedMessage, e.getMessage());
+        } catch (Exception e) {
+            throw new AssertionError("Unexpected exception", e);
+        }
+        try {
+            future.getNow(null);
```
chia7712: How about using following style.

```java
        ExecutionException e = assertThrows(ExecutionException.class, () -> future.getNow(null));
        assertEquals(expectedException, e.getCause().getClass());
        assertEquals(expectedMessage, e.getCause().getMessage());
        return e.getCause();
```

clients/src/main/java/org/apache/kafka/common/KafkaFuture.java
```diff
@@ -184,7 +194,7 @@ public abstract T get(long timeout, TimeUnit unit) throws InterruptedException,
      * Returns the result value (or throws any encountered exception) if completed, else returns
      * the given valueIfAbsent.
      */
-    public abstract T getNow(T valueIfAbsent) throws InterruptedException, ExecutionException;
+    public abstract T getNow(T valueIfAbsent) throws ExecutionException;
```
chia7712: Could you please revert this change as you mentioned that it can cause compatible issue.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
```diff
@@ -27,217 +28,111 @@
...
-                        return value;
-                    if (delta >= waitTimeMs) {
-                        throw new TimeoutException();
-                    }
-                    this.wait(waitTimeMs - delta);
-                    delta = System.currentTimeMillis() - startMs;
-                }
-            }
-        }
+    public KafkaFutureImpl(KafkaCompletableFuture<T> completableFuture) {
```
chia7712: Is this method still useful?

chia7712 (commented):
@tombentley thanks for updating code. a couple of comments below. PTAL

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: I wouldn't call it `get` because it has different semantics to the existing get methods, so I think it should have a distinct name. And I can't think of a succinct alternative to `getNow`. Something like `getOrDefault` doesn't convey they key difference from `get` (the lack of blocking). `getImmediately` is just a bit verbose. 

The API of KafkaFuture has always been a subset of `CompletableFuture`. Introducing a new method would break that. While it would be discoverable while the deprecated method remained to point the way it wouldn't be as discoverable once the old `getNow` was removed.

All deprecation really does is let the programmer defer fixing the problem. But it's most likely not a huge problem to fix (i.e. they've likely caught the exception close to the method call, rather than defining many layers of API with `throws InterruptedException`).

A new method also means their code is then not binary compatible with older client versions. 

So on balance I personally favour just removing the `throws InterruptedException` in 3.0 and letting the programmer clean up the code sooner rather than later. 

There is always the "do nothing" alternative.

ijuma: "do nothing" is probably the right thing here.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: No, but it makes me wonder whether we'd find factory methods for already completed and already failed `KafkaFutures` useful.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
tombentley: OK, I've reverted it.

tombentley:
@chia7712 thanks, I've implemented your most recent comments.

clients/src/main/java/org/apache/kafka/common/internals/KafkaFutureImpl.java
chia7712: > A new method also means their code is then not binary compatible with older client versions.

yep, the BC could be broken. What I really care is the source compatibility. It seems to me a public API should be source compatible to next major release :)

This patch is good to go and we need more time (rather than vespene gas ... StarCraft is my favor game :) ) to reach consensus.

tombentley:
If no further comments I'll merge this in the next hour or so.

chia7712 (approved):
LGTM

tombentley:
Thanks for the reviews @chia7712 @dajac @kkonstantine

PR #9880, opened 2021-01-13
KAFKA-10792 (2.5 backport): Prevent source task shutdown from blocking herder thread (#9669)

C0urante:
Backports the changes from #9669.

Changes the `WorkerSourceTask` class to only call `SourceTask::stop` from the task thread when the task is actually stopped (via `Source:task::close` just before `WorkerTask::run` completes), and only if an attempt has been made to start the task (which will not be the case if it was created in the paused state and then shut down before being started). This prevents `SourceTask::stop` from being indirectly invoked on the herder's thread, which can have adverse effects if the task is unable to shut down promptly.

Unit tests are tweaked where necessary to account for this new logic, which covers some edge cases mentioned in PR #5020 that were unaddressed up until now.

The existing integration tests for blocking connectors added later in 2.6 are backported here to provide cases for blocking source and sink tasks. Full coverage of every source/sink task method is intentionally omitted from these expanded tests in order to avoid inflating test runtime (each one adds an extra 5 seconds at minimum) and because the tests that are added here were sufficient to reproduce the bug with source task shutdown.

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
@@ -96,9 +96,7 @@ class WorkerSourceTask extends WorkerTask {
     private CountDownLatch stopRequestedLatch;
 
     private Map<String, String> taskConfig;
-    private boolean finishedStart = false;
-    private boolean startedShutdownBeforeStartCompleted = false;
-    private boolean stopped = false;
+    private boolean started = false;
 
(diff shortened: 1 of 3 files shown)


rhauch (approved):
LGTM. Thanks for the backport, @C0urante.

rhauch:
I'm trying to get the builds to run properly, but I did pull locally and the build of that local branch does pass.

rhauch:
There seem to be environmental build issues; the builds are failing with "JDK" unknown and "maven" unknown.

I'm going to go ahead and merge this PR to the `2.5` branch.

C0urante:
Thanks Randall!

PR #9881, opened 2021-01-13
MINOR: Initialize QuorumState lazily in RaftClient.initialize()

aloknnikhil:
- There are cases where the quorum voters connect string is not ready
  up until after the controller starts.
  Eg: If the controller is configured to choose a free port, we need to
      wait until after the controller starts to configure the right
      host:port connect string for the RaftClient to connect to

### DIFF
--- a/core/src/main/scala/kafka/raft/RaftManager.scala
+++ b/core/src/main/scala/kafka/raft/RaftManager.scala
@@ -16,11 +16,6 @@
  */
 package kafka.raft
 
-import java.io.File
-import java.nio.file.Files
-import java.util.Random
-import java.util.concurrent.CompletableFuture
-
 import kafka.log.{Log, LogConfig, LogManager}
 import kafka.raft.KafkaRaftManager.RaftIoThread
 import kafka.server.{BrokerTopicStats, KafkaConfig, KafkaServer, LogDirFailureChannel}
@@ -34,8 +29,11 @@ import org.apache.kafka.common.protocol.ApiMessage
 import org.apache.kafka.common.requests.RequestHeader
 import org.apache.kafka.common.security.JaasContext
 import org.apache.kafka.common.utils.{LogContext, Time}
-import org.apache.kafka.raft.{FileBasedStateStore, KafkaRaftClient, QuorumState, RaftClient, RaftConfig, RaftRequest, RecordSerde}
+import org.apache.kafka.raft.{FileBasedStateStore, KafkaRaftClient, RaftClient, RaftConfig, RaftRequest, RecordSerde}
 
+import java.io.File
+import java.nio.file.Files
+import java.util.concurrent.CompletableFuture
 import scala.jdk.CollectionConverters._
 
 object KafkaRaftManager {
@@ -119,7 +117,7 @@ class KafkaRaftManager[T](
 
   def startup(): Unit = {
     netChannel.start()
-    raftClient.initialize()
+    raftClient.initialize(config.getString(RaftConfig.QUORUM_VOTERS_CONFIG))
     raftIoThread.start()
   }
 
@@ -168,16 +166,6 @@ class KafkaRaftManager[T](
   }
 
   private def buildRaftClient(): KafkaRaftClient[T] = {
-    val quorumState = new QuorumState(
-      nodeId,
-      raftConfig.quorumVoterIds,
-      raftConfig.electionTimeoutMs,
-      raftConfig.fetchTimeoutMs,
-      new FileBasedStateStore(new File(dataDir, "quorum-state")),
-      time,
-      logContext,
-      new Random()
-    )
 
     val expirationTimer = new SystemTimer("raft-expiration-executor")
     val expirationService = new TimingWheelExpirationService(expirationTimer)
@@ -187,11 +175,12 @@ class KafkaRaftManager[T](
       recordSerde,
       netChannel,
       metadataLog,
-      quorumState,
+      new FileBasedStateStore(new File(dataDir, "quorum-state")),
       time,
       metrics,
       expirationService,
-      logContext
+      logContext,
+      nodeId
     )
   }
 
--- a/core/src/main/scala/kafka/raft/RaftManager.scala
+++ b/core/src/main/scala/kafka/raft/RaftManager.scala
@@ -117,7 +117,7 @@ class KafkaRaftManager[T](
 
   def startup(): Unit = {
     netChannel.start()
-    raftClient.initialize(config.getString(RaftConfig.QUORUM_VOTERS_CONFIG))
+    raftClient.initialize(raftConfig)
     raftIoThread.start()
   }
 
@@ -171,7 +171,6 @@ class KafkaRaftManager[T](
     val expirationService = new TimingWheelExpirationService(expirationTimer)
 
     new KafkaRaftClient(
-      raftConfig,
       recordSerde,
       netChannel,
       metadataLog,
--- a/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
+++ b/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
@@ -96,7 +96,6 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import static java.util.concurrent.CompletableFuture.completedFuture;
-import static org.apache.kafka.raft.RaftConfig.quorumVoterStringsToNodes;
 import static org.apache.kafka.raft.RaftUtil.hasValidTopicPartition;
 
 /**
@@ -142,13 +141,7 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
     private final Logger logger;
     private final LogContext logContext;
     private final Time time;
-    private final int electionBackoffMaxMs;
-    private final int electionTimeoutMs;
     private final int fetchMaxWaitMs;
-    private final int fetchTimeoutMs;
-    private final int appendLingerMs;
-    private final int retryBackoffMs;
-    private final int requestTimeoutMs;
     private final int nodeId;
     private final NetworkChannel channel;
     private final ReplicatedLog log;
@@ -168,9 +161,9 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
     private RequestManager requestManager;
     private QuorumState quorum;
     private KafkaRaftMetrics kafkaRaftMetrics;
+    private RaftConfig raftConfig;
 
     public KafkaRaftClient(
-        RaftConfig raftConfig,
         RecordSerde<T> serde,
         NetworkChannel channel,
         ReplicatedLog log,
@@ -190,13 +183,7 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
             time,
             metrics,
             expirationService,
-            raftConfig.electionTimeoutMs(),
-            raftConfig.fetchTimeoutMs(),
-            raftConfig.electionBackoffMaxMs(),
-            raftConfig.retryBackoffMs(),
-            raftConfig.requestTimeoutMs(),
             FETCH_MAX_WAIT_MS,
-            raftConfig.appendLingerMs(),
             nodeId,
             logContext,
             new Random());
@@ -212,13 +199,7 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
         Time time,
         Metrics metrics,
         ExpirationService expirationService,
-        int electionTimeoutMs,
-        int fetchTimeoutMs,
-        int electionBackoffMaxMs,
-        int retryBackoffMs,
-        int requestTimeoutMs,
         int fetchMaxWaitMs,
-        int appendLingerMs,
         int nodeId,
         LogContext logContext,
         Random random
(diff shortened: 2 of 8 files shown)


raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -325,8 +341,30 @@ private void fireHandleResign() {
     }
 
     @Override
-    public void initialize() throws IOException {
-        quorum.initialize(new OffsetAndEpoch(log.endOffset().offset, log.lastFetchedEpoch()));
+    public void initialize(String quorumVoterStrings) throws IOException {
```
hachikuji: This is a little awkward given the presence of the voter configuration in `RaftConfig`. Perhaps we could let `initialize` take the full config?
```java
public void initialize(RaftConfig config) throws IOException
```

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -325,8 +341,30 @@ private void fireHandleResign() {
...
+                quorumVoterIds,
+                electionTimeoutMs,
+                fetchTimeoutMs,
+                quorumStateStore,
+                time,
+                logContext,
+                random);
+        quorumState.initialize(new OffsetAndEpoch(log.endOffset().offset, log.lastFetchedEpoch()));
+        this.quorum = quorumState;
+        this.kafkaRaftMetrics = new KafkaRaftMetrics(metrics, "raft", quorum);
```
hachikuji, line 340: Since we are no longer initializing this in the constructor, we might need a null check in `close`.

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
```diff
@@ -753,11 +751,13 @@ void start(int nodeId) {
                 FETCH_TIMEOUT_MS, persistentState.store, time, logContext, random);
             Metrics metrics = new Metrics(time);
 
-            Map<Integer, InetSocketAddress> voterConnectionMap = voters.stream()
-                .collect(Collectors.toMap(
-                    Function.identity(),
-                    this::nodeAddress
-                ));
+            StringBuilder votersString = new StringBuilder();
```
hachikuji: nit: since we have the same code in a couple places, maybe we can factor it into a `RaftTestUtil`?

hachikuji (commented):
Thanks for the patch. Left a few comments.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: How about storing the `RaftConfig` as a field of `KafkaRaftClient`? This would allow us to remove a few of the configuration parameter passed through the constructor.

@hachikuji is there a reason why `KafkaRaftClient` has one constructor which accepts a `RaftConfig` and one constructor that accepts the configurations as explicit parameters?

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -325,8 +341,30 @@ private void fireHandleResign() {
     }
 
     @Override
-    public void initialize() throws IOException {
-        quorum.initialize(new OffsetAndEpoch(log.endOffset().offset, log.lastFetchedEpoch()));
+    public void initialize(String quorumVoterStrings) throws IOException {
+        List<Node> quorumVoterNodes = quorumVoterStringsToNodes(quorumVoterStrings);
+        Set<Integer> quorumVoterIds = quorumVoterNodes.stream().map(Node::id).collect(Collectors.toSet());
```
jsancio: This is the same as `voterAddresses.keySet()` which is computed below.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
```diff
@@ -62,11 +62,12 @@ default void handleResign() {}
     }
 
     /**
-     * Initialize the client. This should only be called once on startup.
+     * Initialize the client with the given voter string config.
+     * This should only be called once on startup.
```
jsancio, line 66: Add a `@param` to the documentation.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -149,6 +152,15 @@ public static ConfigDef configDef() {
         return new ConfigDef(CONFIG);
     }
 
+    public static List<Node> quorumVoterStringsToNodes(String quorumVotersString) {
+        return parseVoterConnections(Arrays.stream(quorumVotersString.split(","))
+                .filter(part -> !part.isEmpty())
+                .collect(Collectors.toList())).entrySet().stream()
```
jsancio: I think we can avoid this pattern if we refactor `parseVoterConnections` to use a new method called `parseVoterConnection` that knows how to convert a `String` to a `Node`.

jsancio (commented):
Thanks for the PR!

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: @jsancio Not really. The one constructor is used in test cases. However, the tests could just as well build a `RaftConfig`, which is what I assume we would have to do to implement my suggestion above.

> How about storing the RaftConfig as a field of KafkaRaftClient? This would allow us to remove a few of the configuration parameter passed through the constructor.

I debated suggesting that as well. I decided not to because it seemed vexing to do a `Map.get` and parse every time we need the fetch timeout... On the other hand, we could move the fields into `RaftConfig`.

aloknnikhil: I had the same idea as @jsancio. I think it makes sense to store a reference to RaftConfig locally and access these parameters on-demand. And caching these in private fields on construction of the RaftConfig makes sense to me. That would prevent the `Map.get` on every access.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
aloknnikhil: Good catch!

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
aloknnikhil: Ack. Addressing these in the changes. I am moving all computation of voter nodes/ids into RaftConfig. That should simplify this and avoid double work

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Same as above. Will refactor into helper functions on RaftConfig.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -129,16 +133,41 @@ public String toString() {
...
     }
 
+    private final int requestTimeoutMs;
+    private final int retryBackoffMs;
+    private final int electionTimeoutMs;
+    private final int electionBackoffMaxMs;
+    private final int fetchTimeoutMs;
+    private final int appendLingerMs;
+
     public RaftConfig(Properties props) {
```
hachikuji: nit: I think this constructor is redundant. can we remove it?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -129,16 +133,41 @@ public String toString() {
...
+        electionTimeoutMs = getInt(QUORUM_ELECTION_TIMEOUT_MS_CONFIG);
+        electionBackoffMaxMs = getInt(QUORUM_ELECTION_BACKOFF_MAX_MS_CONFIG);
+        fetchTimeoutMs = getInt(QUORUM_FETCH_TIMEOUT_MS_CONFIG);
+        appendLingerMs = getInt(QUORUM_LINGER_MS_CONFIG);
     }
 
-    public RaftConfig(Map<String, Object> props) {
+    public RaftConfig(Map<?, ?> props) {
         super(CONFIG, props);
+        requestTimeoutMs = getInt(QUORUM_REQUEST_TIMEOUT_MS_CONFIG);
```
hachikuji: nit: the typical thing to do is invoke the other constructor. e.g. `this(props, true)`

raft/src/test/java/org/apache/kafka/raft/RaftConfigTest.java
```diff
@@ -24,6 +24,7 @@
 import java.util.HashMap;
 import java.util.Properties;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
```
hachikuji: This does not seem used?

raft/src/test/java/org/apache/kafka/raft/RaftTestUtil.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.net.InetSocketAddress;
+import java.util.List;
+import java.util.Properties;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class RaftTestUtil {
+    public static RaftConfig buildRaftConfig(int requestTimeoutMs,
```
hachikuji: nit: convention in the raft code is to align first parameter on next line:
```
public static RaftConfig buildRaftConfig(
    int requestTimeoutMs,
    int retryBackoffMs,
    ...
    List<Node> voterNodes
) {
```

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -2163,7 +2177,13 @@ public Long scheduleAppend(int epoch, List<T> records) {
 
     @Override
     public void close() {
-        kafkaRaftMetrics.close();
+        if (kafkaRaftMetrics != null) {
+            kafkaRaftMetrics.close();
+        }
+    }
+
+    public QuorumState quorum() {
```
hachikuji: Any chance we can make this default access?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Looks like there are a bunch of tests that seem to be using this constructor
- `testSingleQuorumVoterConnections`
- `testMultiQuorumVoterConnections`
- `testInvalidQuorumVotersConfig`

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Ah thanks for catching that. Turns out both public constructors can use the protected constructor (Properties implements Map). Will refactor both.

aloknnikhil: Actually, that means I can remove the `public RaftConfig(Properties props)` constructor. Does that make sense @hachikuji?

raft/src/test/java/org/apache/kafka/raft/RaftConfigTest.java
aloknnikhil: Done. Removed.

raft/src/test/java/org/apache/kafka/raft/RaftTestUtil.java
aloknnikhil: Fixed.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
hachikuji: Yeah, that's why I said it was redundant in the other comment 🙂

aloknnikhil: Perfect. Removed.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -129,16 +130,25 @@ public String toString() {
...
     }
 
     protected RaftConfig(Map<?, ?> props, boolean doLog) {
         super(CONFIG, props, doLog);
+        requestTimeoutMs = getInt(QUORUM_REQUEST_TIMEOUT_MS_CONFIG);
+        retryBackoffMs = getInt(QUORUM_RETRY_BACKOFF_MS_CONFIG);
+        electionTimeoutMs = getInt(QUORUM_ELECTION_TIMEOUT_MS_CONFIG);
+        electionBackoffMaxMs = getInt(QUORUM_ELECTION_BACKOFF_MAX_MS_CONFIG);
+        fetchTimeoutMs = getInt(QUORUM_FETCH_TIMEOUT_MS_CONFIG);
+        appendLingerMs = getInt(QUORUM_LINGER_MS_CONFIG);
```
hachikuji, line 152: Is it worth storing a field for voter connections as well to be consistent?

aloknnikhil: Yea, I think that's reasonable. Added.

hachikuji (approved):
LGTM. Thanks for the patch!

core/src/main/scala/kafka/raft/RaftManager.scala
```diff
@@ -168,30 +166,20 @@ class KafkaRaftManager[T](
...
     val expirationTimer = new SystemTimer("raft-expiration-executor")
     val expirationService = new TimingWheelExpirationService(expirationTimer)
 
     new KafkaRaftClient(
-      raftConfig,
       recordSerde,
       netChannel,
       metadataLog,
-      quorumState,
+      new FileBasedStateStore(new File(dataDir, "quorum-state")),
```
jsancio, line 177: This is a really important file name and should be stored and documented in a static final field. How about defining this in `FileBasedStateStore` and changing the constructor to `FileBasedStateStore(Path dataDir)`?

If you agree, I can file an issue for this.

cc @aloknnikhil @hachikuji

jsancio (commented):
Excuse the delayed feedback.

core/src/main/scala/kafka/raft/RaftManager.scala
aloknnikhil: Makes sense to me.

aloknnikhil: Wondering if it makes sense to introduce a config (or maybe just the final field) as part of #9916

jsancio: I don't think this should be configurable as it is part of the data (topic partition directory). It cannot conflict with other files in that directory and it cannot change between restart. KIP-595 documents this file and its use.

aloknnikhil: Yea, that's fair. We can leave it for another PR.

PR #9883, opened 2021-01-14
MINOR: Generalize server startup to make way for KIP-500

hachikuji:
This patch attempts to generalize server initialization for KIP-500. It adds a `Server` trait which `KafkaServer` extends for the legacy Zookeeper server, and a new `KafkaRaftServer` for the new server. I have also added stubs for `KafkaRaftBroker` and `KafkaRaftController` to give a clearer idea how this will be used.

This patch attempts to be non-intrusive, but it's worth pointing out that it does away with `KafkaServerStartable`. I believe this was intended to support custom startup logic, but it is not an official API and we do not plan to support it after KIP-500. 


clients/src/main/java/org/apache/kafka/common/config/ConfigDef.java
```diff
@@ -404,13 +404,47 @@ public ConfigDef define(String name, Type type, Importance importance, String do
...
+            "",
+            "",
+            -1,
+            Width.NONE,
+            name,
+            Collections.<String>emptyList(),
+            null,
+            true
+        ));
+    }
```
ijuma, line 426: Nit: it looks pretty odd to have widely differing code styles in the same file.

core/src/main/scala/kafka/server/KafkaRaftBroker.scala
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package kafka.server
+
+class KafkaRaftBroker {
```
ijuma: Maybe we can add a brief comment explaining that this is a stub and not used yet. Similar for KafkaRaftController.

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1453,6 +1456,21 @@ class KafkaConfig(val props: java.util.Map[_, _], doLog: Boolean, dynamicConfigO
...
+  val processRoles = parseProcessRoles()
+
+  private def parseProcessRoles(): Set[ProcessRole] = {
+    val roles = getList(KafkaConfig.ProcessRolesProp).asScala.map {
+      case "broker" => BrokerRole
+      case "controller" => ControllerRole
+      case role => throw new ConfigException(s"Unknown process role $role")
+    }
+
+    if (roles.distinct.length != roles.size) {
```
ijuma: This duplicates the work done in `roles.toSet`, seems like we can assign that to a variable to avoid that.

core/src/main/scala/kafka/server/KafkaRaftController.scala
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package kafka.server
+
+class KafkaRaftController {
```
ijuma: Do you think the `Kafka` prefix adds much value? Would `RaftController`, `RaftBroker` and `RaftServer` be confusing/less clear?

ijuma (commented):
Thanks for the PR. Initial comments/questions below.

core/src/main/scala/kafka/server/KafkaRaftController.scala
hachikuji: Hmm, I found it a little strange if "Raft" is the only prefix. The `RaftController` is not controlling Raft... I've considered calling it `MetadataController` previously. To go along with that, we could have `DataBroker` or just `Broker`.

clients/src/main/java/org/apache/kafka/common/config/ConfigDef.java
hachikuji: Fair enough, I'll conform. I just find the other style hard to read.

core/src/main/scala/kafka/server/KafkaRaftController.scala
ijuma: Yeah, I agree that it reads a bit strange without the `Kafka` prefix. I think what you have in the PR is fine.

core/src/main/scala/kafka/raft/RaftManager.scala
```diff
@@ -95,15 +95,15 @@ trait RaftManager[T] {
...
-  baseLogDir: String,
   recordSerde: RecordSerde[T],
   topicPartition: TopicPartition,
   config: KafkaConfig,
   time: Time,
   metrics: Metrics
 ) extends RaftManager[T] with Logging {
 
+  private val nodeId = config.brokerId
+  private val baseLogDir = config.logDirs.head
```
cmccabe: this isn't quite correct since we could also have a specifically configured metadata.log.dir

core/src/main/scala/kafka/server/KafkaRaftController.scala
cmccabe: How about RaftBrokerServer and ControllerServer?  That would help differentiate this from the Controller / QuorumController classes.  It would also be consistent with KafkaServer.

I think it's also more accurate-- these classes manage "all the server stuff" including socket server, etc.

core/src/main/scala/kafka/raft/RaftManager.scala
hachikuji: Right. There are a few placeholders in here until we get all of the configs in.

core/src/main/scala/kafka/server/KafkaRaftController.scala
hachikuji: Or maybe just `Controller` and `Broker`?

hachikuji: Decided to go with `ControllerServer` and `BrokerServer`. As Colin mentioned offline, the "server" suffix is helpful conveying that this component is responsible for starting sockets and such. This distinguishes `ControllerServer` from the actual implementation of the controller logic.

core/src/main/scala/kafka/server/KafkaRaftServer.scala
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.kafka.common.TopicPartition
+import org.apache.kafka.common.utils.{AppInfoParser, Time}
+import org.apache.kafka.raft.internals.StringSerde
+
+/**
+ * Partially stubbed implementation of the KIP-500 server which relies on a self-managed
+ * Raft quorum for replication of the `@metadata` topic, which stores all of
+ * the cluster metadata.
+ */
+class KafkaRaftServer(
```
cmccabe, line 37: Can we have a comment at least describing the fact that it can contain either a controller, broker, or both

hachikuji: Makes sense. Added a comment above.

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1453,6 +1456,23 @@ class KafkaConfig(val props: java.util.Map[_, _], doLog: Boolean, dynamicConfigO
   val brokerIdGenerationEnable: Boolean = getBoolean(KafkaConfig.BrokerIdGenerationEnableProp)
   val maxReservedBrokerId: Int = getInt(KafkaConfig.MaxReservedBrokerIdProp)
   var brokerId: Int = getInt(KafkaConfig.BrokerIdProp)
+  val processRoles = parseProcessRoles()
+
+  private def parseProcessRoles(): Set[ProcessRole] = {
+    val roles = getList(KafkaConfig.ProcessRolesProp).asScala.map {
+      case "broker" => BrokerRole
+      case "controller" => ControllerRole
+      case role => throw new ConfigException(s"Unknown process role $role")
```
cmccabe: can you list the valid roles in this exception?

cmccabe:
LGTM.  I had one small comment.

Thanks, @hachikuji

PR #9886, opened 2021-01-14
MINOR: Rename examples to example

wenbingshen:
As the title


chia7712 (approved):
@wenbingshen Thanks for your patch. LGTM

PR #9887, opened 2021-01-14
KAFKA-12195 Fix synchronization issue happening in KafkaStreams

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-12195
related to aedb53a4e6313398ed9626fbd2fbd7106e2ce94e

## Root Cause
It seems to me there are two issues on ```AdjustStreamThreadCountTest```.

### 1) synchronization issue (the number of threads is inconsistent)
```
threads = Collections.synchronizedList(new LinkedList<>());
```

The synchronization list requires us to manually synchronize the iterator. non-synchronizing the list results in inconsistent results and consequently unstabilize the AdjustStreamThreadCountTest.

### 2) duplicate thread index when adding new stream thread

```createAndAddStreamThread``` is not called with holding ```changeThreadCount``` so it is possible that we create two threads with same thread index. It makes different threads have same metadata and then ```localThreadsMetadata``` returns incorrect number of metadata.

### 3) assert intermediate state (it gets failed if it run too fast)

This PR already reverts the fix and it is traced by #9888

## Test Result from PR

looped the test 100 times on my local. all pass

### CODE REVIEW

File: streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java [hunk 1]
```diff
@@ -119,7 +119,6 @@ public void shouldAddStreamThread() throws Exception {
                     .sorted().toArray(),
                 equalTo(new String[] {"1", "2", "3"})
             );
-            waitForApplicationState(Collections.singletonList(kafkaStreams), KafkaStreams.State.REBALANCING, DEFAULT_DURATION);
```
chia7712: ```REBALANCING``` is an intermediate state so this check can fail if the machine running this test case is too fast.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -989,16 +986,20 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+                            streamThread.shutdown();
+                            if (!streamThread.getName().equals(Thread.currentThread().getName())) {
+                                streamThread.waitOnThreadState(StreamThread.State.DEAD);
+                            }
+                            iter.remove();
+                            final long cacheSizePerThread = getCacheSizePerThread(threads.size());
+                            resizeThreadCache(cacheSizePerThread);
+                            return Optional.of(streamThread.getName());
                         }
-                        threads.remove(streamThread);
```
chia7712: Removing element from collection when looping it is error-prone so I rewrite it by iterator.

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
chia7712: @showuon I can revert this change for your PR #9888 if you brings a better solution :)

showuon: Agree! The test change I have investigated and did much refactor on that. But the synchronization issue is not what I covered. Nice catch!

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -148,7 +146,6 @@ public void shouldAddAndRemoveThreads() throws InterruptedException {
             one.start();
             latch.await(30, TimeUnit.SECONDS);
             assertThat(kafkaStreams.localThreadsMetadata().size(), equalTo(oldThreadCount));
```
chia7712: This assert gets failed due to synchronization issue.

```
java.lang.AssertionError: 
Expected: <2>
     but: was <1>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.apache.kafka.streams.integration.AdjustStreamThreadCountTest.shouldAddAndRemoveThreads(AdjustStreamThreadCountTest.java:149)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:288)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:282)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -230,4 +235,33 @@ public void shouldAddAndRemoveStreamThreadsWhileKeepingNamesCorrect() throws Exc
             waitForApplicationState(Collections.singletonList(kafkaStreams), KafkaStreams.State.RUNNING, DEFAULT_DURATION);
         }
     }
+
+    @Test
+    public void testConcurrentlyAccessThreads() throws InterruptedException {
```
chia7712, line 240: This test case can cause following exception if we don't apply this patch.
```
org.apache.kafka.streams.integration.AdjustStreamThreadCountTest > testConcurrentlyAccessThreads FAILED
    java.lang.AssertionError: expected null, but was:<java.util.ConcurrentModificationException>
        at org.junit.Assert.failNotNull(Assert.java:756)
        at org.junit.Assert.assertNull(Assert.java:738)
        at org.junit.Assert.assertNull(Assert.java:748)
        at org.apache.kafka.streams.integration.AdjustStreamThreadCountTest.testConcurrentlyAccessThreads(AdjustStreamThreadCountTest.java:264)
```

chia7712, line 291: This test case can cause following error if we don't apply this patch.

```
org.apache.kafka.streams.integration.AdjustStreamThreadCountTest > testConcurrentlyAccessThreads FAILED
    java.lang.AssertionError: expected null, but was:<java.util.ConcurrentModificationException>
        at org.junit.Assert.failNotNull(Assert.java:756)
        at org.junit.Assert.assertNull(Assert.java:738)
        at org.junit.Assert.assertNull(Assert.java:748)
        at org.apache.kafka.streams.integration.AdjustStreamThreadCountTest.testConcurrentlyAccessThreads(AdjustStreamThreadCountTest.java:264)
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -951,12 +948,15 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
             final int threadIdx;
             final long cacheSizePerThread;
+            final StreamThread streamThread;
             synchronized (changeThreadCount) {
                 threadIdx = getNextThreadIndex();
                 cacheSizePerThread = getCacheSizePerThread(threads.size() + 1);
                 resizeThreadCache(cacheSizePerThread);
+                // Creating thread should hold the lock in order to avoid duplicate thread index.
+                // If the duplicate index happen, the metadata of thread may be duplicate too.
+                streamThread = createAndAddStreamThread(cacheSizePerThread, threadIdx);
```
chia7712, line 957: fix for case 2)

chia7712:
@cadonna Could you take a look at those concurrent issues (see description) ? I'm digging in it to observe whether there are more issues. Thanks!

ableegoldman:
Hey @wcarlson5 can you also take a look at this?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -989,16 +989,20 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
     public Optional<String> removeStreamThread() {
         if (isRunningOrRebalancing()) {
             synchronized (changeThreadCount) {
-                for (final StreamThread streamThread : threads) {
-                    if (streamThread.isAlive() && (!streamThread.getName().equals(Thread.currentThread().getName()) || threads.size() == 1)) {
-                        streamThread.shutdown();
-                        if (!streamThread.getName().equals(Thread.currentThread().getName())) {
-                            streamThread.waitOnThreadState(StreamThread.State.DEAD);
+                synchronized (threads) {
```
wcarlson5: Why are we using the threads object as a lock here? is that was the `Collections.synchronizedList(new LinkedList<>());` uses internally?

I am a little concerned that we are holding `changeThreadCount` while waiting for `threads`

wcarlson5 (commented):
This doesn't seems to change the actual logic so that shouldn't be a problem. The copy of the thread list to avoid an extra lock was a good idea. With #9888 this should fix the issues we are seeing and I don't see any other problems with it. @ableegoldman the synchronization of the thread list iterators was a possible issue we talked about. I didn't think that it would be a problem if we were not altering the list but it appears that I was wrong. I only have one concern otherwise LGTM. @chia7712  thanks for cleaning this up!

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
chia7712: make sense. Let me use 'copy' to handle lock issue.

guozhangwang:
I made a pass and it looked good. @chia7712 leaving it to your call on the merging order between this one and #9888.

guozhangwang:
The unit tests may not be stable until these PRs are merged.

chia7712:
merge #9888 to run QA again.

PR #9895, opened 2021-01-14
KAFKA-9924: Add docs for RocksDB properties-based metrics

cadonna:

### DIFF
--- a/docs/ops.html
+++ b/docs/ops.html
@@ -2032,9 +2032,11 @@ dropped-records-rate and dropped-records-total which have a recording level of <
  </table>
 
   <h5 class="anchor-heading"><a id="kafka_streams_rocksdb_monitoring" class="anchor-link"></a><a href="#kafka_streams_rocksdb_monitoring">RocksDB Metrics</a></h5>
-  RocksDB Metrics are grouped into statistics-based metrics and properties-based metrics.
+  RocksDB metrics are grouped into statistics-based metrics and properties-based metrics.
   The former are recorded from statistics that a RocksDB state store collects whereas the latter are recorded from
   properties that RocksDB exposes.
+  Statistics collected by RocksDB provide cumulative measurements over time, e.g. bytes written to the state store.
+  Properties exposed by RocksDB provide current measurements, e.g., the amount of memory currently used.
   Note that the <code>store-scope</code> for built-in RocksDB state stores are currently the following:
   <ul>
     <li><code>rocksdb-state</code> (for RocksDB backed key-value store)</li>
@@ -2042,7 +2044,6 @@ dropped-records-rate and dropped-records-total which have a recording level of <
     <li><code>rocksdb-session-state</code> (for RocksDB backed session store)</li>
   </ul>
 
-  <!-- h6 class="anchor-heading"><a id="kafka_streams_rocksdb_statistics_based_monitoring" class="anchor-link"></a><a href="#kafka_streams_rocksdb_statistics_based_monitoring">RocksDB Statistics-based Metrics</a></h6-->
   <strong>RocksDB Statistics-based Metrics:</strong>
   All of the following statistics-based metrics have a recording level of <code>debug</code> because collecting
   statistics in <a href="https://github.com/facebook/rocksdb/wiki/Statistics#stats-level-and-performance-costs">RocksDB
@@ -2141,7 +2142,6 @@ dropped-records-rate and dropped-records-total which have a recording level of <
     </tbody>
   </table>
 
-  <!-- h6 class="anchor-heading"><a id="kafka_streams_rocksdb_properties_based_monitoring" class="anchor-link"></a><a href="#kafka_streams_rocksdb_properties_based_monitoring">RocksDB Properties-based Metrics</a></h6-->
   <strong>RocksDB Properties-based Metrics:</strong>
   All of the following properties-based metrics have a recording level of <code>info</code> and are recorded when the
   metrics are accessed.
--- a/docs/ops.html
+++ b/docs/ops.html
@@ -2048,7 +2048,7 @@ dropped-records-rate and dropped-records-total which have a recording level of <
   statistics in <a href="https://github.com/facebook/rocksdb/wiki/Statistics#stats-level-and-performance-costs">RocksDB
   may have an impact on performance</a>.
   Statistics-based metrics are collected every minute from the RocksDB state stores.
-  If a state store consists of multiple RocksDB instances as it is the case for aggregations over time and session windows,
+  If a state store consists of multiple RocksDB instances, as is the case for aggregations over time and session windows,
   each metric reports an aggregation over the RocksDB instances of the state store.
 
   <table class="data-table">
@@ -2145,8 +2145,8 @@ dropped-records-rate and dropped-records-total which have a recording level of <
   <strong>RocksDB Properties-based Metrics:</strong>
   All of the following properties-based metrics have a recording level of <code>info</code> and are recorded when the
   metrics are accessed.
-  If a state store consists of multiple RocksDB instances as it is the case for aggregations over time and session windows,
-  each metric reports the sum over the RocksDB instances of the state store, except for the block cache metrics
+  If a state store consists of multiple RocksDB instances, as is the case for aggregations over time and session windows,
+  each metric reports the sum over all the RocksDB instances of the state store, except for the block cache metrics
   <code>block-cache-*</code>. The block cache metrics report the sum over all RocksDB instances if each instance uses its
   own block cache, and they report the recorded value from only one instance if a single block cache is shared
   among all instances.
@@ -2221,7 +2221,7 @@ dropped-records-rate and dropped-records-total which have a recording level of <
     <tr>
       <td>estimate-pending-compaction-bytes</td>
       <td>The estimated total number of bytes a compaction needs to rewrite on disk to get all levels down to under
-        target size (only valid for level compaction.</td>
+        target size (only valid for level compaction).</td>
       <td>kafka.streams:type=stream-state-metrics,thread-id=([-.\w]+),task-id=([-.\w]+),[store-scope]-id=([-.\w]+)</td>
     </tr>
     <tr>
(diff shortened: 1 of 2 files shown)


cadonna:
Call for review: @guozhangwang @ableegoldman @lct45

docs/ops.html
```diff
@@ -2032,17 +2032,25 @@ <h5 class="anchor-heading"><a id="kafka_streams_store_monitoring" class="anchor-
  </table>
 
   <h5 class="anchor-heading"><a id="kafka_streams_rocksdb_monitoring" class="anchor-link"></a><a href="#kafka_streams_rocksdb_monitoring">RocksDB Metrics</a></h5>
-  All of the following metrics have a recording level of <code>debug</code>.
-  The metrics are collected every minute from the RocksDB state stores.
-  If a state store consists of multiple RocksDB instances as it is the case for aggregations over time and session windows,
-  each metric reports an aggregation over the RocksDB instances of the state store.
+  RocksDB Metrics are grouped into statistics-based metrics and properties-based metrics.
+  The former are recorded from statistics that a RocksDB state store collects whereas the latter are recorded from
+  properties that RocksDB exposes.
```
lct45, line 2037: I'm still a little confused about the difference between the two after reading this. Maybe an example of something that is exposed by RocksDB but not collected by RocksDB would help

docs/ops.html
```diff
@@ -2032,17 +2032,25 @@ <h5 class="anchor-heading"><a id="kafka_streams_store_monitoring" class="anchor-
...
     <li><code>rocksdb-session-state</code> (for RocksDB backed session store)</li>
   </ul>
 
+  <!-- h6 class="anchor-heading"><a id="kafka_streams_rocksdb_statistics_based_monitoring" class="anchor-link"></a><a href="#kafka_streams_rocksdb_statistics_based_monitoring">RocksDB Statistics-based Metrics</a></h6-->
+  <strong>RocksDB Statistics-based Metrics:</strong>
+  All of the following statistics-based metrics have a recording level of <code>debug</code> because collecting
+  statistics in <a href="https://github.com/facebook/rocksdb/wiki/Statistics#stats-level-and-performance-costs">RocksDB
+  may have an impact on performance</a>.
+  Statistics-based metrics are collected every minute from the RocksDB state stores.
+  If a state store consists of multiple RocksDB instances as it is the case for aggregations over time and session windows,
```
lct45:
```suggestion
  If a state store consists of multiple RocksDB instances, as is the case for aggregations over time and session windows,
```

docs/ops.html
```diff
@@ -2133,6 +2141,137 @@ <h5 class="anchor-heading"><a id="kafka_streams_rocksdb_monitoring" class="ancho
     </tbody>
   </table>
 
+  <!-- h6 class="anchor-heading"><a id="kafka_streams_rocksdb_properties_based_monitoring" class="anchor-link"></a><a href="#kafka_streams_rocksdb_properties_based_monitoring">RocksDB Properties-based Metrics</a></h6-->
+  <strong>RocksDB Properties-based Metrics:</strong>
+  All of the following properties-based metrics have a recording level of <code>info</code> and are recorded when the
+  metrics are accessed.
+  If a state store consists of multiple RocksDB instances as it is the case for aggregations over time and session windows,
```
lct45:
```suggestion
  If a state store consists of multiple RocksDB instances, as is the case for aggregations over time and session windows,
```

docs/ops.html
```diff
@@ -2133,6 +2141,137 @@ <h5 class="anchor-heading"><a id="kafka_streams_rocksdb_monitoring" class="ancho
...
+  each metric reports the sum over the RocksDB instances of the state store, except for the block cache metrics
```
lct45:
```suggestion
  each metric reports the sum over all the RocksDB instances of the state store, except for the block cache metrics
```

docs/ops.html
```diff
@@ -2133,6 +2141,137 @@ <h5 class="anchor-heading"><a id="kafka_streams_rocksdb_monitoring" class="ancho
...
+    </tr>
+    <tr>
+      <td>num-running-compactions</td>
+      <td>The number of currently running compactions.</td>
+      <td>kafka.streams:type=stream-state-metrics,thread-id=([-.\w]+),task-id=([-.\w]+),[store-scope]-id=([-.\w]+)</td>
+    </tr>
+    <tr>
+      <td>estimate-pending-compaction-bytes</td>
+      <td>The estimated total number of bytes a compaction needs to rewrite on disk to get all levels down to under
+        target size (only valid for level compaction.</td>
```
lct45:
```suggestion
        target size (only valid for level compaction).</td>
```

lct45 (approved):
Thanks for the pr @cadonna ! Just left a few nits

docs/ops.html
cadonna: I elaborated on the two types of metrics. Let me know if it is better now.

ableegoldman: Thanks for the explanation, I also didn't understand the difference at first

docs/ops.html
```diff
@@ -2032,17 +2032,26 @@ <h5 class="anchor-heading"><a id="kafka_streams_store_monitoring" class="anchor-
...
     <li><code>rocksdb-window-state</code> (for RocksDB backed window store)</li>
     <li><code>rocksdb-session-state</code> (for RocksDB backed session store)</li>
   </ul>
 
+  <strong>RocksDB Statistics-based Metrics:</strong>
+  All of the following statistics-based metrics have a recording level of <code>debug</code> because collecting
+  statistics in <a href="https://github.com/facebook/rocksdb/wiki/Statistics#stats-level-and-performance-costs">RocksDB
+  may have an impact on performance</a>.
+  Statistics-based metrics are collected every minute from the RocksDB state stores.
+  If a state store consists of multiple RocksDB instances, as is the case for aggregations over time and session windows,
```
ableegoldman:
```suggestion
  If a state store consists of multiple RocksDB instances, as is the case for WindowStores and SessionStores,
```
I think we should be more specific here. Also users can have a Window/SessionStore that's unattached to an aggregation, eg in the PAPI, so we should avoid explicitly tying this to aggregations

docs/ops.html
```diff
@@ -2133,6 +2142,136 @@ <h5 class="anchor-heading"><a id="kafka_streams_rocksdb_monitoring" class="ancho
     </tbody>
   </table>
 
+  <strong>RocksDB Properties-based Metrics:</strong>
+  All of the following properties-based metrics have a recording level of <code>info</code> and are recorded when the
+  metrics are accessed.
+  If a state store consists of multiple RocksDB instances, as is the case for aggregations over time and session windows,
```
ableegoldman:
```suggestion
  If a state store consists of multiple RocksDB instances, as is the case for WindowStores and SessionStores,
```

docs/ops.html
```diff
@@ -2133,6 +2142,136 @@ <h5 class="anchor-heading"><a id="kafka_streams_rocksdb_monitoring" class="ancho
...
+  each metric reports the sum over all the RocksDB instances of the state store, except for the block cache metrics
+  <code>block-cache-*</code>. The block cache metrics report the sum over all RocksDB instances if each instance uses its
+  own block cache, and they report the recorded value from only one instance if a single block cache is shared
```
ableegoldman, line 2151: Technically I think it's possible for users to share a single block cache among only some stores, but not others. Or they could have two block caches shared across different state stores, etc. Would we detect this case and report the correct block cache for a given state store?
(Idk how common that pattern could possibly be, but this made me wonder)

docs/ops.html
```diff
@@ -2133,6 +2142,136 @@ <h5 class="anchor-heading"><a id="kafka_streams_rocksdb_monitoring" class="ancho
...
+      <td>kafka.streams:type=stream-state-metrics,thread-id=([-.\w]+),task-id=([-.\w]+),[store-scope]-id=([-.\w]+)</td>
+    </tr>
+    <tr>
+      <td>num-running-flushes</td>
+      <td>The number of currently running flushes.</td>
+      <td>kafka.streams:type=stream-state-metrics,thread-id=([-.\w]+),task-id=([-.\w]+),[store-scope]-id=([-.\w]+)</td>
+    </tr>
+    <tr>
+      <td>compaction-pending</td>
+      <td>This metric reports 1 if at least one compaction is pending, otherwise it reports 0.</td>
```
ableegoldman, line 2213: Do you know if this applies only to a single rocksdb instance, or across all instances? By default rocksdb has a single shared Environment (basically a thread pool) for compactions so it seems like it could go either way. But it's ok if you don't know, I wouldn't waste hours trying to understand the rocksdb code trying to figure it out 😉

docs/ops.html
cadonna: Good point! Currently, we treat this mixed pattern as an illegal state and throw an `IllegalStateException`. Probably not the best way to handle it. Allowing such a mixed pattern complicates the measurement of the cache metrics. I opened [KAFKA-12223](https://issues.apache.org/jira/browse/KAFKA-12223) to document this.

docs/ops.html
cadonna: No, I do not know. I supposed that it only applies to one single RocksDB instance. I am wondering how RocksDB can share a thread pool between state stores that are started independently from each other.

docs/ops.html
lct45: Ahh I see now, thanks for the additional info! LGTM

docs/ops.html
ableegoldman: I think it has to do with the `Env` class which specifies the thread pool. By default it uses a static `Env`, so the Env -- and also the underlying threads -- are shared between all stores within the process. Something like that

docs/ops.html
ableegoldman: Thanks!

ableegoldman (approved):
LGTM

ableegoldman:

ijuma:
Looks like this broke the build, see https://github.com/apache/kafka/pull/9935.

I can see a number of test failures in the last Jenkins job, let's please verify the tests before merging.

ableegoldman:
@ijuma I'm confused, I don't see any failures for the test that this broke (RocksDBMetricsTest) in the build on the last commit. I saw some test failures, but all unrelated. Am I looking in the wrong place? 
(I followed the link on the "Apply suggestions from code review" commit build --> https://github.com/apache/kafka/runs/1722608276 )

ijuma:
Jenkins is the best place to look:

Screenshot from 2021-01-20 10-36-13

PR #9902, opened 2021-01-15
KAFKA-12193: Re-resolve IPs after a client disconnects

bob-barrett:
This patch changes the NetworkClient behavior to resolve the target node's hostname after disconnecting from an established connection, rather than waiting until the previously-resolved addresses are exhausted. This is to handle the scenario when the node's IP addresses have changed during the lifetime of the connection, and means that the client does not have to try to connect to invalid IP addresses until it has tried each address.


clients/src/main/java/org/apache/kafka/clients/NetworkClient.java
```diff
@@ -239,24 +241,63 @@ public NetworkClient(Selectable selector,
...
-                          int defaultRequestTimeoutMs,
-                          long connectionSetupTimeoutMs,
-                          long connectionSetupTimeoutMaxMs,
-                          ClientDnsLookup clientDnsLookup,
-                          Time time,
-                          boolean discoverBrokerVersions,
-                          ApiVersions apiVersions,
-                          Sensor throttleTimeSensor,
-                          LogContext logContext) {
+    public NetworkClient(MetadataUpdater metadataUpdater,
```
mimaison: Do we really need this constructor? As far as I can tell, it's only called by the other 3 above. These could directly call the real one below instead of going through this new one. WDYT?

clients/src/main/java/org/apache/kafka/clients/ClusterConnectionStates.java
```diff
@@ -183,6 +186,10 @@ public void disconnected(String id, long now) {
             connectingNodes.remove(id);
         } else {
             resetConnectionSetupTimeout(nodeState);
+            if (nodeState.state.isConnected()) {
+                // If a connection had previously been established, re-resolve DNS because the IPs may have changed
+                nodeState.addresses = Collections.emptyList();
```
mimaison: Slightly weird that we're updating a "private" field here.

Also the comment is a bit misleading. We're not re-resolving DNS here but instead clearing state so if we reconnect later, the client will be forced to re-resolve then.

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
```diff
@@ -907,6 +924,140 @@ public void testCorrelationId() {
...
+        AtomicInteger initialAddressConns = new AtomicInteger();
+        AtomicInteger newAddressConns = new AtomicInteger();
+        MockSelector selector = new MockSelector(this.time, inetSocketAddress -> {
+            InetAddress inetAddress = inetSocketAddress.getAddress();
+            if (initialAddresses.contains(inetAddress)) {
+                initialAddressConns.incrementAndGet();
+            } else if (newAddresses.contains(inetAddress)) {
+                newAddressConns.incrementAndGet();
+            }
+            return (mockHostResolver.getUseNewAddresses() && newAddresses.contains(inetAddress)) ||
```
mimaison: I was slightly confused until I realized `getUseNewAddresses()` is a boolean. Maybe `useNewAddresses()` would be a better name?

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
```diff
@@ -907,6 +924,140 @@ public void testCorrelationId() {
...
+
+        // Second connection attempt to new addresses should succeed
+        time.sleep(reconnectBackoffMaxMsTest);
+        client.ready(node, time.milliseconds());
+        time.sleep(connectionSetupTimeoutMaxMsTest);
+        client.poll(0, time.milliseconds());
+        assertTrue(client.isReady(node, time.milliseconds()));
+
+        // We should have tried to connect to two of the initial addresses, none of the new address, and should
+        // only have performed one DNS resolution
```
mimaison: This comment needs updating

mimaison (changes requested):
Thanks for the PR. It looks good overall, I left a few comments for minor issues

clients/src/main/java/org/apache/kafka/clients/ClusterConnectionStates.java
satishd: @mimaison Java allows outer classes to access inner member's private members. This is described in detail at [JLS-6.6.1](https://docs.oracle.com/javase/specs/jls/se7/html/jls-6.html#jls-6.6.1)

> A member (class, interface, field, or method) of a reference (class, interface, or array) type or a constructor of a class type is accessible only if the type is accessible and the member or constructor is declared to permit access:
>  - if the member or constructor is declared private, then access is permitted if and only if it occurs within the body of the top level class (§7.6) that encloses the declaration of the member or constructor.

satishd: Agree with @mimaison on the comment, good to make it clear by mentioning that the addresses are cleared to re-resolve later when it reconnects.

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
bob-barrett: Yep, thanks for catching that

clients/src/main/java/org/apache/kafka/clients/NetworkClient.java
bob-barrett: Good call, fixed. These are also weirdly inconsistent (`selector` is before `metadata` and `metadataUpdater` sometimes, after them other times), which I guess we could clean up, but I didn't want to remove any existing public constructors.

clients/src/main/java/org/apache/kafka/clients/ClusterConnectionStates.java
bob-barrett: Accessing the field is allowed, but I think @mimaison meant that it makes the code a little confusing and inconsistent. `NodeConnectionState` has some private fields and some package-private ones, and `addresses` previous was only modified by methods on `NodeConnectionState`, not accessed directly by the top-level class. I've moved the empty list assignment into a `NodeConnectionState.clearAddresses` method

Agree about the comment, I've updated it

satishd (approved):
thanks @bob-barrett for addressing the review comments. LGTM

dajac:
@bob-barrett Thanks for the PR. For my understanding, have we considered changing the way backoff is applied instead of re-resolving when an established connection is closed. For instance, we could try all the resolved IPs before applying the backoff. However, we would still have to wait until the connection timeouts for each IP.

Also, I suppose that we assume that closing connection is rather infrequent in normal circumstances so the cost of the DNS resolution is low. On the other hand, we relies on the DNS cache if the resolution happens really frequently. Am I getting this right? That seems reasonable. 

I am trying to convince myself that this is the correct approach to address the issue.

bob-barrett:
> For instance, we could try all the resolved IPs before applying the backoff. However, we would still have to wait until the connection timeouts for each IP.

@dajac That's right, and that's my concern with exhausting the currently-resolved IPs. The default connection setup timeout is 10 seconds. Even without applying the backoff, 10 seconds of delay per additional IP address is pretty noticeable.

> Also, I suppose that we assume that closing connection is rather infrequent in normal circumstances so the cost of the DNS resolution is low. On the other hand, we relies on the DNS cache if the resolution happens really frequently. Am I getting this right?

I think this is the right way to think about it, yes. There's already a DNS cache, which can be configured as needed. We're effectively adding our own additional cache (with a very long TTL) on top of that by preserving the resolved IPs. If a client has been connected to a server for hours or days and then disconnects, it makes complete sense to me to re-resolve DNS.

dajac:
@bob-barrett Thanks for your answers. That makes sense and I do agree with the approach.

clients/src/main/java/org/apache/kafka/clients/ClientUtils.java
```diff
@@ -106,8 +106,9 @@ public static ChannelBuilder createChannelBuilder(AbstractConfig config, Time ti
                 clientSaslMechanism, time, true, logContext);
     }
 
-    static List<InetAddress> resolve(String host, ClientDnsLookup clientDnsLookup) throws UnknownHostException {
-        InetAddress[] addresses = InetAddress.getAllByName(host);
+    static List<InetAddress> resolve(String host, ClientDnsLookup clientDnsLookup,
+                                     HostResolver hostResolver) throws UnknownHostException {
```
dajac, line 110: This method and its usages look a bit weird to me now. It resolves an `host` based on `clientDnsLookup` and a `hostResolver` which resolves an `host` at its turn. We also have to pass the `HostResolver` wherever it is called now. I wonder if we should avoid this level of indirection here and directly encapsulate the entire logic in the `HostResolver`.

`HostResolver` would look like this:
```
public interface HostResolver {
    List<InetAddress> resolve(String host, ClientDnsLookup clientDnsLookup) throws UnknownHostException;
}
```

In `NodeConnectionState#currentAddress`, we could replace `addresses = ClientUtils.resolve(host, clientDnsLookup, hostResolver)` by `addresses = hostResolver.resolve(host, clientDnsLookup)`.

The `DefaultHostResolver` would do what `ClientUtils.resolve` does today and we can mock it in tests like you did.

Is this something that you have considered?

dajac (commented):
@bob-barrett Thanks for the PR. I have left a meta-comment. Please, let me know what you think. Besides this, the rest looks good.

clients/src/main/java/org/apache/kafka/clients/ClientUtils.java
bob-barrett: @dajac I started making this change, but while working on it, I realized that I don't think I like requiring the `HostResolver` to understand the `client.dns.lookup` config. In the same way that `MockTime` is only used to mock Java built-in methods, I think the `HostResolver` should only be responsible for the DNS resolution (either real or mocked), and that `ClientUtils` should be responsible for understanding the Kafka application behavior. That way, tests that mock DNS still test the actual `ClientDnsLookup` behavior, rather than relying on both the default resolver and any mocked resolvers handle it correctly. It also means that if we ever add an option for `ClientDnsLookup` or change its behavior, we don't need to remember to update all `HostResolver` implementations.

dajac: Thanks @bob-barrett. I do agree with your point. Let's keep it as you suggested.

clients/src/main/java/org/apache/kafka/clients/DefaultHostResolver.java
```diff
@@ -0,0 +1,29 @@
...
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+public class DefaultHostResolver implements HostResolver {
+
+    @Override
+    public InetAddress[] resolve(String host) throws UnknownHostException {
+        return InetAddress.getAllByName(host);
+    }
+}
```
dajac: nit: Could we add an empty line here?

clients/src/main/java/org/apache/kafka/clients/HostResolver.java
```diff
@@ -0,0 +1,26 @@
...
+
+package org.apache.kafka.clients;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+public interface HostResolver {
+
+    InetAddress[] resolve(String host) throws UnknownHostException;
+}
```
dajac: nit: Could we add an empty line here?

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
```diff
@@ -81,6 +85,19 @@
     private final NetworkClient clientWithNoExponentialBackoff = createNetworkClient(reconnectBackoffMsTest);
     private final NetworkClient clientWithStaticNodes = createNetworkClientWithStaticNodes();
     private final NetworkClient clientWithNoVersionDiscovery = createNetworkClientWithNoVersionDiscovery();
+    private ArrayList<InetAddress> initialAddresses = new ArrayList<>(Arrays.asList(
```
dajac: nit: Should we make these two static?

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
```diff
@@ -960,4 +1112,29 @@ public KafkaException getAndClearFailure() {
             return failure;
         }
     }
+
+    private class AddressChangeHostResolver implements HostResolver {
+        private boolean useNewAddresses;
+        private InetAddress[] initialAddressArray = initialAddresses.toArray(new InetAddress[0]);
+        private InetAddress[] newAddressArray = newAddresses.toArray(new InetAddress[0]);
```
dajac: nit: I wonder if it would be better to pass these two in a constructor? It makes reading the tests a bit more easy.

dajac (commented):
@bob-barrett Thanks for your reply. I left few minor comments.

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
bob-barrett: Catching the `UnknownHostException` makes it a bit annoying, I threw it in a static block to make the compiler happy. Let me know if you think there's something cleaner.

bob-barrett:
@dajac @mimaison @satishd Thanks for your reviews!

In addition to PR feedback, I pushed a new change to `ClusterConnectionStatesTest`. I noticed while testing locally that tests that depended on `kafka.apache.org` were failing. In https://github.com/apache/kafka/pull/9294, we changed the tests to assume that the site resolved to 3 IPv4 addresses. However, there appear to now be 2 IPv4 addresses and one IPv6, which broke the tests, since `ClientUtils#resolve` will only return the addresses of one version or the other, not both. With the `HostResolver` interface, it was straightforward to mock the DNS out instead of actually resolving `kafka.apache.org`, which should make the tests less fragile. This also may accomplish the intent of https://issues.apache.org/jira/browse/KAFKA-10496 without needing to create in-memory DNS.

mimaison (approved):
Thanks for the updates. LGTM

mimaison:
Tests passed on JDK 8.
The failure on JDK 11 (`kafka.api.TransactionsTest.testAbortTransactionTimeout()`) looks unrelated.

dajac:
I have picked the commit into 2.7 but could not pick it into 2.6 cleanly. @bob-barrett Could you open a PR for 2.6?

clients/src/test/java/org/apache/kafka/clients/NetworkClientTest.java
itweixiang: hi bob-barrett , I hava a quetion about your issue ,can you help me ? 
our kafka are deloying in k8s , where kafka cluster restart , result in producer and consumer will disconnect.
kafka cluster restart will generate new ips,  but kafka client store old ips . so we must restart producer and consumer , so tired.

in kafka client 2.8.1 version ,I see  org.apache.kafka.clients.NetworkClient#initiateConnect , `node.host()` maybe return a ip instead of domain.   if return a ip , will result in lose efficacy about your updated version .

my english level is match awful ， can you know my words?  

```
private void initiateConnect(Node node, long now) {
        String nodeConnectionId = node.idString();
        try {
            connectionStates.connecting(nodeConnectionId, now, node.host(), clientDnsLookup);
            InetAddress address = connectionStates.currentAddress(nodeConnectionId);
            log.debug("Initiating connection to node {} using address {}", node, address);
            selector.connect(nodeConnectionId,
                    new InetSocketAddress(address, node.port()),
                    this.socketSendBuffer,
                    this.socketReceiveBuffer);
        } catch (IOException e) {
            log.warn("Error connecting to node {}", node, e);
            // Attempt failed, we'll try again after the backoff
            connectionStates.disconnected(nodeConnectionId, now);
            // Notify metadata updater of the connection failure
            metadataUpdater.handleServerDisconnect(now, nodeConnectionId, Optional.empty());
        }
    }
```

xuchenxin:
您好，我已收到您的来信，会尽快回复。谢谢！

itweixiang:
hi bob-barrett , I hava a quetion about your issue ,can you help me ? 
our kafka are deloying in k8s , where kafka cluster restart , result in producer and consumer will disconnect.
kafka cluster restart will generate new ips,  but kafka client store old ips . so we must restart producer and consumer , so tired.

in kafka client 2.8.1 version ,I see  org.apache.kafka.clients.NetworkClient#initiateConnect , `node.host()` maybe return a ip instead of domain.   if return a ip , will result in lose efficacy about your updated version .

my english level is match awful ， can you know my words?  

```
private void initiateConnect(Node node, long now) {
        String nodeConnectionId = node.idString();
        try {
            connectionStates.connecting(nodeConnectionId, now, node.host(), clientDnsLookup);
            InetAddress address = connectionStates.currentAddress(nodeConnectionId);
            log.debug("Initiating connection to node {} using address {}", node, address);
            selector.connect(nodeConnectionId,
                    new InetSocketAddress(address, node.port()),
                    this.socketSendBuffer,
                    this.socketReceiveBuffer);
        } catch (IOException e) {
            log.warn("Error connecting to node {}", node, e);
            // Attempt failed, we'll try again after the backoff
            connectionStates.disconnected(nodeConnectionId, now);
            // Notify metadata updater of the connection failure
            metadataUpdater.handleServerDisconnect(now, nodeConnectionId, Optional.empty());
        }
    }
```

PR #9903, opened 2021-01-15
KAFKA-12204; Implement DescribeCluster API in the broker

dajac:
This PR implements the DescribeCluster API in the broker.


ijuma:
Thanks for the PR. We should have some unit tests for the protocol serdes, right?

dajac:
@ijuma Good point. Added them.

clients/src/main/resources/common/message/DescribeClusterResponse.json
```diff
@@ -0,0 +1,47 @@
...
+  "validVersions": "0",
+  "flexibleVersions": "0+",
+  "fields": [
+    { "name": "ThrottleTimeMs", "type": "int32", "versions": "0+",
+      "about": "The duration in milliseconds for which the request was throttled due to a quota violation, or zero if the request did not violate any quota." },
+    { "name": "ErrorCode", "type": "int16", "versions": "0+",
+      "about": "The top-level error code, or 0 if there was no error" },
+    { "name": "ErrorMessage", "type": "string", "versions": "0+", "nullableVersions": "0+", "default": "null",
+      "about": "The top-level error message, or null if there was no error." },
+    { "name": "ClusterId", "type": "string", "nullableVersions": "0+", "versions": "0+", "default": "null",
```
rajinisivaram: Does this need to be nullable or have a default?

clients/src/main/resources/common/message/DescribeClusterResponse.json
```diff
@@ -0,0 +1,47 @@
...
+      "about": "The cluster ID that responding broker belongs to." },
+    { "name": "ControllerId", "type": "int32", "versions": "0+", "default": "-1", "entityType": "brokerId",
+      "about": "The ID of the controller broker." },
+    { "name": "Brokers", "type": "[]DescribeClusterBroker", "versions": "0+",
+      "about": "Each broker in the response.", "fields": [
+      { "name": "NodeId", "type": "int32", "versions": "0+", "mapKey": true, "entityType": "brokerId",
```
rajinisivaram: BrokerId instead of NodeId? We seem to be using both elsewhere, but since it is in `Brokers`, `BrokerId` seems appropriate? Doesn't matter much either way I guess since this is internal.

rajinisivaram (commented):
@dajac Thanks for the PR, looks good. Left a couple of minor comments. We don't yet have an authorization test for returning authorized operations, like those in `DescribeAuthorizedOperationsTest` or `AuthorizerIntegrationTest`. Admin client changes will ensure the existing tests cover Metadata or DescribeCluster, but not both, so may be good to add an authorization test in this PR.

dajac:
@rajinisivaram Thanks for your review. I have addressed your points and I have added few unit tests to verify the cluster authorized operations as you suggested. Could you have another look at it?

rajinisivaram (approved):
@dajac Thanks for the updates, LGTM

clients/src/main/resources/common/message/DescribeClusterResponse.json
twmb: Does it? Does MetadataResponse? Is there a reason Metadata does and this doesn't?

dajac: `ClusterId` is always filled in the `DescribeClusterResponse` so it does not have to be `nullable`. In the `MetadataResponse` case, `ClusterId` is not supported by all the versions so having `null` as default helps to differentiate the cases where it is provided from the cases where it is not.

PR #9906, opened 2021-01-15
KAFKA-10885 Refactor MemoryRecordsBuilderTest/MemoryRecordsTest to avoid a lot of…

g1geordie:
Refactor MemoryRecordsBuilderTest/MemoryRecordsTest to avoid a lot of  ignored test cases

I add different Parameter Source to prevent assume method.

if method uses the source only itself ,  I use the `MethodSource`  with default name (same as test name)
The others I specify the provider as before


g1geordie:
There are no ignore test in those tests . 
@chia7712  can you check whether it satisfies your expectation or not ?

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -72,19 +70,28 @@ public String toString() {
...
     }
 
     private final Time time = Time.SYSTEM;
 
     @ParameterizedTest
-    @ArgumentsSource(MemoryRecordsBuilderArgumentsProvider.class)
+    @ArgumentsSource(NotZstd.class)
     public void testWriteEmptyRecordSet(Args args) {
         byte magic = RecordBatch.MAGIC_VALUE_V0;
-        assumeAtLeastV2OrNotZstd(magic, args.compressionType);
```
ijuma, line 86: It previously tested zstd with v2, but now it doesn't. Right?

g1geordie: @ijuma  hello
The assume condition is `compressionType != CompressionType.ZSTD || magic >= MAGIC_VALUE_V2`
We specify the magic code ` byte magic = RecordBatch.MAGIC_VALUE_V0` which is less than `MAGIC_VALUE_V2` .
then  `magic >= MAGIC_VALUE_V2` is always false .

So I think the condition is `compressionType != CompressionType.ZSTD  `

ijuma: Ah, yes, the magic is hardcoded here.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -201,38 +218,37 @@ public void testHasRoomForMethod(Args args) {
...
+        return getMemoryRecordsArguments((magic, type) -> type == CompressionType.NONE || type == CompressionType.LZ4);
     }
 
     /**
      * This test verifies that the checksum returned for various versions matches hardcoded values to catch unintentional
      * changes to how the checksum is computed.
      */
     @ParameterizedTest
-    @ArgumentsSource(MemoryRecordsArgumentsProvider.class)
+    @MethodSource
```
chia7712: I prefer to add explicit parameter name. Otherwise, the method offering parameters looks like a unused method.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -390,150 +449,96 @@ protected boolean shouldRetainRecord(RecordBatch recordBatch, Record record) {
...
-            assertEquals(1, deserializedMessage.voters().size());
-            assertEquals(voterId, deserializedMessage.voters().get(0).voterId());
-        }
+    private static Stream<Arguments> testFilterToBatchDiscard() {
+        return getMemoryRecordsArguments((magic, type) -> !asList(CompressionType.NONE, CompressionType.ZSTD).contains(type) || magic >= MAGIC_VALUE_V2);
     }
 
     @ParameterizedTest
-    @ArgumentsSource(MemoryRecordsArgumentsProvider.class)
+    @MethodSource
```
chia7712: ditto

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -81,26 +82,43 @@ public String toString() {
         }
     }
 
+    private static Stream<Arguments> getMemoryRecordsArguments(BiPredicate<Byte, CompressionType> accept) {
```
chia7712: maybe just me. ```getMemoryRecordsArguments``` is too verbose. How about ```allArguments```?

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -81,26 +82,43 @@ public String toString() {
...
+
+    private static class AtLeastV2OrNotZstd implements ArgumentsProvider {
+        @Override
+        public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+            return getMemoryRecordsArguments((magic, type) -> type != CompressionType.ZSTD || magic >= MAGIC_VALUE_V2);
         }
     }
 
     private final long logAppendTime = System.currentTimeMillis();
     private final int partitionLeaderEpoch = 998;
```
chia7712: this variable is used by ```testIterator``` only so we can make it be local variable.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -261,105 +277,149 @@ public void testChecksum(Args args) {
...
-                assertEquals(TimestampType.CREATE_TIME, batch.timestampType());
-                assertEquals(baseOffset, batch.baseOffset());
-                assertEquals(baseOffset + 1, batch.lastOffset());
-                assertEquals(baseSequence, batch.baseSequence());
-                assertEquals(baseSequence + 1, batch.lastSequence());
-                assertEquals(isTransactional, batch.isTransactional());
-            }
-        }
+    @ArgumentsSource(MagicAtLeastV2.class)
+    public void testEmptyBatchRetention(Args args) {
```
chia7712: This is not related to your PR but this test case seems not to require parameters.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -261,105 +277,149 @@ public void testChecksum(Args args) {
...
+        MemoryRecords filteredRecords = MemoryRecords.readableRecords(filtered);
+        assertEquals(DefaultRecordBatch.RECORD_BATCH_OVERHEAD, filteredRecords.sizeInBytes());
     }
 
     @ParameterizedTest
-    @ArgumentsSource(MemoryRecordsArgumentsProvider.class)
-    public void testEmptyBatchRetention(Args args) {
-        if (args.magic >= RecordBatch.MAGIC_VALUE_V2) {
+    @ArgumentsSource(MagicAtLeastV2.class)
+    public void testEmptyBatchDeletion(Args args) {
```
chia7712: ditto

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -390,150 +449,96 @@ protected boolean shouldRetainRecord(RecordBatch recordBatch, Record record) {
...
-                assertEquals(0, filteredRecords.sizeInBytes());
-            }
+            assertEquals(0, filteredRecords.sizeInBytes());
         }
     }
 
     @ParameterizedTest
-    @ArgumentsSource(MemoryRecordsArgumentsProvider.class)
+    @ArgumentsSource(MagicAtLeastV2.class)
     public void testBuildEndTxnMarker(Args args) {
```
chia7712: ditto

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -390,150 +449,96 @@ protected boolean shouldRetainRecord(RecordBatch recordBatch, Record record) {
...
+        assertTrue(record.isValid());
+        EndTransactionMarker deserializedMarker = EndTransactionMarker.deserialize(record);
+        assertEquals(ControlRecordType.COMMIT, deserializedMarker.controlType());
+        assertEquals(coordinatorEpoch, deserializedMarker.coordinatorEpoch());
     }
 
     @ParameterizedTest
-    @ArgumentsSource(MemoryRecordsArgumentsProvider.class)
+    @ArgumentsSource(MagicAtLeastV2.class)
     public void testBuildLeaderChangeMessage(Args args) {
```
chia7712: ditto

chia7712 (commented):
@g1geordie Thanks for this nice patch. left some comments. please take a look.

g1geordie:
@chia7712  thanks for all suggestions

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -72,19 +70,28 @@ public String toString() {
...
+                    if (accept(compressionType))
+                        values.add(Arguments.of(new Args(bufferOffset, compressionType)));
             return values.stream();
         }
+        protected boolean accept(CompressionType type) {
+            return true;
+        }
+    }
+
+    private static class NotZstd extends MemoryRecordsBuilderArgumentsProvider {
```
ijuma: I think `NotZstd` is not the best way to capture this. Basically we are excluding the compression types based on the magic version. Is that a better way to describe it?

ijuma: Also, not sure we want to do this inheritance based approach. How about we have the Stream returning part as a method and then you just do the filter on the Stream?

chia7712:
@g1geordie Personally, current approach is a bit hard to read to me. It seems to me the "parameterized" arguments should be simple and readable. We should be able to understand what is injected by reading the "value" of ```ArgumentsSource```.

For example, ```MemoryRecordsBuilderTest``` needs two kind of "parameters":
1) exclude zstd
    - testWriteEmptyRecordSet
    - testCompressionRateV0
    - testCompressionRateV1
    - buildUsingLogAppendTime
    - buildUsingCreateTime
    - writePastLimit
    - convertToV1WithMixedV0AndV2Data
    - shouldThrowIllegalStateExceptionOnBuildWhenAborted
    - shouldResetBufferToInitialPositionOnAbort
    - shouldThrowIllegalStateExceptionOnCloseWhenAborted
    - shouldThrowIllegalStateExceptionOnAppendWhenAborted
2) other test cases can run with all parameters

Hence, we can add a new provider to offer parameters excluding zstd and then remove the ```assumeTrue``` from ```MemoryRecordsBuilderTest```. WDYT?

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
chia7712: @g1geordie It seems you changed the approach according to above comments. I'm +1 to the new provider "NotZstd" as it is a simple solution. In order to make it readable, we can add some comment for "NotZstd" (why we need this provider). Also, the inheritance is unnecessary since it can be replaced by filtered Stream (from a private static method which offer all parameters)

@ijuma WDYT?

chia7712: BTW, ```testAppendedChecksumConsistency``` need to test with zstd. The "magic array" should be changed to magic_2 when input is zstd.

ijuma: What we have here is not exactly what I had intended. I have to take a closer look, but just wanted to share that.

g1geordie: Let me explain 

The annotation `@RecordSource` supply valid `( compress , magic )` mapping for all magic .  
If minMaigc or maxMaigc is set , it will retrun valid `(compress , magic in range)` and symbolize the test  test in what version more clearly .
 
I don't know this style for both of you is more meaningful or complex.

chia7712: However, the ```@RecordBuilderSource(haveInvalidCompress = true)``` is a inscrutable to me :(

g1geordie: @chia7712  
Sorry about my bad english .

In some test  , we want to test unsupported compress .  like  `(compress = ZSTD, magic = V1 ) `
`haveInvalidCompress  = true  `, will include those non valid pair .

chia7712: My point was ```haveInvalidCompress``` is hard to understand which compression is invalid. In current test cases, only zstd is possible to be excluded, right? If so, the name like "NotZstd" is more readable to me.

g1geordie:
@chia7712  sorry for late reply . 
I revert to` NotZstd` format  . Can you help me take a look

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -423,30 +455,32 @@ public void buildUsingCreateTime(Args args) {
     }
 
     @ParameterizedTest
-    @RecordBuilderSource
+    @ArgumentsSource(NotZstd.class)
     public void testAppendedChecksumConsistency(Args args) {
```
chia7712, line 434: this test case should test zstd + magic_2. We can changes the magic code according to compression type.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -27,112 +27,81 @@
...
+    private static class MemoryRecordsBuilderArgumentsProvider implements ArgumentsProvider {
         @Override
         public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
-            Predicate<Byte> predicate = magic -> recordSource.minMagic() <= magic && magic <= recordSource.maxMagic();
-            return allArguments((magic, type) -> (recordSource.haveInvalidCompress() || validCompress.test(magic, type)) && predicate.test(magic));
+            return allArguments(type -> true);
         }
+    }
 
+    private static class NotZstd implements ArgumentsProvider {
```
chia7712: Please add comment for this suite.

chia7712 (commented):
@g1geordie Thanks for your patch.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
g1geordie: also test it

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
g1geordie: have added

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
g1geordie: haha . Although I have reverted .
 
`haveInvalidCompress= false`   => (ZSTD , 2)  ...other
`haveInvalidCompress= true`   => (ZSTD , 0) , (ZSTD , 1) ,(ZSTD , 2) ...other

`NotZstd` =>  ...other

Little diff . Never mind me.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -66,25 +65,38 @@ public String toString() {
...
-                for (CompressionType compressionType : CompressionType.values())
-                    values.add(Arguments.of(new Args(bufferOffset, compressionType)));
-            return values.stream();
+            return allArguments(type -> true);
+        }
+    }
+
+    /**
+     *   CompressionType.ZSTD only support when magic >= RecordBatch.MAGIC_VALUE_V2
+     */
```
g1geordie: @chia7712   Is it the comment that  you want to add?

chia7712: How about "ZSTD is not supported for MAGIC_VALUE_V0 and MAGIC_VALUE_V1"?

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -455,20 +460,19 @@ public void buildUsingCreateTime(Args args) {
     @ParameterizedTest
     @ArgumentsSource(MemoryRecordsBuilderArgumentsProvider.class)
     public void testAppendedChecksumConsistency(Args args) {
-        assumeAtLeastV2OrNotZstd(RecordBatch.MAGIC_VALUE_V0, args.compressionType);
-        assumeAtLeastV2OrNotZstd(RecordBatch.MAGIC_VALUE_V1, args.compressionType);
-
         ByteBuffer buffer = ByteBuffer.allocate(512);
         for (byte magic : Arrays.asList(RecordBatch.MAGIC_VALUE_V0, RecordBatch.MAGIC_VALUE_V1, RecordBatch.MAGIC_VALUE_V2)) {
```
chia7712: Could we add a local variable to includes the "valid" magic code. For example:
```java
List<Byte> magicCodes = args.compressionType == CompressionType.ZSTD
  ? Collections.singletonList(RecordBatch.MAGIC_VALUE_V2)
  : Arrays.asList(RecordBatch.MAGIC_VALUE_V0, RecordBatch.MAGIC_VALUE_V1, RecordBatch.MAGIC_VALUE_V2)
  for (byte magic : magicCodes) {
    // ...
  }

```

It seems to me that is more readable. WDYT?

chia7712:
@g1geordie Could you merge trunk to include the recent bug fixes?

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
g1geordie: sure. It's more readable .

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -1004,10 +998,6 @@ public void testWithRecords(Args args) {
         }
```
chia7712: Is ```AtLeastV2OrNotZstd``` more suitable to ```testWithRecords```?

chia7712 (commented):
@g1geordie Thanks for your patch. one last comment. Please take a look.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
g1geordie: @chia7712 In my opinion , It checks the IllegalArgumentException( unsupport compress).
So I think it is necessary

chia7712: It is another test style - we can pass all parameters to all test cases and then explicitly call assertThrow to handle the unsupported method(s). This style can bring some benefits: 

1. Single parameter suite
2. explicitly show the “unsupported” method

WDYT?

g1geordie: @chia7712  It seems like before . 
But I think that checking all `new  MemoryRecordsBuilder ` will throw by some args is not easy.

what do you think

chia7712:
For another, ```testWriteControlBatchNotAllowedMagicV0``` and ```testWriteControlBatchNotAllowedMagicV1``` are almost same. Could we merge them into single test case?

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
chia7712: @g1geordie I file a patch for aforementioned idea. Please take a look at https://github.com/chia7712/kafka/pull/1/files

it uses explicit assert (exception or expected value) for all parameters instead of just ignoring them

g1geordie: It sound like you change `assume (condition)` to  `if (condition) ... else ...`  in all method

g1geordie:
@chia7712  
Thank for your patch .
I follow the style and change `MemoryRecordsBuilderTest`

Can you help me take a look? :)

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsTest.java
```diff
@@ -995,17 +988,20 @@ public void testWithRecords(Args args) {
         byte magic = args.magic;
         Supplier<MemoryRecords> recordsSupplier = () -> MemoryRecords.withRecords(magic, compression,
```
chia7712: This supplier is not necessary as we don't assert the exception anymore.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -71,35 +80,46 @@ public String toString() {
...
     @ParameterizedTest
     @ArgumentsSource(MemoryRecordsBuilderArgumentsProvider.class)
     public void testWriteEmptyRecordSet(Args args) {
-        byte magic = RecordBatch.MAGIC_VALUE_V0;
-        assumeAtLeastV2OrNotZstd(magic, args.compressionType);
-
+        byte magic = args.magic;
         ByteBuffer buffer = allocateBuffer(128, args);
 
         Supplier<MemoryRecordsBuilder> builderSupplier = () -> new MemoryRecordsBuilder(buffer, magic,
```
chia7712: unnecessary supplier.

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -352,124 +364,87 @@ public void testEstimatedSizeInBytes(Args args) {
...
-        MemoryRecords records = builder.build();
-
-        MemoryRecordsBuilder.RecordsInfo info = builder.info();
-        assertEquals(2L, info.maxTimestamp);
-
-        if (args.compressionType == CompressionType.NONE)
-            assertEquals(1L, info.shallowOffsetOfMaxTimestamp);
-        else
-            assertEquals(2L, info.shallowOffsetOfMaxTimestamp);
+        if (args.magic >= MAGIC_VALUE_V1) {
```
chia7712: Could we apply same pattern for this test case?

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -124,207 +144,199 @@ public void testWriteTransactionalRecordSet(Args args) {
...
+                    args.compressionType, TimestampType.CREATE_TIME, 0L, 0L, pid, epoch, sequence,
+                    false, true, RecordBatch.NO_PARTITION_LEADER_EPOCH, buffer.capacity()));
+        }
     }
 
     @ParameterizedTest
     @ArgumentsSource(MemoryRecordsBuilderArgumentsProvider.class)
     public void testWriteTransactionalWithInvalidPID(Args args) {
-        ByteBuffer buffer = allocateBuffer(128, args);
+        if (args.magic >= MAGIC_VALUE_V2) {
```
chia7712: Is this check necessary?

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -124,207 +144,199 @@ public void testWriteTransactionalRecordSet(Args args) {
...
+            assertThrows(IllegalArgumentException.class, () -> builder.appendEndTxnMarker(RecordBatch.NO_TIMESTAMP,
+                    new EndTransactionMarker(ControlRecordType.ABORT, 0)));
+        }
     }
 
     @ParameterizedTest
     @ArgumentsSource(MemoryRecordsBuilderArgumentsProvider.class)
     public void testWriteEndTxnMarkerNonControlBatch(Args args) {
-        ByteBuffer buffer = allocateBuffer(128, args);
+        if (args.magic >= MAGIC_VALUE_V2) {
```
chia7712: How about replacing it by ```assertThrow```?

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -124,207 +144,199 @@ public void testWriteTransactionalRecordSet(Args args) {
...
-        final int leaderId = 1;
-        MemoryRecordsBuilder builder = new MemoryRecordsBuilder(buffer, RecordBatch.CURRENT_MAGIC_VALUE, args.compressionType, TimestampType.CREATE_TIME,
-            0L, 0L,
-            RecordBatch.NO_PRODUCER_ID, RecordBatch.NO_PRODUCER_EPOCH, RecordBatch.NO_SEQUENCE,
-            false, true, RecordBatch.NO_PARTITION_LEADER_EPOCH, buffer.capacity());
-        assertThrows(IllegalArgumentException.class, () -> builder.appendLeaderChangeMessage(RecordBatch.NO_TIMESTAMP,
-            new LeaderChangeMessage()
-                .setLeaderId(leaderId)
-                .setVoters(Collections.emptyList())));
+        if (args.magic >= MAGIC_VALUE_V2) {
```
chia7712: ditto

chia7712 (commented):
@g1geordie Thanks for your patch. The new pattern is LGTM. Could you apply it to all test cases?

g1geordie:
@chia7712  I have filled those `else`  with `assertthrow`.  What do you think ?

chia7712 (approved):

@ijuma I'd like to merge it tomorrow and it would be helpful if you could take a look :)

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -22,47 +22,56 @@
...
             this.bufferOffset = bufferOffset;
             this.compressionType = compressionType;
+            this.magic = magic;
         }
 
         @Override
         public String toString() {
-            return "bufferOffset=" + bufferOffset +
-                ", compressionType=" + compressionType;
+            return "Args{" +
```
ijuma: Why do we have `Args` here? The `toString` is used in the JUnit test display and hence why it doesn't include anything besides the parameters before this change.

g1geordie: I will remove it .

clients/src/test/java/org/apache/kafka/common/record/MemoryRecordsBuilderTest.java
```diff
@@ -778,7 +737,7 @@ public void testBuffersDereferencedOnClose(Args args) {
         int iterations =  0;
         while (iterations++ < 100) {
             buffer.rewind();
-            MemoryRecordsBuilder builder = new MemoryRecordsBuilder(buffer, RecordBatch.MAGIC_VALUE_V2, args.compressionType,
+            MemoryRecordsBuilder builder = new MemoryRecordsBuilder(buffer, args.magic, args.compressionType,
                     TimestampType.CREATE_TIME, 0L, 0L, RecordBatch.NO_PRODUCER_ID,
                     RecordBatch.NO_PRODUCER_EPOCH, RecordBatch.NO_SEQUENCE, false, false,
                     RecordBatch.NO_PARTITION_LEADER_EPOCH, 0);
```
chia7712, line 743: This is unrelated to this PR. However, could you replace `new byte[0]` by `key` (line#744)?

g1geordie: @chia7712  @ijuma  I have fixed  it .  Thanks for both of you to review.

chia7712:
@g1geordie Thanks for your updating. I will merge it tomorrow :)

PR #9910, opened 2021-01-15
KAFKA-10877

smccauliff:
This moves the Logging trait into companion objects so the loggers are only instantiated once.  Classes import the methods from their companion objects.  Classes that were using inheritance to gain access to logging methods are now accessing them though importing their super class' companion object.  In this way any log4j(2) configuration that were configuring the super class logger should still work.

Since classes no longer inherit from Logging they can't change the logIdent string through inheritance.  This PR introduces a new class LogIndent which is used to wrap String so it can be passed as an implicit variable to calls to info(), debug(), etc...  So none of the actual log statements need to change.

CoreUtils.swallow is changed so that it accepts the LogIdent objects.  It seems like CoreUtils.swallow never used to logger that was passed to it only it's own logger which seems like bug.  Calls to swallow need pass the companion object and not this.

There are 10 spot bugs problems flagged.  But I'm not able to figure out what spotbugs is complaining about.

### DIFF
--- a/core/src/main/scala/kafka/admin/AclCommand.scala
+++ b/core/src/main/scala/kafka/admin/AclCommand.scala
@@ -86,12 +86,7 @@ object AclCommand extends Logging {
     def listAcls(): Unit
   }
 
-  object AdminClientService extends Logging {
-
-  }
-
   class AdminClientService(val opts: AclCommandOptions) extends AclCommandService {
-    import AdminClientService._
 
     private def withAdminClient(opts: AclCommandOptions)(f: Admin => Unit): Unit = {
       val props = if (opts.options.has(opts.commandConfigOpt))
@@ -198,7 +193,6 @@ object AclCommand extends Logging {
 
   }
   class AuthorizerService(val authorizerClassName: String, val opts: AclCommandOptions) extends AclCommandService {
-    import AuthorizerService._
 
     private def withAuthorizer()(f: Authorizer => Unit): Unit = {
       // It is possible that zookeeper.set.acl could be true without SASL if mutual certificate authentication is configured.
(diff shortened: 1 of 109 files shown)


smccauliff:
@gitlw This is the more comprehensive change I was discussing with you.

core/src/main/scala/kafka/admin/AclCommand.scala
```diff
@@ -189,7 +189,10 @@ object AclCommand extends Logging {
     }
   }
 
-  class AuthorizerService(val authorizerClassName: String, val opts: AclCommandOptions) extends AclCommandService with Logging {
+  object AuthorizerService extends Logging {
```
ijuma, line 192: If you do it like this, it changes the logger name. You probably want to have a val that sets the logger name to match the actual class versus the companion.

ijuma: Also, for long-running services, not sure this buys us much. I think this is most helpful for instances that get allocated a reasonable amount. Can we focus on those perhaps?

smccauliff: For 1.  Doesn't that create a circular dependency between the class and the companion object?
For 2. Is it a good idea to leave a dangerous anti-pattern lying around in the code base for others to cut-n-paste"

ijuma: 1. It does, but companions and class have a tight relationship. Is it a problem?
2. I think if we want to do more than fix the bug, then we should evaluate what is a good pattern. And I don't think the `Logging` trait is a good pattern. :)

gharris1727:
@smccauliff Are you still interested in making this change?
I think this is still affecting users and wasting CPU cycles :)

PR #9912, opened 2021-01-15
MINOR: Move `RequestChannel.Response` creation logic into `RequestChannel`

hachikuji:
This patch moves some common response creation logic from `RequestHandlerHelper` and into `RequestChannel`. This refactor has the following benefits:

- It allows us to get rid of some logic that was previously duplicated in both `RequestHandlerHelper` and `TestRaftRequestHandler`. 
- It ensures that we do not need to rely on the caller to ensure that `updateErrorMetrics` gets called since this is handled internally in `RequestChannel`.
- It provides better encapsulation of the quota workflow which relies on custom `Response` objects. Previously it was quite confusing for `KafkaApis` to handle this directly through the `sendResponse` API.


core/src/main/scala/kafka/server/RequestHandlerHelper.scala
```diff
@@ -58,16 +57,68 @@ object RequestHandlerHelper {
...
+  this.logIdent = logPrefix
 
+  def handleError(request: RequestChannel.Request, e: Throwable): Unit = {
+    error("Error when handling request: " +
+      s"clientId=${request.header.clientId}, " +
+      s"correlationId=${request.header.correlationId}, " +
+      s"api=${request.header.apiKey}, " +
+      s"version=${request.header.apiVersion}, " +
+      s"body=${request.body[AbstractRequest]}", e)
+    sendErrorOrCloseConnection(request, e, 0)
```
ijuma: Did we consider having the throttle calls (`maybeRecordExempt`, `throttle`, `maybeRecordAndGetThrottleTimeMs`, etc.) split into a separate interface and then have an unthrottled implementation of that?

hachikuji: The approach I took here was intended as a lighter touch. Many of the APIs exposed by `RequestHandlerHelper` set throttling expectations explicitly (e.g. `sendResponseExemptThrottle`). That makes it a little awkward to push down the throttling logic. If we only depended on the request quota, then I would be tempted to try, but we also depend on the controller mutation quota. I did consider changing the dependence from `QuotaManagers` to `ClientRequestQuotaManager` and `ControllerMutationQuotaManager`. Then we could pull out a trait for both of these. But all of this seemed like more work with little benefit.

ijuma: Thanks for the explanation. The `handleError` duplication seems a bit ugly. Also, inheriting from concrete classes is generally not great. So, to me, the current approach is lighter touch, but doesn't feel right.

hachikuji: There is no state inside `UnthrottledRequestHandlerHelper`, so we could pull these methods into the companion I guess. Not sure that's really better, but at least it removes the inheritance. Let me see how this looks.

hachikuji:
So what I ended up doing is moving the helpers that `TestRaftRequestHandler` needed into `RequestChannel`. Basically this encapsulates the construction of the `RequestChannel.Response`, which seems a bit nicer since it was confusing to see `sendResponse` used for throttling events. It is also nice because it ensures that we do not need to rely on the caller to ensure that `updateErrorMetrics` gets called.

clients/src/main/java/org/apache/kafka/common/requests/RequestContext.java
```diff
@@ -175,4 +175,19 @@ public String clientId() {
...
+        return "RequestContext(" +
+            "header=" + header +
+            ", connectionId='" + connectionId + '\'' +
+            ", clientAddress=" + clientAddress +
+            ", principal=" + principal +
+            ", listenerName=" + listenerName +
+            ", securityProtocol=" + securityProtocol +
+            ", clientInformation=" + clientInformation +
+            ", fromPrivilegedListener=" + fromPrivilegedListener +
+            ", principalSerde=" + principalSerde +
```
ijuma, line 190: Have we checked that no privileged information (i.e. credentials) are logged by any of these fields? Seems OK to me, but worth double checking.

core/src/test/scala/unit/kafka/server/ThrottledChannelExpirationTest.scala
```diff
@@ -44,28 +33,13 @@ class ThrottledChannelExpirationTest {
...
-      new RequestHeader(builder.apiKey, request.version, "", 0), request)
-    val requestChannelMetrics: RequestChannel.Metrics = EasyMock.createNiceMock(classOf[RequestChannel.Metrics])
-
-    // read the header from the buffer first so that the body can be read next from the Request constructor
-    val header = RequestHeader.parse(buffer)
-    val context = new RequestContext(header, "1", InetAddress.getLocalHost, KafkaPrincipal.ANONYMOUS,
-      listenerName, SecurityProtocol.PLAINTEXT, ClientInformation.EMPTY, false)
-    (request, new RequestChannel.Request(processor = 1, context = context, startTimeNanos =  0, MemoryPool.NONE, buffer,
-      requestChannelMetrics))
-  }
```
ijuma, line 63: Nice that we don't need this anymore.

core/src/main/scala/kafka/network/RequestChannel.scala
```diff
@@ -371,9 +372,44 @@ class RequestChannel(val queueSize: Int,
...
+    // This case is used when the request handler has encountered an error, but the client
+    // does not expect a response (e.g. when produce request has acks set to 0)
+    updateErrorMetrics(request.header.apiKey, errorCounts.asScala)
+    sendResponse(new RequestChannel.CloseConnectionResponse(request))
+  }
+
+  def sendResponse(
+    request: RequestChannel.Request,
+    response: AbstractResponse,
+    onComplete: Option[Send => Unit]
```
ijuma, line 389: I wonder if this should be `Send => Unit` instead. It doesn't seem like there is much value in the `Option` here. A function that does nothing is equivalent to `None`. I know this is an existing issue, so ok if you prefer not to address now.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -227,37 +225,33 @@ class KafkaApisTest {
...
+    val capturedResponse = EasyMock.newCapture[AbstractResponse]()
+    val capturedRequest = EasyMock.newCapture[RequestChannel.Request]()
 
-    createKafkaApis(authorizer = Some(authorizer), enableForwarding = true).handle(request)
-
-    val envelopeRequest = request.body[EnvelopeRequest]
-    val response = readResponse(envelopeRequest, capturedResponse)
-      .asInstanceOf[EnvelopeResponse]
-
-    assertEquals(Errors.NONE, response.error)
```
ijuma, line 317: We don't need this assert anymore? Can you explain a bit more what we're trying to do with the changes in this method? That will probably help with some other envelope related test changes in this class too.

ijuma (commented):
Thanks for the update, this is looking pretty good!

One thing I didn't fully understand: the previous version of the patch had an unthrottled variant. The latest version extracts the throttling, but I didn't see where we are actually skipping the throttling. Can you please point me to that?

ijuma:
Also, looks like there is a compiler error:

> 02:22:44  [Error] /home/jenkins/jenkins-agent/workspace/Kafka_kafka-pr_PR-9912/core/src/main/scala/kafka/server/KafkaApis.scala:148: value closeConnection is not a member of kafka.server.RequestHandlerHelper

clients/src/main/java/org/apache/kafka/common/requests/RequestContext.java
hachikuji: Yeah, I considered it, but didn't see anything in here that looked sensitive. Most of this information is already included in the request logging. The client address is not, but that seems ok. Including the principal serde is annoying, but doesn't seem like a problem.

core/src/main/scala/kafka/network/RequestChannel.scala
hachikuji: That's fair. I will check the scope of this change.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
hachikuji: There remains some awkwardness in the handling of envelope requests. Basically the flow is like this:

1. KafkaApis.handle(envelope(alterConfigRequest))
2. KafkaApis.handle(alterConfigRequest)
3. RequestChannel.sendResponse(alterConfigResponse)
4. Request.buildResponseSend() -> envelope(alterConfigResponse)

Previously `KafkaApisTest` had to work by parsing the response send, so we had to unwrap the envelope. But now we get the direct call to `sendResponse` with the embedded `AbstractResponse` instance. So basically we do not hit step 4 anymore.

I think a nicer way to structure this which we can consider separately is to change `KafkaApis.handle` so that it takes a callback rather than assuming that responses are sent directly to the request channel. Then when we make the recursive call to `handle` after unwrapping the envelope, we can provide a callback which wraps the underlying response with the response envelope. Alternatively, we can have `KafkaApis.handle` return a `CompletableFuture`. The main point is that we allow for some custom behavior when the response is ready.

core/src/main/scala/kafka/network/RequestChannel.scala
hachikuji: I decided not to do this here. I didn't like replacing `None` with `_ => {}` in uses, and neither did I like making the argument optional. The alternative is to introduce a constant "no-op" function, but I found this also a little awkward. If you think of a nice way to do it, I can review. I do think it is better having a simpler type.

ijuma: Sounds good.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
ijuma: Thanks for the explanation.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -529,15 +518,14 @@ class KafkaApisTest {
     new IncrementalAlterConfigsRequest.Builder(resourceMap, false)
   }
 
+  // FIXME: Unused arg
```
ijuma: Does this still have to be addressed? Either way, the FIXME needs to be removed.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -579,11 +566,11 @@ class KafkaApisTest {
     testForwardableAPI(ApiKeys.ALTER_CLIENT_QUOTAS, requestBuilder)
   }
 
+  // FIXME: Unused arg
```
ijuma: Similar to the other comment.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -2127,21 +2118,15 @@ class KafkaApisTest {
...
+    val response = capturedResponse.getValue.asInstanceOf[JoinGroupResponse]
 
     assertEquals(Errors.NONE, response.error)
     assertEquals(0, response.data.members.size)
     assertEquals(memberId, response.data.memberId)
     assertEquals(0, response.data.generationId)
     assertEquals(memberId, response.data.leader)
     assertEquals(protocolName, response.data.protocolName)
-
-    if (version >= 7) {
```
ijuma, line 2353: Why is this not needed?

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -2200,17 +2183,11 @@ class KafkaApisTest {
...
     ))
 
-    val response = readResponse(syncGroupRequest, capturedResponse)
-      .asInstanceOf[SyncGroupResponse]
+    val response = capturedResponse.getValue.asInstanceOf[SyncGroupResponse]
 
     assertEquals(Errors.NONE, response.error)
     assertArrayEquals(Array.empty[Byte], response.data.assignment)
-
-    if (version >= 5) {
```
ijuma, line 2422: Similar question, why is this not needed?

ijuma (commented):
Looks good overall, a couple of FIXMEs to resolve and a couple of questions.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
hachikuji: Now these test cases are working with the raw response object returned by `KafkaApis`. This means we don't have to deal with fields which are dropped during serialization.

ijuma (approved):
LGTM, thanks.

PR #9913, opened 2021-01-15
MINOR: Move a few more methods to AuthHelper

ijuma:
Also move some tests to `AuthHelperTest`.

### DIFF
--- a/core/src/main/scala/kafka/server/AuthHelper.scala
+++ b/core/src/main/scala/kafka/server/AuthHelper.scala
@@ -19,9 +19,9 @@ package kafka.server
 
 import java.lang.{Byte => JByte}
 import java.util.Collections
-
 import kafka.network.RequestChannel
 import kafka.security.authorizer.AclEntry
+import kafka.utils.CoreUtils
 import org.apache.kafka.common.acl.AclOperation
 import org.apache.kafka.common.errors.ClusterAuthorizationException
 import org.apache.kafka.common.requests.RequestContext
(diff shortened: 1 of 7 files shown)


core/src/main/scala/kafka/server/AuthHelper.scala
```diff
@@ -70,4 +70,64 @@ class AuthHelper(authorizer: Option[Authorizer]) {
...
+      case Some(_) =>
+        val authorizedResourceNames = filterByAuthorized(requestContext, operation, resourceType,
+          resources.keySet, logIfAllowed, logIfDenied)(resourceName)
+        resources.partition { case (k, _) => authorizedResourceNames.contains(resourceName(k)) }
+      case None => (resources, Map.empty)
+    }
+  }
+
+  def filterByAuthorized[T](requestContext: RequestContext,
+                                            operation: AclOperation,
```
mumrah: nit: indentation

core/src/test/scala/unit/kafka/utils/TestUtils.scala
```diff
@@ -130,6 +131,19 @@ object TestUtils extends Logging {
     f
   }
 
+  /**
+   * Returns true if the elements in both lists are the same irrespective of ordering.
+   */
+  def matchSameElements[T](seq: Seq[T]): Seq[T] = {
+    EasyMock.reportMatcher(new IArgumentMatcher {
```
mumrah: I know this is just moving existing code, but is the comment here accurate?  Looks like this is some specific EasyMock thing rather than a simple assertion on Seq elements which the scaladoc suggests

mumrah (approved):
LGTM, few minor comments

core/src/test/scala/unit/kafka/utils/TestUtils.scala
ijuma: Good point. The documentation is confusing. It looks like I also broke some tests when I tweaked this method. Let me fix that.

ijuma:
@mumrah Pushed an update. I reverted the `swallow` change as it is unrelated and it caused a few spotBugs failures due to how the scala inliner works. I will submit a separate change for that.

PR #9914, opened 2021-01-15
KAFKA-12214: Generated code does not include UUID or struct fields in toString()

cmccabe:


ijuma (approved):
LGTM

generator/src/main/java/org/apache/kafka/message/MessageDataGenerator.java
```diff
@@ -1901,7 +1901,9 @@ private void generateFieldToString(String prefix, FieldSpec field) {
             buffer.printf("+ \"%s%s=\" + %s%n",
                     prefix, field.camelCaseName(), field.camelCaseName());
         } else if (field.type().isStruct() ||
-            field.type() instanceof FieldType.UUIDFieldType) {
+                field.type() instanceof FieldType.UUIDFieldType) {
+            buffer.printf("+ \"%s%s=\" + %s.toString()%n",
+                prefix, field.camelCaseName(), field.camelCaseName());
         } else if (field.type().isStruct()) {
```
jolshan: Should this case be removed if it is covered by the case above?

cmccabe: good point.  I will consolidate these.

PR #9916, opened 2021-01-15
MINOR: Import RaftConfig parameters into KafkaConfig

aloknnikhil:
- Typically, Raft Quorum Configuration is provided through the same
properties file as the Broker/Controller a.k.a KafkaConfig
- During the lifecycle of the Broker/Controller, we do reference the
Quorum configuration outside of the RaftConfig (Eg: RaftManager.scala)


raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -37,49 +36,56 @@
     private static final String QUORUM_PREFIX = "quorum.";
 
     public static final String QUORUM_VOTERS_CONFIG = QUORUM_PREFIX + "voters";
-    private static final String QUORUM_VOTERS_DOC = "Map of id/endpoint information for " +
+    public static final String QUORUM_VOTERS_DOC = "Map of id/endpoint information for " +
         "the set of voters in a comma-separated list of `{id}@{host}:{port}` entries. " +
         "For example: `1@localhost:9092,2@localhost:9093,3@localhost:9094`";
+    public static final String QUORUM_VOTERS_DEFAULT = "";
```
ijuma: The typical approach used by other config files is to have `DEFAULT` as a prefix. See `ConsumerConfig` for example. Unless there's a reason to deviate, I'd follow the same approach.

aloknnikhil: Makes sense. No reason to keep it this way. Fixed. Thanks!

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -37,49 +36,56 @@
...
 
-    // Package-private for testing
-    static final String QUORUM_RETRY_BACKOFF_MS_CONFIG = QUORUM_PREFIX +
+    public static final String QUORUM_RETRY_BACKOFF_MS_CONFIG = QUORUM_PREFIX +
         CommonClientConfigs.RETRY_BACKOFF_MS_CONFIG;
+    public static final String QUORUM_RETRY_BACKOFF_MS_DOC = CommonClientConfigs.RETRY_BACKOFF_MS_DOC;
+    public static final int DEFAULT_QUORUM_RETRY_BACKOFF_MS = 100;
 
     static {
         CONFIG = new ConfigDef()
```
hachikuji: Do we need to keep the definitions here?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -37,49 +36,56 @@
     private static final String QUORUM_PREFIX = "quorum.";
```
hachikuji: We need the `controller` prefix that comes from KIP-631. Also if you could change raft/README to reflect the new names, that would be helpful.

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1255,6 +1282,15 @@ object KafkaConfig {
       .define(PasswordEncoderCipherAlgorithmProp, STRING, Defaults.PasswordEncoderCipherAlgorithm, LOW, PasswordEncoderCipherAlgorithmDoc)
       .define(PasswordEncoderKeyLengthProp, INT, Defaults.PasswordEncoderKeyLength, atLeast(8), LOW, PasswordEncoderKeyLengthDoc)
       .define(PasswordEncoderIterationsProp, INT, Defaults.PasswordEncoderIterations, atLeast(1024), LOW, PasswordEncoderIterationsDoc)
+
+      /** ********* Raft Quorum Configuration *********/
+      .define(QuorumVotersProp, LIST, Defaults.QuorumVoters, HIGH, QuorumVotersDoc)
```
hachikuji: I'm considering if we should define these as internal configurations for now until we have a plan for how to expose them when KIP-500 is initially released in alpha.

aloknnikhil:
> This looks unintentional.
Hmm. Any specific part of the code?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Raft cannot depend on core. So, it needs an implementation of AbstractConfig. ConfigDef here is necessary for that implementation. Is that what you meant by definitions?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Done.

core/src/main/scala/kafka/server/KafkaConfig.scala
aloknnikhil: That makes sense. I didn't know about the internal definition. Changed.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
hachikuji: I guess what I'm saying is that RaftConfig does not need to extend AbstractConfig if the definitions are in KafkaConfig. Instead, we can let the RaftConfig constructor take an AbstractConfig, which will be the reference to KafkaConfig. Does that make sense?

aloknnikhil: Right. That would be fine except that would mean the tests in the raft module would then have to depend on the core module to create a KafkaConfig out of the props. There's no circular dependency (between the raft test module and core), but that would mean we would introduce some sort of dependency on core in raft.

hachikuji: `RaftConfig` just becomes a convenient holder for config values. We can have two constructors:
```java
public RaftConfig(AbstractConfig config)
public RaftConfig(String quorumVoters, int requestTimeoutMs, ...)
```

aloknnikhil: Yea, that makes sense. I think devolving it to a util class of sorts (parsing the voter nodes/ids/connections and such) would dedupe the ConfigDef. I'll go ahead and make this change.

aloknnikhil: Fixed. @hachikuji

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -138,31 +75,35 @@ public String toString() {
...
+                abstractConfig.getInt(QUORUM_RETRY_BACKOFF_MS_CONFIG),
+                abstractConfig.getInt(QUORUM_ELECTION_TIMEOUT_MS_CONFIG),
+                abstractConfig.getInt(QUORUM_ELECTION_BACKOFF_MAX_MS_CONFIG),
+                abstractConfig.getInt(QUORUM_FETCH_TIMEOUT_MS_CONFIG),
+                abstractConfig.getInt(QUORUM_LINGER_MS_CONFIG),
+                abstractConfig.getString(QUORUM_VOTERS_CONFIG));
+    }
+
+    public RaftConfig(
+            int requestTimeoutMs,
```
hachikuji: nit: usual alignment is just 4 spaces

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1687,6 +1714,15 @@ class KafkaConfig(val props: java.util.Map[_, _], doLog: Boolean, dynamicConfigO
   val deleteTopicEnable = getBoolean(KafkaConfig.DeleteTopicEnableProp)
   def compressionType = getString(KafkaConfig.CompressionTypeProp)
 
+  /** ********* Raft Quorum Configuration *********/
+  val quorumVoters = getString(KafkaConfig.QuorumVotersProp)
```
hachikuji: Can we add a check somewhere that `quorum.voters` is required if `process.roles` is non-empty?

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -574,6 +583,15 @@ object KafkaConfig {
   val PasswordEncoderKeyLengthProp =  "password.encoder.key.length"
   val PasswordEncoderIterationsProp =  "password.encoder.iterations"
 
+  /** ********* Raft Quorum Configuration *********/
+  val QuorumVotersProp = RaftConfig.QUORUM_VOTERS_CONFIG
```
hachikuji: I know we have done so for other configs, but do we really need to duplicate these property strings?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -18,117 +18,54 @@
...
-                ConfigDef.NO_DEFAULT_VALUE,
-                new ConfigDef.Validator() {
-                    @Override
-                    public void ensureValid(String name, Object value) {
-                        if (value == null) {
-                            throw new ConfigException(name, null);
-                        }
-
-                        @SuppressWarnings("unchecked")
-                        Map<Integer, InetSocketAddress> voterConnections = parseVoterConnections((List) value);
```
hachikuji, line 94: We seem to have lost this validation. Can we just move it to `KafkaConfig` instead?

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1258,6 +1276,15 @@ object KafkaConfig {
       .define(PasswordEncoderCipherAlgorithmProp, STRING, Defaults.PasswordEncoderCipherAlgorithm, LOW, PasswordEncoderCipherAlgorithmDoc)
       .define(PasswordEncoderKeyLengthProp, INT, Defaults.PasswordEncoderKeyLength, atLeast(8), LOW, PasswordEncoderKeyLengthDoc)
       .define(PasswordEncoderIterationsProp, INT, Defaults.PasswordEncoderIterations, atLeast(1024), LOW, PasswordEncoderIterationsDoc)
+
+      /** ********* Raft Quorum Configuration *********/
+      .defineInternal(QuorumVotersProp, STRING, Defaults.QuorumVoters, HIGH)
```
hachikuji: Why don't we make the type a LIST like we had previously?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -138,31 +75,35 @@ public String toString() {
...
+    }
+
+    public RaftConfig(
+            int requestTimeoutMs,
+            int retryBackoffMs,
+            int electionTimeoutMs,
+            int electionBackoffMaxMs,
+            int fetchTimeoutMs,
+            int appendLingerMs,
+            String votersConnectString
```
hachikuji: nit: doesn't matter too much, but could we move this argument to the top to emphasize its importance?

core/src/main/scala/kafka/server/KafkaConfig.scala
aloknnikhil: Spoke offline. The ClusterTestKit on the `kip-500` branch currently requires the `quorum.voters` to be empty at startup and to be set at a later point-in-time (once the controller starts and has chosen a free port). We don't want testing to dictate the norm here. So, to allow for that one alternative is to add a `skipVoterValidation` flag in a later PR. For now, I'll move the validation from RaftConfig to KafkaConfig and skip the validation if the voter string is empty

core/src/main/scala/kafka/server/KafkaConfig.scala
aloknnikhil: Probably not. And as you said, only to keep it consistent with the other props. I can go ahead and refer to the exported final string directly

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: See https://github.com/apache/kafka/pull/9916#discussion_r560611932

core/src/main/scala/kafka/server/KafkaConfig.scala
aloknnikhil: With the validation being moved to KafkaConfig, I don't see any reason not to make it a LIST. Will change it back

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Ack.

core/src/main/scala/kafka/server/KafkaConfig.scala
hachikuji: Sounds good. Since this is an internal config at the moment, we have some time to consider a better long-term solution. My feeling is that we need a separate path for constructing the dependencies we need for testing, but I guess we can cross that bridge when we get there.

core/src/test/scala/unit/kafka/server/KafkaConfigTest.scala
```diff
@@ -941,4 +950,38 @@ class KafkaConfigTest {
...
+  def testValidQuorumVotersConfig(): Unit = {
+    assertValidQuorumVoters("", 0)
+    assertValidQuorumVoters("1@127.0.0.1:9092", 1)
+    assertValidQuorumVoters("1@kafka1:9092,2@kafka2:9092,3@kafka3:9092", 3)
+  }
+
+  private def assertValidQuorumVoters(value: String, expectedVoterCount: Int): Unit = {
+    val props = TestUtils.createBrokerConfig(0, TestUtils.MockZkConnect)
+    props.put(RaftConfig.QUORUM_VOTERS_CONFIG, value)
+    assertDoesNotThrow(() => KafkaConfig.fromProps(props))
```
hachikuji: Hmm.. `assertDoesNotThrow` seems kind of redundant. The test would fail anyway if it did throw.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -236,4 +176,31 @@ private static Integer parseVoterId(String idString) {
...
+            if (value == null) {
+                throw new ConfigException(name, null);
+            }
+
+            @SuppressWarnings("unchecked")
+            List<String> voterStrings = (List) value;
+
+            if (voterStrings.size() == 0) {
+                // TODO: Add a flag to skip validation for an empty voter string, conditionally.
+                //       For now, skip anyway. See https://github.com/apache/kafka/pull/9916#discussion_r560611932
```
hachikuji: We typically do not leave TODOs in the code. We can file a jira if we think it's important to remember. I'd suggest we just leave this check out and skip the empty check below.

raft/src/test/java/org/apache/kafka/raft/RaftConfigTest.java
```diff
@@ -1,76 +0,0 @@
...
-        Properties properties = new Properties();
-        properties.put(RaftConfig.QUORUM_VOTERS_CONFIG, "1@kafka1:9092,2@kafka2:9092,3@kafka3:9092");
-        RaftConfig config = new RaftConfig(properties);
-
-        HashMap<Integer, InetSocketAddress> expected = new HashMap<>();
-        expected.put(1, new InetSocketAddress("kafka1", 9092));
-        expected.put(2, new InetSocketAddress("kafka2", 9092));
-        expected.put(3, new InetSocketAddress("kafka3", 9092));
-
-        assertEquals(expected, config.quorumVoterConnections());
```
hachikuji, line 52: This test case seems stronger than the one that was ported to `KafkaConfigTest`. It is validating the endpoints in addition to the number of voters. Is there any way we can recover this?

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1258,6 +1267,15 @@ object KafkaConfig {
       .define(PasswordEncoderCipherAlgorithmProp, STRING, Defaults.PasswordEncoderCipherAlgorithm, LOW, PasswordEncoderCipherAlgorithmDoc)
       .define(PasswordEncoderKeyLengthProp, INT, Defaults.PasswordEncoderKeyLength, atLeast(8), LOW, PasswordEncoderKeyLengthDoc)
       .define(PasswordEncoderIterationsProp, INT, Defaults.PasswordEncoderIterations, atLeast(1024), LOW, PasswordEncoderIterationsDoc)
+
+      /** ********* Raft Quorum Configuration *********/
+      .defineInternal(RaftConfig.QUORUM_VOTERS_CONFIG, LIST, Defaults.QuorumVoters, new RaftConfig.ControllerQuorumVotersValidator(), HIGH)
```
hachikuji: Can we change `defineInternal` to include the documentation string so that we don't forget to add it back later?

raft/src/test/java/org/apache/kafka/raft/RaftTestUtil.java
```diff
@@ -35,24 +35,10 @@ public static RaftConfig buildRaftConfig(
...
-            votersString.append(prefix);
-            votersString.append(voter.id()).append('@').append(voter.host()).append(':').append(voter.port());
-            prefix = ",";
-        }
-        properties.put(RaftConfig.QUORUM_VOTERS_CONFIG, votersString.toString());
-
-        return new RaftConfig(properties);
+        Map<Integer, InetSocketAddress> voterConnections = voterNodes.stream()
+            .collect(Collectors.toMap(Node::id, node -> new InetSocketAddress(node.host(), node.port())));
+        return new RaftConfig(voterConnections, requestTimeoutMs, retryBackoffMs, electionTimeoutMs, electionBackoffMs,
```
hachikuji: The usage is a tad odd in the tests. We use `voterNodesFromIds` to build `Node` objects from `InetSocketAddress`, then we call `buildRaftConfig` which then turns `Node` back into `InetSocketAddress`. It would be simpler to let `voterNodesFromIds` return `Map<Integer, InetSocketAddress>`. Then we could remove `buildRaftConfig`.

core/src/test/scala/unit/kafka/server/KafkaConfigTest.scala
aloknnikhil: Makes sense. Removed.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Fair enough. Removed.

core/src/main/scala/kafka/server/KafkaConfig.scala
aloknnikhil: Ack. Fixed.

raft/src/test/java/org/apache/kafka/raft/RaftTestUtil.java
aloknnikhil: You're right. This is an artifact from the previous constructor usage. Fixed.

raft/src/test/java/org/apache/kafka/raft/RaftConfigTest.java
aloknnikhil: Ah, good catch. Fixed

raft/src/test/java/org/apache/kafka/raft/RaftTestUtil.java
```diff
@@ -16,36 +16,15 @@
...
-            List<Node> voterNodes
-    ) {
-        Map<Integer, InetSocketAddress> voterConnections = voterNodes.stream()
-            .collect(Collectors.toMap(Node::id, node -> new InetSocketAddress(node.host(), node.port())));
-        return new RaftConfig(voterConnections, requestTimeoutMs, retryBackoffMs, electionTimeoutMs, electionBackoffMs,
-            fetchTimeoutMs, appendLingerMs);
-    }
-
-    public static List<Node> voterNodesFromIds(Set<Integer> voterIds,
+    public static Map<Integer, InetSocketAddress> voterNodesFromIds(Set<Integer> voterIds,
```
hachikuji: nit: maybe `buildAddressMap` or something like that?

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
```diff
@@ -750,9 +748,9 @@ void start(int nodeId) {
             PersistentState persistentState = nodes.get(nodeId);
             MockNetworkChannel channel = new MockNetworkChannel(correlationIdCounter);
             MockMessageQueue messageQueue = new MockMessageQueue();
-            List<Node> voterNodes = voterNodesFromIds(voters, Cluster::nodeAddress);
-            RaftConfig raftConfig = buildRaftConfig(REQUEST_TIMEOUT_MS, RETRY_BACKOFF_MS, ELECTION_TIMEOUT_MS,
-                    ELECTION_JITTER_MS, FETCH_TIMEOUT_MS, LINGER_MS, voterNodes);
+            Map<Integer, InetSocketAddress> voterNodes = voterNodesFromIds(voters, Cluster::nodeAddress);
```
hachikuji: nit: `voterNodes` -> `voterAddresses` or `voterAddressMap`?

raft/src/test/java/org/apache/kafka/raft/RaftTestUtil.java
```diff
@@ -16,36 +16,15 @@
...
                                                Function<Integer, InetSocketAddress> voterAddressGenerator) {
-        return voterIds.stream().map(voterId -> {
-            InetSocketAddress voterAddress = voterAddressGenerator.apply(voterId);
-            return new Node(voterId, voterAddress.getHostName(), voterAddress.getPort());
-        }).collect(Collectors.toList());
+        return voterIds.stream().collect(Collectors.toMap(id -> id, voterAddressGenerator));
```
hachikuji: Guess we can keep it, but this helper doesn't seem to be doing much for us anymore.

aloknnikhil: I'm up for removing any extra files when we don't need them. Aaaaaaand, it's gone.

raft/src/test/java/org/apache/kafka/raft/RaftTestUtil.java
aloknnikhil: Removed in favor of inlined code 
https://github.com/apache/kafka/pull/9916#discussion_r561375612

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
aloknnikhil: Done

hachikuji (approved):
LGTM. Thanks for the patch!

PR #9920, opened 2021-01-17
MINOR: Fix error message in KafkaRaftClient.java

wenbingshen:
before

`logger.trace(
                "Leader doesn't know about snapshot id {}, returned error {} and snapshot id {}",
                state.fetchingSnapshot(),
                partitionSnapshot.errorCode(),
                partitionSnapshot.snapshotId()
            );`

after

`logger.trace(
                "Leader doesn't know about snapshot {}, returned error {} and snapshot id {}",
                state.fetchingSnapshot(),
                partitionSnapshot.errorCode(),
                partitionSnapshot.snapshotId()
            );`

### DIFF
--- a/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
+++ b/raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
@@ -1311,7 +1311,7 @@ public class KafkaRaftClient<T> implements RaftClient<T> {
              * reseting the fetching snapshot state and sending another fetch request.
              */
             logger.trace(
-                "Leader doesn't know about snapshot id {}, returned error {} and snapshot id {}",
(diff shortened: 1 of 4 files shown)


wenbingshen:
@abbccdda  Please take a look at this pr, thanks.

ijuma:
I think the question is whether we intended to only output the snapshot id or more information.

ijuma:
cc @hachikuji

wenbingshen:
> I think the question is whether we intended to only output the snapshot id or more information.

I agree with you.I think we could choose to just print the snapshot Id or we can print more information about the snapshot.But the current printing information is not very clear, there are some problems.

wenbingshen:
@hachikuji  Hello, can you take a look at this pr? I think we can choose to just print the snapshot Id or we can print more information about the snapshot.Here's one way to do it.

PR #9921, opened 2021-01-17
MINOR: Ensure compile and runtime classpaths have consistent versions

ijuma:
Ensure that runtime-only dependencies don't cause a different version to be used.
More specifically:

> The relationship is directed, which means that if the runtimeClasspath configuration
has to be resolved, Gradle will first resolve the compileClasspath and then "inject" the
result of resolution as strict constraints into the runtimeClasspath.

For more details, see:
https://docs.gradle.org/6.8/userguide/resolution_strategy_tuning.html#sec:configuration_consistency

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -194,6 +194,13 @@ subprojects {
   sourceCompatibility = minJavaVersion
   targetCompatibility = minJavaVersion
 
+  java {
+    consistentResolution {
+      // resolve the compileClasspath and then "inject" the result of resolution as strict constraints into the runtimeClasspath
+      useCompileClasspathVersions()
+    }
+  }
+
   tasks.withType(JavaCompile) {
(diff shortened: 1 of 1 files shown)


chia7712 (approved):
This patch impacts only the dependencies used by testing. I check ```core```, ```client```,  ```connect-runtime``` and ```stream```. Only one dependency is changed.

1. org.javassist:javassist:3.27.0-GA -> 3.26.0-GA

so +1

ijuma:
Yeah, we don't use runtime dependencies much at the moment, but you imagine us doing that for log4j in the broker (as one example). So, I think this is a good change to make sure compilation and runtime versions are consistent.

ijuma:
Failures are flaky and unrelated.

PR #9925, opened 2021-01-18
MINOR: Tag `RaftEventSimulationTest` as `integration` and tweak it

ijuma:
The test takes over 1 minute to run, so it should not be considered a
unit test.

Also:
* Replace `test` prefix with `check` prefix for helper methods. A common
mistake is to forget to add the @Test annotation, so it's good to use a
different naming convention for methods that should have the annotation
versus methods that should not.
* Replace `Action` functional interface with built-in `Runnable`.
* Remove unnecessary `assumeTrue`.
* Remove `@FunctionalInterface` from `Invariant` since it's not used
in that way.

### DIFF
--- a/raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
+++ b/raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
@@ -30,6 +30,7 @@ import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.raft.MockLog.LogBatch;
 import org.apache.kafka.raft.MockLog.LogEntry;
 import org.apache.kafka.raft.internals.BatchMemoryPool;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


showuon (approved):
@ijuma , I love the refactor to replace `Action` functional interface with built-in `Runnable`. LGTM!

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
```diff
@@ -59,6 +60,7 @@
 import static org.junit.jupiter.api.Assertions.fail;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
```
chia7712: Is ```assumeTrue``` used in this test? I test ```RaftEventSimulationTest`` but there is no ignored test cases.

chia7712 (commented):
There are other functional interfaces (```Validation``` and ```Invariant```). The setter (```addInvariant``` and ```addValidation```) should be clear enough so we don't need to keep those function interfaces. WDTY?

ijuma:
@chia7712 What are you suggesting instead of `Validation` and `Invariant`? `Runnable`?

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
ijuma: Good catch, removed since it's not currently needed.

chia7712:
> What are you suggesting instead of Validation and Invariant? Runnable?

Runnable

ijuma:
I removed the @FunctionalInterface annotation from `Invariant`. I don't think this interfaces are used as functional interfaces and they make the system more strongly typed. `Action`, on the other hand, was playing a very similar role to `Runnable` and hence why I thought it was not worth keeping.

ijuma:
@chia7712 Does this look ok to you? I am interested in making `unitTest` fast, so would like to get this merged.

chia7712 (approved):
LGTM!

PR #9928, opened 2021-01-19
MINOR: Exclude junit 3 transitive dependency from jfreechart

ijuma:
This was causing IntelliJ to choose the vintage runner when running `core` tests
which would then fail.

Verified that `core` tests work in IntelliJ again after this change.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -787,7 +787,9 @@ project(':core') {
     testCompile libs.apachedsJdbmPartition
     testCompile libs.junitJupiter
     testCompile libs.slf4jlog4j
-    testCompile libs.jfreechart
+    testCompile(libs.jfreechart) {
+      exclude group: 'junit', module: 'junit'
+    }
   }
 
   if (userEnableTestCoverage) {


chia7712 (approved):
There is an unrelated issue. 

```jfreechart ```is used by ```ReplicationQuotasTestRig```.  ```ReplicationQuotasTestRig``` does not have tests and there is no script for it. How we handle such tool class?

ijuma:
@chia7712 Unfortunately that has to be tested manually currently.

ijuma:
It looks like the JDK 8 build is occasionally exiting during the run. Since this change fixes an important developer issue and has nothing that would only fail on JDK 8, I'll go ahead and merge.

chia7712:
> Unfortunately that has to be tested manually currently.

I don't think this patch can break the tool. However, it is weird that we have a orphan tool :(

ijuma:
@chia7712 that tool is a manual test. It is meant to verify that throttling works well in more complicated scenarios.

PR #9930, opened 2021-01-19
MINOR: Fix always-passing validation in TestRecordTest#testProducerRecord

dongjinleekr:
Since the types of `ProducerRecord` and `TestRecord` are different, calling `assertNotEquals` for these instances always succeeds - and meaningless.

This commit improves `TestRecordTest` by comparing the instances deeply.

cc/ @chia7712

### DIFF
--- a/streams/test-utils/src/test/java/org/apache/kafka/streams/test/TestRecordTest.java
+++ b/streams/test-utils/src/test/java/org/apache/kafka/streams/test/TestRecordTest.java
@@ -23,6 +23,7 @@ import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.header.internals.RecordHeader;
 import org.apache.kafka.common.header.internals.RecordHeaders;
 import org.apache.kafka.common.record.TimestampType;
+
 import org.junit.jupiter.api.Test;
 
 import java.time.Instant;
@@ -166,6 +167,5 @@ public class TestRecordTest {
         final TestRecord<String, Integer> testRecord = new TestRecord<>(producerRecord);
         final TestRecord<String, Integer> expectedRecord = new TestRecord<>(key, value, headers, recordTime);
         assertEquals(expectedRecord, testRecord);
-        assertNotEquals(expectedRecord, producerRecord);
     }
 }


streams/test-utils/src/test/java/org/apache/kafka/streams/test/TestRecordTest.java
```diff
@@ -166,6 +166,9 @@ public void testProducerRecord() {
         final TestRecord<String, Integer> testRecord = new TestRecord<>(producerRecord);
         final TestRecord<String, Integer> expectedRecord = new TestRecord<>(key, value, headers, recordTime);
         assertEquals(expectedRecord, testRecord);
-        assertNotEquals(expectedRecord, producerRecord);
+        assertFalse(expectedRecord.getHeaders() == producerRecord.headers() &&
```
chia7712: Could we separate this assert by ```assertEquals```? It can produce more meaningful error message.

chia7712: Also, it should make sure the input arguments (from ```ProducerRecord```) are equals to getter of ```TestRecord```. For example:
```java
assertEquals(expectedRecord.getHeaders(), producerRecord.headers())
```

chia7712 (commented):
@dongjinleekr  Thanks for your patch!

streams/test-utils/src/test/java/org/apache/kafka/streams/test/TestRecordTest.java
dongjinleekr: @chia7712 Sorry for the confusion.  Since `expectedRecord` is logically identical to `testRecord` and `producerRecord` is created fro `testRecord`, `expectedRecord` and `testRecord` are logically equal. The reason why the previous assertion was always true was not for its logical state, but the difference between their classes - `ProducerRecord` vs. `TestRecord`.

So, after some thought, I concluded that just removing the meaningless assertion is enough. Here is the update.

chia7712: fair enough :)

chia7712 (approved):
@dongjinleekr Thanks for your patch. LGTM. will commit it tomorrow.

streams/test-utils/src/test/java/org/apache/kafka/streams/test/TestRecordTest.java
chia7712: Is the assert  intentional? If so, we should keep it and then add clear comments to avoid confusing readers in the future.

WDYT?

dongjinleekr:
ping @chia7712

chia7712:
@dongjinleekr Please take a look at this comment (https://github.com/apache/kafka/pull/9930#discussion_r561021090)

streams/test-utils/src/test/java/org/apache/kafka/streams/test/TestRecordTest.java
dongjinleekr: > Is the assert intentional?

It seems not. If it was interntional, the same statement must be included in its consumer-side counterpart test method, `testConsumerRecord`, which is actually not.

dongjinleekr:
Good job, @chia7712! :congratulations:

PR #9932, opened 2021-01-19
MINOR: Upgrade ducktape to version 0.7.11

stan-is-hate:
Ducktape 0.7.11 fixes a bug where a unicode exception message would cause test runner to hang up and never finish.
This change should be backported to all the branches using ducktape 0.7.10. 
An update to ducktape 0.8.0 will be released separately (and if used with python 3 it probably won't be an issue anyway)


stan-is-hate:
@andrewegel @omkreddy

kkonstantine (approved):
The update LGTM

I assume you've run tests both locally and on a jenkins builder?

kkonstantine:
Few flaky failures. Green builds otherwise on jdk8 and jdk11

kkonstantine:
Merged to 2.6 and backported up to 2.0.

stan-is-hate:
Thanks for review and backport! Changes to ducktape were simply a bugfix to move a bunch of code into the `finally` block to make sure it executes, so they shouldn't affect test results.

PR #9934, opened 2021-01-20
MINOR: Drop enable.metadata.quorum config

hachikuji:
The primary purpose of this patch is to remove the internal `enable.metadata.quorum` configuration. Instead, we rely on `process.roles` to determine if the self-managed quorum has been enabled. As a part of this, I've done the following:

1. Replace the notion of "disabled" APIs with "controller-only" APIs. We previously marked some APIs which were intended only for the KIP-500 as "disabled" so that they would not be unintentionally exposed. For example, the Raft quorum APIs were disabled. Marking them as "controller-only" carries the same effect, but makes the intent that they should be only exposed by the KIP-500 controller clearer.
2. Make `ForwardingManager` optional in `KafkaServer` and `KafkaApis`. Previously we used `null` if forwarding was enabled and relied on the metadata quorum check.
3. Make `zookeeper.connect` an optional configuration if `process.roles` is defined.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
+++ b/clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
@@ -90,15 +90,15 @@ public enum ApiKeys {
     ALTER_CLIENT_QUOTAS(ApiMessageType.ALTER_CLIENT_QUOTAS, false, true),
     DESCRIBE_USER_SCRAM_CREDENTIALS(ApiMessageType.DESCRIBE_USER_SCRAM_CREDENTIALS),
     ALTER_USER_SCRAM_CREDENTIALS(ApiMessageType.ALTER_USER_SCRAM_CREDENTIALS, false, true),
-    VOTE(ApiMessageType.VOTE, true, RecordBatch.MAGIC_VALUE_V0, false, false),
-    BEGIN_QUORUM_EPOCH(ApiMessageType.BEGIN_QUORUM_EPOCH, true, RecordBatch.MAGIC_VALUE_V0, false, false),
-    END_QUORUM_EPOCH(ApiMessageType.END_QUORUM_EPOCH, true, RecordBatch.MAGIC_VALUE_V0, false, false),
-    DESCRIBE_QUORUM(ApiMessageType.DESCRIBE_QUORUM, true, RecordBatch.MAGIC_VALUE_V0, false, false),
+    VOTE(ApiMessageType.VOTE, true, RecordBatch.MAGIC_VALUE_V0, false, true),
+    BEGIN_QUORUM_EPOCH(ApiMessageType.BEGIN_QUORUM_EPOCH, true, RecordBatch.MAGIC_VALUE_V0, false, true),
+    END_QUORUM_EPOCH(ApiMessageType.END_QUORUM_EPOCH, true, RecordBatch.MAGIC_VALUE_V0, false, true),
+    DESCRIBE_QUORUM(ApiMessageType.DESCRIBE_QUORUM, true, RecordBatch.MAGIC_VALUE_V0, false, true),
     ALTER_ISR(ApiMessageType.ALTER_ISR, true),
     UPDATE_FEATURES(ApiMessageType.UPDATE_FEATURES, false, true),
-    ENVELOPE(ApiMessageType.ENVELOPE, true, RecordBatch.MAGIC_VALUE_V0, false, false),
-    FETCH_SNAPSHOT(ApiMessageType.FETCH_SNAPSHOT, false, RecordBatch.MAGIC_VALUE_V0, false, false),
-    DESCRIBE_CLUSTER(ApiMessageType.DESCRIBE_CLUSTER, false, RecordBatch.MAGIC_VALUE_V0, false, true);
+    ENVELOPE(ApiMessageType.ENVELOPE, true, RecordBatch.MAGIC_VALUE_V0, false, true),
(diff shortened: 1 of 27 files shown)


abbccdda:
@hachikuji Do we still need this PR since Colin already did one?

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -638,6 +637,9 @@ object KafkaConfig {
   val RequestTimeoutMsDoc = CommonClientConfigs.REQUEST_TIMEOUT_MS_DOC
   val ConnectionSetupTimeoutMsDoc = CommonClientConfigs.SOCKET_CONNECTION_SETUP_TIMEOUT_MS_DOC
   val ConnectionSetupTimeoutMaxMsDoc = CommonClientConfigs.SOCKET_CONNECTION_SETUP_TIMEOUT_MAX_MS_DOC
+  val ProcessRolesDoc = "[ALPHA] The roles that this process plays: 'broker', 'controller', or 'broker,controller' if it is both. " +
```
ijuma: Can we call it "Early Access" instead of Alpha?

hachikuji: I think I'm going to keep this as internal for now. Let's reconsider this after we have something that can actually start up.

cmccabe:
> @hachikuji Do we still need this PR since Colin already did one?

That PR was for the KIP-500 branch, not for trunk.  It also didn't refactor the "disabled" APIs concept, which I think is useful.

> Replace the notion of "disabled" APIs with "controller-only" APIs. We previously marked some APIs which were intended only for the KIP-500 as "disabled" so that they would not be unintentionally exposed

That seems reasonable.  I guess the idea is that the broker will not include those APIs in its ApiVersionsResponse.  However, the controller always will (not that clients will talk directly to the controller anyway...)

hachikuji:
@cmccabe Yeah, that's right. Let me see if I can improve this a little bit to make the intent clearer. We have some APIs which will be exposed by both the broker and the controller, so I think this should probably be a set rather than a flag.

hachikuji:
I decided to leave the controller-only flag as it is. I think there are further improvements here to make the scope of the API clearer, but the implications for compatibility are subtle enough that we should consider it separately.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -116,8 +116,8 @@
     /** indicates the minimum required inter broker magic required to support the API */
     public final byte minRequiredInterBrokerMagic;
 
-    /** indicates whether the API is enabled and should be exposed in ApiVersions **/
-    public final boolean isEnabled;
+    /** indicates whether this is an API which is only exposed by the KIP-500 controller **/
+    public final boolean isControllerOnlyApi;
```
abbccdda, line 120: Could we get a JIRA to track the work for refactoring these boolean flags into static collections of api keys? To me it is not easy to use constructor correctly when there are multiple of them.

raft/README.md
```diff
@@ -12,17 +12,14 @@ Below we describe the details to set this up.
     bin/test-raft-server-start.sh config/raft.properties
 
 ### Run Multi Node Quorum ###
-Create 3 separate raft quorum properties as the following
```
abbccdda, line 15: Do we need to define `process.roles` here?

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1876,5 +1874,9 @@ class KafkaConfig(val props: java.util.Map[_, _], doLog: Boolean, dynamicConfigO
         s"${KafkaConfig.FailedAuthenticationDelayMsProp}=$failedAuthenticationDelayMs should always be less than" +
         s" ${KafkaConfig.ConnectionsMaxIdleMsProp}=$connectionsMaxIdleMs to prevent failed" +
         s" authentication responses from timing out")
+
+    if (requiresZookeeper && zkConnect == null) {
```
abbccdda, line 1909: Do we have test coverage for this check?

core/src/main/scala/kafka/Kafka.scala
```diff
@@ -65,11 +65,12 @@ object Kafka extends Logging {
 
   private def buildServer(props: Properties): Server = {
     val config = KafkaConfig.fromProps(props, false)
-    if (config.processRoles.isEmpty) {
+    if (config.requiresZookeeper) {
```
abbccdda, line 68: What if we call `processRoles.isDefined` here?

raft/README.md
hachikuji: Yeah, you're right. That's pretty annoying. Let me see if I can do anything about it.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
hachikuji: Filed this JIRA: https://issues.apache.org/jira/browse/KAFKA-12232.

core/src/main/scala/kafka/server/KafkaConfig.scala
hachikuji: Ack. Will add a test case.

core/src/main/scala/kafka/Kafka.scala
hachikuji: Hmm `requiresZookeeper` seemed more explicit. I thought using `processRoles` was a little obscure.

abbccdda (approved):
LGTM

PR #9938, opened 2021-01-20
MINOR: Refactor DescribeAuthorizedOperationsTest

dajac:
I was looking at `DescribeAuthorizedOperationsTest` in the context of KIP-700. I have made few changes to improve the readability of the code/tests. The PR only moves code around.

### DIFF
--- a/core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
+++ b/core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
@@ -20,7 +20,7 @@ import kafka.security.authorizer.{AclAuthorizer, AclEntry}
 import kafka.server.KafkaConfig
 import kafka.utils.{CoreUtils, JaasTestUtils, TestUtils}
 import org.apache.kafka.clients.admin._
-import org.apache.kafka.common.acl.AclOperation.{ALTER, CLUSTER_ACTION, DESCRIBE}
+import org.apache.kafka.common.acl.AclOperation.{ALL, ALTER, CLUSTER_ACTION, DELETE, DESCRIBE}
 import org.apache.kafka.common.acl.AclPermissionType.ALLOW
 import org.apache.kafka.common.acl._
 import org.apache.kafka.common.resource.{PatternType, Resource, ResourcePattern, ResourceType}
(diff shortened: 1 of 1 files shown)


core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
```diff
@@ -86,54 +122,36 @@ class DescribeAuthorizedOperationsTest extends IntegrationTestHarness with SaslS
...
-
-  def createConfig(): Properties = {
+  private def createConfig(): Properties = {
     val adminClientConfig = new Properties()
     adminClientConfig.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, brokerList)
     adminClientConfig.put(AdminClientConfig.REQUEST_TIMEOUT_MS_CONFIG, "20000")
     val securityProps: util.Map[Object, Object] =
       TestUtils.adminClientSecurityConfigs(securityProtocol, trustStoreFile, clientSaslProperties)
-    securityProps.forEach { (key, value) => adminClientConfig.put(key.asInstanceOf[String], value) }
+    securityProps.forEach((key, value) => adminClientConfig.put(key.asInstanceOf[String], value))
```
chia7712: How about ```adminClientConfig.putAll(securityProps)```?

core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
```diff
@@ -32,17 +32,55 @@ import org.junit.jupiter.api.{AfterEach, BeforeEach, Test}
...
+    accessControlEntry(JaasTestUtils.KafkaClientPrincipalUnqualifiedName2, ALL, ALLOW))
+
+  val Topic2All = new AclBinding(
+    new ResourcePattern(ResourceType.TOPIC, Topic2, PatternType.LITERAL),
+    accessControlEntry(JaasTestUtils.KafkaClientPrincipalUnqualifiedName2, DELETE, ALLOW))
+
+  private def accessControlEntry(
+    userName: String,
+    operation: AclOperation,
+    permissionType: AclPermissionType
```
chia7712: It seems this filed is always ```ALLOW```

core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
```diff
@@ -32,17 +32,55 @@ import org.junit.jupiter.api.{AfterEach, BeforeEach, Test}
 
 import scala.jdk.CollectionConverters._
 
+object DescribeAuthorizedOperationsTest {
+  val Group1 = "group1"
```
chia7712, line 36: Is ```GROUP_1" more better? it is global variable in this test.

core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
```diff
@@ -73,11 +114,6 @@ class DescribeAuthorizedOperationsTest extends IntegrationTestHarness with SaslS
...
   }
 
-  private def accessControlEntry(userName: String, permissionType: AclPermissionType, operation: AclOperation): AccessControlEntry = {
-    new AccessControlEntry(new KafkaPrincipal(KafkaPrincipal.USER_TYPE, userName).toString,
-      AclEntry.WildcardHost, operation, permissionType)
-  }
-
   @AfterEach
   override def tearDown(): Unit = {
     if (client != null)
```
chia7712: unnecessary check

core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
```diff
@@ -32,17 +32,55 @@ import org.junit.jupiter.api.{AfterEach, BeforeEach, Test}
...
+}
+
 class DescribeAuthorizedOperationsTest extends IntegrationTestHarness with SaslSetup {
+  import DescribeAuthorizedOperationsTest._
+
   override val brokerCount = 1
   this.serverConfig.setProperty(KafkaConfig.ZkEnableSecureAclsProp, "true")
   this.serverConfig.setProperty(KafkaConfig.AuthorizerClassNameProp, classOf[AclAuthorizer].getName)
 
   var client: Admin = _
```
chia7712, line 82: all test cases need client so we should create it in construction or ```BeforeEach``` block.

chia7712 (approved):
@dajac Nice refactor. overall + 1

core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
dajac: Indeed, let me remove that field.

core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
dajac: In Scala, we tend to only capitalize the first letter of constants. At least, this how I have been doing it so far. I prefer to keep it as is.

core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
dajac: I think that it is better to keep it as `client` could not be initialized. For instance, let's say that there is an exception raised when the client is constructed. In this case, `client` would remain `null`.

dajac:
@chia7712 Thanks for your review. I have addressed your comments. Do you mind having a second quick look at it?

core/src/test/scala/integration/kafka/api/DescribeAuthorizedOperationsTest.scala
chia7712: Sorry for unclear comment. 

My point was ```Utils.closeQuietly``` includes null check so we don't need to add same check again.

dajac: Got it, thanks for the clarification. I have removed the null check.

chia7712 (approved):
@dajac +1 to nice patch!

dajac:
The failed test is not related to this PR:
* Build / JDK 15 / org.apache.kafka.clients.consumer.internals.FetcherTest.testEarlierOffsetResetArrivesLate()

PR #9941, opened 2021-01-20
MINOR: further reduce StreamThread loop INFO logging to 2min summary

ableegoldman:
One more try to get the logging levels right..the only way to log something within the main StreamThread loop without absolutely flooding the logs at a level that isn't appropriate for INFO is to just set some kind of interval to stick to. I chose to log the summary every 2 min, since this is long enough to prevent log spam but short enough to fit at least 2 summaries within the (default) poll interval of 5 min

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -275,6 +275,12 @@ public class StreamThread extends Thread {
     private final Sensor commitRatioSensor;
     private final Sensor failedStreamThreadSensor;
 
+    private static final long LOG_SUMMARY_INTERVAL_MS = 2 * 60 * 1000L; // log a summary of processing every 2 minutes
+    private long lastLogSummaryMs = -1L;
+    private long totalRecordsProcessedSinceLastSummary = 0L;
+    private long totalPunctuatorsSinceLastSummary = 0L;
+    private long totalCommittedSinceLastSummary = 0L;
+
     private long now;
     private long lastPollMs;
     private long lastCommitMs;
@@ -407,7 +413,7 @@ public class StreamThread extends Thread {
             referenceContainer.nextScheduledRebalanceMs,
             shutdownErrorHook,
             streamsUncaughtExceptionHandler,
-            cacheSize -> cache.resize(cacheSize)
+            cache::resize
         );
 
         taskManager.setPartitionResetter(partitions -> streamThread.resetOffsets(partitions, null));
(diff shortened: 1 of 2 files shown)


ableegoldman:
call for review @lct45 @wcarlson5 @cadonna

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -1053,7 +1054,7 @@ int maybeCommitActiveTasksPerUserRequested() {
     }
 
     private void commitOffsetsOrTransaction(final Map<Task, Map<TopicPartition, OffsetAndMetadata>> offsetsPerTask) {
-        log.debug("Committing task offsets {}", offsetsPerTask);
+        log.debug("Committing task offsets {}", offsetsPerTask.entrySet().stream().collect(Collectors.toMap(t -> t.getKey().id(), Entry::getValue))); // avoid logging actual Task objects
```
ableegoldman, line 1057: An unrelated but equally annoying thing I noticed in the logs: we should never log a full `Task` object because it prints literally everything about the task, including for example the topology description which is not that useful but sometimes VERY long

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -407,7 +413,7 @@ public static StreamThread create(final InternalTopologyBuilder builder,
             referenceContainer.nextScheduledRebalanceMs,
             shutdownErrorHook,
             streamsUncaughtExceptionHandler,
-            cacheSize -> cache.resize(cacheSize)
```
wcarlson5, line 410: My ide tried to optimize this as well. At the time not passing in cacheSize caused some expections. I would be careful about making this change without need

wcarlson5:
Overall LGTM. I am not sure about the cache change but the changes to the log makes a lot of sense

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
ableegoldman: Interesting. It *should* be exactly the same, but of course who knows with Java. Did it cause a test to fail or was it something more subtle?

wcarlson5: I think it caused a test to fail but not everytime. It also could have been fixed since then as changes have been made. If all the tests pass it's probably fine

guozhangwang:
LGTM.

ableegoldman:
All tests passed except the broken `StoreQueryIntegrationTest`, merging to trunk

ableegoldman:
Cherrypicked to 2.7

PR #9945, opened 2021-01-21
KAFKA-12212; Bump Metadata API version to remove `ClusterAuthorizedOperations` fields

dajac:
This PR bumps the version of the Metadata API and deprecates the `IncludeClusterAuthorizedOperations` and the `IncludeClusterAuthorizedOperations` fields from version 11 and onward.

### DIFF
--- a/clients/src/main/resources/common/message/MetadataRequest.json
+++ b/clients/src/main/resources/common/message/MetadataRequest.json
@@ -34,7 +34,8 @@
     //
     // Version 10 adds topicId.
     //
-    // Version 11 deprecates IncludeClusterAuthorizedOperations field (KIP-700).
+    // Version 11 deprecates IncludeClusterAuthorizedOperations field. This is now exposed
+    // by the DescribeCluster API (KIP-700).
     { "name": "Topics", "type": "[]MetadataRequestTopic", "versions": "0+", "nullableVersions": "1+",
       "about": "The topics to fetch metadata for.", "fields": [
       { "name": "TopicId", "type": "uuid", "versions": "10+", "ignorable": true, "about": "The topic id." },
--- a/clients/src/main/resources/common/message/MetadataRequest.json
+++ b/clients/src/main/resources/common/message/MetadataRequest.json
@@ -17,7 +17,7 @@
   "apiKey": 3,
   "type": "request",
   "name": "MetadataRequest",
-  "validVersions": "0-10",
+  "validVersions": "0-11",
   "flexibleVersions": "9+",
   "fields": [
     // In version 0, an empty array indicates "request metadata for all topics."  In version 1 and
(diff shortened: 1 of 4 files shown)


core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1246,13 +1246,15 @@ class KafkaApis(val requestChannel: RequestChannel,
       }
 
     var clusterAuthorizedOperations = Int.MinValue
```
chia7712: the default value of ```clusterAuthorizedOperations``` in auto-generated protocol is ```-2147483648```. If the version of request is bigger than 10, does it cause error if we set ```Int.MinValue``` to ```clusterAuthorizedOperations``` of ```MetadataResponseData```?

clients/src/main/resources/common/message/MetadataRequest.json
```diff
@@ -31,7 +31,10 @@
     // Starting in version 8, authorized operations can be requested for cluster and topic resource.
     //
     // Version 9 is the first flexible version.
-    // Version 10 add topicId
+    //
+    // Version 10 adds topicId.
+    //
+    // Version 11 deprecates IncludeClusterAuthorizedOperations field (KIP-700).
```
chia7712: How about saying this function is migrated to ```DescribeClusterRequest```?

chia7712 (commented):
@dajac thanks for your patch!

core/src/main/scala/kafka/server/KafkaApis.scala
dajac: `-2147483648` is actually `Int.MinValue`. It would cause an error if we would write something different from the default value for versions < 8 and > 10.

clients/src/main/resources/common/message/MetadataRequest.json
dajac: Yeah, why not. I will add it.

core/src/main/scala/kafka/server/KafkaApis.scala
chia7712: Could you add the comments for that value (```Int.MinValue```) for dumb readers like me :)

dajac: Will do.

dajac:
@chia7712 Thanks for your comments. I have updated the PR to address them.

chia7712 (approved):
@dajac Thanks for your patch. +1

rajinisivaram (approved):
@dajac Thanks for the PR, LGTM

dajac:
Failed tests are not related:
* Build / JDK 8 / org.apache.kafka.clients.consumer.KafkaConsumerTest.testCloseWithTimeUnit()
* Build / JDK 8 / org.apache.kafka.clients.consumer.internals.FetcherTest.testEarlierOffsetResetArrivesLate()

Merging to trunk.

PR #9947, opened 2021-01-21
KAFKA-12190: Fix setting of file permissions on non-POSIX filesystems

wilkinsona:
Previously, `StateDirectory` used `PosixFilePermissions` to configure its directories' permissions which fails on Windows as its file system is not POSIX-compliant. This PR updates `StateDirectory` to fall back to the `File` API on non-POSIX-compliant file systems. The File API doesn't allow as much control over the permissions so they're as close as the API permits.

The unit tests have been updated to also verify the behaviour on non-POSIX-compliant file systems.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
@@ -31,7 +31,6 @@ import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
 import java.nio.channels.OverlappingFileLockException;
 import java.nio.file.NoSuchFileException;
-import java.nio.file.Paths;
 import java.nio.file.Path;
 import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
@@ -110,12 +109,13 @@ public class StateDirectory {
                     " due to the fact that this directory can be cleared by the OS");
             }
             // change the dir permission to "rwxr-x---" to avoid world readable
-            configurePermissions(Paths.get(baseDir.getPath()));
-            configurePermissions(Paths.get(stateDir.getPath()));
+            configurePermissions(baseDir);
+            configurePermissions(stateDir);
         }
     }
     
-    private void configurePermissions(final Path path) {
+    private void configurePermissions(final File file) {
+        final Path path = file.toPath();
         if (path.getFileSystem().supportedFileAttributeViews().contains("posix")) {
             final Set<PosixFilePermission> perms = PosixFilePermissions.fromString("rwxr-x---");
             try {
@@ -124,7 +124,6 @@ public class StateDirectory {
                 log.error("Error changing permissions for the directory {} ", path, e);
             }
         } else {
-            final File file = path.toFile();
             boolean set = file.setReadable(true, true);
             set &= file.setWritable(true, true);
             set &= file.setExecutable(true, true);
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
@@ -109,16 +109,27 @@ public class StateDirectory {
                 log.warn("Using /tmp directory in the state.dir property can cause failures with writing the checkpoint file" +
                     " due to the fact that this directory can be cleared by the OS");
             }
-
             // change the dir permission to "rwxr-x---" to avoid world readable
-            final Path basePath = Paths.get(baseDir.getPath());
-            final Path statePath = Paths.get(stateDir.getPath());
+            configurePermissions(Paths.get(baseDir.getPath()));
+            configurePermissions(Paths.get(stateDir.getPath()));
+        }
+    }
+    
+    private void configurePermissions(final Path path) {
+        if (path.getFileSystem().supportedFileAttributeViews().contains("posix")) {
             final Set<PosixFilePermission> perms = PosixFilePermissions.fromString("rwxr-x---");
             try {
-                Files.setPosixFilePermissions(basePath, perms);
-                Files.setPosixFilePermissions(statePath, perms);
+                Files.setPosixFilePermissions(path, perms);
             } catch (final IOException e) {
-                log.error("Error changing permissions for the state or base directory {} ", stateDir.getPath(), e);
+                log.error("Error changing permissions for the directory {} ", path, e);
+            }
+        } else {
+            final File file = path.toFile();
+            boolean set = file.setReadable(true, false);
+            set &= file.setWritable(true, true);
+            set &= file.setExecutable(true, false);
+            if (!set) {
+                log.error("Failed to change permissions for the directory {}", file);
             }
         }
     }
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
@@ -125,9 +125,9 @@ public class StateDirectory {
             }
         } else {
             final File file = path.toFile();
-            boolean set = file.setReadable(true, false);
+            boolean set = file.setReadable(true, true);
             set &= file.setWritable(true, true);
-            set &= file.setExecutable(true, false);
+            set &= file.setExecutable(true, true);
             if (!set) {
                 log.error("Failed to change permissions for the directory {}", file);
             }
--- a/streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
@@ -128,8 +128,8 @@ public class StateDirectoryTest {
                     PosixFilePermission.GROUP_EXECUTE,
                     PosixFilePermission.OWNER_READ);
             try {
-                final Set<PosixFilePermission> baseFilePermissions = Files.getPosixFilePermissions(path);
-                assertThat(expectedPermissions, equalTo(baseFilePermissions));
+                final Set<PosixFilePermission> filePermissions = Files.getPosixFilePermissions(path);
+                assertThat(expectedPermissions, equalTo(filePermissions));
             } catch (final IOException e) {
                 fail("Should create correct files and set correct permissions");
             }
--- a/streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
@@ -115,22 +115,29 @@ public class StateDirectoryTest {
 
     @Test
     public void shouldHaveSecurePermissions() {
-        final Set<PosixFilePermission> expectedPermissions = EnumSet.of(
-            PosixFilePermission.OWNER_EXECUTE,
-            PosixFilePermission.GROUP_READ,
-            PosixFilePermission.OWNER_WRITE,
-            PosixFilePermission.GROUP_EXECUTE,
-            PosixFilePermission.OWNER_READ);
-
-        final Path statePath = Paths.get(stateDir.getPath());
-        final Path basePath = Paths.get(appDir.getPath());
-        try {
-            final Set<PosixFilePermission> baseFilePermissions = Files.getPosixFilePermissions(statePath);
-            final Set<PosixFilePermission> appFilePermissions = Files.getPosixFilePermissions(basePath);
-            assertThat(expectedPermissions, equalTo(baseFilePermissions));
-            assertThat(expectedPermissions, equalTo(appFilePermissions));
-        } catch (final IOException e) {
-            fail("Should create correct files and set correct permissions");
+        assertPermissions(Paths.get(stateDir.getPath()));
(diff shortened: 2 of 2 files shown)


streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
```diff
@@ -115,22 +115,29 @@ public void shouldCreateBaseDirectory() {
...
+    private void assertPermissions(final Path path) {
+        if (path.getFileSystem().supportedFileAttributeViews().contains("posix")) {
+            final Set<PosixFilePermission> expectedPermissions = EnumSet.of(
+                    PosixFilePermission.OWNER_EXECUTE,
+                    PosixFilePermission.GROUP_READ,
+                    PosixFilePermission.OWNER_WRITE,
+                    PosixFilePermission.GROUP_EXECUTE,
+                    PosixFilePermission.OWNER_READ);
+            try {
+                final Set<PosixFilePermission> baseFilePermissions = Files.getPosixFilePermissions(path);
```
showuon: This path could be `baseFile` and also `appFile`. I think we'd better to rename it, ex: `filePermissions` or any better idea?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -109,16 +109,27 @@ public StateDirectory(final StreamsConfig config, final Time time, final boolean
...
-                Files.setPosixFilePermissions(statePath, perms);
+                Files.setPosixFilePermissions(path, perms);
             } catch (final IOException e) {
-                log.error("Error changing permissions for the state or base directory {} ", stateDir.getPath(), e);
+                log.error("Error changing permissions for the directory {} ", path, e);
+            }
+        } else {
+            final File file = path.toFile();
+            boolean set = file.setReadable(true, false);
+            set &= file.setWritable(true, true);
```
ableegoldman, line 128: If you only have `file.setWritable(true, true)` then the directory will still be writeable by non-users, I assume?  I actually don't know the details of the `File#setXXX` methods -- but we don't want it to be writeable by just anyone. Should we instead do something like

```suggestion
            set &= file.setWritable(false) && file.setWritable(true, true);
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -109,16 +109,27 @@ public StateDirectory(final StreamsConfig config, final Time time, final boolean
...
-                Files.setPosixFilePermissions(basePath, perms);
-                Files.setPosixFilePermissions(statePath, perms);
+                Files.setPosixFilePermissions(path, perms);
             } catch (final IOException e) {
-                log.error("Error changing permissions for the state or base directory {} ", stateDir.getPath(), e);
+                log.error("Error changing permissions for the directory {} ", path, e);
+            }
+        } else {
+            final File file = path.toFile();
+            boolean set = file.setReadable(true, false);
```
ableegoldman: I think we actually want it to be readable _only_ by the user, and explicitly restrict permissions for all other users. The patch which originally broke things for Windows users was trying to tighten up the security in exactly this way

ableegoldman (commented):
Thanks for the fix! I had a few questions about how to set the permissions for non-users, but I may be out of my depth when it comes to non-posix permissions so let me know if I'm just not on the same page here.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -109,16 +109,27 @@ public StateDirectory(final StreamsConfig config, final Time time, final boolean
...
             try {
-                Files.setPosixFilePermissions(basePath, perms);
-                Files.setPosixFilePermissions(statePath, perms);
+                Files.setPosixFilePermissions(path, perms);
             } catch (final IOException e) {
-                log.error("Error changing permissions for the state or base directory {} ", stateDir.getPath(), e);
+                log.error("Error changing permissions for the directory {} ", path, e);
+            }
+        } else {
+            final File file = path.toFile();
```
showuon: We pass the `path` parameter which is converted from `File` instance (i.e. `Paths.get(stateDir.getPath()`), and then we converted it back to `File` instance in the method. It's redundant. Could we just pass the `File` instance into the method directly?

streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
```diff
@@ -115,22 +115,29 @@ public void shouldCreateBaseDirectory() {
...
+                assertThat(expectedPermissions, equalTo(baseFilePermissions));
+            } catch (final IOException e) {
+                fail("Should create correct files and set correct permissions");
+            }
+        } else {
+            final File file = path.toFile();
```
showuon: Same here, could we pass `File` instance into method directly?

showuon (commented):
Thanks for the fix. Overall looks good. Left some comments. Thanks.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
wilkinsona: The `File` API only allows you to set permissions for the owner or for everyone. A call to `setWritable(writable)` just delegates to `setWritable(writable, true)`. I prefer to call `setWritable(writable, ownerOnly)` directly as I think the intent is slightly clearer. Calling `setWritable(true, true)` or `setWritable(true)` makes the file writable only by the owner.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
wilkinsona: There's an imperfect mapping between POSIX-style permissions and what can be configured using the File API as it's just owner and everyone else. I'll change it so it's only readable by the owner.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
wilkinsona: We'd have to pass in both the `Path` and the `File` and introduce the risk of them not being the "same". We need the `Path` to check if `PosixFilePermissions` can be set. It's purely stylistic, but I prefer it as currently written. Your call, though.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
wilkinsona: `filePermissions` sounds good to me.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
wilkinsona: For symmetry, I'll do the same for `setExecutable` as well.

wilkinsona:
Thanks for the reviews. I've addressed the comments other than possibly passing the `File` into the `configurePermissions` and `assertPermissions` methods. Let me know which way you want to go there.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
showuon: I know we need both `File` and `Path` instance to complete the permission setting. What I meant is that:
1. We already had the `File` instances, i.e. baseDir, stateDir
2. `Paths.get(stateDir.getPath())`
--> what happened behind the scene, is we created a Path instance from the path string in the file instance (i.e.`stateDir.getPath()`)
```java
public final Path getPath(String var1, String... var2) {
... 
    return new UnixPath(this, var3);
}
```
3. `path.toFile()`
--> what happened behind the scene, is we created a File instance from the path string
```java
public final File toFile() {
        return new File(this.toString());
    }
```

So, currently, we will created an additional File instance via `path.toFile()`. If we pass the File instance into `configurePermissions()` directly, it'll only created a `Path` instance. ex: 
```java
configurePermissions(stateDir);  // pass the file instance directly
```
```java
private void configurePermissions(final File file) {
    final Path path = Paths.get(file.getPath()); // created path instance
    if (path.getFileSystem().supportedFileAttributeViews().contains("posix")) {
    ....
    }
    else {
        boolean set = file.setReadable(true, true);  // won't create additional File instance here
        ....
    }
}
```

wilkinsona: Sorry, I see what you mean now. I've pushed another commit that avoids the additional `File` creation.

streams/src/test/java/org/apache/kafka/streams/processor/internals/StateDirectoryTest.java
wilkinsona: Done.

wilkinsona:
I think all the review comments have now been addressed. Please let me know if you'd like the commits to be squashed into one.

showuon (approved):
LGTM. Thanks a lot.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
ableegoldman: Ah, I see. Thanks for the explanation

ableegoldman:
Don't worry about squashing all the commits yourself, we'll do that automatically when merging it.

Three test failures in the build, all unrelated:

```
org.apache.kafka.clients.consumer.KafkaConsumerTest.testCloseWithTimeUnit()
org.apache.kafka.clients.consumer.internals.FetcherTest.testEarlierOffsetResetArrivesLate()
org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationTest.testReplication()
```

ableegoldman (approved):
LGTM, thanks for the patch

ableegoldman:

ableegoldman:
I tried to cherrypick this back to the 2.6 branch but there was a merge conflict. I would be happy to review & merge another PR if you'd like to open one against the 2.6 branch, but realistically I think it's unlikely that we'll have a 2.6.2 release just given the lack of x.x.2 releases in recent history.

We can also just keep an eye out for a proposed 2.6.2 release and then work on another patch for the 2.6 branch if that does happen

wilkinsona:
Thanks for the merge.

Given that the regression in 2.6.1 makes Streams unusable for developers working on Windows, I really think that a 2.6.2 release is needed here. To that end, I've opened [a new pull request](https://github.com/apache/kafka/pull/9968) based on the 2.6 branch.

garyrussell:
As I [commented on the JIRA issue](https://issues.apache.org/jira/browse/KAFKA-12190?focusedCommentId=17267959&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17267959) I think it is important to get a 2.6.2 release out sooner rather than later.

ableegoldman:
@garyrussell I just merged the 2.6 PR so we'll be ready to go if/when the 2.6.2 release is started. To be honest I'm not sure how likely it is for a 2.6.2 release to be rushed out. It's a lot of work to roll out an AK release, even a bugfix. It might be difficult to convince someone to take all that on as a release manager in order to fix a regression that impacts a relatively small subset of users (this is not my personal philosophy, just my expectation).

That said, it's always worth asking, and the dev mailing list is probably the best way to get some attention.

garyrussell:
@ableegoldman I understand, but for the reasons I stated on the JIRA, this was a critical regression and some users have no choice regarding upgrading.

I have no personal ax to grind; I don't use Windows, but many developers have no choice there either.

I wonder if there could be a new mechanism to just release the kafka-clients, e.g. 2.6.1.1 ?

ableegoldman:
Yeah that's completely valid, but unfortunately it's not up to me 🙂 

I do think you should raise this on the mailing list -- maybe someone there will have an idea of how to proceed. I also think it's worth bringing up to get some discussion going on how we can avoid things like this from happening again, since testing on Windows is (obviously) not part of the current release process.

notme159:
@wilkinsona Hello thank you for the fix, I still have an issue, using KeyValueStore and getting
org.apache.kafka.streams.processor.internals.StateDirectory  - Failed to change permissions for the directory C:\Users\user1\AppData\Local\Temp\kafka-streams
when calling         streams = new KafkaStreams(topology, props);

Tried to debug and this always returns false ( set &= file.setWritable(true, true) )
Running win10, admin user, openJdk11 latest, Kafka 3.0.0
The folder has setup Full Control for user1/SYSTEM/Administrators/Everyone, I'm running out of ideas. Thanks for every tip.

aybefox:
@notme159 
Same here :( Did you find out something?

notme159:
@aybefox Well the folder permissions are correct (it is writable), it just logs error when the setWritable is called. So I just suppress the StateDirectory log in log4j config using FATAL level only for this class, not to show ERROR in log.
Of course it's not greatest work around as you might suppress another errors in that class.

BrokenWingsIcarus:
by casuse see [http ://bugs.sun.com/bugdatabase/view_bug.do?bug_id=6728842](http://bugs.sun.com/bugdatabase/view_bug.do?bug_id=6728842)

PR #9949, opened 2021-01-22
MINOR: replace FlattenedIterator by java stream

chia7712:
as title


clients/src/main/java/org/apache/kafka/common/requests/LeaderAndIsrRequest.java
```diff
@@ -183,8 +182,7 @@ public long brokerEpoch() {
 
     public Iterable<LeaderAndIsrPartitionState> partitionStates() {
         if (version() >= 2)
-            return () -> new FlattenedIterator<>(data.topicStates().iterator(),
-                topicState -> topicState.partitionStates().iterator());
+            return () -> data.topicStates().stream().flatMap(topicState -> topicState.partitionStates().stream()).iterator();
```
dajac, line 185: @chia7712 Is it also equivalent performance wise?

ijuma: Indeed, we would want to verify that. It didn't look like it was when I originally checked, but I may have missed something.

ijuma: We should be generally careful about replacing simple code paths with more complex ones (in terms of the underlying bytecode) in Kafka. Kafka is a high performance system, so code readability is one factor, but not the only one.

chia7712: @dajac @ijuma Thanks for your response. You are right. the performance should be more important than code readability. will close this PR :)

PR #4820, opened 2018-04-03
KAFKA-6684: Cast transform bytes

amitsela:
Allow to cast LogicalType to string by calling the serialized (Java) object's toString().   

Added tests for `BigDecimal` and `Date` as whole record and as fields. 

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9950, opened 2021-01-22
KAFKA-12170: Fix for Connect Cast SMT to correctly transform a Byte array into a string

sknop:
Cast SMT transformation for bytes -> string.
Without this fix, the conversion becomes

ByteBuffer.toString(), which always gives one of these useless results:
    "java.nio.HeapByteBuffer[pos=0 lim=4 cap=4]" 
    "[B@5ec0a365" (for byte array)
    

With this change, the byte array is converted into a hex string of the
byte buffer content, for example

    "FEDCBA9876543210"

Completed with test case and successfully tried out in a
real database conversion.

- [X ] Verify design and implementation 
- [ X] Verify test coverage and CI build status

### DIFF
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
@@ -361,6 +361,7 @@ public abstract class Cast<R extends ConnectRecord<R>> implements Transformation
             throw new DataException("Unexpected type in Cast transformation: " + value.getClass());
     }
 
+
     private static String castToString(Object value) {
         if (value instanceof java.util.Date) {
             java.util.Date dateValue = (java.util.Date) value;
@@ -369,15 +370,22 @@ public abstract class Cast<R extends ConnectRecord<R>> implements Transformation
         else if (value instanceof ByteBuffer) {
             ByteBuffer byteBuffer = (ByteBuffer) value;
 
-            StringBuilder sbuf = new StringBuilder();
-            for (byte b : byteBuffer.array()) {
-                sbuf.append(String.format("%02X", b));
-            }
-            return sbuf.toString();
+            return castByteArrayToString(byteBuffer.array());
+        }
+        else if (value.getClass() == byte[].class) {
+            return castByteArrayToString((byte[]) value);
         }
         else {
             return value.toString();
         }
+    }
+
+    private static String castByteArrayToString(byte[] array) {
+        StringBuilder sbuf = new StringBuilder();
+        for (byte b : array) {
+            sbuf.append(String.format("%02X", b));
+        }
+        return sbuf.toString();
     }
 
     protected abstract Schema operatingSchema(R record);
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
@@ -41,7 +41,13 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.nio.ByteBuffer;
-import java.util.*;
+import java.util.Base64;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
 
 import static org.apache.kafka.connect.transforms.util.Requirements.requireMap;
 import static org.apache.kafka.connect.transforms.util.Requirements.requireStruct;
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
@@ -59,7 +59,8 @@ public abstract class Cast<R extends ConnectRecord<R>> implements Transformation
     // allow casting nested fields.
     public static final String OVERVIEW_DOC =
             "Cast fields or the entire key or value to a specific type, e.g. to force an integer field to a smaller "
-                    + "width. Only simple primitive types are supported -- integers, floats, boolean, and string. "
+                    + "width. Cast from integers, floats, boolean and string to any other type, "
+                    + "and cast binary to string (base64 encoded)."
                     + "<p/>Use the concrete transformation type designed for the record key (<code>" + Key.class.getName() + "</code>) "
                     + "or value (<code>" + Value.class.getName() + "</code>).";
 
@@ -85,7 +86,7 @@ public abstract class Cast<R extends ConnectRecord<R>> implements Transformation
             ConfigDef.Importance.HIGH,
             "List of fields and the type to cast them to of the form field1:type,field2:type to cast fields of "
                     + "Maps or Structs. A single type to cast the entire value. Valid types are int8, int16, int32, "
-                    + "int64, float32, float64, boolean, and string.");
+                    + "int64, float32, float64, boolean, and string. Note that binary fields can only be cast to string.");
 
     private static final String PURPOSE = "cast types";
 
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
@@ -22,6 +22,7 @@ import org.apache.kafka.common.cache.LRUCache;
 import org.apache.kafka.common.cache.SynchronizedCache;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
+import org.apache.kafka.common.utils.Utils;
 import org.apache.kafka.connect.connector.ConnectRecord;
 import org.apache.kafka.connect.data.ConnectSchema;
 import org.apache.kafka.connect.data.Date;
@@ -367,16 +368,10 @@ public abstract class Cast<R extends ConnectRecord<R>> implements Transformation
             return Values.dateFormatFor(dateValue).format(dateValue);
         } else if (value instanceof ByteBuffer) {
             ByteBuffer byteBuffer = (ByteBuffer) value;
-            if (byteBuffer.hasArray()) {
-                return castByteArrayToString(byteBuffer.array());
-            }
-            else {
-                byte[] array = new byte[byteBuffer.remaining()];
-                byteBuffer.get(array);
-                return castByteArrayToString(array);
-            }
+            return castByteArrayToString(Utils.readBytes(byteBuffer));
         } else if (value instanceof byte[]) {
-            return castByteArrayToString((byte[]) value);
+            byte[] rawBytes = (byte[]) value;
+            return castByteArrayToString(rawBytes);
         } else {
             return value.toString();
         }
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
@@ -367,7 +367,14 @@ public abstract class Cast<R extends ConnectRecord<R>> implements Transformation
             return Values.dateFormatFor(dateValue).format(dateValue);
         } else if (value instanceof ByteBuffer) {
             ByteBuffer byteBuffer = (ByteBuffer) value;
-            return castByteArrayToString(byteBuffer.array());
+            if (byteBuffer.hasArray()) {
+                return castByteArrayToString(byteBuffer.array());
+            }
+            else {
+                byte[] array = new byte[byteBuffer.remaining()];
+                byteBuffer.get(array);
+                return castByteArrayToString(array);
+            }
         } else if (value instanceof byte[]) {
             return castByteArrayToString((byte[]) value);
         } else {
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
@@ -41,12 +41,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.nio.ByteBuffer;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 import static org.apache.kafka.connect.transforms.util.Requirements.requireMap;
 import static org.apache.kafka.connect.transforms.util.Requirements.requireStruct;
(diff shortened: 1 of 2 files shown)


g1geordie:
@sknop  hello , It seems like there are some style error .
You can use `./gradlew checkstyleMain checkstyleTest` to cehck .

sknop:
> @sknop hello , It seems like there are some style error .
> You can use `./gradlew checkstyleMain checkstyleTest` to cehck .

Thank you @g1geordie for your feedback. I have adjusted the code and it now passes the style check.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
```diff
@@ -360,15 +361,30 @@ else if (value instanceof String)
             throw new DataException("Unexpected type in Cast transformation: " + value.getClass());
     }
 
+
     private static String castToString(Object value) {
         if (value instanceof java.util.Date) {
             java.util.Date dateValue = (java.util.Date) value;
             return Values.dateFormatFor(dateValue).format(dateValue);
+        } else if (value instanceof ByteBuffer) {
+            ByteBuffer byteBuffer = (ByteBuffer) value;
```
mimaison, line 372: We can inline this like you've done for the byte[] case below

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
```diff
@@ -360,15 +361,30 @@ else if (value instanceof String)
...
+
+            return castByteArrayToString(byteBuffer.array());
+        } else if (value.getClass() == byte[].class) {
```
mimaison: Let's use `value instanceof byte[]` so it's like the other cases

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
```diff
@@ -360,15 +361,30 @@ else if (value instanceof String)
             throw new DataException("Unexpected type in Cast transformation: " + value.getClass());
     }
 
+
```
mimaison: We don't need this extra blank line

connect/transforms/src/test/java/org/apache/kafka/connect/transforms/CastTest.java
```diff
@@ -425,7 +426,21 @@ public void castLogicalToString() {
...
+        byteBuffer.put((byte) 0xDC);
+        byteBuffer.put((byte) 0xBA);
+        byteBuffer.put((byte) 0x98);
+        byteBuffer.put((byte) 0x76);
+        byteBuffer.put((byte) 0x54);
+        byteBuffer.put((byte) 0x32);
+        byteBuffer.put((byte) 0x10);
+        byteBuffer.flip();
+
+        byte[] byteArray = Arrays.copyOf(byteBuffer.array(), byteBuffer.array().length);
```
mimaison: What about doing something like:
```java
byte[] byteArray = new byte[] {(byte) 0xFE, (byte) 0xDC, (byte) 0xBA, (byte) 0x98, 0x76, 0x54, 0x32, 0x10};
ByteBuffer byteBuffer = ByteBuffer.wrap(Arrays.copyOf(byteArray, byteArray.length));
```

mimaison (changes requested):
Thanks for the PR!
It looks good overall, I just left a few minor comments

sknop:
Hi Mickael

Thank you so much for your code review. I have adjusted the code according to your suggestions, they all make sense to me.

Cheers

Sven Erik
seknop@gmail.com

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
```diff
@@ -364,11 +365,24 @@ private static String castToString(Object value) {
         if (value instanceof java.util.Date) {
             java.util.Date dateValue = (java.util.Date) value;
             return Values.dateFormatFor(dateValue).format(dateValue);
+        } else if (value instanceof ByteBuffer) {
+            ByteBuffer byteBuffer = (ByteBuffer) value;
+            return castByteArrayToString(byteBuffer.array());
```
kkonstantine: formally we should almost always check that a `ByteBuffer` is indeed backed by an array with `hasArray` or this part my throw exceptions now: https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#array--

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
```diff
@@ -364,11 +365,24 @@ private static String castToString(Object value) {
...
+            return castByteArrayToString((byte[]) value);
         } else {
             return value.toString();
         }
     }
 
+    private static String castByteArrayToString(byte[] array) {
+        StringBuilder sbuf = new StringBuilder();
+        for (byte b : array) {
+            sbuf.append(String.format("%02X", b));
```
kkonstantine: Also, I'm a bit curious. Why are we selecting hex here and not base64 for example?
Hex of course is less efficient. 
https://issues.apache.org/jira/browse/KAFKA-12170 does not mention anything about the rational, so I thought I'd ask. 

@rhauch wdyt about the representation here?

kkonstantine (commented):
Thanks for the PR @sknop 
Left a few comments inline.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
sknop: That is a good point, but what should happen in that case? 

I assume we start with hasArray() and only process the content if an array is available, as I understand it? ByteBuffer.remaining() would contain the number of bytes available ... I'll go and write some test cases to understand this better.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
sknop: The reason I suggested Hex representation was that in the use case I encountered - database PK field encoded as BINARY retrieved via CDC Source Connector - the representation in the SQL CLI was Hex, so choosing the same format made it easy to compare source and target.

I would suggest that base64 encoding would validate its own dedicated transformer, with an ability to choose padding, for example

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
sknop: I added code that checks for the existence of the array and otherwise, I extract the content through ByteBuffer.get(byte[] array).

rhauch:
Before this fix, the changes made in #4820 ([KAFKA-6684](https://issues.apache.org/jira/browse/KAFKA-6684)) resulted in `toString()` being called on `byte[]` and `ByteBuffer`. As highlighted in this PR and issue that `ByteBuffer.toString()` is not useful, but the `toString()` on `byte[]` still works. This PR seems to change that behavior, which would not be backward compatible.

The discussion on PR #4820 also talked about making this compatible with `Values.convertToString(...)`, which for `byte[]` and `ByteBuffer` results in a base 64 encoded string (with `ISO-8859-1` encoding). See the [Values code for details](https://github.com/apache/kafka/blob/95f51539c8d0b88bd7f285011d42e2d1117107de/connect/api/src/main/java/org/apache/kafka/connect/data/Values.java#L673-L682).

~Because of this, it seems much more sensible to also use base64 here for `ByteBuffer` so it matches the existing behavior with `byte[]`.~

Actually, the existing code simply outputs the `toString()` result of a `byte[]` object, which is of the form `[B@22ef9844`, which is the `Object.toString()` implementation that includes the class alias (e.g., `[B` is a byte array) and the hex representation of the object's `hashCode()`.

However, given that none of the cast forms of `byte[]` or `ByteBuffer[]` are useful, then perhaps it's worth adding the previous discussion on #4820 mentioning that:
> As @ewencp suggested, for consistency perhaps we can use the same string formats as SimpleHeaderConverter: https://github.com/apache/kafka/blob/trunk/connect/api/src/main/java/org/apache/kafka/connect/data/Values.java,

As mentioned above, the `Values.convertToString(...)` uses a base 64 encoding. And this aligns with @kkonstantine's suggestion:
> Why are we selecting hex here and not base64 for example? Hex of course is less efficient.

I agree that it maybe doesn't suffice in all user situations, so if we also want to support other encodings we'd need other config changes that will require using the KIP mechanism to propose such enhancements.

WDYT?

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
```diff
@@ -364,11 +365,31 @@ private static String castToString(Object value) {
...
+        } else if (value instanceof ByteBuffer) {
+            ByteBuffer byteBuffer = (ByteBuffer) value;
+            if (byteBuffer.hasArray()) {
+                return castByteArrayToString(byteBuffer.array());
+            }
+            else {
+                byte[] array = new byte[byteBuffer.remaining()];
+                byteBuffer.get(array);
+                return castByteArrayToString(array);
+            }
```
rhauch: The use of `ByteBuffer.get(...)` here does not account for the fact that it may not be positioned at the beginning. Kafka has two `Utils.readBytes(...)` methods that we should probably use. This code would then simplify to:
```suggestion
            byte[] rawBytes = Utils.readBytes(byteBuffer);
            return castByteArrayToString(rawBytes);
```

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
rhauch: As noted in my previous comment, I agree with @kkonstantine that base64 would be preferable. Doing that would align better with the existing `Values` class used in Connect's header converter mechanism.

If we want to add support for multiple encodings, we would need to have a KIP since it would likely mean changing the SMT configuration.

rhauch (changes requested):

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
sknop: That sounds like an excellent plan, I did not know this class existed. Changed.

sknop: This sounds like an excellent plan and passes the tests.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
sknop: If base64 is the better choice we shall go with that. Would we need to adjust the JIRA ticket to reflect this?

I just ran a little test in the JShell:

jshell> byte[] byteArray = new byte[] {(byte) 0xFE, (byte) 0xDC, (byte) 0xBA, (byte) 0x98, 0x76, 0x54, 0x32, 0x10};
byteArray ==> byte[8] { -2, -36, -70, -104, 118, 84, 50, 16 }

jshell> Base64.getEncoder().encode(byteArray)
$2 ==> byte[12] { 47, 116, 121, 54, 109, 72, 90, 85, 77, 104, 65, 61 }

jshell> Base64.getEncoder().encodeToString(byteArray)
$3 ==> "/ty6mHZUMhA="

jshell> Base64.getEncoder().withoutPadding().encodeToString(byteArray)
$4 ==> "/ty6mHZUMhA"

Would you suggest using padding or not?

rhauch: Connect's `Values` code just uses `Base64.getEncoder().encodeToString((byte[]) value)` and does not disable padding. Seems like we should follow the same pattern here. Thoughts, @kkonstantine ?

kkonstantine: Yes, I agree. `encodeToString` is the method I had in mind. Used in `Values` and JsonConverter for a conversion of bytes to string. It's also more efficient.

kkonstantine (commented):
Thanks for the detailed comments @rhauch 
@sknop I agree with @rhauch that base64 would be preferable and consistent with our previous choices. I also agree to punt support of multiple encodings for now.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
```diff
@@ -39,12 +40,8 @@
...
 import org.slf4j.LoggerFactory;
 
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
+import java.nio.ByteBuffer;
+import java.util.*;
```
rhauch: Wildcard imports should be caught by checkstyle, and should fail the build. In any case, please replace with non-wildcard imports.

rhauch (commented):
One import-related fix, but otherwise looks good.

connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java
sknop: IntelliJ did this automatically for me - I have changed the code accordingly.

sknop:
What about documentation? 

Where is the documentation stored and what is the process to change it? It might be useful to point out that the Cast SMT will create base64 strings for byte arrays.

rhauch:
The documentation is auto-generated from the transforms code, via [TransformationDoc](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/tools/TransformationDoc.java). Since this is an existing SMT, there are no changes that that `TransformationDoc` class, but we still have to change the `Cast` class [here](https://github.com/apache/kafka/blob/958f90e710f0de164dce1dda5f45d75a8b8fb8d4/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java#L57-L61) and [here](https://github.com/apache/kafka/blob/958f90e710f0de164dce1dda5f45d75a8b8fb8d4/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Cast.java#L83-L85) to mention that byte arrays can be cast to strings.

sknop:
Thank you @rhauch. Just to make sure I understand this correctly - the types mentioned in the cast string are the _target_ casts, not the source tasks. The current code does not allow casting to a byte array, only casting _from_ a byte array.
As I understand it, therefore no documentation changes are required (except maybe for release notes)?

rhauch:
If we don't clarify the documentation, then I think users will be very confused. Can you take a stab at improving/expanding the documentation a bit to clarify the input and output types?

sknop:
Yes of course, totally agree.

rhauch (approved):
Thanks, @sknop! Looks great.

@mimaison, you have a "request changes", so can you please review? Thanks!

mimaison (approved):
Thanks for the PR. LGTM

kkonstantine (approved):
This LGTM
Thanks @sknop

mimaison:
Test failures are not related:
```
JDK 8 / org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationSSLTest.testOneWayReplicationWithAutoOffsetSync()
JDK 11 / org.apache.kafka.connect.integration.RebalanceSourceConnectorsIntegrationTest.testMultipleWorkersRejoining
```

Merging to trunk

PR #9952, opened 2021-01-23
MINOR: A few small group coordinator cleanups

hachikuji:
A few small cleanups in `GroupCoordinator` and related classes.

- Remove redundant `groupId` field from `MemberMetadata`
- Remove redundant `isStaticMember` field from `MemberMetadata`
- Fix broken log message in `GroupMetadata.loadGroup` and apply it to all loaded members
- Remove ancient TODOs and no-op methods from `GroupCoordinator`
- Move removal of static members into `GroupMetadata.remove`

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9958, opened 2021-01-24
MINOR: Clean up group instance id handling in `GroupCoordinator`

hachikuji:
This is a continuation of a refactor started in #9952. The logic in `GroupCoordinator` is loose and inconsistent in the handling of the `groupInstanceId`. In some cases, such as in the JoinGroup hander, we verify that the groupInstanceId from the request is mapped to the memberId precisely. In other cases, such as Heartbeat, we check the mapping, but only to validate fencing. The patch consolidates the member validation so that all handlers follow the same logic. 

A second problem is that many of the APIs where a `groupInstanceId` is expected use optional arguments. For example:
```scala
def hasStaticMember(groupInstanceId: Option[String]): Boolean

def addStaticMember(groupInstanceId: Option[String], newMemberId: String): Unit
```
If `groupInstanceId` is `None`, then `hasStaticMember` is guaranteed to return `false` while `addStaticMember` raises an `IllegalStateException`. So the APIs suggest a generality which is not supported and does not make sense.

Finally,  the patch attempts to introduce stronger internal  invariants inside `GroupMetadata`. Currently it is possible for an inconsistent `groupInstanceId` to `memberId` mapping to exist because we expose separate APIs to modify `members` and `staticMembers`. We rely on the caller to ensure this doesn't happen.  Similarly, it is possible for a member to be in the `pendingMembers` set as well as the stable `members` map. The patch fixes this by consolidating the paths to addition and removal from these collections and adding assertions to ensure that invariants are maintained. 

### DIFF
--- a/core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
+++ b/core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
@@ -404,7 +404,7 @@ class GroupCoordinator(val brokerId: Int,
         )
 
         memberErrorOpt match {
-          case Some(error) => JoinGroupResult(memberId, error)
+          case Some(error) => responseCallback(JoinGroupResult(memberId, error))
 
           case None => group.currentState match {
             case PreparingRebalance =>
@@ -938,7 +938,7 @@ class GroupCoordinator(val brokerId: Int,
         group,
         memberId,
         groupInstanceId,
-        operation = "offset-commit"
+        operation = if (isTransactional) "txn-offset-commit" else "offset-commit"
       ).orElse {
         if (generationId != group.generationId) {
           Some(Errors.ILLEGAL_GENERATION)
--- a/core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
+++ b/core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
@@ -547,7 +547,7 @@ class GroupCoordinator(val brokerId: Int,
       )
 
       validationErrorOpt match {
-        case Some(error) =>  responseCallback(SyncGroupResult(error))
+        case Some(error) => responseCallback(SyncGroupResult(error))
 
         case None => group.currentState match {
           case Empty =>
@@ -778,7 +778,7 @@ class GroupCoordinator(val brokerId: Int,
         group,
         memberId,
         groupInstanceId,
-        operation = "sync-group"
+        operation = "heartbeat"
       ).orElse {
         if (generationId != group.generationId) {
           Some(Errors.ILLEGAL_GENERATION)
--- a/core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
+++ b/core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
@@ -184,9 +184,9 @@ class GroupCoordinator(val brokerId: Int,
               group.remove(memberId)
               responseCallback(JoinGroupResult(JoinGroupRequest.UNKNOWN_MEMBER_ID, Errors.GROUP_MAX_SIZE_REACHED))
             } else if (isUnknownMember) {
-              doUnknownJoinGroup(group, groupInstanceId, requireKnownMemberId, clientId, clientHost, rebalanceTimeoutMs, sessionTimeoutMs, protocolType, protocols, responseCallback)
+              doNewMemberJoinGroup(group, groupInstanceId, requireKnownMemberId, clientId, clientHost, rebalanceTimeoutMs, sessionTimeoutMs, protocolType, protocols, responseCallback)
             } else {
-              doJoinGroup(group, memberId, groupInstanceId, clientId, clientHost, rebalanceTimeoutMs, sessionTimeoutMs, protocolType, protocols, responseCallback)
+              doCurrentMemberJoinGroup(group, memberId, groupInstanceId, clientId, clientHost, rebalanceTimeoutMs, sessionTimeoutMs, protocolType, protocols, responseCallback)
             }
 
             // attempt to complete JoinGroup
@@ -198,16 +198,18 @@ class GroupCoordinator(val brokerId: Int,
     }
   }
 
-  private def doUnknownJoinGroup(group: GroupMetadata,
-                                 groupInstanceId: Option[String],
-                                 requireKnownMemberId: Boolean,
-                                 clientId: String,
-                                 clientHost: String,
-                                 rebalanceTimeoutMs: Int,
-                                 sessionTimeoutMs: Int,
-                                 protocolType: String,
-                                 protocols: List[(String, Array[Byte])],
-                                 responseCallback: JoinCallback): Unit = {
+  private def doNewMemberJoinGroup(
+    group: GroupMetadata,
+    groupInstanceId: Option[String],
+    requireKnownMemberId: Boolean,
+    clientId: String,
+    clientHost: String,
+    rebalanceTimeoutMs: Int,
+    sessionTimeoutMs: Int,
+    protocolType: String,
+    protocols: List[(String, Array[Byte])],
+    responseCallback: JoinCallback
+  ): Unit = {
     group.inLock {
       if (group.is(Dead)) {
         // if the group is marked as dead, it means some other thread has just removed the group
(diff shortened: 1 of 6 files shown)


hachikuji:
cc @dajac No hurry, but if you have time, please take a look.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -185,9 +185,9 @@ class GroupCoordinator(val brokerId: Int,
               group.remove(memberId)
               responseCallback(JoinGroupResult(JoinGroupRequest.UNKNOWN_MEMBER_ID, Errors.GROUP_MAX_SIZE_REACHED))
             } else if (isUnknownMember) {
-              doUnknownJoinGroup(group, groupInstanceId, requireKnownMemberId, clientId, clientHost, rebalanceTimeoutMs, sessionTimeoutMs, protocolType, protocols, responseCallback)
+              doNewMemberJoinGroup(group, groupInstanceId, requireKnownMemberId, clientId, clientHost, rebalanceTimeoutMs, sessionTimeoutMs, protocolType, protocols, responseCallback)
```
dajac: nit:  Could we break this long line and the one below as well?

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -220,67 +222,163 @@ class GroupCoordinator(val brokerId: Int,
...
+  private def validateCurrentMember(
+    group: GroupMetadata,
+    memberId: String,
+    groupInstanceId: Option[String],
+    operation: String
+  ): Option[Errors] = {
+    // We are validating two things:
+    // 1. If `groupInstanceId` is present, then it exists and is mapped to `memberId`
+    // 2. The `memberId` exists in the group
+
```
dajac: nit: This empty line could be removed.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -470,33 +600,31 @@ class GroupCoordinator(val brokerId: Int,
...
-                        s"from pending member bucket with member id $memberId")
+                    removePendingMemberAndUpdateGroup(group, memberId)
+                    heartbeatPurgatory.checkAndComplete(MemberKey(group.groupId, memberId))
+                    info(s"Pending member with memberId=$memberId has left group ${group.groupId} " +
+                      s"through explicit `LeaveGroup` request")
+                    memberLeaveError(leavingMember, Errors.NONE)
+                  } else {
+                    val memberErrorOpt = validateCurrentMember(group, memberId, groupInstanceId, "leave-group")
+                    if (memberErrorOpt.isDefined) {
+                      memberLeaveError(leavingMember, memberErrorOpt.get)
```
dajac: nit: It might be better to use `match` here instead of using `isDefined` and `get` on `memberErrorOpt`.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -603,6 +731,25 @@ class GroupCoordinator(val brokerId: Int,
...
+  private def validateHeartbeat(
+    group: GroupMetadata,
+    generationId: Int,
+    memberId: String,
+    groupInstanceId: Option[String]
+  ): Option[Errors] = {
+    if (group.is(Dead)) {
+      Some(Errors.COORDINATOR_NOT_AVAILABLE)
+    } else {
+      validateCurrentMember(group, memberId, groupInstanceId, "sync-group").orElse {
```
dajac: `heartbeat` instead of `sync-group` should be used here.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -725,50 +890,67 @@ class GroupCoordinator(val brokerId: Int,
...
-        responseCallback(offsetMetadata.map { case (k, _) => k -> Errors.ILLEGAL_GENERATION })
+      val validationErrorOpt = validateTxnOffsetCommit(
+        group,
+        generationId,
+        memberId,
+        groupInstanceId
+      )
+
+      if (validationErrorOpt.isDefined) {
+        responseCallback(offsetMetadata.map { case (k, _) => k -> validationErrorOpt.get })
```
dajac, line 920: nit: I would use `match` here as well.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -725,50 +890,67 @@ class GroupCoordinator(val brokerId: Int,
...
+    if (group.is(Dead)) {
+      Some(Errors.COORDINATOR_NOT_AVAILABLE)
+    } else if (generationId >= 0 || memberId != JoinGroupRequest.UNKNOWN_MEMBER_ID || groupInstanceId.isDefined) {
+      validateCurrentMember(group, memberId, groupInstanceId, "offset-commit").orElse {
+        if (generationId != group.generationId) {
+          Some(Errors.ILLEGAL_GENERATION)
+        } else {
+          None
+        }
+      }
```
dajac: This block is similar to the one in `validateTxnOffsetCommit` module the reason. I wonder if we could define a common helper method for both cases.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -725,50 +890,67 @@ class GroupCoordinator(val brokerId: Int,
...
-        responseCallback(offsetMetadata.map { case (k, _) => k -> Errors.ILLEGAL_GENERATION })
+      val validationErrorOpt = validateOffsetCommit(
+        group,
+        generationId,
+        memberId,
+        groupInstanceId
+      )
+
+      if (validationErrorOpt.isDefined) {
+        responseCallback(offsetMetadata.map { case (k, _) => k -> validationErrorOpt.get })
```
dajac, line 976: nit: I would use `match` here as well.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -1017,26 +1198,17 @@ class GroupCoordinator(val brokerId: Int,
...
-    } else {
-      group.removePendingMember(memberId)
-    }
     maybePrepareRebalance(group, s"Adding new member $memberId with group instance id $groupInstanceId")
   }
 
   private def updateStaticMemberAndRebalance(group: GroupMetadata,
                                              newMemberId: String,
-                                             groupInstanceId: Option[String],
+                                             oldMemberId: String,
```
dajac, line 1233: nit: I would put `oldMemberId` before `newMemberId`. It feels a bit more natural when reading it.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -1055,11 +1227,15 @@ class GroupCoordinator(val brokerId: Int,
           val groupAssignment: Map[String, Array[Byte]] = group.allMemberMetadata.map(member => member.memberId -> member.assignment).toMap
           groupManager.storeGroup(group, groupAssignment, error => {
             if (error != Errors.NONE) {
+
+              // TODO: This logic seems questionable. The write was not committed, but that doesn't
+              //  mean it wasn't written to the log and cannot eventually become committed.
```
dajac: I do agree. Should we turn this onto a Jira and address it in a follow up PR?

core/src/main/scala/kafka/coordinator/group/GroupMetadata.scala
```diff
@@ -247,11 +250,15 @@ private[group] class GroupMetadata(val groupId: String, initialState: GroupState
 
     if (leaderId.isEmpty)
       leaderId = Some(member.memberId)
+
     members.put(member.memberId, member)
     member.supportedProtocols.foreach{ case (protocol, _) => supportedProtocols(protocol) += 1 }
```
dajac: nit: Not related to your PR but a space is missing before `{`.

dajac (commented):
@hachikuji Great refactoring! It really improve the readability of the code. It seems to me that all the invariants are preserved so it looks good overall. I have left a couple of minor comments. I need to make a second pass over it.

dajac:
Failed tests are not related. Merging to trunk.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
hachikuji: Yeah, I agree. I think I was trying to avoid a big indent of the block below. I found a reasonable way to do it in the latest commit.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
hachikuji: Done. I filed https://issues.apache.org/jira/browse/KAFKA-12363.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -378,24 +537,19 @@ class GroupCoordinator(val brokerId: Int,
...
+        group,
+        generationId,
+        memberId,
+        protocolType,
+        protocolName,
+        groupInstanceId
+      )
+
+      validationErrorOpt match {
+        case Some(error) =>  responseCallback(SyncGroupResult(error))
```
dajac: nit: There is an extra space before `responseCallback`.

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
```diff
@@ -602,6 +765,30 @@ class GroupCoordinator(val brokerId: Int,
...
+    groupInstanceId: Option[String]
+  ): Option[Errors] = {
+    if (group.is(Dead)) {
+      Some(Errors.COORDINATOR_NOT_AVAILABLE)
+    } else {
+      validateCurrentMember(
+        group,
+        memberId,
+        groupInstanceId,
+        operation = "sync-group"
```
dajac: `sync-group` => `heartbeat`?

core/src/main/scala/kafka/coordinator/group/GroupCoordinator.scala
dajac: Ah.. I see your point regarding the indentation. That makes sense to avoid them. I do not see the reasonable way that you are referring to though.

core/src/main/scala/kafka/coordinator/group/GroupMetadata.scala
```diff
@@ -430,7 +429,7 @@ private[group] class GroupMetadata(val groupId: String, initialState: GroupState
   private def candidateProtocols: Set[String] = {
     // get the set of protocols that are commonly supported by all members
     val numMembers = members.size
-    supportedProtocols.filter(_._2 == numMembers).map(_._1).toSet
+    supportedProtocols.filter(_._2 == numMembers).keys.toSet
```
dajac, line 432: nit: Could we directly use `keysSet` instead of `.keys.toSet`?

core/src/main/scala/kafka/coordinator/group/GroupMetadata.scala
```diff
@@ -247,16 +250,20 @@ private[group] class GroupMetadata(val groupId: String, initialState: GroupState
...
     if (member.isAwaitingJoin)
       numMembersAwaitingJoin += 1
+
+    pendingMembers.remove(member.memberId)
   }
 
   def remove(memberId: String): Unit = {
     members.remove(memberId).foreach { member =>
-      member.supportedProtocols.foreach{ case (protocol, _) => supportedProtocols(protocol) -= 1 }
+      member.supportedProtocols.foreach { case (protocol, _) => supportedProtocols(protocol) -= 1 }
```
dajac, line 266: I have noticed that we update `supportedProtocols` in multiple places with the exact same line (modulo +/-). I think that we should consolidate them into one or two helper methods. We can address this separately though. I just wanted to point that out.

dajac (commented):
@hachikuji Thanks for the updates. I went through the logic again and it looks good to me. I have left few minor comments.

core/src/main/scala/kafka/coordinator/group/GroupMetadata.scala
hachikuji: I tried this, but I think we still need the `toSet` because we are converting to the immutable `Set` type.

core/src/main/scala/kafka/coordinator/group/GroupMetadata.scala
hachikuji: Yeah, I agree. I tried a few options for this, but I could not find one I felt satisfied with. If you have a good suggestion, I'm happy to give it a shot here or to help review if you want to do a separate PR.

dajac (approved):
LGTM

PR #9959, opened 2021-01-25
MINOR: Remove redundant casting and if condition from ConnectSchema

g1geordie:
as title


chia7712 (approved):

PR #9960, opened 2021-01-25
MINOR: Remove `toStruct` and `fromStruct` methods from generated protocol classes

ijuma:
Update few classes that were still using the removed methods (including
tests that are no longer required).

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/security/authenticator/DefaultKafkaPrincipalBuilder.java
+++ b/clients/src/main/java/org/apache/kafka/common/security/authenticator/DefaultKafkaPrincipalBuilder.java
@@ -24,7 +24,8 @@ import org.apache.kafka.common.errors.SerializationException;
 import org.apache.kafka.common.message.DefaultPrincipalData;
 import org.apache.kafka.common.network.Authenticator;
 import org.apache.kafka.common.network.TransportLayer;
-import org.apache.kafka.common.protocol.types.Struct;
+import org.apache.kafka.common.protocol.ByteBufferAccessor;
+import org.apache.kafka.common.protocol.MessageUtil;
 import org.apache.kafka.common.security.auth.AuthenticationContext;
 import org.apache.kafka.common.security.auth.KafkaPrincipal;
 import org.apache.kafka.common.security.auth.KafkaPrincipalBuilder;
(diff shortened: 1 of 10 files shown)


clients/src/main/java/org/apache/kafka/common/security/authenticator/DefaultKafkaPrincipalBuilder.java
```diff
@@ -188,9 +185,7 @@ public KafkaPrincipal deserialize(byte[] bytes) {
             throw new SerializationException("Invalid principal data version " + version);
```
chia7712, line 185: Could we use ```HIGHEST_SUPPORTED_VERSION``` and ```LOWEST_SUPPORTED_VERSION``` to replace origin condition ("0" and ```DefaultPrincipalData.SCHEMAS.length```) ?

chia7712 (approved):
@ijuma Nice cleanup!

clients/src/main/java/org/apache/kafka/common/security/authenticator/DefaultKafkaPrincipalBuilder.java
ijuma: Good point, addressed.

chia7712 (approved):

clients/src/test/java/org/apache/kafka/common/requests/OffsetFetchResponseTest.java
```diff
@@ -107,20 +107,21 @@ public void testStructBuild() {
...
         for (short version = 0; version <= ApiKeys.OFFSET_FETCH.latestVersion(); version++) {
-            Struct struct = latestResponse.data().toStruct(version);
+            OffsetFetchResponseData data = new OffsetFetchResponseData(
+                new ByteBufferAccessor(latestResponse.serialize(version)), version);
 
-            OffsetFetchResponse oldResponse = OffsetFetchResponse.parse(latestResponse.serialize(version), version);
+            OffsetFetchResponse oldResponse = new OffsetFetchResponse(data, version);
 
             if (version <= 1) {
-                assertFalse(struct.hasField(ERROR_CODE));
```
chia7712, line 115: the field "ERROR_CODE" is never used.

clients/src/main/resources/common/message/README.md
```diff
@@ -157,9 +157,6 @@ meaning of the request, such as a "validateOnly" boolean, should not be ignored.
 It's often useful to know how much space a message will take up before writing
 it out to a buffer.  You can find this out by calling the Message#size method.
 
-You can also convert a message to a Struct by calling Message#toStruct.  This
-allows you to use the functions that serialize Structs to buffers.
-
 Deserializing Messages
 ----------------------
 Message objects may be deserialized using the Message#read method.  This method
```
chia7712, line 162: There is another invalid description.

```
You can also deserialize a message from a Struct by calling Message#fromStruct.
The Struct will not be modified.
```

dengziming:
This is nice, maybe we will also remove `struct.writeTo` and `struct.read` after #9641  #9766  and KAFKA-10863 is accomplished.

ijuma:
@dengziming Looks like `StickyAssignor` also uses `Struct` still. Once all usages are removed, maybe we can delete `Struct` altogether instead of only some methods in it.

ijuma:
Hmm, we may need to keep `Struct` as a type for `Schema`, however. So you may be right that we can only remove the serialization/deserialization methods.

dengziming:
Yes, #9766 is trying to replace Struct in`StickyAssignor`. I am not sure about removing Struct class, I will try to check it, but we can remove serialization/deserialization methods.

ijuma:
JDK 11 build passed, JDK 8 and 15 had unrelated flaky failures.

PR #9963, opened 2021-01-25
MINOR: Extract ApiVersions logic from the `SocketServer` to the `KafkaApis`

dajac:
When we implement KIP-511, we put the logic to extract the `SoftwareName` and the `SoftwareVersion` from the `ApiVersionsRequest` in the `SocketServer` to avoid having to wire the `ChannelMetadataRegistry` up to the `KafkaApis` layer. In retrospect, this was not a good idea because it spreads the logic to handle that request in multiple places. This patch is an attempt to move that logic to `KafkaApis`.


purplefox:
LGTM!

dajac:
@hachikuji Could you take a look at this one when you have a bit of time? It is not urgent.

dajac:
Closing.

PR #9964, opened 2021-01-25
MINOR: remove duplicate code of serializing auto-generated data

chia7712:
as title

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfo.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfo.java
@@ -24,7 +24,7 @@ import java.util.Map;
 
 import org.apache.kafka.common.Uuid;
 import org.apache.kafka.common.protocol.ByteBufferAccessor;
-import org.apache.kafka.common.protocol.ObjectSerializationCache;
+import org.apache.kafka.common.protocol.MessageUtil;
 import org.apache.kafka.streams.errors.TaskAssignmentException;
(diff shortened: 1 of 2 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/assignment/SubscriptionInfo.java
```diff
@@ -265,14 +265,7 @@ public ByteBuffer encode() {
                 "Should never try to encode a SubscriptionInfo with version [" +
                     data.version() + "] > LATEST_SUPPORTED_VERSION [" + LATEST_SUPPORTED_VERSION + "]"
             );
-        } else {
-            final ObjectSerializationCache cache = new ObjectSerializationCache();
-            final ByteBuffer buffer = ByteBuffer.allocate(data.size(cache, (short) data.version()));
-            final ByteBufferAccessor accessor = new ByteBufferAccessor(buffer);
-            data.write(accessor, cache, (short) data.version());
-            buffer.rewind();
```
dajac, line 273: `MessageUtil.toByteBuffer` uses `flip` whereas `rewind` is used here. Does it make any difference in this particular case?

chia7712: In this case, the state of ```flip``` is coincide with ```rewind```. This buffer is filled so the ```limit``` is always equal to ```capability``` even though ```rewind``` does not reset ```limit```.

dajac (approved):
LGTM

chia7712:
the flay is traced by https://issues.apache.org/jira/browse/KAFKA-12245

PR #9966, opened 2021-01-25
MINOR: Fix visibility of Log.{unflushedMessages, addSegment} methods

kowshik:
I've changed the visibility of the following methods, since these need not be `public` methods:
 * The method `Log.unflushedMessages` from public to private.
 * The method `Log.addSegment` from public to package private.

**Test plan:**
Rely on existing tests.


kowshik:
cc @junrao for review

junrao (approved):
@kowshik : Thanks for the PR. LGTM. Waiting for the tests to pass.

chia7712 (approved):

PR #9967, opened 2021-01-25
KAFKA-12236; New meta.properties logic for KIP-500

hachikuji:
This patch contains the new handling of `meta.properties` required by the KIP-500 server as specified in KIP-631. When using the self-managed quorum, the `meta.properties` file is required in each log directory with the new `version` property set to 1. It must include the `cluster.id` property and it must have a `node` matching that in the configuration.

The behavior of `meta.properties` for the Zookeeper-based `KafkaServer` does not change. We treat `meta.properties` as optional and as if it were `version=0`. We continue to generate the clusterId and/or the brokerId through Zookeeper as needed.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/Uuid.java
+++ b/clients/src/main/java/org/apache/kafka/common/Uuid.java
@@ -105,10 +105,18 @@ public class Uuid implements Comparable<Uuid> {
      * Creates a UUID based on a base64 string encoding used in the toString() method.
      */
     public static Uuid fromString(String str) {
+        if (str.length() > 24) {
+            throw new IllegalArgumentException("Input string with prefix `"
+                + str.substring(0, 24) + "` is too long to be decoded as a base64 UUID");
+        }
+
         ByteBuffer uuidBytes = ByteBuffer.wrap(Base64.getUrlDecoder().decode(str));
         if (uuidBytes.remaining() != 16) {
-            throw new IllegalArgumentException("Failed to parse `" + str + "` as a base64-encoded UUID");
+            throw new IllegalArgumentException("Input string `" + str + "` decoded as "
+                + uuidBytes.remaining() + " bytes, which is more than the expected 16 bytes "
+                + "of a base64-encoded UUID");
         }
+
         return new Uuid(uuidBytes.getLong(), uuidBytes.getLong());
     }
 
--- a/clients/src/main/java/org/apache/kafka/common/Uuid.java
+++ b/clients/src/main/java/org/apache/kafka/common/Uuid.java
@@ -113,7 +113,7 @@ public class Uuid implements Comparable<Uuid> {
         ByteBuffer uuidBytes = ByteBuffer.wrap(Base64.getUrlDecoder().decode(str));
         if (uuidBytes.remaining() != 16) {
             throw new IllegalArgumentException("Input string `" + str + "` decoded as "
-                + uuidBytes.remaining() + " bytes, which is more than the expected 16 bytes "
+                + uuidBytes.remaining() + " bytes, which is not equal to the expected 16 bytes "
                 + "of a base64-encoded UUID");
         }
 
--- a/clients/src/main/java/org/apache/kafka/common/Uuid.java
+++ b/clients/src/main/java/org/apache/kafka/common/Uuid.java
@@ -106,6 +106,9 @@ public class Uuid implements Comparable<Uuid> {
      */
     public static Uuid fromString(String str) {
         ByteBuffer uuidBytes = ByteBuffer.wrap(Base64.getUrlDecoder().decode(str));
+        if (uuidBytes.remaining() != 16) {
+            throw new IllegalArgumentException("Failed to parse `" + str + "` as a base64-encoded UUID");
+        }
         return new Uuid(uuidBytes.getLong(), uuidBytes.getLong());
     }
 
--- a/core/src/main/scala/kafka/common/InconsistentControllerIdException.scala
+++ b/core/src/main/scala/kafka/common/InconsistentControllerIdException.scala
@@ -1,0 +1,24 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kafka.common
+
+class InconsistentControllerIdException(message: String, cause: Throwable) extends RuntimeException(message, cause) {
+  def this(message: String) = this(message, null)
+  def this(cause: Throwable) = this(null, cause)
+  def this() = this(null, null)
+}
--- a/core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
+++ b/core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
@@ -22,7 +22,6 @@ import java.nio.file.{Files, NoSuchFileException}
 import java.util.Properties
 
 import kafka.common.{InconsistentBrokerMetadataException, KafkaException}
-import kafka.server.KafkaRaftServer.{BrokerRole, ControllerRole, ProcessRole}
 import kafka.server.RawMetaProperties._
 import kafka.utils._
 import org.apache.kafka.common.Uuid
@@ -34,7 +33,7 @@ import scala.jdk.CollectionConverters._
 object RawMetaProperties {
   val ClusterIdKey = "cluster.id"
   val BrokerIdKey = "broker.id"
-  val ControllerIdKey = "controller.id"
+  val NodeIdKey = "node.id"
   val VersionKey = "version"
 }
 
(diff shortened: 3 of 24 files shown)


core/src/main/scala/kafka/common/InconsistentControllerIdException.scala
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kafka.common
+
+class InconsistentControllerIdException(message: String, cause: Throwable) extends RuntimeException(message, cause) {
+  def this(message: String) = this(message, null)
+  def this(cause: Throwable) = this(null, cause)
+  def this() = this(null, null)
```
ijuma: Do we need this constructor?

ijuma (commented):
Thanks for the PR. @hachikuji does version 1 use the base64 encoding for cluster id?

core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
```diff
@@ -21,38 +21,220 @@ import java.io._
...
+
+    for (logDir <- logDirs) {
+      val brokerCheckpointFile = new File(logDir + File.separator + "meta.properties")
+      val brokerCheckpoint = new BrokerMetadataCheckpoint(brokerCheckpointFile)
+
+      try {
+        brokerCheckpoint.read() match {
+          case Some(properties) => brokerMetadataMap += logDir -> properties
+          case None => if (ignoreMissing) {
+            logDir -> new Properties()
```
chia7712: unnecessary body?

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1495,6 +1517,13 @@ class KafkaConfig(val props: java.util.Map[_, _], doLog: Boolean, dynamicConfigO
     distinctRoles
   }
 
+  def metadataLogDir: String = {
+    Option(getString(KafkaConfig.MetadataLogDirProp)) match {
+      case Some(dir) => dir
+      case None => logDirs.head
```
chia7712, line 1524: Does it cause trouble if we change the order of directories from "log.dirs" when restarting server?

core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
```diff
@@ -21,38 +21,220 @@ import java.io._
...
+    logDirs: collection.Seq[String],
+    ignoreMissing: Boolean
+  ): (RawMetaProperties, collection.Seq[String]) = {
+    require(logDirs.nonEmpty, "Must have at least one log dir to read meta.properties")
+
+    val brokerMetadataMap = mutable.HashMap[String, Properties]()
+    val offlineDirs = mutable.ArrayBuffer.empty[String]
+
+    for (logDir <- logDirs) {
+      val brokerCheckpointFile = new File(logDir + File.separator + "meta.properties")
```
chia7712: How about using ```new File(logDir, "meta.properties")```?

core/src/main/scala/kafka/server/KafkaRaftServer.scala
```diff
@@ -100,4 +104,36 @@ object KafkaRaftServer {
   sealed trait ProcessRole
   case object BrokerRole extends ProcessRole
   case object ControllerRole extends ProcessRole
+
+  def loadMetaProperties(config: KafkaConfig): (MetaProperties, Seq[String]) = {
+    val logDirs = config.logDirs ++ Seq(config.metadataLogDir)
```
chia7712: How about ```config.logDirs :+ config.metadataLogDir```?

chia7712 (commented):
> the meta.properties file is required in each log directory with the new version property set to 1

Pardon me, what did you mean "in each log directory"?  I read this PR and it seems to me the new meta.properties must exists in "only one" log directory (the head of directories by default)

core/src/main/scala/kafka/server/KafkaConfig.scala
hachikuji: Good question. I think the implementation will have to be smart enough to find the metadata log dir in case we have already created the partition directory. Let me see if it is straightforward to add that here or if we should leave it for a follow-up. I guess the alternative would be to require `metadata.log.dir`, but I think we were trying to avoid that.

hachikuji:
> Pardon me, what did you mean "in each log directory"? I read this PR and it seems to me the new meta.properties must exists in "only one" log directory (the head of directories by default)

I guess this is proper chastisement for the shortage of tests verifying this behavior. I have added some and caught a bug in the process. Perhaps the bug was the source of the confusion.

hachikuji:
Another thing we need to do here is replace `controller.id` and `broker.id` with a single `node.id`. I will get to this tomorrow.

hachikuji:
@ijuma To answer your question, newly generated ClusterIds do indeed use the Base64 encoding. I think it probably needs to be smart enough to accept the old encoding as well though. Let me add a test case for this.

core/src/main/scala/kafka/server/KafkaConfig.scala
hachikuji: What I am thinking for the moment is the following:

1. Assume a deterministic mapping to the metadata dir from the configuration.
2. Raise an exception on startup if we find the `@metadata` log dir in another directory.

As a future improvement, we might relax this in case the metadata log dir has not been explicitly defined and is offline.

hachikuji:
@chia7712 @ijuma Thanks for the comments thus far. This is ready for another look when you have time.

core/src/main/scala/kafka/server/Server.scala
```diff
@@ -46,6 +46,22 @@ object Server {
     new Metrics(metricConfig, reporters, time, true, metricsContext)
   }
 
+  def initializeMetrics(
+    config: KafkaConfig,
+    time: Time,
+    metaProps: MetaProperties
```
chia7712: It seems to me the properties in ```MetaProperties``` is duplicate to ```KafkaConfig``` in this case. Is there any reason that we need to pass ```MetaProperties```?

clients/src/main/java/org/apache/kafka/common/Uuid.java
```diff
@@ -106,6 +106,9 @@ public String toString() {
      */
     public static Uuid fromString(String str) {
         ByteBuffer uuidBytes = ByteBuffer.wrap(Base64.getUrlDecoder().decode(str));
+        if (uuidBytes.remaining() != 16) {
+            throw new IllegalArgumentException("Failed to parse `" + str + "` as a base64-encoded UUID");
```
ijuma: It is worth being a bit more explicit about the issue, maybe something like "Expected 16 bytes, received Y bytes."? It may be a bit difficult to figure out why the parsing failed otherwise (since the string will have only valid characters).

ijuma: Also, would it be better to check the string size before decoding? If there's a bug and a really large string is passed, it's a bit safer to error before parsing (and to only include the first N characters in the exception).

core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
```diff
@@ -21,38 +21,203 @@ import java.io._
...
+import scala.jdk.CollectionConverters._
+
+object RawMetaProperties {
+  val ClusterIdKey = "cluster.id"
+  val BrokerIdKey = "broker.id"
+  val NodeIdKey = "node.id"
+  val VersionKey = "version"
+}
+
+case class RawMetaProperties(props: Properties = new Properties()) {
```
ijuma: Does it make sense for this to be a case class? We override `toString`, so what do we actually get from making it a case class here? Do we rely on the generated equals/hashCode?

core/src/main/scala/kafka/server/Server.scala
hachikuji: @chia7712 We need the clusterId from `MetaProperties`. I modified this to pass it through directly avoid the duplicate methods.

core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
```diff
@@ -21,38 +21,203 @@ import java.io._
...
 
-case class BrokerMetadata(brokerId: Int,
-                          clusterId: Option[String]) {
+import scala.collection.mutable
+import scala.jdk.CollectionConverters._
+
+object RawMetaProperties {
+  val ClusterIdKey = "cluster.id"
+  val BrokerIdKey = "broker.id"
+  val NodeIdKey = "node.id"
```
chia7712, line 36: Both "broker.id" and "node.id" are defined by ```KafkaConfig``` also. Should we unify them?

chia7712 (approved):
LGTM. just one trivial question is left.

core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
hachikuji: Yeah, that is the plan. Once KIP-500 is enabled, we will deprecate `broker.id` so that `node.id` is the only ID going forward. That said, I'm trying to give us a little wiggle room by defining this config as internal for now. So we still have an opportunity to change this. If you have any thoughts about it, definitely share them.

core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
hachikuji: I can't think of a reason to keep it as a case class. Will change.

clients/src/main/java/org/apache/kafka/common/Uuid.java
hachikuji: Thanks. Tried to improve this in the latest commit. Please take a look.

ijuma (commented):
@hachikuji Thanks for the updates. Looks reasonable to me at a high level. Since @chia7712 has approved it, I think we can go ahead and merge if the tests pass.

ijuma:
I think the case class change caused a compiler error, we probably need to make the field accessible.

hachikuji:
@ijuma Yep, got a little lazy verifying the case class conversion. Hopefully fixed now. We were in fact relying on the `equals`, but I modified the test case.

hachikuji:
One test failure fixed by https://github.com/apache/kafka/pull/10006. Merging to trunk.

PR #9971, opened 2021-01-26
MINOR: Call logSegments.toBuffer only when required

kowshik:
I've improved `Log.fetchOffsetByTimestamp` implementation slightly, such that we only copy the `segments` map values when `targetTimestamp` is neither `EARLIEST_TIMESTAMP` nor `LATEST_TIMESTAMP`. Previously, it was cached outside the `if` block, and that seemed a little bit of an overkill to me.

**Test plan:**
Rely on existing tests.

### DIFF
--- a/core/src/main/scala/kafka/log/Log.scala
+++ b/core/src/main/scala/kafka/log/Log.scala
@@ -1648,9 +1648,6 @@ class Log(@volatile private var _dir: File,
           s"for partition $topicPartition is ${config.messageFormatVersion} which is earlier than the minimum " +
           s"required version $KAFKA_0_10_0_IV0")
 
-      // Cache to avoid race conditions. `toBuffer` is faster than most alternatives and provides
-      // constant time access while being safe to use with concurrent collections unlike `toArray`.
-      val segmentsCopy = logSegments.toBuffer
       // For the earliest and latest, we do not need to return the timestamp.
       if (targetTimestamp == ListOffsetsRequest.EARLIEST_TIMESTAMP) {
         // The first cached epoch usually corresponds to the log start offset, but we have to verify this since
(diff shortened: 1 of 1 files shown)


kowshik:
cc @junrao for review

chia7712:
```logSegments``` is thread-safe so is the copy necessary in this case?

kowshik:
@chia7712: Going by the documentation, the results of the iteration are undefined if the map is modified while an iteration over the collection is in progress. IIUC we would end up iterating the values in the `segments` map via the line: `val targetSeg = segmentsCopy.find(_.largestTimestamp >= targetTimestamp)`.

Please refer to the code pointers in the following order:

https://github.com/confluentinc/ce-kafka/blob/master/core/src/main/scala/kafka/log/Log.scala#L318
https://github.com/confluentinc/ce-kafka/blob/master/core/src/main/scala/kafka/log/Log.scala#L2420
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentNavigableMap.html
Particularly this: https://docs.oracle.com/javase/8/docs/api/java/util/SortedMap.html#values--
> Collection<V> values()
Returns a Collection view of the values contained in this map. The collection's iterator returns the values in ascending order of the corresponding keys. The collection is backed by the map, so changes to the map are reflected in the collection, and vice-versa. If the map is modified while an iteration over the collection is in progress (except through the iterator's own remove operation), the results of the iteration are undefined. The collection supports element removal, which removes the corresponding mapping from the map, via the Iterator.remove, Collection.remove, removeAll, retainAll and clear operations. It does not support the add or addAll operations.

So... to avoid the above thread safety issue, prior to the iteration, a copy of the segments map values is made.

Please go ahead and merge this PR if it looks good to you.

chia7712:
@kowshik Thanks for your response. This patch is LGTM and I just want to clarify the usage of ```segmentsCopy```

The instance of ```segmentsCopy``` is ```ConcurrentSkipListMap``` so the return ```Iterable``` should be thread safe (see https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentSkipListMap.html#values--).

kowshik:
@chia7712 Good point. I understand now that iteration over `ConcurrentSkipListMap.values()` is weakly consistent, which seems good enough for this case. But since my intention with this patch was to not attempt to change the iteration behavior, but rather just the caching part, I'm hoping the PR in its existing form is acceptable. Please let me know how you feel.

chia7712:
> But since my intention with this patch was to not change the iteration behavior, I'm hoping the PR in its existing form is acceptable.

make sense to me :)

chia7712 (approved):

ijuma (approved):
Thanks, LGTM. I think the PR title is a bit confusing. I would write something like "Call `logSegments.toBuffer` only when required" since caching tends to imply more than just allocating to a local val.

satishd (approved):
+1 LGTM

kowshik:
@ijuma Good point, I've fixed the PR title now.

PR #9972, opened 2021-01-26
KAFKA-8779: Reintroduce flaky tests

tombentley:
Enable some previously-flaky PlaintextAdminIntegrationTests.

The cause of the flakiness appears to have been lack of test isolation between those tests and other tests which manipulate log levels (which are effectively global to the JVM). In general there are quite a few tests which manipulate log4j and/or make assertions on log4j global state, so introduce some of helpers to try to get test isolation across the whole of the broker code.

1. `logLock` in `TestUtils` provides the mutual exclusion. We can't use `synchronized` on an `Object` because some tests change the logging in their `@Before(Class)` and restore in ``@After(Class)`.
2. `TestUtils.withLogLock` encapsulates the isolation logic, and it used by most of the log-changing tests, rather than them using `logLock` directly.
3. `LogCaptureAppender` now uses `withLogLock` too, and there's no reason not to apply the same state management pattern via `captureLogging`.

### DIFF
--- a/core/src/test/scala/integration/kafka/api/AuthorizerIntegrationTest.scala
+++ b/core/src/test/scala/integration/kafka/api/AuthorizerIntegrationTest.scala
@@ -25,6 +25,7 @@ import kafka.security.authorizer.AclEntry
 import kafka.security.authorizer.AclEntry.WildcardHost
 import kafka.server.{BaseRequestTest, KafkaConfig}
 import kafka.utils.TestUtils
+import kafka.utils.TestUtils.withLogReset
 import org.apache.kafka.clients.admin.{Admin, AdminClientConfig, AlterConfigOp}
 import org.apache.kafka.clients.consumer._
 import org.apache.kafka.clients.consumer.internals.NoOpConsumerRebalanceListener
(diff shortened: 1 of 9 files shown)


tombentley:
I've now run these tests 4 or 5 times in CI and not seen any failures in the previously flaky tests where previously I did, so I'm hopeful that these changes to indeed fix the issue.

@chia7712 or @mimaison perhaps one of you could review this change?

cc @stanislavkozlovski who wrote the tests originally.

chia7712:
> The cause of the flakiness appears to have been lack of test isolation between those tests and other tests which manipulate log levels (which are effectively global to the JVM). In general there are quite a few tests which manipulate log4j and/or make assertions on log4j global state, so introduce some of helpers to try to get test isolation across the whole of the broker code.

IIRC, gradle forks `n` JVMs to run `n` tests parallel and tests in same JVM are executed sequentially. Hence, could we reset log4j state (maybe in `@AfterEach`) instead of global locking? Please correct me if I misunderstand the root cause.

tombentley:
Thanks @chia7712 that's an excellent point! Maybe it's worth inlining `withLogReset` since it offers little value now?

tombentley:
Rebased to trigger CI again.

@chia7712 please could to take another look when you have the chance?

tombentley:
Rebased to trigger CI again.

@chia7712 would you be able to take another look?

ijuma:
@chia7712 Does this look good to you?

mjsax:
> KAFKA-8779: Reintroduce flaky tests

@tombentley -- I don't think we should re-introduce flaky tests 😁 (sorry, not sorry; couldn't resist)

chia7712:
all disabled tests are enabled now (see #17382)

PR #9973, opened 2021-01-26
KAFKA-12239 Detail log 'Error getting JMX attribute'

gquintana:
When collecting bulk metrics, this warning message in logs is unhelpful, it is impossible to determine which MBean is missing this attribute and fix the metric collector:

``[2021-01-26T15:43:41,078][WARN ][org.apache.kafka.common.metrics.JmxReporter] Error getting JMX attribute 'records-lag-max'
javax.management.AttributeNotFoundException: Could not find attribute records-lag-max
```
This pull request just adds the MBean object name in the log to ease debugging.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/metrics/JmxReporter.java
+++ b/clients/src/main/java/org/apache/kafka/common/metrics/JmxReporter.java
@@ -262,7 +262,7 @@ public class JmxReporter implements MetricsReporter {
                 try {
                     list.add(new Attribute(name, getAttribute(name)));
                 } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


chia7712:
@gquintana thanks for your patch. Could you share the new error message to me?

nitesh-jain-kr:
Hi @gquintana, this PR was not merged. Do we have a fix for this issue?

gquintana:
Hi @nitesh-jain-kr AFAIK the issue is still here. However as Kafka seems to be moving away from JMX, it may not be relevant once all metrics can be collected through OpenTelemetry

PR #9978, opened 2021-01-27
KAFKA-10716: persist UUID in state directory for stable processId across restarts

ableegoldman:
To stabilize the task assignment across restarts of the JVM we need some way to persist the process-specific UUID. We can just write it to a file in the state directory, and initialize it from there or create a new one if no prior UUID exists. 

This also adds a state directory-level lock -- partly to protect the "process file" where we write the UUID, and partly to address an issue that has been plaguing users for a long time: with this top-level lock, we can now strictly enforce that only one instance of Streams is running on the same physical state directory. This configuration has never been supported, but it's never been explicitly protected against, leading to often murky symptoms such as tasks being stuck uninitialized. With this change, we will fail fast if another process holds the lock for this state directory


streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -782,8 +783,27 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+        taskTopology = internalTopologyBuilder.buildTopology();
+        globalTaskTopology = internalTopologyBuilder.buildGlobalStateTopology();
+
+        final boolean hasGlobalTopology = globalTaskTopology != null;
+        final boolean hasPersistentStores = taskTopology.hasPersistentLocalStore() ||
+            (hasGlobalTopology && globalTaskTopology.hasPersistentGlobalStore());
+
+        try {
+            stateDirectory = new StateDirectory(config, time, hasPersistentStores);
+            processId = stateDirectory.getProcessId();
```
ableegoldman: This is the only real change in the constructor, but I had to move a few things around and tried to organize them as I went

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -782,8 +782,27 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
+        taskTopology = internalTopologyBuilder.buildTopology();
+        globalTaskTopology = internalTopologyBuilder.buildGlobalStateTopology();
+
+        final boolean hasGlobalTopology = globalTaskTopology != null;
+        final boolean hasPersistentStores = taskTopology.hasPersistentLocalStore() ||
+            (hasGlobalTopology && globalTaskTopology.hasPersistentGlobalStore());
+
+        try {
+            stateDirectory = new StateDirectory(config, time, hasPersistentStores);
+            processId = stateDirectory.initializeProcessId();
```
ableegoldman, line 804: this is the only logical change in the KafkaStreams constructor: the rest of the diff is due to moving things around in order to get everything initialized in the proper order

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -927,6 +912,39 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+        } else {
+            numStreamThreads = config.getInt(StreamsConfig.NUM_STREAM_THREADS_CONFIG);
+        }
+
+        if (numStreamThreads == 0 && !hasGlobalTopology) {
+            log.error("Topology with no input topics will create no stream threads and no global thread.");
+            throw new TopologyException("Topology has no stream threads and no global threads, " +
+                                            "must subscribe to at least one source topic or global table.");
+        }
+        return numStreamThreads;
```
ableegoldman, line 950: Just tried to factor some of the self-contained logic into helper methods, since I found it incredibly difficult to get oriented within the super-long KafkaStreams constructor

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1208,24 +1227,28 @@ private Thread shutdownHelper(final boolean error) {
     }
 
     private boolean close(final long timeoutMs) {
-        if (state == State.ERROR) {
-            log.info("Streams client is already in the terminal state ERROR, all resources are closed and the client has stopped.");
+        if (state == State.ERROR || state == State.NOT_RUNNING) {
```
ableegoldman, line 1298: Something I noticed during testing, I feel it makes sense for the handling of ERROR and NOT_RUNNING to parallel (same for the PENDING_ flavors). This is a slight change in behavior; now if a user calls `close()` while the instance is already closing, it will wait for the ongoing shutdown to complete before returning (with timeout).

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -54,14 +61,27 @@
...
     static final String LOCK_FILE_NAME = ".lock";
 
+    /* The process file is used to persist the process id across restarts.
+     * The version 0 schema consists only of the version number and UUID
+     *
+     * If you need to store additional metadata of the process you can bump the version numberand append new fields.
+     * For compatibility reasons you should only ever add fields, and only by appending them to the end
+     */
+    private static final String PROCESS_FILE_NAME = "kafka-streams-process-metadata";
+    private static final int PROCESS_FILE_VERSION = 0;
```
ableegoldman: No idea if we'll ever want to add anything else to this file, but better to be safe and forward compatible than sad

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -416,11 +524,15 @@ private void cleanRemovedTasksCalledByUser() throws Exception {
                                 logPrefix(), dirName, id),
                             exception
                         );
-                        throw exception;
```
ableegoldman, line 420: IDE was giving me a warning

streams/src/test/java/org/apache/kafka/streams/integration/StandbyTaskEOSIntegrationTest.java
```diff
@@ -112,6 +118,19 @@ public void createTopics() throws Exception {
...
+    public void cleanUp() {
+        if (streamInstanceOne != null) {
+            streamInstanceOne.close();
+        }
+        if (streamInstanceTwo != null) {
+            streamInstanceTwo.close();
+        }
+        if (streamInstanceOneRecovery != null) {
+            streamInstanceOneRecovery.close();
+        }
```
ableegoldman, line 131: There are no logical changes to this test, I just had to refactor it a bit because we were creating two copies of the same KafkaStreams at the same time (with the same app.dir & state.dir), even though one of them wasn't started until much later. Since we do the state initialization inside the KafkaStreams constructor, this was no good

ableegoldman:
Not done with the tests, but I'd appreciate some feedback on the non-testing code and general idea -- any takers for review? @cadonna @vvcephei @guozhangwang @wcarlson5 @lct45

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1208,24 +1227,28 @@ private Thread shutdownHelper(final boolean error) {
...
+            log.info("Streams client is in {}, all resources are being closed and the client will be stopped.", state);
+            if (state == State.PENDING_ERROR && waitOnState(State.ERROR, timeoutMs)) {
                 log.info("Streams client stopped to ERROR completely");
                 return true;
+            } else if (state == State.PENDING_SHUTDOWN && waitOnState(State.NOT_RUNNING, timeoutMs)) {
+                log.info("Streams client stopped to NOT_RUNNING completely");
+                return true;
             } else {
-                log.info("Streams client cannot transition to ERROR completely within the timeout");
+                log.warn("Streams client cannot transition to {}} completely within the timeout", state);
```
wcarlson5: the state here doesn't make the log make sense. If the state is `PENDING_ERROR` then the log should say ERROR

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I think this change makes a lot of sense. I don't think it changes the final behavior besides avoiding extra state change rejections from the logs, but it looks like they are replaced.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -133,6 +152,72 @@ private void configurePermissions(final File file) {
...
+            stateDirLock = tryLock(stateDirLockChannel);
+        } catch (final IOException e) {
+            log.error("Unable to lock the state directory due to unexpected exception", e);
+            throw new ProcessorStateException("Failed to lock the state directory during startup", e);
+        }
+
+        return stateDirLock != null;
+    }
+
+    public UUID initializeProcessId() {
```
wcarlson5, line 184: since it doesn't seem that we need to be very thrifty with space for this file would it make sense to write it in a more friendly format that would be easier to maintain? i.e. json or something, we are giving it a version number...

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -133,6 +152,72 @@ private void configurePermissions(final File file) {
...
     }
 
+    /**
+     * @return true if the state directory was successfully locked
+     */
+    private boolean lockStateDirectory() {
+        final File lockFile = new File(stateDir, LOCK_FILE_NAME);
+        try {
+            stateDirLockChannel = FileChannel.open(lockFile.toPath(), StandardOpenOption.CREATE, StandardOpenOption.WRITE);
+            stateDirLock = tryLock(stateDirLockChannel);
```
wcarlson5, line 175: Is there any case where we might want to release the lock of this state directory? It looks like we just hold it

wcarlson5 (commented):
Overall it looks good, I just had a few comments

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
ableegoldman: Nope, we just hold it until the KafkaStreams is closed

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
```diff
@@ -133,6 +152,72 @@ private void configurePermissions(final File file) {
...
+                                           "Kafka Streams are running in the same state directory");
+        }
+
+        final File processFile = new File(stateDir, PROCESS_FILE_NAME);
+        try {
+            if (processFile.exists()) {
+                try (final BufferedReader reader = Files.newBufferedReader(processFile.toPath())) {
+                    // only field in version 0 is the UUID
+                    final int version = Integer.parseInt(reader.readLine());
+                    if (version > 0) {
```
lct45: Do we want to make this set to PROCESS_FILE_VERSION on the off chance anyone needs to increment it?

lct45 (commented):
thanks for the PR! overall LGTM

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
ableegoldman: Ah good catch

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
ableegoldman: Oh yeah definitely, thanks

streams/src/main/java/org/apache/kafka/streams/processor/internals/StateDirectory.java
ableegoldman: I'll try it out with json. If we do use json, then we don't even need the version number right? As long as we only ever add fields then it should always be compatible. At least that's my understanding, I'm not a json expert

ableegoldman: Oh my god. I hate json lol

ableegoldman: alright done

wcarlson5: LGTM

ableegoldman:
A few unrelated test failures:
```
Build / JDK 11 / org.apache.kafka.clients.consumer.internals.FetcherTest.testEarlierOffsetResetArrivesLate()
Build / JDK 8 / kafka.api.CustomQuotaCallbackTest.testCustomQuotaCallback()
Build / JDK 8 / kafka.api.PlaintextConsumerTest.testMultiConsumerStickyAssignment()
Build / JDK 15 / kafka.api.CustomQuotaCallbackTest.testCustomQuotaCallback()
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
guozhangwang: cc @cadonna for the pre-creation of internal topics, we would also need to build the topology in order to populate the `RepartitionTopics` etc, let's just make sure we only build it once when we add that logic.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -798,73 +817,42 @@ private KafkaStreams(final InternalTopologyBuilder internalTopologyBuilder,
...
         ClientMetrics.addCommitIdMetric(streamsMetrics);
         ClientMetrics.addApplicationIdMetric(streamsMetrics, config.getString(StreamsConfig.APPLICATION_ID_CONFIG));
         ClientMetrics.addTopologyDescriptionMetric(streamsMetrics, internalTopologyBuilder.describe().toString());
         ClientMetrics.addStateMetric(streamsMetrics, (metricsConfig, now) -> state);
-        log.info("Kafka Streams version: {}", ClientMetrics.version());
-        log.info("Kafka Streams commit ID: {}", ClientMetrics.commitId());
-        this.internalTopologyBuilder = internalTopologyBuilder;
-        // re-write the physical topology according to the config
-        internalTopologyBuilder.rewriteTopology(config);
+        ClientMetrics.addNumAliveStreamThreadMetric(streamsMetrics, (metricsConfig, now) ->
```
guozhangwang, line 841: Thanks for the code cleanup and re-ordering! BTW if there's no logical changes at all maybe add a comment on the PR next time.

guozhangwang (approved):
LGTM! Could you file a JIRA ticket for clearing the lower-level locks as follow-ups? My understanding is that, not only task-level locks, but also global state dir locks, can be removed.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
ableegoldman: Ah sorry, I did but it got covered up when I pushed some changes: https://github.com/apache/kafka/pull/9978#discussion_r565746851

ableegoldman:
Opened https://issues.apache.org/jira/browse/KAFKA-12288 for followup cleaning out of filesystem locks

ableegoldman:
Merged to trunk

ableegoldman:
The cherrypick seems to have a lot of merge conflicts so I'll open a separate PR for 2.7/2.6

PR #9979, opened 2021-01-27
KAFKA-12238; Implement `DescribeProducers` API from KIP-664

hachikuji:
Implements the `DescribeProducers` API specified by KIP-664: https://cwiki.apache.org/confluence/display/KAFKA/KIP-664%3A+Provide+tooling+to+detect+and+abort+hanging+transactions.

Note that this patch contains only the server implementation. The `Admin` changes will come in a follow-up.

### DIFF
--- a/clients/src/main/resources/common/message/DescribeProducersResponse.json
+++ b/clients/src/main/resources/common/message/DescribeProducersResponse.json
@@ -1,0 +1,46 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+{
+  "apiKey": 61,
+  "type": "response",
+  "name": "DescribeProducersResponse",
+  "validVersions": "0",
+  "flexibleVersions": "0+",
+  "fields": [
+    { "name": "ThrottleTimeMs", "type": "int32", "versions": "0+", "ignorable": true,
+      "about": "The duration in milliseconds for which the request was throttled due to a quota violation, or zero if the request did not violate any quota." },
+    { "name": "Topics", "type": "[]TopicResponse", "versions": "0+",
+      "about": "Each topic in the response.", "fields": [
+      { "name": "Name", "type": "string", "versions": "0+", "entityType": "topicName",
+        "about": "The topic name" },
+      { "name": "Partitions", "type": "[]PartitionResponse", "versions": "0+",
+        "about": "Each partition in the response.", "fields": [
+        { "name": "PartitionIndex", "type": "int32", "versions": "0+",
+          "about": "The partition index." },
+        { "name": "ErrorCode", "type": "int16", "versions": "0+",
+          "about": "The partition error code, or 0 if there was no error." },
+        { "name": "ActiveProducers", "type": "[]ProducerState", "versions": "0+", "fields": [
+          { "name": "ProducerId", "type": "int64", "versions": "0+" },
+          { "name": "ProducerEpoch", "type": "int32", "versions": "0+" },
+          { "name": "LastSequence", "type": "int32", "versions": "0+", "default":  "-1" },
+          { "name": "LastTimestamp", "type": "int64", "versions": "0+", "default": "-1" },
+          { "name": "CoordinatorEpoch", "type":  "int32", "versions":  "0+" },
+          { "name": "CurrentTxnStartOffset", "type": "int64", "versions": "0+", "default": "-1" }
+        ]}
+      ]}
+    ]}
+  ]
+}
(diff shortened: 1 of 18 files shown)


clients/src/main/resources/common/message/DescribeProducersResponse.json
```diff
@@ -0,0 +1,46 @@
...
+// limitations under the License.
+
+{
+  "apiKey": 61,
+  "type": "response",
+  "name": "DescribeProducersResponse",
+  "validVersions": "0",
+  "flexibleVersions": "0+",
+  "fields": [
+    { "name": "ThrottleTimeMs", "type": "int32", "versions": "0+", "ignorable": true,
```
dajac: nit: `ignorable` not really necessary as we start from version 0.

clients/src/main/resources/common/message/DescribeProducersResponse.json
```diff
@@ -0,0 +1,46 @@
...
+    { "name": "Topics", "type": "[]TopicResponse", "versions": "0+",
+      "about": "Each topic in the response.", "fields": [
+      { "name": "Name", "type": "string", "versions": "0+", "entityType": "topicName",
+        "about": "The topic name" },
+      { "name": "Partitions", "type": "[]PartitionResponse", "versions": "0+",
+        "about": "Each partition in the response.", "fields": [
+        { "name": "PartitionIndex", "type": "int32", "versions": "0+",
+          "about": "The partition index." },
+        { "name": "ErrorCode", "type": "int16", "versions": "0+",
+          "about": "The partition error code, or 0 if there was no error." },
```
dajac, line 34: Should we consider adding `ErrorMessage` as well? We tend to have it nowadays in many responses but I am not sure if it is a generality.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -511,6 +513,9 @@ public void testSerialization() throws Exception {
         checkRequest(createAlterClientQuotasRequest(), true);
         checkErrorResponse(createAlterClientQuotasRequest(), unknownServerException, true);
         checkResponse(createAlterClientQuotasResponse(), 0, true);
+        checkRequest(createDescribeProducersRequest(), true);
+        checkErrorResponse(createDescribeProducersRequest(), unknownServerException, true);
+        checkResponse(createDescribeProducersResponse(), 0, true);
```
dajac: nit: This test case is getting quite big, too big. Last time I added a new request/response, I went with a dedicated unit test for it. See `testDescribeClusterSerialization`. I wanted to refactor it to generalize this pattern but did not have the time to do it yet. What do you think about this?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3365,6 +3366,44 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+        val partitionResponse = topicError match {
+          case Some(error) => partitionError(topicPartition, error)
+          case None => replicaManager.activeProducerState(topicPartition)
+        }
+        topicResponse.partitions.add(partitionResponse)
+      }
+
+      if (!topicResponse.partitions.isEmpty) {
+        response.topics.add(topicResponse)
+      }
```
dajac: What's the rational for this? Is it because we can't convey a possible topic-level error in the response if we don't have at least one partition for it?

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -3082,4 +3082,76 @@ class KafkaApisTest {
...
+    assertEquals(15, fooProducer.producerEpoch)
+    assertEquals(100, fooProducer.lastSequence)
+    assertEquals(time.milliseconds(), fooProducer.lastTimestamp)
+    assertEquals(-1, fooProducer.currentTxnStartOffset)
+    assertEquals(200, fooProducer.coordinatorEpoch)
+
+    val barTopic = response.data.topics.asScala.find(_.name == tp2.topic).get
+    val barPartition = barTopic.partitions.asScala.find(_.partitionIndex == tp2.partition).get
+    assertEquals(Errors.TOPIC_AUTHORIZATION_FAILED.code, barPartition.errorCode)
+  }
```
dajac, line 3174: I wonder if we could add a third topic to test the `UNKNOWN_TOPIC_OR_PARTITION` case. What do you think?

dajac (commented):
@hachikuji Thanks for the PR. I left few minor comments. We should add the request to `RequestConvertToJson`. I also wonder if we should add the new API in `AuthorizerIntegrationTest` for completeness.

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
hachikuji: Sounds reasonable. I'm also not super fond of this test pattern.

clients/src/main/resources/common/message/DescribeProducersResponse.json
hachikuji: I don't feel strongly about it. Perhaps it's better to have it than not. I will add it.

core/src/main/scala/kafka/server/KafkaApis.scala
hachikuji: Hmm... It's been a while since I wrote this. I agree it looks a little strange. I guess there's probably no harm echoing back the same structure that was sent.

dajac:
@hachikuji Thanks for the updates. The code LGTM. However, there might be something wrong with `testIllegalRequiredAcks` which failed in the three builds. Could you verify?
* Build / JDK 15 / kafka.server.ReplicaManagerTest.testIllegalRequiredAcks()
* Build / JDK 8 / kafka.server.ReplicaManagerTest.testIllegalRequiredAcks()
* Build / JDK 11 / kafka.server.ReplicaManagerTest.testIllegalRequiredAcks()

hachikuji:
@dajac Thanks, I tried that test locally a few times and couldn't get it to fail. I triggered another build, so let's see how it goes.

PR #9980, opened 2021-01-27
MINOR: Reduce size of the ProducerStateEntry batchMetadata queue.

gardnervickers:
This reduces the size at construction of the batchMetadata queue from 16
entries to 5. The backing array will be sized to 8 elements, cutting the array size
in half.

We know that we will only allow a max of 5 entries in this queue, so this change will not
change the alloc/copy behavior.

### DIFF
--- a/core/src/main/scala/kafka/log/ProducerStateManager.scala
+++ b/core/src/main/scala/kafka/log/ProducerStateManager.scala
@@ -63,7 +63,7 @@ private[log] object ProducerStateEntry {
   private[log] val NumBatchesToRetain = 5
 
   def empty(producerId: Long) = new ProducerStateEntry(producerId,
-    batchMetadata = new mutable.Queue[BatchMetadata](5),
+    batchMetadata = new mutable.Queue[BatchMetadata](NumBatchesToRetain),
     producerEpoch = RecordBatch.NO_PRODUCER_EPOCH,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/kafka/log/ProducerStateManager.scala
```diff
@@ -63,7 +63,7 @@ private[log] object ProducerStateEntry {
   private[log] val NumBatchesToRetain = 5
 
   def empty(producerId: Long) = new ProducerStateEntry(producerId,
-    batchMetadata = mutable.Queue[BatchMetadata](),
+    batchMetadata = new mutable.Queue[BatchMetadata](5),
```
satishd: minor: you may want to have it as `new mutable.Queue[BatchMetadata](NumBatchesToRetain)`  instead of harcoding directly.

satishd (commented):
thanks @gardnervickers for the PR, left a minor comment.

core/src/main/scala/kafka/log/ProducerStateManager.scala
gardnervickers: Yes, good suggestion :)

satishd (approved):
thanks @gardnervickers for addressing the review comments. LGTM.

PR #9984, opened 2021-01-27
KAFKA-12247: add timeout and static group rebalance to remove thread

wcarlson5:
add timeout and static group rebalance to remove thread

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -25,10 +25,10 @@ import org.apache.kafka.clients.admin.RemoveMembersFromConsumerGroupResult;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
-import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.common.Metric;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.common.metrics.JmxReporter;
 import org.apache.kafka.common.metrics.KafkaMetricsContext;
 import org.apache.kafka.common.metrics.MetricConfig;
@@ -1056,7 +1056,7 @@ public class KafkaStreams implements AutoCloseable {
                                 throw new StreamsException(
                                         "Could not remove static member " + groupInstanceID.get()
                                             + " from consumer group " + config.getString(StreamsConfig.APPLICATION_ID_CONFIG)
-                                            + " the following reason: ",
+                                            + " for the following reason: ",
                                         e.getCause()
                                 );
                             }
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -25,7 +25,7 @@ import org.apache.kafka.clients.admin.RemoveMembersFromConsumerGroupResult;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
-import org.apache.kafka.common.KafkaException;
+import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.common.Metric;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.TopicPartition;
@@ -73,7 +73,6 @@ import org.apache.kafka.streams.state.internals.GlobalStateStoreProvider;
 import org.apache.kafka.streams.state.internals.QueryableStoreProvider;
 import org.apache.kafka.streams.state.internals.RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapter;
 import org.apache.kafka.streams.state.internals.StreamThreadStateStoreProvider;
-import org.apache.kafka.common.errors.TimeoutException;
 import org.slf4j.Logger;
 
 import java.time.Duration;
@@ -1046,13 +1045,20 @@ public class KafkaStreams implements AutoCloseable {
                             try {
                                 removeMembersFromConsumerGroupResult.memberResult(memberToRemove).get(timeoutMs - begin, TimeUnit.MILLISECONDS);
                             } catch (final java.util.concurrent.TimeoutException e) {
-                                log.warn("Rethrowing a java TimeoutException as a kafka TimeoutException " + Arrays.toString(e.getStackTrace()));
-                                throw new TimeoutException(e.getMessage());
+                                log.error("Could not remove static member {} from consumer group {} due to a timeout: {}",
+                                        groupInstanceID.get(), config.getString(StreamsConfig.APPLICATION_ID_CONFIG), e);
+                                throw new TimeoutException(e.getMessage(), e);
                             } catch (final InterruptedException e) {
                                 Thread.currentThread().interrupt();
                             } catch (final ExecutionException e) {
-                                log.error("Getting the member result threw this ExecutionException: " + e.getMessage());
-                                throw new StreamsException("Could not remove member from group for the following reason: ", e.getCause());
+                                log.error("Could not remove static member {} from consumer group {} due to: {}",
+                                        groupInstanceID.get(), config.getString(StreamsConfig.APPLICATION_ID_CONFIG), e);
+                                throw new StreamsException(
+                                        "Could not remove static member " + groupInstanceID.get()
+                                            + " from consumer group " + config.getString(StreamsConfig.APPLICATION_ID_CONFIG)
+                                            + " the following reason: ",
+                                        e.getCause()
+                                );
                             }
                         }
                         if (timeout) {
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -25,6 +25,7 @@ import org.apache.kafka.clients.admin.RemoveMembersFromConsumerGroupResult;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.KafkaException;
 import org.apache.kafka.common.Metric;
 import org.apache.kafka.common.MetricName;
 import org.apache.kafka.common.TopicPartition;
@@ -1051,6 +1052,7 @@ public class KafkaStreams implements AutoCloseable {
                                 Thread.currentThread().interrupt();
                             } catch (final ExecutionException e) {
                                 log.error("Getting the member result threw this ExecutionException: " + e.getMessage());
+                                throw new StreamsException("Could not remove member from group for the following reason: ", e.getCause());
                             }
                         }
                         if (timeout) {
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -1034,7 +1034,7 @@ public class KafkaStreams implements AutoCloseable {
                         final long cacheSizePerThread = getCacheSizePerThread(threads.size());
                         resizeThreadCache(cacheSizePerThread);
                         if (groupInstanceID.isPresent() && !streamThread.getName().equals(Thread.currentThread().getName())) {
-                            final MemberToRemove memberToRemove = new MemberToRemove(streamThread.getGroupInstanceID().get());
+                            final MemberToRemove memberToRemove = new MemberToRemove(groupInstanceID.get());
                             final Collection<MemberToRemove> membersToRemove = Collections.singletonList(memberToRemove);
                             final RemoveMembersFromConsumerGroupResult removeMembersFromConsumerGroupResult = 
                                 adminClient.removeMembersFromConsumerGroup(
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -21,6 +21,7 @@ import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.admin.ListOffsetsResult.ListOffsetsResultInfo;
 import org.apache.kafka.clients.admin.MemberToRemove;
 import org.apache.kafka.clients.admin.RemoveMembersFromConsumerGroupOptions;
+import org.apache.kafka.clients.admin.RemoveMembersFromConsumerGroupResult;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
@@ -90,10 +91,11 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import org.apache.kafka.common.errors.TimeoutException;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 
@@ -986,30 +988,7 @@ public class KafkaStreams implements AutoCloseable {
      *         no stream threads are alive
      */
     public Optional<String> removeStreamThread() {
-        if (isRunningOrRebalancing()) {
-            synchronized (changeThreadCount) {
-                // make a copy of threads to avoid holding lock
-                for (final StreamThread streamThread : new ArrayList<>(threads)) {
-                    if (streamThread.isAlive() && (!streamThread.getName().equals(Thread.currentThread().getName())
-                            || threads.size() == 1)) {
-                        streamThread.shutdown();
-                        if (!streamThread.getName().equals(Thread.currentThread().getName())) {
-                            streamThread.waitOnThreadState(StreamThread.State.DEAD, -1);
-                        }
-                        threads.remove(streamThread);
-                        final long cacheSizePerThread = getCacheSizePerThread(threads.size());
-                        resizeThreadCache(cacheSizePerThread);
-                        final Collection<MemberToRemove> membersToRemove = Collections.singletonList(new MemberToRemove(streamThread.getGroupInstanceID()));
-                        adminClient.removeMembersFromConsumerGroup(config.getString(StreamsConfig.APPLICATION_ID_CONFIG), new RemoveMembersFromConsumerGroupOptions(membersToRemove));
-                        return Optional.of(streamThread.getName());
-                    }
-                }
-            }
-            log.warn("There are no threads eligible for removal");
-        } else {
-            log.warn("Cannot remove a stream thread when Kafka Streams client is in state  " + state());
-        }
-        return Optional.empty();
+        return removeStreamThread(Long.MAX_VALUE);
     }
 
     /**
@@ -1030,6 +1009,11 @@ public class KafkaStreams implements AutoCloseable {
     public Optional<String> removeStreamThread(final Duration timeout) throws TimeoutException {
         final String msgPrefix = prepareMillisCheckFailMsgPrefix(timeout, "timeout");
         final long timeoutMs = validateMillisecondDuration(timeout, msgPrefix);
+        return removeStreamThread(timeoutMs);
+    }
+
+    private Optional<String> removeStreamThread(final long timeoutMs) throws TimeoutException {
+        final long begin = time.milliseconds();
         if (isRunningOrRebalancing()) {
             synchronized (changeThreadCount) {
                 // make a copy of threads to avoid holding lock
@@ -1046,8 +1030,20 @@ public class KafkaStreams implements AutoCloseable {
                         threads.remove(streamThread);
                         final long cacheSizePerThread = getCacheSizePerThread(threads.size());
                         resizeThreadCache(cacheSizePerThread);
-                        Collection<MemberToRemove> membersToRemove = Collections.singletonList(new MemberToRemove(streamThread.getGroupInstanceID()));
-                        adminClient.removeMembersFromConsumerGroup(config.getString(StreamsConfig.APPLICATION_ID_CONFIG), new RemoveMembersFromConsumerGroupOptions(membersToRemove));
+                        if (streamThread.getGroupInstanceID().isPresent()) {
+                            final MemberToRemove memberToRemove = new MemberToRemove(streamThread.getGroupInstanceID().get());
+                            final Collection<MemberToRemove> membersToRemove = Collections.singletonList(memberToRemove);
+                            final RemoveMembersFromConsumerGroupResult removeMembersFromConsumerGroupResult = adminClient.removeMembersFromConsumerGroup(config.getString(StreamsConfig.APPLICATION_ID_CONFIG), new RemoveMembersFromConsumerGroupOptions(membersToRemove));
+                            try {
+                                removeMembersFromConsumerGroupResult.memberResult(memberToRemove).get(timeoutMs - begin, TimeUnit.MILLISECONDS);
+                            } catch (final java.util.concurrent.TimeoutException e) {
+                                throw new TimeoutException(e.getMessage());
+                            } catch (final InterruptedException e) {
+                                Thread.currentThread().interrupt();
+                            } catch (final ExecutionException e) {
+                                e.printStackTrace();
+                            }
+                        }
                         return Optional.of(streamThread.getName());
                     }
                 }
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -1036,7 +1036,11 @@ public class KafkaStreams implements AutoCloseable {
                         if (groupInstanceID.isPresent() && !streamThread.getName().equals(Thread.currentThread().getName())) {
                             final MemberToRemove memberToRemove = new MemberToRemove(streamThread.getGroupInstanceID().get());
                             final Collection<MemberToRemove> membersToRemove = Collections.singletonList(memberToRemove);
-                            final RemoveMembersFromConsumerGroupResult removeMembersFromConsumerGroupResult = adminClient.removeMembersFromConsumerGroup(config.getString(StreamsConfig.APPLICATION_ID_CONFIG), new RemoveMembersFromConsumerGroupOptions(membersToRemove));
+                            final RemoveMembersFromConsumerGroupResult removeMembersFromConsumerGroupResult = 
+                                adminClient.removeMembersFromConsumerGroup(
+                                    config.getString(StreamsConfig.APPLICATION_ID_CONFIG), 
+                                    new RemoveMembersFromConsumerGroupOptions(membersToRemove)
+                                );
                             try {
                                 removeMembersFromConsumerGroupResult.memberResult(memberToRemove).get(timeoutMs - begin, TimeUnit.MILLISECONDS);
                             } catch (final java.util.concurrent.TimeoutException e) {
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -1026,7 +1026,7 @@ public class KafkaStreams implements AutoCloseable {
                         streamThread.shutdown();
                         if (!streamThread.getName().equals(Thread.currentThread().getName())) {
                             if (!streamThread.waitOnThreadState(StreamThread.State.DEAD, timeoutMs)) {
-                                log.warn("Thread " + streamThread.getName() + " did not stop in the allotted time");
+                                log.warn("Thread " + streamThread.getName() + " did not shutdown in the allotted time");
                                 timeout = true;
                             }
                         }
--- a/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
+++ b/streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
@@ -72,6 +72,7 @@ import org.apache.kafka.streams.state.internals.GlobalStateStoreProvider;
 import org.apache.kafka.streams.state.internals.QueryableStoreProvider;
 import org.apache.kafka.streams.state.internals.RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapter;
 import org.apache.kafka.streams.state.internals.StreamThreadStateStoreProvider;
+import org.apache.kafka.common.errors.TimeoutException;
 import org.slf4j.Logger;
 
 import java.time.Duration;
@@ -95,7 +96,6 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
-import org.apache.kafka.common.errors.TimeoutException;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 
@@ -1002,11 +1002,11 @@ public class KafkaStreams implements AutoCloseable {
      * cache size specified in configuration {@link StreamsConfig#CACHE_MAX_BYTES_BUFFERING_CONFIG}.
      *
      * @param timeout The the length of time to wait for the thread to shutdown
-     * @throws TimeoutException if the thread does not stop in time
+     * @throws org.apache.kafka.common.errors.TimeoutException if the thread does not stop in time
      * @return name of the removed stream thread or empty if a stream thread could not be removed because
      *         no stream threads are alive
      */
-    public Optional<String> removeStreamThread(final Duration timeout) throws TimeoutException {
+    public Optional<String> removeStreamThread(final Duration timeout) {
         final String msgPrefix = prepareMillisCheckFailMsgPrefix(timeout, "timeout");
         final long timeoutMs = validateMillisecondDuration(timeout, msgPrefix);
         return removeStreamThread(timeoutMs);
@@ -1014,6 +1014,7 @@ public class KafkaStreams implements AutoCloseable {
 
     private Optional<String> removeStreamThread(final long timeoutMs) throws TimeoutException {
         final long begin = time.milliseconds();
+        boolean timeout = false;
         if (isRunningOrRebalancing()) {
             synchronized (changeThreadCount) {
                 // make a copy of threads to avoid holding lock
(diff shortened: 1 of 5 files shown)


streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1005,11 +1007,56 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
                             || threads.size() == 1)) {
                         streamThread.shutdown();
                         if (!streamThread.getName().equals(Thread.currentThread().getName())) {
-                            streamThread.waitOnThreadState(StreamThread.State.DEAD);
+                            streamThread.waitOnThreadState(StreamThread.State.DEAD, -1);
                         }
                         threads.remove(streamThread);
                         final long cacheSizePerThread = getCacheSizePerThread(threads.size());
                         resizeThreadCache(cacheSizePerThread);
+                        Collection<MemberToRemove> membersToRemove = Collections.singletonList(new MemberToRemove(streamThread.getGroupInstanceID()));
```
wcarlson5: I ended up getting the `group.instance.id` from the streamThread

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -610,17 +610,32 @@ public void setStreamsUncaughtExceptionHandler(final java.util.function.Consumer
         this.streamsUncaughtExceptionHandler = streamsUncaughtExceptionHandler;
     }
 
-    public void waitOnThreadState(final StreamThread.State targetState) {
+    public boolean waitOnThreadState(final StreamThread.State targetState, long timeoutMs) {
+        if (timeoutMs < 0) {
```
wcarlson5: for the non timeout uses

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -1147,6 +1162,10 @@ public String toString(final String indent) {
         return indent + "\tStreamsThread threadId: " + getName() + "\n" + taskManager.toString(indent);
     }
 
+    public String getGroupInstanceID(){
+        return mainConsumer.groupMetadata().groupInstanceId().orElse("");
```
wcarlson5: It seems easier to get it form here than the config. It looked like I might have how to manipulate strings in that case

wcarlson5 (commented):
@ableegoldman I am going to add some tests but how does this look?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -91,6 +93,7 @@
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
```
ableegoldman: There's actually a kafka-specific version of `TimeoutException` that you should use to keep in line with other kafka APIs. It's `org.apache.kafka.common.errors.TimeoutException`

ableegoldman: (Tbh that drives me crazy, I once spent like 4 hours debugging something only to realize that I wasn't using the correct TimeoutException 😠 )

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -610,17 +610,32 @@ public void setStreamsUncaughtExceptionHandler(final java.util.function.Consumer
         this.streamsUncaughtExceptionHandler = streamsUncaughtExceptionHandler;
     }
 
-    public void waitOnThreadState(final StreamThread.State targetState) {
+    public boolean waitOnThreadState(final StreamThread.State targetState, long timeoutMs) {
+        if (timeoutMs < 0) {
+            timeoutMs = 0;
+        } else if (timeoutMs == 0) {
+            timeoutMs = Long.MAX_VALUE;
```
ableegoldman: We definitely shouldn't modify the passed in timeout like this -- a user should be able to pass in `0` to mean "don't block at all". Mysteriously blocking forever when they do so would be pretty weird

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -1147,6 +1162,10 @@ public String toString(final String indent) {
         return indent + "\tStreamsThread threadId: " + getName() + "\n" + taskManager.toString(indent);
     }
 
+    public String getGroupInstanceID() {
```
cadonna: Why not an `Optional`?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
ableegoldman: I think if you fix the semantics of `removeStreamThread()` to match that of `close()` then there's no need for a `-1` sentinel, in which case we should just throw an `IllegalArgumentException` here (or it's probably better to check and throw that in the actual `removeStreamThread(timeout)` call to fail fast

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
cadonna: I would do it the same way.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1005,11 +1008,60 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
                             || threads.size() == 1)) {
                         streamThread.shutdown();
                         if (!streamThread.getName().equals(Thread.currentThread().getName())) {
-                            streamThread.waitOnThreadState(StreamThread.State.DEAD);
+                            streamThread.waitOnThreadState(StreamThread.State.DEAD, -1);
```
ableegoldman: To be consistent with the semantics of `KafkaStreams#close`, the overload with no parameter should probably default to be fully blocking, ie with a timeout of `Long.MAX_VALUE`. Also, to avoid duplicate code, I would just have this method call `removeStreamThread(final Duration timeout)` instead of doing everything twice. Again, something like what we do for `#close`

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1005,11 +1008,60 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+                            if (!streamThread.waitOnThreadState(StreamThread.State.DEAD, timeoutMs)) {
+                                log.warn("Thread " + streamThread.getName() + " did not stop in the allotted time");
+                                throw new TimeoutException("Thread " + streamThread.getName() + " did not stop in the allotted time");
+                            }
+                        }
+                        threads.remove(streamThread);
+                        final long cacheSizePerThread = getCacheSizePerThread(threads.size());
+                        resizeThreadCache(cacheSizePerThread);
+                        Collection<MemberToRemove> membersToRemove = Collections.singletonList(new MemberToRemove(streamThread.getGroupInstanceID()));
+                        adminClient.removeMembersFromConsumerGroup(config.getString(StreamsConfig.APPLICATION_ID_CONFIG), new RemoveMembersFromConsumerGroupOptions(membersToRemove));
```
ableegoldman: Ok, this is going to be a little tricky...`removeMembersFromConsumerGroup` is async so we have two options. (1) just ignore the returned result and hope that it succeeded, or (2) check the returned `KafkaFuture` and wait/make sure that it succeeded.

Probably we should go with (2) and just apply the remaining time of the timeout. If you haven't mucked around with the KafkaFuture class before, I believe `KafkaFuture#get(long timeout, TimeUnit unit)` is what you'd need here

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1005,11 +1008,60 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+                        if (!streamThread.getName().equals(Thread.currentThread().getName())) {
+                            if (!streamThread.waitOnThreadState(StreamThread.State.DEAD, timeoutMs)) {
+                                log.warn("Thread " + streamThread.getName() + " did not stop in the allotted time");
+                                throw new TimeoutException("Thread " + streamThread.getName() + " did not stop in the allotted time");
+                            }
+                        }
+                        threads.remove(streamThread);
+                        final long cacheSizePerThread = getCacheSizePerThread(threads.size());
+                        resizeThreadCache(cacheSizePerThread);
+                        Collection<MemberToRemove> membersToRemove = Collections.singletonList(new MemberToRemove(streamThread.getGroupInstanceID()));
```
ableegoldman: I'm not sure how `removeMembersFromConsumerGroup` would behave if you passed in `""` as the `group.instance.id`, do you know? If not then let's just be safe and check what `streamThread.getGroupInstanceID()` returns, and skip this call if there is no group.instance.id (ie if not static)

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: Either way we need to deal with it. I thought it would be easier to just do it once. But It probably better practice to handle it later. I will change it to Optional

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: I did not know that. good catch

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: ah yeah, I had to fix this when I was writing my test

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: good idea, I reworked that a bit

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: that works

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Sounds good. How should you handle the `ExecutionException`?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1005,11 +1036,28 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
                         threads.remove(streamThread);
                         final long cacheSizePerThread = getCacheSizePerThread(threads.size());
                         resizeThreadCache(cacheSizePerThread);
+                        if (streamThread.getGroupInstanceID().isPresent()) {
+                            final MemberToRemove memberToRemove = new MemberToRemove(streamThread.getGroupInstanceID().get());
+                            final Collection<MemberToRemove> membersToRemove = Collections.singletonList(memberToRemove);
+                            final RemoveMembersFromConsumerGroupResult removeMembersFromConsumerGroupResult = adminClient.removeMembersFromConsumerGroup(config.getString(StreamsConfig.APPLICATION_ID_CONFIG), new RemoveMembersFromConsumerGroupOptions(membersToRemove));
+                            try {
+                                removeMembersFromConsumerGroupResult.memberResult(memberToRemove).get(timeoutMs - begin, TimeUnit.MILLISECONDS);
+                            } catch (final java.util.concurrent.TimeoutException e) {
```
wcarlson5, line 1047: have to make this a kafkaTimeout

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -88,9 +91,11 @@
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.UUID;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import org.apache.kafka.common.errors.TimeoutException;
```
ableegoldman: nit: move the import to the other `o.a.k.*` imports

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -997,19 +1002,63 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+     * <p>
+     * The removed stream thread is gracefully shut down. This method does not specify which stream
+     * thread is shut down.
+     * <p>
+     * Since the number of stream threads decreases, the sizes of the caches in the remaining stream
+     * threads are adapted so that the sum of the cache sizes over all stream threads equals the total
+     * cache size specified in configuration {@link StreamsConfig#CACHE_MAX_BYTES_BUFFERING_CONFIG}.
+     *
+     * @param timeout The the length of time to wait for the thread to shutdown
+     * @throws TimeoutException if the thread does not stop in time
```
ableegoldman:
```suggestion
     * @throws org.apache.kafka.common.errors.TimeoutException if the thread does not stop in time
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -997,19 +1002,63 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+     * @return name of the removed stream thread or empty if a stream thread could not be removed because
+     *         no stream threads are alive
+     */
+    public Optional<String> removeStreamThread(final Duration timeout) throws TimeoutException {
```
ableegoldman: We generally don't explicitly make this part of the API, and just inform users through the javadocs as you've done
```suggestion
    public Optional<String> removeStreamThread(final Duration timeout) {
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
ableegoldman: Ugh I forgot that `KafkaFuture` still throws a regular java TimeoutException. Such a mess -- btw we should log an error here (or a warn?) before rethrowing

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -319,9 +319,9 @@ private void prepareStreamThread(final StreamThread thread, final boolean termin
                 StreamThread.State.PARTITIONS_ASSIGNED);
             return null;
         }).anyTimes();
+        EasyMock.expect(thread.getGroupInstanceID()).andReturn(Optional.empty()).anyTimes();
```
ableegoldman:
```suggestion
        EasyMock.expect(thread.getGroupInstanceID()).andStubReturn(Optional.empty());
```

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -180,6 +182,19 @@ public void shouldRemoveStreamThread() throws Exception {
         }
     }
 
+    @Test
+    public void shouldnNotRemoveStreamThreadWithTimeout() throws Exception {
```
ableegoldman:
```suggestion
    public void shouldNotRemoveStreamThreadWithinTimeout() throws Exception {
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -997,19 +1002,63 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
-                            streamThread.waitOnThreadState(StreamThread.State.DEAD);
+                            if (!streamThread.waitOnThreadState(StreamThread.State.DEAD, timeoutMs)) {
+                                log.warn("Thread " + streamThread.getName() + " did not stop in the allotted time");
+                                throw new TimeoutException("Thread " + streamThread.getName() + " did not stop in the allotted time");
+                            }
                         }
                         threads.remove(streamThread);
                         final long cacheSizePerThread = getCacheSizePerThread(threads.size());
                         resizeThreadCache(cacheSizePerThread);
+                        if (groupInstanceID.isPresent()) {
```
wcarlson5: Ok, so we just do something like if (groupInstanceID.isPresent() && !streamThread.getName().equals(Thread.currentThread().getName())  when deciding whether to remove it from the group?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: We should. And I think maybe we should log the original stack trace

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -180,6 +182,19 @@ public void shouldRemoveStreamThread() throws Exception {
...
+        try (final KafkaStreams kafkaStreams = new KafkaStreams(builder.build(), properties)) {
+            addStreamStateChangeListener(kafkaStreams);
+            startStreamsAndWaitForRunning(kafkaStreams);
+
+            final int oldThreadCount = kafkaStreams.localThreadsMetadata().size();
+            stateTransitionHistory.clear();
+            assertThrows(TimeoutException.class, () -> kafkaStreams.removeStreamThread(Duration.ZERO.minus(DEFAULT_DURATION)));
```
ableegoldman, line 207: It's a bit weird to test this by passing in a negative timeout but I don't have any good ideas for forcing it to exceed the timeout 😕

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -997,19 +1002,63 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
                 for (final StreamThread streamThread : new ArrayList<>(threads)) {
                     if (streamThread.isAlive() && (!streamThread.getName().equals(Thread.currentThread().getName())
                             || threads.size() == 1)) {
+                        final Optional<String> groupInstanceID = streamThread.getGroupInstanceID();
                         streamThread.shutdown();
                         if (!streamThread.getName().equals(Thread.currentThread().getName())) {
-                            streamThread.waitOnThreadState(StreamThread.State.DEAD);
+                            if (!streamThread.waitOnThreadState(StreamThread.State.DEAD, timeoutMs)) {
+                                log.warn("Thread " + streamThread.getName() + " did not stop in the allotted time");
+                                throw new TimeoutException("Thread " + streamThread.getName() + " did not stop in the allotted time");
```
ableegoldman: Hm actually now that I think about it, we should probably continue with the cleanup to leave the app in a good state even if the shutdown is taking a long time. eg just store the exception for now and then rethrow it at the end of this method -- WDYT?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
ableegoldman: Yeah I think that makes sense here

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -180,6 +182,19 @@ public void shouldRemoveStreamThread() throws Exception {
         }
```
ableegoldman, line 183: One last thing, can you add a version of the `shouldRemoveStreamThread()` test that uses static membership?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Hmmmm. That is interesting. I am not sure. If the thread hasn't been removed then we don't want to resize the cache. The timeout is essentially saying that removing the thread failed. So is it right to then remove it anyways?

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
wcarlson5: I don't either...

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
ableegoldman: It does seem like kind of a gray area. Still, the TimeoutException isn't necessarily saying that it failed, just that we didn't wait long enough for it to finish the shutdown. But we have at least definitely initiated the shutdown -- besides, if the thread really is stuck in its shutdown then it's probably a benefit to go ahead with the `removeMembersFromConsumerGroup` call to get it kicked out all the sooner.

But, in the end, we really make no guarantees about the application should a user choose to ignore the  TimeoutException (though they absolutely can). I can imagine that some users might choose to just swallow it and decide that they don't care if the shutdown is taking a long time. It's hard to say

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
wcarlson5: But it isn't consistent because if the thread removes itself then the timeout its started

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
wcarlson5: yes

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: Okay I buy it I'll delay the exception

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
cadonna: Have you thought of using the `KafkaStreams` constructor that takes a `Time` object? You can pass `MockTime` to the constructor and progress time as you want.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -983,19 +988,68 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+                                timeout = true;
+                            }
                         }
                         threads.remove(streamThread);
                         final long cacheSizePerThread = getCacheSizePerThread(threads.size());
                         resizeThreadCache(cacheSizePerThread);
+                        if (groupInstanceID.isPresent() && !streamThread.getName().equals(Thread.currentThread().getName())) {
+                            final MemberToRemove memberToRemove = new MemberToRemove(streamThread.getGroupInstanceID().get());
+                            final Collection<MemberToRemove> membersToRemove = Collections.singletonList(memberToRemove);
+                            final RemoveMembersFromConsumerGroupResult removeMembersFromConsumerGroupResult = adminClient.removeMembersFromConsumerGroup(config.getString(StreamsConfig.APPLICATION_ID_CONFIG), new RemoveMembersFromConsumerGroupOptions(membersToRemove));
```
cadonna: This line is a bit long.
 
```suggestion
                            final RemoveMembersFromConsumerGroupResult removeMembersFromConsumerGroupResult = 
                                adminClient.removeMembersFromConsumerGroup(
                                    config.getString(StreamsConfig.APPLICATION_ID_CONFIG), 
                                    new RemoveMembersFromConsumerGroupOptions(membersToRemove)
                                );
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -983,19 +988,68 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+                            if (!streamThread.waitOnThreadState(StreamThread.State.DEAD, timeoutMs)) {
+                                log.warn("Thread " + streamThread.getName() + " did not stop in the allotted time");
+                                timeout = true;
+                            }
                         }
                         threads.remove(streamThread);
                         final long cacheSizePerThread = getCacheSizePerThread(threads.size());
                         resizeThreadCache(cacheSizePerThread);
+                        if (groupInstanceID.isPresent() && !streamThread.getName().equals(Thread.currentThread().getName())) {
+                            final MemberToRemove memberToRemove = new MemberToRemove(streamThread.getGroupInstanceID().get());
```
cadonna:
```suggestion
                            final MemberToRemove memberToRemove = new MemberToRemove(groupInstanceID.get());
```

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -1074,6 +1089,9 @@ private void completeShutdown(final boolean cleanRun) {
         streamsMetrics.removeAllThreadLevelSensors(getName());
 
         setState(State.DEAD);
+        if (leaveGroup.get()) {
+            mainConsumer.unsubscribe();
```
cadonna: It is illegal to unsubscribe a closed consumer.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -1158,6 +1176,14 @@ public String toString(final String indent) {
         return indent + "\tStreamsThread threadId: " + getName() + "\n" + taskManager.toString(indent);
     }
 
+    public Optional<String> getGroupInstanceID() {
+        return getGroupInstanceID;
+    }
+
+    public void leaveGroup() {
+        this.leaveGroup.set(true);
```
cadonna: Could you please rename the method and the field to something like `requestLeaveGroupDuringShutdown()` and `leaveGroupRequested`?

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -983,19 +988,68 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
                 for (final StreamThread streamThread : new ArrayList<>(threads)) {
                     if (streamThread.isAlive() && (!streamThread.getName().equals(Thread.currentThread().getName())
                             || threads.size() == 1)) {
+                        final Optional<String> groupInstanceID = streamThread.getGroupInstanceID();
+                        streamThread.leaveGroup();
                         streamThread.shutdown();
                         if (!streamThread.getName().equals(Thread.currentThread().getName())) {
-                            streamThread.waitOnThreadState(StreamThread.State.DEAD);
+                            if (!streamThread.waitOnThreadState(StreamThread.State.DEAD, timeoutMs)) {
+                                log.warn("Thread " + streamThread.getName() + " did not stop in the allotted time");
```
cadonna: nit:
```suggestion
                                log.warn("Thread " + streamThread.getName() + " did not shutdown in the allotted time");
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -983,19 +988,68 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+    }
+
+    private Optional<String> removeStreamThread(final long timeoutMs) throws TimeoutException {
+        final long begin = time.milliseconds();
+        boolean timeout = false;
         if (isRunningOrRebalancing()) {
             synchronized (changeThreadCount) {
                 // make a copy of threads to avoid holding lock
                 for (final StreamThread streamThread : new ArrayList<>(threads)) {
                     if (streamThread.isAlive() && (!streamThread.getName().equals(Thread.currentThread().getName())
```
cadonna: Since we use this check mutliple times, could you please extract `!streamThread.getName().equals(Thread.currentThread().getName()` to a variable named `callingThreadIsNotCurrentStreamThread` or similar.

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -983,19 +988,68 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+                            final Collection<MemberToRemove> membersToRemove = Collections.singletonList(memberToRemove);
+                            final RemoveMembersFromConsumerGroupResult removeMembersFromConsumerGroupResult = adminClient.removeMembersFromConsumerGroup(config.getString(StreamsConfig.APPLICATION_ID_CONFIG), new RemoveMembersFromConsumerGroupOptions(membersToRemove));
+                            try {
+                                removeMembersFromConsumerGroupResult.memberResult(memberToRemove).get(timeoutMs - begin, TimeUnit.MILLISECONDS);
+                            } catch (final java.util.concurrent.TimeoutException e) {
+                                throw new TimeoutException(e.getMessage());
+                            } catch (final InterruptedException e) {
+                                Thread.currentThread().interrupt();
+                            } catch (final ExecutionException e) {
+                                e.printStackTrace();
```
cadonna: Shall we throw the cause of the `ExecutionException` and log an error?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
cadonna: Or shall we simply add the following overload `shutdown(final boolean shouldUnsubscribe)` instead of a new method? If yes, I would still rename the field as proposed above.

streams/src/test/java/org/apache/kafka/streams/KafkaStreamsTest.java
```diff
@@ -330,12 +331,12 @@ private void prepareStreamThread(final StreamThread thread, final boolean termin
                 Collections.emptySet()
             )
         ).anyTimes();
-        EasyMock.expect(thread.threadMetadata()).andStubReturn(threadMetadata);
-        thread.waitOnThreadState(StreamThread.State.DEAD);
-        EasyMock.expectLastCall().anyTimes();
+        EasyMock.expect(thread.waitOnThreadState(StreamThread.State.DEAD, Long.MAX_VALUE)).andReturn(true).anyTimes();
```
cadonna:
```suggestion
        EasyMock.expect(thread.waitOnThreadState(StreamThread.State.DEAD, Long.MAX_VALUE)).andStubReturn(true);
```

cadonna (commented):
@wcarlson5 Thanks for the PR!

Here is my feedback!

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: good to know

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
wcarlson5: Either way we will need to set a flag and let thread take care of it so I will just rename it

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: sure

streams/src/test/java/org/apache/kafka/streams/integration/AdjustStreamThreadCountTest.java
```diff
@@ -180,6 +182,31 @@ public void shouldRemoveStreamThread() throws Exception {
...
     }
 
+    @Test
+    public void shouldRemoveStreamThreadWithStaticMembership() throws Exception {
+        properties.put("group.instance.id", "test");
```
ableegoldman:
```suggestion
        properties.put(ConsumerConfig.GROUP_INSTANCE_ID_CONFIG, "member-A");
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -69,6 +73,7 @@
 import org.apache.kafka.streams.state.internals.QueryableStoreProvider;
 import org.apache.kafka.streams.state.internals.RocksDBGenericOptionsToDbOptionsColumnFamilyOptionsAdapter;
 import org.apache.kafka.streams.state.internals.StreamThreadStateStoreProvider;
+import org.apache.kafka.common.errors.TimeoutException;
```
ableegoldman: nit: put this import above with the others (IDE often misplaces these since we follow a weird import ordering in places)

ableegoldman (approved):
Two minor nits but overall LGTM

ableegoldman:
FYI
```
15:42:08  Execution failed for task ':streams:checkstyleMain'.
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -983,19 +989,75 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+                            final RemoveMembersFromConsumerGroupResult removeMembersFromConsumerGroupResult = 
+                                adminClient.removeMembersFromConsumerGroup(
+                                    config.getString(StreamsConfig.APPLICATION_ID_CONFIG), 
+                                    new RemoveMembersFromConsumerGroupOptions(membersToRemove)
+                                );
+                            try {
+                                removeMembersFromConsumerGroupResult.memberResult(memberToRemove).get(timeoutMs - begin, TimeUnit.MILLISECONDS);
+                            } catch (final java.util.concurrent.TimeoutException e) {
+                                log.warn("Rethrowing a java TimeoutException as a kafka TimeoutException " + Arrays.toString(e.getStackTrace()));
+                                throw new TimeoutException(e.getMessage());
```
cadonna: nit: I would throw it as an error since I do not see any difference in severity compared with the execution exception. I think if you add the java timeout exception as the cause of the Kafka timeout exception, it will print also the stack trace of the Kafka timeout exception with `Caused by`. However, it will print the stack trace later in the logs. But that is OK, IMO, and we do that also in other places I guess.  

```suggestion
                                log.error("Could not remove static member {} from consumer group {} due to a timeout: {}",
                                    groupInstanceID.get(), config.getString(StreamsConfig.APPLICATION_ID_CONFIG), e);
                                throw new TimeoutException(e.getMessage(), e);
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -983,19 +989,75 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+                            } catch (final InterruptedException e) {
+                                Thread.currentThread().interrupt();
+                            } catch (final ExecutionException e) {
+                                log.error("Getting the member result threw this ExecutionException: " + e.getMessage());
+                                throw new StreamsException("Could not remove member from group for the following reason: ", e.getCause());
```
cadonna: nit:
Just added some more information to the messages.

```suggestion
                                log.error("Could not remove static member {} from consumer group {} due to: {}",
                                    groupInstanceID.get(), config.getString(StreamsConfig.APPLICATION_ID_CONFIG), e);
                                throw new StreamsException("Could not remove static member {} from consumer group {} the following reason: ", e.getCause());
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -983,19 +989,75 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
+     * threads are adapted so that the sum of the cache sizes over all stream threads equals the total
+     * cache size specified in configuration {@link StreamsConfig#CACHE_MAX_BYTES_BUFFERING_CONFIG}.
+     *
+     * @param timeout The the length of time to wait for the thread to shutdown
+     * @throws org.apache.kafka.common.errors.TimeoutException if the thread does not stop in time
+     * @return name of the removed stream thread or empty if a stream thread could not be removed because
+     *         no stream threads are alive
+     */
+    public Optional<String> removeStreamThread(final Duration timeout) {
+        final String msgPrefix = prepareMillisCheckFailMsgPrefix(timeout, "timeout");
```
cadonna, line 1010: I thought we agreed to not allow non-positive timeouts, i.e., `timeout <= Duration.ZERO`, for now. Or did I miss something?

cadonna (commented):
Thanks for the updates, @wcarlson5 !

I had just some minor comments on the log and exception messages and a major about `timeout <= Duration.ZERO`

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
wcarlson5: We are not allowing them as they will timeout but we are not stoping someone from entering them

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -1046,13 +1045,20 @@ private StreamThread createAndAddStreamThread(final long cacheSizePerThread, fin
...
                                 Thread.currentThread().interrupt();
                             } catch (final ExecutionException e) {
-                                log.error("Getting the member result threw this ExecutionException: " + e.getMessage());
-                                throw new StreamsException("Could not remove member from group for the following reason: ", e.getCause());
+                                log.error("Could not remove static member {} from consumer group {} due to: {}",
+                                        groupInstanceID.get(), config.getString(StreamsConfig.APPLICATION_ID_CONFIG), e);
+                                throw new StreamsException(
+                                        "Could not remove static member " + groupInstanceID.get()
+                                            + " from consumer group " + config.getString(StreamsConfig.APPLICATION_ID_CONFIG)
+                                            + " the following reason: ",
```
cadonna:
```suggestion
                                            + " for the following reason: ",
```

streams/src/main/java/org/apache/kafka/streams/KafkaStreams.java
```diff
@@ -25,7 +25,7 @@
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerConfig;
-import org.apache.kafka.common.KafkaException;
+import org.apache.kafka.common.errors.TimeoutException;
```
cadonna, line 28: I think you need to move this line before line 47 to have the imports lexicographically sorted.

cadonna: I think you need this line before line 32 to have the imports lexicographically sorted.

wcarlson5: which is it 47 or 32?

cadonna (approved):
LGTM

ableegoldman:
All tests passed 🥳  Merging to trunk

PR #9985, opened 2021-01-27
KAFKA-12237: Support non-routable quorum voter addresses

aloknnikhil:
With KIP-595, we expect the RaftConfig to specify the quorum voter endpoints upfront on startup. In the general case, this works fine. However, for testing we need a more lazy approach that discovers the other voters in the quorum after startup (i.e. controller port bind). This approach also lends itself well to cases where we might have an observer that discovers the voter endpoints from, say a `DescribeQuorum` event.

### DIFF
--- a/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
+++ b/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
@@ -16,9 +16,6 @@
  */
 package kafka.raft
 
-import java.net.InetSocketAddress
-import java.util.concurrent.ConcurrentLinkedQueue
-import java.util.concurrent.atomic.AtomicInteger
 import kafka.common.{InterBrokerSendThread, RequestAndCompletionHandler}
 import kafka.utils.Logging
 import org.apache.kafka.clients.{ClientResponse, KafkaClient}
@@ -30,6 +27,8 @@ import org.apache.kafka.common.utils.Time
 import org.apache.kafka.raft.RaftConfig.InetAddressSpec
 import org.apache.kafka.raft.{NetworkChannel, RaftRequest, RaftResponse, RaftUtil}
 
+import java.util.concurrent.ConcurrentLinkedQueue
+import java.util.concurrent.atomic.AtomicInteger
 import scala.collection.mutable
 
 object KafkaNetworkChannel {
--- a/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
+++ b/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
@@ -34,7 +34,6 @@ import scala.collection.mutable
 
 object KafkaNetworkChannel {
 
-  val nonRoutableAddress = new InetSocketAddress("0.0.0.0", 0)
   private[raft] def buildRequest(requestData: ApiMessage): AbstractRequest.Builder[_ <: AbstractRequest] = {
     requestData match {
       case voteRequest: VoteRequestData =>
--- a/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
+++ b/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
@@ -34,6 +34,7 @@ import scala.collection.mutable
 
 object KafkaNetworkChannel {
 
+  val nonRoutableAddress = new InetSocketAddress("0.0.0.0", 0)
   private[raft] def buildRequest(requestData: ApiMessage): AbstractRequest.Builder[_ <: AbstractRequest] = {
     requestData match {
       case voteRequest: VoteRequestData =>
@@ -165,9 +166,12 @@ class KafkaNetworkChannel(
     RaftUtil.errorResponse(apiKey, error)
   }
 
-  override def updateEndpoint(id: Int, address: InetSocketAddress): Unit = {
-    val node = new Node(id, address.getHostString, address.getPort)
-    endpoints.put(id, node)
+  def updateEndpoint(id: Int, address: InetSocketAddress): Unit = {
+    // We update the endpoint only if it's routable/valid
+    if (!address.equals(nonRoutableAddress)) {
+      val node = new Node(id, address.getHostString, address.getPort)
+      endpoints.put(id, node)
+    }
   }
 
   def start(): Unit = {
--- a/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
+++ b/core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
@@ -19,7 +19,6 @@ package kafka.raft
 import java.net.InetSocketAddress
 import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.atomic.AtomicInteger
-
 import kafka.common.{InterBrokerSendThread, RequestAndCompletionHandler}
 import kafka.utils.Logging
 import org.apache.kafka.clients.{ClientResponse, KafkaClient}
@@ -28,6 +27,7 @@ import org.apache.kafka.common.message._
 import org.apache.kafka.common.protocol.{ApiKeys, ApiMessage, Errors}
 import org.apache.kafka.common.requests._
 import org.apache.kafka.common.utils.Time
+import org.apache.kafka.raft.RaftConfig.InetAddressSpec
 import org.apache.kafka.raft.{NetworkChannel, RaftRequest, RaftResponse, RaftUtil}
 
 import scala.collection.mutable
@@ -166,12 +166,9 @@ class KafkaNetworkChannel(
     RaftUtil.errorResponse(apiKey, error)
   }
 
-  def updateEndpoint(id: Int, address: InetSocketAddress): Unit = {
-    // We update the endpoint only if it's routable/valid
-    if (!address.equals(nonRoutableAddress)) {
-      val node = new Node(id, address.getHostString, address.getPort)
-      endpoints.put(id, node)
-    }
+  def updateEndpoint(id: Int, spec: InetAddressSpec): Unit = {
+    val node = new Node(id, spec.address.getHostString, spec.address.getPort)
+    endpoints.put(id, node)
   }
 
   def start(): Unit = {
--- a/core/src/main/scala/kafka/raft/RaftManager.scala
+++ b/core/src/main/scala/kafka/raft/RaftManager.scala
@@ -29,10 +29,10 @@ import org.apache.kafka.common.protocol.ApiMessage
 import org.apache.kafka.common.requests.RequestHeader
 import org.apache.kafka.common.security.JaasContext
 import org.apache.kafka.common.utils.{LogContext, Time}
+import org.apache.kafka.raft.RaftConfig.{AddressSpec, InetAddressSpec}
 import org.apache.kafka.raft.{FileBasedStateStore, KafkaRaftClient, RaftClient, RaftConfig, RaftRequest, RecordSerde}
 
 import java.io.File
-import java.net.InetSocketAddress
 import java.nio.file.Files
 import java.util
 import java.util.OptionalInt
@@ -119,10 +119,18 @@ class KafkaRaftManager[T](
   private val raftIoThread = new RaftIoThread(raftClient)
 
   def startup(): Unit = {
-    // Update the voter endpoints with what's in RaftConfig
-    val voterAddresses: util.Map[Integer, InetSocketAddress] = raftConfig.quorumVoterConnections
+    // Update the voter endpoints (if valid) with what's in RaftConfig
+    val voterAddresses: util.Map[Integer, AddressSpec] = raftConfig.quorumVoterConnections
     for (voterAddressEntry <- voterAddresses.entrySet.asScala) {
-      netChannel.updateEndpoint(voterAddressEntry.getKey, voterAddressEntry.getValue)
+      voterAddressEntry.getValue match {
+        case spec: InetAddressSpec => {
+          netChannel.updateEndpoint(voterAddressEntry.getKey, spec)
+        }
+        case invalid: AddressSpec => {
+          logger.warn(s"Skipping channel update for destination ID: ${voterAddressEntry.getKey} " +
+            s"because of invalid endpoint: ${invalid.address().toString}")
+        }
+      }
     }
     netChannel.start()
     raftIoThread.start()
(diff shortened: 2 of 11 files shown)


core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
```diff
@@ -34,6 +34,7 @@ import scala.collection.mutable
 
 object KafkaNetworkChannel {
 
+  val nonRoutableAddress = new InetSocketAddress("0.0.0.0", 0)
```
hachikuji: Let me suggest an alternative for the sake of argument. Currently, `RaftConfig.parseVoterConnections` return `Map<Integer, InetSocketAddress>`. This works for the case we're interested in, but there is a risk of our sentinel non-routable address leaking into unexpected cases (a common source of bugs in Kafkaland). Alternatively, what if we add something like this to `RaftConfig`:

```java
public class RaftConfig {
  ...

  public Map<Integer, AddressSpec> quorumVoterConnections();

  public static interface AddressSpec {
  }

  public static class InetAddressSpec implements AddressSpec {
    final InetSocketAddress address;
  }

  public static class UnknownAddressSpec implements AddressSpec {
  }
}
```

The advantage is that this lets the type checker help us ensure that we are checking for a sentinel.

raft/src/test/java/org/apache/kafka/raft/MockNetworkChannel.java
```diff
@@ -25,20 +25,25 @@
...
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 public class MockNetworkChannel implements NetworkChannel {
     private final AtomicInteger correlationIdCounter;
+    private final Map<Integer, InetSocketAddress> addressCache;
```
hachikuji: I don't think we are using the address here. Can we use `Set<Integer>`? Potentially we could even get rid of this collection. It was more useful when the RaftClient itself was expected to discover the voter endpoints.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -208,8 +209,9 @@ public KafkaRaftClient(
         int fetchMaxWaitMs,
         OptionalInt nodeId,
         LogContext logContext,
-        Random random
-    ) {
+        Random random,
+        RaftConfig raftConfig
+    ) throws IOException {
```
hachikuji: Is anything in here throwing `IOException`?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -36,7 +36,9 @@
     public static final String QUORUM_VOTERS_CONFIG = QUORUM_PREFIX + "voters";
     public static final String QUORUM_VOTERS_DOC = "Map of id/endpoint information for " +
         "the set of voters in a comma-separated list of `{id}@{host}:{port}` entries. " +
-        "For example: `1@localhost:9092,2@localhost:9093,3@localhost:9094`";
+        "For example: `1@localhost:9092,2@localhost:9093,3@localhost:9094.`" +
+        "If the voter endpoints are not known at startup, a non-routable address can be provided instead." +
```
hachikuji: Perhaps we can keep this as an internal feature for now. It is not something that a user would be able to leverage. We can document it in the class javadoc perhaps.

raft/src/test/java/org/apache/kafka/raft/MockNetworkChannel.java
hachikuji: On second thought, it seems worth keeping this as a set. It helps us ensure that no requests are sent to non-voters. I would just change the field name to `Set<Integer> voterIds`.

aloknnikhil: Ack. I considered it. Figured it might be useful to have the endpoints for any future tests. Looks like the MockNetworkChannel doesn't test anything endpoint specific. Will remove

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
aloknnikhil: Yea. This is an artifact from the `quorumState.initialize` change. Since that's moved down to the `client.initialize`, we can remove it.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Fair enough. I can move it there.

core/src/main/scala/kafka/raft/KafkaNetworkChannel.scala
aloknnikhil: Yea, good catch. The AddressSpec makes sense.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -76,7 +87,48 @@
     private final int electionBackoffMaxMs;
     private final int fetchTimeoutMs;
     private final int appendLingerMs;
-    private final Map<Integer, InetSocketAddress> voterConnections;
+    private final Map<Integer, AddressSpec> voterConnections;
+
+    public static abstract class AddressSpec {
+       public abstract InetSocketAddress address();
```
hachikuji: Do we need this in the abstract class? I was thinking we would only be able to access `InetSocketAddress` if the type is `InetAddressSpec`. Otherwise the type protection from `AddressSpec` loses its bite.

core/src/test/scala/unit/kafka/raft/KafkaNetworkChannelTest.scala
```diff
@@ -139,6 +145,25 @@ class KafkaNetworkChannelTest {
     }
   }
 
+  @Test
+  def testNonRoutableAddressUpdateRequest(): Unit = {
+    val destinationId = 2
+    assertThrows(classOf[IllegalArgumentException],
+      () => new InetAddressSpec(new InetSocketAddress("0.0.0.0", 0)))
```
hachikuji: Can we move this to `RaftConfigTest`? It's not really part of the behavior of `KafkaNetworkChannel`.

core/src/test/scala/unit/kafka/raft/KafkaNetworkChannelTest.scala
```diff
@@ -139,6 +145,25 @@ class KafkaNetworkChannelTest {
...
+
+    // Update channel with a valid endpoint
```
hachikuji: Not sure there's much value in the rest of this test. Seems effectively the same as `testSendAndReceiveOutboundRequest`.

core/src/main/scala/kafka/raft/RaftManager.scala
```diff
@@ -118,9 +119,20 @@ class KafkaRaftManager[T](
...
   def startup(): Unit = {
+    // Update the voter endpoints (if valid) with what's in RaftConfig
+    val voterAddresses: util.Map[Integer, AddressSpec] = raftConfig.quorumVoterConnections
+    for (voterAddressEntry <- voterAddresses.entrySet.asScala) {
+      voterAddressEntry.getValue match {
+        case spec: InetAddressSpec => {
+          netChannel.updateEndpoint(voterAddressEntry.getKey, spec)
+        }
+        case invalid: AddressSpec => {
+          logger.warn(s"Skipping channel update for destination ID: ${voterAddressEntry.getKey} " +
```
hachikuji: This could be `info` I think in the case of `UnknownAddressSpec`. It is expected behavior to skip the update. We could add a third case for unexpected `AddressSpec` types.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -76,7 +87,48 @@
...
+            this.address = address;
+        }
+
+        @Override
+        public InetSocketAddress address() {
+            return address;
+        }
+    }
+
+    public static class UnknownAddressSpec extends AddressSpec {
```
hachikuji: A common pattern for classes like this without any state is to create a static instance.
```java
  public static final UnknownAddressSpec INSTANCE = new UnknownAddressSpec();
```

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Yea, I did a second pass later and figured the address defeats the purpose of the type enforcement (especially returning an InetSocketAddress for any address spec). I reverted to an interface and have only InetSocketAddressSpec contain an InetSocketAddress

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Ack. Refactored as stated above.

core/src/test/scala/unit/kafka/raft/KafkaNetworkChannelTest.scala
aloknnikhil: All the tests in `RaftConfigTest` were migrated to `KafkaConfigTest` in an earlier PR. I'll move this also.

core/src/test/scala/unit/kafka/raft/KafkaNetworkChannelTest.scala
aloknnikhil: Fair enough. Removed.

core/src/main/scala/kafka/raft/RaftManager.scala
aloknnikhil: Ack.

core/src/main/scala/kafka/raft/RaftManager.scala
```diff
@@ -126,9 +126,13 @@ class KafkaRaftManager[T](
         case spec: InetAddressSpec => {
           netChannel.updateEndpoint(voterAddressEntry.getKey, spec)
         }
+        case _: UnknownAddressSpec => {
```
hachikuji: nit: the braces in these `case` matches are not needed

core/src/test/scala/unit/kafka/server/KafkaConfigTest.scala
```diff
@@ -1001,6 +1001,12 @@ class KafkaConfigTest {
     assertEquals(expectedVoters, raftConfig.quorumVoterConnections())
   }
 
+  @Test
+  def testNonRoutableAddressSpecException(): Unit = {
+    assertThrows(classOf[IllegalArgumentException],
+      () => new InetAddressSpec(new InetSocketAddress("0.0.0.0", 0)))
```
hachikuji: nit: I still think `RaftConfigTest` is a better home for this. It doesn't involve `KafkaConfig` at all. We could also just delete the test since it probably isn't buying us much.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -102,31 +120,23 @@ public boolean equals(Object obj) {
...
-    public static class InetAddressSpec extends AddressSpec {
-        private final InetSocketAddress address;
-
-        public InetAddressSpec(InetSocketAddress address) {
-            if (address.equals(UNROUTABLE_ADDRESS)) {
-                throw new IllegalArgumentException("Address not routable");
-            }
-            this.address = address;
+    public static class UnknownAddressSpec implements AddressSpec {
+        private UnknownAddressSpec() {
```
hachikuji, line 129: Since we have a private constructor, I think we will only have the `UNKNOWN_ADDRESS_SPEC_INSTANCE` instance. So can we rely on the default equals/hashcode?

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -89,8 +92,23 @@
...
-    public static abstract class AddressSpec {
-       public abstract InetSocketAddress address();
+    public interface AddressSpec {
+    }
+
+    public static class InetAddressSpec implements AddressSpec {
+        public final InetSocketAddress address;
+
+        public InetAddressSpec(InetSocketAddress address) {
+            if (address.equals(NON_ROUTABLE_ADDRESS)) {
```
hachikuji: Maybe worth adding a null check here?

core/src/main/scala/kafka/raft/RaftManager.scala
aloknnikhil: Ah makes sense. Removed.

core/src/test/scala/unit/kafka/server/KafkaConfigTest.scala
aloknnikhil: Yea, the only thing this is testing is that the `InetAddressSpec` never accepts a `0.0.0.0:0` as a parameter. We could remove it instead.

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Yea, that's true actually. Should work fine for the map equals test case containing an instance of the UnknownAddressSpec. Removed

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
aloknnikhil: Good catch! Added

raft/src/main/java/org/apache/kafka/raft/RaftConfig.java
```diff
@@ -99,7 +99,7 @@
         public final InetSocketAddress address;
 
         public InetAddressSpec(InetSocketAddress address) {
-            if (address.equals(NON_ROUTABLE_ADDRESS)) {
+            if (address != null && address.equals(NON_ROUTABLE_ADDRESS)) {
```
hachikuji: We also don't want to accept null addresses, right? I was thinking we could do this:
```java
if (address == null || address.equals(NON_ROUTABLE_ADDRESS)) {
  throw new IllegalArgumentException("Invalid address " + address);
}
```

aloknnikhil: Yea, I thought of that. But would there ever be a case where we'd need to construct the address spec with a null? Probably not. Might as well use a nullptr for the address spec at that point I guess. Fair enough, will change.

hachikuji: Currently our equals method assumes non-null addresses. I can't think of a case where we would want it.

aloknnikhil: Right. Makes sense. Fixed.

hachikuji (approved):
Thanks, LGTM

PR #9986, opened 2021-01-27
JUnit extensions for integration tests

mumrah:
This PR adds a set of annotations that can be used to generate one or more Kafka clusters for integration testing. The primary motivation for this is to reuse existing integration tests for both Zk and Raft based Kafka clusters. A secondary motivation is to refactor the complex test hierarchy that exists and instead provide reusable functionality to tests through helper classes.

# Annotations

A new `@ClusterTest` annotation is introduced which allows for a test to declaratively configure an underlying Kafka cluster.

```scala
@ClusterTest
def testSomething(): Unit = { ... }
```

This annotation has fields for cluster type and number of brokers, as well as commonly parameterized configurations. Arbitrary server properties can also be provided in the annotation:

```java
@ClusterTest(clusterType = Type.Zk, securityProtocol = "PLAINTEXT", properties = {
  @ClusterProperty(key = "inter.broker.protocol.version", value = "2.7-IV2"),
  @ClusterProperty(key = "socket.send.buffer.bytes", value = "10240"),
})
void testSomething() { ... }
```

Multiple `@ClusterTest` annotations can be given to generate more than one test invocation for the annotated method.

```scala
@ClusterTests(Array(
    @ClusterTest(securityProtocol = "PLAINTEXT"),
    @ClusterTest(securityProtocol = "SASL_PLAINTEXT")
))
def testSomething(): Unit = { ... }
```

A class-level `@ClusterTestDefaults` annotation is added to provide default values for `@ClusterTest` defined within the class. The intention here is to reduce repetitive annotation declarations and also make changing defaults easier for a class with many test cases.

# Dynamic Configuration

In order to allow for more flexible cluster configuration, a `@ClusterTemplate` annotation is also introduced. This annotation takes a single string value which references a static method on the test class. This method is used to produce any number of test configurations using a fluent builder style API.

```java
@ClusterTemplate("generateConfigs")
void testSomething() { ... }

static void generateConfigs(ClusterGenerator clusterGenerator) {
  clusterGenerator.accept(ClusterConfig.defaultClusterBuilder()
      .name("Generated Test 1")
      .serverProperties(props1)
      .ibp("2.7-IV1")
      .build());
  clusterGenerator.accept(ClusterConfig.defaultClusterBuilder()
      .name("Generated Test 2")
      .serverProperties(props2)
      .ibp("2.7-IV2")
      .build());
  clusterGenerator.accept(ClusterConfig.defaultClusterBuilder()
      .name("Generated Test 3")
      .serverProperties(props3)
      .build());
}
```

This "escape hatch" from the simple declarative style configuration makes it easy to dynamically configure clusters.

# JUnit Extension

One thing to note is that our "test*" methods are no longer _tests_, but rather they are test templates. We have added a JUnit extension called `ClusterForEach` which knows how to process these annotations in order to generate test invocations. Test classes that wish to make use of these annotations need to explicitly register this extension:

```scala
@ExtendWith(value = Array(classOf[ClusterForEach]))
class ApiVersionsRequestTest {
   ...
}
```

# JUnit Lifecycle

The lifecycle of a test class that is extended with `ClusterForEach` follows:

* JUnit discovers test template methods that are annotated with `@ClusterTest`, `@ClusterTests`, or `@ClusterTemplate`
* `ClusterForEach` is called for each of these template methods in order to generate some number of test invocations

For each generated invocation:
* Static `@BeforeAll` methods are called
* Test class is instantiated
* Non-static `@BeforeEach` methods are called
* Kafka Cluster is started
* Test method is invoked
* Kafka Cluster is stopped
* Non-static `@AfterEach` methods are called
* Static `@AfterAll` methods are called

`@BeforeEach` methods give an opportunity to setup additional test dependencies before the cluster is started. For example, in [SaslApiVersionsRequestTest](https://github.com/apache/kafka/pull/9986/files#diff-ba6b8459d0e1cb7e5672a259b8e58fa066ca3d656f71c79d3da809d6da6c4a33), a MiniKDC is started before Kafka comes up (in order to provide a SASL backend). As seen in the example `@BeforeEach` (and other places) can have parameters which are provided through JUnit dependency injection.

# Dependency Injection

A few classes are introduced to provide context to the underlying cluster and to provide reusable functionality that was previously garnered from the test hierarchy. 

* ClusterConfig: a mutable cluster configuration, includes cluster type, number of brokers, properties, etc
* ClusterInstance: a shim to the underlying class that actually runs the cluster, provides access to things like SocketServers
* IntegrationTestHelper: connection related functions taken from IntegrationTestHarness and BaseRequestTest

In order to have one of these objects injected, simply add it as a parameter to your test class, `@BeforeEach` method, or test method. 

| Injection | Class | BeforeEach | Test | Notes
| --- | --- | --- | --- | --- |
| ClusterConfig | yes | yes | yes* | Once in the test, changing config has no effect |
| ClusterInstance | yes* | no | yes | Injectable at class level for convenience, can only be accessed inside test |
| IntegrationTestHelper | yes | yes | yes | - |

```scala
@ExtendWith(value = Array(classOf[ClusterForEach]))
class SomeTestClass(helper: IntegrationTestHelper) {
 
  @BeforeEach
  def setup(config: ClusterConfig): Unit = {
    config.serverProperties().put("foo", "bar")
  }

  @ClusterTest
  def testSomething(cluster: ClusterInstance): Unit = {
    val topics = cluster.createAdminClient().listTopics()
  }
}
```

# Gradle

In order to overcome some compilation issues, the core module test compile phase is now using the Scala compile for both Java and Scala sources. This is because the new annotations and JUnit extension are written in Java, but depend on some Scala classes in "core".

```gradle
    test {
      java {
        srcDirs = []
      }
      scala {
        srcDirs = ["src/generated/java", "src/test/java", "src/test/scala"]
      }
    }
```

Since most of the core module test sources are Scala already, this should have little impact on build time. 

# KIP-500

As mentioned above, one of the main motivations of this work is to allow for easy reuse of test code between Raft and Zk modes. This will be accomplished using the `clusterType` field on the annotations. For example:

```scala
@ClusterTest(clusterType=Type.Both)
def testSomething(): Unit = { ... }
```

would generate two invocations: one for Zk and one for Raft. 

Once more of the KIP-500 code is merged into trunk, we will add a Raft-based test invocation context. The prototype version can be seen on the [original prototype PR](https://github.com/confluentinc/kafka/pull/492/files#diff-5517148d95f0c6ac63152b332014d8aefd705b55e1ae153adfe2603bb5cc8361)

# Gotchas
* Test methods annotated with JUnit's `@Test` will still be run, but no cluster will be started and no dependency injection will happen. This is generally not what you want
* Even though ClusterConfig is accessible and mutable inside the test method, changing it will have no affect on the cluster


mumrah:
Example of test report from one of the converted tests:

https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-9986/1/testReport/kafka.server/ApiVersionsRequestTest/

image

build.gradle
```diff
@@ -1011,7 +1011,10 @@ project(':core') {
     }
     test {
       java {
-        srcDirs = ["src/generated/java", "src/test/java"]
+        srcDirs = []
+      }
+      scala {
+        srcDirs = ["src/generated/java", "src/test/java", "src/test/scala"]
```
hachikuji: Do we need `src/generated/java` to be built with scala? I think it shouldn't have any dependence on anything in core.

core/src/test/java/kafka/test/ClusterInstance.java
```diff
@@ -0,0 +1,66 @@
...
+
+    ClusterConfig config();
+
+    ListenerName listener();
+
+    String brokerList();
+
+    Collection<SocketServer> brokers();
+
+    Collection<SocketServer> controllers();
```
hachikuji: Might be useful to have some documentation for these methods. I think `controllers` will always return empty list if this is a zk cluster?

core/src/test/java/kafka/test/ClusterForEachTest.java
```diff
@@ -0,0 +1,112 @@
...
+import kafka.test.annotation.Type;
+import kafka.test.junit.ClusterForEach;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+
+@ClusterTestDefaults(clusterType = Type.Zk)   // Set defaults for a few params in @ClusterTest(s)
+@ExtendWith(ClusterForEach.class)
```
hachikuji: Wow, so meta. Is it possible to have any validation that this gets executed?

core/src/test/java/kafka/test/junit/ZkClusterInvocationContext.java
```diff
@@ -0,0 +1,251 @@
...
+                        return Math.max(clusterConfig.brokers(), clusterConfig.controllers());
+                    }
+
+                    @Override
+                    public Option<File> trustStoreFile() {
+                        return OptionConverters.toScala(clusterConfig.trustStoreFile());
+                    }
+                };
+
+                // TODO consumer and producer configs
```
hachikuji: File a jira instead?

core/src/test/java/kafka/test/annotation/ClusterProperty.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Documented
+@Target({ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface ClusterProperty {
```
hachikuji: Maybe `ClusterConfig`? Would be nice to clarify the scope. I assume this is only for `KafkaConfig`?

core/src/test/java/kafka/test/annotation/ClusterTest.java
```diff
@@ -0,0 +1,45 @@
...
+    int brokers() default 0;
+    int controllers() default 0;
+    AutoStart autoStart() default AutoStart.Default;
+
+    String name() default "";
+    String securityProtocol() default "PLAINTEXT";
+    String listener() default "";
+    ClusterProperty[] serverProperties() default {};
+}
+
```
hachikuji: nit: unneeded newlines

core/src/test/java/kafka/test/junit/ClusterForEach.java
```diff
@@ -0,0 +1,219 @@
...
+ *     assertTrue(condition)
+ *   }
+ * }
+ * </pre>
+ *
+ * will generate two invocations of "someTest" (since ClusterType.Both was given). For each invocation, the test class
+ * SomeIntegrationTest will be instantiated, lifecycle methods (before/after) will be run, and "someTest" will be invoked.
+ *
+ **/
+public class ClusterForEach implements TestTemplateInvocationContextProvider {
```
hachikuji: Wonder if the name could be improved. It's boring, but how about `ClusterTestExtensions`?

```scala
@ExtendWith(Array(classOf[ClusterTestExtensions]))
class SomeIntegrationTest
```

core/src/test/java/kafka/test/ClusterInstance.java
mumrah: It will return just the one SocketServer of the broker which has the active controller

build.gradle
mumrah: The existing java test sourceSet referenced the generated sources, so I copied them over. I could try without and see what happens

core/src/test/java/kafka/test/annotation/ClusterProperty.java
mumrah: We already have `ClusterConfig` which is likely to get imported into tests, so we'd end up needing to qualify one of them.

core/src/test/java/kafka/test/ClusterForEachTest.java
mumrah: testClusterTest below exercises the class level annotation

core/src/test/java/kafka/test/ClusterInstance.java
hachikuji: Can we document that? Another one that is not so obvious is `listener`. What would this return if the test involved multiple listeners?

core/src/test/java/kafka/test/annotation/ClusterProperty.java
hachikuji: Maybe `ClusterConfigProperty` or `ClusterConfigOverride`?

build.gradle
hachikuji: Hmm, come to think of it, I'm not sure why "src/generated/java" is even in here. I tried removing it locally and it seems we don't need it. Maybe I'm missing something.

hachikuji (commented):
Replied to a few comments. Any chance we could pull the great description from the PR  into a package doc or something?

build.gradle
ijuma: Yeah, I don't think we need any of them since the generated classes are in the `clients` module.

core/src/test/scala/unit/kafka/server/SaslApiVersionsRequestTest.scala
```diff
@@ -16,83 +16,94 @@
...
-      validateApiVersionsResponse(apiVersionsResponse)
+      validateApiVersionsResponse(apiVersionsResponse, cluster.listener())
       sendSaslHandshakeRequestValidateResponse(socket)
     } finally {
       socket.close()
     }
   }
 
-  @Test
+  @ClusterTest(securityProtocol = "SASL_PLAINTEXT", clusterType = Type.Zk)
```
ijuma: It's a bit weird to repeat the security protocol in every test method given the class name. Can we not do it in the setup method?

mumrah: We could do it in a method specified by `@ClusterTemplate`, or we could add `securityProtocol ` to the `@ClusterTestDefaults` class-level annotation. I'll leave this for a follow-up PR

core/src/test/java/kafka/test/annotation/Type.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kafka.test.annotation;
+
+public enum Type {
```
cmccabe, line 23: Java Enums should be capitalized.  Also can we add a JavaDoc for each value describing what it is?

core/src/test/java/kafka/test/ClusterConfig.java
```diff
@@ -0,0 +1,207 @@
...
+ */
+public class ClusterConfig {
+
+    private final Type type;
+    private final int brokers;
+    private final int controllers;
+    private final String name;
+    private final boolean autoStart;
+
+    private final String securityProtocol;
```
cmccabe: What's the advantage of making this a string rather than using the SecurityProtocol enum?

core/src/test/java/kafka/test/ClusterInstance.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.kafka.clients.admin.Admin;
+import org.apache.kafka.common.network.ListenerName;
+
+import java.util.Collection;
+import java.util.Optional;
+import java.util.Properties;
+
+public interface ClusterInstance {
+
+    enum ClusterType {
```
cmccabe, line 30: Enum values should be capitalized

core/src/test/java/kafka/test/ClusterInstance.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * The cluster configuration used to create this cluster. Changing data in this instance through this accessor will
+     * have no affect on the cluster since it is already provisioned.
+     */
+    ClusterConfig config();
+
+    /**
+     * The listener for this cluster as configured by {@link ClusterTest} or by {@link ClusterConfig}. If
+     * unspecified by those sources, this will return the listener for the default security protocol PLAINTEXT
+     */
```
cmccabe, line 49: A cluster can have multiple listeners, right?  Is this intended to be the listener that should be used by clients?  If so, calling it the client listener makes sense.

(Technically we could have multiple client listeners too, but that's much less common)

Many clusters run with 3 listeners:
* client
* inter-broker NON-REPLICATION
* replication

and we are going to add a 4th:
* controller

core/src/test/java/kafka/test/ClusterInstance.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * The broker connect string which can be used by clients for bootstrapping
+     */
+    String brokerList();
+
+    /**
+     * A collection of all brokers in the cluster. In ZK-based clusters this will also include the broker which is
+     * acting as the controller (since ZK controllers serve both broker and controller roles).
+     */
+    Collection<SocketServer> brokers();
```
cmccabe: I would prefer to call this method something like `brokerSocketServers`.  After all, the `SocketServer` is just one small part of the broker, not the broker itself.

core/src/test/java/kafka/test/ClusterInstance.java
```diff
@@ -0,0 +1,98 @@
...
+
+    Admin createAdminClient(Properties configOverrides);
+
+    default Admin createAdminClient() {
+        return createAdminClient(new Properties());
+    }
+
+    void start();
+
+    void stop();
```
cmccabe, line 96: should this implement `AutoCloseable`?  Does code need to remember to close this cluster?

Or is that handled automatically by the framework...

core/src/test/java/kafka/test/ClusterInstance.java
mumrah: Yes, multiple listeners can be defined. This is something to solve/fix in this new framework. Only one security protocol and listener are exposed here in order to stay similar to how KafkaServerTestHarness works. It defines one security protocol and listener as members:

```
protected def securityProtocol: SecurityProtocol = SecurityProtocol.PLAINTEXT
protected def listenerName: ListenerName = ListenerName.forSecurityProtocol(securityProtocol)
``` 

Of course, more listeners can be defined through configuration, but this one is exposed by the test harness and used by many tests.

core/src/test/java/kafka/test/ClusterInstance.java
mumrah: Originally I left these out of the interface because the framework handle starting and stopping the cluster. However, after @ijuma's feedback, I added `boolean autoStart` to ClusterConfig along with these methods so that a test could control cluster startup/shutdown explicitly. 

If a test does not call `stop`, the framework will stop the cluster. Stopping it a second time does nothing (per the implementation).

core/src/test/java/kafka/test/ClusterConfig.java
mumrah: No particular reason. We can address this when we look more into issue of multiple listeners you point out below

cmccabe: If there's no reason to use a string then let's just use the `SecurityProtocol` enum.

core/src/test/java/kafka/test/ClusterConfig.java
```diff
@@ -0,0 +1,207 @@
...
+        this.securityProtocol = securityProtocol;
+        this.listenerName = listenerName;
+        this.trustStoreFile = trustStoreFile;
+    }
+
+    public Type clusterType() {
+        return type;
+    }
+
+    public int brokers() {
```
cmccabe: let's use "numBrokers" , etc. to reflect the fact that we're not returning the actual brokers.  Same for "controllers"

core/src/test/java/kafka/test/ClusterConfig.java
```diff
@@ -0,0 +1,207 @@
...
+        return brokers;
+    }
+
+    public int controllers() {
+        return controllers;
```
cmccabe, line 72: What does this return in ZK mode?

core/src/test/java/kafka/test/ClusterInstance.java
cmccabe: How about naming these "clientSecurityProtocol", "clientListenerName", etc. to reflect the fact that the clients will be using these security protocols, listener names, etc.  Then we can add more listeners later for specific tests without doing a big rename.

core/src/test/java/kafka/test/ClusterInstance.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * The listener for this cluster as configured by {@link ClusterTest} or by {@link ClusterConfig}. If
+     * unspecified by those sources, this will return the listener for the default security protocol PLAINTEXT
+     */
+    ListenerName listener();
+
+    /**
+     * The broker connect string which can be used by clients for bootstrapping
+     */
+    String brokerList();
```
cmccabe: I realize `brokerList` is the existing name, but given that we've standardized on `--bootstrap-server` for the command-line argument, maybe it makes sense to use that name here?

core/src/test/java/kafka/test/ClusterInstance.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * A collection of all controllers in the cluster. For ZK-based clusters, this will return the broker which is also
+     * currently the active controller. For Raft-based clusters, this will return all controller servers.
+     */
+    Collection<SocketServer> controllerSocketServers();
+
+    /**
+     * Any one of the broker servers.
+     */
+    Optional<SocketServer> anyBrokerSocketServer();
```
cmccabe: Why does this return an `Optional<>`?  Zero-node clusters seem like a corner case we don't really care about (we can just throw an exception there)....

core/src/test/java/kafka/test/ClusterInstance.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Any one of the controller servers.
+     */
+    Optional<SocketServer> anyControllerSocketServer();
```
cmccabe: Same question... can we just throw an exception if there are no controller socket servers?

core/src/test/java/kafka/test/ClusterInstance.java
cmccabe: ok

core/src/test/java/kafka/test/annotation/AutoStart.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kafka.test.annotation;
+
+public enum AutoStart {
```
cmccabe, line 20: capitalize enum names please.  also, do we need this enum?  Why does DEFAULT need to be distinct from YES?  Once we have a lot of tests depending on the default being "on" I can't see us changing this (or how having 3 values would help us change it if so...)

core/src/test/java/kafka/test/junit/ZkClusterInvocationContext.java
```diff
@@ -0,0 +1,250 @@
...
+            return ClusterType.ZK;
+        }
+
+        @Override
+        public ClusterConfig config() {
+            return config;
+        }
+
+        @Override
+        public Object getUnderlying() {
```
cmccabe: Hmm... why return `Object` rather than `IntegrationTestHarness`?

core/src/test/scala/integration/kafka/server/IntegrationTestHelper.scala
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.kafka.common.utils.Utils
+
+import java.io.{DataInputStream, DataOutputStream}
+import java.net.Socket
+import java.nio.ByteBuffer
+import java.util.Properties
+import scala.annotation.nowarn
+import scala.reflect.ClassTag
+
+class IntegrationTestHelper {
```
cmccabe: These are all static methods, right?  Can we just make this an "object" rather than a class and have something like `object IntegrationTestUtils` ?

core/src/test/java/kafka/test/ClusterInstance.java
mumrah: Yea we can ditch the Optional for these

core/src/test/java/kafka/test/annotation/AutoStart.java
mumrah: Since we can't use `null` as defaults for annotations, I added DEFAULT enum constants for cluster type and auto start attributes. When processing the `@ClusterTest` method annotations, if we see Type.DEFAULT or AutoStart.DEFAULT, we know the test method is saying "use the default". We could simply put the defaults right in the `ClusterTest` annotation, but then we would have an issue when method-level `@ClusterTest` and class-level `@ClusterTestDefaults` are both in use. How would we know which to use? 

I added AutoStart.DEFAULT in order to let the method-level annotation explicitly defer to the default. Here's an example:

```java
// indicate that for this test, we want to disable auto-start by default
@ClusterTestDefaults(autoStart = false)
public class TestClass {
  
  // but this test does want auto start
  @ClusterTest(autoStart = AutoStart.YES)
  public void test1() { }

  // this test is unnecessarily setting auto start as false
  @ClusterTest(autoStart = AutoStart.NO)
  public void test2() { }

  // this test will get its auto start value from the class level
  @ClusterTest
  public void test3() { }
}
```

An alternative would be to allow for a class-level annotation that completely overrides any method-level `@ClusterTest`, but that seems a little backwards.

cmccabe: Thanks for the explanation.  That makes sense.

core/src/test/java/kafka/test/ClusterConfig.java
mumrah: In ZK mode, number of controllers is basically ignored. I can add some logic to validate that someone hasn't mistakenly requested more than one controller for a ZK cluster.

cmccabe: ok

cmccabe:
LGTM.  Thanks, @mumrah

cmccabe (approved):

cmccabe:
looks like there is a checkstyle issue:
```
[2021-02-08T23:40:02.927Z] [Error] /home/jenkins/jenkins-agent/workspace/Kafka_kafka-pr_PR-9986/core/src/test/scala/unit/kafka/server/ApiVersionsRequestTest.scala:20: Unused import
[2021-02-08T23:40:02.927Z] [Error] /home/jenkins/jenkins-agent/workspace/Kafka_kafka-pr_PR-9986/core/src/test/scala/integration/kafka/api/AdminClientWithPoliciesIntegrationTest.scala:91: @nowarn annotation does not suppress any warnings
```

PR #9987, opened 2021-01-27
KAFKA-10895: Gracefully handle invalid JAAS configs (follow up fix)

C0urante:
Follow-up to https://github.com/apache/kafka/pull/9806

If an invalid JAAS config is present on the worker, invoking `Configuration::getConfiguration` throws an exception. The changes from #9806 cause that exception to be thrown during plugin scanning, which causes the worker to fail even if it is not configured to use the basic auth extension at all.

This follow-up handles invalid JAAS configurations more gracefully, and only throws them if the worker is actually configured to use the basic auth extension, at the time that the extension is instantiated and configured.

Two unit tests are added to test the green-path and red-path behavior of the extension when it encounters well-formed and ill-formed JAAS configurations, respectively.

### DIFF
--- a/connect/basic-auth-extension/src/main/java/org/apache/kafka/connect/rest/basic/auth/extension/BasicAuthSecurityRestExtension.java
+++ b/connect/basic-auth-extension/src/main/java/org/apache/kafka/connect/rest/basic/auth/extension/BasicAuthSecurityRestExtension.java
@@ -81,10 +81,21 @@ public class BasicAuthSecurityRestExtension implements ConnectRestExtension {
         }
     }
 
+    private final Supplier<Configuration> configuration;
+
+    public BasicAuthSecurityRestExtension() {
+        this(CONFIGURATION);
+    }
+
+    // For testing
+    BasicAuthSecurityRestExtension(Supplier<Configuration> configuration) {
+        this.configuration = configuration;
+    }
+
     @Override
     public void register(ConnectRestExtensionContext restPluginContext) {
         log.trace("Registering JAAS basic auth filter");
-        restPluginContext.configurable().register(new JaasBasicAuthFilter(CONFIGURATION.get()));
+        restPluginContext.configurable().register(new JaasBasicAuthFilter(configuration.get()));
         log.trace("Finished registering JAAS basic auth filter");
     }
 
@@ -96,7 +107,7 @@ public class BasicAuthSecurityRestExtension implements ConnectRestExtension {
     @Override
     public void configure(Map<String, ?> configs) {
         // If we failed to retrieve a JAAS configuration during startup, throw that exception now
-        CONFIGURATION.get();
+        configuration.get();
     }
 
     @Override
--- a/connect/basic-auth-extension/src/main/java/org/apache/kafka/connect/rest/basic/auth/extension/BasicAuthSecurityRestExtension.java
+++ b/connect/basic-auth-extension/src/main/java/org/apache/kafka/connect/rest/basic/auth/extension/BasicAuthSecurityRestExtension.java
@@ -18,6 +18,7 @@
 package org.apache.kafka.connect.rest.basic.auth.extension;
 
 import org.apache.kafka.common.utils.AppInfoParser;
+import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.rest.ConnectRestExtension;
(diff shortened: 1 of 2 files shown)


connect/basic-auth-extension/src/main/java/org/apache/kafka/connect/rest/basic/auth/extension/BasicAuthSecurityRestExtension.java
```diff
@@ -80,7 +95,8 @@ public void close() throws IOException {
 
     @Override
     public void configure(Map<String, ?> configs) {
-
+        // If we failed to retrieve a JAAS configuration during startup, throw that exception now
+        CONFIGURATION.get();
```
gharris1727: Could you add a test which confirms that we're propagating the exception here? At the moment, the test verifies that the wrapping method works, but doesn't verify that it's used by the rest extension during the loading phase.

connect/basic-auth-extension/src/test/java/org/apache/kafka/connect/rest/basic/auth/extension/BasicAuthSecurityRestExtensionTest.java
```diff
@@ -63,4 +69,25 @@ public void testJaasConfigurationNotOverwritten() {
...
+            throw jaasConfigurationException;
+        });
+
+        ConnectException thrownException = assertThrows(ConnectException.class, configuration::get);
+        assertEquals(jaasConfigurationException, thrownException.getCause());
+    }
+
+    @Test
+    public void testGoodJaasConfiguration() {
+        Configuration mockConfiguration = EasyMock.mock(Configuration.class);
```
gharris1727, line 90: The identity function could pass this test, but wouldn't have the behavior we need in the BasicAuthSecurityRestExtension. I wonder if there's a way to confirm that the mockConfiguration has been evaluated prior to calling `get()` on the returned supplier.

gharris1727 (commented):
Thanks for the fix @C0urante, it's important that we don't regress users that don't use this plugin. I think most people that don't configure this extension also aren't going to give a valid jaas config :)

connect/basic-auth-extension/src/main/java/org/apache/kafka/connect/rest/basic/auth/extension/BasicAuthSecurityRestExtension.java
C0urante: Ack, done.

connect/basic-auth-extension/src/test/java/org/apache/kafka/connect/rest/basic/auth/extension/BasicAuthSecurityRestExtensionTest.java
C0urante: Not sure if you really meant evaluating `mockConfiguration`, or perhaps just validating that the wrapping `Supplier<Configuration>` was evaluated... I added a test case for the latter; LMKWYT.

C0urante:
Thanks for your comments, @gharris1727. I've expanded on the testing logic to try to give us stronger guarantees about the behavior of the extension; interested in your thoughts on the changes.

gharris1727 (approved):
Lookin good 🚀

C0urante:
Ping @kkonstantine this needs to be merged before any release with the changes from https://github.com/apache/kafka/pull/9806 is put out.

connect/basic-auth-extension/src/test/java/org/apache/kafka/connect/rest/basic/auth/extension/BasicAuthSecurityRestExtensionTest.java
kkonstantine: Another way is to use a mocked supplier and verify that the methods you expect to be called are indeed called (obviously only `get` here) and return or throw as you expect. This test does that, but with an additional level for indirection with this boolean, which basically implements what Easymock does already. Probably fine here since it results in more condensed code. But in the general case you'd pass a mock to verify that methods on that mock are called, which I also assume is referred to as evaluation here.

kkonstantine (approved):
Thanks for the fix @C0urante 
Indeed, static initialization can be tricky. I should have scrutinized the instantiation more on the original PR. 

This LGTM

PR #7353, opened 2019-09-17
KAFKA-8471: Replace control requests/responses with automated protocol

ijuma:
Replaced UpdateMetadata{Request, Response}, LeaderAndIsr{Request, Response}
and StopReplica{Request, Response} with the automated protocol classes.

Updated the JSON schema for the 3 request types to be more consistent and
less strict (if needed to avoid duplication).

The general approach is to avoid generating new collections in the request
classes. Normalization happens in the constructor to make this possible. Builders
still have to group by topic to maintain the external ungrouped view.

Introduced new tests for LeaderAndIsrRequest and UpdateMetadataRequest to
verify that the new logic is correct.

A few other clean-ups/fixes in code that was touched due to these changes:
* KAFKA-8956: Refactor DelayedCreatePartitions#updateWaiting to avoid modifying
collection in foreach.
* Avoid unnecessary allocation for state change trace logging if trace logging is not enabled
* Use `toBuffer` instead of `toList`, `toIndexedSeq` or `toSeq` as it generally performs
better and it matches the performance characteristics of `java.util.ArrayList`. This is
particularly important when passing such instances to Java code.
* Minor refactoring for clarity and readability.
* Removed usage of deprecated `/:`, unused imports and unnecessary `var`s.
* Include exception in `AdminClientIntegrationTest` failure message.
* Move StopReplicaRequest verification in `AuthorizerIntegrationTest` to the end
to match the comment.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #9988, opened 2021-01-28
MINOR: remove useless repeated method call in KafkaApiTest

dengziming:
SetReplicas was invoked twice with same params, I think the author means setIsrs

QA

### DIFF
--- a/core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
+++ b/core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
@@ -2992,7 +2992,7 @@ class KafkaApisTest {
       .setLeaderEpoch(1)
       .setReplicas(replicas)
       .setZkVersion(0)
-      .setReplicas(replicas)
+      .setIsr(replicas)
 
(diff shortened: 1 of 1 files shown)


dengziming:
@ijuma Hello, this is your code, PTAL.

chia7712 (commented):
@dengziming nice find and thanks for your patch.

There are 94 test cases but no one does use ISR info?

ijuma (approved):
LGTM. It may make sense to assert this in some test too.

dengziming:
@chia7712 , I checked the code again and found no one uses ISR info. In fact this code is copied when replacing req/resp with auto-generated protocol in #7353, and the original code set ISR=allReplicas.

chia7712 (approved):

chia7712:
>  the original code set ISR=allReplicas.

Could you share the link to me? thanks!

dengziming:
@chia7712 , see https://github.com/apache/kafka/pull/7353/files#diff-d418e862d65bd68453ec359a3b4f08e38be8948b4d772909ece6ff2a07e5f714L818

PR #9990, opened 2021-01-28
KAFKA-12235: Fix ZkAdminManager.describeConfigs on 2+ config keys

ivanyu:

### DIFF
--- a/core/src/main/scala/kafka/server/ConfigHelper.scala
+++ b/core/src/main/scala/kafka/server/ConfigHelper.scala
@@ -47,11 +47,11 @@ class ConfigHelper(metadataCache: MetadataCache, config: KafkaConfig, configRepo
 
       def createResponseConfig(configs: Map[String, Any],
                                createConfigEntry: (String, Any) => DescribeConfigsResponseData.DescribeConfigsResourceResult): DescribeConfigsResponseData.DescribeConfigsResult = {
-        val filteredConfigPairs = if (resource.configurationKeys == null)
+        val filteredConfigPairs = if (resource.configurationKeys == null || resource.configurationKeys.isEmpty)
           configs.toBuffer
         else
           configs.filter { case (configName, _) =>
-            resource.configurationKeys.asScala.forall(_.contains(configName))
+            resource.configurationKeys.asScala.contains(configName)
           }.toBuffer
 
         val configEntries = filteredConfigPairs.map { case (name, value) => createConfigEntry(name, value) }
--- a/core/src/test/scala/unit/kafka/server/ZkAdminManagerTest.scala
+++ b/core/src/test/scala/unit/kafka/server/ZkAdminManagerTest.scala
@@ -34,6 +34,8 @@ import java.util.Properties
 
 import kafka.server.metadata.ZkConfigRepository
 
+import scala.jdk.CollectionConverters._
+
 class ZkAdminManagerTest {
 
   private val zkClient: KafkaZkClient = EasyMock.createNiceMock(classOf[KafkaZkClient])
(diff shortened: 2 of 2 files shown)


ijuma:
Thanks for the PR. Is this a regression? Also, don't we have any admin client tests that exercise this functionality?

ivanyu:
Yes, looks like a regression. The test ported to 2.6 ([here](https://github.com/apache/kafka/blob/2.6/core/src/test/scala/unit/kafka/server/AdminManagerTest.scala)) works:
```scala
  @Test
  def testDescribeConfigsWithConfigurationKeys(): Unit = {
    EasyMock.expect(zkClient.getEntityConfigs(ConfigType.Topic, topic)).andReturn(TestUtils.createBrokerConfig(brokerId, "zk"))
    EasyMock.expect(metadataCache.contains(topic)).andReturn(true)

    EasyMock.replay(zkClient, metadataCache)

    val resources = Map[ConfigResource, Option[Set[String]]](
      new ConfigResource(ConfigResource.Type.TOPIC, topic) -> Some(Set("retention.ms", "retention.bytes", "segment.bytes"))
    )

    val adminManager = createAdminManager()
    val r = adminManager.describeConfigs(resources, true, false)
    val resultConfigKeys = r.head._2.entries().asScala.map(e => e.name()).toSet
    assertEquals(Set("retention.ms", "retention.bytes", "segment.bytes"), resultConfigKeys)
  }
```
There were some changes of `describeConfigs` signature and code, maybe the bug got in there.

The admin client and command line tools can't specify configuration keys ([it's always `null`](https://github.com/apache/kafka/blob/287b600cf712a4a7e9139e15e1151768f67eec75/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java#L2100)) - I guess that's why the tests are green.

ivanyu:
Resolved conflicts, rebased on the latest `trunk`.
@ijuma could you please have a look?

core/src/main/scala/kafka/server/ConfigHelper.scala
```diff
@@ -47,11 +47,11 @@ class ConfigHelper(metadataCache: MetadataCache, config: KafkaConfig, configRepo
 
       def createResponseConfig(configs: Map[String, Any],
                                createConfigEntry: (String, Any) => DescribeConfigsResponseData.DescribeConfigsResourceResult): DescribeConfigsResponseData.DescribeConfigsResult = {
-        val filteredConfigPairs = if (resource.configurationKeys == null)
+        val filteredConfigPairs = if (resource.configurationKeys == null || resource.configurationKeys.isEmpty)
```
cmccabe, line 50: Hmm... this change doesn't seem correct.  null does not mean the same thing as empty.

This is how `ConfigurationKeys` is described in the schema:
```
      { "name": "ConfigurationKeys", "type": "[]string", "versions": "0+", "nullableVersions": "0+",
        "about": "The configuration keys to list, or null to list all configuration keys." }
```
It doesn't say "null or empty to list all configuration keys."  Just null.  Also, I looked at the code from the 2.7 release and it does not treat empty the same as null.

core/src/main/scala/kafka/server/ConfigHelper.scala
```diff
@@ -47,11 +47,11 @@ class ConfigHelper(metadataCache: MetadataCache, config: KafkaConfig, configRepo
...
           configs.toBuffer
         else
           configs.filter { case (configName, _) =>
-            resource.configurationKeys.asScala.forall(_.contains(configName))
+            resource.configurationKeys.asScala.contains(configName)
```
cmccabe, line 54: I think this bug was introduced when `configurationKeys` went from being an `Option[something]` to being a non-optional type in 2.7.  

The code in 2.6 explains the "cleverness" here-- `forall` always returns true if we're dealing with `None`.  The code in 2.7 drops the comment and the `Option` and introduces the bug.  So we'll need to fix it in that branch too.

cmccabe:
ping @ivanyu

core/src/main/scala/kafka/server/ConfigHelper.scala
ivanyu: I agree with this. However it seems treating empty as null is de facto behavior at the moment (before the fix). To be more precise, here
https://github.com/apache/kafka/blob/e2a0d0c90e1916d77223a420e3595e8aba643001/core/src/main/scala/kafka/server/ConfigHelper.scala#L53-L55
`resource.configurationKeys` being empty means `true` for each element of `configs`, so effectively no filtering. Which, in turn, is equivalent to `resource.configurationKeys == null`.

For example, `testDescribeConfigsWithDocumentation` and `testDescribeConfigsWithEmptyConfigurationKeys` in `ZkAdminManagerTest` break after the fix if treating empty as null is not kept.

I guess it originates from that the default constructor of `DescribeConfigsResource` makes `configurationKeys` an empty array.

We can either 1) keep the current behavior (the current approach); or 2) stick to the spec (also potentially adding `"default": "null"` to the field description). I'd prefer 2. What do you think, @cmccabe?

cmccabe: As you said earlier, the code in 2.7 is broken.  So we should't be going based off of that. The correct code in 2.6 does not special-case the empty string.

cmccabe (approved):
LGTM. We need to get this in ASAP. I'll file a separate PR for the null / empty discussion

PR #9994, opened 2021-01-28
KAFKA-12248: Add BrokerHeartbeat/BrokerRegistration RPCs for KIP-500

aloknnikhil:


hachikuji:
Thanks for the PR. Just to make this a little more manageable, could we split this into two PRs? One for the register/heartbeat controller APIs? And a separate one for decommission, which is a client API. 

Also, it would be nice to have JIRAs since these are significant additions. Thanks!

aloknnikhil:
@hachikuji Fixed it up to merge only the Broker-Controller RPCs (Heartbeat/Registration)

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -98,7 +98,9 @@
     UPDATE_FEATURES(ApiMessageType.UPDATE_FEATURES, false, true),
     ENVELOPE(ApiMessageType.ENVELOPE, true, RecordBatch.MAGIC_VALUE_V0, false, true),
     FETCH_SNAPSHOT(ApiMessageType.FETCH_SNAPSHOT, false, RecordBatch.MAGIC_VALUE_V0, false, true),
-    DESCRIBE_CLUSTER(ApiMessageType.DESCRIBE_CLUSTER);
+    DESCRIBE_CLUSTER(ApiMessageType.DESCRIBE_CLUSTER),
+    BROKER_REGISTRATION(ApiMessageType.BROKER_REGISTRATION, false, RecordBatch.MAGIC_VALUE_V0, false, true),
```
hachikuji: I think the second flag should be true for both of these. They do require CLUSTER_ACTION.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -232,6 +232,9 @@ class KafkaApis(val requestChannel: RequestChannel,
         case ApiKeys.DESCRIBE_QUORUM => requestHelper.closeConnection(request, util.Collections.emptyMap())
         case ApiKeys.FETCH_SNAPSHOT => requestHelper.closeConnection(request, util.Collections.emptyMap())
 
+        // Handle requests that should have been sent to the KIP-500 controller.
```
hachikuji: nit: this applies to all of the Raft apis above as well

hachikuji (commented):
Left a couple comments. You will need to update `RequestConvertToJson` as well. Probably we should have some tests in `RequestResponseTest` as well.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
aloknnikhil: Ack.

core/src/main/scala/kafka/server/KafkaApis.scala
aloknnikhil: Ack.

aloknnikhil:
> Left a couple comments. You will need to update RequestConvertToJson as well. Probably we should have some tests in RequestResponseTest as well.

Done.

hachikuji (approved):
LGTM. Thanks for the patch!

hachikuji:
@aloknnikhil Looks like we will need to fix some failures in `ApiKeysTest` before we can merge this. I suspect that the test does not like the throttle time fields. One way we could handle this is to ignore `controllerOnly` APIs in this test case. I guess we haven't really figured out the throttling picture for broker-to-controller APIs.

PR #9995, opened 2021-01-29
Add log about broker info when getting UNKNOWN_TOPIC_OR_PARTITION error

sudoa:
Description: This patch adds additional logs in the `NetworkClient` to print metadata info such as leader id and epoch for partitions that fail metadata fetch with UNKNOWN_TOPIC_OR_PARTITION error. This will help with a Venice investigation: https://jira01.corp.linkedin.com:8443/browse/LIKAFKA-33540
Testing: N/A, since it's for logs.


sudoa:
done by mistake

PR #9996, opened 2021-01-29
KAFKA-12249: KIP-500 - Add client-side Decommission Broker RPC

aloknnikhil:


clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
```diff
@@ -1449,6 +1449,39 @@ default DescribeFeaturesResult describeFeatures() {
      */
     UpdateFeaturesResult updateFeatures(Map<String, FeatureUpdate> featureUpdates, UpdateFeaturesOptions options);
 
+    /**
+     * Unregisters a broker.
```
hachikuji: We should document the fact that this will only work for self-managed kafka clusters.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4503,6 +4506,47 @@ void handleFailure(Throwable throwable) {
         return new UpdateFeaturesResult(new HashMap<>(updateFutures));
     }
 
+    @Override
+    public DecommissionBrokerResult decommissionBroker(int brokerId, DecommissionBrokerOptions options) {
+        final KafkaFutureImpl<Void> future = new KafkaFutureImpl<>();
+        final long now = time.milliseconds();
+        final Call call = new Call("decommissionBroker", calcDeadlineMs(now, options.timeoutMs()),
+            new ControllerNodeProvider()) {
```
hachikuji: Shouldn't we use `LeastLoadedNodeProvider`? The kip-500 controller will not be reachable anyway.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -98,7 +98,8 @@
     UPDATE_FEATURES(ApiMessageType.UPDATE_FEATURES, false, true),
     ENVELOPE(ApiMessageType.ENVELOPE, true, RecordBatch.MAGIC_VALUE_V0, false, true),
     FETCH_SNAPSHOT(ApiMessageType.FETCH_SNAPSHOT, false, RecordBatch.MAGIC_VALUE_V0, false, true),
-    DESCRIBE_CLUSTER(ApiMessageType.DESCRIBE_CLUSTER);
+    DESCRIBE_CLUSTER(ApiMessageType.DESCRIBE_CLUSTER),
+    DECOMMISSION_BROKER(ApiMessageType.DECOMMISSION_BROKER, false, RecordBatch.MAGIC_VALUE_V0, true, true);
```
hachikuji, line 108: This is not a controller-only API.

clients/src/main/resources/common/message/DecommissionBrokerResponse.json
```diff
@@ -0,0 +1,28 @@
...
+{
+  "apiKey": 63,
+  "type": "response",
+  "name": "DecommissionBrokerResponse",
+  "validVersions": "0",
+  "flexibleVersions": "0+",
+  "fields": [
+    { "name": "ThrottleTimeMs", "type": "int32", "versions": "0+",
+      "about": "Duration in milliseconds for which the request was throttled due to a quota violation, or zero if the request did not violate any quota." },
+    { "name": "ErrorCode", "type": "int16", "versions": "0+",
```
hachikuji, line 25: Considering if we may as well add an ErrorMessage field here.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4503,6 +4506,47 @@ void handleFailure(Throwable throwable) {
         return new UpdateFeaturesResult(new HashMap<>(updateFutures));
     }
 
+    @Override
+    public DecommissionBrokerResult decommissionBroker(int brokerId, DecommissionBrokerOptions options) {
```
hachikuji, line 4612: Can you add some unit tests for this api? Nice to cover at least the success/error cases.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3296,6 +3298,13 @@ class KafkaApis(val requestChannel: RequestChannel,
     handle(forwardedRequest)
   }
 
+  def handleDecommissionBrokerRequest(request: RequestChannel.Request): Unit = {
+    val decommissionBrokerRequest = request.body[DecommissionBrokerRequest]
+    requestHelper.sendResponseMaybeThrottle(request, requestThrottleMs =>
+      decommissionBrokerRequest.getErrorResponse(requestThrottleMs,
+        Errors.UNSUPPORTED_VERSION.exception))
```
hachikuji: If we advertise this API in the ApiVersions response, then it would be surprising to get an UNSUPPORTED_VERSION error. I think we might need augment the ApiVersion handler here so that we are able to exclude KIP-500 APIs when it is not enabled.

hachikuji:
Note also the test failure `RequestConvertToJsonTest.testAllResponseTypesHandled`.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
aloknnikhil: Hmm. Isn't decommissioning a broker something only a controller can do? Or maybe I am missing something here.

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
aloknnikhil: Ack.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
aloknnikhil: Ah, didn't know a `LeastLoadedNodeProvider` existed. Fixed!

clients/src/main/resources/common/message/DecommissionBrokerResponse.json
aloknnikhil: Sure. That would be consistent with the other AdminClient APIs

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
aloknnikhil: Ack.

core/src/main/scala/kafka/server/KafkaApis.scala
aloknnikhil: That makes sense.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
hachikuji: You are right that only the controller can properly handle it, but the broker needs to be able to forward it, so it must also expose it.

aloknnikhil: Ah. Right. I'm curious, then, what would happen if we send a decommission request through the broker being decommissioned? Wouldn't the broker shutdown (as a result of being unregistered) before responding to the AdminClient call?

hachikuji: I am not sure. It seems like the decommissioned broker should wait around long enough to give the controller a chance to move replicas elsewhere.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -123,6 +124,9 @@
     /** indicates whether the API is enabled for forwarding **/
     public final boolean forwardable;
 
+    /** indicates if this API is exclusive to the KIP-500 mode **/
+    public final boolean isKip500OnlyApi;
```
aloknnikhil: @hachikuji I am not entirely sure we need this additional flag but I don't think there's a cleaner way to mark certain APIs as being available only under KIP-500

aloknnikhil: Spoke offline. Will mark the API as `controllerOnly` for now to not advertise as part of the supported BrokerAPIs. We will allow the broker to process requests for this API once the KIP-500 controller integration is complete

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
```diff
@@ -1485,6 +1485,43 @@ default DescribeFeaturesResult describeFeatures() {
...
+     * <p>
+     * This operation is supported only on self-managed Kafka clusters.
```
hachikuji: Let's emphasize that this means brokers which do not use Zookeeper.

> This operation is supported only on self-managed Kafka clusters (i.e. brokers which do not rely on Zookeeper).

Same thing for the other API.

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
```diff
@@ -1485,6 +1485,43 @@ default DescribeFeaturesResult describeFeatures() {
      */
     UpdateFeaturesResult updateFeatures(Map<String, FeatureUpdate> featureUpdates, UpdateFeaturesOptions options);
 
+    /**
+     * Unregisters a broker.
```
hachikuji: What does this mean exactly? The verb "decommission" sounds like we are permanently removing the broker and reassigning any partitions that were assigned to it.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4605,6 +4608,47 @@ void handleFailure(Throwable throwable) {
...
+            void handleResponse(AbstractResponse abstractResponse) {
+                final DecommissionBrokerResponse response =
+                    (DecommissionBrokerResponse) abstractResponse;
+                Errors error = Errors.forCode(response.data().errorCode());
+                switch (error) {
+                    case NONE:
+                        future.complete(null);
+                        break;
+                    case NOT_CONTROLLER:
+                        handleNotControllerError(error);
```
hachikuji: Do we actually expect to get back `NOT_CONTROLLER`? I would have thought that this error would be swallowed by the broker when it forwards the request. In the current `ForwardingManager`, the broker will try to send the request to the controller. If the controller returns `NOT_CONTROLLER`, it will just retry. It will continue doing so until it ultimately times out. So maybe `REQUEST_TIMED_OUT` is the one we should be looking for?

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -101,7 +101,10 @@
     DESCRIBE_CLUSTER(ApiMessageType.DESCRIBE_CLUSTER),
     DESCRIBE_PRODUCERS(ApiMessageType.DESCRIBE_PRODUCERS),
     BROKER_REGISTRATION(ApiMessageType.BROKER_REGISTRATION, true, RecordBatch.MAGIC_VALUE_V0, false, true),
-    BROKER_HEARTBEAT(ApiMessageType.BROKER_HEARTBEAT, true, RecordBatch.MAGIC_VALUE_V0, false, true);
+    BROKER_HEARTBEAT(ApiMessageType.BROKER_HEARTBEAT, true, RecordBatch.MAGIC_VALUE_V0, false, true),
+
+    // Once we have the controller integration for supporting broker decommissioning, we will support forwarding from the broker
```
hachikuji, line 106: Maybe emphasize that we have temporarily marked this as "controller-only" as a short-term workaround to avoid exposing it on zk-based brokers.

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
```diff
@@ -5169,6 +5171,72 @@ public void testShouldRefreshCoordinator() {
...
+        }
+    }
+
+    @Test
+    public void testDecommissionBrokerFailure() {
+        int decommissionedBrokerNode = 1;
+        try (AdminClientUnitTestEnv env = mockClientEnv()) {
+            env.kafkaClient().setNodeApiVersions(
+                    NodeApiVersions.create(ApiKeys.DECOMMISSION_BROKER.id, (short) 0, (short) 0));
+            env.kafkaClient().prepareResponse(prepareDecommissionBrokerResponse(Errors.BROKER_NOT_AVAILABLE, 0));
```
hachikuji: Hmm.. I would actually expect BROKER_NOT_AVAILABLE to be retriable. Maybe we should use an error code like `UNKNOWN_SERVER_ERROR`.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -235,6 +235,11 @@ class KafkaApis(val requestChannel: RequestChannel,
         case ApiKeys.FETCH_SNAPSHOT => requestHelper.closeConnection(request, util.Collections.emptyMap())
         case ApiKeys.BROKER_REGISTRATION => requestHelper.closeConnection(request, util.Collections.emptyMap())
         case ApiKeys.BROKER_HEARTBEAT => requestHelper.closeConnection(request, util.Collections.emptyMap())
+
+        // We need to fix this to forward the request to the active controller.
```
hachikuji: We can move this comment to a jira.

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
aloknnikhil: Done.

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
aloknnikhil: Right. I guess at some point the comment and the name of the API seem to have diverged. Updated the comment to reflect the use of the API

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
aloknnikhil: Hmm. Good point. We don't need a special-case for the REQUEST_TIMED_OUT then. Will update the test and case here

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
aloknnikhil: Ack.

clients/src/test/java/org/apache/kafka/clients/admin/KafkaAdminClientTest.java
aloknnikhil: Makes sense. Updated

core/src/main/scala/kafka/server/KafkaApis.scala
aloknnikhil: https://issues.apache.org/jira/browse/KAFKA-12275

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4605,6 +4609,56 @@ void handleFailure(Throwable throwable) {
...
+            void handleResponse(AbstractResponse abstractResponse) {
+                final DecommissionBrokerResponse response =
+                        (DecommissionBrokerResponse) abstractResponse;
+                Errors error = Errors.forCode(response.data().errorCode());
+                switch (error) {
+                    case NONE:
+                        context.future().complete(null);
+                        break;
+                    case REQUEST_TIMED_OUT:
+                        Call nextCall = getDecommissionBrokerCall(context);
```
hachikuji: I think you can just raise the exception? It will get caught and `Call.fail` will be called. This does all the `tries` bookkeeping and will resubmit it if the exception is retriable (which `TimeoutException` is).

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -4605,6 +4609,56 @@ void handleFailure(Throwable throwable) {
...
+                        nextCall.tries = super.tries + 1;
+                        nextCall.nextAllowedTryMs = calculateNextAllowedRetryMs();
+                        runnable.call(nextCall, time.milliseconds());
+                        break;
+                    default:
+                        context.future().completeExceptionally(error.exception());
```
hachikuji: Can we log an error message here which includes the request information and the unexpected error?

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
aloknnikhil: Ah. That's much easier :)

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
aloknnikhil: Ack.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3317,6 +3318,13 @@ class KafkaApis(val requestChannel: RequestChannel,
     handle(forwardedRequest)
   }
 
+  def handleDecommissionBrokerRequest(request: RequestChannel.Request): Unit = {
```
hachikuji: Can we remove this for now since we are not using it?

hachikuji (approved):
LGTM overall. Just one more thing to fix.

ijuma:
Restarted the build, it timed out for some reason.

ijuma:
Two flaky tests:

> JDK 8 / org.apache.kafka.common.metrics.MetricsTest.testPercentilesWithRandomNumbersAndLinearBucketing()
> JDK 8 / org.apache.kafka.connect.integration.BlockingConnectorTest.testBlockInConnectorValidate

Aatirhassanpir:
Title
[FEATURE] Add client-side Decommission Broker RPC for KIP-500

Repository Version
Apache Kafka 2.x.x

Key Words / Search Terms
Decommission Broker, RPC, KIP-500, Kafka Admin Client, Broker Removal, Self-managed Kafka

Description
This feature introduces the client-side implementation of the Decommission Broker RPC as part of KIP-500. This API will allow clients to request the removal of brokers from a Kafka cluster, ensuring smooth broker lifecycle management in self-managed (ZooKeeper-less) Kafka environments. It includes methods for decommissioning brokers and the necessary result classes to track the operation.

Expected Behavior
The client should be able to call the decommissionBroker method to decommission brokers by their ID.
The operation should be supported only for self-managed Kafka clusters.
The method should return a result object (DecommissionBrokerResult) that can track the status of the operation via KafkaFuture<Void>.
The decommissioning process should trigger partition reassignment for the removed broker.
Current Behavior
Kafka currently does not provide an API for clients to directly decommission brokers.
The broker decommissioning process requires interaction with the cluster, and there’s no clear way to manage it from the client side.
Steps to reproduce (for bug reports)
Call the existing Admin client API to decommission a broker.
Observe that no such functionality is available, leading to an error or no result.
Use cases / motivating examples (for feature requests)
Cluster Scaling: When scaling down a Kafka cluster, it’s essential to decommission brokers gracefully without affecting data availability or cluster health.
Self-managed Clusters: This feature is crucial for Kafka clusters that do not rely on ZooKeeper and are managed using KIP-500.
Automated Broker Management: By providing this API, users can automate broker decommissioning processes within their operations, making cluster management more efficient.

PR #9997, opened 2021-01-29
KAFKA-9274: Add timeout handling for `StreamPartitioner`

mjsax:
- part of KIP-572

When a custom `StreamPartitioner` is used, we need to get the number of partitions of output topics from the producer.
This `partitionFor(topic)` call may through a `TimeoutException` that we now handle gracefully.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
@@ -113,7 +113,7 @@ public class RecordCollectorImpl implements RecordCollector {
             try {
                 partitions = streamsProducer.partitionsFor(topic);
             } catch (final TimeoutException timeoutException) {
-                log.debug("Could not get partitions for topic {}, will retry", topic);
+                log.warn("Could not get partitions for topic {}, will retry", topic);
 
                 // re-throw to trigger `task.timeout.ms`
                 throw timeoutException;
@@ -121,7 +121,9 @@ public class RecordCollectorImpl implements RecordCollector {
                 // here we cannot drop the message on the floor even if it is a transient timeout exception,
                 // so we treat everything the same as a fatal exception
                 throw new StreamsException("Could not determine the number of partitions for topic '" + topic +
-                    "' for task " + taskId + " due to " + fatal.toString());
+                    "' for task " + taskId + " due to " + fatal.toString(),
+                    fatal
+                );
             }
             if (partitions.size() > 0) {
                 partition = partitioner.partition(topic, key, value, partitions.size());
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
@@ -31,6 +31,7 @@ import org.apache.kafka.common.errors.ProducerFencedException;
 import org.apache.kafka.common.errors.RetriableException;
 import org.apache.kafka.common.errors.SecurityDisabledException;
 import org.apache.kafka.common.errors.SerializationException;
+import org.apache.kafka.common.errors.TimeoutException;
 import org.apache.kafka.common.errors.UnknownServerException;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.common.metrics.Sensor;
@@ -111,13 +112,16 @@ public class RecordCollectorImpl implements RecordCollector {
             final List<PartitionInfo> partitions;
             try {
                 partitions = streamsProducer.partitionsFor(topic);
-            } catch (final KafkaException e) {
-                // TODO: KIP-572 need to handle `TimeoutException`
-                // -> should we throw a `TaskCorruptedException` for this case to reset the task and retry (including triggering `task.timeout.ms`) ?
+            } catch (final TimeoutException timeoutException) {
+                log.debug("Could not get partitions for topic {}, will retry", topic);
+
+                // re-throw to trigger `task.timeout.ms`
+                throw timeoutException;
+            } catch (final KafkaException fatal) {
                 // here we cannot drop the message on the floor even if it is a transient timeout exception,
                 // so we treat everything the same as a fatal exception
                 throw new StreamsException("Could not determine the number of partitions for topic '" + topic +
-                    "' for task " + taskId + " due to " + e.toString());
+                    "' for task " + taskId + " due to " + fatal.toString());
             }
             if (partitions.size() > 0) {
                 partition = partitioner.partition(topic, key, value, partitions.size());
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
@@ -31,6 +31,7 @@ import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.errors.DeserializationExceptionHandler;
 import org.apache.kafka.streams.errors.LockException;
 import org.apache.kafka.streams.errors.StreamsException;
+import org.apache.kafka.streams.errors.TaskCorruptedException;
 import org.apache.kafka.streams.errors.TaskMigratedException;
 import org.apache.kafka.streams.errors.TopologyException;
 import org.apache.kafka.streams.processor.Cancellable;
@@ -99,6 +100,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
     private final InternalProcessorContext processorContext;
     private final RecordQueueCreator recordQueueCreator;
 
+    private StampedRecord record;
     private boolean commitNeeded = false;
     private boolean commitRequested = false;
     private boolean hasPendingTxCommit = false;
@@ -621,6 +623,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
                 throw new IllegalStateException("Unknown state " + state() + " while closing active task " + id);
         }
 
+        record = null;
         partitionGroup.clear();
         closeTaskSensor.record();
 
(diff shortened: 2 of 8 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -99,6 +100,7 @@
     private final InternalProcessorContext processorContext;
     private final RecordQueueCreator recordQueueCreator;
 
+    private StampedRecord record;
```
mjsax, line 103: We pull this variable out form the method to use it as a "cache" -- if we fail on send(), we can process the record a second time.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -663,18 +666,21 @@ public boolean isProcessable(final long wallClockTime) {
      */
     @SuppressWarnings("unchecked")
     public boolean process(final long wallClockTime) {
-        if (!isProcessable(wallClockTime)) {
-            return false;
-        }
+        if (record == null) {
```
mjsax, line 669: If we have a cached record, it implies we failed to process it before -- thus we don't pull a new record from the buffer but retry to process the cached record.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -714,19 +720,34 @@ public boolean process(final long wallClockTime) {
...
-        } catch (final StreamsException e) {
-            throw e;
+
+            record = null;
+        } catch (final TimeoutException timeoutException) {
+            if (!eosEnabled) {
+                throw timeoutException;
+            } else {
+                record = null;
+                throw new TaskCorruptedException(Collections.singletonMap(id, changelogPartitions()));
```
mjsax, line 730: For this case, we don't need the cached record, as we need to reset the task anyway to cleanup potentially "corrupted" state store.

mjsax: Note that we don't trigger `task.timeout.ms` for this case atm. Because we need to restore state what might talk some time, it seems questionable if we should tigger `task.timeout.ms` for this case of not.

Cf. TaskManager#process() that catches `TimeoutException` and trigger the timeout, but does not catch `TaskCorruptedException`

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
```diff
@@ -111,13 +112,16 @@ public void initialize() {
             final List<PartitionInfo> partitions;
             try {
                 partitions = streamsProducer.partitionsFor(topic);
-            } catch (final KafkaException e) {
-                // TODO: KIP-572 need to handle `TimeoutException`
-                // -> should we throw a `TaskCorruptedException` for this case to reset the task and retry (including triggering `task.timeout.ms`) ?
+            } catch (final TimeoutException timeoutException) {
+                log.debug("Could not get partitions for topic {}, will retry", topic);
```
ableegoldman: seems like it should at least be info, if not warn

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
ableegoldman: I agree, we shouldn't trigger `task.timeout.ms` until it's back to RUNNING right?

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
```diff
@@ -1109,22 +1109,32 @@ int process(final int maxNumRecords, final Time time) {
...
-                int processed = 0;
-                final long then = now;
                 while (processed < maxNumRecords && task.process(now)) {
+                    task.clearTaskTimeout();
                     processed++;
                 }
-                now = time.milliseconds();
-                totalProcessed += processed;
-                task.recordProcessBatchTime(now - then);
+            } catch (final TimeoutException timeoutException) {
```
ableegoldman, line 1119: When would we hit this TimeoutException vs the one about (in Task)?

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
vvcephei: +1

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
```diff
@@ -111,13 +112,16 @@ public void initialize() {
...
+
+                // re-throw to trigger `task.timeout.ms`
+                throw timeoutException;
+            } catch (final KafkaException fatal) {
                 // here we cannot drop the message on the floor even if it is a transient timeout exception,
                 // so we treat everything the same as a fatal exception
                 throw new StreamsException("Could not determine the number of partitions for topic '" + topic +
-                    "' for task " + taskId + " due to " + e.toString());
+                    "' for task " + taskId + " due to " + fatal.toString());
```
vvcephei: Can we also include the cause when we throw exceptions? It's not always helpful, but it has been invaluable for debugging many times since we started to include the cause.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -714,19 +720,34 @@ public boolean process(final long wallClockTime) {
             if (recordInfo.queue().size() == maxBufferedSize) {
                 mainConsumer.resume(singleton(partition));
             }
-        } catch (final StreamsException e) {
-            throw e;
+
+            record = null;
+        } catch (final TimeoutException timeoutException) {
+            if (!eosEnabled) {
+                throw timeoutException;
```
vvcephei, line 727: It seems like KIP-572 should retry only when it's safe to do so. Under ALOS, it's not safe to retry, so we should just crash here, right?

vvcephei: If I'm reading the code right, this would also mean we don't need to mess with caching the StampedRecord, which would be nice.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
vvcephei: It looks like it's only here to handle the case of ALOS. On the contrary, if we just crash in that case, we don't need this block here at all, right?

vvcephei (commented):
Thanks, @mjsax !

I really just have one comment, that it seems safer and simpler to just crash (as we do today) when we get a TimeoutException under ALOS.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
ableegoldman: @vvcephei what do you mean by "crash" here? Throw TaskCorrutpedException?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
mjsax: Not sure if starting the timeout at all makes sense for this case?

Also, starting it when we transit to RUNNING seems to be a non-trivial change.

streams/src/main/java/org/apache/kafka/streams/processor/internals/RecordCollectorImpl.java
mjsax: INFO does not make sense I guess. So WARN it it.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
mjsax: Why would it not be safe to retry the commit for at-least-once? -- The indention of the PR was to retry for this case. -- It may lead to duplication, but that seems fine with at-least-once.

streams/src/main/java/org/apache/kafka/streams/processor/internals/TaskManager.java
mjsax: > vs the one about (in Task)?

Not sure what you mean. Can you elaborate?

> It looks like it's only here to handle the case of ALOS. On the contrary, if we just crash in that case, we don't need this block here at all, right?

Yes, it's only to at-least-once -- for exactly-once we need to roll-back the state stores -- we cannot just retry. -- And the goal is to _avoid_ to just crash... :D

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
vvcephei: Thanks, all, sorry for the delay. I meant "crash" as in, crash the thread. This is what happens today, right? We throw the TimeoutException, convert it to a StreamsException, and throw it up to the top level, killing the thread?

What I am thinking is that it's silly to just crash a thread when it has gotten a timeout in `poll()` or in checking `partitions()` because we can harmlessly retry those operations. Along those lines, retrying `commit` should be safe as well.

But it looks like this PR is not retrying just `commit`, but is retrying the entire processing of a record. Is that a wrong conclusion?

Retrying "process a record" is definitely not "safe" or "harmless" it will absolutely result in producing incorrect results, therefore we should not do it.

mjsax: > But it looks like this PR is not retrying just commit, but is retrying the entire processing of a record. Is that a wrong conclusion?
>
> Retrying "process a record" is definitely not "safe" or "harmless" it will absolutely result in producing incorrect results, therefore we should not do it.

Well, overall I agree, but if we crash the thread with at-least-once, a we rebalance, the other thread will reprocess based on the last committed offset, so the net result is, that a rebalance might reprocess even _more_ records -- if we handle the timeout, we only try to reprocess a single record.

ableegoldman (approved):
LGTM. I share some of John's reservations, but I am fine with proceeding with this PR and we can always revisit the issue later

PR #9998, opened 2021-01-29
KAFKA-12250; Add metadata record serde for KIP-631

hachikuji:
This patch adds a `RecordSerde` implementation for the metadata record format expected by KIP-631. 

### DIFF
--- a/metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
+++ b/metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
@@ -27,23 +27,15 @@ import org.apache.kafka.raft.RecordSerde;
 public class MetadataRecordSerde implements RecordSerde<ApiMessageAndVersion> {
 
     @Override
-    public ObjectSerializationCache newWriteContext() {
-        return new ObjectSerializationCache();
-    }
-
-    @Override
-    public int recordSize(ApiMessageAndVersion data, Object context) {
-        ObjectSerializationCache serializationCache = (ObjectSerializationCache) context;
-        int size = 0;
-        size += ByteUtils.sizeOfUnsignedVarint(data.message().apiKey());
+    public int recordSize(ApiMessageAndVersion data, ObjectSerializationCache serializationCache) {
+        int size = ByteUtils.sizeOfUnsignedVarint(data.message().apiKey());
         size += ByteUtils.sizeOfUnsignedVarint(data.version());
         size += data.message().size(serializationCache, data.version());
         return size;
     }
 
     @Override
-    public void write(ApiMessageAndVersion data, Object context, Writable out) {
-        ObjectSerializationCache serializationCache = (ObjectSerializationCache) context;
+    public void write(ApiMessageAndVersion data, ObjectSerializationCache serializationCache, Writable out) {
         out.writeUnsignedVarint(data.message().apiKey());
         out.writeUnsignedVarint(data.version());
         data.message().write(out, serializationCache, data.version());
--- a/metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
+++ b/metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.metadata;
+
+import org.apache.kafka.common.metadata.MetadataRecordType;
+import org.apache.kafka.common.protocol.ApiMessage;
+import org.apache.kafka.common.protocol.ObjectSerializationCache;
+import org.apache.kafka.common.protocol.Readable;
+import org.apache.kafka.common.protocol.Writable;
+import org.apache.kafka.common.utils.ByteUtils;
+import org.apache.kafka.raft.RecordSerde;
+
+public class MetadataRecordSerde implements RecordSerde<ApiMessageAndVersion> {
+
+    @Override
+    public ObjectSerializationCache newWriteContext() {
+        return new ObjectSerializationCache();
+    }
+
+    @Override
+    public int recordSize(ApiMessageAndVersion data, Object context) {
+        ObjectSerializationCache serializationCache = (ObjectSerializationCache) context;
+        int size = 0;
+        size += ByteUtils.sizeOfUnsignedVarint(data.message().apiKey());
+        size += ByteUtils.sizeOfUnsignedVarint(data.version());
+        size += data.message().size(serializationCache, data.version());
+        return size;
+    }
+
+    @Override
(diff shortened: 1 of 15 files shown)


ijuma:
Is this ready for review @hachikuji?

metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public ObjectSerializationCache newWriteContext() {
+        return new ObjectSerializationCache();
+    }
+
+    @Override
+    public int recordSize(ApiMessageAndVersion data, Object context) {
+        ObjectSerializationCache serializationCache = (ObjectSerializationCache) context;
+        int size = 0;
```
ijuma: Nit: seems a bit redundant. Can we not assign the size in the line below?

metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
```diff
@@ -0,0 +1,62 @@
...
+public class MetadataRecordSerde implements RecordSerde<ApiMessageAndVersion> {
+
+    @Override
+    public ObjectSerializationCache newWriteContext() {
+        return new ObjectSerializationCache();
+    }
+
+    @Override
+    public int recordSize(ApiMessageAndVersion data, Object context) {
+        ObjectSerializationCache serializationCache = (ObjectSerializationCache) context;
```
ijuma: Why do we cast instead of having the context type as a type parameter?

metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
```diff
@@ -0,0 +1,62 @@
...
+        out.writeUnsignedVarint(data.message().apiKey());
+        out.writeUnsignedVarint(data.version());
+        data.message().write(out, serializationCache, data.version());
+    }
+
+    @Override
+    public ApiMessageAndVersion read(Readable input, int size) {
+        short apiKey = (short) input.readUnsignedVarint();
+        short version = (short) input.readUnsignedVarint();
```
ijuma: I know this is in the KIP but is it really worth deviating from the request format (two shorts) to save two bytes for low api key/version? We end up with additional complexity on writes/reads (see size code above for example).

cmccabe: It is a big savings overall, right?  2 bytes per record, for every record.  We use varints in the record format for the same reasons.

Actually the question that @hachikuji raised earlier was whether we should have a third varint which comes first for message frame version, which is not unreasonable.

ijuma: @cmccabe That really depends on the typical record size. And hence why I asked the question.

metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
hachikuji: The types here leak into `RaftClient` and `RaftManager`, which is really annoying. Having `RaftClient<ApiMessageAndVersion>` seems reasonable, but `RaftClient<ApiMessageAndVersion, ObjectSerializationCache>` is cumbersome. An alternative is to change the type to always be `ObjectSerializationCache` since there isn't actually any other use case.

metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
hachikuji: I don't feel too strongly about the use of the unsigned type. I have seen the varint size functions pop up on some flame graphs, so the concern is valid. If it becomes a real issue, one idea would be to precompute the size during the code generation. Really we could even hard-code the value 1 here and then write a test which fails if there are any api keys/versions which do not fit in a single byte.

@cmccabe Yeah, I still think it's probably worth having a version byte or something in case we need to change this in the future. We have regretted not having a header version in the RPC schema.

metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
hachikuji: Fixed. I did it that way because the asymmetry in the following calls kind of annoyed me 😆 .

metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
hachikuji: I pushed a commit to do this. Let me know if that addresses the concern.

metadata/src/main/java/org/apache/kafka/metadata/MetadataRecordSerde.java
ijuma: OK, so with the frame version, we're now talking about 6 bytes. I'm happy to go with this and we can re-evaluate before 3.0 once we have clear data on storage savings vs cpu impact. It may be worth filing a JIRA so that we don't forget.

hachikuji: I filed this JIRA: https://issues.apache.org/jira/browse/KAFKA-12286.

raft/src/main/java/org/apache/kafka/raft/metadata/MetadataRecordSerde.java
```diff
@@ -0,0 +1,66 @@
...
+    public void write(ApiMessageAndVersion data, ObjectSerializationCache serializationCache, Writable out) {
+        out.writeUnsignedVarint(DEFAULT_FRAME_VERSION);
+        out.writeUnsignedVarint(data.message().apiKey());
+        out.writeUnsignedVarint(data.version());
+        data.message().write(out, serializationCache, data.version());
+    }
+
+    @Override
+    public ApiMessageAndVersion read(Readable input, int size) {
+        short frameVersion = (short) input.readUnsignedVarint();
```
ijuma, line 52: Do we need to verify that the cast is safe?

raft/src/test/java/org/apache/kafka/raft/metadata/MetadataRecordSerdeTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+            ByteBuffer buffer = ByteBuffer.allocate(size);
+            ByteBufferAccessor bufferAccessor = new ByteBufferAccessor(buffer);
+
+            serde.write(messageAndVersion, cache, bufferAccessor);
+            buffer.flip();
+
+            assertEquals(size, buffer.remaining());
+            ApiMessageAndVersion readMessageAndVersion = serde.read(bufferAccessor, size);
+            Assertions.assertEquals(messageAndVersion, readMessageAndVersion);
```
ijuma: Nit: remove `Assertions.` prefix.

ijuma (approved):
LGTM. One nit and one potential improvement we can tackle via a separate PR/JIRA.

cmccabe:
LGTM

raft/src/main/java/org/apache/kafka/raft/metadata/MetadataRecordSerde.java
hachikuji: Yeah, we should have a helper here. I will create a follow-up.

raft/src/test/java/org/apache/kafka/raft/metadata/MetadataRecordSerdeTest.java
```diff
@@ -0,0 +1,72 @@
...
+        ByteBuffer buffer = ByteBuffer.allocate(16);
+        ByteUtils.writeUnsignedVarint(15, buffer);
+        buffer.flip();
+
+        MetadataRecordSerde serde = new MetadataRecordSerde();
+        assertThrows(SerializationException.class,
+            () -> serde.read(new ByteBufferAccessor(buffer), 16));
+    }
+
+}
```
jsancio, line 72: Missing newline character.

hachikuji: Nod. Will fix in the follow-up mentioned above. Did not want to trigger another build.

PR #10000, opened 2021-01-29
KAFKA-9274: handle TimeoutException on task reset

mjsax:
- part of KIP-572

This changes move the offset reset for the internal "main consumer" when we revive a corrupted task, from the "task cleanup" code path, to the "task init" code path. For this case, we have already logic in place to handle `TimeoutException` that might be thrown by `consumer#committed()` method call.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
@@ -214,7 +214,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
      * @throws StreamsException fatal error, should close the thread
      */
     @Override
-    public void initializeIfNeeded(java.util.function.Consumer<Set<TopicPartition>> offsetResetter) {
+    public void initializeIfNeeded(final java.util.function.Consumer<Set<TopicPartition>> offsetResetter) {
         if (state() == State.CREATED) {
             initOffsetsIfNeeded(offsetResetter);
             recordCollector.initialize();
@@ -232,7 +232,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
         }
     }
 
-    private void initOffsetsIfNeeded(java.util.function.Consumer<Set<TopicPartition>> offsetResetter) {
+    private void initOffsetsIfNeeded(final java.util.function.Consumer<Set<TopicPartition>> offsetResetter) {
         final Map<TopicPartition, OffsetAndMetadata> committed = mainConsumer.committed(resetOffsetsForPartitions);
         for (final Map.Entry<TopicPartition, OffsetAndMetadata> committedEntry : committed.entrySet()) {
             final OffsetAndMetadata offsetAndMetadata = committedEntry.getValue();
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
@@ -85,6 +85,7 @@ public class StreamTask extends AbstractTask implements ProcessorNodePunctuator,
     private final RecordCollector recordCollector;
     private final PartitionGroup.RecordInfo recordInfo;
     private final Map<TopicPartition, Long> consumedOffsets;
+    private final Set<TopicPartition> resetOffsetsForPartitions;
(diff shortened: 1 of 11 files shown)


chia7712:
congratulations on PR 10,000 :)

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -227,6 +230,27 @@ public void initializeIfNeeded() {
         }
     }
 
+    private void initOffsetsIfNeeded(final java.util.function.Consumer<Set<TopicPartition>> offsetResetter) {
```
mjsax: I was considering to maybe merge this method into `initMetadata()` but it might convolute different code path, and we should execute this method rarely anyway so I don't think we should have concerns about calling `mainConsumer.committed` twice for rare cases.

Let me know what you think.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -227,6 +230,27 @@ public void initializeIfNeeded() {
...
+        final Map<TopicPartition, OffsetAndMetadata> committed = mainConsumer.committed(resetOffsetsForPartitions);
+        for (final Map.Entry<TopicPartition, OffsetAndMetadata> committedEntry : committed.entrySet()) {
+            final OffsetAndMetadata offsetAndMetadata = committedEntry.getValue();
+            if (offsetAndMetadata != null) {
+                mainConsumer.seek(committedEntry.getKey(), offsetAndMetadata);
+                resetOffsetsForPartitions.remove(committedEntry.getKey());
+            }
+        }
+
+        if (!resetOffsetsForPartitions.isEmpty()) {
```
mjsax: this `if` is not strictly required, however, it allows us to just pass `null` as `offsetResetter` in tests, so might be worth it.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
ableegoldman: Hm...I'm not necessarily that concerned about calling `mainConsumer.committed` twice in rare cases (although maybe that would not be so good, since those rare cases happen to be those in which this is probably more likely to time out, right?)
But personally, just coming into this code from the outside, it's super confusing to have two different methods for initializing the offsets. It seems more convoluted that way, to me. Also maybe I am missing some context here but why do we call `initOffsetsIfNeeded`  from `initializeIfNeeded` rather than from `completeRestoration` in the first place? We don't need to initialize main consumer offsets until it transitions to running

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
ableegoldman: Can we just pass in a no-op lambda instead? I'd rather avoid special handling for null input that isn't supposed to be null, just so we can use null in the tests (which are therefore not realistic tests since it should never be null, no?)

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
mjsax: Good point!

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
mjsax: Fair enough.

mjsax:
Updated this PR.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamTask.java
```diff
@@ -229,17 +231,22 @@ public void initializeIfNeeded() {
...
     @Override
-    public void completeRestoration() {
+    public void completeRestoration(final java.util.function.Consumer<Set<TopicPartition>> offsetResetter) {
         switch (state()) {
             case RUNNING:
                 return;
 
             case RESTORING:
-                initializeMetadata();
+                resetOffsetsIfNeededAndInitializeMetadata(offsetResetter);
```
ableegoldman, line 249: cool, thanks, this seems much cleaner to me

ableegoldman (approved):
Ok I think this LG -- but I'll be happy to get this soaking asap

PR #10004, opened 2021-01-29
MINOR : add memberId to JoinGroup Success message

nicolasguyomar:
Hi team,

Minor improvement to log the received memberId upon joining a group at INFO level on the memberside (consumer, worker) so that we can easily track broker logs with the member logs without the need to increase log verbosity

Thank you

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
@@ -591,7 +591,7 @@ public abstract class AbstractCoordinator implements Closeable {
                                 joinResponse.data().generationId(),
                                 joinResponse.data().memberId(), joinResponse.data().protocolName());
 
-                            log.info("Successfully joined group with generation {}", AbstractCoordinator.this.generation);
+                            log.info("Successfully joined group with generation {} and memberId {}", generation.generationId, generation.memberId);
 
                             if (joinResponse.isLeader()) {
                                 onJoinLeader(joinResponse).chain(future);


clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
```diff
@@ -591,7 +591,7 @@ public void handle(JoinGroupResponse joinResponse, RequestFuture<ByteBuffer> fut
                                 joinResponse.data().generationId(),
                                 joinResponse.data().memberId(), joinResponse.data().protocolName());
 
-                            log.info("Successfully joined group with generation {}", AbstractCoordinator.this.generation);
+                            log.info("Successfully joined group with generation {} and memberId {}", generation.generationId, generation.memberId);
```
tombentley, line 594: If we're removing the redundant `AbstractCoordinator.this` here, we might as well do it 4 lines above too, imo.

tombentley (approved):
@abbccdda @guozhangwang any chance one of you could review this?

clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractCoordinator.java
dajac: @nicolasguyomar We already log the memberId here as we log the entire generation object (which includes the memberId). This was changed recently: https://github.com/apache/kafka/commit/7e7bb184d2abe34280a7f0eb0f0d9fc0e32389f2#diff-15efe9b844f78b686393b6c2e2ad61306c3473225742caed05c7edab9a138832L504. Previously, it was logging the generationId only.

nicolasguyomar: Thank you @dajac this is exactly what I needed at INFO level yes, I'll close that PR

PR #9577, opened 2020-11-09
KAFKA-9837: KIP-589 new RPC for notifying controller log dir failure

dengziming:
This patch implements [KIP-589](https://cwiki.apache.org/confluence/display/KAFKA/KIP-589+Add+API+to+update+Replica+state+in+Controller), which introduces an asynchronous API for brokers to notifying the controller of log dir failure.

*Summary of testing strategy (including rationale)
for the feature or bug fix. Unit and/or integration
tests are expected for any behaviour change and
system tests should be considered for larger changes.*
1. Unit test for LogDirEventManagerImpl
2. Integration test for new behavior

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #10005, opened 2021-01-29
MINOR: Add ConfigRepository, use in Partition and KafkaApis

rondagostino:
`Partition` objects are able to retrieve topic configs when creating their log, and currently they do so with an implementation of `trait TopicConfigFetcher` that uses ZooKeeper.  ZooKeeper is not available when using a Raft-based metadata log, so we need to abstract the retrieval of configs so it can work either with or without ZooKeeper.  This PR introduces `trait ConfigRepository` with `ZkConfigRepository` and `CachedConfigRepository` implementations.  `Partition` objects now use a provided `ConfigRepository` to retrieve topic configs, and we eliminate `TopicConfigFetcher` as it is no longer needed.

`ReplicaManager` now contains an instance of `ConfigRepository` so it can provide it when creating `Partition` instances.

`KafkaApis` needs to be able to handle describe-config requests; it currently delegates that to `ZkAdminManager`, which of course queries ZooKeeper.  To make this work with or without ZooKeeper we move the logic from `ZkAdminManager` into a new `ConfigHelper` class that goes through a `ConfigRepository` instance.  We provide `KafkaApis` with such an instance, and it creates an instance of the helper so it can use that instead of going through `ZkAdminManager`.

Existing tests are sufficient to identify bugs and regressions in `Partition`, `ReplicaManager`, `KafkaApis`, and `ConfigHelper`.  The `ConfigRepository` implementations have their own unit tests.


core/src/main/scala/kafka/log/LogConfig.scala
```diff
@@ -374,40 +374,4 @@ object LogConfig {
     MessageTimestampDifferenceMaxMsProp -> KafkaConfig.LogMessageTimestampDifferenceMaxMsProp,
     MessageDownConversionEnableProp -> KafkaConfig.LogMessageDownConversionEnableProp
   )
-
-
-  /**
-   * Copy the subset of properties that are relevant to Logs. The individual properties
```
hachikuji: We moved this here from `KafkaServer`. Not sure why we want to revert that. `LogConfig` seems like the right place to have a function to extract the log config properties.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -119,7 +122,21 @@ class KafkaApis(val requestChannel: RequestChannel,
 
   type FetchResponseStats = Map[TopicPartition, RecordConversionStats]
   this.logIdent = "[KafkaApi-%d] ".format(brokerId)
-  val adminZkClient = new AdminZkClient(zkClient)
+
+  val usingRaftMetadataQuorum = if (adminManager != null && controller != null && zkClient != null) {
```
hachikuji: Can we use `config.requiresZookeeper` here?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -119,7 +122,21 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+    false
+  } else { // at least one is null, so make sure they are all null, plus we must have a forwardingManager
+    if (adminManager != null || controller != null || zkClient != null) {
```
hachikuji: This kind of logic might be better suited for a factory. Something like this:
```scala
object KafkaApis {
  def buildZk(zkClient: KafkaZkClient, ...): KafkaApis = {
     // verify adminManager, controller, etc
  }
  def buildSelfManaged(...): KafkaApis = {
    // verify forwarding manager, etc.
  }
}
```

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -97,12 +100,12 @@ import scala.annotation.nowarn
...
-                val adminManager: ZkAdminManager,
+                adminManager: ZkAdminManager,
                 val groupCoordinator: GroupCoordinator,
                 val txnCoordinator: TransactionCoordinator,
-                val controller: KafkaController,
-                val forwardingManager: Option[ForwardingManager],
-                val zkClient: KafkaZkClient,
+                controller: KafkaController,
+                forwardingManager: Option[ForwardingManager],
+                zkClient: KafkaZkClient,
```
hachikuji: Can we use `Option` for all of these types that might be null?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -150,7 +167,11 @@ class KafkaApis(val requestChannel: RequestChannel,
     }
 
     forwardingManager match {
-      case Some(mgr) if !request.isForwarded && !controller.isActive =>
+      case Some(mgr) if request.isForwarded && usingRaftMetadataQuorum =>
```
hachikuji: The extra checks in here seem unnecessary. If the forwarding manager is provided, we can use it.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -1654,6 +1702,7 @@ class KafkaApis(val requestChannel: RequestChannel,
           .format(response, request.header.correlationId, request.header.clientId))
         response
       }
+
```
hachikuji: nit: do we need all these additional newlines? they are a tad distracting

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -249,6 +270,7 @@ class KafkaApis(val requestChannel: RequestChannel,
   }
 
   def handleLeaderAndIsrRequest(request: RequestChannel.Request): Unit = {
+    rejectWithUnsupportedVersionIfUsingRaftMetadataQuorum(request)
```
hachikuji: Do we excludes these APIs from the ApiVersions response when raft is enabled? It would be a little weird if `ApiVersions` indicates that an API is supported, but then we return UNSUPPORTED_VERSION when the client tries to use it.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -2609,14 +2672,27 @@ class KafkaApis(val requestChannel: RequestChannel,
         case rt => throw new InvalidRequestException(s"Unexpected resource type $rt")
       }
     }
-    val authorizedResult = adminManager.alterConfigs(authorizedResources, alterConfigsRequest.validateOnly)
-    val unauthorizedResult = unauthorizedResources.keys.map { resource =>
-      resource -> configsAuthorizationApiError(resource)
+    // Should never get here when using a Raft metadata quorum since it would be forwarded,
```
hachikuji: If you think the protection is worthwhile, then can we do something like `rejectWithUnsupportedVersionIfUsingRaftMetadataQuorum`?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -2965,10 +3064,10 @@ class KafkaApis(val requestChannel: RequestChannel,
     val electionRequest = request.body[ElectLeadersRequest]
 
     def sendResponseCallback(
-      error: ApiError
-    )(
-      results: Map[TopicPartition, ApiError]
-    ): Unit = {
+                              error: ApiError
```
hachikuji: nit: not sure what happened with the indentation here

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3008,7 +3107,8 @@ class KafkaApis(val requestChannel: RequestChannel,
       })
     }
 
-    if (!authHelper.authorize(request.context, ALTER, CLUSTER, CLUSTER_NAME)) {
+    // not yet supported when using a Raft-based metadata quorum
+    if (usingRaftMetadataQuorum || !authHelper.authorize(request.context, ALTER, CLUSTER, CLUSTER_NAME)) {
```
hachikuji: `CLUSTER_AUTHORIZATION_FAILED` seems like a surprising error for an unsupported operation. Any of the following errors might be better:
-  `UNKNOWN_SERVER_ERROR` 
- `OPERATION_NOT_ATTEMPTED`
- `UNSUPPORTE_VERSION`

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3316,8 +3440,8 @@ class KafkaApis(val requestChannel: RequestChannel,
   }
 
   private def parseForwardedClientAddress(
-    address: Array[Byte]
-  ): InetAddress = {
+                                           address: Array[Byte]
```
hachikuji: nit: fix indentation

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3450,7 +3574,187 @@ class KafkaApis(val requestChannel: RequestChannel,
       brokerEpochInRequest < controller.brokerEpoch
     }
   }
+}
+
+class ConfigHelper(metadataCache: MetadataCache, config: KafkaConfig, configRepository: ConfigRepository) extends Logging {
```
hachikuji: Can we move this to a separate file?

core/src/main/scala/kafka/server/KafkaApis.scala
rondagostino: I reworked it a bit to clarify the intent, which is to add a log message if a Raft-based broker receives a forwarded request; it'll still handle the request (and we'll give `NOT_CONTROLLER`).

I'm not sure we should change the logic here and forward a request that itself is already forwarded as that opens up the possibility of endless loops.

core/src/main/scala/kafka/server/KafkaApis.scala
rondagostino: I think `NOT_CONTROLLER` is the clearest thing we could return.

core/src/main/scala/kafka/server/KafkaApis.scala
rondagostino: I'm not sure how all of these spacing changes happened -- I didn't explicitly do them.  My guess is the editor tried to be helpful by formatting things at some point.  I reverted these.

core/src/main/scala/kafka/server/KafkaApis.scala
rondagostino: We probably should, yes.  When looking at it, the exclusion should be done in `ApiVersion.apiVersionsResponse()`.  But how we can know there if we are using a Raft-based metadata quorum is unclear.  We could brute-force send in the `KafkaConfig`, but it appears perhaps `Features` might be another way to do it.  Have we considered creating a feature for that, do you know?

core/src/main/scala/kafka/server/KafkaApis.scala
rondagostino: Changed to `UnsupportedVersionException`

hachikuji:
@rondagostino Thanks for the updates. I am wondering if we can pull out the changes in `KafkaApis` since most of them are unrelated to the `ConfigRepository` refactor. In general, I think we need a better approach for managing the dependency requirements for the zk/self-managed implementations. One idea here is try to encapsulate the zk-specific dependencies into a separate object. For example:

```scala
trait MetadataSupport

class ZkSupport(
  adminManager: ZkAdminManager,
  zkClient: KafkaZkClient,
  ...
) extends MetadataSupport
```
Then we can pass through `MetadataSupport` and get rid of the optional types. We can then have helpers like the following:
```scala
def requireZk(): ZkSupport = {
  support match {
    case zkSupport: ZkSupport => zkSupport
    case _ => throw new IllegalStateException
  }
}
```  

The advantage is that we can make the dependencies that are available when zk support is enabled explicit. We don't have to deal with all the permutations that are possible when we have 5 different `Option` classes.

We can then start making this trait more useful by adding methods. For example, instead of reaching into the `KafkaController` to get the broker epoch, we can add a `brokerEpoch` method. Eventually we might get to a point where we don't need so many special cases.

The second thing I think we can do better is how we handle the acceptable scope of the APIs when using zk or the self-managed quorum. The patch does this by adding separate checks in each of the individual handlers. I filed https://issues.apache.org/jira/browse/KAFKA-12278 to try and handle this problem in a more general way, and so that we can also keep ApiVersions consistent. I would rather address this problem in that JIRA if it is ok with you.

core/src/main/scala/kafka/server/KafkaApis.scala
rondagostino: This no longer applies to this PR as we have removed the `KafkaApis` changes from it.

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -89,8 +88,7 @@ object Partition extends KafkaMetricsGroup {
 
     val configProvider = new TopicConfigFetcher {
       override def fetch(): Properties = {
-        val adminZkClient = new AdminZkClient(replicaManager.zkClient)
-        adminZkClient.fetchEntityConfig(ConfigType.Topic, topicPartition.topic)
+        replicaManager.configRepository.topicConfigs(topicPartition.topic())
```
hachikuji: Hmm.. It seems like `ConfigRepository` can just replace `TopicConfigFetcher`?

core/src/main/scala/kafka/server/metadata/ConfigRepository.scala
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kafka.server.metadata
+
+import java.util.Properties
+
+trait ConfigRepository {
+  def topicConfigs(name: String): Properties
```
hachikuji: nit: can we be explicit and rename the args to `topicName` and `brokerId`?

Also, it would be useful to document the fact that the returned value is a copy of current configuration and will not reflect future changes.

core/src/main/scala/kafka/server/metadata/LocalConfigRepository.scala
```diff
@@ -0,0 +1,93 @@
...
+
+import scala.jdk.CollectionConverters._
+
+/**
+ * A ConfigRepository that stores configurations locally.
+ */
+class LocalConfigRepository extends ConfigRepository {
+  val configMap = new ConcurrentHashMap[ConfigResource, util.HashMap[String, String]]
+
+  def setTopicConfig(topic: String, key: String, value: String): Unit = {
```
hachikuji: Useful to add some documentation about the contract. For example, it looks like we expect value=null to indicate deletion (and on that note, it might be clearer to have an explicit `removeTopicConfig`).

core/src/main/scala/kafka/server/metadata/LocalConfigRepository.scala
```diff
@@ -0,0 +1,93 @@
...
+            newConfigs.putAll(curConfigs)
+            newConfigs.put(key, value)
+            newConfigs
+          }
+        }
+      }
+    })
+  }
+
+  def config(configResource: ConfigResource): Properties = {
```
hachikuji: nit: can we be consistent on pluralization? I think I tend to see `config` as plural already, so maybe we can use `topicConfig` and `brokerConfig`.

Really we could even pull this up to `ConfigRepository`. Maybe something like this:

```scala
trait ConfigRepository {
  def config(configResource: ConfigResource): Properties
  
  def topicConfig(topic: String): Properties = {
    config(new ConfigResource(Type.TOPIC, topic))
  }

  def brokerConfig(id: Int): Properties = {
    config(new ConfigResource(Type.BROKER, id.toString()))
  }
}
```

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -120,6 +120,7 @@ class KafkaApis(val requestChannel: RequestChannel,
   type FetchResponseStats = Map[TopicPartition, RecordConversionStats]
   this.logIdent = "[KafkaApi-%d] ".format(brokerId)
   val adminZkClient = new AdminZkClient(zkClient)
+  val configHelper = new ConfigHelper(metadataCache, config, replicaManager.configRepository)
```
hachikuji: nit: it's a little ugly to reach into `ReplicaManager` in order to get the dependency we need. This also makes the mocking more difficult. Can we instead add `ConfigRepository` as a formal parameter?

core/src/main/scala/kafka/cluster/Partition.scala
rondagostino: `TopicConfigFetcher` is specific to a single topic; `ConfigRepository` is not.  I don't think one can directly replace the other -- unless I am misunderstanding/missing something?

hachikuji: We have access to the topic name in `Partition`, so my thought was to replace the calls to `TopicConfigFetcher.fetch()` with `ConfigRepository.topicConfig(topic)`

core/src/main/scala/kafka/cluster/Partition.scala
```diff
@@ -342,7 +331,7 @@ class Partition(val topicPartition: TopicPartition,
   // Visible for testing
   private[cluster] def createLog(isNew: Boolean, isFutureReplica: Boolean, offsetCheckpoints: OffsetCheckpoints): Log = {
     def fetchLogConfig: LogConfig = {
-      val props = topicConfigProvider.fetch()
+      val props = configRepository.topicConfig(topicPartition.topic())
```
hachikuji: nit: we can use `configRepository.topicConfig(topic)`

core/src/main/scala/kafka/server/metadata/LocalConfigRepository.scala
```diff
@@ -0,0 +1,136 @@
...
+
+import org.apache.kafka.common.config.ConfigResource
+import org.apache.kafka.common.config.ConfigResource.Type
+
+import scala.jdk.CollectionConverters._
+
+/**
+ * A ConfigRepository that stores configurations locally.
+ */
+class LocalConfigRepository extends ConfigRepository {
```
hachikuji: I found the term "local" a little confusing. How about `CachedConfigRepository`?

core/src/main/scala/kafka/server/metadata/LocalConfigRepository.scala
```diff
@@ -0,0 +1,136 @@
...
+    setConfig(new ConfigResource(Type.TOPIC, topicName), key, value)
+  }
+
+  /**
+   * Remove the broker config for the given broker ID and the given key.
+   *
+   * @param brokerId the ID of the broker for which the config will be removed
+   * @param key the key identifying the topic config to remove
+   */
+  def removeBrokerConfig(brokerId: Int, key: String): Unit = {
```
hachikuji: nit: I think it is better to have just one way to remove configs. I know I suggested this, but I was assuming we would not also support removal through `set`. If you think using `set` with a null value is clear enough, we can remove these.

core/src/main/scala/kafka/server/KafkaServer.scala
```diff
@@ -348,15 +348,15 @@ class KafkaServer(
 
         /* start processing requests */
         dataPlaneRequestProcessor = new KafkaApis(socketServer.dataPlaneRequestChannel, replicaManager, adminManager, groupCoordinator, transactionCoordinator,
-          kafkaController, forwardingManager, zkClient, config.brokerId, config, metadataCache, metrics, authorizer, quotaManagers,
+          kafkaController, forwardingManager, zkClient, config.brokerId, config, replicaManager.configRepository, metadataCache, metrics, authorizer, quotaManagers,
```
hachikuji: nit: Similarly here, it would be cleaner to construct the config repository separately in this class and pass it down to the components that need it including `ReplicaManager` and `KafkaApis`.

core/src/main/scala/kafka/server/metadata/ZkConfigRepository.scala
```diff
@@ -0,0 +1,39 @@
...
+
+object ZkConfigRepository {
+  def apply(zkClient: KafkaZkClient): ZkConfigRepository =
+    new ZkConfigRepository(new AdminZkClient(zkClient))
+}
+
+class ZkConfigRepository(adminZkClient: AdminZkClient) extends ConfigRepository {
+  override def config(configResource: ConfigResource): Properties = {
+     adminZkClient.fetchEntityConfig(
+       if (configResource.`type`() == Type.TOPIC) ConfigType.Topic else ConfigType.Broker,
```
hachikuji: Hmm... It would be better to have an explicit check for the `Broker` type. We can raise an exception for any remaining types.

hachikuji (commented):
Thanks for the updates. Looks good overall. Left a few additional comments.

core/src/main/scala/kafka/server/KafkaServer.scala
rondagostino: @hachikuji After this change we have only 1 place left where we reach into `ReplicaManager` to grab its `ConfigRepository` -- when we create a `Partition` instance.  Do you think we should change that so that we explicitly provide the instance to the `Partition` constructor and the companion object's `apply()` method?  Aside form this last question I think we're all set here.

rondagostino: It's just the `apply()` method that reaches in, so I'm going to add an explicit `ConfigRepository` parameter there.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -1881,7 +1884,11 @@ class ReplicaManager(val config: KafkaConfig,
     logManager.handleLogDirFailure(dir)
 
     if (sendZkNotification)
-      zkClient.propagateLogDirEvent(localBrokerId)
+      if (zkClient == null) {
```
hachikuji: Can we make `zkClient` an `Option`? By the way, we are really close to being able to remove `KafkaZkClient` as a dependency, but I guess we will first need #9577.

hachikuji (approved):
LGTM

PR #10006, opened 2021-01-29
KAFKA-12245; Fix flaky `FetcherTest.testEarlierOffsetResetArrivesLate`

hachikuji:
This test has been flaky due to a race condition of some kind. I rewrote it so that it no longer relies on multiple threads.


clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
```diff
@@ -1745,51 +1742,77 @@ public void testSeekWithInFlightReset() {
...
+        }, listOffsetResponse(Errors.NONE, 1L, 0L));
+        consumerClient.pollNoWakeup();
+
+        // The list offset result should be ignored
+        assertTrue(subscriptions.isOffsetResetNeeded(tp0));
+        assertEquals(OffsetResetStrategy.LATEST, subscriptions.resetStrategy(tp0));
+
+        fetcher.resetOffsetsIfNeeded();
+        client.prepareResponse(req -> {
+            return listOffsetMatchesExpectedReset(tp0, OffsetResetStrategy.LATEST, req);
```
abbccdda: `return` is not necessary

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
```diff
@@ -1745,51 +1742,77 @@ public void testSeekWithInFlightReset() {
...
+
+        return true;
+    }
+
     @Test
-    public void testEarlierOffsetResetArrivesLate() throws InterruptedException {
-        LogContext lc = new LogContext();
-        buildFetcher(spy(new SubscriptionState(lc, OffsetResetStrategy.EARLIEST)), lc);
+    public void testEarlierOffsetResetArrivesLate() {
+        buildFetcher();
```
abbccdda, line 1771: Why we don't need to set reset strategy here?

abbccdda (commented):
I'm not sure I found evidence to support that the flakiness was due to multiple threads. Does the PR just try to reduce the test complexity?

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
hachikuji: Because the test case sets it explicitly below.

hachikuji:
@abbccdda See for example this trace, which is often found in an unrelated test case (https://github.com/apache/kafka/runs/1794689847):
```
org.mockito.exceptions.misusing.UnfinishedStubbingException: 
Unfinished stubbing detected here:
-> at org.apache.kafka.clients.consumer.internals.FetcherTest.testEarlierOffsetResetArrivesLate(FetcherTest.java:1761)

E.g. thenReturn() may be missing.
Examples of correct stubbing:
    when(mock.isOk()).thenReturn(true);
    when(mock.isOk()).thenThrow(exception);
    doThrow(exception).when(mock).someVoidMethod();
Hints:
 1. missing thenReturn()
 2. you are trying to stub a final method, which is not supported
 3. you are stubbing the behaviour of another mock inside before 'thenReturn' instruction is completed
```
I could not come up with an explanation for this, nor could I reproduce it. What I suspect is that mockito is not really safe for multi-threaded access as the test case assumes. After looking at it for a little while, it seemed better to rewrite it since we do not need multiple threads.

chia7712:
@hachikuji According to my experience of using Mockito, it can caused undefined behavior if we change class loader in mocking/spying object. Hence, I traced client tests and there is a test case ```AbstractConfigTest#testClassConfigs``` (https://github.com/apache/kafka/blob/trunk/clients/src/test/java/org/apache/kafka/common/config/AbstractConfigTest.java#L280) which does not reset the class loader. Hence, the class loader is changed if ```AbstractConfigTest``` and ```FetcherTest``` are executed in same JVM. I reset the class loader and loop clients tests 30 times. ```FetcherTest``` pass.

This patch is ok to me. Personally, we could make ```AbstractConfigTest#testClassConfigs``` reset class loader to avoid similar issues in the future.

hachikuji:
@chia7712 That makes sense. Do you want to do that in a separate PR? I think it still makes sense to simplify this test case.

chia7712:
> Do you want to do that in a separate PR? 

sure. Please take a look at #10010

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
```diff
@@ -1745,51 +1742,78 @@ public void testSeekWithInFlightReset() {
...
     }
 
-    @Timeout(10)
+    private boolean listOffsetMatchesExpectedReset(
+        TopicPartition tp,
+        OffsetResetStrategy strategy,
+        AbstractRequest request
+    ) {
+        if (!(request instanceof ListOffsetsRequest)) {
+            return false;
```
chia7712: Why not throwing exception straightforwardly when receiving those unexpected requests?

chia7712 (approved):
+1 to this nice refactor.

one small question is left. Please take a look.

clients/src/test/java/org/apache/kafka/clients/consumer/internals/FetcherTest.java
hachikuji: I pushed an update to turn these into assertions. I also filed https://issues.apache.org/jira/browse/KAFKA-12263.

PR #10007, opened 2021-01-31
KAFKA-10700 - Support mutual TLS authentication for SASL_SSL listeners (KIP-684)

rajinisivaram:
mTLS is enabled if listener-prefixed `ssl.client.auth` is configured for SASL_SSL listeners. Broker-wide `ssl.client.auth` is not applied to SASL_SSL listeners as before, but we now print a warning.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/common/network/ChannelBuilders.java
+++ b/clients/src/main/java/org/apache/kafka/common/network/ChannelBuilders.java
@@ -18,6 +18,7 @@ package org.apache.kafka.common.network;
 
 import org.apache.kafka.common.Configurable;
 import org.apache.kafka.common.config.AbstractConfig;
+import org.apache.kafka.common.config.SslClientAuth;
 import org.apache.kafka.common.config.internals.BrokerSecurityConfigs;
 import org.apache.kafka.common.errors.InvalidConfigurationException;
 import org.apache.kafka.common.security.auth.SecurityProtocol;
@@ -31,13 +32,18 @@ import org.apache.kafka.common.security.token.delegation.internals.DelegationTok
 import org.apache.kafka.common.utils.LogContext;
 import org.apache.kafka.common.utils.Time;
 import org.apache.kafka.common.utils.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 public class ChannelBuilders {
+    private static final Logger log = LoggerFactory.getLogger(ChannelBuilders.class);
+
     private ChannelBuilders() { }
 
     /**
@@ -112,6 +118,7 @@ public class ChannelBuilders {
                                          Time time,
                                          LogContext logContext) {
         Map<String, Object> configs = channelBuilderConfigs(config, listenerName);
+        String sslClientAuthOverride = null;
 
         ChannelBuilder channelBuilder;
         switch (securityProtocol) {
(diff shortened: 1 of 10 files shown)


clients/src/test/java/org/apache/kafka/common/security/authenticator/SaslAuthenticatorTest.java
```diff
@@ -1777,6 +1785,74 @@ public void testInsufficientScopeSaslOauthBearerMechanism() throws Exception {
...
+        verifySslClientAuthForSaslSslListener(false, SslClientAuth.REQUIRED);
+    }
+
+    private void verifySslClientAuthForSaslSslListener(boolean useListenerPrefix,
+                                                       SslClientAuth configuredClientAuth) throws Exception {
+
+        SecurityProtocol securityProtocol = SecurityProtocol.SASL_SSL;
+        configureMechanisms("PLAIN", Collections.singletonList("PLAIN"));
+        String listenerPrefix = useListenerPrefix ? ListenerName.forSecurityProtocol(securityProtocol).configPrefix() : "";
+        saslServerConfigs.put(listenerPrefix  + BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG, configuredClientAuth.name());
```
dajac: nit: There is an extra space before `+`.

clients/src/main/java/org/apache/kafka/common/network/ChannelBuilders.java
```diff
@@ -112,6 +118,7 @@ private static ChannelBuilder create(SecurityProtocol securityProtocol,
                                          Time time,
                                          LogContext logContext) {
         Map<String, Object> configs = channelBuilderConfigs(config, listenerName);
+        String sslClientAuthOverride = null;
```
dajac: This variable is only used for `SASL_SSL` and `SASL_PLAINTEXT`. Could we move its declaration within the `case` block?

clients/src/main/java/org/apache/kafka/common/network/ChannelBuilders.java
```diff
@@ -129,6 +136,22 @@ private static ChannelBuilder create(SecurityProtocol securityProtocol,
                     jaasContexts = new HashMap<>(enabledMechanisms.size());
                     for (String mechanism : enabledMechanisms)
                         jaasContexts.put(mechanism, JaasContext.loadServerContext(listenerName, mechanism, configs));
+
+                    // SSL client authentication is enabled in brokers for SASL_SSL only if listener-prefixed config is specified.
+                    if (listenerName != null && securityProtocol == SecurityProtocol.SASL_SSL) {
+                        String configuredClientAuth = (String) configs.get(BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG);
+                        String listenerClientAuth = (String) config.originalsWithPrefix(listenerName.configPrefix(), true)
+                                .get(BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG);
```
dajac, line 144: Constructing the config for the listener prefix seems a bit wasteful just to check that a config exists. I wonder if this is really necessary and if we could check `listenerName.configPrefix() + BrokerSecurityConfigs.SSL_CLIENT_AUTH_CONFIG` in `config` directly instead. Maybe it does not matter too much here as we do this only once after all. What do you think?

clients/src/main/java/org/apache/kafka/common/network/ChannelBuilders.java
```diff
@@ -129,6 +136,22 @@ private static ChannelBuilder create(SecurityProtocol securityProtocol,
...
+                        if (listenerClientAuth == null) {
+                            sslClientAuthOverride = SslClientAuth.NONE.name().toLowerCase(Locale.ROOT);
```
dajac, line 151: I was a bit confused by this at first as I was wondering why `listenerClientAuth` is not set to `listenerClientAuth` here are well. `null` means that we let the ssl factory get the configuration from `configs` later on. We could perhaps add a comment to make this clearer.

Alternatively, could we just override `SSL_CLIENT_AUTH_CONFIG` in `configs` directly here instead of doing it in `SslFactory#configure`? This might make the handling clearer here. I don't feel strong about this. What do you think?

dajac (commented):
@rajinisivaram Thanks for the PR. I have left few minor comments.

clients/src/main/java/org/apache/kafka/common/network/ChannelBuilders.java
rajinisivaram: We can't do `config.getString(prefix+configName)` because `AbstractConfig` only allows configs from its definition to be obtained using `get` and the prefixed config is not treated as a defined config. I couldn't think of alternative way that doesn't need the originals in some form. As you said, it is only used once when broker starts up, so should be ok. But happy to change if you can think of an alternative.

clients/src/main/java/org/apache/kafka/common/network/ChannelBuilders.java
rajinisivaram: We let SslFactory deal with the override since it needs to apply the override when it is reconfigured.  Added a comment. Let me know what you think.

rajinisivaram:
@dajac Thanks for the review. Addressed comments and left a couple of responses.

clients/src/main/java/org/apache/kafka/common/network/ChannelBuilders.java
dajac: Ah right. I can't think of a better way neither.

clients/src/main/java/org/apache/kafka/common/network/ChannelBuilders.java
dajac: That makes sense. I did not think about the reconfiguration case.

dajac (approved):
LGTM. Thanks for the updates.

rajinisivaram:
@dajac Thanks for the review. Test failures not related, merging to trunk.

PR #10008, opened 2021-01-31
MINOR: Remove ZK dependency for coordinator topics' partition counts

rondagostino:
`GroupMetadataManager` currently uses ZooKeeper to obtain any non-default value for the `__consumer_offsets` partition count. ZooKeeper will not be available when the broker uses a Raft-based metadata quorum. This PR removes the hard dependency on ZooKeeper by allowing a function returning an `Int` to be provided instead. Providing a ZooKeeper client is still supported in `apply()` methods, but those methods transparently wrap the client with a function and pass that function instead.

`TransactionStateManager` currently uses ZooKeeper to obtain any non-default value for the `__transaction_state` partition count.  ZooKeeper will not be available when the broker uses a Raft-based metadata quorum.  This PR removes the hard dependency on ZooKeeper by allowing a function returning an `Int` to be provided instead.  Providing a ZooKeeper client is still supported in `apply()` methods, but those methods transparently wrap the client with a function and pass that function instead.

Existing tests are sufficient to prevent bugs and regressions.


ijuma:
Did we consider passing a function instead of zk client?

core/src/main/scala/kafka/server/KafkaServer.scala
```diff
@@ -318,8 +318,9 @@ class KafkaServer(
 
         /* start group coordinator */
         // Hardcode Time.SYSTEM for now as some Streams tests fail otherwise, it would be good to fix the underlying issue
-        groupCoordinator = GroupCoordinator(config, zkClient, replicaManager, Time.SYSTEM, metrics)
-        groupCoordinator.startup()
+        groupCoordinator = GroupCoordinator(config, replicaManager, Time.SYSTEM, metrics)
+        groupCoordinator.startup(zkClient.getTopicPartitionCount(Topic.GROUP_METADATA_TOPIC_NAME).
```
ijuma: Why not pass the number of partitions in the constructor instead?

rondagostino: Good question.  We won't know what value to provide when using a Raft-based metadata quorum until we have caught up on the metadata log and know our configs.  Providing it in the `startup()` function has no impact in the ZooKeeper world because we invoke that immediately after constructing the instance.  In the Raft-based world, if we provide it in the constructor, it constrains us to not construct the instance until we have caught up and have our configs.  By providing it in the `startup()` method it gives us additional flexibility to potentially instantiate the group coordinator but not yet "start" it.

ijuma: What's the advantage of having a coordinator that is not started?

rondagostino: Yeah, it is difficult to justify beyond the fact that `startup()` exists -- maybe it deserves to be gotten rid of.  But that's perhaps a discussion for a separate ticket/PR.

ijuma: The fact that we have stuff like the following implies that the approach taken here isn't so good:
* https://github.com/apache/kafka/pull/10008/files#diff-8bdfc05a89875ddd25c9bd355c102ca2a0556795bb18fba25efc3f9959d789e2R199
* https://github.com/apache/kafka/pull/10008/files#diff-8bdfc05a89875ddd25c9bd355c102ca2a0556795bb18fba25efc3f9959d789e2R188

rondagostino: Ok, I reverted and used the same approach as is proposed for the `__transaction_state` partition count in https://github.com/apache/kafka/pull/10013.

cmccabe: > What's the advantage of having a coordinator that is not started?

This is a pattern that we are using for a lot of components in the kip-500 branch.  It reflects the fact that in the kip-500 world, we have a "before metadata is available" state and an "after metadata is available" state.  The advantage is that it makes this state explicit rather than being implicit.  When the state is implicit mistakes can easily happen-- for example, we don't want someone to modify the code to ask for some piece of metadata that is not yet available.  The original author of the code may have "just known" that this was not allowed, but it's better to formalize it so that people don't make mistakes.

I think the callback approach is really clunky and tends to obfuscate the code.  What's going on here is really that the number of partitions gets initialized once.  Having a complex callback chain just obscures that and makes it tough to read.

core/src/main/scala/kafka/coordinator/transaction/TransactionStateManager.scala
```diff
@@ -96,7 +108,20 @@ class TransactionStateManager(brokerId: Int,
   private[transaction] val transactionMetadataCache: mutable.Map[Int, TxnMetadataCacheEntry] = mutable.Map()
 
   /** number of partitions for the transaction log topic */
-  private val transactionTopicPartitionCount = getTransactionTopicPartitionCount
+  private var _transactionTopicPartitionCount: Option[Int] = Option.empty // lazy, once-only evaluation
```
mumrah: I think this should probably be `@volatile` since the reads are not synchronized. Maybe we can use `AtomicReference` or some future thing here?

core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala
```diff
@@ -78,7 +78,20 @@ class GroupMetadataManager(brokerId: Int,
   private val shuttingDown = new AtomicBoolean(false)
 
   /* number of partitions for the consumer metadata topic */
-  private val groupMetadataTopicPartitionCount = getGroupMetadataTopicPartitionCount
+  private var _groupMetadataTopicPartitionCount: Option[Int] = Option.empty // lazy, once-only evaluation
```
mumrah: Same volatile question as below

mumrah:
What is the motivation for making transactionTopicPartitionCount and groupMetadataTopicPartitionCount lazy? Is it because in the KIP-500 world, we may not yet have this metadata from the Raft quorum so we can't look this up on startup? 

If that's the case, are we guaranteed to have loaded the metadata before the first usages of GroupMetadataManager and TransactionManager?

core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala
```diff
@@ -78,7 +78,20 @@ class GroupMetadataManager(brokerId: Int,
...
+  private def groupMetadataTopicPartitionCount: Int = {
+    _groupMetadataTopicPartitionCount match {
+      case Some(partitionCount) => partitionCount
+      case None => synchronized { // make sure we only invoke the function once
+        _groupMetadataTopicPartitionCount match {
+          case Some(partitionCount) => partitionCount // another thread beat us to it
+          case None =>
+            _groupMetadataTopicPartitionCount = Some(offsetsTopicPartitionCountFunc())
```
mumrah: It's worth noting that if our underlying function throws an error, this will throw as well for any usages of `groupMetadataTopicPartitionCount` (e.g., `GroupMetadataManager#partitionFor`). That wasn't possible in the ZK world (since this value was cached on startup), so we should probably make sure the callers can all tolerate an exception.

ijuma:
We discussed this offline and the current suggestion is to pass the function in the `start` method.

core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala
```diff
@@ -78,7 +76,7 @@ class GroupMetadataManager(brokerId: Int,
   private val shuttingDown = new AtomicBoolean(false)
 
   /* number of partitions for the consumer metadata topic */
-  private val groupMetadataTopicPartitionCount = getGroupMetadataTopicPartitionCount
+  private var groupMetadataTopicPartitionCount: Int = _
```
rondagostino: Wonder if this needs `@volatile`.  It is only set once, upon invoking `startup()`, at which time it changes from 0 to the partition count.  The only place this value is used is in `partitionFor()` where it is the denominator and will cause division-by-zero if the original value is read instead of the updated value.

ijuma: If it's used by multiple threads and not set in the constructor, then it needs to be volatile. If it's only set in the constructor and there is at least one final field, then it would not need volatile.

core/src/main/scala/kafka/coordinator/transaction/TransactionStateManager.scala
```diff
@@ -96,7 +94,8 @@ class TransactionStateManager(brokerId: Int,
   private[transaction] val transactionMetadataCache: mutable.Map[Int, TxnMetadataCacheEntry] = mutable.Map()
 
   /** number of partitions for the transaction log topic */
-  private val transactionTopicPartitionCount = getTransactionTopicPartitionCount
+  private var retrieveTransactionTopicPartitionCount: () => Int = _
+  private var transactionTopicPartitionCount: Int = _
```
rondagostino: Same question here about `@volatile`.  The value is set once at startup and is used in a denominator within `partitionFor()`

core/src/main/scala/kafka/coordinator/group/GroupMetadataManager.scala
rondagostino: This is no longer an issue since we invoke the function once in `startup()`.  We do it again in `TransactionStateManager` whenever we become leader for a partition, but that happens in the scheduler thread where any exception will be caught/logged.

ijuma:
We finally got a non-red build, some unrelated flaky tests in JDK 11 and 15. JDK 8 passed.

cmccabe:
Failing tests are 
`org.apache.kafka.connect.integration.RebalanceSourceConnectorsIntegrationTest.testMultipleWorkersRejoining`
and `org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationSSLTest.testOneWayReplicationWithAutoOffsetSync()` which are not related.

cmccabe (approved):
LGTM

PR #10009, opened 2021-02-01
MINOR: Introduce ProducerIdGenerator trait

rondagostino:
`ProducerIdManager` is an existing class that talks to ZooKeeper directly.  We won't have ZooKeeper when using a Raft-based metadata quorum, so we need an abstraction for the functionality of generating producer IDs.  This PR introduces `ProducerIdGenerator` for this purpose, and we pass an implementation when instantiating `TransactionCoordinator` rather than letting `TransactionCoordinator.apply()` itself always create a ZooKeeper-based instance.

comment to ping reviewers.*

### DIFF
--- a/core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
+++ b/core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
@@ -35,7 +35,7 @@ object TransactionCoordinator {
   def apply(config: KafkaConfig,
             replicaManager: ReplicaManager,
             scheduler: Scheduler,
-            producerIdGenerator: ProducerIdGenerator,
+            producerIdGeneratorFactory: () => ProducerIdGenerator,
             zkClient: KafkaZkClient,
             metrics: Metrics,
             metadataCache: MetadataCache,
@@ -59,7 +59,7 @@ object TransactionCoordinator {
     val txnMarkerChannelManager = TransactionMarkerChannelManager(config, metrics, metadataCache, txnStateManager,
       time, logContext)
 
-    new TransactionCoordinator(config.brokerId, txnConfig, scheduler, producerIdGenerator, txnStateManager, txnMarkerChannelManager,
+    new TransactionCoordinator(config.brokerId, txnConfig, scheduler, producerIdGeneratorFactory, txnStateManager, txnMarkerChannelManager,
       time, logContext)
   }
 
@@ -83,7 +83,7 @@ object TransactionCoordinator {
 class TransactionCoordinator(brokerId: Int,
                              txnConfig: TransactionConfig,
                              scheduler: Scheduler,
-                             producerIdGenerator: ProducerIdGenerator,
+                             producerIdGeneratorFactory: () => ProducerIdGenerator,
                              txnManager: TransactionStateManager,
                              txnMarkerChannelManager: TransactionMarkerChannelManager,
                              time: Time,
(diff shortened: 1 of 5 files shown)


rondagostino:
Only test failure is a known flaky test: https://issues.apache.org/jira/browse/KAFKA-7940

cmccabe:
This one looks good to me...

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
```diff
@@ -615,7 +615,7 @@ class TransactionCoordinator(brokerId: Int,
     info("Shutting down.")
     isActive.set(false)
     scheduler.shutdown()
-    producerIdManager.shutdown()
+    producerIdGenerator.shutdown()
```
mumrah, line 620: Since we're externalizing this trait (`ProducerIdGenerator`), it seems odd that TransactionCoordinator takes it as a dependency, but then later assumes the responsibility of closing it. I think we might need to move the shutdown call to the same place we create the instance, just so the ownership is clear.

Alternatively (and maybe preferably?) we can keep the ProducerIdGenerator construction in the companion object and add a second `apply` method when we add the KIP-500 implementation.

mumrah: Either way, the new trait looks good

ijuma: Yeah, I agree that the shutdown of producer id generator should move to `KafkaServer` if it's where we create it. Alternatively, `TransactionCoordinator` could take a factory (i.e. () => ProducerIdGenerator) for creating a `ProducerIdGenerator` and keep lifecycle management in `TransactionCoordinator`.

ijuma: I think the factory with the `apply` method would work reasonably well.

rondagostino: It felt weird to create the instance in `apply()` and hand it off to `TransactionCoordinator`, so I passed the factory and had the coordinator generate the instance.

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
```diff
@@ -83,7 +83,7 @@ object TransactionCoordinator {
 class TransactionCoordinator(brokerId: Int,
                              txnConfig: TransactionConfig,
                              scheduler: Scheduler,
-                             producerIdManager: ProducerIdManager,
+                             producerIdGeneratorFactory: () => ProducerIdGenerator,
```
ijuma: Nit: since this is a function, I'd call it `createProduceIdGenerator` or a verb like that.

ijuma (approved):
LGTM, just one nit.

ijuma (approved):

ijuma:
JDK 8 build exited (JDK 11 and 15 passed), so restarted the PR builder.

ijuma:
Weird, red build again. Restarted it once again.

ijuma:
Two unrelated flaky failures:

> Build / JDK 15 / org.apache.kafka.connect.integration.RebalanceSourceConnectorsIntegrationTest.testMultipleWorkersRejoining
> Build / JDK 8 / org.apache.kafka.connect.integration.BlockingConnectorTest.testBlockInSinkTaskStart

PR #10011, opened 2021-02-01
MINOR: remove org.apache.kafka.streams.errors.BrokerNotFoundException

chia7712:
```BrokerNotFoundException``` is useless after 234ec8a8af76bfb7874dd99714a65089d6048953 was merged.


mjsax:
Even if we don't throw this exception any longer, it seems to be a public API change, and thus we should rather do a KIP an first deprecate the class to remove it only later in a major release?

\cc @bbejeck WDYT?

ijuma:
Yeah, we should not remove public classes without a deprecation cycle.

chia7712:
@mjsax @ijuma Thanks for your response. I will file a KIP to deprecate it.

chia7712:
superseded by https://issues.apache.org/jira/browse/KAFKA-12281

PR #10013, opened 2021-02-01
MINOR: Remove ZK dependency for __transaction_state partition count

rondagostino:
`TransactionStateManager` currently uses ZooKeeper to obtain any non-default value for the `__transaction_state` partition count.  ZooKeeper will not be available when the broker uses a Raft-based metadata quorum.  This PR removes the hard dependency on ZooKeeper by allowing a function returning an `Int` to be provided instead.  Providing a ZooKeeper client is still supported in `apply()` methods, but those methods transparently wrap the client with a function and pass that function instead.

### DIFF
--- a/core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
+++ b/core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
@@ -30,6 +30,13 @@ import org.apache.kafka.common.record.RecordBatch
 import org.apache.kafka.common.requests.TransactionResult
 import org.apache.kafka.common.utils.{LogContext, ProducerIdAndEpoch, Time}
 
+class TransactionStateTopicPartitionCountViaZooKeeper(zkClient: KafkaZkClient,
+                                                      defaultTransactionTopicPartitionCount: Int) {
+  def transactionStateTopicPartitionCount(): Int = {
+    zkClient.getTopicPartitionCount(Topic.TRANSACTION_STATE_TOPIC_NAME).getOrElse(defaultTransactionTopicPartitionCount)
+  }
+}
+
 object TransactionCoordinator {
 
   def apply(config: KafkaConfig,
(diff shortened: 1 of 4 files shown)


rondagostino:
The combination of this PR and https://github.com/apache/kafka/pull/10009/ will remove hard ZooKeeper dependencies from the `TransactionCoordinator` and `TransactionStateManager` classes.

core/src/main/scala/kafka/coordinator/transaction/TransactionCoordinator.scala
```diff
@@ -30,6 +30,13 @@ import org.apache.kafka.common.record.RecordBatch
 import org.apache.kafka.common.requests.TransactionResult
 import org.apache.kafka.common.utils.{LogContext, ProducerIdAndEpoch, Time}
 
+class TransactionStateTopicPartitionCountViaZooKeeper(zkClient: KafkaZkClient,
```
ijuma, line 33: Can we not simply have a method in the companion object that we can lift into a function?

rondagostino: The logic to create a function `() => zkClient.getTopicPartitionCount(Topic.TRANSACTION_STATE_TOPIC_NAME).getOrElse(defaultValue)` is needed in 2 separate places (once in `TransactionCoordinator` and once in `TransactionStateManager`), so based on that I refactored it out into a common place here.  I can remove this class if you prefer.

ijuma: Yeah, I am suggesting you can still have it shared, but as a method in the companion object of one of the classes. And then you can have a simpler method name like `transactionStatePartitionCount` or something.

ijuma: To clarify, it's a bit of an anti pattern to have a whole class when a method would do in Scala.

cmccabe: I left a comment in the other PR.  We do not want to get rid of startup, or have a callback function.  See #10008.  I agree with @ijuma that we should consolidate these into one PR to keep the discussion in one place.

rondagostino: Closing this PR -- I consolidated these changes into https://github.com/apache/kafka/pull/10008 so we can review them together.

rondagostino:
Closing this PR -- I consolidated these changes into https://github.com/apache/kafka/pull/10008 so we can review them together.

PR #10014, opened 2021-02-01
KAFKA-12252 and KAFKA-12262: Fix session key rotation when leadership changes

C0urante:
Addressed [KAFKA-12252](https://issues.apache.org/jira/browse/KAFKA-12252) and [KAFKA-12262](https://issues.apache.org/jira/browse/KAFKA-12262).

- To address `KAFKA-12262`, all workers now track key expiration instead of just the leader.
- To address `KAFKA-12252`, a conditional check is added in the herder tick thread to only cut rebalance polling short for key rotation if the worker is the leader.

Unit tests for the distributed herder are added for both bugs. The first ensures that if a worker becomes the leader after reading a session key from the config topic, it wakes up from polling for rebalance in time for key rotation. The second ensures that if a worker loses leadership, it stops taking key expiration into account when calculating the time to poll for rebalance.

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
@@ -402,7 +402,7 @@ public class DistributedHerder extends AbstractHerder implements Runnable {
             log.debug("Scheduled rebalance at: {} (now: {} nextRequestTimeoutMs: {}) ",
                     scheduledRebalance, now, nextRequestTimeoutMs);
         }
-        if (internalRequestValidationEnabled() && keyExpiration < Long.MAX_VALUE) {
+        if (isLeader() && internalRequestValidationEnabled() && keyExpiration < Long.MAX_VALUE) {
             nextRequestTimeoutMs = Math.min(nextRequestTimeoutMs, Math.max(keyExpiration - now, 0));
             log.debug("Scheduled next key rotation at: {} (now: {} nextRequestTimeoutMs: {}) ",
                     keyExpiration, now, nextRequestTimeoutMs);
@@ -1573,10 +1573,11 @@ public class DistributedHerder extends AbstractHerder implements Runnable {
 
             synchronized (DistributedHerder.this) {
                 DistributedHerder.this.sessionKey = sessionKey.key();
-                // Track the expiration of the key if and only if this worker is the leader
+                // Track the expiration of the key.
                 // Followers will receive rotated keys from the leader and won't be responsible for
-                // tracking expiration and distributing new keys themselves
-                if (isLeader() && keyRotationIntervalMs > 0) {
+                // tracking expiration and distributing new keys themselves, but may become leaders
+                // later on and will need to know when to update the key.
+                if (keyRotationIntervalMs > 0) {
                     DistributedHerder.this.keyExpiration = sessionKey.creationTimestamp() + keyRotationIntervalMs;
                 }
             }
--- a/connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
+++ b/connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
@@ -28,6 +28,7 @@ import org.apache.kafka.connect.runtime.ConnectMetrics.MetricGroup;
 import org.apache.kafka.connect.runtime.ConnectorConfig;
 import org.apache.kafka.connect.runtime.Herder;
 import org.apache.kafka.connect.runtime.MockConnectMetrics;
+import org.apache.kafka.connect.runtime.SessionKey;
 import org.apache.kafka.connect.runtime.SinkConnectorConfig;
 import org.apache.kafka.connect.runtime.TargetState;
 import org.apache.kafka.connect.runtime.TaskConfig;
@@ -67,6 +68,7 @@ import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
 import org.powermock.reflect.Whitebox;
 
+import javax.crypto.SecretKey;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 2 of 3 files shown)


C0urante:
@gharris1727 @ncliang would either of you like to take a look?

connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
```diff
@@ -2161,6 +2163,83 @@ public Boolean answer() throws Throwable {
...
+
+        PowerMock.replayAll();
+
+        herder.tick();
+        configUpdateListener.onSessionKeyUpdate(initialKey);
+        herder.tick();
+        herder.tick();
+
+        assertTrue(pollTimeout.getValue() <= DistributedConfig.INTER_WORKER_KEY_TTL_MS_MS_DEFAULT);
+    }
```
gharris1727, line 2119: nit: Do these tests need a `PowerMock.verifyAll()` at the end? I see a `PowerMock.replayAll()` call but not any `PowerMock.expectLastCall()` or anything that makes it seem like powermock is being used functionally here.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
```diff
@@ -381,7 +381,7 @@ public void tick() {
             log.debug("Scheduled rebalance at: {} (now: {} nextRequestTimeoutMs: {}) ",
                     scheduledRebalance, now, nextRequestTimeoutMs);
         }
-        if (internalRequestValidationEnabled() && keyExpiration < Long.MAX_VALUE) {
+        if (isLeader() && internalRequestValidationEnabled() && keyExpiration < Long.MAX_VALUE) {
```
gharris1727, line 405: I don't think the `keyExpiration < MAX_VALUE` condition is strictly necessary any more, since if we're the leader, we must either have a valid key by this point. It made sense before, as it was basically an implcit `isLeader()` check that explicitly happened much earlier during the callback. Now, we either had a key read before, or added one at the beginning of this loop (and read to the end of the log, executing all callbacks including the one which updates `keyExpiration`). 
Neither the `isLeader()` or `internalRequestValidationEnabled()` conditions should change in between the `checkForKeyRotation(now)` call and here, so I don't think there's danger of hitting this condition when the `checkForKeyRotation(now)` didn't ensure that we have a key.

If `putSessionKey()` ever changed implementation to not read back the key and ensure that callbacks were executed prior to returning, then this condition would guard in that case, and prevent us from scheduling a `nextRequestTimeoutMs` 50y earlier than the otherwise very reasonable 292 million years in the future. We would "forget" to schedule a key expiration because of a race condition while installing a new key, and only schedule a rebalance the next time a rebalance occurred due to some other factor.

Seems harmless to remove, and only changes behavior for a case where we introduce a (worse) concurrency bug. Removing it will keep someone from wondering "why is that condition there" at some point in the future though :)

C0urante: There's a rationale for leaving the check in here but I'm not sold either way. Let me know what you think:

The `Scheduled next key rotation at...` log message should ideally only be emitted if key rotation is actually enabled. Key rotation can be [disabled by setting the ttl to 0](https://github.com/apache/kafka/blob/821d867c1bd01b88aa695827cb57c9aa764857a4/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L1579). If someone has key rotation disabled, emitting this log message doesn't make a lot of sense (even if the scheduled rotation time is millions of years in the future).

This also mirrors the structure of the scheduled rebalance delay check above, although that check is more necessary because of the tweaking of the `rebalanceResolved` field.

connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
C0urante: There are a bunch of expectations set up with the `expectRebalance` and `expectPostRebalance` methods; I've added the `PowerMock::verifyAll` call as a result. Good catch, thanks!

C0urante:
Thanks @gharris1727; I've tried to address your comments. Let me know what you think.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
gharris1727: Got it, duh. should have checked the other condition before leaving an uninformed comment 🙃

gharris1727 (approved):
LGTM, thanks @C0urante!

C0urante:
Ah whoops, the `PowerMock::verifyAll` broke the unit tests. Patched them up; should be good to go now.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
```diff
@@ -402,7 +402,7 @@ public void tick() {
...
```
rhauch, line 405: IIUC, `keyExpiration` is set to `Long.MAX_VALUE` above (line 365) when the key is to be rotated, but then after that is done this herder is the leader and its `keyExpiration` field is modified only through its own `ConfigUpdateListener`. But it's also possible that when all requests to be processed on lines 380-397 take less time than the config store requires to consume its own session key record, then this block is not entered until the next `tick()` call? If so, then isn't it possible that the tick blocks longer than the time until the next key expiration (assuming there are no further config updates or rebalance-inducing conditions)?

connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
```diff
@@ -2071,6 +2073,84 @@ public void testPutConnectorConfig() throws Exception {
...
+        member.poll(Long.MAX_VALUE);
+        EasyMock.expectLastCall();
+
+        expectRebalance(2, Collections.emptyList(), Collections.emptyList(), "member", MEMBER_URL);
+        Capture<SessionKey> updatedKey = EasyMock.newCapture();
+        configBackingStore.putSessionKey(EasyMock.capture(updatedKey));
+        EasyMock.expectLastCall().andAnswer(() -> {
+            configUpdateListener.onSessionKeyUpdate(updatedKey.getValue());
+            return null;
+        });
```
rhauch, line 2103: This is the condition that in the test (and probably the majority of real-world cases) always happens quickly and immediately, but which at least theoretically could take a while if the config backing store's consumer is somehow blocked. If that were the case, then the leader's `member.poll(...)` invocation blocks indefinitely. If there are no rebalance-inducing events, config changes, or even herder requests, then isn't is possible that even if the config back store calls the listener regarding the session key update _after_ the `member.poll(...)` is called, leading to the session key expiring before the herder's `tick()` method returns?

rhauch (commented):
Thanks, @C0urante. I have one question about a possible edge case below that this change may not actually affect, but one which seems like a potential issue nonetheless.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
C0urante: Good question--this shouldn't be possible since `KafkaConfigBackingStore::putSessionKey` does a flush and read-to-end after dispatching the new session key to the producer:

https://github.com/apache/kafka/blob/0e8a84e5d7203765240894873fd2896906dfbc2b/connect/runtime/src/main/java/org/apache/kafka/connect/storage/KafkaConfigBackingStore.java#L456-L471

connect/runtime/src/test/java/org/apache/kafka/connect/runtime/distributed/DistributedHerderTest.java
C0urante: I believe I've addressed this in my other comment; LMK what you think.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
rhauch: Actually, the `configLog.readToEnd().get(30000, TimeUnit.MILLISECONDS)` effective call only blocks up to 30 seconds for the consumer to read to the end. So it is *possible* for the `putSessionKey(...)` to return without the consumer having caught up and the session key notification method called. One example scenario is a network event after the producer successfully sends the record, but before the consumer is able to read to the end of the log, at which point the consumer continues retrying with effectively no timeout.

It still seems like it's possible in worker with no effective changes for it to block much longer than the key expiration. Obviously, if anything does cause the leader to wake up, it should be recover. But my question is what happens to one of the followers (which maybe isn't partitioned from the Kafka cluster) when they discover the session key is expired? Do they do nothing until they are called (again, assuming no change in membership and no config changes)?

C0urante: If the read to end doesn't complete after 30 seconds, the call doesn't return silently, but throws an exception. This in itself may not be great (we could catch exceptions caused by the call to `KafkaConfigBackingStore::putSessionKey` instead of allowing them to kill the entire worker), but it's still impossible for an accidental call to `WorkerGroupMember::poll` with `Long.MAX_VALUE` to happen under those circumstances.

C0urante:
@rhauch ping 🙂 
This has been waiting for a while and the only concern that's been raised is unrelated to the fix at hand. Can you take another look?

C0urante:
I've also filed https://issues.apache.org/jira/browse/KAFKA-12474 and https://issues.apache.org/jira/browse/KAFKA-12476 as follow-up items, since they are unaffected by and unrelated to the bug being addressed here.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java
rhauch: Sounds good. Thanks for logging [KAFKA-12474](https://issues.apache.org/jira/browse/KAFKA-12474) and [KAFKA-12476](https://issues.apache.org/jira/browse/KAFKA-12476).

rhauch (approved):
Thanks, @C0urante. LGTM.

FYI, we're still blocked on the `2.6` and `2.8` branches. I'll wait a few more days to merge since hopefully the former is unblocked soon, but after I merge this to `trunk`, `2.7` and `2.6` I will probably create a PR for `2.8`.

PR #10015, opened 2021-02-01
MINOR: Factor out controller node provider `BrokerToControllerChannelManager`

hachikuji:
This patch factors out a trait to allow for other ways to provide the controller `Node` object to `BrokerToControllerChannelManager`. In KIP-500, the controller will be provided from the Raft client and not the metadata cache.


mumrah (approved):
LGTM. Always nice to pull out tightly coupled things like this (BTCCM and MetadataCache). I think you could probably factor out MetadataCache from AlterIsrManager since it's only needed for getting the controller.

PR #10017, opened 2021-02-01
KAFKA-12260: Avoid hitting NPE for partitionsFor

abbccdda:
As mentioned in the ticket, returning `null` for `consumer#partitionsFor` is not a good client side agreement. Addressing this problem by returning the empty list instead.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
@@ -1924,7 +1924,7 @@ public class KafkaConsumer<K, V> implements Consumer<K, V> {
      * @param topic The topic to get partition metadata for
      * @param timeout The maximum of time to await topic metadata
      *
-     * @return The list of partitions
+     * @return The list of partitions, which will be empty when the given topic is not found
      * @throws org.apache.kafka.common.errors.WakeupException if {@link #wakeup()} is called before or while this
      *             function is called
      * @throws org.apache.kafka.common.errors.InterruptException if the calling thread is interrupted before or while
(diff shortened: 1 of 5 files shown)


clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
```diff
@@ -1948,7 +1948,7 @@ public OffsetAndMetadata committed(TopicPartition partition, final Duration time
             Timer timer = time.timer(timeout);
             Map<String, List<PartitionInfo>> topicMetadata = fetcher.getTopicMetadata(
                     new MetadataRequest.Builder(Collections.singletonList(topic), metadata.allowAutoTopicCreation()), timer);
-            return topicMetadata.get(topic);
+            return topicMetadata.getOrDefault(topic, Collections.emptyList());
```
chia7712, line 1951: just curious.

Is it possible that an existent topic has no partitions? If so, returning empty list may confuse users who querying nonexistent topic. Maybe we should throw ```UnknownTopicOrPartitionException``` for nonexistent topic and return empty list for existent topic.

chia7712 (approved):
+1 to eliminate null from Public APIs.

one (unrelated) question is left.

clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
chia7712: BTW, it is worth writing docs for this case (return empty list).

abbccdda: I don't think you could have a topic with 0 partition.

chia7712:
@abbccdda Thanks for updating code. Two minor comments are left.

1. The docs of ```KafkaConsumer#partitionsFor(String topic, Duration timeout)``` is not updated. 
1. ```MockConsumer``` still return null when `topic` is nonexistent. Should we make it return empty collection as well?

abbccdda:
@chia7712 Sounds good

PR #10018, opened 2021-02-01
Upstream MetadataImage and related classes

mumrah:
Mostly a wholesale copy of the files from confluentinc/kip-500, with a few fixes for the test classes

### DIFF
--- a/core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
+++ b/core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
@@ -1,0 +1,143 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kafka.server.metadata
+
+import java.util
+import java.util.Collections
+import java.util.concurrent.ThreadLocalRandom
+
+import kafka.cluster.BrokerEndPoint
+import kafka.common.BrokerEndPointNotAvailableException
+import org.apache.kafka.common.Node
+import org.apache.kafka.common.metadata.RegisterBrokerRecord
+import org.apache.kafka.common.network.ListenerName
+import org.slf4j.Logger
+
+import scala.jdk.CollectionConverters._
+
+object MetadataBroker {
+  def apply(record: RegisterBrokerRecord): MetadataBroker = {
+    new MetadataBroker(record.brokerId(), record.rack(),
+      record.endPoints().asScala.map {
+        case e => e.name() ->
+          new Node(record.brokerId(), e.host(), e.port(), record.rack())
+      }.toMap,
+      true)
+  }
+}
+
+case class MetadataBroker(id: Int,
+                          rack: String,
+                          endpoints: collection.Map[String, Node],
+                          fenced: Boolean) {
+  def brokerEndPoint(listenerName: ListenerName): BrokerEndPoint = {
+    endpoints.get(listenerName.value()) match {
+      case None => throw new BrokerEndPointNotAvailableException(
+        s"End point with listener name ${listenerName.value} not found for broker $id")
+      case Some(node) => new BrokerEndPoint(node.id(), node.host(), node.port())
+    }
+  }
+}
+
+class MetadataBrokersBuilder(log: Logger, prevBrokers: MetadataBrokers) {
+  private var newBrokerMap = prevBrokers.cloneBrokerMap()
+
+  def add(broker: MetadataBroker): Unit = {
+    newBrokerMap.put(broker.id, broker)
+  }
+
+  def changeFencing(id: Int, fenced: Boolean): Unit = {
+    val broker = newBrokerMap.get(id)
+    if (broker == null) {
+      throw new RuntimeException(s"Unknown broker id ${id}")
+    }
+    val newBroker = new MetadataBroker(broker.id, broker.rack, broker.endpoints, fenced)
+    newBrokerMap.put(id, newBroker)
+  }
+
+  def remove(id: Int): Unit = {
+    newBrokerMap.remove(id)
+  }
+
+  def get(brokerId: Int): Option[MetadataBroker] = Option(newBrokerMap.get(brokerId))
+
+  def build(): MetadataBrokers = {
+    val result = MetadataBrokers(log, newBrokerMap)
+    newBrokerMap = Collections.unmodifiableMap(newBrokerMap)
+    result
+  }
+}
+
+object MetadataBrokers {
+  def apply(log: Logger,
+            brokerMap: util.Map[Integer, MetadataBroker]): MetadataBrokers = {
+    var listenersIdenticalAcrossBrokers = true
+    var prevListeners: collection.Set[String] = null
+    val _aliveBrokers = new util.ArrayList[MetadataBroker](brokerMap.size())
+    brokerMap.values().iterator().asScala.foreach {
+      case broker => if (!broker.fenced) {
+        if (prevListeners == null) {
+          prevListeners = broker.endpoints.keySet
+        } else if (!prevListeners.equals(broker.endpoints.keySet)) {
+          listenersIdenticalAcrossBrokers = false
+        }
+        _aliveBrokers.add(broker)
+      }
+    }
+    if (!listenersIdenticalAcrossBrokers) {
+      log.error("Listeners are not identical across alive brokers. " +
+        _aliveBrokers.asScala.map(
+          broker => s"${broker.id}: ${broker.endpoints.keySet.mkString(", ")}"))
+    }
+    new MetadataBrokers(_aliveBrokers, brokerMap)
+  }
+}
+
+case class MetadataBrokers(private val _aliveBrokers: util.List[MetadataBroker],
+                           private val brokerMap: util.Map[Integer, MetadataBroker]) {
+  def size(): Int = brokerMap.size()
+
+  def iterator(): Iterator[MetadataBroker] = brokerMap.values().iterator().asScala
+
+  def cloneBrokerMap(): util.Map[Integer, MetadataBroker] = {
+    val result = new util.HashMap[Integer, MetadataBroker]
+    result.putAll(brokerMap)
+    result
+  }
+
+  def getAlive(id: Int): Option[MetadataBroker] = {
+    val broker = get(id)
+    if (broker.isDefined && !broker.get.fenced) {
+      broker
+    } else {
+      None
+    }
+  }
+
+  def randomAliveBrokerId(): Option[Int] = {
+    if (_aliveBrokers.isEmpty()) {
+      None
+    } else {
+      Some(_aliveBrokers.get(ThreadLocalRandom.current().nextInt(_aliveBrokers.size())).id)
+    }
+  }
+
+  def aliveBrokers(): util.List[MetadataBroker] = _aliveBrokers
+
+  def get(id: Int): Option[MetadataBroker] = Option(brokerMap.get(id))
+}
--- a/core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
+++ b/core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
@@ -32,10 +32,11 @@ import scala.jdk.CollectionConverters._
 
 object MetadataBroker {
   def apply(record: RegisterBrokerRecord): MetadataBroker = {
-    new MetadataBroker(record.brokerId(), record.rack(),
+    new MetadataBroker(record.brokerId, record.rack,
       record.endPoints().asScala.map {
-        case e => e.name() ->
-          new Node(record.brokerId(), e.host(), e.port(), record.rack())
+        endPoint =>
+          endPoint.name() ->
+            new Node(record.brokerId, endPoint.host, endPoint.port, record.rack)
       }.toMap,
       true)
   }
@@ -46,10 +47,10 @@ case class MetadataBroker(id: Int,
                           endpoints: collection.Map[String, Node],
                           fenced: Boolean) {
   def brokerEndPoint(listenerName: ListenerName): BrokerEndPoint = {
-    endpoints.get(listenerName.value()) match {
+    endpoints.get(listenerName.value) match {
       case None => throw new BrokerEndPointNotAvailableException(
         s"End point with listener name ${listenerName.value} not found for broker $id")
-      case Some(node) => new BrokerEndPoint(node.id(), node.host(), node.port())
+      case Some(node) => new BrokerEndPoint(node.id, node.host, node.port)
     }
   }
 }
@@ -90,14 +91,15 @@ object MetadataBrokers {
     var prevListeners: collection.Set[String] = null
     val _aliveBrokers = new util.ArrayList[MetadataBroker](brokerMap.size())
     brokerMap.values().iterator().asScala.foreach {
-      case broker => if (!broker.fenced) {
-        if (prevListeners == null) {
-          prevListeners = broker.endpoints.keySet
-        } else if (!prevListeners.equals(broker.endpoints.keySet)) {
-          listenersIdenticalAcrossBrokers = false
+      broker =>
+        if (!broker.fenced) {
+          if (prevListeners == null) {
+            prevListeners = broker.endpoints.keySet
+          } else if (!prevListeners.equals(broker.endpoints.keySet)) {
+            listenersIdenticalAcrossBrokers = false
+          }
+          _aliveBrokers.add(broker)
         }
-        _aliveBrokers.add(broker)
-      }
     }
     if (!listenersIdenticalAcrossBrokers) {
       log.error("Listeners are not identical across alive brokers. " +
@@ -108,7 +110,7 @@ object MetadataBrokers {
   }
 }
 
-case class MetadataBrokers(private val _aliveBrokers: util.List[MetadataBroker],
+case class MetadataBrokers(private val aliveBrokersList: util.List[MetadataBroker],
                            private val brokerMap: util.Map[Integer, MetadataBroker]) {
   def size(): Int = brokerMap.size()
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
```diff
@@ -0,0 +1,143 @@
...
+}
+
+object MetadataBrokers {
+  def apply(log: Logger,
+            brokerMap: util.Map[Integer, MetadataBroker]): MetadataBrokers = {
+    var listenersIdenticalAcrossBrokers = true
+    var prevListeners: collection.Set[String] = null
+    val _aliveBrokers = new util.ArrayList[MetadataBroker](brokerMap.size())
+    brokerMap.values().iterator().asScala.foreach {
+      case broker => if (!broker.fenced) {
```
hachikuji: nit: we don't need `case` here. Also, we usually put this on the previous line:
```scala
...foreach { broker =>
```
There are a few of these in the PR.

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
```diff
@@ -0,0 +1,143 @@
...
+import org.slf4j.Logger
+
+import scala.jdk.CollectionConverters._
+
+object MetadataBroker {
+  def apply(record: RegisterBrokerRecord): MetadataBroker = {
+    new MetadataBroker(record.brokerId(), record.rack(),
+      record.endPoints().asScala.map {
+        case e => e.name() ->
+          new Node(record.brokerId(), e.host(), e.port(), record.rack())
```
hachikuji: nit: a bunch of cases in here we could drop the unneeded parenthesis

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.kafka.common.network.ListenerName
+import org.slf4j.Logger
+
+import scala.jdk.CollectionConverters._
+
+object MetadataBroker {
+  def apply(record: RegisterBrokerRecord): MetadataBroker = {
+    new MetadataBroker(record.brokerId(), record.rack(),
+      record.endPoints().asScala.map {
+        case e => e.name() ->
```
hachikuji: nit: unneeded `case`

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
```diff
@@ -0,0 +1,143 @@
...
+    result
+  }
+}
+
+object MetadataBrokers {
+  def apply(log: Logger,
+            brokerMap: util.Map[Integer, MetadataBroker]): MetadataBrokers = {
+    var listenersIdenticalAcrossBrokers = true
+    var prevListeners: collection.Set[String] = null
+    val _aliveBrokers = new util.ArrayList[MetadataBroker](brokerMap.size())
```
hachikuji, line 91: nit: any reason this has the underscore prefix?

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  def iterator(): Iterator[MetadataBroker] = brokerMap.values().iterator().asScala
+
+  def cloneBrokerMap(): util.Map[Integer, MetadataBroker] = {
+    val result = new util.HashMap[Integer, MetadataBroker]
+    result.putAll(brokerMap)
+    result
+  }
+
+  def getAlive(id: Int): Option[MetadataBroker] = {
```
hachikuji: nit: `getAlive` -> `aliveBroker`? Seems more consistent with `aliveBrokers`.

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
```diff
@@ -0,0 +1,143 @@
...
+    val broker = get(id)
```
hachikuji: Maybe rewrite this as a match?

core/src/main/scala/kafka/server/metadata/MetadataImage.scala
```diff
@@ -0,0 +1,122 @@
...
+  def partitionsBuilder(): MetadataPartitionsBuilder = {
+    if (_partitionsBuilder == null) {
+      _partitionsBuilder = new MetadataPartitionsBuilder(brokerId, prevImage.partitions)
+    }
+    _partitionsBuilder
+  }
+
+  def hasPartitionChanges(): Boolean = _partitionsBuilder != null
+
+  def topicIdToName(uuid: Uuid): Option[String] = {
```
hachikuji: nit: `uuid` -> `topicId`

core/src/main/scala/kafka/server/metadata/MetadataImage.scala
```diff
@@ -0,0 +1,122 @@
...
+      None,
+      new MetadataBrokers(Collections.emptyList(), new util.HashMap[Integer, MetadataBroker]()))
+  }
+
+  def contains(partition: TopicPartition): Boolean =
+    partitions.get(partition.topic(), partition.partition()).isDefined
+
+  def contains(topic: String): Boolean = partitions.topicPartitions(topic).hasNext
+
+  def getAliveBroker(id: Int): Option[MetadataBroker] = brokers.getAlive(id)
```
hachikuji: nit: drop `get`

core/src/main/scala/kafka/server/metadata/MetadataImage.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  def numAliveBrokers(): Int = brokers.aliveBrokers().size
+
+  def controller(): Option[MetadataBroker] = controllerId.flatMap(id => brokers.getAlive(id))
+
+  def topicIdToName(uuid: Uuid): Option[String] = {
```
hachikuji, line 118: nit: `uuid` -> `topicId`

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+
+object MetadataPartition {
+  def apply(name: String, record: PartitionRecord): MetadataPartition = {
+    MetadataPartition(name,
+      record.partitionId(),
+      record.leader(),
+      record.leaderEpoch(),
+      record.replicas(),
+      record.isr(),
+      Collections.emptyList(), // TODO: handle offline replicas
```
hachikuji: Can we turn the TODO into a jira?

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+  private var newIdMap = prevPartitions.copyIdMap()
+  private val changed = new util.IdentityHashMap[Any, Boolean]()
+  private val _localChanged = new util.HashSet[MetadataPartition]
+  private val _localRemoved = new util.HashSet[MetadataPartition]
+
+  def topicIdToName(id: Uuid): Option[String] = Option(newIdMap.get(id))
+
+  def removeTopicById(id: Uuid): Iterable[MetadataPartition] = {
+    Option(newIdMap.remove(id)) match {
+      case None => throw new RuntimeException(s"Unable to locate topic with ID ${id}")
```
hachikuji: nit: `${id}` -> `$id` (there are a few of these in here)

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+      case Some(name) => newNameMap.remove(name).values().asScala
+    }
+  }
+
+  def handleIsrChange(record: IsrChangeRecord): Unit = {
+    Option(newIdMap.get(record.topicId())) match {
```
hachikuji, line 117: Some helpers might make this less awkward. Maybe a shortcut to get to the partition map?

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+
+  def addUuidMapping(name: String, id: Uuid): Unit = {
+    newIdMap.put(id, name)
+  }
+
+  def removeUuidMapping(id: Uuid): Unit = {
+    newIdMap.remove(id)
+  }
+
+  def get(topicName: String, partitionId: Int): Option[MetadataPartition] = {
```
hachikuji, line 137: nit: I think a more explicit name would help. This naming assumes that it is clear enough in the context of `MetadataPartitions`, but it can get a little confusing when you get a few methods chained together.

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+      removingReplicas)
+  }
+}
+
+class MetadataPartitionsBuilder(val brokerId: Int,
+                                val prevPartitions: MetadataPartitions) {
+  private var newNameMap = prevPartitions.copyNameMap()
+  private var newIdMap = prevPartitions.copyIdMap()
+  private val changed = new util.IdentityHashMap[Any, Boolean]()
+  private val _localChanged = new util.HashSet[MetadataPartition]
```
hachikuji, line 104: nit: the naming convention is not very clear to me. Not sure why these two collections get an underscore, but `changed` does not.

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+
+  def removeUuidMapping(id: Uuid): Unit = {
+    newIdMap.remove(id)
+  }
+
+  def get(topicName: String, partitionId: Int): Option[MetadataPartition] = {
+    Option(newNameMap.get(topicName)).flatMap(m => Option(m.get(partitionId)))
+  }
+
+  def set(partition: MetadataPartition): Unit = {
```
hachikuji, line 141: We can probably leave this as a future improvement, but I'm a little concerned that we don't have the topicId here. I am guessing that the caller has other state assertions which ensure that partition state is updated consistently with respect to the topic id map, but it might also be nice if we could be more defensive about protecting internal state.

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+      addingReplicas,
+      removingReplicas)
+  }
+}
+
+class MetadataPartitionsBuilder(val brokerId: Int,
+                                val prevPartitions: MetadataPartitions) {
+  private var newNameMap = prevPartitions.copyNameMap()
+  private var newIdMap = prevPartitions.copyIdMap()
+  private val changed = new util.IdentityHashMap[Any, Boolean]()
```
hachikuji: I think this is a map just because there is no `IdentityHashSet`. Perhaps we could use `Collections.newSetFromMap`?

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+        if (prevPartition.isReplicaFor(brokerId)) {
+          _localRemoved.add(prevPartition)
+        }
+      } else {
+        Option(prevPartitionMap.get(partitionId)).foreach { prevPartition =>
+          if (prevPartition.isReplicaFor(brokerId)) {
+            _localRemoved.add(prevPartition)
+          }
+          val newPartitionMap = new util.HashMap[Int, MetadataPartition](prevPartitionMap.size() - 1)
+          prevPartitionMap.forEach(new BiConsumer[Int, MetadataPartition]() {
```
hachikuji: nit: can rewrite this as a lambda:
```scala
prevPartitionMap.forEach { (partitionId, metadataPartition =>
```

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+  }
+
+  def build(): MetadataPartitions = {
+    val result = new MetadataPartitions(newNameMap, newIdMap)
+    newNameMap = Collections.unmodifiableMap(newNameMap)
+    newIdMap = Collections.unmodifiableMap(newIdMap)
+    result
+  }
+
+  def localChanged(): Set[MetadataPartition] = _localChanged.asScala.toSet
```
hachikuji: Hmm.. I wonder if the copy here is necessary. If it is, perhaps we should emphasize it in the name. For example `copyLocalChanged`.

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+
+  def copyIdMap(): util.Map[Uuid, String] = {
+    val copy = new util.HashMap[Uuid, String](idMap.size())
+    copy.putAll(idMap)
+    copy
+  }
+
+  def allPartitions(): Iterator[MetadataPartition] = new AllPartitionsIterator(nameMap).asScala
+
+  def allTopicNames(): Set[String] = nameMap.keySet().asScala.toSet
```
hachikuji: Another potentially dangerous copy here. I'm assuming this is unintended since I think we are trying to make the image immutable. We could fix this by using `collection.Set` as the return type and getting rid of the call to `toSet`.

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -0,0 +1,282 @@
...
+    result
+  }
+
+  def localChanged(): Set[MetadataPartition] = _localChanged.asScala.toSet
+
+  def localRemoved(): Set[MetadataPartition] = _localRemoved.asScala.toSet
+}
+
+case class MetadataPartitions(private val nameMap: util.Map[String, util.Map[Int, MetadataPartition]],
+                              private val idMap: util.Map[Uuid, String]) {
```
hachikuji, line 203: Intuitively, this seems a little backwards. I would have expected to have the partitions keyed by topicId with a separate map from topic name to topic id. If we were going to support renaming for example, then we would just update the name to id map.

I suspect we are ultimately going to have to change this because it gives us no good way to lookup the UUID corresponding to a given topic name. We will need to be able to do this to handle `Metadata` queries (see https://github.com/apache/kafka/commits/trunk/core/src/main/scala/kafka/server/MetadataCache.scala).

core/src/test/scala/kafka/server/metadata/MetadataBrokersTest.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  @Test
+  def testChangeFencing(): Unit = {
+    val builder = new MetadataBrokersBuilder(log, emptyBrokers)
+    assertEquals(None, builder.get(0))
+    assertThrows(classOf[RuntimeException], () => builder.changeFencing(0, false))
+    builder.add(TestUtils.createMetadataBroker(0, fenced = true))
+    assertTrue(builder.get(0).get.fenced)
+    builder.changeFencing(0, false)
+    assertFalse(builder.get(0).get.fenced)
```
hachikuji, line 63: Should we verify the result after `build()` as well?

core/src/test/scala/kafka/server/metadata/MetadataPartitionsTest.scala
```diff
@@ -0,0 +1,156 @@
...
+  def testAllPartitionsIterator(): Unit = {
+    val builder = new MetadataPartitionsBuilder(0, emptyPartitions)
+    val expected = new mutable.HashSet[MetadataPartition]()
+    expected += newPartition("foo", 0)
+    expected += newPartition("foo", 1)
+    expected += newPartition("foo", 2)
+    expected += newPartition("bar", 0)
+    expected += newPartition("bar", 1)
+    expected += newPartition("baz", 0)
+    expected.foreach { builder.set(_) }
```
hachikuji: nit: simplify a little bit
```scala
    expected.foreach(builder.set)
```

core/src/test/scala/kafka/server/metadata/MetadataPartitionsTest.scala
```diff
@@ -0,0 +1,156 @@
...
+  val log = LoggerFactory.getLogger(classOf[MetadataPartitionsTest])
+
+  val emptyPartitions = new MetadataPartitions(Collections.emptyMap(), Collections.emptyMap())
+
+  private def newPartition(topicName: String,
+                           partitionIndex: Int,
+                           replicas: Option[Seq[Int]] = None,
+                           isr: Option[Seq[Int]] = None): MetadataPartition = {
+    val effectiveReplicas = replicas match {
+      case None => util.Arrays.asList(Integer.valueOf(partitionIndex),
```
hachikuji: nit: can probably clean this up a bit. For example:
```scala
    val effectiveReplicas = replicas
      .getOrElse(List(partitionIndex, partitionIndex + 1, partitionIndex + 2))
      .map(Int.box)
      .toList.asJava
```

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
mumrah: Looks like there's an `aliveBrokers` method defined below.

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
mumrah: How about simply `get(id).filter(!_.fenced)` ?

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
mumrah: Seems to be due to methods with the same name defined on the class

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.kafka.common.network.ListenerName
+import org.slf4j.Logger
+
+import scala.jdk.CollectionConverters._
+
+object MetadataBroker {
+  def apply(record: RegisterBrokerRecord): MetadataBroker = {
+    new MetadataBroker(record.brokerId, record.rack,
+      record.endPoints().asScala.map {
+        endPoint =>
```
hachikuji: nit: move to previous line

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
```diff
@@ -0,0 +1,140 @@
...
+}
+
+object MetadataBrokers {
+  def apply(log: Logger,
+            brokerMap: util.Map[Integer, MetadataBroker]): MetadataBrokers = {
+    var listenersIdenticalAcrossBrokers = true
+    var prevListeners: collection.Set[String] = null
+    val _aliveBrokers = new util.ArrayList[MetadataBroker](brokerMap.size())
+    brokerMap.values().iterator().asScala.foreach {
+      broker =>
```
hachikuji: nit: move to previous line

cmccabe:
@hachikuji : yes, the copies are unintentional.  we should should avoid those...

Also we should be able to access topics / partitions by id, but it requires an extra hop, as you saw.

hachikuji (approved):
Thanks for the updates. LGTM.

mumrah:
Manually merged to trunk to fixup authorship

PR #10019, opened 2021-02-01
MINOR: Introduce KafkaBroker trait for use in dynamic reconfiguration

rondagostino:
Dynamic broker reconfiguration needs to occur for both ZooKeeper-based brokers and brokers that use a Raft-based metadata quorum.  `DynamicBrokerConfig` currently operates on `KafkaServer`, but it needs to operate on `BrokerServer` (the broker implementation that will use the Raft metadata log) as well. This PR introduces a `KafkaBroker` trait to allow dynamic reconfiguration to work with either implementation.


core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
```diff
@@ -669,7 +669,7 @@ class DynamicLogConfig(logManager: LogManager, server: KafkaServer) extends Brok
     updateLogsConfig(newBrokerDefaults.asScala)
 
     if (logManager.currentDefaultConfig.uncleanLeaderElectionEnable && !origUncleanLeaderElectionEnable) {
-      server.kafkaController.enableDefaultUncleanLeaderElection()
+      server.zkBasedKafkaController.foreach {_.enableDefaultUncleanLeaderElection() }
```
jsancio: It is idiomatic to use `()` instead of `{}` when the lambda is a single expression. E.g.
```scala
       server.zkBasedKafkaController.foreach(_.enableDefaultUncleanLeaderElection())
```

core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
```diff
@@ -936,7 +936,7 @@ class DynamicListenerConfig(server: KafkaServer) extends BrokerReconfigurable wi
     if (listenersAdded.nonEmpty)
       server.socketServer.addListeners(listenersAdded)
 
-    server.kafkaController.updateBrokerInfo(server.createBrokerInfo)
+    server.zkBasedKafkaController.foreach {_.updateBrokerInfo(server.createBrokerInfo) }
```
jsancio: It is idiomatic to use `()` instead of `{}` when the lambda is a single expression. E.g.
```scala
       server.zkBasedKafkaController.foreach(_.enableDefaultUncleanLeaderElection())
```

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -0,0 +1,113 @@
...
+  def kafkaYammerMetrics: KafkaYammerMetrics
+  def logManager: LogManager
+  def metrics(): Metrics
+  def quotaManagers: QuotaFactory.QuotaManagers
+  def replicaManager: ReplicaManager
+  def socketServer: SocketServer
+
+  // must override the following in KafkaServer since they only apply there
+  def zkBasedKafkaController: Option[KafkaController] = None
+  def createBrokerInfo: BrokerInfo = throw new UnsupportedOperationException("Unsupported in with a Raft-based meadata quorum")
```
jsancio: Why is "Raft-based" mentioned in generic `KafkaBroker` trait? Maybe we can remove the default implementation and force the classes implementing this trait to override the method.

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -0,0 +1,113 @@
...
+  def kafkaScheduler: KafkaScheduler
+  def kafkaYammerMetrics: KafkaYammerMetrics
+  def logManager: LogManager
+  def metrics(): Metrics
+  def quotaManagers: QuotaFactory.QuotaManagers
+  def replicaManager: ReplicaManager
+  def socketServer: SocketServer
+
+  // must override the following in KafkaServer since they only apply there
+  def zkBasedKafkaController: Option[KafkaController] = None
```
jsancio: Ideally, we should not have a reference to ZooKeeper on a trait that is implemented by both the ZK quorum and Raft quorum. Maybe `kafkaController: Option[KafkaController]`.

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -0,0 +1,113 @@
...
+trait KafkaBroker extends Logging with KafkaMetricsGroup {
+  def authorizer: Option[Authorizer]
+  def clusterId(): String
+  def config: KafkaConfig
+  def brokerState: BrokerState
+  def dataPlaneRequestHandlerPool: KafkaRequestHandlerPool
+  def kafkaScheduler: KafkaScheduler
+  def kafkaYammerMetrics: KafkaYammerMetrics
+  def logManager: LogManager
+  def metrics(): Metrics
```
jsancio: The methods `clusterId` and `metrics` have parenthesis `()` while the rest of the methods do not. Should we just remove the parenthesis for these methods?

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -0,0 +1,113 @@
...
+                                             metricsReporters: Seq[AnyRef]): Unit = {
+    val metricsContext = createKafkaMetricsContext(clusterId, config)
+    metricsReporters.foreach {
+      case x: MetricsReporter => x.contextChange(metricsContext)
+      case _ => //do nothing
+    }
+  }
+}
+
+trait KafkaBroker extends Logging with KafkaMetricsGroup {
```
jsancio: Why does this trait extends `Logging` and `KafkaMetricsGroup`? These feel like an implementation detail.

This looks like a very important trait and concept. Should we add documentation to all of the public methods?

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -0,0 +1,113 @@
...
+  // must override the following in KafkaServer since they only apply there
+  def zkBasedKafkaController: Option[KafkaController] = None
+  def createBrokerInfo: BrokerInfo = throw new UnsupportedOperationException("Unsupported in with a Raft-based meadata quorum")
+
+
+  newKafkaServerGauge("BrokerState", () => brokerState.currentState)
+  newKafkaServerGauge("ClusterId", () => clusterId())
+  newKafkaServerGauge("yammer-metrics-count", () =>  KafkaYammerMetrics.defaultRegistry.allMetrics.size)
+
+  val linuxIoMetricsCollector = new LinuxIoMetricsCollector("/proc", Time.SYSTEM, logger.underlying)
```
jsancio: I think you can make this `private[this]`.

If you want to make sure that these fields don't get leaked to implementing classes maybe we can move this default metric initialization to a companion object (static in Java) method:

```scala
object KafkaBroker {
  def initializeDefaultMetrics(metricsGroup: KafkaMetricsGroup, kafkaServer: KafkaBroker): Unit
}
```

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -0,0 +1,113 @@
...
+    val metricsContext = new KafkaMetricsContext(metricsPrefix, contextLabels)
+    metricsContext
+  }
+
+  private[server] def notifyClusterListeners(clusterId: String,
+                                             clusterListeners: Seq[AnyRef]): Unit = {
+    val clusterResourceListeners = new ClusterResourceListeners
+    clusterResourceListeners.maybeAddAll(clusterListeners.asJava)
+    clusterResourceListeners.onUpdate(new ClusterResource(clusterId))
+  }
```
jsancio, line 60: This code is duplicated in `KafkaServer`.

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  private[server] def notifyMetricsReporters(clusterId: String,
+                                             config: KafkaConfig,
+                                             metricsReporters: Seq[AnyRef]): Unit = {
+    val metricsContext = createKafkaMetricsContext(clusterId, config)
+    metricsReporters.foreach {
+      case x: MetricsReporter => x.contextChange(metricsContext)
+      case _ => //do nothing
+    }
+  }
```
jsancio, line 70: This code is duplicated in `KafkaServer`.

jsancio (commented):
Thanks for the PR @rondagostino

core/src/main/scala/kafka/server/KafkaBroker.scala
ijuma: Is there a shared interface for the Controller though? It seems that maybe you want to extract the methods that are needed into the trait instead of this approach.

core/src/main/scala/kafka/server/KafkaBroker.scala
rondagostino: I'm going to remove these in favor or having the caller cast.  See comment below for details.

core/src/main/scala/kafka/server/KafkaBroker.scala
rondagostino: It's straightforward for the calling code to cast to a `KafkaServer` or `BrokerServer` since the way to deal with the two cases can be totally different.  I'l push a commit doing this.

core/src/main/scala/kafka/server/KafkaBroker.scala
rondagostino: Metrics need to exist for compatibility, of course, and while a pure interface sounds like a great idea, it feels practical to put the metrics here since they are shared by all types of brokers -- they're basically part of the interface we've agreed to maintain for the sake of compatibility.

I started to do the documentation as you suggested, but they're all pretty useless (`// the authorizer`, for example).

I was able to remove `Logging` since it is redundant -- `KafkaMetricsGroup` already has it.

core/src/main/scala/kafka/server/KafkaBroker.scala
rondagostino: Agree on access -- I'll make it `private`.  Moving everything else feels like a bit of overkill to me.

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  newKafkaServerGauge("BrokerState", () => brokerState.currentState)
+  newKafkaServerGauge("ClusterId", () => clusterId)
+  newKafkaServerGauge("yammer-metrics-count", () =>  KafkaYammerMetrics.defaultRegistry.allMetrics.size)
+
+  private val linuxIoMetricsCollector = new LinuxIoMetricsCollector("/proc", Time.SYSTEM, logger.underlying)
+
+  if (linuxIoMetricsCollector.usable()) {
+    newGauge("linux-disk-read-bytes", () => linuxIoMetricsCollector.readBytes())
+    newGauge("linux-disk-write-bytes", () => linuxIoMetricsCollector.writeBytes())
```
ijuma, line 101: Why don't you have to override the metric name here?

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -0,0 +1,106 @@
...
+  // For backwards compatibility, we need to keep older metrics tied
+  // to their original name when this class was named `KafkaServer`
+  private def newKafkaServerGauge[T](metricName: String, gauge: yammer.Gauge[T]): yammer.Gauge[T] = {
+    val explicitName = explicitMetricName(
+      group = "kafka.server",
+      typeName = "KafkaServer",
+      name = metricName,
+      tags = Map.empty
+    )
+    KafkaYammerMetrics.defaultRegistry().newGauge(explicitName, gauge)
```
ijuma: Check `KafkaZkClient` for an example of how this can be done more concisely.

core/src/main/scala/kafka/server/DynamicBrokerConfig.scala
```diff
@@ -936,7 +939,10 @@ class DynamicListenerConfig(server: KafkaServer) extends BrokerReconfigurable wi
     if (listenersAdded.nonEmpty)
       server.socketServer.addListeners(listenersAdded)
 
-    server.kafkaController.updateBrokerInfo(server.createBrokerInfo)
+    server match {
+      case kafkaServer: KafkaServer => kafkaServer.kafkaController.updateBrokerInfo(kafkaServer.createBrokerInfo)
+      case _ =>
```
cmccabe, line 944: Can you log an error message here?  This something we will need to implement (re-registration on listener change) but we don't have time right now.

core/src/main/scala/kafka/server/KafkaBroker.scala
cmccabe: Yes, I think the metric name needs to be overridden...

cmccabe:
I agree with the comments from earlier that we don't want to expose the pre-KIP-500 `KafkaController` in the `KafkaServer` interface.  I think casting to `KafkaServer` or `RaftBrokerServer` is a reasonable solution.  It fits with the general (anti?) pattern in `DynamicBrokerConfig` of reaching directly into broker fields to do stuff.  Arguably a lot of this code should be in the broker class itself for just that reason, so that these broker fields can be private. But, one step at a time.

We should disallow changing broker listener configurations if KIP-500 is enabled, until we implement re-registering the broker.  (This isn't a lot of work but we just haven't done it yet...)

cmccabe:
LGTM

cmccabe (approved):

PR #10020, opened 2021-02-01
KAFKA-10689: fix windowed FKJ topology and put checks in assignor to avoid infinite loops

mjsax:
Backport of #9568 to `2.6`

Fix infinite loop in assignor when trying to resolve the number of partitions in a topology with a windowed FKJ. Also adds a check to this loop to break out and fail the application if we detect that we are/will be stuck in an infinite loop.

### DIFF
--- a/streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignorTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignorTest.java
@@ -16,7 +16,6 @@
  */
 package org.apache.kafka.streams.processor.internals;
 
-import java.util.SortedSet;
 import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.ListOffsetsResult;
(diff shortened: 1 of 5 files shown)


streams/src/test/java/org/apache/kafka/streams/processor/internals/StreamsPartitionAssignorTest.java
```diff
@@ -1987,6 +1996,65 @@ public void testUniqueFieldOverflow() {
...
+        configureDefault();
+
+        subscriptions.put("consumer",
+                          new Subscription(
+                              singletonList("topic"),
+                              defaultSubscriptionInfo.encode()
+                          ));
+        final Map<String, Assignment> assignments = partitionAssignor.assign(metadata, new GroupSubscription(subscriptions)).groupAssignment();
+        assertThat(AssignmentInfo.decode(assignments.get("consumer").userData()).errCode(),
+                   equalTo(AssignorError.INCOMPLETE_SOURCE_TOPIC_METADATA.code()));
```
mjsax, line 2031: The change to the original PR is, tat we expect a different error code here. But I guess it's still ok.

ableegoldman: Yeah it's a little awkward/misleading, but ideally no one should be hitting this exception in the 1st place once we have this fix. So this seems fine for 2.6.2

ableegoldman (approved):
LGTM, thanks for backporting this. Feel free to merge when the build passes

PR #10021, opened 2021-02-02
KAFKA-12205: Delete snapshots less than the snapshot at the log start

dengziming:
3 times to delete a snapshot:
1. When a follower completing fetch snapshot and truncate the log to the latest snapshot
2. When a controller pollCurrentState and delete old segment
3. When restarting the RaftClient.

Unit test

### DIFF
--- a/core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
+++ b/core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
@@ -284,7 +284,7 @@ final class KafkaMetadataLog private (
    * Removes all snapshots on the log directory whose epoch and end offset is less than the giving epoch and end offset.
    */
   private def removeSnapshotFilesBefore(logStartSnapshotId: OffsetAndEpoch): Unit = {
-    val expiredSnapshotIdsIter = snapshotIds.headSet(logStartSnapshotId).iterator()
+    val expiredSnapshotIdsIter = snapshotIds.headSet(logStartSnapshotId, false).iterator()
     while (expiredSnapshotIdsIter.hasNext) {
       val snapshotId = expiredSnapshotIdsIter.next()
       // If snapshotIds contains a snapshot id, the KafkaRaftClient and Listener can expect that the snapshot exists
@@ -353,10 +353,14 @@ object KafkaMetadataLog {
     Files
       .walk(log.dir.toPath, 1)
       .map[Optional[SnapshotPath]] { path =>
-        if (path != log.dir.toPath) {
-          Snapshots.parse(path)
-        } else {
+
+        if (path == log.dir.toPath) {
           Optional.empty()
+        } else if (path.endsWith(Snapshots.DELETE_SUFFIX)) {
+          Files.deleteIfExists(path)
+          Optional.empty()
+        } else {
+          Snapshots.parse(path)
         }
       }
       .forEach { path =>
--- a/core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
+++ b/core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
@@ -19,9 +19,8 @@ package kafka.raft
 import java.io.{File, IOException}
 import java.nio.file.{Files, NoSuchFileException}
 import java.util.concurrent.ConcurrentSkipListSet
-import java.util.{NoSuchElementException, Optional, Properties}
+import java.util.{NoSuchElementException, Optional}
 
-import kafka.api.ApiVersion
 import kafka.log.{AppendOrigin, Log, LogConfig, LogOffsetSnapshot, SnapshotGenerated}
 import kafka.server.{BrokerTopicStats, FetchHighWatermark, FetchLogEnd, LogDirFailureChannel}
 import kafka.utils.{Logging, Scheduler}
@@ -35,6 +34,7 @@ import scala.compat.java8.OptionConverters._
 
 final class KafkaMetadataLog private (
   log: Log,
+  config: LogConfig,
   scheduler: Scheduler,
   // This object needs to be thread-safe because it is used by the snapshotting thread to notify the
   // polling thread when snapshots are created.
@@ -303,7 +303,7 @@ final class KafkaMetadataLog private (
       scheduler.schedule(
         "delete-snapshot-file",
         () => Snapshots.deleteSnapshotIfExists(log.dir.toPath, snapshotId),
-        60 * 1000L)
+        config.fileDeleteDelayMs)
     }
   }
 
@@ -316,18 +316,13 @@ object KafkaMetadataLog {
 
   def apply(
     topicPartition: TopicPartition,
+    defaultLogConfig: LogConfig,
     dataDir: File,
     time: Time,
     scheduler: Scheduler,
     maxBatchSizeInBytes: Int,
     maxFetchSizeInBytes: Int
   ): KafkaMetadataLog = {
-    val props = new Properties()
-    props.put(LogConfig.MaxMessageBytesProp, maxBatchSizeInBytes.toString)
-    props.put(LogConfig.MessageFormatVersionProp, ApiVersion.latestVersion.toString)
-
-    LogConfig.validateValues(props)
-    val defaultLogConfig = LogConfig(props)
 
     val log = Log(
       dir = dataDir,
@@ -345,6 +340,7 @@ object KafkaMetadataLog {
 
     val metadataLog = new KafkaMetadataLog(
       log,
+      defaultLogConfig,
       scheduler,
       recoverSnapshots(log),
       topicPartition,
@@ -367,20 +363,18 @@ object KafkaMetadataLog {
     Files
       .walk(log.dir.toPath, 1)
       .map[Optional[SnapshotPath]] { path =>
-
-        if (path == log.dir.toPath) {
-          Optional.empty()
-        } else if (path.endsWith(Snapshots.DELETE_SUFFIX)) {
-          Files.deleteIfExists(path)
-          Optional.empty()
-        } else {
+        if (path != log.dir.toPath) {
           Snapshots.parse(path)
+        } else {
+          Optional.empty()
         }
       }
       .forEach { path =>
         path.ifPresent { snapshotPath =>
-          if (snapshotPath.partial || snapshotPath.snapshotId.offset < log.logStartOffset) {
-            // Delete partial snapshot and older snapshot
+          if (snapshotPath.partial ||
+            snapshotPath.deleted ||
+            snapshotPath.snapshotId.offset < log.logStartOffset) {
+            // Delete partial snapshot, deleted snapshot and older snapshot
             Files.deleteIfExists(snapshotPath.path)
           } else {
             snapshotIds.add(snapshotPath.snapshotId)
--- a/core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
+++ b/core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
@@ -19,8 +19,9 @@ package kafka.raft
 import java.io.{File, IOException}
 import java.nio.file.{Files, NoSuchFileException}
 import java.util.concurrent.ConcurrentSkipListSet
-import java.util.{NoSuchElementException, Optional}
+import java.util.{NoSuchElementException, Optional, Properties}
 
+import kafka.api.ApiVersion
 import kafka.log.{AppendOrigin, Log, LogConfig, LogOffsetSnapshot, SnapshotGenerated}
 import kafka.server.{BrokerTopicStats, FetchHighWatermark, FetchLogEnd, LogDirFailureChannel}
 import kafka.utils.{Logging, Scheduler}
@@ -34,13 +35,13 @@ import scala.compat.java8.OptionConverters._
 
 final class KafkaMetadataLog private (
   log: Log,
-  config: LogConfig,
   scheduler: Scheduler,
   // This object needs to be thread-safe because it is used by the snapshotting thread to notify the
   // polling thread when snapshots are created.
   snapshotIds: ConcurrentSkipListSet[OffsetAndEpoch],
   topicPartition: TopicPartition,
-  maxFetchSizeInBytes: Int
+  maxFetchSizeInBytes: Int,
+  val fileDeleteDelayMs: Long // Visible for testing,
 ) extends ReplicatedLog with Logging {
 
   override def read(startOffset: Long, readIsolation: Isolation): LogFetchInfo = {
@@ -303,7 +304,7 @@ final class KafkaMetadataLog private (
       scheduler.schedule(
         "delete-snapshot-file",
         () => Snapshots.deleteSnapshotIfExists(log.dir.toPath, snapshotId),
-        config.fileDeleteDelayMs)
+        fileDeleteDelayMs)
     }
   }
 
@@ -316,13 +317,18 @@ object KafkaMetadataLog {
 
   def apply(
     topicPartition: TopicPartition,
-    defaultLogConfig: LogConfig,
     dataDir: File,
     time: Time,
     scheduler: Scheduler,
     maxBatchSizeInBytes: Int,
     maxFetchSizeInBytes: Int
   ): KafkaMetadataLog = {
+    val props = new Properties()
+    props.put(LogConfig.MaxMessageBytesProp, maxBatchSizeInBytes.toString)
+    props.put(LogConfig.MessageFormatVersionProp, ApiVersion.latestVersion.toString)
+
+    LogConfig.validateValues(props)
+    val defaultLogConfig = LogConfig(props)
 
     val log = Log(
       dir = dataDir,
@@ -340,11 +346,11 @@ object KafkaMetadataLog {
 
     val metadataLog = new KafkaMetadataLog(
       log,
-      defaultLogConfig,
       scheduler,
       recoverSnapshots(log),
       topicPartition,
-      maxFetchSizeInBytes
+      maxFetchSizeInBytes,
+      defaultLogConfig.fileDeleteDelayMs
     )
 
     // When recovering, truncate fully if the latest snapshot is after the log end offset. This can happen to a follower
--- a/core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
+++ b/core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
@@ -42,15 +42,6 @@ final class KafkaMetadataLog private (
   maxFetchSizeInBytes: Int
 ) extends ReplicatedLog {
 
-  /* The oldest snapshot id is the snapshot at the log start offset. Since the KafkaMetadataLog doesn't
-   * currently delete snapshots, it is possible for the oldest snapshot id to not be the smallest
-   * snapshot id in the snapshotIds set.
-   */
-  private[this] var oldestSnapshotId = snapshotIds
-    .stream()
-    .filter(_.offset == startOffset)
-    .findAny()
-
   override def read(startOffset: Long, readIsolation: Isolation): LogFetchInfo = {
     val isolation = readIsolation match {
       case Isolation.COMMITTED => FetchHighWatermark
@@ -109,7 +100,7 @@ final class KafkaMetadataLog private (
 
   override def lastFetchedEpoch: Int = {
     log.latestEpoch.getOrElse {
-      latestSnapshotId.map[Int] { snapshotId =>
+      latestSnapshotId().map[Int] { snapshotId =>
         val logEndOffset = endOffset().offset
         if (snapshotId.offset == startOffset && snapshotId.offset == logEndOffset) {
           // Return the epoch of the snapshot when the log is empty
@@ -126,12 +117,12 @@ final class KafkaMetadataLog private (
   }
 
   override def endOffsetForEpoch(epoch: Int): OffsetAndEpoch = {
-    (log.endOffsetForEpoch(epoch), oldestSnapshotId.asScala) match {
+    (log.endOffsetForEpoch(epoch), earliestSnapshotId().asScala) match {
       case (Some(offsetAndEpoch), Some(snapshotId)) if (
         offsetAndEpoch.offset == snapshotId.offset &&
         offsetAndEpoch.leaderEpoch == epoch) =>
 
-        // The epoch is smaller thant the smallest epoch on the log. Overide the diverging
+        // The epoch is smaller than the smallest epoch on the log. Override the diverging
         // epoch to the oldest snapshot which should be the snapshot at the log start offset
         new OffsetAndEpoch(snapshotId.offset, snapshotId.epoch)
 
@@ -164,13 +155,14 @@ final class KafkaMetadataLog private (
 
   override def truncateToLatestSnapshot(): Boolean = {
     val latestEpoch = log.latestEpoch.getOrElse(0)
-    latestSnapshotId.asScala match {
+    latestSnapshotId().asScala match {
       case Some(snapshotId) if (snapshotId.epoch > latestEpoch ||
         (snapshotId.epoch == latestEpoch && snapshotId.offset > endOffset().offset)) =>
         // Truncate the log fully if the latest snapshot is greater than the log end offset
 
         log.truncateFullyAndStartAt(snapshotId.offset)
-        oldestSnapshotId = latestSnapshotId
+        // Delete snapshot after truncating
+        removeSnapshotFileBefore(snapshotId)
 
         true
 
@@ -228,7 +220,7 @@ final class KafkaMetadataLog private (
       if (latest.epoch > snapshotId.epoch || latest.offset > snapshotId.offset) {
         // Since snapshots are less than the high-watermark absolute offset comparison is okay.
         throw new IllegalArgumentException(
-          s"Attemting to create a snapshot ($snapshotId) that is not greater than the latest snapshot ($latest)"
+          s"Attempting to create a snapshot ($snapshotId) that is not greater than the latest snapshot ($latest)"
         )
       }
     }
@@ -258,8 +250,13 @@ final class KafkaMetadataLog private (
     }
   }
 
-  override def oldestSnapshotId(): Optional[OffsetAndEpoch] = {
-    oldestSnapshotId
+  override def earliestSnapshotId(): Optional[OffsetAndEpoch] = {
+    try {
+      Optional.of(snapshotIds.first)
+    } catch {
+      case _: NoSuchElementException =>
+        Optional.empty()
+    }
   }
 
   override def onSnapshotFrozen(snapshotId: OffsetAndEpoch): Unit = {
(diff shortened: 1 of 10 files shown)


core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -173,13 +164,16 @@ final class KafkaMetadataLog private (
...
       case Some(snapshotId) if (snapshotId.epoch > latestEpoch ||
         (snapshotId.epoch == latestEpoch && snapshotId.offset > endOffset().offset)) =>
         // Truncate the log fully if the latest snapshot is greater than the log end offset
 
         log.truncateFullyAndStartAt(snapshotId.offset)
-        oldestSnapshotId = latestSnapshotId
+        // Delete snapshot after truncating
+        val olderSnapshotIds = snapshotIds.headSet(snapshotId)
+        olderSnapshotIds.forEach(id => Snapshots.deleteSnapshotIfExists(log.dir.toPath, id))
+        snapshotIds.removeAll(olderSnapshotIds)
```
CaoManhDat: Should we call `olderSnapshotIds.clear()` here (since the olderSnapshotIds is backed by the original snapshotIds)

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -267,23 +261,31 @@ final class KafkaMetadataLog private (
...
         logStartSnapshotId.offset <= snapshotId.offset &&
         log.maybeIncrementLogStartOffset(logStartSnapshotId.offset, SnapshotGenerated)) =>
-
         log.deleteOldSegments()
-        oldestSnapshotId = Optional.of(logStartSnapshotId)
+
+        // Delete snapshot after increasing LogStartOffset
+        val expiredSnapshotIds = snapshotIds.headSet(logStartSnapshotId)
+        expiredSnapshotIds.forEach(snapshotId => Snapshots.deleteSnapshotIfExists(log.dir.toPath, snapshotId))
+        snapshotIds.removeAll(expiredSnapshotIds)
```
CaoManhDat: same here (should we create a method for this block?)

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
dengziming: In fact, after `snapshotIds.removeAll(olderSnapshotIds)`, the  element in olderSnapshotIds is also removed.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
dengziming: Done!

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
CaoManhDat: my point here is calling `olderSnapshotIds.clear()` instead of calling `removeAll()`

dengziming: Thank you, will fix this!

dengziming:
@jsancio @hachikuji , Hello, this pr is ready for review.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
```diff
@@ -261,7 +258,7 @@ default long truncateToEndOffset(OffsetAndEpoch endOffset) {
      * @return an Optional snapshot id at the log start offset if nonzero, otherwise returns an empty
      *         Optional
      */
-    Optional<OffsetAndEpoch> oldestSnapshotId();
+    Optional<OffsetAndEpoch> earliestSnapshotId();
```
jsancio, line 263: Okay. Earliest and latest are better adjectives. I think we picked these names because there is a precedence of using oldest and latest in the `ProducerStateManager`.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -267,30 +259,46 @@ final class KafkaMetadataLog private (
...
 
         true
 
       case _ => false
     }
   }
 
+  /**
+   * remove all snapshots whose end offset is less than the giving offset, also delete the corresponding
+   * snapshot files.
```
jsancio:
```suggestion
   * Removes all snapshots on the log directory whose epoch and end offset is less than the giving epoch and end offset.
```

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -267,30 +259,46 @@ final class KafkaMetadataLog private (
...
+   */
+  private def removeSnapshotFileBefore(logStartSnapshotId: OffsetAndEpoch): Unit = {
```
jsancio: How about `removeSnapshotsBefore` or `removeSnapshotFilesBefore`?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -315,7 +324,8 @@ object KafkaMetadataLog {
       }
       .forEach { path =>
         path.ifPresent { snapshotPath =>
-          if (snapshotPath.partial) {
+          if (snapshotPath.partial || snapshotPath.snapshotId.offset < log.logStartOffset) {
```
jsancio: Right. I think we need to do this because the `RaftClient` can fail in `deleteBeforeSnapshot` after updating the log start offset but before deleting the snapshots older than the log start offset. Does this match your thinking?

core/src/test/scala/kafka/raft/KafkaMetadataLogTest.scala
```diff
@@ -196,6 +227,53 @@ final class KafkaMetadataLogTest {
...
+      snapshot.freeze()
+    }
+
+    append(log, numberOfRecords - 2, epoch, log.endOffset.offset)
+    val oldSnapshotId3 = new OffsetAndEpoch(numberOfRecords, epoch)
+    TestUtils.resource(log.createSnapshot(oldSnapshotId3)) { snapshot =>
+      snapshot.freeze()
+    }
+
+    val greaterEpochSnapshotId = new OffsetAndEpoch(3 * numberOfRecords, epoch)
```
jsancio: Minor but this snapshot has the same epoch as some of the other snapshots and the "last fetched epoch" of the log. Did you mean something like `greaterSnapshotId` or just `latestSnapshotId`?

core/src/test/scala/kafka/raft/KafkaMetadataLogTest.scala
```diff
@@ -261,11 +339,60 @@ final class KafkaMetadataLogTest {
...
```
jsancio: Minor but this snapshot has the same epoch as some of the other snapshots and the "last fetched epoch" of the log. Did you mean something like `greaterSnapshotId` or just `latestSnapshotId`?

raft/src/test/java/org/apache/kafka/snapshot/SnapshotsTest.java
```diff
@@ -75,4 +77,23 @@ public void testInvalidSnapshotFilenames() {
...
+    public void testDeleteSnapshot() throws IOException {
+
+        OffsetAndEpoch snapshotId = new OffsetAndEpoch(
+            TestUtils.RANDOM.nextInt(Integer.MAX_VALUE),
+            TestUtils.RANDOM.nextInt(Integer.MAX_VALUE)
+        );
+
+        Path logDirPath = TestUtils.tempDirectory().toPath();
+        FileRawSnapshotWriter snapshot = FileRawSnapshotWriter.create(logDirPath, snapshotId, Optional.empty());
+        snapshot.freeze();
```
jsancio: Let's close the snapshot writer. How about?

```suggestion
        try (FileRawSnapshotWriter snapshot = FileRawSnapshotWriter.create(logDirPath, snapshotId, Optional.empty())) {
            snapshot.freeze();
        }
```

jsancio (commented):
Thanks for the PR @dengziming!

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
dengziming: Here I am thinking of uncleaned broker shutdown, the case you describe is one of them.

raft/src/main/java/org/apache/kafka/raft/ReplicatedLog.java
dengziming: I didn't notice `ProducerStateManager`, I choose `earliest` because of the consumer offset reset strategy, now I think both are OK.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -267,30 +259,45 @@ final class KafkaMetadataLog private (
...
       case _ => false
     }
   }
 
+  /**
+   * Removes all snapshots on the log directory whose epoch and end offset is less than the giving epoch and end offset.
+   */
+  private def removeSnapshotFilesBefore(logStartSnapshotId: OffsetAndEpoch): Unit = {
+    val expiredSnapshotIds = snapshotIds.headSet(logStartSnapshotId)
+    expiredSnapshotIds.forEach(snapshotId => Snapshots.deleteSnapshotIfExists(log.dir.toPath, snapshotId))
```
jsancio: One of the invariant for this data structure has is that if `snaspshotIds` contains a snapshot id the `KafkaRaftClient` and `Listener` can expect that the snapshot exists on the file system. With this implementation this is not strictly the case since we are first deleting the snapshot from the file system and then we are removing it from `snapshotIds`. What if we do the opposite? E.g.

```scala
import Breaks.{breakable, break}

breakable {
  val iterator = snapshotIds.iterator()
  while (iterator.hasNext()) {
    val snapshotId = iterator.next()
    if (snapshotId < logStartSnapshotId) {
      iterator.remove()
      Snapshots.deleteSnapshotIfExists(log.dir.toPath, snapshotId)
    } else {
      break
    }
  }
}
```

jsancio (commented):
Once last round from me.

The title of this PR says "Delete snapshot if leo less than the log start offset". LEO stands for log end offset. We are actually comparing the snapshot's end offset and epoch against the log start snapshot or "earliest snapshot". How about "Delete snapshots less than the snapshot at the log start"?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
dengziming: Your concern is reasonable, I changed the code and added the comments to the annotation.

jsancio (approved):
LGTM. Thanks for the PR!

dengziming:
ping @hachikuji .
retest this, please.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
```diff
@@ -92,4 +93,13 @@ public static Path createTempFile(Path logDir, OffsetAndEpoch snapshotId) throws
 
         return Optional.of(new SnapshotPath(path, new OffsetAndEpoch(endOffset, epoch), partial));
     }
+
+    /**
+     * Delete this snapshot from the filesystem.
+     */
+    public static boolean deleteSnapshotIfExists(Path logDir, OffsetAndEpoch snapshotId) throws IOException {
+        Path path = snapshotPath(logDir, snapshotId);
+        return Files.deleteIfExists(path);
```
hachikuji: When we remove other log files in `Log`, we first rename them to `{file}.deleted`. Then we schedule them for background deletion. Do you think we should do similarly for snapshot files?

hachikuji (commented):
Looks good. Just had one question.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
dengziming: This is a good catch, I changed the procedure to rename and async delete.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
```diff
@@ -92,4 +100,22 @@ public static Path createTempFile(Path logDir, OffsetAndEpoch snapshotId) throws
...
+        Path destination = Snapshots.deleteRename(path, snapshotId);
+        Utils.atomicMoveWithFallback(path, destination);
+
+        return CompletableFuture.supplyAsync(() -> {
+            try {
+                return Files.deleteIfExists(destination);
+            } catch (IOException e) {
+                throw new RuntimeException("Error deleting snapshot file " + destination + ":" + e.getMessage());
+            }
+        });
```
jsancio: This uses the default Executor in Java. I think we should instead use the same async scheduler that is used for the `KafkaMetadataLog`.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
```diff
@@ -92,4 +100,22 @@ public static Path createTempFile(Path logDir, OffsetAndEpoch snapshotId) throws
...
+
+    /**
+     * Delete this snapshot from the filesystem.
+     */
+    public static CompletableFuture<Boolean> deleteSnapshotIfExists(Path logDir, OffsetAndEpoch snapshotId) throws IOException {
+        Path path = snapshotPath(logDir, snapshotId);
+        Path destination = Snapshots.deleteRename(path, snapshotId);
+        Utils.atomicMoveWithFallback(path, destination);
+
+        return CompletableFuture.supplyAsync(() -> {
```
jsancio: This looks like the Jira and commit that introduced a delay when deleting segments: https://issues.apache.org/jira/browse/KAFKA-636. It looks like the delays is there to reduce the probability that the file is deleted while Kafka is still reading the file.

Even though `CompletableFuture::supplyAsync` run async it will attempt to run immediately.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
dengziming: I found that KafkaMetadataLog doesn't use a scheduler currently, should we add a scheduler to it?

jsancio: That's correct, it indirectly uses the scheduler in `Log.scala` when deleting segment for example.

Yes. We can pass the same scheduler that we pass to `Log.scala` to this type when it is constructed.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -299,7 +302,16 @@ final class KafkaMetadataLog private (
...
       expiredSnapshotIdsIter.remove()
-      Snapshots.deleteSnapshotIfExists(log.dir.toPath, snapshotId)
+
+      val path = Snapshots.snapshotPath(log.dir.toPath, snapshotId)
+      val destination = Snapshots.deleteRename(path, snapshotId)
+      try {
+        Utils.atomicMoveWithFallback(path, destination)
+      } catch {
+        case e: IOException =>
+          warn("Error renaming snapshot file: " + path + " to :" + destination + ", " + e.getMessage)
```
hachikuji: nit: Could this be error level? Can we use `$` substitutions? Also, let's include the full exception instead of just the message.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
dengziming: Added schedule to `KafkaMetadataLog`, PTAL. Do you think we should add a `MetadataLogSnapshot` class instead of using `OffsetAndEpoch` to represent a snapshot?

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
dengziming: If the file is deleted while Kafka is still reading the file, won't the reading fail since we have already renamed the log file?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -299,7 +290,16 @@ final class KafkaMetadataLog private (
...
+      }
+      scheduler.schedule("delete-snapshot-file", () => Snapshots.deleteSnapshotIfExists(log.dir.toPath, snapshotId))
```
hachikuji: Similar to `Log.deleteSegmentFiles`, we should probably use a delay here. I think it would be ok to either hard-code this to the default of 60s, or use `file.delete.delay.ms` from the default configuration.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -341,7 +342,7 @@ object KafkaMetadataLog {
         }
       }
 
-    val replicatedLog = new KafkaMetadataLog(log, snapshotIds, topicPartition, maxFetchSizeInBytes)
+    val replicatedLog = new KafkaMetadataLog(log, scheduler, snapshotIds, topicPartition, maxFetchSizeInBytes)
```
hachikuji: Not something we need to solve here, but I think we should put some thought into consolidating file management. Right now it's a little awkward to divide responsibility between `KafkaMetadataLog` and `Log`. For example, I think we are trying to say that `KafkaMetadataLog` is responsible for snapshots. That is mostly true, but we are relying on `Log.loadSegmentFiles` for the deletion of orphaned `.delete` snapshots.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
hachikuji: Probably not since the reference would be through the file descriptor.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
dengziming: I also find it's awkward to represent snapshot files using `OffsetAndEpoch` but we need to rely on `Snapshots` to remove and delete snapshots, will file a ticket to discuss consolidating file management, the first thought is to add  `MetadataLogSnapshot` class to represent a snapshot file.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
dengziming: Done!

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -267,30 +250,62 @@ final class KafkaMetadataLog private (
...
 
       case _ => false
     }
   }
 
+  /**
+   * Removes all snapshots on the log directory whose epoch and end offset is less than the giving epoch and end offset.
+   */
+  private def removeSnapshotFilesBefore(logStartSnapshotId: OffsetAndEpoch): Unit = {
+    val expiredSnapshotIdsIter = snapshotIds.headSet(logStartSnapshotId, false).iterator()
```
CaoManhDat, line 289: why we needs `false` here, it seems the default value of calling `headSet`?

dengziming: I add `false` to remind us that this headSet is not inclusive, I think it doesn't matter whether add or not, both are OK.

CaoManhDat: makes sense, ty!

dengziming:
Rebase onto trunk, ping @hachikuji .

raft/src/test/java/org/apache/kafka/snapshot/SnapshotsTest.java
```diff
@@ -74,5 +79,44 @@ public void testInvalidSnapshotFilenames() {
...
+        }
+
+        try (FileRawSnapshotWriter snapshot = FileRawSnapshotWriter.create(logDirPath, snapshotId, Optional.empty())) {
+            snapshot.freeze();
+
+            Path snapshotPath = Snapshots.snapshotPath(logDirPath, snapshotId);
+            assertTrue(Files.exists(snapshotPath));
+
+            // rename snapshot before deleting
+            Utils.atomicMoveWithFallback(snapshotPath, Snapshots.deleteRename(snapshotPath, snapshotId));
```
hachikuji: nit: I think this is the only difference between this block and the previous. Can we factor out a helper?

hachikuji (approved):
@dengziming Thanks for the patch, LGTM. I had just one nitpick.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -258,30 +251,62 @@ final class KafkaMetadataLog private (
...
+      try {
+        Utils.atomicMoveWithFallback(path, destination)
+      } catch {
+        case e: IOException =>
+          error(s"Error renaming snapshot file: $path to $destination", e)
+      }
+      scheduler.schedule(
+        "delete-snapshot-file",
+        () => Snapshots.deleteSnapshotIfExists(log.dir.toPath, snapshotId),
+        60 * 1000L)
```
jsancio: @dengziming, should we file an issue under the Snapshot "epic" to use the log configuration for this?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -336,19 +362,25 @@ object KafkaMetadataLog {
...
     Files
       .walk(log.dir.toPath, 1)
       .map[Optional[SnapshotPath]] { path =>
-        if (path != log.dir.toPath) {
-          Snapshots.parse(path)
-        } else {
+
+        if (path == log.dir.toPath) {
+          Optional.empty()
+        } else if (path.endsWith(Snapshots.DELETE_SUFFIX)) {
```
jsancio: Can we move the check `if (path.endsWith(Snapshots.DELETE_SUFFIX))` to `Snapshots.parse`? We can do that by improving the `SnapshotPath` to be able to represent immutable snapshot, partial snapshots and deleting snapshot.

We can do this if we replace the internal `boolean partial` in `SnapshotPath` with an enum that tracks the 3 possible states for a snapshot.

What do you think?

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -336,19 +362,25 @@ object KafkaMetadataLog {
...
+          Files.deleteIfExists(path)
```
jsancio: If you implement my other comment, I think you can move this delete operation to `forEach` below.

It is nice to have `map` operations that are side effect free.

core/src/test/scala/kafka/raft/KafkaMetadataLogTest.scala
```diff
@@ -32,13 +32,13 @@ import org.apache.kafka.common.utils.Utils
 import org.apache.kafka.raft.internals.BatchBuilder
 import org.apache.kafka.raft.{KafkaRaftClient, LogAppendInfo, LogOffsetMetadata, OffsetAndEpoch, RecordSerde, ReplicatedLog}
 import org.apache.kafka.snapshot.{SnapshotPath, Snapshots}
-import org.junit.jupiter.api.Assertions.{assertEquals, assertFalse, assertThrows, assertTrue}
+import org.junit.jupiter.api.Assertions.{assertEquals, assertFalse, assertNotEquals, assertThrows, assertTrue}
```
jsancio, line 35: This comment applies to all of the tests/changes in this file.

For all of the tests that were added to this file can we also add them to `MockLogTest`, if they apply? It is good to have tests that validate that these two implementations behave similarly.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
```diff
@@ -16,16 +16,22 @@
...
 import java.util.Optional;
-import org.apache.kafka.raft.OffsetAndEpoch;
 
 public final class Snapshots {
-    private static final String SUFFIX =  ".checkpoint";
-    private static final String PARTIAL_SUFFIX = String.format("%s.part", SUFFIX);
+    private static final Logger log = LoggerFactory.getLogger(Snapshots.class);
+    public static final String SUFFIX = ".checkpoint";
+    public static final String PARTIAL_SUFFIX = String.format("%s.part", SUFFIX);
+    public static final String DELETE_SUFFIX = ".deleted";
```
jsancio: If you implement my suggestions to `Snapshot.parse` and `SnapshotPath`, I think we can make all of these fields private.

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
```diff
@@ -92,4 +102,24 @@ public static Path createTempFile(Path logDir, OffsetAndEpoch snapshotId) throws
...
+     */
+    public static boolean deleteSnapshotIfExists(Path logDir, OffsetAndEpoch snapshotId) {
+        Path path = Snapshots.snapshotPath(logDir, snapshotId);
+        Path destination = Snapshots.deleteRename(path, snapshotId);
+        try {
+            // rename before deleting if target file is not renamed
+            if (Files.exists(path) || Files.notExists(destination)) {
+                Utils.atomicMoveWithFallback(path, destination);
+            }
+            return Files.deleteIfExists(destination);
```
jsancio: Hmm. How about?

```suggestion
            return Files.deleteIfExists(path) | Files.deleteIfExists(destination);
```

And maybe rename the variables `path` and `destination` to something like `immutablePath` and `deletingPath` respectively.

jsancio (commented):
Thanks for the changes. A few more suggestions.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
hachikuji: Perhaps we may as well do it here?

raft/src/test/java/org/apache/kafka/snapshot/SnapshotsTest.java
dengziming: Firstly I didn't understand and now I got it!

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
```diff
@@ -16,16 +16,21 @@
...
 import java.text.NumberFormat;
 import java.util.Optional;
-import org.apache.kafka.raft.OffsetAndEpoch;
 
 public final class Snapshots {
-    private static final String SUFFIX =  ".checkpoint";
+    private static final Logger log = LoggerFactory.getLogger(Snapshots.class);
+    private static final String SUFFIX = ".checkpoint";
     private static final String PARTIAL_SUFFIX = String.format("%s.part", SUFFIX);
+    private static final String DELETE_SUFFIX = ".deleted";
```
jsancio: @dengziming Sorry that I didn't notice this earlier. Why don't we make the `DELETE_SUFIX` similar to the `PARTIAL_SUFFIX`? E.g. `public static final String DELETE_SUFFIX = String.format("%s.deleted", SUFFIX);`

This will give us the follow filename suffixes:

1. Immutable snapshots - `.checkpoint`
2. Partial snapshots - `.checkpoint.part`
3. Deleted snapshots - `checkpoint.deleted`

dengziming: Sorry I made a mistake when understanding `Log.deleteSegmentFiles`, I will change it.

core/src/test/scala/kafka/raft/KafkaMetadataLogTest.scala
dengziming: I add some changes in `MockLog` to delete the elder snapshots, and 2 unit tests.

dengziming (commented):
Resolve comments:
1. Delete older snapshot in MockLog and add corresponding unit tests
2. Use LogConfig.fileDeleteDelayMs when deleting snapshot file
3. optimize methods in `Snapshots`

raft/src/main/java/org/apache/kafka/snapshot/Snapshots.java
dengziming: This looks better, I change the code and the annotations.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
```diff
@@ -35,21 +34,14 @@ import scala.compat.java8.OptionConverters._
 
 final class KafkaMetadataLog private (
   log: Log,
+  config: LogConfig,
```
jsancio: The `LogConfig` in `log: Log` should be accessible to this class as `log.config`.

jsancio (approved):
LGTM! One minor suggestion.

core/src/main/scala/kafka/raft/KafkaMetadataLog.scala
hachikuji: I'm slightly more inclined to pass through the configuration value as a parameter in the constructor here rather than getting in the habit of relying on `LogConfig`. We're in a an awkward state at the moment in regard to `LogConfig`. We depend on some of the configurations, but others we ignore. It relates to this JIRA I filed: https://issues.apache.org/jira/browse/KAFKA-12423. My thinking for now is perhaps to try and keep the coupling with `LogConfig` from getting too tight. Maybe we could do something like this in `apply`:
```scala
    val metadataLog = new KafkaMetadataLog(
      log,
      recoverSnapshots(log),
      topicPartition,
      maxFetchSizeInBytes,
      defaultLogConfig.fileDeleteDelayMs
    )
```
This is probably borderline overkill, but the point is to try and keep the configurations that we rely on in this layer separable.

jsancio: Sounds good to me. We can revisit this in https://issues.apache.org/jira/browse/KAFKA-12423.

hachikuji (approved):
LGTM

PR #9999, opened 2021-01-29
MINOR: Ensure `InterBrokerSendThread` closes `NetworkClient`

hachikuji:
We should ensure `NetworkClient` is closed properly when `InterBrokerSendThread` is shutdown. Also use `initiateShutdown` instead of `wakeup()` to alert polling thread.

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #10024, opened 2021-02-02
KAFKA-12273 InterBrokerSendThread#pollOnce throws FatalExitError even…

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-12273

kafka tests sometimes shutdown gradle process with non-zero code. The (one of) root cause is that ```InterBrokerSendThread#pollOnce``` encounters ```DisconnectException``` when ```NetworkClient``` is closing. ```DisconnectException``` should be viewed as "expected" error as we do close thread. In other words, ```InterBrokerSendThread#pollOnce``` should swallow such exception.


core/src/test/scala/kafka/common/InterBrokerSendThreadTest.scala
```diff
@@ -51,6 +52,26 @@ class InterBrokerSendThreadTest {
...
+      try super.pollOnce(maxTimeoutMs)
+      catch {
+        case e: Throwable => exceptionCallback(e)
+      }
+    }
+
+  }
+
+  @Test
+  def shutdownThreadShouldNotCauseException(): Unit = {
```
ijuma, line 65: I don't think we should have a test inside a non test class. Gradle won't run it for once and it's quite confusing.

ijuma (commented):
FatalExitException is mentioned in the PR description. What throws that?

chia7712:
> FatalExitException is mentioned in the PR description. What throws that?

please take a look at following log (from gradle)

`
    [2021-02-03 02:03:58,599] ERROR [Transaction Marker Channel Manager 0]: unhandled exception caught in InterBrokerSendThread (kafka.coordinator.transaction.TransactionMarkerChannelManager:76)
    org.apache.kafka.common.errors.DisconnectException: NetworkClient is no longer active, state is CLOSING
`

It can reproduce by looping ```AuthorizerIntegrationTest``` or ```PlaintextAdminIntegrationTest```.

core/src/test/scala/kafka/common/InterBrokerSendThreadTest.scala
chia7712: @ijuma Pardon me. I failed to catch your point. ```InterBrokerSendThreadTest.scala``` is a true test case and it seems to me the test make sure InterBrokerSendThreadTest does not throw fatal exception when it is shutdown correctly.

ijuma: My bad, I thought this was `TestInterBrokerSendThread`, but that's an inner class.

chia7712:
@ijuma any feedback for this PR? It happens occasionally on our jenkins (for example: https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-10068/3/execution/node/86/log/?cloudbees-analytics-link=scm-reporting%2Ferror%2Fstep)

chia7712:
another related build (https://ci-builds.apache.org/job/Kafka/job/kafka-pr/job/PR-10012/5/console)

ijuma:
@chia7712 Do you know why this started happening more often?

chia7712:
> Do you know why this started happening more often?

I believe it is related to #9999. `networkClient.initiateClose()` changes the state of `networkClient` to inactive and so the loop in thread encountered state exception.

ijuma:
Thanks for the reference, that's helpful. cc @dajac @hachikuji

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
```diff
@@ -78,6 +77,9 @@ abstract class InterBrokerSendThread(
       failExpiredRequests(now)
       unsentRequests.clean()
     } catch {
+      case _: DisconnectException if isShutdownInitiated =>
+        // DisconnectException is caused by NetworkClient#ensureActive
+        // this thread is closing so this error is acceptable
```
ijuma: I am not sure about this. We would ideally drain the existing requests (up to a timeout), not accept new ones and then close the network client. This change seems like it would hide the problem, but not necessarily fix it. Let's see what @dajac and/or @hachikuji say.

chia7712:
@dajac @hachikuji Could you please take a look? This issue causes trouble to QA since it obstructs gradle from generating test reports.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
dajac: `DisconnectException` is indeed expected after `initiateClose` is called. It seems OK to handle it here but I would rather use `client.active()` instead of `isShutdownInitiated`. This is something that we should have done in https://github.com/apache/kafka/pull/9999.

@ijuma I think that your point is legitimate. It seems that we brutally stop the send thread without really draining the inflight requests. This maybe OK as we stop this thread only when the broker is shutdown anyway. In order to restore the CI, I suggest to proceed with the minor fix and to review/address this point in a followup PR if necessary.

chia7712: @dajac Thanks for your response.

>  I would rather use client.active() instead of isShutdownInitiated. This is something that we should have done in #9999.

It seems to me the exception is available only if the thread is going to shutdown. If network client get closed and it is not caused by calling Thread#shutdown, we still throw fatal exception as it is totally unexpected behavior.

chia7712: @dajac Could you take a look at above comment? thanks.

chia7712:
`Build / JDK 11 / org.apache.kafka.connect.integration.InternalTopicsIntegrationTest.testCreateInternalTopicsWithFewerReplicasThanBrokers`

It is fixed by #10152 and #10158

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
dajac: @chia7712 I understand your point. I was thinking that the exception is actually expected when `networkClient.initiateClose()` is called so it makes sense to rely on the `networkClient` to decide if it is expected or not. Today, this is tight to the thread being shutdown as you pointed out. If we would change this in the future, it will be really easy to mis changing this condition and this could led to a subtile bug like this one. I don't feel strong for either approaches so I leave it up to you.

chia7712: >  If we would change this in the future, it will be really easy to mis changing this condition and this could led to a subtile bug like this one

make sense :)

chia7712:
@dajac I have addressed your comment in latest commit. Please take a look.

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
```diff
@@ -78,6 +77,8 @@ abstract class InterBrokerSendThread(
       failExpiredRequests(now)
       unsentRequests.clean()
     } catch {
+      case _: DisconnectException if !networkClient.active() =>
+        // Other threads can shutdown network client by calling InterBrokerSendThread#shutdown
```
dajac: Could we also state the `DisconnectException` is expected when `NetworkClient#initiateClose` is called?

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
dajac: @chia7712 Could we file a Jira to follow up on @ijuma's point? I think that we should review this but it is fine to do this separately from this one.

chia7712: sure. see https://issues.apache.org/jira/browse/KAFKA-12356

core/src/main/scala/kafka/common/InterBrokerSendThread.scala
chia7712: done

dajac (approved):
LGTM, assuming tests pass. @chia7712 Please, excuse me for the delay on this one.

chia7712:
@dajac don’t worry. Thanks for your reviews!

chia7712:
push to trunk and 2.8

PR #10025, opened 2021-02-02
Upstream QuotaConfigs

mumrah:
This PR moves static property definitions for user client quotas into a new class called QuotaConfigs in the clients module under the o.a.k.common.config.internals package. This is needed to support the client quotas work in the quorum based controller.


cmccabe:
So this is just moving the configs but not changing them.

LGTM pending jenkins.

Thanks, @mumrah

cmccabe (approved):

cmccabe:
pushed

PR #10026, opened 2021-02-02
MINOR: Add mock implementation of `BrokerToControllerChannelManager`

hachikuji:
Tests involving `BrokerToControllerChannelManager` are simplified by being able to leverage `MockClient`. This patch introduces a `MockBrokerToControllerChannelManager` implementation which makes that possible.

The patch updates `ForwardingManagerTest` to use `MockBrokerToControllerChannelManager`. We also add a couple additional timeout cases, which exposed a minor bug. Previously we were using the wrong `TimeoutException`, which meant that expected timeout errors were in fact translated to `UNKNOWN_SERVER_ERROR`.

### DIFF
--- a/core/src/test/scala/unit/kafka/server/ForwardingManagerTest.scala
+++ b/core/src/test/scala/unit/kafka/server/ForwardingManagerTest.scala
@@ -19,15 +19,17 @@ package kafka.server
 import java.net.InetAddress
 import java.nio.ByteBuffer
 import java.util.Optional
-import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicReference
 
 import kafka.network
 import kafka.network.RequestChannel
 import kafka.utils.MockTime
-import org.apache.kafka.clients.{ClientResponse, RequestCompletionHandler}
+import org.apache.kafka.clients.{MockClient, NodeApiVersions}
+import org.apache.kafka.clients.MockClient.RequestMatcher
+import org.apache.kafka.common.Node
 import org.apache.kafka.common.config.{ConfigResource, TopicConfig}
 import org.apache.kafka.common.memory.MemoryPool
-import org.apache.kafka.common.message.AlterConfigsResponseData
+import org.apache.kafka.common.message.{AlterConfigsResponseData, ApiVersionsResponseData}
 import org.apache.kafka.common.network.{ClientInformation, ListenerName}
 import org.apache.kafka.common.protocol.{ApiKeys, Errors}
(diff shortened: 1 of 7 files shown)


core/src/test/scala/unit/kafka/server/ForwardingManagerTest.scala
```diff
@@ -19,135 +19,139 @@ package kafka.server
...
-import org.mockito.ArgumentMatchers._
 import org.mockito.Mockito
 
 import scala.jdk.CollectionConverters._
 
 class ForwardingManagerTest {
-  private val brokerToController = Mockito.mock(classOf[BrokerToControllerChannelManager])
   private val time = new MockTime()
+  private val client = new MockClient(time)
+  private val controllerNodeProvider = Mockito.mock(classOf[ControllerNodeProvider])
```
mumrah, line 47: Since this is just a single method interface, can we use a lambda instead of a mock?

core/src/test/scala/unit/kafka/server/ForwardingManagerTest.scala
```diff
@@ -19,135 +19,139 @@ package kafka.server
...
+    forwardingManager.forwardRequest(request, res => res.foreach(response.set))
+    brokerToController.poll()
+    assertNull(response.get)
+
+    // After reaching the retry timeout, we get a disconnect. Instead of retrying,
+    // we should fail the request with a timeout error.
+    time.sleep(brokerToController.retryTimeoutMs)
+    client.respond(testAlterConfigRequest.getErrorResponse(0, Errors.UNKNOWN_SERVER_ERROR.exception), true)
+    brokerToController.poll()
+    brokerToController.poll()
```
mumrah, line 150: Why the second poll here? Is this to trigger the timeout error?

mumrah (approved):
Nice, I like it. Two questions inline, but otherwise LGTM

core/src/test/scala/unit/kafka/server/ForwardingManagerTest.scala
hachikuji: Yeah, this is one thing that is kind of annoying with this model. We need one `poll` to receive the disconnect response and then one more to trigger the callback.

core/src/test/scala/unit/kafka/server/ForwardingManagerTest.scala
hachikuji: `ControllerNodeProvider` also is used to expose `listenerName` and `securityProtocol`, so the mock might still be simpler.

mumrah: Gotcha, yea a mock is probably simpler in this case. If we end up doing a lot of ControllerNodeProvider mocking down the road, it might make sense for a full test implementation, but not for this PR

PR #10027, opened 2021-02-02
KAFKA-12270: Handle race condition when Connect tasks attempt to create topic

rhauch:
**_This is option 1 for the fix, which is simpler but less ideal because it results in an extra admin client call. See #10032 for option 2 that is a bit more complicated but results in no extra admin client calls._**

When a source connector is configured to create missing topics has multiple tasks that generate records for the same topic, it is possible that multiple tasks may simultaneously describe the topic, find it does not exist, and attempt to create the task. One of those create topic requests will succeed, and the other concurrent tasks will receive the response from the topic admin as having not created the task and will fail unnecessarily.

This change corrects the logic after the create topic method returns false (signifying that it did not create the task) to check again whether the topic exists (i.e., was created since the worker task last described the topic) and to continue if it does exist and fail only if the topic did not exist after the second check. This is not as efficient as would be ideal, but this ensures that tasks do not fail in this race condition and that tasks will continue to fail if the topic could not be created and did not exist after the create topic request did not create the topic.

Modified one unit test that mocks the behavior, and added a new test with the new behavior.

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
@@ -416,6 +416,15 @@ class WorkerSourceTask extends WorkerTask {
             topicCreation.addTopic(topic);
             log.info("Created topic '{}' using creation group {}", newTopic, topicGroup);
         } else {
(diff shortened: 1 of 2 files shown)


rhauch:
Closing in favor of #10032.

PR #10028, opened 2021-02-02
Replace BrokerStates.scala with BrokerState.java

cmccabe:
Making BrokerState a straightforward Java enum makes it easier to use from Java code.

The existing BrokerState class is also awkward to use because it is mutable
(it holds a volatile variable inside).  While there is also an immutable
case class, there is no way to go from a numeric type to the case class.


connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedKafkaCluster.java
```diff
@@ -241,13 +240,13 @@ public String zKConnectString() {
     }
 
     /**
-     * Get the brokers that have a {@link RunningAsBroker} state.
+     * Get the brokers that have a {@link BrokerState.RUNNING} state.
```
chia7712: ```{@link BrokerState#RUNNING} ```

chia7712 (approved):

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -60,7 +61,7 @@ class LogManager(logDirs: Seq[File],
                  val retentionCheckMs: Long,
                  val maxPidExpirationMs: Int,
                  scheduler: Scheduler,
-                 val brokerState: BrokerState,
+                 val brokerState: AtomicReference[BrokerState],
```
ijuma, line 64: Is there a reason why you're using an AtomicReference instead of a volatile var?

cmccabe: The log manager needs to alter the broker state using the `AtomicReference`.  Since the broker state is owned by `KafkaBroker`, this would not be possible by just using a volatile variable.

(This is an ugly pattern, I think, and a later patch set removes it.  But in the interest of keeping this PR small, we preserve the original behavior where LogManager changes the broker state directly for now)

PR #10030, opened 2021-02-03
MINOR: Add KafkaEventQueue

cmccabe:
Add KafkaEventQueue, which is used by the KIP-500 controller to manage its event queue. Compared to using an Executor, KafkaEventQueue has the following advantages:

* Events can be given "deadlines."  If an event lingers in the queue beyond the deadline, it will be completed with a timeout exception.  This is useful for implementing timeouts for controller RPCs.

* Events can be prepended to the queue as well as appended.

* Events can be given tags to make them easier to manage.  This is especially useful for rescheduling or cancelling events which were previously scheduled to execute in the future.


ijuma:
Thanks for the PR. Why does this have to be in the `common` package?

cmccabe:
> Why does this have to be in the common package?

I put it in the clients module since it's fairly generic and could be used in other places.

However, thinking about it more, maybe we should put it in the metadata module for now, to avoid adding more stuff to clients.

ijuma:
Makes sense. Yes, I think it makes sense to start in metadata if that's the main place we intend to use it (for now).

cmccabe:
It's also used in the broker, but that depends on the metadata module

ijuma:
@junrao Do you have time to help review this?

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
```diff
@@ -0,0 +1,232 @@
...
+     * @param tag                   If this is non-null, the unique tag to use for this
+     *                              event.  If an event with this tag already exists, it
+     *                              will be cancelled.
+     * @param deadlineNsCalculator  A function which takes as an argument the existing
+     *                              deadline for the event with this tag (or null if the
+     *                              event has no tag, or if there is none such), and
+     *                              produces the deadline to use for this event.
+     * @param event                 The event to schedule.
+     */
+    default void scheduleDeferred(String tag,
```
junrao, line 167: Perhaps we could document the ordering guarantee with the deferred event compared with other types of events?

cmccabe:
re-trigger jenkins to try to get a green build

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+                    }
+                } finally {
+                    lock.unlock();
+                }
+            }
+        }
+
+        private void enqueue(EventContext eventContext,
+                             Function<Long, Long> deadlineNsCalculator) {
+            lock.lock();
```
junrao, line 263: Do we need the lock here again since the caller already locks?

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+                        break;
+                    case PREPEND:
+                        head.insertAfter(eventContext);
+                        if (queueWasEmpty) {
+                            shouldSignal = true;
+                        }
+                        break;
+                    case DEFERRED:
+                        if (deadlineNs == null) {
+                            eventContext.completeWithException(new RuntimeException(
```
junrao: Since this is unexpected, should we just throw the exception directly back to the caller?

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+                        long now = time.nanoseconds();
+                        long timeoutNs = entry.getKey();
+                        EventContext eventContext = entry.getValue();
+                        if (timeoutNs <= now) {
+                            if (eventContext.insertionType == EventInsertionType.DEFERRED) {
+                                // The deferred event is ready to run.  Prepend it to the
+                                // queue.  (The value for deferred events is a schedule time
+                                // rather than a timeout.)
+                                remove(eventContext);
+                                head.insertAfter(eventContext);
```
junrao: Since now can change after each while loop, it's possible that deferred events are prepended to the queue in reverse order. Does that matter?

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    @Override
+    public void close() throws InterruptedException {
+        beginShutdown("KafkaEventQueue#close");
+        eventHandlerThread.join();
+        log.info("closed event queue.");
```
junrao, line 424: Is this necessary since we already have an info logging in beginShutdown().

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
```diff
@@ -0,0 +1,232 @@
...
+        enqueue(EventInsertionType.APPEND, null, null, event);
+    }
+
+    /**
+     * Enqueue an event to be run in FIFO order.
+     *
+     * @param deadlineNs        The time in monotonic nanoseconds after which the future
+     *                          is completed with a
+     *                          @{org.apache.kafka.common.errors.TimeoutException},
+     *                          and the event is cancelled.
```
jsancio, line 145: Which future? Did you mean `Event`. I assume that `deadlineNs` is the deadline for scheduling/executing the `run` method. The `run` method can take longer than `deadlineNs`.

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * Schedule an event to be run at a specific time.
+     *
+     * @param tag                   If this is non-null, the unique tag to use for this
+     *                              event.  If an event with this tag already exists, it
+     *                              will be cancelled.
+     * @param deadlineNsCalculator  A function which takes as an argument the existing
+     *                              deadline for the event with this tag (or null if the
+     *                              event has no tag, or if there is none such), and
+     *                              produces the deadline to use for this event.
```
jsancio, line 161: The headline description says "Schedule an event to be run at a specific time" yet this param's description seem to indicate a different behavior.

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /**
+     * Enqueue an event to be run in FIFO order.
+     *
+     * @param insertionType         How to insert the event.
+     *                              PREPEND means insert the event as the first thing
+     *                              to run.  APPEND means insert the event as the last
+     *                              thing to run.  DEFERRED means insert the event to
+     *                              run after a delay.
```
jsancio, line 195: The headline description says "Enqueue an event to be run in FIFO order." but looking the param description it doesn't looking like it is FIFO order since it support prepend, append and deferred.

What do you mean by DEFEREED? How is the delay specified? I see the param `deadlineNsCalculator` but it looks like this describe the maximum amount of time this event is allowed to stay in the queue before it is cancelled by the queue.

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
```diff
@@ -0,0 +1,232 @@
...
+     * @param source        The source of the shutdown.
+     * @param cleanupEvent  The mandatory event to invoke after all other events have
+     *                      been processed.
+     * @param timeUnit      The time unit to use for the timeout.
+     * @param timeSpan      The amount of time to use for the timeout.
+     *                      Once the timeout elapses, any remaining queued
+     *                      events will get a
+     *                      @{org.apache.kafka.common.errors.TimeoutException}.
+     */
+    void beginShutdown(String source, Event cleanupEvent, TimeUnit timeUnit, long timeSpan);
```
jsancio: In Java the unit of measure `timeUnit` and scalar `timeSpan` are declare in reverse order. For example:
```java
     void beginShutdown(String source, Event cleanupEvent, long timeSpan, TimeUnit timeUnit);
```

metadata/src/main/java/org/apache/kafka/queue/EventQueueClosedException.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.kafka.queue;
+
+
+public class EventQueueClosedException extends RuntimeException {
```
jsancio: Can we use Java's existing type `java.util.concurrent.RejectedExecutionException`?

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+        /**
+         * The next pointer in our circular doubly-linked list.
+         */
+        private EventContext next = this;
+
+        /**
+         * If this event is in the delay map, this is the key it is there under.
+         * If it is not in the map, this is null.
+         */
+        private Long deadlineNs = null;
```
jsancio: We can use `OptionalLong`.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+
+        /**
+         * The tag associated with this event.
+         */
+        private String tag;
```
jsancio, line 71: We can use `Optional<String>`.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+
+        EventContext(Event event, EventInsertionType insertionType, String tag) {
+            this.event = event;
+            this.insertionType = insertionType;
+            this.tag = tag;
+        }
+
+        /**
+         * Insert a new node in the circularly linked list after this node.
```
jsancio: How about "Insert the EventContext other in the circularly linked list after this node." I would also mention that this requires that `other.isSingleton()` is true. Should we check for that in this implementation?

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+        /**
+         * Run the event associated with this EventContext.
+         */
+        void run() throws InterruptedException {
+            try {
+                event.run();
+            } catch (InterruptedException e) {
+                throw e;
+            } catch (Exception e) {
+                event.handleException(e);
```
jsancio: Do we want to handle the case where `Event::handleException` throws?

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
```diff
@@ -0,0 +1,232 @@
...
+
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+
+
+public interface EventQueue extends AutoCloseable {
+    interface Event {
+        void run() throws Exception;
+        default void handleException(Throwable e) {}
+    }
```
jsancio, line 46: Let's document these two methods. For example, looking at the implementation it looks like `handleException` needs to handle:
1. Exceptions thrown by `Event::run`
2. Exception passed by the `EventQueue` like `EventQueueClosedException`, `TimeoutException`, anything else?

Any thoughts on why this design instead of not allowing `run` to throw a checked exception? For example:
```java
interface Event extends Runnable {
    default void cancel(Throwable e) {}
}
```

metadata/src/test/java/org/apache/kafka/queue/KafkaEventQueueTest.java
```diff
@@ -0,0 +1,239 @@
...
+        KafkaEventQueue queue =
+            new KafkaEventQueue(Time.SYSTEM, new LogContext(), "testHandleEvents");
+        AtomicInteger numEventsExecuted = new AtomicInteger(0);
+        CompletableFuture<Integer> future1 = new CompletableFuture<>();
+        queue.prepend(new FutureEvent<>(future1, () -> {
+            assertEquals(1, numEventsExecuted.incrementAndGet());
+            return 1;
+        }));
+        CompletableFuture<Integer> future2 = new CompletableFuture<>();
+        queue.appendWithDeadline(Time.SYSTEM.nanoseconds() + TimeUnit.SECONDS.toNanos(30),
```
junrao: Strictly speaking, with system time, this event could time out before the first future is executed. Should we use mock time?

metadata/src/test/java/org/apache/kafka/queue/KafkaEventQueueTest.java
```diff
@@ -0,0 +1,239 @@
...
+        assertEquals(3, numEventsExecuted.get());
+    }
+
+    @Test
+    public void testScheduleDeferred() throws Exception {
+        KafkaEventQueue queue =
+            new KafkaEventQueue(Time.SYSTEM, new LogContext(), "testAppendDeferred");
+
+        // Wait for the deferred event to happen after the non-deferred event.
+        // It may not happpen every time, so we keep trying until it does.
```
junrao: typo happpen

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+                    long awaitNs = Long.MAX_VALUE;
+                    Map.Entry<Long, EventContext> entry = delayMap.firstEntry();
+                    if (entry != null) {
+                        // Search for timed-out events or deferred events that are ready
+                        // to run.
+                        long now = time.nanoseconds();
+                        long timeoutNs = entry.getKey();
+                        EventContext eventContext = entry.getValue();
+                        if (timeoutNs <= now) {
+                            if (eventContext.insertionType == EventInsertionType.DEFERRED) {
```
jsancio, line 214: Hmm. Does this mean that `DEFERRED` events cannot timeout? How about making the two threshold described independently when the event is enqueued?

junrao (commented):
@cmccabe : Thanks for the PR. A few comments below.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+
+        /**
+         * The head of the event queue.
+         */
+        private final EventContext head = new EventContext(null, null, null);
+
+        /**
+         * An ordered map of times in monotonic nanoseconds to events to time out.
+         */
+        private final TreeMap<Long, EventContext> delayMap = new TreeMap<>();
```
jsancio: Hmm. The word "delay" suggests relative time but it looks like the keys are absolute times. How about `scheduleMap`?

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+                    case PREPEND:
+                        head.insertAfter(eventContext);
+                        if (queueWasEmpty) {
+                            shouldSignal = true;
+                        }
+                        break;
+                    case DEFERRED:
+                        if (deadlineNs == null) {
+                            eventContext.completeWithException(new RuntimeException(
+                                "You must specify a deadline for deferred events."));
```
jsancio: Hmm. This is surprising to me. The caller scheduling the event passed illegal arguments to the queue. Shouldn't this method throw an IllegalArgumentException instead of sending an exception to the event?

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+                        }
+                        break;
+                }
+                if (deadlineNs != null) {
+                    long insertNs =  deadlineNs;
+                    long prevStartNs = delayMap.isEmpty() ? Long.MAX_VALUE : delayMap.firstKey();
+                    // If the time in nanoseconds is already taken, take the next one.
+                    while (delayMap.putIfAbsent(insertNs, eventContext) != null) {
+                        insertNs++;
+                    }
```
jsancio, line 306: Hmm. This can cause an event with delay X to be schedule after an event with delay Y even if X < Y.

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
jsancio: Looking at the implementation, it looks like this Long is an absolute time. Is there a reason why absolute times are used instead of relative times like delay and timeout?

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
cmccabe: I added a comment saying that "Once the deadline has arrived, the event will be prepended to the queue"

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
```diff
@@ -0,0 +1,232 @@
...
+    default void appendWithDeadline(long deadlineNs, Event event) {
+        enqueue(EventInsertionType.APPEND, null, __ -> deadlineNs, event);
+    }
+
+    /**
+     * Schedule an event to be run at a specific time.
+     *
+     * @param tag                   If this is non-null, the unique tag to use for this
+     *                              event.  If an event with this tag already exists, it
+     *                              will be cancelled.
```
jsancio, line 157: It would be interesting to explain this design. I get the impression that we are interesting in rescheduling events that haven't triggered.

One way to change the deadline is to specified the previous event that is being rescheduled. E.g.
```java
rescheduleDeferred(Event oldEvent, Function<Long, Long> deadlineNsCalculator, Event newEvent);
```

This API allows the deadline to either increase or decrease. Are both cases used by the Controller?

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+    @Override
+    public void enqueue(EventInsertionType insertionType,
+                        String tag,
+                        Function<Long, Long> deadlineNsCalculator,
+                        Event event) {
+        lock.lock();
+        try {
+            EventContext eventContext = new EventContext(event, insertionType, tag);
+            if (closingTimeNs != Long.MAX_VALUE) {
+                eventContext.completeWithException(new EventQueueClosedException());
```
jsancio: Looks like we are calling unknown code while holding a lock.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+                    deadlineNsCalculator == null ? __ -> null : deadlineNsCalculator);
+            }
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    @Override
+    public void cancelDeferred(String tag) {
+        lock.lock();
```
jsancio: We can move this lock to `EventHandler::cancelDeferred`. `EventHandler` is already handling concurrent access to the fields in this object.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+                              TimeUnit timeUnit, long timeSpan) {
+        if (timeSpan < 0) {
+            throw new IllegalArgumentException("beginShutdown must be called with a " +
+                "non-negative timeout.");
+        }
+        Objects.requireNonNull(newCleanupEvent);
+        lock.lock();
+        try {
+            if (cleanupEvent != null) {
+                log.debug("{}: Event queue is already shut down.", source);
```
jsancio: ".. shutting down"

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+
+    @Override
+    public void wakeup() {
+        lock.lock();
+        try {
+            eventHandler.cond.signal();
+        } finally {
+            lock.unlock();
+        }
+    }
```
jsancio, line 418: How about moving this concurrency complexity to `EventHandler` by adding a method called `EventHandler::wakeup`?

jsancio (commented):
Thanks for the PR.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: good point... we don't need to lock here, since the caller already does.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: I think it's probably better to handle it through the Event callback, since all the other errors are handled that way.  We do not throw exceptions from any of the enqueue functions-- they always succeed, and whatever error or timeout happens is handled through the events.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: Unless I'm missing something, the event that got prepended to the queue will be immediately run on the next iteration of the loop.  There is no opportunity to prepend multiple deferred events to the toRun queue, since only one event can be prepended per loop run, and that event will be immediately run at the top of the loop.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: I think it's useful since events can still be executed after beginShutdown is called.  joining the thread is an important milestone (and won't happen if a buggy event hangs) so good to log.

metadata/src/test/java/org/apache/kafka/queue/KafkaEventQueueTest.java
cmccabe: I will enlarge the timeout to 60 seconds so that it's equal to the test timeout as a whole.  Assuming that timeout is reasonable (and I think it is) then there should be no issue...

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
cmccabe: Good question.  I clarified this comment a bit.

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
cmccabe: We need to support absolute times in order to support things like `EarliestDeadlineFunction`, and mixing relative and absolute times is kind of confusing.

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
cmccabe: I will change the description to reflect the fact that this function supports PREPEND and DEFERRED, so it's not always "FIFO order".

How to use deferred events is described in more detail in `scheduleDeferred`

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
cmccabe: fair enough, I will put timeSpan first

metadata/src/main/java/org/apache/kafka/queue/EventQueueClosedException.java
cmccabe: Good idea.  I will just use that exception instead.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: ok

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: I think it's pretty standard for linked list insertion functions to overwrite the existing prev and next pointers of the object being inserted

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: handleException really should not throw.  But I will add an ERROR log message if it does.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
junrao: Hmm, after the event is prepended, we continue to the next while loop before dequeuing and setting toRun, right? Then, in the next loop, we could be prepending the next deferred item before the previous one runs.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: I'd rather not here.

Checking strings for null seems like idiomatic Java.  Strings have always been nullable reference types.  This is different than the OptionalLong case above, where it always felt awkward checking if a numeric type was null.  At least that's my feeling.

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
cmccabe: > Let's document these two methods

Sure.  I will add some JavaDoc.

> Any thoughts on why this design instead of not allowing run to throw a checked exception? For example:

Checked exceptions are sort of a failed idea.  This would just force you to wrap all the checked exceptions that the event could throw, and unwrap them in handleException.  That would not help the clarity of the controller code.

Also, implementing Runnable isn't really useful here because Java 8 lambdas can be used on any single-function interface, not just Runnable.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: > Hmm. Does this mean that DEFERRED events cannot timeout? 

yes

> How about making the two threshold described independently when the event is enqueued?

there's no use-case for timing out deferred events.  they go to the head of the queue when their time arrives in any case.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: We can call it `deadlineMap`

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: In general I would prefer to do all the error handling in one place (the event).  Having to handle errors in multiple places is more difficult.

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
cmccabe: > One way to change the deadline is to specified the previous event that is being rescheduled

Having to store the previous event object somewhere would be very annoying.  Since the thread triggering deferred events is often not the controller thread itself, it would require complex locking or volatile objects.  Also then you run into issues like should the equals function be used, or object identity.

It's a lot easier to just pass a string (which is immutable).  This also has the nice effect of describing the deferred event.

> This API allows the deadline to either increase or decrease. Are both cases used by the Controller?

In the controller we are mainly interested in making the deadline closer to "now" (decreasing it) or cancelling the deferred event.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: fair point.  let's move this outside the lock.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: I will add a line to the JavaDoc stating "Events whose deadlines are only a few nanoseconds apart may be executed in any order"

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
junrao: > Unless I'm missing something, the event that got prepended to the queue will be immediately run on the next iteration of the loop. There is no opportunity to prepend multiple deferred events to the toRun queue, since only one event can be prepended per loop run, and that event will be immediately run at the top of the loop.

Hmm, after the event is prepended, we continue to the next while loop before dequeuing and setting toRun, right? Then, in the next loop, we could be prepending the next deferred item before the previous one runs.

junrao (commented):
@cmccabe : Thanks for the updated PR. Just one followup comment from previous round.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
jsancio: There is a `continue;` at line 219 so that means that `toRun` is not set in this iteration of the loop did you mean to have this instead: `toRun = eventContext;`

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
```diff
@@ -0,0 +1,420 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Function;
+import org.apache.kafka.common.errors.TimeoutException;
+import org.apache.kafka.common.utils.KafkaThread;
+import org.apache.kafka.common.utils.LogContext;
+import org.apache.kafka.common.utils.Time;
+import org.slf4j.Logger;
+
+
+public final class KafkaEventQueue implements EventQueue {
```
jsancio, line 37: @cmccabe What do you think about splitting this functionality into two types? For example:

1. `EventQueue` is a type which is responsible for ordering events given the the insertion type and deadline. This type is thread-safe but doesn't instantiate thread(s). This type exposes methods for enqueuing and dequeuing events. The dequeuing method(s) can take in a "time" parameter and polls to see if there is an event ready. The dequeue method(s) would need to return the difference between "time" and the next closest event in the queue.
2. `SingleThreadEventExecutor` is a type which spawns a thread to dequeue events from the `EventQueue`, executes the `run` or `handleException` methods of the event and it is `AutoCloseable`.

cmccabe: That's an interesting idea, but I'm not sure I see an advantage for this use-case.  We only want a single thread here-- otherwise we would have to have locking in the controller and in the parts of the broker which use this queue.  So the potential benefit that I can see from your proposal (allowing multiple executors) doesn't really apply here.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
cmccabe: I looked at this again, and yes, you are right: it should just be `toRun = eventContext`.  Fixed.

metadata/src/main/java/org/apache/kafka/queue/EventQueue.java
```diff
@@ -0,0 +1,263 @@
...
+     * @param deadlineNsCalculator  A function which takes as an argument the existing
+     *                              deadline for the event with this tag (or empty if the
+     *                              event has no tag, or if there is none such), and
+     *                              produces the deadline to use for this event.
+     *                              Once the deadline has arrived, the event will be
+     *                              prepended to the queue.  Events whose deadlines are
```
junrao: "the event will be prepended to the queue" : This is no longer true in the implementation.

junrao (approved):
@cmccabe : Thanks for the updated PR. LGTM. Just a minor comment below.

metadata/src/main/java/org/apache/kafka/queue/KafkaEventQueue.java
jsancio: Okay. I suggested it because maybe unittests would be easier to write since the tests would have to deal with concurrency.

cmccabe:
Test failure is `org.apache.kafka.connect.mirror.integration.MirrorConnectorsIntegrationSSLTest` which is not related.

PR #10032, opened 2021-02-03
KAFKA-12270: Handle race condition when Connect tasks attempt to create topics

rhauch:
_**This is option #2 for the fix, which is a bit more ideal since no additional admin client calls are made. See #10027 for option 1 that is simpler but results in additional admin client calls.**_

When a source connector is configured to create missing topics has multiple tasks that generate records for the same topic, it is possible that multiple tasks may simultaneously describe the topic, find it does not exist, and attempt to create the task. One of those create topic requests will succeed, and the other concurrent tasks will receive the response from the topic admin as having not created the task and will fail unnecessarily.

This change corrects the logic by moving the `TopicAdmin` logic to create a topic to a new `createOrFindTopics(…)` method that returns the set of created topic names and the set of existing topic names. This allows the existing `createTopics(…)` and `createTopic(…)` methods to retain the same behavior, but also allows the `WorkerSourceTask` to know from its single call to this new method whether the topic was created or was found to exist.

This adds one unit test and modifies several unit tests in `WorkerSourceTaskWithTopicCreationTest` that use mocks to verify the behavior, and modifies several existing unit tests for `TopicAdminTest` to ensure the logic of the new method is as expected.

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSourceTask.java
@@ -412,10 +412,15 @@ class WorkerSourceTask extends WorkerTask {
         log.debug("Topic '{}' matched topic creation group: {}", topic, topicGroup);
         NewTopic newTopic = topicGroup.newTopic(topic);
 
-        if (admin.createTopic(newTopic)) {
+        TopicAdmin.TopicCreationResponse response = admin.createOrFindTopics(newTopic);
+        if (response.isCreated(newTopic.name())) {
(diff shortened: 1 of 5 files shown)


kkonstantine (approved):
I like this approach. Thanks @rhauch 

LGTM

rhauch:
We're having infrastructure and build issues. I rebased this on the latest from `trunk`, which includes #10038 to disable Travis builds. No other changes were made since approval.

rhauch:
JDK 11 passed, but JDK 8 and 15 each failed on spurious and seemingly unrelated tests (e.g., rebalance failure conditions, timeouts).

PR #10034, opened 2021-02-03
MINOR: Fix windows startup scripts to use connect-log4j.properties instead of tools-log4j.properties

ssugar:
Very small change.  Found that the connect-standalone.bat file uses tools-log4j.properties, but if this is used and no connector properties file is specified, the connect-standalone.bat file will exit with no errors or other output.

Output when connect-standalone.bat uses tools-log4j.properties:

image

If connect-standalone.bat uses connect-log4j.properties, it will return an error message noting that no connector properties file is specified.  This looks to be the correct behaviour.

Output when connect-standalone.bat uses connect-log4j.properties:

image

### DIFF
--- a/bin/windows/connect-distributed.bat
+++ b/bin/windows/connect-distributed.bat
@@ -27,7 +27,7 @@ popd
 
 rem Log4j settings
 IF ["%KAFKA_LOG4J_OPTS%"] EQU [""] (
-	set KAFKA_LOG4J_OPTS=-Dlog4j.configuration=file:%BASE_DIR%/config/tools-log4j.properties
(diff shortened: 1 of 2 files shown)


kkonstantine (commented):
Thanks for surfacing this @ssugar !

Would you mind fixing `./bin/windows/connect-distributed.bat` too? Seems to suffer from the same issue

ssugar:
@kkonstantine  - i've made the same change to the connect-distributed.bat file as well

kkonstantine (approved):
LGTM
Thanks!

PR #10039, opened 2021-02-03
MINOR: Defer log recovery until LogManager startup

rondagostino:
Currently log recovery begins as soon as we instantiate `LogManager`, but when using a Raft-based metadata quorum we won't have configs until after we catch up on the metadata log.  We therefore defer log recovery until we actually invoke `startup()` on the `LogManager` instance.  This timing difference has no effect when using ZooKeeper because we immediately invoke `startup()` on the instantiated instance, but it gives us the necessary flexibility for accurate log recovery with updated configs when using a Raft-based metadata quorum.

The `LogCleaner` is currently instantiated during construction just after log recovery completes, and then it is started in `startup()`.  As an extra precaution, since we are no longer performing recovery during construction, we both instantiate and start the log cleaner in `startup()` after log recovery completes.

We also convert `LogManager` to use a `ConfigRepository` to load topic configs (which can override the default log configs) instead of having a hard-coded dependency on ZooKeeper.  We retrieve the topic configs when we invoke `startup()` -- which again is effectively no different from a timing perspective than what we do today for the ZooKeeper case.

One subtlety is that currently we create the log configs for every topic at this point -- if a topic has no config overrides then we associate a copy of the default configuration with the topic inside a map, and we retrieve the log configs for that topic's partitions from from that map during recovery.  This PR makes a change to this series of events as follows.  We do not associate a copy of the the default configuration with a topic in the map if the topic has no configs set when we query for them.  This saves some memory -- we don't unnecessarily copy the default config many times -- but it also means we have use the default log configs for that topic later on when recovery for each of its partitions begins.

The difference is that the default configs are dynamically reconfigurable, and they could potentially change between the time when we invoke `startup()` and when log recovery begins (log recovery can begin quite some time after `startup()` is invoked if shutdown was unclean).  Prior to this patch such a change would not be used; with this patch they could be if they happen before recovery begins.  This actually is better -- we are performing log recovery with the most recent known defaults when a topic had no overrides at all. Also, `Partition.createLog` has logic to handle missed config updates, so the behavior is eventually the same.

The transition of the broker state from `STARTING` to `RECOVERY` currently happens within the `LogManager`, and it only occurs if the shutdown was unclean.  We move this transition into the broker as it avoids passing a reference to the broker state into the `LogManager`.  We also now always transition the broker into the `RECOVERY` state as dictated by [the KIP-631 broker state machine](https://cwiki.apache.org/confluence/display/KAFKA/KIP-631%3A+The+Quorum-based+Kafka+Controller#KIP631:TheQuorumbasedKafkaController-TheBrokerStateMachine).

Finally, a few clean-ups were included. One worth highlighting is that `Partition` no longer requires a `ConfigRepository`.


core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -402,7 +403,26 @@ class LogManager(logDirs: Seq[File],
...
+
+  // visible for testing
+  private[log] def generateTopicLogConfigs(retrieveTopicNames: () => Set[String]): Map[String, LogConfig] = {
+    val topicLogConfigs: mutable.Map[String, LogConfig] = mutable.Map()
+    val defaultProps = currentDefaultConfig.originals()
+    retrieveTopicNames().foreach { topicName =>
+      val overrides = configRepository.topicConfigs(topicName)
+      val logConfig = LogConfig.fromProps(defaultProps, overrides)
+      topicLogConfigs(topicName) = logConfig
+    }
```
rondagostino: I wonder about the efficiency of this loop when using ZooKeeper.  Today we ask ZooKeeper for all of the topics, then we ask ZooKeeper for all of the configs for all of those topics.  However, under the hood it does appear that `KafkaZkClient.getTopicConfigs(Set)` is iterating through all of the topics and querying each one individually.  So I raise it just to get another pair of eyes to confirm that this explicit loop will not perform worse than what appears to be a single round-trip but is in fact not.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -256,9 +254,10 @@ class LogManager(logDirs: Seq[File],
   private[log] def loadLog(logDir: File,
                       hadCleanShutdown: Boolean,
                       recoveryPoints: Map[TopicPartition, Long],
-                      logStartOffsets: Map[TopicPartition, Long]): Log = {
+                      logStartOffsets: Map[TopicPartition, Long],
+                      topicLogConfigs: Map[String, LogConfig]): Log = {
     val topicPartition = Log.parseTopicPartitionName(logDir)
-    val config = topicConfigs.getOrElse(topicPartition.topic, currentDefaultConfig)
+    val config = topicLogConfigs.getOrElse(topicPartition.topic, currentDefaultConfig)
```
rondagostino: We can ignore any topic that has no explicit configs in `startup()` below since this is doing `.getOrElse()` -- it'll save memory.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -402,7 +403,26 @@ class LogManager(logDirs: Seq[File],
...
+  }
+
+  // visible for testing
+  private[log] def generateTopicLogConfigs(retrieveTopicNames: () => Set[String]): Map[String, LogConfig] = {
+    val topicLogConfigs: mutable.Map[String, LogConfig] = mutable.Map()
+    val defaultProps = currentDefaultConfig.originals()
+    retrieveTopicNames().foreach { topicName =>
+      val overrides = configRepository.topicConfigs(topicName)
+      val logConfig = LogConfig.fromProps(defaultProps, overrides)
+      topicLogConfigs(topicName) = logConfig
```
rondagostino: only add it if there are some overides

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -403,7 +401,30 @@ class LogManager(logDirs: Seq[File],
   /**
    *  Start the background threads to flush logs and do log cleanup
    */
-  def startup(): Unit = {
+  def startup(retrieveTopicNames: => Set[String]): Unit = {
+    startupWithTopicLogConfigOverrides(generateTopicLogConfigs(retrieveTopicNames))
```
mumrah: We are able to load the topic configs right away because they are coming from ZK, right?

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -403,7 +401,30 @@ class LogManager(logDirs: Seq[File],
...
+  }
+
+  // visible for testing
+  private[log] def generateTopicLogConfigs(topicNames: Set[String]): Map[String, LogConfig] = {
+    val topicLogConfigs: mutable.Map[String, LogConfig] = mutable.Map()
```
mumrah: nit: can move type info to right hand side and just have `val topicLogConfigs = ...`

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -403,7 +401,30 @@ class LogManager(logDirs: Seq[File],
   /**
    *  Start the background threads to flush logs and do log cleanup
    */
-  def startup(): Unit = {
+  def startup(retrieveTopicNames: => Set[String]): Unit = {
```
mumrah: Feels slightly odd to pass in the set of topics to load here, but I can't think of a good way to avoid it. Perhaps we could pass MetadataCache into LogManager and let startup call MetadataCache#getAllTopics? That might be more risky though since it changes the startup order in KafkaServer, maybe we can look into this as a follow-up.

Besides that, the name here seems strange. Maybe something like "topicsToLoad"?

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -352,13 +348,15 @@ class LogManager(logDirs: Seq[File],
         val numLogsLoaded = new AtomicInteger(0)
         numTotalLogs += logsToLoad.length
 
-        val jobsForDir = logsToLoad.map { logDir =>
+        val jobsForDir = logsToLoad
+          .filter(logDir => Log.parseTopicPartitionName(logDir).topic != KafkaRaftServer.MetadataTopic)
```
mumrah: Will the metadata topic get passed into LogManager? I would guess not since it's a Raft topic and not a regular Kafka topic. 

Also style nit: you can do `logsToLoad.filter { logDir => ... }` similar to the `map` below

mumrah (commented):
Thanks for the patch @rondagostino! A few comments and questions inline

core/src/main/scala/kafka/log/LogManager.scala
rondagostino: Yes, `startup()` is invoked immediately after the instance is constructed when we are using ZooKeeper, and we can get the configs at that point since we can just query ZK.

core/src/main/scala/kafka/log/LogManager.scala
rondagostino: Would like to keep it as a verb: `retrieveTopicsToLoad`

rondagostino: Actually, this isn't the list of topics to load -- it's the list of topics for which we will look for log config overrides.  `retrieveTopicsForLogConfigOverrideCheck`

core/src/main/scala/kafka/log/LogManager.scala
rondagostino: Yeah, we use the list of topics just to know what to check for log config overrides.  This will see anything in the log dir.  Now that I realize it, we could probably use this to know what topics we need to look at for log config overrides.  But I don't want to change too much at this point.

rondagostino: We could get rid of the parameter in `startup()` and create a `TopicLogConfigOverrideRetriever` class that held the `ConfigRepository` and would retrieve any topic configs in a thread-safe manner -- then we just pass that to `loadLog()` and it asks for the overrides, and the class would retrieve iff they haven't been retrieved before.  WDYT?

rondagostino: I suspect we shouldn’t load topic configs on-demand because it would change the current behavior.  Specifically, if we can’t load the configs, currently the object instantiation will raise an exception.  The current patch will cause the exception to be raised in startup(), which is effectively the same thing.  But if we load them on-demand then both instantiation and startup() will succeed, and the exception will occur in a thread doing recovery.

core/src/main/scala/kafka/log/LogManager.scala
```diff
@@ -118,8 +116,6 @@ class LogManager(logDirs: Seq[File],
     logDirsSet
   }
 
-  loadLogs()
```
ijuma, line 121: Looks like we need to look if it's still safe for the things after this to be run in the constructor. Perhaps the cleaner should be started after the logs are loaded?

rondagostino: Good point.  Yes, it seems safer to delay starting anything until we actually have recovered.

core/src/test/scala/unit/kafka/cluster/PartitionLockTest.scala
```diff
@@ -305,15 +304,6 @@ class PartitionLockTest extends Logging {
     partition
   }
 
-  private def createLogProperties(overrides: Map[String, String]): Properties = {
```
ijuma: The tests pass without this, but we may want to pass these to the LogManager.

ijuma:
One unrelated flaky failure:
`Build / JDK 15 / kafka.controller.ControllerIntegrationTest.testPartitionReassignmentToBrokerWithOfflineLogDir()`

One of the builds exited early, which is unrelated too. To be sure, restarted the tests.

ijuma (approved):
LGTM after the latest updates. @rondagostino Let me know if you're good with the changes I pushed.

rondagostino:
Latest changes LGTM

ijuma:
The JDK 8 build had unrelated flaky test failures this time while JDK 11 and 15 terminated before completion (again unrelated). Will go ahead and merge.

ijuma:
Merged to trunk and 2.8 branches.

PR #10040, opened 2021-02-03
KAFKA-12259: Use raw config to infer connector type when returning connect status response

Cyril-Engels:
Addressed [KAFKA-12259](https://issues.apache.org/jira/browse/KAFKA-12259).

Problem: when calling the connect status endpoint, a 500 error is returned, e.g.
```
{
  "error_code": 500,
  "message": "Could not read properties from file /tmp/somefile.properties"
}
```
when any of the connector's has an exception from the config provider. This is because the `AbstractHerder` is trying to use resolved config. However, only the `connector.class` is needed from the config to infer if this connector is of source or sink type. The endpoint should still return the status of the connector instead of a 500 error.

This change uses raw config from the config backing store to retrieve the connector class to avoid any variable resolution.

Unit tests have been updated to reflect this change.

when any of the connector's has an exception from the config provider. 

### DIFF
--- a/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/AbstractHerder.java
+++ b/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/AbstractHerder.java
@@ -229,9 +229,9 @@ public abstract class AbstractHerder implements Herder, TaskStatus.Listener, Con
     }
 
     /*
-     * Retrieves config map by connector name
+     * Retrieves raw config map by connector name.
      */
-    protected abstract Map<String, String> config(String connName);
+    protected abstract Map<String, String> rawConfig(String connName);
 
     @Override
     public void connectorConfig(String connName, Callback<Map<String, String>> callback) {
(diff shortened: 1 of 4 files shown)


connect/runtime/src/main/java/org/apache/kafka/connect/runtime/AbstractHerder.java
```diff
@@ -284,7 +284,8 @@ public ConnectorStateInfo connectorStatus(String connName) {
 
         Collections.sort(taskStates);
 
-        Map<String, String> conf = config(connName);
+        final ClusterConfigState configState = configBackingStore.snapshot();
```
kkonstantine: I'm suspecting we might be able to skip the `snapshot` here. Since we are asking for a specific connector and since we check the status topic before, it's highly probable that the connector config is already in the `configState` that has been snapshotted in distributed or standalone herders. 

Should we just change the implementations of `AbstractHerder#config` to return the raw config?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/AbstractHerder.java
```diff
@@ -284,7 +284,8 @@ public ConnectorStateInfo connectorStatus(String connName) {
 
         Collections.sort(taskStates);
 
-        Map<String, String> conf = config(connName);
```
kkonstantine, line 287: we are removing the single use of this method. 

Either we should remove it completely (including its implementations) or we could choose to extend it to return the raw config upon request. Eg. change it to `protected abstract Map<String, String> config(String connName, boolean returnRawConfig);` 

`config(connName, true)` still won't have a use but maybe that's an ok exception for now. Or we can change it to `rawConfig(String connName)`. 

See my comment below about whether we need to create a shallow copy of the config cluster state with the `snapshot` method.

kkonstantine (commented):
Thanks @Cyril-Engels. That's a nice fix. 
Added a few comments to suggest a couple of small adjustments

kkonstantine (approved):
LGTM!

PR #10041, opened 2021-02-03
MINOR: Prevent creating partition.metadata until ID can be written

jolshan:
Currently the partition.metadata file is created when the log is created. However, clusters with older inter-broker protocols will never use this file. This PR moves the creation of the file to when we write to the file. 

This PR also deletes the partition.metadata file on startup if the IBP version is lower than 2.8. 

I will be looking at benchmarks to see how these changes affect LISR request processing.


core/src/main/scala/kafka/log/Log.scala
```diff
@@ -256,7 +256,8 @@ class Log(@volatile private var _dir: File,
           val topicPartition: TopicPartition,
           val producerStateManager: ProducerStateManager,
           logDirFailureChannel: LogDirFailureChannel,
-          private val hadCleanShutdown: Boolean = true) extends Logging with KafkaMetricsGroup {
+          private val hadCleanShutdown: Boolean = true,
+          val usesTopicId: Boolean = true) extends Logging with KafkaMetricsGroup {
```
junrao: Could we add the new param to javadoc? Also, will keepPartitionMetdataFile be better than usesTopicId?

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -341,10 +342,15 @@ class Log(@volatile private var _dir: File,
...
 
-    // Recover topic ID if present
+    // Delete partition metadata file if the version does not support topic IDs.
+    // Recover topic ID if present and topic IDs are supported
     partitionMetadataFile.foreach { file =>
-      if (!file.isEmpty())
-        topicId = file.read().topicId
+      if (file.exists()) {
+        if (!usesTopicId)
+          file.delete()
```
junrao: Does partitionMetadataFile need to be of Some? It seems that we can just always instantiate the object.

core/src/main/scala/kafka/server/PartitionMetadataFile.scala
```diff
@@ -91,11 +91,10 @@ class PartitionMetadataFile(val file: File,
   private val lock = new Object()
   private val logDir = file.getParentFile.getParent
 
-
-  try Files.createFile(file.toPath) // create the file if it doesn't exist
-  catch { case _: FileAlreadyExistsException => }
-
   def write(topicId: Uuid): Unit = {
+    try Files.createFile(file.toPath) // create the file if it doesn't exist
```
junrao: Do we need to create the file first? It seems that later on we always rename the temp file to this one.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
```diff
@@ -217,6 +217,7 @@ class LogManagerTest {
     }
     assertTrue(log.numberOfSegments > 1, "There should be more than one segment now.")
     log.updateHighWatermark(log.logEndOffset)
+    log.partitionMetadataFile.get.write(Uuid.randomUuid())
```
junrao: Is this needed? It seems Log never reads UUID? Ditto below.

junrao (commented):
@jolshan : Thanks for the PR. A few comments below.

core/src/main/scala/kafka/log/Log.scala
jolshan: I named this based on the config's name. But the name you suggested is more descriptive.

core/src/main/scala/kafka/log/Log.scala
jolshan: I modeled this off of `leaderEpochCache` which is also an option. But I see how that is different now.

core/src/test/scala/unit/kafka/log/LogManagerTest.scala
jolshan: I wanted to test the correct files are kept/deleted. I could also change the math in line 234 if we don't need this test.

junrao: Got it. Since the partitionMetadataFile is now created on demand, perhaps we could just change the math on the expected number of files?

jolshan: So remove the creation and subtract one file?

junrao: That's what I am thinking. It's kind of weird to add the new file just to get the existing math work.

core/src/main/scala/kafka/log/Log.scala
```diff
@@ -256,7 +256,8 @@ class Log(@volatile private var _dir: File,
           val topicPartition: TopicPartition,
           val producerStateManager: ProducerStateManager,
           logDirFailureChannel: LogDirFailureChannel,
-          private val hadCleanShutdown: Boolean = true) extends Logging with KafkaMetricsGroup {
+          private val hadCleanShutdown: Boolean = true,
+          val keepPartitionMetadataFile: Boolean = true) extends Logging with KafkaMetricsGroup {
```
junrao, line 266: Could we add the new param to the javadoc? In the javadoc, it would be useful to explain a bit how this helps with re-upgrade.

junrao (commented):
@jolshan : Thanks for the updated PR. Just one more comment below.

core/src/main/scala/kafka/log/Log.scala
jolshan: good point. will do!

jolshan: Something like this work for an explanation?
```
* boolean flag to indicate whether the partition.metadata file should be kept in the 
* log directory. A partition.metadata file is only created when the controller's 
* inter-broker protocol version is at least 2.8. This file will persist the topic ID on
* the broker. If inter-broker protocol is downgraded below 2.8, a topic ID may be lost
* and a new ID generated upon re-upgrade. If the inter-broker protocol version is below
* 2.8, partition.metadata will be deleted to avoid ID conflicts upon re-upgrade. 
 ```

Comment (junrao) [reply core/src/main/scala/kafka/log/Log.scala:266]: Yes, looks good to me.

**Review by junrao — APPROVED** [review]
@jolshan : Thanks for the latest PR. LGTM

Are the test failure related? Also, I guess you want this PR in the 2.8 branch too?

**Comment by jolshan** [comment]
Yes. I will want this on the 2.8 branch. I'll check the failed tests. I've also had trouble with building at least one of the three JDKs, but it seems like it is not the same one each time.

**Comment by jolshan** [comment]
`MirrorConnectorsIntegrationSSLTest.testOneWayReplicationWithAutoOffsetSync()` seems to be flaky (is failing on other open PRs) It passed when I ran locally.
`ConnectionQuotasTest.testListenerConnectionRateLimitWhenActualRateAboveLimit()` seems to be unrelated. I also ran locally and it passed.

**Comment by junrao** [comment]
@jolshan : Thanks. For those transient failures, could you file a jira if it's not tracked already?

Are the JDK 8 failures also transient?

**Comment by jolshan** [comment]
I had 2 JDK 11 failures previously. I also noticed JDK failures on other PRs. I'll check JIRA for these issues

**Comment by jolshan** [comment]
Looks like one was already created.
https://issues.apache.org/jira/browse/KAFKA-12284
https://issues.apache.org/jira/browse/KAFKA-12319

**Comment by junrao** [comment]
@jolshan : Thanks. Were the JDK 8 tests ok too?

**Comment by jolshan** [comment]
@junrao On the previous commit JDK8 built and only MirrorConnectorsIntegrationSSLTest.testOneWayReplicationWithAutoOffsetSync() failed. (The only difference between this commit and the most recent was the javadoc change. Before that, all JDK8 tests passed.

**Comment by junrao** [comment]
@jolshan Could you rebase the PR for trunk? Also, I am not sure if the PR ports to 2.8 cleanly, if not, could you submit a separate PR for 2.8? Thanks.

**Comment by junrao** [comment]
@jolshan : Do you have any benchmark results that you want to share?

**Comment by jolshan** [comment]
Sure. Using the LeaderAndIsrBenchmark (and the async profiler) [LISRbench.zip](https://github.com/apache/kafka/files/5961818/LISRbench.zip)
 from https://github.com/apache/kafka/pull/10071 I had these results:
```
Benchmark                                                          (partitionCount)  (topicCount)  Mode  Cnt        Score       Error  Units
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       10            10  avgt   15    50920.020 ±  6450.121  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       10            20  avgt   15    81505.310 ±  3821.451  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       10           100  avgt   15   451907.331 ±  2157.775  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       20            10  avgt   15    75708.157 ±  4316.357  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       20            20  avgt   15   136747.799 ±  3886.797  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       20           100  avgt   15   980846.811 ±  4017.757  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       50            10  avgt   15   156896.027 ±  4952.014  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       50            20  avgt   15   318294.323 ±  4729.402  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       50           100  avgt   15  2904827.652 ± 74264.592  ns/op
```
I took a look at the flame graph for these tests and the file I/O impact seemed to be very minimal

This was trunk before the changes:
```
Benchmark                                                          (partitionCount)  (topicCount)  Mode  Cnt        Score        Error  Units
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       10            10  avgt   15    53887.240 ±   5645.067  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       10            20  avgt   15    83015.916 ±   3560.633  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       10           100  avgt   15   481109.947 ±  18767.713  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       20            10  avgt   15    75932.544 ±   6528.281  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       20            20  avgt   15   148323.073 ±   6009.839  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       20           100  avgt   15  1146604.372 ±  54073.673  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       50            10  avgt   15   171529.923 ±   9940.990  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       50            20  avgt   15   327926.279 ±   9292.803  ns/op
LeaderAndIsrRequestBenchmark.testHandleLeaderAndIsrRequest                       50           100  avgt   15  3512983.407 ± 570652.231  ns/op
```

junrao:
@jolshan : It seems this PR can't be applied cleanly in 2.8. Could you submit a separate PR for 2.8? Thanks.

jolshan:
@junrao  https://github.com/apache/kafka/pull/10100

PR #10042, opened 2021-02-03
KAFKA-9527: fix NPE when using time-based argument for Stream Resetter

MarcoLotz:
**Description:**
A NPE is happening when using the StreamResetter tool with the arguments "to-timedate" and "by-duration" on empty partitions.

**Cause:**
This happens because [fetcher](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/Fetcher.java#L516) returns a null value on the map for empty partitions.

**Solution:**
The null values needs to be correctly handled with Optional. In the case of empty partition, consumer.seek() should not be called - letting it fallback to the "auto.offset.reset" behaviour. The user should be notified by it.

**Implementation comments:**

- System.out.println was used instead of log.warn - since the whole class uses syso for interacting with the user.
- There was no test for empty partition on reset to offset [streamsResetter.resetOffsetsTo()]. Test was added.
- Added test for empty partition when resetToDatetime() is called - that was the cause of the bug.
- MockConsumer doesn't have an implementation of offsetsForTimes() method - it throws an exception when called, breaching Liskov Substitution Principle. This method is used by resetToDatetime(). Since I couldn't find Mockito in the project dependencies to make a Spy, I decided to extends the MockConsumer class as a nested class and especialize the method.
- Looking at Gradle, I saw that the min. java version is 1.8 - thus is decided to use 1.8 implementation of Optional instead of 1.9 - which would be a lot cleaner specially for the Optional.ProcessOrElse(). 

### DIFF
--- a/core/src/main/scala/kafka/tools/StreamsResetter.java
+++ b/core/src/main/scala/kafka/tools/StreamsResetter.java
@@ -35,6 +35,7 @@ import org.apache.kafka.clients.consumer.OffsetAndTimestamp;
 import org.apache.kafka.common.KafkaFuture;
 import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.annotation.InterfaceStability;
+import org.apache.kafka.common.requests.ListOffsetsResponse;
 import org.apache.kafka.common.serialization.ByteArrayDeserializer;
 import org.apache.kafka.common.utils.Exit;
 import org.apache.kafka.common.utils.Utils;
@@ -52,6 +53,7 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
@@ -492,9 +494,10 @@ public class StreamsResetter {
         resetToDatetime(client, inputTopicPartitions, Instant.now().minus(duration).toEpochMilli());
     }
 
-    private void resetToDatetime(final Consumer<byte[], byte[]> client,
-                                 final Set<TopicPartition> inputTopicPartitions,
-                                 final Long timestamp) {
+    // visible for testing
+    public void resetToDatetime(final Consumer<byte[], byte[]> client,
+                                final Set<TopicPartition> inputTopicPartitions,
+                                final Long timestamp) {
         final Map<TopicPartition, Long> topicPartitionsAndTimes = new HashMap<>(inputTopicPartitions.size());
         for (final TopicPartition topicPartition : inputTopicPartitions) {
             topicPartitionsAndTimes.put(topicPartition, timestamp);
@@ -503,7 +506,16 @@ public class StreamsResetter {
         final Map<TopicPartition, OffsetAndTimestamp> topicPartitionsAndOffset = client.offsetsForTimes(topicPartitionsAndTimes);
 
         for (final TopicPartition topicPartition : inputTopicPartitions) {
-            client.seek(topicPartition, topicPartitionsAndOffset.get(topicPartition).offset());
+            final Optional<Long> partitionOffset = Optional.ofNullable(topicPartitionsAndOffset.get(topicPartition))
+                    .map(OffsetAndTimestamp::offset)
+                    .filter(offset -> offset != ListOffsetsResponse.UNKNOWN_OFFSET);
+            if (partitionOffset.isPresent()) {
+                client.seek(topicPartition, partitionOffset.get());
+            } else {
+                client.seekToEnd(Collections.singletonList(topicPartition));
+                System.out.println("Partition " + topicPartition.partition() + " from topic " + topicPartition.topic() +
+                        " is empty, without a committed record. Falling back to latest known offset.");
+            }
         }
     }
 
--- a/streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
@@ -21,6 +21,7 @@ import org.apache.kafka.clients.admin.MockAdminClient;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.clients.consumer.MockConsumer;
+import org.apache.kafka.clients.consumer.OffsetAndTimestamp;
 import org.apache.kafka.clients.consumer.OffsetResetStrategy;
 import org.apache.kafka.common.Cluster;
 import org.apache.kafka.common.Node;
@@ -31,6 +32,7 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.time.Duration;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -74,6 +76,28 @@ public class StreamsResetterTest {
 
         final ConsumerRecords<byte[], byte[]> records = consumer.poll(Duration.ofMillis(500));
         assertEquals(3, records.count());
+    }
+
+    @Test
+    public void testResetOffsetToSpecificOffsetWhenAfterEndOffset() {
+        final long beginningOffset = 5L;
+        final long endOffset = 10L;
+        final MockConsumer<byte[], byte[]> emptyConsumer = new MockConsumer<>(OffsetResetStrategy.EARLIEST);
+        emptyConsumer.assign(Collections.singletonList(topicPartition));
+
+        final Map<TopicPartition, Long> beginningOffsetsMap = new HashMap<>();
+        beginningOffsetsMap.put(topicPartition, beginningOffset);
+        emptyConsumer.updateBeginningOffsets(beginningOffsetsMap);
+
+        final Map<TopicPartition, Long> endOffsetsMap = new HashMap<>();
+        endOffsetsMap.put(topicPartition, endOffset);
+        emptyConsumer.updateEndOffsets(endOffsetsMap);
+        // resetOffsetsTo only seeks the offset, but does not commit.
+        streamsResetter.resetOffsetsTo(emptyConsumer, inputTopicPartitions, endOffset + 2L);
+
+        final long position = emptyConsumer.position(topicPartition);
+
+        assertEquals(endOffset, position);
     }
 
     @Test
@@ -247,6 +271,29 @@ public class StreamsResetterTest {
         assertTrue(streamsResetter.matchesInternalTopicFormat("appId-KTABLE-FK-JOIN-SUBSCRIPTION-REGISTRATION-12323232-topic"));
     }
 
+    @Test
+    public void testResetToDatetimeWhenPartitionIsEmptyResetsToLatestOffset() {
+        final long beginningAndEndOffset = 5L; // Empty partition implies beginning offset == end offset
+        final MockConsumer<byte[], byte[]> emptyConsumer = new EmptyPartitionConsumer<>(OffsetResetStrategy.EARLIEST);
+        emptyConsumer.assign(Collections.singletonList(topicPartition));
+
+        final Map<TopicPartition, Long> beginningOffsetsMap = new HashMap<>();
+        beginningOffsetsMap.put(topicPartition, beginningAndEndOffset);
+        emptyConsumer.updateBeginningOffsets(beginningOffsetsMap);
+
+        final Map<TopicPartition, Long> endOffsetsMap = new HashMap<>();
+        endOffsetsMap.put(topicPartition, beginningAndEndOffset);
+        emptyConsumer.updateEndOffsets(endOffsetsMap);
+
+        final long yesterdayTimestamp = Instant.now().minus(Duration.ofDays(1)).toEpochMilli();
+        // resetToDatetime only seeks the offset, but does not commit.
+        streamsResetter.resetToDatetime(emptyConsumer, inputTopicPartitions, yesterdayTimestamp);
+
+        final long position = emptyConsumer.position(topicPartition);
+
+        assertEquals(beginningAndEndOffset, position);
+    }
+
     private Cluster createCluster(final int numNodes) {
         final HashMap<Integer, Node> nodes = new HashMap<>();
         for (int i = 0; i < numNodes; ++i) {
@@ -255,6 +302,20 @@ public class StreamsResetterTest {
         return new Cluster("mockClusterId", nodes.values(),
             Collections.<PartitionInfo>emptySet(), Collections.<String>emptySet(),
             Collections.<String>emptySet(), nodes.get(0));
+    }
+
+    private static class EmptyPartitionConsumer<K, V> extends MockConsumer<K, V> {
+
+        public EmptyPartitionConsumer(final OffsetResetStrategy offsetResetStrategy) {
+            super(offsetResetStrategy);
+        }
+
+        @Override
+        public synchronized Map<TopicPartition, OffsetAndTimestamp> offsetsForTimes(final Map<TopicPartition, Long> timestampsToSearch) {
+            final Map<TopicPartition, OffsetAndTimestamp> topicPartitionToOffsetAndTimestamp = new HashMap<>();
+            timestampsToSearch.keySet().forEach(k -> topicPartitionToOffsetAndTimestamp.put(k, null));
+            return topicPartitionToOffsetAndTimestamp;
+        }
     }
 
 }


core/src/main/scala/kafka/tools/StreamsResetter.java
```diff
@@ -503,7 +505,15 @@ private void resetToDatetime(final Consumer<byte[], byte[]> client,
         final Map<TopicPartition, OffsetAndTimestamp> topicPartitionsAndOffset = client.offsetsForTimes(topicPartitionsAndTimes);
 
         for (final TopicPartition topicPartition : inputTopicPartitions) {
-            client.seek(topicPartition, topicPartitionsAndOffset.get(topicPartition).offset());
+            final Optional<Long> partitionOffset = Optional.ofNullable(topicPartitionsAndOffset.get(topicPartition))
+                    .map(OffsetAndTimestamp::offset);
```
jeqo: just in case, latest consumer-group reset-offset is comparing against `ListOffsetsResponse.UNKNOWN_OFFSET` instead of `null`.

https://github.com/apache/kafka/blob/0bc394cc1d19f1e41dd6646e9ac0e09b91fb1398/core/src/main/scala/kafka/admin/ConsumerGroupCommand.scala#L680-L681

MarcoLotz: You are right. ListOffsetsResponse.UNKNOWN_OFFSET value is -1. Clearly it's not enough to check if it is null (the cause of the NPE bug) but also we need to make sure that it is not ListOffsetsResponse.UNKNOWN_OFFSET value.

I have added a filter to the Optional.

MarcoLotz:
@jeqo changed to seekLatest and aligned with UNKOWN_OFFSET recommendation.

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
```diff
@@ -247,6 +268,26 @@ public void shouldDetermineInternalTopicBasedOnTopicName1() {
         assertTrue(streamsResetter.matchesInternalTopicFormat("appId-KTABLE-FK-JOIN-SUBSCRIPTION-REGISTRATION-12323232-topic"));
     }
 
+    @Test
+    public void emptyPartitionsAreCorrectlyHandledWhenResettingByDateAndTime() {
+        final MockConsumer<byte[], byte[]> emptyConsumer = new EmptyPartitionConsumer<>(OffsetResetStrategy.EARLIEST);
+        final Duration yesterday = Duration.ofDays(1);
```
jeqo: nit: could we move this initialization closer to its usage in L285? or just `.minus(Duration.ofDays(1))` could be simpler.

core/src/main/scala/kafka/tools/StreamsResetter.java
```diff
@@ -503,7 +506,16 @@ private void resetToDatetime(final Consumer<byte[], byte[]> client,
...
-            client.seek(topicPartition, topicPartitionsAndOffset.get(topicPartition).offset());
+            final Optional<Long> partitionOffset = Optional.ofNullable(topicPartitionsAndOffset.get(topicPartition))
+                    .map(OffsetAndTimestamp::offset)
+                    .filter(offset -> offset != ListOffsetsResponse.UNKNOWN_OFFSET);
+            if (partitionOffset.isPresent()) {
+                client.seek(topicPartition, partitionOffset.get());
+            } else {
+                client.seekToEnd(Collections.singletonList(topicPartition));
+                System.out.println("Partition " + topicPartition.partition() + " from topic " + topicPartition.topic() +
+                        " is empty, without a committed offset. Falling back to latest offset.");
```
jeqo: nit:
```suggestion
                        " is empty, without a committed record. Falling back to latest known offset.");
```

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
MarcoLotz: I've moved to the line just before L285. I would rather keep it there instead of provide it straight on line L285. Keeping it on a variable makes the //given //when //then more readable then providing straight in the //when call.

jeqo (approved):
LGTM, thanks @MarcoLotz !

core/src/main/scala/kafka/tools/StreamsResetter.java
```diff
@@ -503,7 +506,16 @@ private void resetToDatetime(final Consumer<byte[], byte[]> client,
...
-            client.seek(topicPartition, topicPartitionsAndOffset.get(topicPartition).offset());
+            final Optional<Long> partitionOffset = Optional.ofNullable(topicPartitionsAndOffset.get(topicPartition))
+                    .map(OffsetAndTimestamp::offset)
+                    .filter(offset -> offset != ListOffsetsResponse.UNKNOWN_OFFSET);
+            if (partitionOffset.isPresent()) {
+                client.seek(topicPartition, partitionOffset.get());
+            } else {
+                client.seekToEnd(Collections.singletonList(topicPartition));
+                System.out.println("Partition " + topicPartition.partition() + " from topic " + topicPartition.topic() +
+                        " is empty, without a committed record. Falling back to latest known offset.");
```
mjsax, line 517: > without a committed record

Not sure what this means? Does `is empty` need any clarification?

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
```diff
@@ -76,6 +78,25 @@ public void testResetToSpecificOffsetWhenBetweenBeginningAndEndOffset() {
...
+        emptyConsumer.updateEndOffsets(endOffsets);
+
+        final Map<TopicPartition, Long> beginningOffsets = new HashMap<>();
+        beginningOffsets.put(topicPartition, 0L);
+        emptyConsumer.updateBeginningOffsets(beginningOffsets);
+
+        streamsResetter.resetOffsetsTo(emptyConsumer, inputTopicPartitions, 2L);
+
+        final ConsumerRecords<byte[], byte[]> records = emptyConsumer.poll(Duration.ofMillis(500));
+        assertEquals(0, records.count());
```
mjsax: Not sure if I understand this test. As we use a `MockConsumer` and we never call `addRecord()` this condition should be `true` independent of `StreamsResetter`.

Should we not rather verify if `streamsResetter` did _commit_ offsets as expected?

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
```diff
@@ -76,6 +78,25 @@ public void testResetToSpecificOffsetWhenBetweenBeginningAndEndOffset() {
         assertEquals(3, records.count());
     }
 
+    @Test
+    public void testResetToSpecificOffsetWhenPartitionIsEmpty() {
```
mjsax: This test seems to be orthogonal to the fix. Was just wondering why we add it and what it's purpose is? It's always good to close testing gaps, just not sure if I understand what this test really tests?

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
```diff
@@ -247,6 +268,27 @@ public void shouldDetermineInternalTopicBasedOnTopicName1() {
...
+        final Map<TopicPartition, Long> beginningOffsets = new HashMap<>();
+        beginningOffsets.put(topicPartition, 0L);
+        emptyConsumer.updateBeginningOffsets(beginningOffsets);
+
+        final long yesterdayTimestamp = Instant.now().minus(Duration.ofDays(1)).toEpochMilli();
+
+        streamsResetter.resetToDatetime(emptyConsumer, inputTopicPartitions, yesterdayTimestamp);
+
+        final ConsumerRecords<byte[], byte[]> records = emptyConsumer.poll(Duration.ofMillis(500));
+        assertEquals(0, records.count());
```
mjsax: Some comment as above.

mjsax (commented):
Thanks for the PR. Couple of question on the tests.

core/src/main/scala/kafka/tools/StreamsResetter.java
MarcoLotz: I see your point, I don't mind removing "without a committed record" part of the message. @jeqo This would have to be updated on the scala code too, since I saw that the messages are about the same. Would that be ok?

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
MarcoLotz: You are right - indeed the test is orthogonal to the fix. I have decided to add it there due to the lack of any test scenario for reseting partition offset (to a specific offset) on an empty partition. The original CUT passed this test without the bug-fix. I considered an important scenario to test.

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
MarcoLotz: @mjsax I guess got your point. In the CUT the call ```streamsResetter.resetOffsetsTo(emptyConsumer, inputTopicPartitions, 2L);``` makes the consumer client to seek that offset (if available) without a commit. The call to commit offset happens in another section of the code that is not under test there (line 407 of StreamsResetter.java).

I will change the test logic so that the "given condition" uses ```streamsResetter.resetOffsetsTo(emptyConsumer, inputTopicPartitions, 2L);``` and "when condition" is actually the call of ```client.commitSync();```.

This way, the "then condition" would be able to test using the committed offsets. Same applies for the other test.

MarcoLotz: @mjsax I have changed the PR to make sure that the tests verify the committed offset. It seems to me, however, that the original test implementer intent was to verify the position of the consumer after the method call - not if it committed the offset. Specially because the ```position()``` and ```commitAsync()``` calls happen outside the methods under test.

MarcoLotz:
@mjsax can you please have a quick look on the updated tests?

core/src/main/scala/kafka/tools/StreamsResetter.java
mjsax: If @jeqo has not objections, I think we should remove it for both cases.

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
```diff
@@ -76,6 +79,30 @@ public void testResetToSpecificOffsetWhenBetweenBeginningAndEndOffset() {
...
+        endOffsets.put(topicPartition, 0L);
+        emptyConsumer.updateEndOffsets(endOffsets);
+
+        final Map<TopicPartition, Long> beginningOffsets = new HashMap<>();
+        beginningOffsets.put(topicPartition, 0L);
+        emptyConsumer.updateBeginningOffsets(beginningOffsets);
+        streamsResetter.resetOffsetsTo(emptyConsumer, inputTopicPartitions, 2L);
+        emptyConsumer.position(topicPartition);
+
+        emptyConsumer.commitSync();
```
mjsax: Thanks for explaining on the other comment that `resetOffsetsTo` only seeks but does not commit.

For this test, I am wondering if we can only verify if the seek happened without calling commit in the test code? This may make the test "cleaner". If fact, `emptyConsumer.position(topicPartition);` should return the seek position and it seems sufficient to verify its return value?

I also hope we actually have a "integration test" that tests that both seek and commit happens when using the resetting (ie, an full test for this case instead of just a partial helper-method test).

We might also add a comment that the helper method `resetOffsetsTo` does only seeks but not commits (for others to make sense of the test more easily).

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
```diff
@@ -76,6 +79,30 @@ public void testResetToSpecificOffsetWhenBetweenBeginningAndEndOffset() {
         assertEquals(3, records.count());
     }
 
+    @Test
+    public void testResetToSpecificOffsetWhenPartitionIsEmptyResetsToLatestOffset() {
+        final MockConsumer<byte[], byte[]> emptyConsumer = new MockConsumer<>(OffsetResetStrategy.EARLIEST);
+        emptyConsumer.assign(Collections.singletonList(topicPartition));
+
+        final Map<TopicPartition, Long> endOffsets = new HashMap<>();
+        endOffsets.put(topicPartition, 0L);
```
mjsax: Nit: To make sure we don't have any default/fall-back offset of zero encoded anywhere, it might be better to test with different offsets values for endOffset/beginningOffset and the target offset?

Atm, if we would `seekToBeginning` as fallback instead of `seektToEnd` this test would still pass. Maybe best to just use 5, 10, 20 (or similar) for start, end, target.

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
```diff
@@ -247,6 +274,32 @@ public void shouldDetermineInternalTopicBasedOnTopicName1() {
         assertTrue(streamsResetter.matchesInternalTopicFormat("appId-KTABLE-FK-JOIN-SUBSCRIPTION-REGISTRATION-12323232-topic"));
     }
 
+    @Test
+    public void testResetToDatetimeWhenPartitionIsEmptyResetsToLatestOffset() {
```
mjsax, line 275: Same comments as for the first test.

mjsax (commented):
Thanks for updating the PR. A few more suggestions to improve the tests.

core/src/main/scala/kafka/tools/StreamsResetter.java
jeqo: "empty" should be clear enough. Will update it in #10092

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
MarcoLotz: Updated the code. I've confirmed that there are IT tests for it at ResetIntegrationTest.java.

MarcoLotz:
@mjsax based on your points I have updated the tests. The new tests verify the seek operations and do test multiple offsets scenarios, to ensure no hardcoded zero offset.

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
```diff
@@ -76,6 +78,35 @@ public void testResetToSpecificOffsetWhenBetweenBeginningAndEndOffset() {
         assertEquals(3, records.count());
     }
 
+    @Test
+    public void testResetOffsetToSpecificOffsetWhenAfterEndOffset() {
+        final long[] beginningOffsets = {0L, 5L, 10L, 20L};
+        final long[] endOffsets = {0L, 5L, 10L, 20L};
+        // Test on multiple beginning and end offset combinations
+        for (int beginningOffsetIndex = 0; beginningOffsetIndex < beginningOffsets.length; beginningOffsetIndex++) {
```
mjsax: I actually think that a single combination beginning=5 end=10 should be sufficient.

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
```diff
@@ -247,6 +278,32 @@ public void shouldDetermineInternalTopicBasedOnTopicName1() {
         assertTrue(streamsResetter.matchesInternalTopicFormat("appId-KTABLE-FK-JOIN-SUBSCRIPTION-REGISTRATION-12323232-topic"));
     }
 
+    @Test
+    public void testResetToDatetimeWhenPartitionIsEmptyResetsToLatestOffset() {
+        final long[] beginningAndEndOffsets = {0L, 5L, 10L, 20L};
```
mjsax: So ensure that we really set to end, beginning and end should be set to different values, otherwise, we cannot distinguish both. Thus, similar to the test above, testing a single combination beginning=5 and end=10 seems sufficient?

MarcoLotz: On this one, seems to me that only one value is sufficient. As the test describes `testResetToDatetimeWhenPartitionIsEmptyResetsToLatestOffset`, the partition is empty - thus beginning offset == end offset. I will change to have a beginning offset equals to 5 anyways.

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
MarcoLotz: Sure. Will simplify then

MarcoLotz:
updated the PR and rebased with trunk. Build seems to be failing on trunk due to RocksDBTimeOrderedWindowStore.java

streams/src/test/java/org/apache/kafka/streams/tools/StreamsResetterTest.java
mjsax: Fair enough :)

mjsax:
Not your fault... `trunk` is broken: https://github.com/apache/kafka/pull/10517

mjsax:
Hotfix merged. Retriggered Jenkins for this PR.

MarcoLotz:
@mjsax thanks! Rebased with the hotfix. Now should build as expected.

mjsax:
Thanks for the fix @MarcoLotz! -- Merged to `trunk`.

PR #10043, opened 2021-02-03
MINOR: Add StorageTool as specified in KIP-631

cmccabe:
Add StorageTool as specified in KIP-631.  It can format and describe storage directories.

### DIFF
--- a/core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
+++ b/core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
@@ -88,9 +88,9 @@ class RawMetaProperties(val props: Properties = new Properties()) {
   }
 
   override def toString: String = {
-    "RawMetaProperties(" + props.keySet().asScala.toList.asInstanceOf[List[String]].sorted.map {
+    "{" + props.keySet().asScala.toList.asInstanceOf[List[String]].sorted.map {
       key => key + "=" + props.get(key)
-    }.mkString(", ") + ")"
+    }.mkString(", ") + "}"
   }
 }
 
@@ -130,7 +130,7 @@ case class ZkMetaProperties(
   }
 
   override def toString: String = {
-    s"LegacyMetaProperties(brokerId=$brokerId, clusterId=$clusterId)"
+    s"ZkMetaProperties(brokerId=$brokerId, clusterId=$clusterId)"
   }
 }
 
--- a/core/src/main/scala/kafka/tools/StorageTool.scala
+++ b/core/src/main/scala/kafka/tools/StorageTool.scala
@@ -158,8 +158,11 @@ object StorageTool extends Logging {
       0
     } else {
       if (foundDirectories.nonEmpty) {
-        stream.println("Found log director%s:".format(
-          if (foundDirectories.size == 1) "y" else "ies"))
+        if (foundDirectories.size == 1) {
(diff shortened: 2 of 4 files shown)


core/src/main/scala/kafka/tools/StorageTool.scala
```diff
@@ -0,0 +1,232 @@
...
+      }
+    }
+
+    if (directories.isEmpty) {
+      stream.println("No directories specified.")
+      0
+    } else {
+      if (foundDirectories.nonEmpty) {
+        stream.println("Found log director%s:".format(
+          if (foundDirectories.size == 1) "y" else "ies"))
```
mumrah: I know it's correct, but this just looks strange to me. Maybe we could do something like:

```
if size == 1
  print "Found log directory:"
else
  print "Found log directories:"
```

core/src/main/scala/kafka/tools/StorageTool.scala
```diff
@@ -0,0 +1,232 @@
...
+      } catch {
+        case e: Throwable => throw new TerseFailure(s"Unable to create storage " +
+          s"directory ${directory}: ${e.getMessage}")
+      }
+      val metaPropertiesPath = Paths.get(directory, "meta.properties")
+      val checkpoint = new BrokerMetadataCheckpoint(metaPropertiesPath.toFile)
+      checkpoint.write(metaProperties.toProperties)
+      stream.println(s"Formatting ${directory}")
+    })
+    0
```
mumrah, line 236: Do we always return zero if we don't throw? What's the use of this return value?

core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
```diff
@@ -88,9 +88,9 @@ class RawMetaProperties(val props: Properties = new Properties()) {
   }
 
   override def toString: String = {
-    "RawMetaProperties(" + props.keySet().asScala.toList.asInstanceOf[List[String]].sorted.map {
+    "{" + props.keySet().asScala.toList.asInstanceOf[List[String]].sorted.map {
```
mumrah, line 91: Is this producing JSON? If so, should we use a JSON encoder instead?

mumrah (approved):
Looks good, few minor comments

core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
cmccabe: It doesn't produce JSON (for example, keys and values are not quoted).

This is intended for human consumption.  It will show up in the command-line output.

core/src/main/scala/kafka/tools/StorageTool.scala
cmccabe: ok

core/src/main/scala/kafka/tools/StorageTool.scala
cmccabe: It's mainly for consistency with `infoCommand` which does sometimes return 1.  But I don't feel that strongly about it.

core/src/main/scala/kafka/server/BrokerMetadataCheckpoint.scala
mumrah: ok, sounds good

PR #10044, opened 2021-02-03
MINOR: Word count should account for extra whitespaces between words

wicknicks:
Currently, the word count demo only accounts for single whitespaces between words. This fix uses a more generic regex to skip multiple spaces or tabs that may occur between words. Corresponding unit tests were updated. 

Also, the quickstart docs already use the correct regex.

Signed-off-by: Arjun Satish <arjun@confluent.io>


mjsax (approved):
LGTM. Can merge after Jenkins passed.

PR #10045, opened 2021-02-03
MINOR: Allow KafkaApis to be configured for Raft controller quorums

rondagostino:
`KafkaApis` is configured differently when it is used in a broker with a Raft-based controller quorum vs. a ZooKeeper quorum.  For example, when using Raft, `ForwardingManager` is required rather than optional, and there is no `AdminManager`, `KafkaController`, or `KafkaZkClient`.  This PR introduces `MetadataSupport` to abstract the two possibilities: `ZkSupport` and `RaftSupport`.  This provides a fluent way to decide what to do based on the type of support that `KafkaApis` has been configured with.  Certain types of requests are not supported when using raft (`AlterIsrRequest`, `UpdateMetadataRequest`, etc.), and `MetadataSupport` gives us an intuitive way to identify the constraints and requirements associated with the different configurations and react accordingly.


core/src/main/scala/kafka/server/MetadataSupport.scala
```diff
@@ -0,0 +1,69 @@
...
+   * Return this instance downcast for use with Raft
+   *
+   * @param createException function to create an exception to throw
+   * @return this instance downcast for use with Raft
+   * @throws Exception if this instance is not for Raft
+   */
+  def requireRaft(createException: () => Exception): RaftSupport = {
+    this match {
+      case raftSupport@RaftSupport(_) => raftSupport
+      case ZkSupport(_, _, _, _) => throw createException()
```
ijuma: It's a bit odd to have methods where we do pattern matching and have some overrides. It's probably simpler to just have different implementations instead of pattern matching.

core/src/main/scala/kafka/server/MetadataSupport.scala
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+}
+
+case class ZkSupport(adminManager: ZkAdminManager,
+                     controller: KafkaController,
+                     zkClient: KafkaZkClient,
+                     forwardingManager: Option[ForwardingManager]) extends MetadataSupport {
+  private val _adminZkClient = new AdminZkClient(zkClient)
```
ijuma: There's no benefit in having a val and def for something. var and def makes sense, but that's not what we have here. Just make the `val` public. Right?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -117,9 +113,17 @@ class KafkaApis(val requestChannel: RequestChannel,
                 val brokerFeatures: BrokerFeatures,
                 val finalizedFeatureCache: FinalizedFeatureCache) extends ApiRequestHandler with Logging {
 
+  metadataSupport match {
+    case ZkSupport(_, _, _, _) if !config.requiresZookeeper =>
+      throw new IllegalStateException("Config specifies Raft but metadata support instance is for ZooKeeper")
+    case RaftSupport(_) if config.requiresZookeeper =>
+      throw new IllegalStateException("Config specifies ZooKeeper but metadata support instance is for Raft")
+    case _ => // consistent configs and metadata support instance
+  }
```
jsancio: We can probably move this to `MetadataSupport::validateRequiresZooKeeper(value: Boolean): Unit`

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -149,12 +153,23 @@ class KafkaApis(val requestChannel: RequestChannel,
...
+          case _ =>
+            handler(request)
+        }
+      // Raft
+      case RaftSupport(fwdMgr) =>
+        if (!request.isForwarded) {
+          fwdMgr.forwardRequest(request, responseCallback)
+        } else {
+          handler(request) // will reject
+        }
```
jsancio: One way to move this complexity to the implementations of `MetadataSupport` is by adding the following method:

```scala
def forwardRequest(request: RequestChannel.Request, callback: Option[AbstractResponse] => Unit): Boolean
```

With this function, here you can:

```scala
if (!metadataSupport.forwardRequest(request, responseCallback)) {
  handler(request);
}
```

core/src/main/scala/kafka/server/MetadataSupport.scala
```diff
@@ -0,0 +1,69 @@
...
+  val forwardingManager: Option[ForwardingManager]
+
+  /**
+   * Return this instance downcast for use with ZooKeeper
+   *
+   * @param createException function to create an exception to throw
+   * @return this instance downcast for use with ZooKeeper
+   * @throws Exception if this instance is not for ZooKeeper
+   */
+  def requireZk(createException: () => Exception): ZkSupport = {
```
jsancio: If you want to delay the execution of some expression, you can do this with `def requireZkOrThrow(thunk: => Exception): ZkSupport`. At the call site you can just do `metadataSupport.requireZkOrThrow(new Exception())`. This will allow to change some of the functions you defined from:
```scala
def createException(...): () => Exception
```
to just
```scala
def createException(...): Exception
```

jsancio (commented):
Thanks for the PR. Did you mean to tag this PR as a draft?

core/src/main/scala/kafka/server/KafkaApis.scala
rondagostino: I added a maybeForward() method to the trait

rondagostino:
This is still in draft because it depends on https://github.com/apache/kafka/pull/10005

core/src/main/scala/kafka/server/MetadataSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+  val forwardingManager: Option[ForwardingManager]
+
+  /**
+   * Return this instance downcast for use with ZooKeeper
+   *
+   * @param createException function to create an exception to throw
+   * @return this instance downcast for use with ZooKeeper
+   * @throws Exception if this instance is not for ZooKeeper
+   */
+  def requireZk(createException: => Exception): ZkSupport
```
jsancio: This comment also applies to `requireRaft`.

How about `requireZkOrThrow`? There is already precedence of this naming convention in Kafka and Java.

jsancio (approved):
LGTM just one minor suggestion.

core/src/main/scala/kafka/server/MetadataSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+  override def confirmConsistentWith(config: KafkaConfig): Unit = {
+    if (!config.requiresZookeeper) {
+      throw new IllegalStateException("Config specifies Raft but metadata support instance is for ZooKeeper")
+    }
+  }
+
+  override def maybeForward(request: RequestChannel.Request,
+                            handler: RequestChannel.Request => Unit,
+                            responseCallback: Option[AbstractResponse] => Unit): Unit = {
+    if (forwardingManager.isDefined && !request.isForwarded && !controller.isActive) {
```
hachikuji: nit: usually when you see `isDefined` followed by `get`, there is likely an opportunity for a `match` or `foreach`. 
```scala
forwardingManager match {
  case Some(mgr) if !request.isForwarded && !controllers.isActive =>
    forwardingManager.get.forwardRequest(request, responseCallback)
  case _ => 
    handler(request)
}
```

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -3197,4 +3200,192 @@ class KafkaApisTest {
...
+  }
+
+  @Test
+  def testRaftHandleStopReplicaRequest(): Unit = {
+    val request: RequestChannel.Request = EasyMock.createNiceMock(classOf[RequestChannel.Request])
+    val kafkaApis = createKafkaApis(raftSupport = true)
+    val e = assertThrows(classOf[UnsupportedVersionException],
+      () => kafkaApis.handleStopReplicaRequest(request))
+    assertEquals(kafkaApis.shouldNeverReceive(request).getMessage, e.getMessage)
+    //    handleUpdateMetadataRequest
```
hachikuji: Did you mean to include these?

core/src/main/scala/kafka/server/MetadataSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+   */
+  def requireRaftOrThrow(createException: => Exception): RaftSupport
+
+  /**
+   * Confirm that this instance is consistent with the given config
+   *
+   * @param config the config to check for consistency with this instance
+   * @throws IllegalStateException if there is an inconsistency (Raft for a ZooKeeper config or vice-versa)
+   */
+  def confirmConsistentWith(config: KafkaConfig): Unit
```
hachikuji: nit: confirm -> ensure?

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3446,17 +3465,41 @@ class KafkaApis(val requestChannel: RequestChannel,
...
       // brokerEpochInRequest > controller.brokerEpoch is possible in rare scenarios where the controller gets notified
       // about the new broker epoch and sends a control request with this epoch before the broker learns about it
-      brokerEpochInRequest < controller.brokerEpoch
+      brokerEpochInRequest < zkSupport.controller.brokerEpoch
     }
   }
 
+  // visible for testing
+  private[server] def shouldNeverReceive(request: RequestChannel.Request): Exception = {
+    new UnsupportedVersionException(s"Should never receive when using a Raft-based metadata quorum: $request")
```
hachikuji: Perhaps no need to print the whole request. Maybe the ApiKey is enough?

core/src/main/scala/kafka/server/MetadataSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+
+import kafka.controller.KafkaController
+import kafka.network.RequestChannel
+import kafka.zk.{AdminZkClient, KafkaZkClient}
+import org.apache.kafka.common.requests.AbstractResponse
+
+sealed trait MetadataSupport {
+  /**
+   * Provide a uniform way of getting to the ForwardingManager, which is a shared concept
+   * despite being optional when using ZooKeeper and required when using Raft
```
hachikuji, line 28: Maybe worth adding a note that this state is temporary. We only allow forwarding manager at the moment with zk broker to enable integration testing. We can remove this as soon as the controller is checked in and we can convert the integration tests.

core/src/main/scala/kafka/server/MetadataSupport.scala
```diff
@@ -0,0 +1,109 @@
...
+  val forwardingManager: Option[ForwardingManager]
+
+  /**
+   * Return this instance downcast for use with ZooKeeper
+   *
+   * @param createException function to create an exception to throw
+   * @return this instance downcast for use with ZooKeeper
+   * @throws Exception if this instance is not for ZooKeeper
+   */
+  def requireZkOrThrow(createException: => Exception): ZkSupport
```
hachikuji, line 39: It's a little odd to see this in the trait. Would it be reasonable to turn them into private defs in `KafkaApis`? I'm ok with it if you think it is better. We'll only ever have the two implementations.

core/src/test/scala/unit/kafka/server/KafkaApisTest.scala
```diff
@@ -3197,4 +3200,192 @@ class KafkaApisTest {
...
+    val request: RequestChannel.Request = EasyMock.createNiceMock(classOf[RequestChannel.Request])
+    val kafkaApis = createKafkaApis(raftSupport = true)
+    val e = assertThrows(classOf[UnsupportedVersionException],
+      () => kafkaApis.handleCreatePartitionsRequest(request))
+    assertEquals(kafkaApis.shouldAlwaysBeForwarded(request).getMessage, e.getMessage)
+  }
+
+  @Test
+  def testRaftHandleDeleteTopicsRequest(): Unit = {
+    val request: RequestChannel.Request = EasyMock.createNiceMock(classOf[RequestChannel.Request])
```
hachikuji: nit: there's a lot of redundancy in these test cases. Could we define helpers? For example:
```scala
def assertUnsupportedWithRaft(handler: RequestChannel.Request => Unit)

assertUnsupportedWithRaft(kafkaApis.handleDeleteTopicsRequest)
```

core/src/test/scala/unit/kafka/utils/TestUtils.scala
```diff
@@ -273,7 +273,8 @@ object TestUtils extends Logging {
                          logDirCount: Int = 1,
                          enableToken: Boolean = false,
                          numPartitions: Int = 1,
-                         defaultReplicationFactor: Short = 1): Properties = {
+                         defaultReplicationFactor: Short = 1,
+                         processRoles: String = ""): Properties = {
```
hachikuji: Adding function args here is a no no. Too many downstream dependencies. Since we have only one usage at the moment, it is better to modify the `Properties` that is returned from `createBrokerConfig`. We can think of something nicer when we have more uses.

hachikuji (commented):
Thanks for the patch. Looks pretty good. Left some comments.

core/src/main/scala/kafka/server/MetadataSupport.scala
rondagostino: Yeah, I prefer it here only because this has worked out so well that I would not be surprised if we tried to use the same approach somewhere else.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3511,4 +3501,14 @@ object KafkaApis {
     FetchResponse.sizeOf(versionId, unconvertedResponse.responseData.entrySet
       .iterator.asScala.filter(element => quota.isThrottled(element.getKey)).asJava)
   }
+
+  // visible for testing
+  private[server] def shouldNeverReceive(request: RequestChannel.Request): Exception = {
```
hachikuji, line 3491: Why move these and not `unsupported`/`notYetSupported`?

rondagostino: Ah, I hadn't tested for those, and I moved based on rewriting tests.  Will move.

hachikuji (approved):
LGTM

PR #10049, opened 2021-02-04
Refactor MetadataCache for Raft metadata

mumrah:
For 2.8 we will keep the existing MetadataCache implementation intact. Instead of replacing it with the changes needed for Raft, we will extract an interface for both implementations to use. KafkaServer (ZK-backed clusters) will only pick the ZK metadata cache impl.

The only real change to production code here will be the use of MetadataBroker instead of Broker for MetadataCache#getAliveBroker and MetadataCache#getAliveBrokers. This was done to eliminate unnecessary divergence between the two implementations.


core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
```diff
@@ -106,13 +103,9 @@ class MetadataCacheTest {
         .setZkVersion(zkVersion)
         .setReplicas(asList(2, 1, 3)))
 
-    val topicIds = new util.HashMap[String, Uuid]()
```
jolshan: Are we just not testing topic Ids here anymore?

core/src/main/scala/kafka/server/MetadataCache.scala
```diff
@@ -157,102 +178,86 @@ class MetadataCache(brokerId: Int) extends Logging {
...
                        errorUnavailableListeners: Boolean = false): Seq[MetadataResponseTopic] = {
-    val snapshot = metadataSnapshot
+    val image = _currentImage
     topics.toSeq.flatMap { topic =>
-      getPartitionMetadata(snapshot, topic, listenerName, errorUnavailableEndpoints, errorUnavailableListeners).map { partitionMetadata =>
+      getPartitionMetadata(image, topic, listenerName, errorUnavailableEndpoints, errorUnavailableListeners).map { partitionMetadata =>
         new MetadataResponseTopic()
           .setErrorCode(Errors.NONE.code)
           .setName(topic)
-          .setTopicId(snapshot.topicIds.getOrElse(topic, Uuid.ZERO_UUID))
```
jolshan: Is there a place where we are setting the topic ID in the metadata response?

jolshan (commented):
I'm not sure if we are still supporting topic IDs here.

core/src/main/scala/kafka/server/MetadataCache.scala
jolshan: Let me know if I missed it somewhere, but we will need the ID included in the response.

I think we might need a MetadataImage method like `topicIdToName` but instead `topicNameToId`

core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
```diff
@@ -483,61 +475,4 @@ class MetadataCacheTest {
     assertEquals(initialBrokerIds.toSet, aliveBrokersFromCache.map(_.id).toSet)
   }
 
-  @Test
-  def testGetClusterMetadataWithOfflineReplicas(): Unit = {
```
hachikuji, line 487: Why remove this?

mumrah: Hmm, I think this was an artifact of the merge, I'll restore this test

core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
```diff
@@ -483,61 +475,4 @@ class MetadataCacheTest {
...
-      .setOfflineReplicas(offline))
-    val version = ApiKeys.UPDATE_METADATA.latestVersion
-    val updateMetadataRequest = new UpdateMetadataRequest.Builder(version, 2, controllerEpoch, brokerEpoch, partitionStates.asJava,
-      brokers.asJava, Collections.emptyMap()).build()
-    cache.updateMetadata(15, updateMetadataRequest)
-
-    val expectedNode0 = new Node(0, "foo", 9092)
-    val expectedNode1 = new Node(1, "", -1)
-
-    val cluster = cache.getClusterMetadata("clusterId", listenerName)
```
mumrah: Since we're looking up the cluster by listener name here, we don't see the offline broker in the MetadataImage because it's endpoints map is empty. This leads to `cluster.leaderFor` on L534 returning null

@hachikuji @cmccabe is a change in metadata behavior, or does this test have bad assumptions

core/src/main/scala/kafka/server/MetadataCache.scala
```diff
@@ -34,16 +34,69 @@ import org.apache.kafka.common.message.UpdateMetadataRequestData.UpdateMetadataP
...
+import org.apache.kafka.common.message.{MetadataResponseData, UpdateMetadataRequestData}
 import org.apache.kafka.common.network.ListenerName
 import org.apache.kafka.common.protocol.Errors
 import org.apache.kafka.common.requests.{MetadataResponse, UpdateMetadataRequest}
 import org.apache.kafka.common.security.auth.SecurityProtocol
 
+trait MetadataCache {
+
+  // errorUnavailableEndpoints exists to support v0 MetadataResponses
+  def getTopicMetadata(
```
cmccabe, line 57: The formatting is kind of weird here.  Why the blank line after `getTopicMetadata(`?

Also, it would be good to have Javadoc for this function

core/src/main/scala/kafka/server/MetadataCache.scala
```diff
@@ -34,16 +34,69 @@ import org.apache.kafka.common.message.UpdateMetadataRequestData.UpdateMetadataP
...
+  def getAliveBrokers: collection.Seq[MetadataBroker]
+
+  def getPartitionInfo(topic: String, partitionId: Int): Option[UpdateMetadataRequestData.UpdateMetadataPartitionState]
+
+  def numPartitions(topic: String): Option[Int]
+
+  // if the leader is not known, return None;
+  // if the leader is known and corresponding node is available, return Some(node)
+  // if the leader is known but corresponding node with the listener name is not available, return Some(NO_NODE)
+  def getPartitionLeaderEndpoint(topic: String, partitionId: Int, listenerName: ListenerName): Option[Node]
```
cmccabe, line 84: can you put this in javadoc format ?  i.e. `@returns`

core/src/main/scala/kafka/server/MetadataCache.scala
```diff
@@ -34,16 +34,69 @@ import org.apache.kafka.common.message.UpdateMetadataRequestData.UpdateMetadataP
...
+    new RaftMetadataCache(brokerId)
+  }
+}
+
 /**
  *  A cache for the state (e.g., current leader) of each partition. This cache is updated through
  *  UpdateMetadataRequest from the controller. Every broker maintains the same cache, asynchronously.
  */
-class MetadataCache(brokerId: Int) extends Logging {
+class ZkMetadataCache(brokerId: Int) extends MetadataCache with Logging {
```
cmccabe, line 118: This is kind of a lot of code so why not have it in its own file?

The JavaDoc should also explain that this is for brokers using ZK and not for brokers using the metadata quorum.

core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
```diff
@@ -16,20 +16,20 @@
...
 import org.apache.kafka.common.requests.UpdateMetadataRequest
 import org.apache.kafka.common.security.auth.SecurityProtocol
-import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.Assertions._
+import org.junit.jupiter.api.Test
 
+import java.util.Collections
 import scala.jdk.CollectionConverters._
 
 class MetadataCacheTest {
```
cmccabe, line 45: we should have some way of running these tests on the raft metadata cache as well as the zk metadata cache.  I guess we can do that in a follow-up PR, though

mumrah: Yea, I think we can probably do some parameterization thing or possibly even use test templating similar to #9986

core/src/main/scala/kafka/server/MetadataCache.scala
```diff
@@ -34,16 +34,69 @@ import org.apache.kafka.common.message.UpdateMetadataRequestData.UpdateMetadataP
...
+  // This method returns the deleted TopicPartitions received from UpdateMetadataRequest
+  def updateMetadata(correlationId: Int, request: UpdateMetadataRequest): collection.Seq[TopicPartition]
+
+  def contains(topic: String): Boolean
+
+  def contains(tp: TopicPartition): Boolean
+}
+
+object MetadataCache {
+  def zkMetadataCache(brokerId: Int): ZkMetadataCache = {
```
hachikuji, line 105: nit: I don't think these factories are providing much over the constructors `MetdataCache.zkMetadataCache` vs `new ZkMetadataCache`

core/src/main/scala/kafka/server/MetadataCache.scala
mumrah: I left the ZK implementation in place since it's really the only production implementation for now. It also reduces the size of the diff for this change. I don't feel very strongly about it either way, so I'm happy to relocate it to a separate file

core/src/main/scala/kafka/server/MetadataCache.scala
mumrah: Yea, this just got pulled up from the class when I extracted the trait. I'll fix up these comments

core/src/main/scala/kafka/server/MetadataCache.scala
hachikuji: Perhaps we can do this as a follow-up? It is nice at the moment to see the diffs clearly.

core/src/main/scala/kafka/server/MetadataCache.scala
mumrah: I was thinking it might be easier to refactor in the future if we only need to rename the factory method rather than changing all the `new Class`.

core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
hachikuji: Should be easy to do with all the fancy junit 5 stuff. Since there are no separate tests for `RaftMetadataCache` (unless I'm missing some), I think we should do it here.

core/src/main/scala/kafka/server/MetadataCache.scala
hachikuji: Ok. Probably not a ton of work either way, but will leave it up to you.

core/src/main/scala/kafka/server/MetadataCache.scala
```diff
@@ -34,16 +34,69 @@ import org.apache.kafka.common.message.UpdateMetadataRequestData.UpdateMetadataP
...
+trait MetadataCache {
+
+  // errorUnavailableEndpoints exists to support v0 MetadataResponses
+  def getTopicMetadata(
+                        topics: collection.Set[String],
+                        listenerName: ListenerName,
+                        errorUnavailableEndpoints: Boolean = false,
+                        errorUnavailableListeners: Boolean = false): collection.Seq[MetadataResponseData.MetadataResponseTopic]
+
+  def getAllTopics(): collection.Set[String]
```
hachikuji, line 63: nit: We can save it for a follow-up, but it would be nice to drop all the `get` prefixes here

core/src/main/scala/kafka/server/metadata/MetadataImage.scala
```diff
@@ -118,5 +118,10 @@ case class MetadataImage(partitions: MetadataPartitions,
   def topicIdToName(uuid: Uuid): Option[String] = {
     partitions.topicIdToName(uuid)
   }
+
+  def topicNameToId(name: String): Option[Uuid] = {
+
```
hachikuji: nit: remove newline

core/src/main/scala/kafka/server/metadata/RaftMetadataCache.scala
```diff
@@ -0,0 +1,389 @@
...
+                               partitionId: Int,
+                               stateInfo: UpdateMetadataPartitionState): Unit = {
+    val infos = partitionStates.getOrElseUpdate(topic, mutable.LongMap.empty)
+    infos(partitionId) = stateInfo
+  }
+}
+
+
+class RaftMetadataCache(val brokerId: Int) extends MetadataCache with Logging {
+  this.logIdent = s"[MetadataCache brokerId=${brokerId}] "
```
hachikuji: nit: unneeded braces

core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
mumrah: Yea easy enough. I added `@ParameterizedTest` to MetadataCacheTest. One is failing with the Raft metadata cache, so I left that for ZK-only now.

core/src/main/scala/kafka/server/metadata/RaftMetadataCache.scala
```diff
@@ -0,0 +1,389 @@
...
+  def stateChangeTraceEnabled(): Boolean = {
+    stateChangeLogger.isTraceEnabled
+  }
+
+  def logStateChangeTrace(str: String): Unit = {
+    stateChangeLogger.trace(str)
+  }
+
+  // This method returns the deleted TopicPartitions received from UpdateMetadataRequest
+  override def updateMetadata(correlationId: Int, request: UpdateMetadataRequest): Seq[TopicPartition] = {
```
hachikuji, line 305: We shouldn't need this for the RaftMetadataCache. Did we end up rewriting the code to update from the image builder?

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
```diff
@@ -37,7 +36,15 @@ object MetadataBroker {
...
-      true)
+      fenced = true)
+  }
+
+  def apply(broker: Broker): MetadataBroker = {
+    new MetadataBroker(broker.id, broker.rack.orNull,
+      broker.endPoints.map { endpoint =>
+        endpoint.listenerName.value -> new Node(broker.id, endpoint.host, endpoint.port, broker.rack.orNull)
+      }.toMap,
+      fenced = false)
```
mumrah, line 47: I wanted to go ahead and conform to MetadataBroker for both implementations. One side-effect is we are exposing the fenced flag to ZK-based clusters. I've set it to false here since we don't have any notion of broker fencing in the ZK-based metadata. Is there any harm in this?

core/src/test/scala/unit/kafka/server/MetadataCacheTest.scala
```diff
@@ -485,7 +494,7 @@ class MetadataCacheTest {
 
   @Test
```
mumrah: This test is failing in Raft mode. I'll investigate

core/src/main/scala/kafka/server/metadata/RaftMetadataCache.scala
hachikuji: Discussed offline. This is useful temporarily to allow reuse of `MetadataCacheTest`. Once 2.8 is cut, we'll hopefully be able to consolidate this logic to use the MetadataImage for both zk and raft.

core/src/main/scala/kafka/server/MetadataCache.scala
mumrah: KAFKA-12299

core/src/main/scala/kafka/server/MetadataCache.scala
mumrah: KAFKA-12299

core/src/main/scala/kafka/server/metadata/MetadataBrokers.scala
cmccabe: as long as this is just used by the ZK code path there is no harm

cmccabe (approved):
LGTM

core/src/main/scala/kafka/server/metadata/RaftMetadataCache.scala
```diff
@@ -0,0 +1,390 @@
...
+    inLock(lock) {
+      val image = _currentImage
+      val builder = MetadataImageBuilder(brokerId, logger.underlying, image)
+
+      builder.controllerId(if (request.controllerId() < 0) None else Some(request.controllerId()))
+
+      // Compare the new brokers with the existing ones.
+      def toMetadataBroker(broker: UpdateMetadataBroker): MetadataBroker = {
+        val endpoints = broker.endpoints().asScala.map { endpoint =>
+          endpoint.listener -> new Node(broker.id(), endpoint.host(), endpoint.port())
```
mumrah, line 315: Note that here we are, somewhat incorrectly, excluding the broker's rack from the Node object. This was done to keep the behavior compatible with the existing MetadataCache (now ZkMetadataCache). Once we have this merged and 2.8 is released, we can work on determining if this was done for a good reason, or if it's a bug in the existing MetadataCache logic. Since this class isn't used for any production code yet, we can leave this as-is for now.

PR #10050, opened 2021-02-04
MINOR: Don't assume number of dns results (stabilize ClusterConnectio…

chia7712:
```ClusterConnectionStatesTest#testMultipleIPsWithUseAll``` assume the number of addresses is 3. That assumption is unstable since it can be changed in the future. For example, the addresses of ```kafka.apache.org``` is changed to 2 recently.

It seems to me the test case is used to make sure all addresses (from resolving) are used when building connections. Hence, we can rewrite it by for-loop to make sure all addresses are used across connections.

### DIFF
--- a/clients/src/test/java/org/apache/kafka/clients/ClusterConnectionStatesTest.java
+++ b/clients/src/test/java/org/apache/kafka/clients/ClusterConnectionStatesTest.java
@@ -29,7 +29,11 @@ import java.lang.reflect.Method;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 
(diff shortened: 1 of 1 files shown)


mumrah:
@chia7712 I think this got fixed yesterday on trunk with the addition of a new HostResolver interface. Now we can precisely control what IPs are resolved during testing. https://github.com/apache/kafka/commit/131d4753cfed65ed6dee0a8c754765c97c3d513f

chia7712:
@mumrah Thanks for kind reminder. close this now.

PR #10052, opened 2021-02-04
KAFKA-12289: Adding test cases for prefix scan in InMemoryKeyValueStore

vamossagar12:
comment to ping reviewers.*

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
@@ -104,9 +104,11 @@ public interface ReadOnlyKeyValueStore<K, V> {
      * prefix into the format in which the keys are stored in the stores needs to be passed to this method.
      * The returned iterator must be safe from {@link java.util.ConcurrentModificationException}s
      * and must not return null values.
-     * Order is not guaranteed as bytes lexicographical ordering might not represent key order.
-     * For example, if the key type is Integer, and the store contains keys [1, 2, 11, 13],
-     * then running store.prefixScan(1, new IntegerSerializer()) would return [1] and not [1,11,13].
+     * Since {@code prefixScan()} relies on byte lexicographical ordering and not on the ordering of the key type, results for some types might be unexpected.
+     * For example, if the key type is {@code Integer}, and the store contains keys [1, 2, 11, 13],
+     * then running {@code store.prefixScan(1, new IntegerSerializer())} will return [1] and not [1,11,13]. 
+     * In contrast, if the key type is {@code String} the keys will be sorted [1, 11, 13, 2] in the store and {@code store.prefixScan(1, new StringSerializer())} will return [1,11,13]. 
+     * In both cases {@code prefixScan()} starts the scan at 1 and stops at 2.
      *
      * @param prefix The prefix.
      * @param prefixKeySerializer Serializer for the Prefix key type
--- a/streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
+++ b/streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
@@ -104,6 +104,9 @@ public interface ReadOnlyKeyValueStore<K, V> {
      * prefix into the format in which the keys are stored in the stores needs to be passed to this method.
      * The returned iterator must be safe from {@link java.util.ConcurrentModificationException}s
      * and must not return null values.
+     * Order is not guaranteed as bytes lexicographical ordering might not represent key order.
+     * For example, if the key type is Integer, and the store contains keys [1, 2, 11, 13],
+     * then running store.prefixScan(1, new IntegerSerializer()) would return [1] and not [1,11,13].
      *
      * @param prefix The prefix.
      * @param prefixKeySerializer Serializer for the Prefix key type
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStoreTest.java
@@ -215,7 +215,7 @@ public class InMemoryKeyValueStoreTest extends AbstractKeyValueStoreTest {
         byteStore.putAll(entries);
         byteStore.flush();
 
-        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = byteStore.prefixScan("d", stringSerializer);
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = byteStore.prefixScan("bb", stringSerializer);
         int numberOfKeysReturned = 0;
 
         while (keysWithPrefix.hasNext()) {
--- a/streams/src/test/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStoreTest.java
+++ b/streams/src/test/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStoreTest.java
@@ -17,17 +17,54 @@
 package org.apache.kafka.streams.state.internals;
 
 import org.apache.kafka.common.serialization.Serde;
+import org.apache.kafka.common.serialization.Serdes;
+import org.apache.kafka.common.serialization.Serializer;
+import org.apache.kafka.common.serialization.StringSerializer;
+import org.apache.kafka.common.utils.Bytes;
+import org.apache.kafka.streams.KeyValue;
 import org.apache.kafka.streams.processor.StateStoreContext;
 import org.apache.kafka.streams.state.KeyValueStore;
-import org.apache.kafka.streams.state.StoreBuilder;
+import org.apache.kafka.streams.state.KeyValueStoreTestDriver;
 import org.apache.kafka.streams.state.Stores;
+import org.apache.kafka.streams.state.StoreBuilder;
+import org.apache.kafka.streams.state.KeyValueIterator;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.Assert.assertEquals;
 
 public class InMemoryKeyValueStoreTest extends AbstractKeyValueStoreTest {
+
+    private KeyValueStore<Bytes, byte[]> byteStore;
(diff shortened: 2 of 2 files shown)


streams/src/test/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStoreTest.java
```diff
@@ -60,4 +67,22 @@ public void shouldRemoveKeysWithNullValues() {
...
+        store = createKeyValueStore(driver.context());
+        final String value = "value";
+        final List<KeyValue<Integer, String>> entries = new ArrayList<>();
+        entries.add(new KeyValue<>(1, value));
+        entries.add(new KeyValue<>(2, value));
+        entries.add(new KeyValue<>(11, value));
+        entries.add(new KeyValue<>(13, value));
+
+        store.putAll(entries);
+        final KeyValueIterator<Integer, String> keysWithPrefix = store.prefixScan(1, new IntegerSerializer());
```
vamossagar12: @cadonna I found something interesting here. As you can see this creates a state store with Integer prefix. Now when I run prefixScan, I get only 1 from the iterator even though we have 11 and 13 as the keys. This is because in prefixScan, we do this: `final Bytes to = Bytes.increment(from);` which sets `to` to 2.
Because of that, it doesn't return all the keys. 
Question is, with prefix Scan we are primarily focussing on lexicographically next key which has the same prefix. That way, 11 should be returned. But with the way it has been implemented for this store, it won't work.

cadonna: Thank you for pointing this out! 
I will have a look at this in the next days.

vamossagar12: @cadonna , I was thinking about this today and the question that I have is what exactly does a prefix scan mean for numeric keys. For the way we have implemented, it will work even for numeric keys if we can retrieve the keys in Lexicographical manner. 
So, if the prefix is `1` and post increment, it becomes `2`, then the range in which I need to search becomes (1,2]. If my map has 1,2,11,13 then I should be able to fetch 1,11,13 as those come before 2 lexicographically. But when I test the underlying map from my test case, it returns 1,2,11,13 and that is why the range is failing. 

In the test class, the KV store being created has Integer keys, but the underlying map still uses a lexicographically map. So, ideally, it should have retured keys lexicographically.

vamossagar12: I spent some more time on this today. What seems to be happening here is that in InMemoryKVStore, even though we are creating a NavigableMap as the inner map and even though Bytes(Key type for the inner map) does implement Comparable, when I insert the following keys:

```final List<KeyValue<Integer, String>> entries = new ArrayList<>();
        entries.add(new KeyValue<>(1, value));
        entries.add(new KeyValue<>(2, value));
        entries.add(new KeyValue<>(11, value));
        entries.add(new KeyValue<>(13, value));
store.putAll(entries);
```
so the expectation(atleast for me) was that i should get the keys stored lexicographically ie the output of 

```
final KeyValueIterator<Integer, String> allKeys = store.all();
        while (allKeys.hasNext()){
            System.out.println(allKeys.next().key);
        }
``` 
should be `1 11 13 2` but what gets returned is `1 2 11 13` i.e the order in which we inserted. 

Just to strengthen the point, when I insert keys in the order:
```final List<KeyValue<Integer, String>> entries = new ArrayList<>();
        entries.add(new KeyValue<>(1, value));
        entries.add(new KeyValue<>(11, value));
        entries.add(new KeyValue<>(13, value));
        entries.add(new KeyValue<>(2, value));

        store.putAll(entries);
```
and when I fetch all keys, what I get back is still `1 2 11 13` i.e the natural ordering of the integer keys. And this is precisely the reason why prefix scan is returning only 1 key in the test case that I had pointed out. I did try to pass in an explicit comparator to the TreeMap but that didn't help as well. What are your thoughts on this?

cadonna: The reason, we get only `1` when we scan for prefix `1` is that the integer serializer serializes `11` and `13` in the least significant byte instead of serializing `1` in the byte before the least significant byte and `1` and `3` in the least significant byte. With the former the **byte** lexicographical order of `1 2 11 13` would be `1 2 11 13` which corresponds to the natural order of integers. With the latter the **byte** lexicographical order of `1 2 11 13` would be `1 11 13 2` which corresponds to the string lexicographical order. So the serializer determines the order of the entries and the store always returns the entries in byte lexicographical order.

You will experience a similar when you call `range(-1, 2)` on the in-memory state store in the unit test. You will get back an empty result since `-1` is larger then `2` in byte lexicographical order   
 when the `IntegerSerializer` is used. Also not the warning that is output, especially this part `... or serdes that don't preserve ordering when lexicographically comparing the serialized bytes ...`
 
I think we should clearly state this limitation in the javadocs of the `prefixScan()` as we have done for `range()`, maybe with an example. 

Currently, to get `prefixScan()` working for all types, we would need to do a complete scan (i.e. `all()`) followed by a filter, right? 

Double checking: Is my understanding correct? @ableegoldman

ableegoldman: Yeah, the underlying store compares the serializer bytes lexicographically, it doesn't have any concept of "Integer" or any other type. And the really tricky thing is that it scans lexicographically, which means from left to right, whereas when we serialize things we usually do so from right to left. eg `2` in binary is `10` whereas 11 in binary is `1011` and 13 is `1101`. 
The problem here is that the serialized version of 2 is a different number of bytes than the serialized form of 11/13, so the lexicographical comparator is effectively comparing digits of a different magnitude.

vamossagar12: Thanks @cadonna , @ableegoldman  for the detailed explanation. I understood the behaviour now.

vamossagar12: > The reason, we get only `1` when we scan for prefix `1` is that the integer serializer serializes `11` and `13` in the least significant byte instead of serializing `1` in the byte before the least significant byte and `1` and `3` in the least significant byte. With the former the **byte** lexicographical order of `1 2 11 13` would be `1 2 11 13` which corresponds to the natural order of integers. With the latter the **byte** lexicographical order of `1 2 11 13` would be `1 11 13 2` which corresponds to the string lexicographical order. So the serializer determines the order of the entries and the store always returns the entries in byte lexicographical order.
> 
> You will experience a similar when you call `range(-1, 2)` on the in-memory state store in the unit test. You will get back an empty result since `-1` is larger then `2` in byte lexicographical order
> when the `IntegerSerializer` is used. Also not the warning that is output, especially this part `... or serdes that don't preserve ordering when lexicographically comparing the serialized bytes ...`
> 
> I think we should clearly state this limitation in the javadocs of the `prefixScan()` as we have done for `range()`, maybe with an example.
> 
> Currently, to get `prefixScan()` working for all types, we would need to do a complete scan (i.e. `all()`) followed by a filter, right?

That is correct. That is the only way currently. 

> 
> Double checking: Is my understanding correct? @ableegoldman

I think adding a warning similar to the range() query would be good. I will do that as part of the PR. However, in this test class, adding test cases for the integer serializer won't make sense. Probably I will create another KVStore and add tests for those. Is that ok, @cadonna ?

cadonna: That sounds good!

vamossagar12:
@cadonna , As discussed, i have added the test cases by adding store with key as Bytes. Also, have added the warning with the example in the javadocs for prefixScan()

streams/src/main/java/org/apache/kafka/streams/state/ReadOnlyKeyValueStore.java
```diff
@@ -104,6 +104,9 @@
      * prefix into the format in which the keys are stored in the stores needs to be passed to this method.
      * The returned iterator must be safe from {@link java.util.ConcurrentModificationException}s
      * and must not return null values.
+     * Order is not guaranteed as bytes lexicographical ordering might not represent key order.
+     * For example, if the key type is Integer, and the store contains keys [1, 2, 11, 13],
+     * then running store.prefixScan(1, new IntegerSerializer()) would return [1] and not [1,11,13].
```
cadonna:
```suggestion
     * Since {@code prefixScan()} relies on byte lexicographical ordering and not on the ordering of the key type, results for some types might be unexpected.
     * For example, if the key type is {@code Integer}, and the store contains keys [1, 2, 11, 13],
     * then running {@code store.prefixScan(1, new IntegerSerializer())} will return [1] and not [1,11,13]. 
     * In contrast, if the key type is {@code String} the keys will be sorted [1, 11, 13, 2] in the store and {@code store.prefixScan(1, new StringSerializer())} will return [1,11,13]. 
     * In both cases {@code prefixScan()} starts the scan at 1 and stops at 2.
```

streams/src/test/java/org/apache/kafka/streams/state/internals/InMemoryKeyValueStoreTest.java
```diff
@@ -60,4 +97,131 @@ public void shouldRemoveKeysWithNullValues() {
...
+        entries.add(new KeyValue<>(
+            new Bytes(stringSerializer.serialize(null, "b")),
+            stringSerializer.serialize(null, "c")));
+        entries.add(new KeyValue<>(
+            new Bytes(stringSerializer.serialize(null, "c")),
+            stringSerializer.serialize(null, "e")));
+        byteStore.putAll(entries);
+        byteStore.flush();
+
+        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = byteStore.prefixScan("d", stringSerializer);
```
cadonna: Here, I would use `bb` as the prefix to cover the case where a key is a prefix of the prefix.  

```suggestion
        final KeyValueIterator<Bytes, byte[]> keysWithPrefix = byteStore.prefixScan("bb", stringSerializer);
```

cadonna (commented):
@vamossagar12 Thanks for the update!

Here my comments.

vamossagar12:
Thanks @cadonna , i have committed your suggestions.

cadonna (approved):
Thanks, @vamossagar12 !

LGTM!

cadonna:
Call for committer approval: @guozhangwang @ableegoldman @mjsax @vvcephei @abbccdda

guozhangwang:
LGTM. Thanks for the contribution @vamossagar12 !

guozhangwang:
Cherry-picked to 2.8 as well cc @vvcephei

PR #2835, opened 2017-04-10
KAFKA-5051: Avoid reverse DNS lookup to obtain hostname for TLS

rajinisivaram:

PR #10059, opened 2021-02-04
KAFKA-8562: SaslChannelBuilder - avoid (reverse) DNS lookup while building underlying SslTransportLayer

dpoldrugo:
Description:
As suggested by @omkreddy in this [comment](https://issues.apache.org/jira/browse/KAFKA-8562?focusedCommentId=16912437&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16912437), implemented avoiding (reverse) DNS lookup while building underlying SslTransportLayer.

How the problem was manifested:
When clients or other brokers are connecting to a broker using SASL_SSL, a broker was doing (reverse) DNS lookup and if there is no PTR Record, the lookup could last several seconds, which in the end caused big latencies on several parts of the system... replication, consume requests and produce requests.
Here you can see a recorded sample: 
KAFKA-8562 reverse DNS sampling
Also, here is a Wireshark packet capture for DNS requests, and in this case you can see that it lasted more then 11 seconds:
KAFKA-8562 wireshark dns packet capture
When using PLAINTEXT or SSL, this problem doesn't manifest.

Solution:
In #2835 , @rajinisivaram already added a helper method `SslChannelBuilder.peerHost`, so I just moved it to a new class called `ChannelBuilderUtils` and used it in `SaslChannelBuilder.buildTransportLayer` method.


dpoldrugo:
@ijuma, @omkreddy, @rajinisivaram could you take a look?

dpoldrugo:
Should this feature have it's own config so users configure it by themselves... to enable / disable reverse dns resolving for SASL_SSL ?

hachikuji (approved):
LGTM. Thanks for the patch!

hachikuji:
Hmm.. I may have been too hasty with my approval. We need to investigate the test failures. I've triggered another build.

hachikuji:
I think the test failures are caused by this: https://github.com/apache/kafka/blob/trunk/clients/src/test/java/org/apache/kafka/common/security/authenticator/SaslAuthenticatorFailureDelayTest.java#L222. We are relying on the reverse DNS to map "127.0.0.1" to "localhost." If we use "localhost" directly, then the tests pass.

It seems reasonable to me to change the test case. I cannot think of a reason why this would change the reasoning that is documented under `ChannelBuilderUtils.peerHost`, which seems to apply just as well to SASL_SSL as SSL.

omkreddy:
@dpoldrugo Thanks for the PR. LGTM.  As @hachikuji  mentioned, we should update  `SaslAuthenticatorFailureNoDelayTest`, `SaslAuthenticatorTest` tests.  I think, we may not need config, as we are already avoiding DNS lookup for SSL.

dpoldrugo:
Thanks @hachikuji and @omkreddy for you comments. 
Changed the tests as you proposed:
https://github.com/apache/kafka/pull/10059/commits/c605ea41fb0b342e9bb17595941c0e658bb63360

clients/src/main/java/org/apache/kafka/common/network/ChannelBuilderUtils.java
```diff
@@ -0,0 +1,63 @@
...
+     *     SubjectAltName (ipaddress). This could be used when Kafka is on a private network.
+     *     If reverse DNS lookup is used, authentication would succeed using IP address if lookup
+     *     fails and IP address is used, but authentication would fail if lookup succeeds and
+     *     dns name is used. For consistency and to avoid dependency on a potentially insecure
+     *     DNS, reverse DNS lookup should be avoided and the IP address specified by the client for
+     *     connection should be used to create the SSL engine.</li>
+     *   </ul></li>
+     * </ul>
+     */
+    static String peerHost(SelectionKey key) {
```
ijuma: Instead of doing this, I think we can have the following 3 methods in `SslFactory`. Thoughts?

```java
    public SSLEngine createSslEngine(Socket socket) {
        return createSslEngine(peerHost(socket), socket.getPort());
    }

    /**
     * Prefer `createSslEngine(Socket)` if a `Socket` instance is available. If using this overload,
     * avoid reverse DNS resolution in the computation of `peerHost`.
     */
    public SSLEngine createSslEngine(String peerHost, int peerPort) {
        if (sslEngineFactory == null) {
            throw new IllegalStateException("SslFactory has not been configured.");
        }
        if (mode == Mode.SERVER) {
            return sslEngineFactory.createServerSslEngine(peerHost, peerPort);
        } else {
            return sslEngineFactory.createClientSslEngine(peerHost, peerPort, endpointIdentification);
        }
    }

    /**
     * Returns host/IP address of remote host without reverse DNS lookup to be used as the host
     * for creating SSL engine. This is used as a hint for session reuse strategy and also for
     * hostname verification of server hostnames.
     * <p>
     * Scenarios:
     * <ul>
     *   <li>Server-side
     *   <ul>
     *     <li>Server accepts connection from a client. Server knows only client IP
     *     address. We want to avoid reverse DNS lookup of the client IP address since the server
     *     does not verify or use client hostname. The IP address can be used directly.</li>
     *   </ul>
     *   </li>
     *   <li>Client-side
     *   <ul>
     *     <li>Client connects to server using hostname. No lookup is necessary
     *     and the hostname should be used to create the SSL engine. This hostname is validated
     *     against the hostname in SubjectAltName (dns) or CommonName in the certificate if
     *     hostname verification is enabled. Authentication fails if hostname does not match.</li>
     *     <li>Client connects to server using IP address, but certificate contains only
     *     SubjectAltName (dns). Use of reverse DNS lookup to determine hostname introduces
     *     a security vulnerability since authentication would be reliant on a secure DNS.
     *     Hence hostname verification should fail in this case.</li>
     *     <li>Client connects to server using IP address and certificate contains
     *     SubjectAltName (ipaddress). This could be used when Kafka is on a private network.
     *     If reverse DNS lookup is used, authentication would succeed using IP address if lookup
     *     fails and IP address is used, but authentication would fail if lookup succeeds and
     *     dns name is used. For consistency and to avoid dependency on a potentially insecure
     *     DNS, reverse DNS lookup should be avoided and the IP address specified by the client for
     *     connection should be used to create the SSL engine.</li>
     *   </ul></li>
     * </ul>
     */
    private String peerHost(Socket socket) {
        return new InetSocketAddress(socket.getInetAddress(), 0).getHostString();
    }

dpoldrugo:
@hachikuji @ijuma @omkreddy will you consider this pull request to be part of of 2.5.2 release?

omkreddy:
@dpoldrugo Currently we don't have plan for 2.5.2 release. also we normally backport only critical/blocker fixes. not sure if this comes under this category.
https://cwiki.apache.org/confluence/display/KAFKA/Future+release+plan

dpoldrugo:
@omkreddy well for as at Infobip it caused an incident which basically blocked all kafka traffic, just because of some clients didn't have PTR records and thus reverse dns blocked the networking thread pool of all kafka brokers. We are currently running a custom built kafka 2.5.0.

PR #10063, opened 2021-02-05
KAFKA-12258: Add support for splitting appending records

jsancio:
1. Type `BatchAccumulator`. Add support for appending records into one or more batches.
2. Type `RaftClient`. Rename `scheduleAppend` to `scheduleAtomicAppend`.
3. Type `RaftClient`. Add a new method `scheduleAppend` which appends records to the log using as many batches as necessary.
4. Increase the batch size from 1MB to 8MB.

comment to ping reviewers.*


raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -79,24 +80,69 @@ public BatchAccumulator(
...
+        }
+
+        ObjectSerializationCache serializationCache = new ObjectSerializationCache();
+
+        appendLock.lock();
+        try {
+            maybeCompleteDrain();
+
+            for (T record : records) {
+                BatchBuilder<T> batch = maybeAllocateBatch(
```
mumrah: Do we need to create a batch for each record? What is the overhead associated with a batch?

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -79,24 +80,69 @@ public BatchAccumulator(
...
+     * @param epoch the expected leader epoch. If this does not match, then {@link Long#MAX_VALUE}
+     *              will be returned as an offset which cannot become committed
+     * @param records the list of records to include in the batches
+     * @return the expected offset of the last record; {@link Long#MAX_VALUE} if the epoch does not
+     *         match; null if no memory could be allocated for the batch at this time
+     * @throws RecordBatchTooLargeException if the size of one record T is greater than the maximum
+     *         batch size; if this exception is throw some of the elements in records may have
+     *         been committed
      */
     public Long append(int epoch, List<T> records) {
```
mumrah, line 101: Since these two `append` methods are mostly identical, can we use a single private method with `isAtomic` as an argument? Otherwise seems we will need to update both methods in most cases as we add features and fix bugs.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
jsancio: No. `maybeAllocateBatch` should only create a new batch if the record with a value of size `serde.recordSize(record, serializationCache)` doesn't fit in the current batch being built. We also don't run the risk of the polling thread draining the batch too early because this method holds on to the lock during this iteration.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
jsancio: Yeah. I thought about doing this but I felt that the implementation was very subtle and hard to read. I use this strategy for `private Long append(int epoch, List<T> records, boolean isAtomic)` in `KafkaRaftClient` since the difference is obvious.

In the `BatchAccumulator` for the "atomic" version we need to calculate the total size and allocate a buffer/batch if needed that fits that total size. For the non-atomic version we need to allocate a buffer/batch for every record if it doesn't fit the current batch.

I'll create a commit for this so you can see the two side by side.

jsancio: @mumrah This one way to merge these two methods. Let me know which one you prefer.

https://gist.github.com/jsancio/06bdffc9f25450127b1dc730ebbd55f7

jsancio:
@mumrah I found a few more issues on how we compute the needed sizes. I need to fix one of the remaining issues and write tests for them. Let's assume that this PR is still a WIP.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
mumrah: The gist looks pretty reasonable 👍

jsancio: Done. Implemented the Gist.

jsancio:
@mumrah Ready for review. Fixed the size calculation issues that I discovered when implementing this PR.

hachikuji (commented):
Thanks for the patch. I think we might need changes in `RaftManager` as well to ensure that `max.message.bytes` gets set correctly for the `@metadata` log. I think currently we take the configuration from the general kafka configuration, which is probably not what we want. I wonder if there are any other overridable configurations that could cause us trouble.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
```diff
@@ -77,6 +79,29 @@ default void handleResign() {}
      */
     void register(Listener<T> listener);
 
+    /**
+     * Append a list of records to the log. The write will be scheduled for some time
+     * in the future. There is no guarantee that appended records will be written to
+     * the log and eventually committed. While the order of the records is preserve, they can
+     * be appended to the log using one or more batches. This means that each record could
+     * be committed independently.
```
hachikuji: It might already be clear enough given the previous sentence, but maybe we could emphasize that if any record becomes committed, then all records ordered before it are guaranteed to be committed as well.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -79,67 +80,110 @@ public BatchAccumulator(
...
-     *              {@link Long#MAX_VALUE} will be returned as an offset which
-     *              cannot become committed.
-     * @param records the list of records to include in a batch
-     * @return the expected offset of the last record (which will be
-     *         {@link Long#MAX_VALUE} if the epoch does not match), or null if
-     *         no memory could be allocated for the batch at this time
+     * The order of the elements in the records argument will match the order in the batches.
+     * This method will use as many batches as necessary to serialize all of the records. Since
+     * this method can split the records into multiple batches it is possible that some of the
+     * recors will get committed while other will not when the leader fails.
```
hachikuji: typo: `recors`

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -79,67 +80,110 @@ public BatchAccumulator(
...
-    private BatchBuilder<T> maybeAllocateBatch(int batchSize) {
-        if (currentBatch == null) {
+    private void maybeResetLinger() {
+        if (!lingerTimer.isRunning()) {
+            lingerTimer.reset(time.milliseconds() + lingerMs);
+        }
+    }
+
+    private BatchBuilder<T> maybeAllocateBatch(int[] recordSizes) {
+        int bytesNeeded = BatchBuilder.batchSizeForRecordSizes(
```
hachikuji: The call to `batchSizeForRecordSizes` delegates to `bytesNeededForRecordSizes`, which does a full pass over the record sizes and adds the size of the header. Then below `hasRoomFor` makes another call to `bytesNeededForRecordSizes`. Can we avoid the re-computation?

In general, I wonder if we can be a little looser about the batch size limit and avoid some of the extra work. When compression is enabled, these are just estimates anyway. What if we treat the batch size limit as the maximum batch size before compression and ignore overhead? Basically we can set max.message.bytes to Int.MaxValue and just enforce the limit here on the pre-compression bytes.

raft/src/main/java/org/apache/kafka/raft/RaftClient.java
jsancio: Done.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
jsancio: We discussed this offline. For this PR we decided to avoid computing the sizes twice.

In the future we want avoid having strict enforcement of batch sizes. We can avoid this we if add support in the Raft client or controller for applying a set of records atomically even if they are not in the same batch.

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -79,70 +82,111 @@ public BatchAccumulator(
...
+    ) {
         if (currentBatch == null) {
             startNewBatch();
-        } else if (!currentBatch.hasRoomFor(batchSize)) {
-            completeCurrentBatch();
-            startNewBatch();
         }
+
+        if (currentBatch != null) {
+            OptionalInt bytesNeeded = currentBatch.roomFor(records, serializationCache);
```
hachikuji: So `roomFor` returns the the bytes needed to accommodate the batch when there is not enough room in the current batch. Perhaps an alternative name might be `overflowBytes` or maybe `bytesNeeded` to match the variable name here?

raft/src/main/java/org/apache/kafka/raft/internals/BatchBuilder.java
```diff
@@ -307,4 +309,38 @@ public int writeRecord(
...
+
+    private int bytesNeededForRecords(
+        Collection<T> records,
+        ObjectSerializationCache serializationCache
+    ) {
+        long expectedNextOffset = nextOffset;
+        int bytesNeeded = 0;
+        for (T record : records) {
+            if (expectedNextOffset - baseOffset >= Integer.MAX_VALUE) {
+                return Integer.MAX_VALUE;
```
hachikuji: Hmm.. If we return `Integer.MAX_VALUE`, then that will prevent us from being able to split when the number of records exceeds `Integer.MAX_VALUE`. Similarly, it might be worth hardening this logic by ensuring `bytesNeeded` does not overflow.  These cases are unlikely in practice, so we can save them for a follow-up if there is no clear way to handle them right now.

jsancio: True. I think the cleanest solution is to throw in this case.

I was thinking of returning `OptionalInt.of(Integer.MAX_VALUE)` and checking for it in the `BatchAccumulator`. With this solution we can't distinguish between the current batch plus the new records exceed this constraint. And the new records that are being added to an empty batch exceed this constraint.

I think throwing an exception is fine since this is a fatal error. This mean that we have at least `Integer.MAX_VALUE` which would most like result in a OOM exception somewhere else in the system.

hachikuji: Since we are handling this case by raising an exception, is it worth checking for overflow of `bytesNeeded` as well?

raft/src/main/java/org/apache/kafka/raft/internals/BatchAccumulator.java
```diff
@@ -298,28 +342,30 @@ public void close() {
...
+        // the MemoryRecords (data) object.
+        private final ByteBuffer pooledBuffer;
 
         private CompletedBatch(
             long baseOffset,
             List<T> records,
             MemoryRecords data,
             MemoryPool pool,
-            ByteBuffer buffer
+            ByteBuffer pooledBuffer
```
hachikuji: nit: might be worth using the same name in `BatchBuilder`. Currently we use `initialBuffer`

raft/src/test/java/org/apache/kafka/raft/internals/BatchAccumulatorTest.java
```diff
@@ -164,47 +170,84 @@ public void testUnflushedBuffersReleasedByClose() {
...
+                appender.call(acc, leaderEpoch, singletonList("foo"));
+            }
+
+            List<BatchAccumulator.CompletedBatch<String>> batches = acc.drain();
+            assertEquals(4, batches.size());
+            assertTrue(batches.stream().allMatch(batch -> batch.data.sizeInBytes() <= maxBatchSize));
+        });
+    }
+
+    @Test void testRecordsAreSplit() {
```
hachikuji: nit: it's a little weird for this test alone to have the annotation on the same line as the method definition

raft/src/main/java/org/apache/kafka/raft/internals/BatchBuilder.java
jsancio: Updated it to use `Math.addExact`.

hachikuji (approved):
LGTM

jsancio:
Running the following commands
```
./gradlew -version

------------------------------------------------------------
Gradle 6.8.1
------------------------------------------------------------

Build time:   2021-01-22 13:20:08 UTC
Revision:     31f14a87d93945024ab7a78de84102a3400fa5b2

Kotlin:       1.4.20
Groovy:       2.5.12
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          1.8.0_282 (Private Build 25.282-b08)
OS:           Linux 5.8.0-7642-generic amd64

./gradlew -PscalaVersion=2.12 clean compileJava compileScala compileTestJava compileTestScala spotlessScalaCheck checkstyleMain checkstyleTest spotbugsMain rat --profile --no-daemon --continue -PxmlSpotBugsReport=true

./gradlew -PscalaVersion=2.13 clean compileJava compileScala compileTestJava compileTestScala spotlessScalaCheck checkstyleMain checkstyleTest spotbugsMain rat --profile --no-daemon --continue -PxmlSpotBugsReport=true

./gradlew -PscalaVersion=2.12 unitTest integrationTest --profile --no-daemon --continue -PtestLoggingEvents=started,passed,skipped,failed -PignoreFailures=true -PmaxTestRetries=1 -PmaxTestRetryFailures=5
```

jsancio:
The following tests failed:
```
core:

    ConfigCommandTest. shouldFailIfUnresolvableHost()
    ConfigCommandTest. shouldFailIfUnresolvableHost()
    ConfigCommandTest. shouldFailIfUnresolvableHostUsingZookeeper()
    ConfigCommandTest. shouldFailIfUnresolvableHostUsingZookeeper()
    DynamicConfigChangeTest. testIpHandlerUnresolvableAddress()
    DynamicConfigChangeTest. testIpHandlerUnresolvableAddress()
    DynamicConfigTest. shouldFailIpConfigsWithBadHost()
    DynamicConfigTest. shouldFailIpConfigsWithBadHost()
```

PR #10066, opened 2021-02-05
KAFKA-12278; Ensure exposed api versions are consistent within listener scopes

hachikuji:
With KIP-500, we have more complex requirements on API accessibility. Previously all APIs were accessible on every listener exposed by the broker, but now that is no longer true. For example:

- the controller exposes some APIs which are not accessible on the broker listener (e.g. quorum/registration/heartbeat APIs)
- most of the client APIs are not exposed on the controller (e.g. consumer group apis)
- there are some APIs which are not implemented by the KIP-500 broker (e.g. `LeaderAndIsr` and `UpdateMetadata`)
- there are some APIs which are only implemented by the KIP-500 broker (e.g. `DecommissionBroker` and `DescribeQuorum`)

All of this means that we need more sophistication in how we expose APIs and keep them consistent with the `ApiVersions` API. Up to now, we have been working around this using the `controllerOnly` flag inside `ApiKeys`, but this is not rich enough to support all of the cases listed above.

In this patch, we address this by problem by introducing a new `listeners` field to the request schema definitions. This field is an array of strings which indicate the listener types in which the API should be exposed. We currently support the following listener types: 

- `zkBroker`: old broker
- `broker`: kip-500 broker
- `controller`: kip-500 controller

For example, the `DecommissionBroker` API has the following listeners tag:
```json
  "listeners": ["broker", "controller"]
```
This indicates that the API is only on the KIP-500 broker and controller (both are needed because the request will be sent by clients and forwarded to the controller).

The patch changes the generator so that the listener type definitions are added to `ApiMessageType` and exposed through convenient helpers. At the same time, we have removed the `controllerOnly` flag from `ApiKeys` since now we can identify all controller APIs through the "controller" tag.

The rest of the patch is dedicated to ensuring that the API listener type is properly set and that the APIs accordingly tagged are carried through to the `ApiVersionsResponse`. We have created a new `ApiVersionManager` which encapsulates the creation of the `ApiVersionsResponse` based on the listener type. Additionally, `SocketServer` is modified to ensure the listener type of received requests before forwarding them to the request handler.

We have also fixed a bug in the handling of the `ApiVersionsResponse` prior to authentication. Previously a static response was sent, which means that changes to features would not get reflected. This also meant that the logic to ensure that only the intersection of version ranges supported by the controller would get exposed did not work. I think this is important because some clients rely on the initial pre-authenticated `ApiVersions` response rather than doing a second round after authentication as the Java client does.

One final cleanup note: I have removed the expectation that envelope requests are only allowed on "privileged" listeners. This made sense initially because we expected to use forwarding before the KIP-500 controller was available. That is not the case anymore and we expect the `Envelope` API to only be exposed on the controller listener. I have nevertheless preserved the existing workarounds to allow verification of the forwarding behavior in integration testing.

### DIFF
--- a/checkstyle/import-control.xml
+++ b/checkstyle/import-control.xml
@@ -50,6 +50,7 @@
 
   <subpackage name="common">
     <allow class="org.apache.kafka.clients.consumer.ConsumerRecord" exact-match="true" />
+    <allow class="org.apache.kafka.common.message.ApiMessageType" exact-match="true" />
     <disallow pkg="org.apache.kafka.clients" />
     <allow pkg="org.apache.kafka.common" exact-match="true" />
     <allow pkg="org.apache.kafka.common.annotation" />
@@ -108,6 +109,7 @@
       <allow pkg="org.apache.kafka.common.config" />
       <allow pkg="org.apache.kafka.common.metrics" />
       <allow pkg="org.apache.kafka.common.security" />
+      <allow class="org.apache.kafka.common.requests.ApiVersionsResponse" />
     </subpackage>
 
     <subpackage name="resource">
--- a/clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
+++ b/clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
@@ -207,8 +207,8 @@ public enum ApiKeys {
         return messageType.responseHeaderVersion(apiVersion);
     }
 
-    public boolean inScope(ApiMessageType.ApiScope scope) {
-        return messageType.scope().contains(scope);
+    public boolean inScope(ApiMessageType.ListenerType listener) {
+        return messageType.listeners().contains(listener);
     }
 
     private static String toHtml() {
@@ -251,12 +251,12 @@ public enum ApiKeys {
     }
 
     public static EnumSet<ApiKeys> zkBrokerApis() {
-        return apisInScope(ApiMessageType.ApiScope.ZK_BROKER);
+        return apisForListener(ApiMessageType.ListenerType.ZK_BROKER);
     }
 
-    public static EnumSet<ApiKeys> apisInScope(ApiMessageType.ApiScope scope) {
+    public static EnumSet<ApiKeys> apisForListener(ApiMessageType.ListenerType listener) {
         List<ApiKeys> controllerApis = Arrays.stream(ApiKeys.values())
-            .filter(apiKey -> apiKey.messageType.scope().contains(scope))
+            .filter(apiKey -> apiKey.messageType.listeners().contains(listener))
             .collect(Collectors.toList());
         return EnumSet.copyOf(controllerApis);
     }
--- a/clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
+++ b/clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
@@ -23,6 +23,7 @@ import org.apache.kafka.common.record.RecordBatch;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.EnumMap;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
@@ -105,6 +106,15 @@ public enum ApiKeys {
     BROKER_REGISTRATION(ApiMessageType.BROKER_REGISTRATION, true, RecordBatch.MAGIC_VALUE_V0, false),
     BROKER_HEARTBEAT(ApiMessageType.BROKER_HEARTBEAT, true, RecordBatch.MAGIC_VALUE_V0, false),
     UNREGISTER_BROKER(ApiMessageType.UNREGISTER_BROKER, false, RecordBatch.MAGIC_VALUE_V0, true);
+
+    private static final Map<ApiMessageType.ListenerType, EnumSet<ApiKeys>> APIS_BY_LISTENER =
+        new EnumMap<>(ApiMessageType.ListenerType.class);
+
+    static {
+        for (ApiMessageType.ListenerType listenerType : ApiMessageType.ListenerType.values()) {
+            APIS_BY_LISTENER.put(listenerType, filterApisForListener(listenerType));
+        }
+    }
 
     // The generator ensures every `ApiMessageType` has a unique id
     private static final Map<Integer, ApiKeys> ID_TO_TYPE = Arrays.stream(ApiKeys.values())
@@ -255,6 +265,10 @@ public enum ApiKeys {
     }
 
     public static EnumSet<ApiKeys> apisForListener(ApiMessageType.ListenerType listener) {
+        return APIS_BY_LISTENER.get(listener);
+    }
+
+    private static EnumSet<ApiKeys> filterApisForListener(ApiMessageType.ListenerType listener) {
         List<ApiKeys> controllerApis = Arrays.stream(ApiKeys.values())
             .filter(apiKey -> apiKey.messageType.listeners().contains(listener))
             .collect(Collectors.toList());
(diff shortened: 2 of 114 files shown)


checkstyle/import-control.xml
```diff
@@ -99,6 +100,7 @@
       <allow pkg="org.apache.kafka.common.config" />
       <allow pkg="org.apache.kafka.common.metrics" />
       <allow pkg="org.apache.kafka.common.security" />
+      <allow class="org.apache.kafka.common.requests.ApiVersionsResponse" />
```
ijuma, line 112: The network package is not meant to access requests, right? Could we make it work by changing the supplier to something more generic?

generator/src/main/java/org/apache/kafka/message/RequestApiScope.java
```diff
@@ -0,0 +1,33 @@
...
+    ZK_BROKER,
+
+    @JsonProperty("broker")
+    BROKER,
+
+    @JsonProperty("controller")
+    CONTROLLER,
+
+    @JsonProperty("raft")
+    RAFT;
```
ijuma: If we're talking about scopes, wouldn't it be more intuitive if one of them was `CLIENT`? Since these definitions go into the protocol definition files that are used by other clients, this could be used to differentiate between protocol apis that clients need to support versus protocol apis that are used for inter broker and broker -> controller communication.

checkstyle/import-control.xml
hachikuji: Hmm.. Do you have any ideas? I guess I would say that the dependency is there regardless of how we choose to hide it since we have to go through `ChannelBuilders` to build the authenticator, and the authenticator does depend on the request APIs.

generator/src/main/java/org/apache/kafka/message/RequestApiScope.java
hachikuji: I had something like that originally, but I found it simpler to have a 1-1 mapping from listener to scope. All "CLIENT" APIs have to be present on both the zk and raft brokers, but the "CLIENT" scope alone doesn't give us a way to distinguish between the two. We could have "ZK_CLIENT" and "RAFT_CLIENT" as an alternative. Then I guess we would also still need "ZK_BROKER" and "RAFT_BROKER." In any case, then each listener might have multiple scopes, which is a little more complex.

Perhaps this would be clearer if we had a name tying this back the listener more directly. Maybe we should get rid of the "scope" term and just call it "listeners"?

```json
"listeners": ["broker", "controller"]
```

By the way, you probably saw it already, but the "raft" scope was added here just for convenience in `TestRaftServer`. I could use the "controller" scope instead and remove this.

ijuma: These are good points. I think if we keep the approach suggested here, `listeners` may be better. However, I am still not sure if it would not be better to have `ZK_CLIENT` and `RAFT_CLIENT` (or something along those lines). Then, as we drop support for ZK, it would work pretty nicely. Generally, I think the json files would ideally be our protocol definition versus something where we capture implementation details.

cmccabe: @ijuma : unfortunately we don't have a clear separation between clients and brokers at the protocol level.  As you know, if a random node connects to the broker and asks for ApiVersions, the broker doesn't (yet) know if the other node is another broker or if it's a client.  So it wouldn't help to label APIs as CLIENT vs. BROKER, since the broker has to send back both anyway.

On the other hand, we can definitely put APIs into "zk broker", "kip-500 zk", and "kip500 controller" and use those buckets to figure out what to send in the ApiVersionsRequest.

@hachikuji : The "raft" scope seems a little weird since all the other scopes map to listeners on servers.  The TestRaftServer is just an internal junit thing, right?  So it's probably fine if it just uses the Controller scope and returns `UnsupportedVersionException` for things it doesn't support.  I don't think we should have a separate scope for this unless it somehow impacts the real servers.

hachikuji: Ok, I will get rid of it.

hachikuji:
@cmccabe @ijuma This is ready for another look. I have changed `scope` to `listeners` and I removed the "raft" listener type.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -246,10 +250,15 @@ public void visit(Type field) {
...
     }
 
-    public static List<ApiKeys> brokerApis() {
-        return Arrays.stream(values())
-            .filter(api -> !api.isControllerOnlyApi)
+    public static EnumSet<ApiKeys> zkBrokerApis() {
+        return apisForListener(ApiMessageType.ListenerType.ZK_BROKER);
+    }
+
+    public static EnumSet<ApiKeys> apisForListener(ApiMessageType.ListenerType listener) {
```
cmccabe, line 267: given that these `EnumSets` are used so much throughout the code, we should just calculate them once and reuse them, rather than recalculating each time.  There are only 3 listener types so it could just be in a static map or something.

core/src/main/scala/kafka/server/ApiVersionManager.scala
```diff
@@ -0,0 +1,126 @@
...
+  def apiVersionResponse(throttleTimeMs: Int): ApiVersionsResponse
+  def isApiEnabled(apiKey: ApiKeys): Boolean = enabledApis.contains(apiKey)
+  def newRequestMetrics: RequestChannel.Metrics = new network.RequestChannel.Metrics(enabledApis)
+}
+
+object ApiVersionManager {
+  def apply(
+    listenerType: ListenerType,
+    config: KafkaConfig,
+    forwardingManager: Option[ForwardingManager],
```
cmccabe, line 41: I would really prefer not to have all this `ForwardingManager` stuff in here.  The `ControllerServer` has been committed to trunk and does handle `ENVELOPE_REQUEST` so I think this is out of date.

It should be as simple as controllers handle envelope requests, brokers don't, right?

checkstyle/import-control.xml
cmccabe: > I guess I would say that the dependency is there regardless of how we choose to hide it since we have to go through ChannelBuilders to build the authenticator, and the authenticator does depend on the request APIs.

Yeah, the API versions response is pretty special in the protocol. Fortunately or unfortunately, it's not treated as a generic response.  I tried to think of some ways to avoid this dependency but they all ended up being kind of like obfuscation.

> Could we make it work by changing the supplier to something more generic?

Maybe I'm misinterpreting the suggestion, but if we choose to make this a `Supplier<AbstractResponse>`, that's still in the requests package....

core/src/main/scala/kafka/server/ApiVersionManager.scala
hachikuji: The `ForwardingManager` is used by the broker. The reason we depend on it here is because of the need to intersect the controller api versions.

I agree with you about the envelope handling. I was planning to submit a follow-up to remove this logic from `KafkaApis`. As you said, only the controller needs to be able to handle the Envelope API. Doing so here, however, would add a significant diff.

clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
hachikuji: Makes sense.

checkstyle/import-control.xml
ijuma: I meant that you could introduce an interface that the response implements. That interface would live in the network package or in the common package.

cmccabe (approved):
LGTM

core/src/main/scala/kafka/server/ApiVersionManager.scala
cmccabe: ok

checkstyle/import-control.xml
cmccabe: Hmm.  Let's revisit this after 2.8.

cmccabe:
pushed

PR #10069, opened 2021-02-05
MINOR: Add RaftReplicaManager

rondagostino:
This adds the logic to apply partition metadata when consuming from the Raft-based metadata log.

`RaftReplicaManager` extends `ReplicaManager` for now to minimize changes to existing code for the 2.8 release.  We will likely adjust this hierarchy at a later time (e.g. introducing a trait and adding a helper to refactor common code).  For now, we expose the necessary fields and methods in `ReplicaManager` by changing their scope from `private` to `protected`, and we refactor out a couple of pieces of logic that are shared between the two implementation (stopping replicas and adding log dir fetchers).

Existing tests are sufficient to expose regressions in the current `ReplicaManager`.

We intend to exercise the new `RaftReplicaManager` code via system tests and unit/integration tests (both to come in later PRs).

### DIFF
--- a/core/src/main/scala/kafka/server/DelayedDeleteRecords.scala
+++ b/core/src/main/scala/kafka/server/DelayedDeleteRecords.scala
@@ -84,7 +84,7 @@ class DelayedDeleteRecords(delayMs: Long,
                 (false, Errors.NOT_LEADER_OR_FOLLOWER, DeleteRecordsResponse.INVALID_LOW_WATERMARK)
             }
 
-          case HostedPartition.Deferred(_, _, _, _, _) =>
+          case _: HostedPartition.Deferred =>
             (false, Errors.UNKNOWN_TOPIC_OR_PARTITION, DeleteRecordsResponse.INVALID_LOW_WATERMARK)
 
           case HostedPartition.Offline =>
--- a/core/src/main/scala/kafka/server/DelayedDeleteRecords.scala
+++ b/core/src/main/scala/kafka/server/DelayedDeleteRecords.scala
@@ -84,7 +84,7 @@ class DelayedDeleteRecords(delayMs: Long,
                 (false, Errors.NOT_LEADER_OR_FOLLOWER, DeleteRecordsResponse.INVALID_LOW_WATERMARK)
             }
 
-          case HostedPartition.Deferred(_) =>
+          case HostedPartition.Deferred(_, _, _, _, _) =>
             (false, Errors.UNKNOWN_TOPIC_OR_PARTITION, DeleteRecordsResponse.INVALID_LOW_WATERMARK)
 
           case HostedPartition.Offline =>
--- a/core/src/main/scala/kafka/server/RaftReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/RaftReplicaManager.scala
@@ -225,7 +225,7 @@ class RaftReplicaManager(config: KafkaConfig,
               None
 
             case HostedPartition.Online(partition) => Some(partition)
-            case HostedPartition.Deferred(_, _, _, _, _) => throw new IllegalStateException(
+            case _: HostedPartition.Deferred => throw new IllegalStateException(
               s"There should never be deferred partition metadata when we aren't deferring changes: $topicPartition")
 
             case HostedPartition.None =>
@@ -536,7 +536,7 @@ class RaftReplicaManager(config: KafkaConfig,
   // after the iterator has been constructed could still be returned by this iterator.
   private def deferredPartitionsIterator: Iterator[HostedPartition.Deferred] = {
     allPartitions.values.iterator.flatMap {
-      case deferred@HostedPartition.Deferred(_, _, _, _, _) => Some(deferred)
+      case deferred: HostedPartition.Deferred => Some(deferred)
       case _ => None
     }
   }
--- a/core/src/main/scala/kafka/server/RaftReplicaManager.scala
+++ b/core/src/main/scala/kafka/server/RaftReplicaManager.scala
@@ -86,7 +86,6 @@ class RaftReplicaManager(config: KafkaConfig,
         val leaderPartitionStates = mutable.Map[Partition, MetadataPartition]()
         val followerPartitionStates = mutable.Map[Partition, MetadataPartition]()
         val partitionsAlreadyExisting = mutable.Set[MetadataPartition]()
-        val mostRecentMetadataOffsets = mutable.Map[Partition, Long]()
         deferredPartitionsIterator.foreach { deferredPartition =>
           val state = deferredPartition.metadata
           val partition = deferredPartition.partition
@@ -98,14 +97,13 @@ class RaftReplicaManager(config: KafkaConfig,
           if (!deferredPartition.isNew) {
             partitionsAlreadyExisting += state
           }
-          mostRecentMetadataOffsets.put(partition, deferredPartition.mostRecentMetadataOffset)
         }
 
         val partitionsMadeLeader = makeLeaders(partitionsAlreadyExisting, leaderPartitionStates,
-          highWatermarkCheckpoints,-1, mostRecentMetadataOffsets)
+          highWatermarkCheckpoints, 0)
         val partitionsMadeFollower = makeFollowers(partitionsAlreadyExisting,
           createMetadataBrokersFromCurrentCache, followerPartitionStates,
-          highWatermarkCheckpoints, -1, mostRecentMetadataOffsets)
+          highWatermarkCheckpoints, 0)
 
         // We need to transition anything that hasn't transitioned from Deferred to Offline to the Online state.
         // We also need to identify the leadership change callback(s) to invoke
@@ -142,7 +140,7 @@ class RaftReplicaManager(config: KafkaConfig,
             val state = metadata.metadata
             val partition = metadata.partition
             val topicPartition = partition.topicPartition
-            val mostRecentMetadataOffset = metadata.mostRecentMetadataOffset
+            val mostRecentMetadataOffset = state.largestDeferredOffsetEverSeen
             val leader = state.leaderId == localBrokerId
             val leaderOrFollower = if (leader) "leader" else "follower"
             val partitionLogMsgPrefix = s"Apply deferred $leaderOrFollower partition $topicPartition last seen in metadata batch $mostRecentMetadataOffset"
@@ -192,6 +190,7 @@ class RaftReplicaManager(config: KafkaConfig,
         val partitionChangesToBeDeferred = mutable.HashMap[Partition, (MetadataPartition, Option[HostedPartition.Deferred])]()
         builder.localChanged().foreach { currentState =>
           val topicPartition = new TopicPartition(currentState.topicName, currentState.partitionIndex)
+          sanityCheckStateDeferredAtOffset(topicPartition, currentState, metadataOffset)
           val (partition, priorDeferredMetadata) = getPartition(topicPartition) match {
             case HostedPartition.Offline =>
               stateChangeLogger.warn(s"Ignoring handlePartitionChanges at $metadataOffset " +
@@ -200,7 +199,7 @@ class RaftReplicaManager(config: KafkaConfig,
               (None, None)
 
             case HostedPartition.Online(partition) => (Some(partition), None)
-            case deferred@HostedPartition.Deferred(partition, _, _, _, _) => (Some(partition), Some(deferred))
+            case deferred@HostedPartition.Deferred(partition, _, _, _) => (Some(partition), Some(deferred))
 
             case HostedPartition.None =>
               // Create the partition instance since it does not yet exist
@@ -215,8 +214,9 @@ class RaftReplicaManager(config: KafkaConfig,
       } else { // not deferring changes, so make leaders/followers accordingly
         val partitionsToBeLeader = mutable.HashMap[Partition, MetadataPartition]()
         val partitionsToBeFollower = mutable.HashMap[Partition, MetadataPartition]()
-        builder.localChanged().foreach { state =>
-          val topicPartition = new TopicPartition(state.topicName, state.partitionIndex)
+        builder.localChanged().foreach { currentState =>
+          val topicPartition = new TopicPartition(currentState.topicName, currentState.partitionIndex)
+          sanityCheckStateNotDeferred(topicPartition, currentState)
           val partition = getPartition(topicPartition) match {
             case HostedPartition.Offline =>
               stateChangeLogger.warn(s"Ignoring handlePartitionChanges at $metadataOffset " +
@@ -234,10 +234,10 @@ class RaftReplicaManager(config: KafkaConfig,
               Some(partition)
           }
           partition.foreach { partition =>
-            if (state.leaderId == localBrokerId) {
-              partitionsToBeLeader.put(partition, state)
+            if (currentState.leaderId == localBrokerId) {
+              partitionsToBeLeader.put(partition, currentState)
             } else {
-              partitionsToBeFollower.put(partition, state)
+              partitionsToBeFollower.put(partition, currentState)
             }
           }
         }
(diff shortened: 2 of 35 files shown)


core/src/main/scala/kafka/server/RaftReplicaManager.scala
```diff
@@ -0,0 +1,538 @@
...
+          if (!deferredPartition.isNew) {
+            partitionsAlreadyExisting += state
+          }
+          mostRecentMetadataOffsets.put(partition, deferredPartition.mostRecentMetadataOffset)
+        }
+
+        val partitionsMadeLeader = makeLeaders(partitionsAlreadyExisting, leaderPartitionStates,
+          highWatermarkCheckpoints,-1, mostRecentMetadataOffsets)
+        val partitionsMadeFollower = makeFollowers(partitionsAlreadyExisting,
+          createMetadataBrokersFromCurrentCache, followerPartitionStates,
```
rondagostino: > Do we have any guarantee that the metadata cache is in a state that is consistent with the deferred changes?

Good question.  Metadata changes up through the point of applying the deferred partition metadata changes should be applied to the metadata cache at this point.

One thing we need to think about is the fact that we currently don't defer metadata cache changes at all.  The metadata cache will contain partition states that are ahead of ReplicaManager during the time when ReplicaManager is deferring its changes.  This means, for example, that the following will reflect deferred partition changes that have been applied to the metadata cache but that have not been applied to ReplicaManager.  We may have to write test cases for each of these conditions so we can be clear on what the expected behavior should be.

- MetadataRequest
- FindCoordinatorRequest
- ElectLeadersRequest with topicPartitions = null
- DelayedCreatePartitions (in topic purgatory)
- DelayedElectLeader (in elect leader purgatory)
- Anything that calls ReplicaManager.fetchMessages() and DelayedFetch (in fetch purgatory), though these seem okay since they wait until they can get enough data?
- TransactionMarkerChannelManager.addTxnMarkersToBrokerQueue
- DescribeConfigsRequest
- OffsetCommitRequest (whether or not to send UNKNOWN_TOPIC_OR_PARTITION)
- ProduceRequest (whether or not to send UNKNOWN_TOPIC_OR_PARTITION)
- FetchRequest (whether or not to send UNKNOWN_TOPIC_OR_PARTITION)
- DeleteRecordsRequest (whether or not to send UNKNOWN_TOPIC_OR_PARTITION)
- AddPartitionsToTxnRequest (whether or not to send UNKNOWN_TOPIC_OR_PARTITION)
- TxnOffsetCommitRequest (whether or not to send UNKNOWN_TOPIC_OR_PARTITION)
- OffsetDeleteRequest (whether or not to send UNKNOWN_TOPIC_OR_PARTITION)

core/src/main/scala/kafka/server/RaftReplicaManager.scala
```diff
@@ -0,0 +1,538 @@
...
+            val mostRecentMetadataOffset = metadata.mostRecentMetadataOffset
+            val leader = state.leaderId == localBrokerId
+            val leaderOrFollower = if (leader) "leader" else "follower"
+            val partitionLogMsgPrefix = s"Apply deferred $leaderOrFollower partition $topicPartition last seen in metadata batch $mostRecentMetadataOffset"
+            stateChangeLogger.error(s"$partitionLogMsgPrefix: error while applying deferred metadata.", e)
+          }
+          stateChangeLogger.info(s"Applied ${partitionsMadeLeader.size + partitionsMadeFollower.size} deferred partitions prior to the error: " +
+            s"${partitionsMadeLeader.size} leader(s) and ${partitionsMadeFollower.size} follower(s)")
+          // Re-throw the exception for it to be caught in BrokerMetadataListener
+          throw e
```
rondagostino, line 195: > If we fail to apply changes, I guess we have to see that as a fatal error? The only possible way of recovering would be to replay the changes.

Yeah, I think so.  We may need to put effort into minimizing the blast radius of these failures.

core/src/main/scala/kafka/server/DelayedDeleteRecords.scala
```diff
@@ -84,7 +84,7 @@ class DelayedDeleteRecords(delayMs: Long,
                 (false, Errors.NOT_LEADER_OR_FOLLOWER, DeleteRecordsResponse.INVALID_LOW_WATERMARK)
             }
 
-          case HostedPartition.Deferred(_) =>
+          case HostedPartition.Deferred(_, _, _, _, _) =>
```
ijuma: It's cleaner to use `case _: HostedPartition.Deferred`. Otherwise, every time a parameter is added, you have to update this for no good reason.

core/src/main/scala/kafka/server/RaftReplicaManager.scala
rondagostino: I added the partition state with respect to deferral to `MetadataPartitions`.  Storing the information in the metadat cache could help here.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -248,19 +250,19 @@ class ReplicaManager(val config: KafkaConfig,
...
+  protected val localBrokerId = config.brokerId
+  protected val allPartitions = new Pool[TopicPartition, HostedPartition](
     valueFactory = Some(tp => HostedPartition.Online(Partition(tp, time, configRepository, this)))
   )
-  private val replicaStateChangeLock = new Object
+  protected val replicaStateChangeLock = new Object
   val replicaFetcherManager = createReplicaFetcherManager(metrics, time, threadNamePrefix, quotaManagers.follower)
   val replicaAlterLogDirsManager = createReplicaAlterLogDirsManager(quotaManagers.alterLogDirs, brokerTopicStats)
   private val highWatermarkCheckPointThreadStarted = new AtomicBoolean(false)
   @volatile var highWatermarkCheckpoints: Map[String, OffsetCheckpointFile] = logManager.liveLogDirs.map(dir =>
```
ijuma: Is this intended to be public? Maybe we should fix it while we're at it.

ijuma:
@rondagostino Looks like a file is missing the license:

> 19:18:36  Execution failed for task ':rat'.
> 19:18:36  > Found 1 files with unknown licenses.

ijuma:
All builds are green!

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -439,71 +414,85 @@ class ReplicaManager(val config: KafkaConfig,
...
+    replicaFetcherManager.removeFetcherForPartitions(partitions)
+    replicaAlterLogDirsManager.removeFetcherForPartitions(partitions)
+
+    // Second remove deleted partitions from the partition map. Fetchers rely on the
+    // ReplicaManager to get Partition's information so they must be stopped first.
+    val partitionsToDelete = mutable.Set.empty[TopicPartition]
+    partitionsToStop.forKeyValue { (topicPartition, shouldDelete) =>
+      if (shouldDelete) {
+        getPartition(topicPartition) match {
+          case hostedPartition: NonOffline => // Online or Deferred (Deferred never occurs when using ZooKeeper)
```
ijuma: This comment seems a bit redundant, no? We should have such documentation in the definition of `NonOffline` and avoid duplicating it everywhere it's used.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -439,71 +414,85 @@ class ReplicaManager(val config: KafkaConfig,
...
+      // If we were the leader, we may have some operations still waiting for completion.
+      // We force completion to prevent them from timing out.
+      completeDelayedFetchOrProduceRequests(topicPartition)
+    }
+
+    // Third delete the logs and checkpoint.
+    val errorMap = new mutable.HashMap[TopicPartition, Throwable]()
+    if (partitionsToDelete.nonEmpty) {
+      // Delete the logs and checkpoint. Confusingly, this function isn't actually
+      // asynchronous-- it just synchronously schedules the directories to be deleted later.
```
ijuma: This comment is a bit confusing. It synchronously schedules and asynchronously deletes it. Why is the function not asynchronous? I think you mean that it doesn't give you a future or something like that, but that is not a requirement for a function to be asynchronous, right?

core/src/test/scala/unit/kafka/server/RaftReplicaManagerTest.scala
```diff
@@ -0,0 +1,76 @@
...
+package kafka.server
+
+import java.io.File
+import java.util.concurrent.atomic.AtomicBoolean
+
+import kafka.server.QuotaFactory.QuotaManagers
+import kafka.server.metadata.{CachedConfigRepository, RaftMetadataCache}
+import kafka.utils.{MockScheduler, MockTime, TestUtils}
+import org.apache.kafka.common.metrics.Metrics
+import org.easymock.EasyMock
```
ijuma: Can we use `Mockito` instead of `EasyMock`?

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -30,7 +31,8 @@ import scala.jdk.CollectionConverters._
 
 
 object MetadataPartition {
-  def apply(name: String, record: PartitionRecord): MetadataPartition = {
+  val OffsetNeverDeferred = 0L // must not be a valid offset we could see (i.e. must not be positive)
```
cmccabe: Can you add JavaDoc for this?

Also, what about `NoDeferredOffset` as a name?

One last question... why is this 0 and not -1?  0 is a valid offset in the log, whereas -1 is not.

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
```diff
@@ -39,11 +41,30 @@ object MetadataPartition {
       record.isr(),
       Collections.emptyList(), // TODO KAFKA-12285 handle offline replicas
       Collections.emptyList(),
-      Collections.emptyList())
+      Collections.emptyList(),
+      largestDeferredOffsetEverSeen = deferredAtOffset.getOrElse(OffsetNeverDeferred),
+      isCurrentlyDeferringChanges = deferredAtOffset.isDefined)
```
cmccabe: Hmm... why do we need this boolean?  Can't we just check if `largestDeferredOffsetEverSeen` is not `OffsetNeverDeferred`

cmccabe:
Thanks for this PR, @rondagostino !

I can see why you wanted to have `RaftReplicaChangeDelegateHelper`.  The `ReplicaManager` is not very easy to unit test because it has grown so large.  I don't think this delegate thing is quite the right abstraction here-- it's pretty confusing-- but I guess let's revisit this after 2.8 is finished.

I suppose one option is, once `ReplicaManager` is a pure interface, we can split the kip-500 update logic off into a separate set of functions that takes a `ReplicaManager` as an input.  Then we can easily unit-test the update logic with a `MockReplicaManager`.

For now I left two small comments... LGTM after those are addressed.

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
rondagostino: We basically use `largestDeferredOffsetEverSeen` only for logging at this point -- we also check it in a few `private def sanityCheckState...()` `RaftReplicaManager` methods.  We could completely eliminate `largestDeferredOffsetEverSeen` if we didn't want to log when the partition was last deferred.  It just tracks when the partition was last seen and the change at that offset was deferred rather than directly applied.  Once the partition is no longer deferred the value remains whatever it was and the boolean flips to `false`.

It does seem on the surface that we could change the declaration to `deferredSinceOffset` and get rid of the boolean -- and `deferredSinceOffset` would change to `-1` once those changes are applied.  But there is a problem with this if the partition changes to not being deferred in the metadata cache before we ask `RaftReplicaManager` to process all of its deferred changes: the value will be -1 in the metadata cache under those circumstances, and we wouldn't have the value to log.

So I think we have a few options.

1. Do the logging, apply the changes to the matadata cache before replica manager, and keep the `Long` and `Boolean` as currently defined
2. Do the logging, apply the changes to the matadata cache **after** replica manager, and use just a `Long` (with the semantics being changed as described above)
3. Just use a Boolean and don't do the logging.

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
rondagostino: As discussed offline, we will remove this and not include the last seen offset in log messages when applying deferred changes.

core/src/main/scala/kafka/server/metadata/MetadataPartitions.scala
rondagostino: As discussed offline, we will eliminate the information from the messages we log when applying deferred changes, and we won't carry that info around in `MetadataPartition`.  Currently `RaftReplicaManager` knows if it is deferring changes or not.  Maybe later when we get `BrokerLifecycleManager` and `BrokerMetadataListener` committed we can think about where a global boolean might live to identify if the broker is fenced or not.  It isn't critical to decide right now because we are only going to defer the application of partition metadata at startup in 2.8.

rondagostino (commented):
@cmccabe I pushed a commit getting rid of the information from `MetadataPartition` and removing the information from the log messages.

core/src/main/scala/kafka/server/RaftReplicaManager.scala
```diff
@@ -0,0 +1,397 @@
...
+            }
+          }
+        }
+
+        val prevPartitions = imageBuilder.prevImage.partitions
+        val changedPartitionsPreviouslyExisting = mutable.Set[MetadataPartition]()
+        builder.localChanged().foreach(metadataPartition =>
+          prevPartitions.topicPartition(metadataPartition.topicName, metadataPartition.partitionIndex).foreach(
+            changedPartitionsPreviouslyExisting.add))
+        val nextBrokers = imageBuilder.nextBrokers()
```
cmccabe: rather than doing this, I would prefer to just give the ReplicaManager access to the BrokersBuilder and put the relevant accessors on that.  Otherwise, we create a bunch of objects that get thrown away (since nextBrokers will have to be called again later).  A small thing but it would be good to set up a good pattern here.

core/src/main/scala/kafka/server/ReplicaManager.scala
```diff
@@ -185,7 +185,9 @@ object HostedPartition {
    * This state only applies to brokers that are using a Raft-based metadata
    * quorum; it never happens when using ZooKeeper.
    */
-  final case class Deferred(partition: Partition) extends NonOffline
+  final case class Deferred(partition: Partition,
+                            isNew: Boolean,
+                            onLeadershipChange: (Iterable[Partition], Iterable[Partition]) => Unit) extends NonOffline
```
cmccabe: I don't think this callback is logically part of the HostedPartition object itself.  We don't want a different callback for each partition.  It will the same one for each (in practice it just hooks into the group manager and one other manager that I forget).  So let's just supply this callback in whatever function it is needed in, not store it in the hostedpartition object.

cmccabe:
Thanks @rondagostino .  I left two small comments... LGTM after those are addressed.

core/src/main/scala/kafka/server/RaftReplicaManager.scala
rondagostino: > would prefer to just give the ReplicaManager access to the BrokersBuilder and put the relevant accessors on that. Otherwise, we create a bunch of objects that get thrown away (since nextBrokers will have to be called again later).

The point in time at which this invocation is occurring is when we are telling `RaftReplicaManager` to process the partition changes from a batch.  This happens after the batch has been otherwise handled by the metadata listener.  I think the most common occurrence for any batch is that there are no broker metadata changes (i.e. no broker has registered/unregistered or fenced/unfenced), in which case invoking `nextBrokers()` simply returns the previous brokers -- we don't create any objects.  If we make the change I think you are suggesting and `RaftReplicaManager` were to invoke `imageBuilder.brokersBuilder().build()` then we would actually invoke `MetadataBrokers(log, newBrokerMap)`, which is much more expensive.  Maybe you are suggesting `MetadataBrokersBuilder` should remember that it was already built?  It sort of does in that any attempt to mutate it after it is built will fail (since its map is now unmodifiable), but calling `build()` a second time could be made more efficient than it is now.

cmccabe: Thanks for the explanation!

Let's just rename this to "brokers" and then I think we're good.  (Everything in the builder is "next" after all :) )

rondagostino: As discussed offline, will keep it like it is but rename `nextBrokers()` to simply `brokers()`.

cmccabe (approved):

PR #10074, opened 2021-02-06
KAFKA-12305: Fix Flatten SMT for array types

C0urante:
[Jira](https://issues.apache.org/jira/browse/KAFKA-12305)

(Copied from Jira):

The `Flatten` SMT fails for array types. A sophisticated approach that tries to flatten arrays might be desirable in some cases, and may have been punted during the early design phase of the transform, but in the interim, it's probably not worth it to make array data and the SMT mutually exclusive.

A naive approach that preserves arrays as-are and doesn't attempt to flatten them seems fair for now.

### DIFF
--- a/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Flatten.java
+++ b/connect/transforms/src/main/java/org/apache/kafka/connect/transforms/Flatten.java
@@ -124,6 +124,7 @@ public abstract class Flatten<R extends ConnectRecord<R>> implements Transformat
                 case BOOLEAN:
                 case STRING:
                 case BYTES:
+                case ARRAY:
                     newRecord.put(fieldName(fieldNamePrefix, entry.getKey()), entry.getValue());
                     break;
                 case MAP:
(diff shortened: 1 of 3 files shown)


C0urante:
@gharris1727 @ncliang @tombentley anyone got time to take a look?

ncliang (approved):
Thanks @C0urante . LGTM

tombentley (commented):
Is it worth having a test to cover the recursive flatten case discussed in the JIRA? 

I think we should also elaborate on the behaviour for arrays in the SMT doc, so people know what the SMT is intended to do with arrays (it was never a problem before when the leaves were scalars).

C0urante:
Thanks @tombentley--good call on updating the docs; I've done that. RE a test for the recursive case--I don't think it'll hurt, so I updated the existing tests to include it. Probably not too valuable right now since arrays are just passed through unmodified but it might save someone else that bit of legwork in the future if we decide we want bona fide flattening for array types as well.

tombentley (approved):
Thanks @C0urante, LGTM. I'll merge it assuming the CI doesn't throw up anything.

C0urante:
@tombentley know it's been a while and we're probably past the boat for 3.0, but wanted to check in and see if there's anything blocking this PR from being merged at the moment?

tombentley:
@C0urante sorry for the delay.

PR #10077, opened 2021-02-07
KAFKA-12309 The revocation algorithm produces uneven distributions

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-12309

# issue description
When adding a new worker, the revocation algorithm tries to revoke some connectors/tasks from existent workers so as to move them to new worker to balance load. However, the algorithm can revoke incorrect number of connectors/tasks for following cases.

### Assignments:
"worker_0" -> 8 connectors/tasks
"worker_1" -> 8 connectors/tasks
(New) "worker_2" -> 0 connectors/tasks

### Result
"worker_0" -> 6 connectors/tasks
"worker_1" -> 6 connectors/tasks
"worker_2" -> 4 connectors/tasks

### (expected) Result
"worker_0" -> 6 connectors/tasks
"worker_1" -> 5 connectors/tasks
"worker_2" -> 5 connectors/tasks
or 
"worker_0" -> 5 connectors/tasks
"worker_1" -> 6 connectors/tasks
"worker_2" -> 5 connectors/tasks

# root cause
the algorithm makes all existent workers keep `ceil` (`numToRevoke = existing.tasksSize() - ceilTasks;`) connectors/tasks. That results in that there is no enough revoked connectors/tasks to be assigned to new worker.

# extra changes included by this PR

1. move revocation algorithm to a helper method in order to test it


connect/runtime/src/test/java/org/apache/kafka/connect/integration/RebalanceSourceConnectorsIntegrationTest.java
```diff
@@ -359,7 +359,9 @@ private boolean assertConnectorAndTasksAreUniqueAndBalanced() {
                     tasks.values().size(),
                     tasks.values().stream().distinct().collect(Collectors.toList()).size());
             assertTrue("Connectors are imbalanced: " + formatAssignment(connectors), maxConnectors - minConnectors < 2);
+            if (minConnectors > 1) assertEquals("Some workers have no connectors", connectors.size(), connect.workers().size());
```
chia7712, line 365: make sure there is no idle worker

chia7712:
@rhauch Could you take a look? this uneven distributions can be reproduced easily so it would be nice to fix it asap.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -562,50 +561,63 @@ private void resetDelay() {
...
+        int numberOfBiggers = numberOfActives % numberOfTotalWorks;
+        int ceil = numberOfBiggers == 0 ? floor : floor + 1;
         for (WorkerLoad existing : existingWorkers) {
-            Iterator<ConnectorTaskId> tasks = existing.tasks().iterator();
-            numToRevoke = existing.tasksSize() - ceilTasks;
-            log.debug("Tasks on worker {} is higher than ceiling, so revoking {} tasks", existing, numToRevoke);
-            for (int i = existing.tasksSize(); i > floorTasks && numToRevoke > 0; --i, --numToRevoke) {
+            int currentSize = forTask ? existing.tasksSize() : existing.connectorsSize();
+            int expectedSize;
+            if (existingWorkers.size() == 1 || currentSize == 1) expectedSize = ceil;
```
showuon: Why do we need this condition? And why under this case, we don't need to have `numberOfBiggers--`? After all, we assigned one ceiling to this worker, didn't we?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -562,50 +561,63 @@ private void resetDelay() {
...
-                resources.connectors().add(connectors.next());
-            }
-        }
+        return revoking;
+    }
 
+    static void computeRevoked(Map<String, ConnectorsAndTasks> revoking,
+                               Collection<WorkerLoad> existingWorkers,
+                               int numberOfTotalWorks,
+                               int numberOfActives,
```
showuon: Could we rename this parameters as `numberOfActiveTasks` to be more clear?

showuon: And they can be `final`

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -562,50 +561,63 @@ private void resetDelay() {
...
+                               boolean forTask) {
+        int floor = numberOfActives / numberOfTotalWorks;
+        int numberOfBiggers = numberOfActives % numberOfTotalWorks;
```
showuon: I believe you're trying to name it `numberOfRemainder`

showuon (commented):
@chia7712 , thanks for the patch! This is a good finding! Left some comments. Thanks.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
showuon: Oh, I got what you mean, maybe `numberOfCeilingMembers` would be more clear?

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
chia7712: That is a specify case as the node having single task is already in balance. We can move the job to another node but it does not balance anything but it brings extra down-time.

I will add comments for that case.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
chia7712: nice catch. will copy that.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
chia7712: ok. will copy that

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
```diff
@@ -562,50 +561,68 @@ private void resetDelay() {
...
-            Iterator<ConnectorTaskId> tasks = existing.tasks().iterator();
-            numToRevoke = existing.tasksSize() - ceilTasks;
-            log.debug("Tasks on worker {} is higher than ceiling, so revoking {} tasks", existing, numToRevoke);
-            for (int i = existing.tasksSize(); i > floorTasks && numToRevoke > 0; --i, --numToRevoke) {
+    static void computeRevoked(final Map<String, ConnectorsAndTasks> revoking,
+                               final Collection<WorkerLoad> existingWorkers,
+                               final int numberOfWorkers,
+                               final int numberOfActiveTasks,
+                               final boolean forTask) {
+        final int floor = numberOfActiveTasks / numberOfWorkers;
```
kkonstantine: Refactoring similar code to common methods is good and improves readability. 

But during code shuffling and especially if we add functional changes let's not apply also minor stylistic changes. 
We don't usually use `final` for local variables in Connect.

kkonstantine (commented):
Thanks for the PR @chia7712 

I'd like to return for a detailed review after we exclude stylistic changes that are not necessary. Also, it'd be good to enhance the description besides the example and call out the corner cases for which logic has been added.

connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/IncrementalCooperativeAssignor.java
kkonstantine: Let's not change local variables and arguments with `final` when there isn't this style in place. Sorry for the back and forth.

chia7712:
@kkonstantine thanks for your feedback.

> I'd like to return for a detailed review after we exclude stylistic changes that are not necessary. Also, it'd be good to enhance the description besides the example and call out the corner cases for which logic has been added.

Sorry that I did not follow the code style in connect module. I have reverted all incorrect style. Also, I have updated the description.

kkonstantine (commented):
Thanks for the quick turnaround @chia7712 !
Much easier to review line my line now (at least for me). 

I'll do my best to review by the end of the week.

PR #10068, opened 2021-02-05
MINOR: StopReplicaResp and StopReplicaReq Test should cover all available version

dengziming:
*More detailed description of your change*
Accidently found that the test case in `StopReplicaRespTest` and `StopReplicaReqTest` didn't cover all versions.

*Summary of testing strategy (including rationale)*
QA

### Committer Checklist (excluded from commit message)
- [ ] Verify design and implementation 
- [ ] Verify test coverage and CI build status
- [ ] Verify documentation (including upgrade notes)

PR #10078, opened 2021-02-07
MINOR: make sure all generated data tests cover all versions

chia7712:
inspired by #10068


clients/src/main/java/org/apache/kafka/common/protocol/ApiKeys.java
```diff
@@ -201,7 +201,7 @@ public short oldestVersion() {
 
     public List<Short> allVersions() {
         List<Short> versions = new ArrayList<>(latestVersion() - oldestVersion() + 1);
-        for (short version = oldestVersion(); version < latestVersion(); version++) {
+        for (short version = oldestVersion(); version <= latestVersion(); version++) {
```
chia7712, line 204: This is not a critical bug as the method `allVersions` is used by testing only.

clients/src/test/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochRequestTest.java
```diff
@@ -34,15 +34,15 @@ public void testForConsumerRequiresVersion3() {
             assertThrows(UnsupportedVersionException.class, () -> builder.build(v));
         }
 
-        for (short version = 3; version < ApiKeys.OFFSET_FOR_LEADER_EPOCH.latestVersion(); version++) {
+        for (short version = 3; version <= ApiKeys.OFFSET_FOR_LEADER_EPOCH.latestVersion(); version++) {
             OffsetsForLeaderEpochRequest request = builder.build((short) 3);
```
dajac: Should we replace `3` by `version` here?

clients/src/test/java/org/apache/kafka/common/requests/ProduceRequestTest.java
```diff
@@ -291,19 +292,11 @@ public void testMixedIdempotentData() {
...
-    }
-
-    private void assertThrowsInvalidRecordException(ProduceRequest.Builder builder, short version) {
-        try {
-            builder.build(version).serialize();
-            fail("Builder did not raise " + InvalidRecordException.class.getName() + " as expected");
-        } catch (RuntimeException e) {
-            assertTrue(InvalidRecordException.class.isAssignableFrom(e.getClass()),
-                "Unexpected exception type " + e.getClass().getName());
+    private static <T extends Throwable> void assertThrowsInvalidRecordExceptionForAllVersions(ProduceRequest.Builder builder,
```
dajac: Should we rename the method now that the exception is passed as an argument?

clients/src/test/java/org/apache/kafka/common/requests/ProduceRequestTest.java
```diff
@@ -291,19 +292,11 @@ public void testMixedIdempotentData() {
...
+                                                                                               Class<T> expectedType) {
+        for (short version = builder.oldestAllowedVersion(); version <= builder.latestAllowedVersion(); version++) {
+            short v = version;
```
dajac: Could we get rid of `v` and use `version` below directly?

clients/src/test/java/org/apache/kafka/common/requests/RequestResponseTest.java
```diff
@@ -295,7 +295,7 @@ public void testSerialization() throws Exception {
         checkRequest(createDeleteGroupsRequest(), true);
         checkErrorResponse(createDeleteGroupsRequest(), unknownServerException, true);
         checkResponse(createDeleteGroupsResponse(), 0, true);
-        for (int i = 0; i < ApiKeys.LIST_OFFSETS.latestVersion(); i++) {
+        for (int i = 0; i <= ApiKeys.LIST_OFFSETS.latestVersion(); i++) {
```
dajac: Could we use `allVersions` here as well?

clients/src/test/java/org/apache/kafka/common/message/MessageTest.java
```diff
@@ -485,7 +485,7 @@ public void testOffsetCommitRequestVersions() throws Exception {
             if (version == 1) {
                 testEquivalentMessageRoundTrip(version, requestData);
             } else if (version >= 2 && version <= 4) {
-                testAllMessageRoundTripsBetweenVersions(version, (short) 4, requestData, requestData);
+                testAllMessageRoundTripsBetweenVersions(version, (short) 5, requestData, requestData);
```
dajac, line 488: Why are we changing this?

dajac (commented):
Thanks for the PR. I have left few comments.

clients/src/test/java/org/apache/kafka/common/requests/OffsetsForLeaderEpochRequestTest.java
chia7712: good catch. fixed.

clients/src/test/java/org/apache/kafka/common/requests/ProduceRequestTest.java
chia7712: It requires final variable. I rewrite code by lambda foreach to eliminate duplicate intermediate variable.

clients/src/test/java/org/apache/kafka/common/message/MessageTest.java
chia7712: `testAllMessageRoundTripsBetweenVersions` exclude the end number so it should pass 5 so as to test version 4

dajac (approved):
LGTM, thanks for the PR.

PR #10080, opened 2021-02-08
MINOR: KafkaBroker.brokerState should be volatile instead of AtomicReference

ijuma:
We don't need or use the additional functionality provided by
AtomicReference.

### DIFF
--- a/core/src/main/scala/kafka/server/KafkaBroker.scala
+++ b/core/src/main/scala/kafka/server/KafkaBroker.scala
@@ -18,7 +18,6 @@
 package kafka.server
 
 import java.util
-import java.util.concurrent.atomic.AtomicReference
 
 import com.yammer.metrics.core.MetricName
 import kafka.log.LogManager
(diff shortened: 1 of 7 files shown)


ijuma:
cc @rondagostino

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -71,8 +70,11 @@ object KafkaBroker {
 }
 
 trait KafkaBroker extends KafkaMetricsGroup {
+  @volatile private var _brokerState: BrokerState = BrokerState.NOT_RUNNING
```
chia7712, line 73: Why not using ```@volatile var brokerState: BrokerState = BrokerState.NOT_RUNNING```?

ijuma: Because that would make it mutable for everyone. The current implementation is mutable only within subclasses.

mumrah (approved):
LGTM

core/src/main/scala/kafka/server/KafkaBroker.scala
```diff
@@ -90,7 +92,7 @@ trait KafkaBroker extends KafkaMetricsGroup {
     explicitMetricName(KafkaBroker.metricsPrefix, KafkaBroker.metricsTypeName, name, metricTags)
   }
 
-  newGauge("BrokerState", () => brokerState.get.value())
+  newGauge("BrokerState", () => brokerState.value)
```
chia7712, line 95: (this is not related to this PR) it is a bit weird to me that we expose "byte" as "broker state".

chia7712 (approved):

core/src/main/scala/kafka/server/KafkaBroker.scala
ijuma: Right. To change this, it would require a KIP.

PR #10082, opened 2021-02-08
MINOR: use 'mapKey' to avoid unnecessary grouped data

chia7712:
as title. Thanks to great KRPC :)

# changes in this PR

1. add 'mapKey=true' to `DescribeLogDirsRequest`
1. rename `PartitionIndex` to `Partitions` for `DescribeLogDirsRequest`
1. add 'mapKey=true' to `ElectLeadersRequest`
1. rename `PartitionId` to `Partitions` for `ElectLeadersRequest`
1. add 'mapKey=true' to `ConsumerProtocolAssignment`


clients/src/main/resources/common/message/ElectLeadersRequest.json
```diff
@@ -28,7 +28,7 @@
     { "name": "TopicPartitions", "type": "[]TopicPartitions", "versions": "0+", "nullableVersions": "0+",
       "about": "The topic partitions to elect leaders.",
       "fields": [
-        { "name": "Topic", "type": "string", "versions": "0+", "entityType": "topicName",
+        { "name": "Topic", "type": "string", "versions": "0+", "entityType": "topicName", "mapKey": true,
           "about": "The name of a topic." },
         { "name": "PartitionId", "type": "[]int32", "versions": "0+",
```
dajac: nit: While we are here, should we rename this field to `Partitions`? I was confused by this while reading the code. When I read `PartitionId`, I don't really expect an array.

chia7712: make sense

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -2581,10 +2581,10 @@ public DescribeReplicaLogDirsResult describeReplicaLogDirs(Collection<TopicParti
                 List<Integer> partitionIndex = new ArrayList<>();
```
ijuma: Change this variable name too?

ijuma:
Looks like ConsumerProtocolAssignment was also changed to have mapKey, so worth updating the PR to include that information.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
dajac: I do agree.

dajac (approved):
@chia7712 LGTM besides @ijuma's comment. It would be good to rename that variable to be consistent.

chia7712:
> Looks like ConsumerProtocolAssignment was also changed to have mapKey, so worth updating the PR to include that information.

done

chia7712:
the error is traced by #10024

PR #10084, opened 2021-02-08
MINOR: Rename DecommissionBrokers to UnregisterBrokers

cmccabe:
Rename DecommissionBrokers to UnregisterBrokers.  Fix an incorrect JavaDoc comment for the Admin API.  Make sure that UNREGISTER_BROKER is marked as forwardable and not as a controller-only API (since it can received by brokers).


core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -217,6 +217,7 @@ class KafkaApis(val requestChannel: RequestChannel,
         case ApiKeys.ENVELOPE => handleEnvelope(request)
         case ApiKeys.DESCRIBE_CLUSTER => handleDescribeCluster(request)
         case ApiKeys.DESCRIBE_PRODUCERS => handleDescribeProducersRequest(request)
+        case ApiKeys.UNREGISTER_BROKER => maybeForwardToController(request, handleUnregisterBrokerRequest)
```
chia7712, line 220: The other APIs from KIP-500 is just closed. Maybe this API could follow same pattern?

cmccabe: This call needs to be forwarded to the KIP-500 controller.  That is different from the other KIP-500 RPCs which are controller RPCs, and are not expected to be received on the broker at all.

chia7712: @cmccabe Thanks for explanation. It seems the [comment](https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/server/KafkaApis.scala#L221) gets invalid now.
```
Handle requests that should have been sent to the KIP-500 controlle
```

Could you revise the comment as well?

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
```diff
@@ -1486,41 +1486,40 @@ default DescribeFeaturesResult describeFeatures() {
...
      * <p>
      *
      * @param brokerId  the broker id to unregister.
      * @param options   the options to use.
      *
-     * @return the {@link DecommissionBrokerResult} containing the result
+     * @return the {@link UnregisterBrokerResult} containing the result
      */
-    DecommissionBrokerResult decommissionBroker(int brokerId, DecommissionBrokerOptions options);
+    UnregisterBrokerResult unregisterBroker(int brokerId, UnregisterBrokerOptions options);
```
ijuma, line 1528: @cmccabe We should mark all KIP-500 methods with `@Unstable` and mention it in the documentation too.

core/src/main/scala/kafka/server/KafkaApis.scala
cmccabe: The comment is not invalid.  It just doesn't apply to UNREGISTER_BROKER.  It does apply to VOTE, BEGIN_QUORUM_EPOCH, etc. etc.  I moved UNREGISTER_BROKER so it no longer appears underneath the comment.

chia7712 (approved):

clients/src/main/java/org/apache/kafka/clients/admin/Admin.java
```diff
@@ -1486,41 +1486,46 @@ default DescribeFeaturesResult describeFeatures() {
...
     /**
-     * Permanently remove a broker and reassign any partitions on the broker.
+     * Unregister a broker.
      * <p>
-     * This operation is supported only on self-managed Kafka clusters (i.e. brokers which do not rely on Zookeeper).
+     * This operation does not have any effect on partition assignments. It is supported
+     * only on Kafka clusters which use Raft to store metadata, rather than ZooKeeper.
      *
      * The following exceptions can be anticipated when calling {@code get()} on the future from the
      * returned {@link DescribeFeaturesResult}:
```
ijuma: "DescribeFeaturesResult" needs to be updated.

core/src/main/scala/kafka/server/KafkaApis.scala
```diff
@@ -3390,6 +3390,11 @@ class KafkaApis(val requestChannel: RequestChannel,
       new DescribeProducersResponse(response.setThrottleTimeMs(requestThrottleMs)))
   }
 
+  def handleUnregisterBrokerRequest(request: RequestChannel.Request): Unit = {
+    throw new UnsupportedVersionException("The broker unregistration API is not available when using " +
+      "Apache ZooKeeper mode.")
```
ijuma, line 3398: This is a bit confusing since this method can be called when the built-in quorum mode is used too. Or am I missing something?

ijuma (commented):
Thanks for the PR. A couple of comments below. Also, we should update the PR description to describe the other changes we include in the PR besides the rename.

core/src/main/scala/kafka/server/KafkaApis.scala
cmccabe: This method can't be called in built-in quorum mode since the API is forwardable and we will always have a forwarding manager

ijuma: Can we add a comment explaining that?

ijuma (approved):
LGTM after the comment is added and the PR description (and commit message) is updated.

core/src/main/scala/kafka/server/KafkaApis.scala
cmccabe: will do

PR #10085, opened 2021-02-09
KAFKA-12154: Snapshot Loading API

jsancio:
Implement Raft Snapshot loading API.

1. Adds a new method `handleSnapshot` to `raft.Listener` which is called whenever the `RaftClient` determines that the `Listener` needs to load a new snapshot before reading the log. This happens when the `Listener`'s next offset is less than the log start offset also known as the earliest snapshot.

2.  Adds a new type `SnapshotReader<T>` which provides a `Iterator<Batch<T>>` interface and de-serializes records in the `RawSnapshotReader` into `T`s

3.  Adds a new type `RecordsIterator<T>` that implements an `Iterator<Batch<T>>` by scanning a `Records` object and deserializes the batches and records into `Batch<T>`. This type is used by both `SnapshotReader<T>` and `RecordsBatchReader<T>` internally to implement the `Iterator` interface that they expose. 

4. Changes the `MockLog` implementation to read one batch at a time. The previous implementation always read from the given offset to the high-watermark. This made it impossible to test interesting snapshot loading scenarios.

5. Removed `throws IOException` from some methods. Some of types were inconsistently throwing `IOException` in some cases and throwing `RuntimeException(..., new IOException(...))` in others. This PR improves the consistent by wrapping `IOException` in `RuntimeException` in a few more places and replacing `Closeable` with `AutoCloseable`.

6. Updated the Kafka Raft simulation test to take into account snapshot. `ReplicatedCounter` was updated to generate snapshot after 10 records get committed. This means that the `ConsistentCommittedData` validation was extended to take snapshots into account.

comment to ping reviewers.*


raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -321,6 +335,23 @@ private void updateListenersProgress(List<ListenerContext> listenerContexts, lon
         }
     }
 
+    private Optional<SnapshotReader<T>> oldestSnapshot() {
+        if (log.oldestSnapshotId().isPresent()) {
```
dengziming: nit: we can use `log.oldestSnapshotId().map(oldestSnapshotId -> {...})` to get rid of if `else`

jsancio: Thanks. Fixed!

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -311,8 +311,18 @@ private void updateListenersProgress(long highWatermark) {
     private void updateListenersProgress(List<ListenerContext> listenerContexts, long highWatermark) {
         for (ListenerContext listenerContext : listenerContexts) {
             listenerContext.nextExpectedOffset().ifPresent(nextExpectedOffset -> {
-                if (nextExpectedOffset < log.startOffset()) {
-                    listenerContext.fireHandleSnapshot(log.startOffset());
+                if (nextExpectedOffset < log.startOffset() && nextExpectedOffset < highWatermark) {
+                    SnapshotReader<T> snapshot = earliestSnapshot().orElseThrow(() -> {
```
hachikuji: Why would we use the earliest snapshot and not the latest?

jsancio: No good reason. I'll change it to use the latest.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -2154,8 +2182,14 @@ private boolean maybeCompleteShutdown(long currentTimeMs) {
         return false;
     }
 
-    private void maybeUpdateOldestSnapshotId() {
-        log.latestSnapshotId().ifPresent(log::deleteBeforeSnapshot);
+    private void maybeUpdateEarliestSnapshotId() {
```
hachikuji: nit: it seems like the objective here is deleting old snapshots, which is why we call `deleteBeforeSnaphshot`. Updating the earliest snapshot id seems more like a side effect.

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
```diff
@@ -87,6 +96,25 @@ public synchronized void handleCommit(BatchReader<Integer> reader) {
...
+                for (List<Integer> batch : snapshot) {
+                    for (Integer value : batch) {
+                        log.debug("Setting value: {}", value);
+                        this.committed = value;
+                        this.uncommitted = value;
+                    }
+                }
+                log.debug("Finished loading snapshot. Set value: {}", this.committed);
+            }
+        } catch (IOException e) {
```
hachikuji: Why do we need the second level here?

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
```diff
@@ -87,6 +96,25 @@ public synchronized void handleCommit(BatchReader<Integer> reader) {
         }
     }
 
+    @Override
+    public synchronized void handleSnapshot(SnapshotReader<Integer> reader) {
+        try {
+            try (SnapshotReader<Integer> snapshot = reader) {
+                log.debug("Loading snapshot {}", snapshot.snapshotId());
+                for (List<Integer> batch : snapshot) {
```
hachikuji: I guess snapshots for this replicated counter would end up with a single value?

raft/src/main/java/org/apache/kafka/raft/metadata/MetaLogRaftShim.java
```diff
@@ -124,6 +126,15 @@ public void handleCommit(BatchReader<ApiMessageAndVersion> reader) {
             }
         }
 
+        @Override
+        public void handleSnapshot(SnapshotReader<ApiMessageAndVersion> reader) {
+            // TODO: Create Jira: Handle loading commit in ListenerShim
```
hachikuji: We can probably take care of this when we merge the two interfaces. I'm hoping to get to that shortly.

raft/src/main/java/org/apache/kafka/snapshot/SnapshotReader.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.List;
+import java.util.NoSuchElementException;
+import org.apache.kafka.common.protocol.ByteBufferAccessor;
+import org.apache.kafka.common.record.Record;
+import org.apache.kafka.common.record.RecordBatch;
+import org.apache.kafka.common.utils.BufferSupplier;
+import org.apache.kafka.raft.OffsetAndEpoch;
+import org.apache.kafka.raft.RecordSerde;
+
+public final class SnapshotReader<T> implements Closeable, Iterable<List<T>> {
```
hachikuji: I guess I'll renew my appeal to converge this interface with `BatchReader`. We really ought to be able to share some code.

jsancio: Fair enough. Let me revisit this and see what code we can reuse.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -1300,18 +1330,18 @@ private FetchSnapshotResponseData handleFetchSnapshotRequest(
...
             if (partitionSnapshot.position() > Integer.MAX_VALUE) {
+                // TODO: This should return an error response instead of throwing an exception
                 throw new IllegalStateException(String.format("Trying to fetch a snapshot with position: %d lager than Int.MaxValue", partitionSnapshot.position()));
             }
 
-            UnalignedRecords records = snapshot.read(partitionSnapshot.position(), Math.min(data.maxBytes(), maxSnapshotSize));
-
-            long snapshotSize = snapshot.sizeInBytes();
+            // TODO: I think this slice of records is closed when the snapshot is close in the try (...) above.
+            UnalignedRecords records = snapshot.slice(partitionSnapshot.position(), Math.min(data.maxBytes(), maxSnapshotSize));
```
jsancio, line 1316: When the implementation is a `FileRawSnapshotReader`, the created slice will be `close` before the network client has had a chance to send the bytes. Created https://issues.apache.org/jira/browse/KAFKA-12543 and I will work on this after this PR.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: My thought process was that what the Raft client cares about is the log start offset and the snapshot at that offset. The fact that snapshots get deleted and how they get deleted is an implementation detail in `ReplicatedLog`.

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
jsancio: This is because `SnapshotReader::close` was declare as throwing an `IOException`.

This made the API for `SnapshotReader` confusing as `hasNext` and `next` don't throw an `IOException` even though they read from disk in some cases. I fixed this by changing `SnapshotReader` to implement `AutoCloseable` instead of `Closeable`.

raft/src/main/java/org/apache/kafka/raft/metadata/MetaLogRaftShim.java
jsancio: Okay. I'll remove the TODO. Do you have a Jira for this? If not let me know and I can create one.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
```diff
@@ -355,33 +354,30 @@ public LogFetchInfo read(long startOffset, Isolation isolation) {
         }
 
         ByteBuffer buffer = ByteBuffer.allocate(512);
-        LogEntry firstEntry = null;
+        LogOffsetMetadata batchStartOffset = null;
```
jsancio, line 366: Note that the changes to this method are so that `read`  doesn't return all of the batches from from `startOffset` to `highWatermark`. This was needed for more interested test cases around snapshot loading.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
```diff
@@ -418,48 +414,49 @@ public RawSnapshotWriter createSnapshot(OffsetAndEpoch snapshotId) {
     public void onSnapshotFrozen(OffsetAndEpoch snapshotId) {}
 
     @Override
-    public boolean deleteBeforeSnapshot(OffsetAndEpoch logStartSnapshotId) {
-        if (logStartOffset() > logStartSnapshotId.offset ||
-            highWatermark.offset < logStartSnapshotId.offset) {
+    public boolean deleteBeforeSnapshot(OffsetAndEpoch snapshotId) {
```
jsancio, line 434: Note that the changes to this method are to relax the log start offset and high-watermark invariant so that we can create more interesting snapshot and log states in the `RaftClientTestContext.Builder`.

shell/src/main/java/org/apache/kafka/shell/MetadataNodeManager.java
```diff
@@ -105,6 +106,22 @@ public void handleCommits(long lastOffset, List<ApiMessage> messages) {
             }, null);
         }
 
+        @Override
+        public void handleSnapshot(SnapshotReader<ApiMessageAndVersion> reader) {
+            // TODO: Create Jira: Need to cover the case where handle snapshot invalidates previous commits
+            //                    Need to handle that reader.snapshotId() means that every record up to that offset is committed
```
jsancio: Created this Jira: https://issues.apache.org/jira/browse/KAFKA-12545

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
jsancio: Yes. I missed this. `ReplicatedCounter` now generates a snapshot after more than 10 records. Fix the existing simulation tests to take snapshot into account.

raft/src/main/java/org/apache/kafka/snapshot/SnapshotReader.java
jsancio: Implemented a `SerdeRecordsIterator` that both `BatchReader` and `SnapshotReader` use internally. The biggest change here was to take into account large `FileRecords`.

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
jsancio: I also changed a few places that were throwing `IOException`

jsancio:
Update the PR's description to match the latest changes included here.

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
```diff
@@ -60,19 +61,21 @@
     @Override
     void close();
 
-    class Batch<T> {
+    final class Batch<T> implements Iterable<T> {
```
jsancio: @hachikuji given that we use this type in `SnapshotReader<T>` I am okay moving this type to something like `o.a.k.r.Batch` instead of `o.a.k.r.BatchReader.Batch`.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -326,6 +336,14 @@ private void updateListenersProgress(List<ListenerContext> listenerContexts, lon
         }
     }
 
+    private Optional<SnapshotReader<T>> latestSnapshot() {
+        return log.latestSnapshotId().flatMap(snapshoId -> {
+            return log
```
mumrah: nit: can omit the curly braces and return here

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -326,6 +336,14 @@ private void updateListenersProgress(List<ListenerContext> listenerContexts, lon
         }
     }
 
+    private Optional<SnapshotReader<T>> latestSnapshot() {
+        return log.latestSnapshotId().flatMap(snapshoId -> {
```
mumrah: typo: snapshoId

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
```diff
@@ -68,20 +70,65 @@ public synchronized void increment() {
...
+            int initialValue = committed;
             while (reader.hasNext()) {
                 BatchReader.Batch<Integer> batch = reader.next();
                 log.debug("Handle commit of batch with records {} at base offset {}",
                     batch.records(), batch.baseOffset());
                 for (Integer value : batch.records()) {
-                    if (value != this.committed + 1) {
-                        throw new AssertionError("Expected next committed value to be " +
-                            (this.committed + 1) + ", but instead found " + value + " on node " + nodeId);
+                    if (value != committed + 1) {
```
mumrah: nit: not really part of this PR, but can we use variable names besides "value" and "initialValue"?

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
```diff
@@ -68,20 +70,65 @@ public synchronized void increment() {
...
-                    this.committed = value;
+                    committed = value;
                 }
+
+                nextReadOffset = batch.lastOffset() + 1;
+                readEpoch = batch.epoch();
             }
             log.debug("Counter incremented from {} to {}", initialValue, committed);
+
+            if (lastSnapshotEndOffset + 10 < nextReadOffset) {
```
mumrah: Where does `10` come from? Should it be configurable?

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
```diff
@@ -0,0 +1,231 @@
...
+        allocatedBuffer = Optional.empty();
+    }
+
+    private void checkIfClosed() {
+        if (isClosed) {
+            throw new IllegalStateException("Serde record batch itererator was closed");
+        }
+    }
+
+    private Optional<Iterator<MutableRecordBatch>> nextBatches() {
```
mumrah: Is the Optional necessary here? Could we instead return an empty Iterator?

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
```diff
@@ -0,0 +1,231 @@
...
+        int recordSize = records.sizeInBytes();
+        if (bytesRead < recordSize) {
+            final MemoryRecords memoryRecords;
+            if (records instanceof MemoryRecords) {
+                bytesRead = recordSize;
+                memoryRecords = (MemoryRecords) records;
+            } else if (records instanceof FileRecords) {
```
mumrah: I'm wondering if it makes sense to have two implementation for the two types of records we support here. It might make the code easier to follow -- especially if we start have lots of `instanceof` checks down the line (as the complexity of this code grows).

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
```diff
@@ -0,0 +1,231 @@
...
+
+        return Optional.empty();
+    }
+
+    private Batch<T> readBatch(DefaultRecordBatch batch) {
+        final Batch<T> result;
+        if (batch.isControlBatch()) {
+            result = Batch.empty(batch.baseOffset(), batch.partitionLeaderEpoch(), batch.lastOffset());
+        } else {
+            Integer numRecords = batch.countOrNull();
```
mumrah: Maybe use OptionalInt here? We're pretty consistently using Optional everywhere else in this class

mumrah (commented):
Thanks for the patch! First pass of questions/comments inline

raft/src/main/java/org/apache/kafka/raft/metadata/MetaLogRaftShim.java
```diff
@@ -124,6 +125,11 @@ public void handleCommit(BatchReader<ApiMessageAndVersion> reader) {
             }
         }
 
+        @Override
+        public void handleSnapshot(SnapshotReader<ApiMessageAndVersion> reader) {
+            reader.close();
```
dengziming, line 131: Should we do something here?

raft/src/main/java/org/apache/kafka/snapshot/RawSnapshotReader.java
```diff
@@ -37,17 +36,21 @@
      *
      * @throws IOException for any IO error while reading the size
```
dengziming: nit: remove @throws

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: nit: missing t in snapshotId

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -355,24 +373,29 @@ private void fireHandleResign(int epoch) {
...
-        // When there is only a single voter, become candidate immediately
-        if (quorum.isVoter()
-            && quorum.remoteVoters().isEmpty()
-            && !quorum.isLeader()
-            && !quorum.isCandidate()) {
-            transitionToCandidate(currentTimeMs);
+            // When there is only a single voter, become candidate immediately
+            if (quorum.isVoter()
+                && quorum.remoteVoters().isEmpty()
+                && !quorum.isLeader()
```
hachikuji: Since we're in here already, this check is not needed since we already ruled it out above.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
hachikuji: Hmm, I guess I still see this the other way around. Why would the raft client care about updating the log start offset if not to delete old snapshots? What would that even mean outside the context of snapshot deletion?

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
```diff
@@ -68,20 +70,65 @@ public synchronized void increment() {
...
+                                value,
+                                nodeId
+                            )
+                        );
                     }
-                    this.committed = value;
+                    committed = value;
                 }
+
+                nextReadOffset = batch.lastOffset() + 1;
```
hachikuji, line 96: As far as I can tell, we should be able to convert these two fields to local variables.

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
hachikuji: Looks like we are trying to do snapshots every 10 records. We could probably get rid of `lastSnapshotEndOffset` and use `committed % 10` or something like that. It may also be useful to be able to control the frequency of snapshots with a parameter.

raft/src/main/java/org/apache/kafka/raft/internals/RecordsBatchReader.java
```diff
@@ -163,48 +85,46 @@ public OptionalLong lastOffset() {
...
+            baseOffset,
+            new SerdeRecordsIterator<>(records, serde, bufferSupplier, maxBatchSize),
+            closeListener
+        );
+    }
 
-        int valueSize = input.readVarint();
-        if (valueSize < 0) {
-            throw new IllegalArgumentException();
+    private void checkIfClosed() {
```
hachikuji: nit: maybe `ensureNotClosed` or `ensureOpen`?

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
```diff
@@ -0,0 +1,231 @@
...
+import org.apache.kafka.common.record.DefaultRecordBatch;
+import org.apache.kafka.common.record.FileRecords;
+import org.apache.kafka.common.record.MemoryRecords;
+import org.apache.kafka.common.record.MutableRecordBatch;
+import org.apache.kafka.common.record.Records;
+import org.apache.kafka.common.utils.BufferSupplier;
+import org.apache.kafka.raft.BatchReader.Batch;
+import org.apache.kafka.raft.RecordSerde;
+
+public final class SerdeRecordsIterator<T> implements Iterator<Batch<T>>, AutoCloseable {
```
hachikuji: nit: it's really more of a _batch_ iterator. Could we cal this `RecordsBatchIterator` to go along with `RecordBatchReader`?

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
```diff
@@ -0,0 +1,231 @@
...
+                }
+
+                int start = buffer.position();
+                try { 
+                    ((FileRecords) records).readInto(buffer, bytesRead);
+                } catch (IOException e) {
+                    throw new RuntimeException("Failed to read records into memory", e);
+                }
+
+                bytesRead += buffer.limit() - start;
```
hachikuji: Maybe I am missing it, but how does this ensure proper alignment? The batch at the end of the buffer could be incomplete, no?

raft/src/test/java/org/apache/kafka/raft/MockLog.java
```diff
@@ -418,48 +414,49 @@ public RawSnapshotWriter createSnapshot(OffsetAndEpoch snapshotId) {
...
+        if (logStartOffset() > snapshotId.offset ||
+            highWatermark.offset < snapshotId.offset) {
 
             throw new OffsetOutOfRangeException(
                 String.format(
                     "New log start (%s) is less than start offset (%s) or is greater than the high watermark (%s)",
```
hachikuji: nit: maybe we can just have two separate cases so that we can give a clearer message

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
```diff
@@ -988,14 +991,43 @@ private int parseSequenceNumber(ByteBuffer value) {
...
+
             OptionalLong highWatermark = manager.highWatermark();
             if (!highWatermark.isPresent()) {
                 // We cannot do validation if the current high watermark is unknown
                 return;
             }
 
-            for (LogBatch batch : log.readBatches(0L, highWatermark)) {
+            AtomicLong startOffset = new AtomicLong(0);
+            log.earliestSnapshotId().ifPresent(snapshoId -> {
```
hachikuji: nit: typo `snapshoId`

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
mumrah: @hachikuji are you talking about the buffer passed to MemoryRecords and bytes that could be left over after iterating through the batches?

jsancio: There are a couple of important points here:

1. To `MemoryRecords` we pass a slice.
2. We update the `allocatedBuffer::position` when a batch is successfully read here https://github.com/apache/kafka/pull/10085/files/2dd49397a1848e0b9b9c11aa72d5455c5fccfc3e#diff-9ecaf1dc29d739f7a29639b070051e3750f49ea0ac73dc5c27680dfb8f7b9ef6R156-R157
3. The buffer is flipped by https://github.com/apache/kafka/pull/10085/files/2dd49397a1848e0b9b9c11aa72d5455c5fccfc3e#diff-9ecaf1dc29d739f7a29639b070051e3750f49ea0ac73dc5c27680dfb8f7b9ef6R130

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
```diff
@@ -0,0 +1,231 @@
...
+                        // The buffer is not big enough to read an entire batch
+                        throw new IllegalStateException(
+                            String.format(
+                                "Unable to read batch from file records buffer %s with maximum batch %s and record size %s",
+                                buffer,
+                                maxBatchSize,
+                                records.sizeInBytes()
+                            )
+                        );
+                    }
```
jsancio: @hachikuji @mumrah I am not super happy with this handling. Specially since it is possible for us to write a batch greater than `maxBatchSize` based on how `BatchAccumulator` works. I am thinking for now that we configure the write size to 8MB and the read size to something like 10MB.

When we have a better solution for `scheduleAtomicAppend` that doesn't rely on batches for atomicity, I think this is less important.

hachikuji: Can we just let the buffer grow as needed? We have a `BufferSupplier` after all.

jsancio: To get this new size we need to read the size from the batch record using the `Readable` for example. I don't know if we want to trust this value for memory allocation.

hachikuji: We have a history of size-related allocation failures when reading from the log, so it would be nice to avoid them. I agree we should probably not allocate blindly, but I also do not think we should put too much faith in the max batch size. Eventually the max batch size may be configurable, but we'll always have to deal with historic data.

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
hachikuji: Ok, I think I get it. I missed the call to `compact` above.

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
hachikuji: Probably a good move. I think this dates back before we had `Optional` available. I'd suggest doing it in a separate PR though since it's going to touch a bunch of otherwise unrelated code.

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
jsancio: Yes. Done.

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
jsancio: I think most of the code will be the same but I can try to have to different implementations. Do you mind if I do that in a future PR?

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
jsancio: Yes. I can do this in a followup PR. Do we want to add a new method or do we want to replace the existing method? Is `interface RecordBatch` considered a public API?

raft/src/main/java/org/apache/kafka/raft/metadata/MetaLogRaftShim.java
jsancio: In the future yes. Simply ignore the snapshot is okay for now for the following reasons:
1. We are getting rid of this shim in https://github.com/apache/kafka/pull/10497
2. None of the kraft listeners generate snapshots so this should never be called
3. We will implement snapshot loading for the controller and broker as part of this jira: https://issues.apache.org/jira/browse/KAFKA-12466

raft/src/main/java/org/apache/kafka/snapshot/RawSnapshotReader.java
jsancio: Done.

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
jsancio: Yep. Not sure what I was thinking when I added those fields.

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
jsancio: Hmm. We should only generate snapshots at batch boundaries. There is no guarantee that the `lastOffset` of a batch is a multiple of some number since batches can any number of records.

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
jsancio: By the way Kafka already has `org.apache.kafka.common.record.RecordBatchIterator<T extends RecordBatch>`.

To me, the name of the iterator should represent what it iterates over not what it generates. In this case it is iterating over `Records` (`MemoryRecords` and `FileRecords`). When we implement @mumrah's suggestion of having separate implements for each, we will have `MemoryRecordsIterator` and `FileRecordsIterator`.

I'll change the name to `RecordsIterator`.

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
jsancio: Yes. The fact that we `compact` in subsequent calls is important since it copies unused byte to the start of the byte buffer.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
jsancio: Done.

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
jsancio: Yeah, got it. I thought about it some more. This is currently only used for the metadata partitions. If the controller or broker cannot read the metadata partition then it cannot make any progress in the rest of the system. In this case, the process should probably shutdown. Causing the process to shutdown because of an `OOMException` is probably okay.

jsancio:
@mumrah @hachikuji @dengziming thanks for the feedback. I should have addressed all of your suggestions. Let me know if I missed anything.

raft/src/main/java/org/apache/kafka/raft/internals/RecordsBatchReader.java
jsancio: Done.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
jsancio: Made it a variable.

raft/src/main/java/org/apache/kafka/raft/ReplicatedCounter.java
jsancio: Done. Not sure if it is any better :smile:

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: I changed the method name.

raft/src/main/java/org/apache/kafka/raft/internals/SerdeRecordsIterator.java
mumrah: Separate PR sounds fine. I'm pretty sure `RecordBatch` is a public interface though. I'm not sure how we normally go about changing signatures in public APIs. Definitely out of scope for this PR

core/src/main/scala/kafka/tools/TestRaftServer.scala
```diff
@@ -226,7 +232,11 @@ class TestRaftServer(
             reader.close()
           }
 
-        case _ =>
+        case HandleSnapshot(reader) =>
+          // Ignore snapshots; only interested on records appended by this leader
```
hachikuji: nit: on -> in?

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
hachikuji: Pulling it up sounds reasonable to me.

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
```diff
@@ -60,19 +61,21 @@
...
         private final long baseOffset;
         private final int epoch;
+        private final long lastOffset;
         private final List<T> records;
 
-        public Batch(long baseOffset, int epoch, List<T> records) {
+        private Batch(long baseOffset, int epoch, long lastOffset, List<T> records) {
```
hachikuji: Perhaps it is clear enough already, but maybe we should document that these offsets are inclusive.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
```diff
@@ -311,8 +311,18 @@ private void updateListenersProgress(long highWatermark) {
     private void updateListenersProgress(List<ListenerContext> listenerContexts, long highWatermark) {
         for (ListenerContext listenerContext : listenerContexts) {
             listenerContext.nextExpectedOffset().ifPresent(nextExpectedOffset -> {
-                if (nextExpectedOffset < log.startOffset()) {
-                    listenerContext.fireHandleSnapshot(log.startOffset());
+                if (nextExpectedOffset < log.startOffset() && nextExpectedOffset < highWatermark) {
+                    SnapshotReader<T> snapshot = latestSnapshot().orElseThrow(() -> {
+                        return new IllegalStateException(
+                            String.format(
+                                "Snapshot expected when next offset is %s, log start offset is %s and high-watermark is %s",
```
hachikuji: Perhaps it is useful to mention the class of the listener since we are referring to its next expected offset?

raft/src/main/java/org/apache/kafka/raft/internals/RecordsIterator.java
```diff
@@ -0,0 +1,238 @@
...
+            buffer = allocatedBuffer.get();
+            buffer.compact();
+        } else {
+            buffer = bufferSupplier.get(Math.min(batchSize, records.sizeInBytes()));
+            allocatedBuffer = Optional.of(buffer);
+        }
+
+        MemoryRecords memoryRecords = readFileRecords(fileRecords, buffer);
+
+        if (memoryRecords.firstBatchSize() < buffer.remaining()) {
```
hachikuji: Since `firstBatchSize()` can return null, it might be worth adding a comment explaining that we guarantee a minimum size of `HEADER_SIZE_UP_TO_MAGIC`.

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
```diff
@@ -1081,24 +1103,25 @@ OptionalInt currentClaimedEpoch() {
...
-        @Override
-        public void handleClaim(int epoch) {
-            // We record the next expected offset as the claimed epoch's start
-            // offset. This is useful to verify that the `handleClaim` callback
-            // was not received early.
-            long claimedEpochStartOffset = lastCommitOffset().isPresent() ?
-                lastCommitOffset().getAsLong() + 1 : 0L;
-            this.currentClaimedEpoch = OptionalInt.of(epoch);
-            this.claimedEpochStartOffsets.put(epoch, claimedEpochStartOffset);
+        Optional<SnapshotReader<String>> takeSnapshot() {
```
hachikuji: This name confused me a little bit. I thought it was generating a snapshot. Maybe something like `drainHandledSnapshot` or something like that would be clearer?

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
```diff
@@ -717,7 +720,8 @@ void start(int nodeId) {
                 persistentState.store,
                 logContext,
                 time,
-                random
+                random,
+                serde
```
hachikuji, line 725: Maybe a useful `Invariant` we can add here is that there always exists a snapshot corresponding to the log start offset.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
```diff
@@ -363,33 +362,30 @@ public LogFetchInfo read(long startOffset, Isolation isolation) {
...
         for (LogBatch batch : batches) {
             // Note that start offset is inclusive while max offset is exclusive. We only return
             // complete batches, so batches which end at an offset larger than the max offset are
             // filtered, which is effectively the same as having the consumer drop an incomplete
             // batch returned in a fetch response.
-            if (batch.lastOffset() >= startOffset) {
-                if (batch.lastOffset() < maxOffset) {
-                    buffer = batch.writeTo(buffer);
-                }
+            if (batch.lastOffset() >= startOffset && batch.lastOffset() < maxOffset && !batch.entries.isEmpty()) {
```
hachikuji, line 373: I think this is a good change. I wonder if it creates a sort of reverse problem though where we end up not exercising multi-batch paths. Perhaps we could randomly choose 1-3 batches to return or something like that.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
```diff
@@ -426,48 +422,55 @@ public RawSnapshotWriter createSnapshot(OffsetAndEpoch snapshotId) {
...
-            OffsetAndEpoch snapshotId = snapshotIdOpt.get();
-            if (startOffset() < logStartSnapshotId.offset &&
-                highWatermark.offset >= logStartSnapshotId.offset &&
-                snapshotId.offset >= logStartSnapshotId.offset) {
+        if (snapshots.containsKey(snapshotId)) {
+            snapshots.headMap(snapshotId, false).clear();
 
-                snapshots.headMap(logStartSnapshotId, false).clear();
+            // Update the high watermark if it is less than the new log start offset
+            if (snapshotId.offset > highWatermark.offset) {
```
hachikuji: Maybe I am missing something, but how could this be possible given the check above?

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
```diff
@@ -184,6 +187,22 @@ Builder appendToLog(int epoch, List<String> records) {
             return this;
         }
 
+        Builder withSnapshot(OffsetAndEpoch snapshotId) throws IOException {
```
hachikuji: Maybe `withEmptySnapshot`?

raft/src/test/java/org/apache/kafka/raft/internals/RecordsIteratorTest.java
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.kafka.raft.internals;
+
+
```
hachikuji: nit: extra newlines

core/src/main/scala/kafka/tools/TestRaftServer.scala
jsancio: Fixed.

raft/src/main/java/org/apache/kafka/raft/BatchReader.java
jsancio: Yes. I documented the static methods for creating objects of this type.

raft/src/main/java/org/apache/kafka/raft/KafkaRaftClient.java
jsancio: Good idea. I decided to use `getClass().getTypeName()`.

raft/src/main/java/org/apache/kafka/raft/internals/RecordsIterator.java
jsancio: Yes. Added a comment.

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
jsancio: Yep. Changed the name as suggested. The name was inspired that a naming pattern used in Rust (https://doc.rust-lang.org/std/option/enum.Option.html#method.take) but I see that we use `takeSnapshot` in other parts of Kafka for creating snapshots.

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
jsancio: Yes. Good idea. Done.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
jsancio: I wanted to keep the tests reproducible. As a comprise I changed this code to read at most 2 batches. We can revisit this in the future and make it random but reproducible.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
jsancio: You are correct. This not needed.

raft/src/test/java/org/apache/kafka/raft/RaftClientTestContext.java
jsancio: Done.

raft/src/test/java/org/apache/kafka/raft/internals/RecordsIteratorTest.java
jsancio: Lines removed.

raft/src/test/java/org/apache/kafka/raft/MockLog.java
hachikuji: Sounds good.

raft/src/test/java/org/apache/kafka/raft/MockLogTest.java
```diff
@@ -75,56 +75,6 @@ public void testTopicId() {
         assertEquals(topicId, log.topicId());
     }
 
-    @Test
-    public void testAppendAsLeaderHelper() {
```
hachikuji, line 79: Why remove the test?

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
```diff
@@ -966,6 +967,54 @@ public void verify() {
         }
     }
 
+    private static class SnapshotAtLogStart implements Invariant {
```
hachikuji, line 970: Is there any noticeable change in simulation run time? It looks reasonable, but thought I'd check.

raft/src/test/java/org/apache/kafka/raft/MockLogTest.java
jsancio: I actually, move this to `testAppendAsLeader`. The two tests were testing almost the same things. I just merged them so that one test now covers all of the cases that didn't overlap.

raft/src/test/java/org/apache/kafka/raft/RaftEventSimulationTest.java
jsancio: Good catch. I made some improvements After the improvements the simulation time went from 1 minute in trunk to a 1:41 minutes for this PR. It looks like a good amount of that time is spent in MockLog.validateOffsetAndEpoch. Let me see what we can do but we can also remove that functionality from the verification.

jsancio: Okay. The simulations now take less than a minute in total on my machine which is what I see on trunk in my machine.

hachikuji (approved):
LGTM. Thanks for the patch!

PR #10090, opened 2021-02-09
Add BrokerMetadataListener and related classes

mumrah:
This adds BrokerMetadataListener which is responsible for processing metadata records received by the broker when running in Raft mode.

Note that several files in this PR are also included in https://github.com/apache/kafka/pull/10106 which will get merged to trunk first.

### DIFF
--- a/core/src/main/scala/kafka/server/metadata/BrokerMetadataListener.scala
+++ b/core/src/main/scala/kafka/server/metadata/BrokerMetadataListener.scala
@@ -46,7 +46,8 @@ class BrokerMetadataListener(val brokerId: Int,
                              val replicaManager: RaftReplicaManager,
                              val txnCoordinator: TransactionCoordinator,
                              val logManager: LogManager,
-                             val threadNamePrefix: Option[String]
+                             val threadNamePrefix: Option[String],
+                             val clientQuotaManager: ClientQuotaMetadataManager
                             ) extends MetaLogListener with KafkaMetricsGroup {
   val logContext = new LogContext(s"[BrokerMetadataListener id=${brokerId}] ")
   val log = logContext.logger(classOf[BrokerMetadataListener])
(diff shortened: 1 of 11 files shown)


core/src/main/scala/kafka/server/metadata/BrokerMetadataListener.scala
```diff
@@ -0,0 +1,260 @@
...
+class BrokerMetadataListener(val brokerId: Int,
+                             val time: Time,
+                             val metadataCache: RaftMetadataCache,
+                             val configRepository: LocalConfigRepository,
+                             val groupCoordinator: GroupCoordinator,
+                             val replicaManager: RaftReplicaManager,
+                             val txnCoordinator: TransactionCoordinator,
+                             val logManager: LogManager,
+                             val threadNamePrefix: Option[String]
+                            ) extends MetaLogListener with KafkaMetricsGroup {
```
rondagostino: Do any of these really need `val`?

core/src/main/scala/kafka/server/metadata/BrokerMetadataListener.scala
```diff
@@ -0,0 +1,260 @@
...
+  val logContext = new LogContext(s"[BrokerMetadataListener id=${brokerId}] ")
+  val log = logContext.logger(classOf[BrokerMetadataListener])
```
rondagostino, line 53: `private`?

core/src/main/scala/kafka/server/metadata/BrokerMetadataListener.scala
```diff
@@ -0,0 +1,260 @@
...
+        try {
+          if (isTraceEnabled) {
+            trace("Metadata batch %d: processing [%d/%d]: %s.".format(lastOffset, index + 1,
+              records.size(), record.toString()))
+          }
+          handleMessage(imageBuilder, record, lastOffset)
+        } catch {
+          case e: Exception => error(s"Unable to handle record ${index} in batch " +
+            s"ending at offset ${lastOffset}", e)
+        }
```
rondagostino, line 104: Perhaps not to resolve now, but I think error conditions like this might need more thought.  Should the broker die?  If so, how would the broker ever come back up if it can't get past a particular batch?  But what is the impact of not processing the remaining messages in a batch, and is that always acceptable or just sometimes acceptable?  Do we need metrics to track these error events?  It seems an error log message might not be sufficient?

PR #10095, opened 2021-02-10
MINOR: Introduce the KIP-500 Broker lifecycle manager

cmccabe:
The broker lifecycle manager handles registering the broker and sending periodic heartbeats, as described in KIP-631.  Based on the responses it receives from the controller, it transitions the broker through the states described in BrokerState.java: STARTING, RECOVERY, RUNNING, PENDING_CONTROLLED_SHUTDOWN, etc.


core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+
+  /**
+   * The broker rack, or null if there is no configured rack.
+   */
+  private val rack = config.rack.orNull
+
+  /**
+   * How long to wait for registration to succeed before failing the startup process.
+   */
+  private val initialTimeoutNs = NANOSECONDS.
```
ijuma: Nit: Isn't `TimeUnit.MILLISECONDS.toNanos(config.initialRegistrationTimeoutMs)` more readable?

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+    convert(config.initialRegistrationTimeoutMs.longValue(), TimeUnit.MILLISECONDS)
+
+  /**
+   * The exponential backoff to use for resending communication.
+   */
+  private val resendExponentialBackoff =
+    new ExponentialBackoff(100, 2, config.brokerSessionTimeoutMs.toLong, 0.1)
```
ijuma: We seem to use `0.2` as jitter elsewhere. Any reason to deviate? Maybe we should have an `ExponentialBackoff` static factory method that only takes the initial interval and the max interval. It seems that the other two parameters can be the same most of the time.

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+import org.apache.kafka.common.protocol.Errors
+import org.apache.kafka.common.requests.{BrokerHeartbeatRequest, BrokerHeartbeatResponse, BrokerRegistrationRequest, BrokerRegistrationResponse}
+import org.apache.kafka.metadata.{BrokerState, VersionRange}
+import org.apache.kafka.queue.EventQueue.DeadlineFunction
+import org.apache.kafka.common.utils.{ExponentialBackoff, LogContext, Time}
+import org.apache.kafka.queue.{EventQueue, KafkaEventQueue}
+import scala.jdk.CollectionConverters._
+
+
+class BrokerLifecycleManager(val config: KafkaConfig,
```
ijuma, line 54: Worth adding a comment explaining the thread safety aspects of this class and also its general purpose. I see a few volatile fields, but `failedAttempts` is not volatile (for example).

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+   * True if the initial registration succeeded.
+   * This variable can only be accessed from the event queue thread.
+   */
+  private var initialRegistrationSucceeded = false
+
+  /**
+   * The cluster ID, or null if this manager has not been started yet.
+   * This variable can only be accessed from the event queue thread.
+   */
+  private var _clusterId: Uuid = null
```
mumrah: Why are some members prefixed with an underscore?

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+   *
+   * @param highestMetadataOffsetProvider Provides the current highest metadata offset.
+   * @param channelManager                The brokerToControllerChannelManager to use.
+   * @param clusterId                     The cluster ID.
+   */
+  def start(highestMetadataOffsetProvider: () => Long,
+            channelManager: BrokerToControllerChannelManager,
+            clusterId: Uuid,
+            advertisedListeners: ListenerCollection,
+            supportedFeatures: util.Map[String, VersionRange]): Unit = {
```
ijuma, line 187: Why are we using a Java collection?

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+
+  def setReadyToUnfence(): Unit = {
+    eventQueue.append(new SetReadyToUnfenceEvent())
+  }
+
+  def brokerEpoch(): Long = _brokerEpoch
+
+  def state(): BrokerState = _state
+
+  class BeginControlledShutdownEvent extends EventQueue.Event {
```
ijuma: Are these events meant to be public?

ijuma (commented):
Thanks for the PR. A few comments below. Also, do you have a reference to how this class is used?

core/src/main/scala/kafka/server/KafkaConfig.scala
```diff
@@ -1497,6 +1509,9 @@ class KafkaConfig(val props: java.util.Map[_, _], doLog: Boolean, dynamicConfigO
   var brokerId: Int = getInt(KafkaConfig.BrokerIdProp)
   val nodeId: Int = getInt(KafkaConfig.NodeIdProp)
   val processRoles: Set[ProcessRole] = parseProcessRoles()
+  val initialRegistrationTimeoutMs = getInt(KafkaConfig.InitialBrokerRegistrationTimeoutMs)
+  val brokerHeartbeatIntervalMs = getInt(KafkaConfig.BrokerHeartbeatIntervalMsProp)
+  val brokerSessionTimeoutMs = getInt(KafkaConfig.BrokerSessionTimeoutMsProp)
```
ijuma: It's worth including the explicit type here since it's a public `val`. Are these meant to be nullable? It may be better to make then `Option[Int]` if so. Otherwise, we should make them `Int` so that they're easier to use (they're currently being inferred as `Integer`).

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+    }
+  }
+
+  class ShutdownEvent extends EventQueue.Event {
+    override def run(): Unit = {
+      info(s"Transitioning from ${_state} to ${BrokerState.SHUTTING_DOWN}.")
+      _state = BrokerState.SHUTTING_DOWN
+      controlledShutdownFuture.complete(null)
+      initialCatchUpFuture.cancel(false)
+      _channelManager.shutdown()
```
mumrah: Do we need to guard against `_channelManager` being null here? Is it possible that `close` is called without `start` having been called, maybe in some faulty startup scenario?

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+              if (message.data().isCaughtUp()) {
+                info(s"The broker has caught up. Transitioning from STARTING to RECOVERY.")
+                _state = BrokerState.RECOVERY
+                initialCatchUpFuture.complete(null)
+              } else {
+                info(s"The broker is STARTING. Still waiting to catch up with cluster metadata.")
+              }
+              // Schedule the heartbeat after only 10 ms so that in the case where
+              //there is no recovery work to be done, we start up a bit quicker.
+              scheduleNextCommunication(NANOSECONDS.convert(10, MILLISECONDS))
```
mumrah, line 379: We are scheduling the next heartbeat in the near future here so we don't have to wait for the full heartbeat interval in order to make it to RUNNING. But i think this will only happen if a SetReadyToUnfenceEvent is handled between the broker entering RECOVERY and the next heartbeat request, in which case that event handler will send a heartbeat right away. If that's the case, do we really need this 10ms here or should we just wait the normal heartbeat interval?

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+    convert(config.initialRegistrationTimeoutMs.longValue(), TimeUnit.MILLISECONDS)
+
+  /**
+   * The exponential backoff to use for resending communication.
+   */
+  private val resendExponentialBackoff =
+    new ExponentialBackoff(100, 2, config.brokerSessionTimeoutMs.toLong, 0.1)
+
+  /**
+   * The number of tries we've tried to communicate.
```
rondagostino: nit: append " and failed" or `s/tried/failed/`

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+  private var readyToUnfence = false
+
+  /**
+   * True if we sent a heartbeat to the active controller requesting controlled
+   * shutdown.
+   */
+  private var gotControlledShutdownResponse = false
+
+  /**
+   * Whether or not we this broker is registered with the controller quorum.
```
rondagostino: nit: `s/we this/this/`

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+            case BrokerState.PENDING_CONTROLLED_SHUTDOWN =>
+              if (!message.data().shouldShutDown()) {
+                info(s"The broker is in PENDING_CONTROLLED_SHUTDOWN state, still waiting " +
+                  "for the active controller.")
+                if (!gotControlledShutdownResponse) {
+                  // If this is the first pending controlled shutdown response we got,
+                  // schedule our next heartbeat a little bit sooner than we usually would.
+                  // In the case where controlled shutdown completes quickly, this will
+                  // speed things up a little bit.
+                  scheduleNextCommunication(NANOSECONDS.convert(50, MILLISECONDS))
```
mumrah, line 400: Since the heartbeat interval is configurable, maybe we should calculate this from the interval instead of a fixed value. Maybe something like (interval / 2) or sqrt(interval)?

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+
+  def brokerEpoch(): Long = _brokerEpoch
+
+  def state(): BrokerState = _state
+
+  class BeginControlledShutdownEvent extends EventQueue.Event {
+    override def run(): Unit = {
+      _state match {
+        case BrokerState.PENDING_CONTROLLED_SHUTDOWN =>
+          info(s"Attempted to enter controlled shutdown state, but we are already in " +
```
rondagostino: nit: `s/controlled/pending controlled/`

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+  /**
+   * The cluster ID, or null if this manager has not been started yet.
+   * This variable can only be accessed from the event queue thread.
+   */
+  private var _clusterId: Uuid = null
+
+  /**
+   * The listeners which this broker advertises.
+   */
+  private var _advertisedListeners: ListenerCollection = null
```
mumrah: nit: I think scala prefers the underscore initializer for cases like this. E.g., 

```scala
private var _advertisedListeners: ListenerCollection = _
```

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+
+  /**
+   * The broker id.
+   */
+  private val nodeId = config.nodeId
+
+  /**
+   * The broker rack, or null if there is no configured rack.
+   */
+  private val rack = config.rack.orNull
```
rondagostino: Make an `Option[String]`?  No need to propagate the unfortunate fact that rack can be null elsewhere?

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+                     advertisedListeners: ListenerCollection,
+                     supportedFeatures: util.Map[String, VersionRange]) extends EventQueue.Event {
+    override def run(): Unit = {
+      _highestMetadataOffsetProvider = highestMetadataOffsetProvider
+      _channelManager = channelManager
+      _channelManager.start()
+      _state = BrokerState.STARTING
+      _clusterId = clusterId
+      _advertisedListeners = advertisedListeners
+      _supportedFeatures = supportedFeatures
```
mumrah: Should we be copying from this map rather than assigning to it? Do we want things outside of this class to mutate the data in this map after start is called?

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
```diff
@@ -0,0 +1,457 @@
...
+
+  class StartupEvent(highestMetadataOffsetProvider: () => Long,
+                     channelManager: BrokerToControllerChannelManager,
+                     clusterId: Uuid,
+                     advertisedListeners: ListenerCollection,
+                     supportedFeatures: util.Map[String, VersionRange]) extends EventQueue.Event {
+    override def run(): Unit = {
+      _highestMetadataOffsetProvider = highestMetadataOffsetProvider
+      _channelManager = channelManager
+      _channelManager.start()
```
mumrah, line 254: If this class is owning the lifecycle for `_channelManager`, should it not also instantiate it?

mumrah (commented):
Thanks @cmccabe, this looks nice. I like the event driven nature of this manager along with the state machine 👍 

One thing I don't love is how BrokerLifecycleManager can be instantiated but some of the members are not initialized until start is called. Is there no way we can defer initialization until all the dependencies of this class are available?

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
rondagostino: I wonder if it might be helpful to assign each `var` into one of two buckets: stuff that can only be written to from the event queue thread but that can be read from any thread (must be `@volatile`), and stuff that is only used from within the event queue (doesn't need to be).  At a minimum put these into different commented sections, but maybe even create a single container object for the `@volatile` ones:


```
  private case class EventQueueThreadOwnedVars(@volatile var _brokerEpoch: Long = -1L,
                                               @volatile var _state: BrokerState = BrokerState.NOT_RUNNING)
  val eventQueueThreadOwnedVars = EventQueueThreadOwnedVars()
```

core/src/test/scala/unit/kafka/server/BrokerLifecycleManagerTest.scala
```diff
@@ -0,0 +1,217 @@
...
+    context.time.sleep(3000)
+    TestUtils.retry(10000) {
+      context.poll()
+      manager.eventQueue.wakeup()
+      Assertions.assertEquals(BrokerState.SHUTTING_DOWN, manager.state())
+    }
+    manager.controlledShutdownFuture.get()
+    manager.close()
+  }
+}
```
rondagostino: nit: missing newline

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: 20% jitter is kind of a lot for the heartbeat, I think.  20% means we could be heartbeating every 2.4 s rather than 2 s.

Thinking about it more I'd be more comfortable with 2%.  The point of jitter is just to avoid having a lot of heartbeats come in all at once, and I think 2% jitter is more than sufficient to accomplish that.  With a very small injection of randomness each time the heartbeat times will drift apart naturally, without the need for big divergences.

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: I added some JavaDoc for the class as a whole explaining the overall paradigm.

I think putting all the mutable state into a separate class is a bit too extreme.  I added notes about how each mutable variable should be used, however.

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: supportedFeatures is a static final map which basically explains what features the software supports.  So the intention is for it to be defined in common code which happens to be written in Java (not included in this PR).  It seemed simpler not to translate the map into scala, although I guess it's not too difficult either way.

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: that's a good point-- these can all be private.  Will fix.

core/src/main/scala/kafka/server/KafkaConfig.scala
cmccabe: Good point.  I added the explicit type of `Int`.

I don't think they need to be nullable... they have reasonable defaults as usual in KafkaConfig.  And they will be ignored altogether when in ZK mode, of course.

cmccabe:
Thanks for the reviews, @mumrah !

> One thing I don't love is how BrokerLifecycleManager can be instantiated but some of the members are not initialized until start is called. Is there no way we can defer initialization until all the dependencies of this class are available?

I think the next PR in the series makes things clearer.  The `BrokerServer` uses `BrokerLifecycleManager` to manage the broker state at all times, not just after the broker has been started or before it has been stopped.  Although there is a little bit of extra complexity in the lifecycle manager to support this usage, it makes the surrounding code much simpler because we can rely on this component to always have the state.  Therefore we don't need more mutable state and if statements in `BrokerServer`.

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: I guess you already know this but unlike Java, Scala doesn't support an accessor function having the same name as an instance variable.  So I kind of got into the habit of prefixing private variables with an underscore.  In this case I guess it's not needed since we don't have a clusterId accessor (yet?) but maybe it's easier just to be consistent.

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: good point.  we should check.

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: If there is no log recovery work to be done, a SetReadyToUnfenceEvent will probably be handled in this time.  Basically BrokerServer will just continue on and tell us it's ready to go.

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: Hmm.  I'm not sure.  Even if the heartbeat interval was 1 second or 3 seconds (just for example) I don't think we'd want this interval to change.

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: We do it this way to make unit tests easier, since we need to supply a mock channel manager at times

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: I'll make a defensive copy

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
cmccabe: I guess we could have the responsibility for closing the channel manager lie with the external callers.  Would that be better here?  It would be clearer, I guess.

mumrah (approved):
Thanks for the explanation on the initialization/startup thing @cmccabe. I think this LGTM 👍

core/src/main/scala/kafka/server/BrokerLifecycleManager.scala
mumrah: Yea, fair enough. Maybe we can at least consolidate on one magic "short" time and make it a constant? Now we have 10ms and 50ms hard coded.

PR #10096, opened 2021-02-10
KAFKA-12268: Make early poll return opt-in

vvcephei:
* Revert the default Consumer#poll behavior back to early return on records only
* Add config to enable early return on metadata or records
* Set the return-on-metadata config in Streams to support KIP-695
* Revert the undesired addition of Hamcrest to the Client module
* Revert the now unnecessary poll-until-records logic in PlaintextConsumerTest

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerConfig.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerConfig.java
@@ -71,6 +71,21 @@ public class ConsumerConfig extends AbstractConfig {
     /** <code>max.poll.interval.ms</code> */
     public static final String MAX_POLL_INTERVAL_MS_CONFIG = CommonClientConfigs.MAX_POLL_INTERVAL_MS_CONFIG;
     private static final String MAX_POLL_INTERVAL_MS_DOC = CommonClientConfigs.MAX_POLL_INTERVAL_MS_DOC;
+
+    /** <code>long.poll.mode</code> */
+    public static final String LONG_POLL_RETURN_ON_RECORDS = "return_on_records";
+    public static final String LONG_POLL_RETURN_ON_RESPONSE = "return_on_response";
+    public static final String LONG_POLL_MODE_CONFIG = "long.poll.mode";
+    public static final String LONG_POLL_MODE_DOC = "Whether a call to Consumer#poll(...) should block until" +
+        " records are received (return_on_records), or should return early if a metadata-only response is received" +
+        " from the brokers (return_on_metadata). The 'return_on_records' mode is the default, and it matches prior" +
+        " behavior. The 'return_on_response' mode allows callers to witness topic metadata changes via" +
+        " ConsumerRecords#metadata() as soon as they are received in a fetch response, even if there are no records" +
+        " included in the response. Callers who are only interested in seeing records should leave the default in" +
+        " place, and callers who are interested in maintaining fresh local information about the current lag should" +
+        " enable 'return_on_response'." +
+        " Note that the Consumer metrics are always updated upon receipt of the fetch responses.";
+
     /**
      * <code>session.timeout.ms</code>
      */
@@ -519,6 +534,12 @@ public class ConsumerConfig extends AbstractConfig {
                                         atLeast(1),
                                         Importance.MEDIUM,
                                         MAX_POLL_INTERVAL_MS_DOC)
+                                .define(LONG_POLL_MODE_CONFIG,
+                                        Type.STRING,
+                                        LONG_POLL_RETURN_ON_RECORDS,
+                                        in(LONG_POLL_RETURN_ON_RECORDS, LONG_POLL_RETURN_ON_RESPONSE),
+                                        Importance.LOW,
+                                        LONG_POLL_MODE_DOC)
                                 .define(EXCLUDE_INTERNAL_TOPICS_CONFIG,
                                         Type.BOOLEAN,
                                         DEFAULT_EXCLUDE_INTERNAL_TOPICS,
--- a/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
+++ b/clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
@@ -586,6 +586,7 @@ public class KafkaConsumer<K, V> implements Consumer<K, V> {
     private final long retryBackoffMs;
     private final long requestTimeoutMs;
     private final int defaultApiTimeoutMs;
+    private final boolean longPollExitEarlyOnMetadata;
     private volatile boolean closed = false;
     private List<ConsumerPartitionAssignor> assignors;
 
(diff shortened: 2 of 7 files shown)


vvcephei:
Hey @rajinisivaram and @chia7712 ,

Here is my proposed fix for https://issues.apache.org/jira/browse/KAFKA-12268 , based on our discussion in https://github.com/apache/kafka/pull/10022 and in the mailing list (https://lists.apache.org/thread.html/ra3a73e74c2bb9768ce33dcf0492e2f0fd2f120044541c6b216a67cca%40%3Cdev.kafka.apache.org%3E)

I don't know if anyone has a preference on the config name itself. I also considered just `enable.early.poll.return.on.metadata=false|true`, but _maybe_ there will be some other behavior we might add as another "mode" in the future. No strong opinion here.

Also, please take a look at the logic in the Consumer. I did my best to make it both efficient and legible, but I'm not sure I achieved either goal.

Thanks for spotting the problem and for the discussion, and thanks in advance for the reviews!

Also, as the original reviewer of the feature, I'd appreciate a review from @guozhangwang if you have time.

clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
```diff
@@ -1236,7 +1243,17 @@ public void assign(Collection<TopicPartition> partitions) {
                 }
 
                 final FetchedRecords<K, V> records = pollForFetches(timer);
-                if (!records.isEmpty()) {
+
+                // We only need to save off the metadata if we are NOT supposed to return early on metadata-only responses
+                if (!longPollExitEarlyOnMetadata) {
```
chia7712, line 1248: How about updating `nullableSeenMetadata` only if we don't return records?
```java
if (longPollShouldReturn(records)) { 
 ...
} else {
  if (nullableSeenMetadata == null) {
    nullableSeenMetadata = new HashMap<>(records.metadata());
  } else {
    nullableSeenMetadata.putAll(records.metadata());
  }
}
```

The benefit from above code is that we don't need to handle duplicate metadata which exists on both `FetchedRecords` and `nullableSeenMetadata` when it succeed to get records and metadata in first loop.

streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
```diff
@@ -1201,6 +1201,9 @@ private void verifyMaxInFlightRequestPerConnection(final Object maxInFlightReque
         // disable auto topic creation
         consumerProps.put(ConsumerConfig.ALLOW_AUTO_CREATE_TOPICS_CONFIG, "false");
 
+        // enable early return on metadata
+        consumerProps.put(ConsumerConfig.LONG_POLL_MODE_CONFIG, ConsumerConfig.LONG_POLL_RETURN_ON_RECORDS);
```
chia7712, line 1205: It should be `LONG_POLL_RETURN_ON_RESPONSE` rather than `LONG_POLL_RETURN_ON_RECORDS`

chia7712 (commented):
@vvcephei Thanks for nice patch. Some minor comments are left. Please take a look.

Also, could you update javadocs of `KafkaConsumer#poll` to mention this new config?

clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
```diff
@@ -2480,4 +2508,38 @@ String getClientId() {
...
+            if (nullableMetadataMap == null || nullableMetadataMap.isEmpty()) {
+                continue;
+            } else {
+                if (result == null) {
+                    result = new HashMap<>(nullableMetadataMap.size());
+                }
+                for (Map.Entry<TopicPartition, FetchedRecords.FetchMetadata> entry : nullableMetadataMap.entrySet()) {
+                    result.put(
+                        entry.getKey(),
+                        new ConsumerRecords.Metadata(
```
chia7712, line 2530: Not sure why we need to have `FetchedRecords.FetchMetadata`? It is always converted to `ConsumerRecords.Metadata` directly. The `records` in `FetchedRecords` is `ConsumerRecord`. Maybe we can make `FetchedRecords` use `ConsumerRecords.Metadata` also.

chia7712 (commented):
It seems to me the constructors of `ConsumerRecords` and `ConsumerRecords.Metadata` should be package-private rather than public. Otherwise, it could cause troubles when we are going to add more data in the future. WDYT?

clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerConfig.java
```diff
@@ -71,6 +71,21 @@
...
+    public static final String LONG_POLL_MODE_CONFIG = "long.poll.mode";
+    public static final String LONG_POLL_MODE_DOC = "Whether a call to Consumer#poll(...) should block until" +
+        " records are received (return_on_records), or should return early if a metadata-only response is received" +
+        " from the brokers (return_on_metadata). The 'return_on_records' mode is the default, and it matches prior" +
+        " behavior. The 'return_on_response' mode allows callers to witness topic metadata changes via" +
+        " ConsumerRecords#metadata() as soon as they are received in a fetch response, even if there are no records" +
+        " included in the response. Callers who are only interested in seeing records should leave the default in" +
+        " place, and callers who are interested in maintaining fresh local information about the current lag should" +
+        " enable 'return_on_response'." +
+        " Note that the Consumer metrics are always updated upon receipt of the fetch responses.";
```
ijuma, line 87: Hmm, this doesn't seem great.

ijuma: In what KIP was this approved?

vvcephei: Thanks @ijuma ,

This is KIP-695. The currently approved version of it includes just a behavior change and no config. The behavior change is that, if you call `Consumer#poll` and there are some completed fetch responses, but none of them have records, we still return a `ConsumerRecords` object with no records so that we can return the new metadata (via `ConsumerRecords#metadata()`.

That behavior change struck @chia7712 and @rajinisivaram as not ok (for reasons discussed in https://lists.apache.org/thread.html/ra3a73e74c2bb9768ce33dcf0492e2f0fd2f120044541c6b216a67cca%40%3Cdev.kafka.apache.org%3E ).

In that thread, we agreed to instead introduce an opt-in config, which is what this doc is about. Which part doesn't seem great to you?

streams/src/main/java/org/apache/kafka/streams/StreamsConfig.java
vvcephei: Oy, you are right. Thanks!

clients/src/main/java/org/apache/kafka/clients/consumer/KafkaConsumer.java
vvcephei: sounds good!

clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerConfig.java
ijuma: Having yet another config for something people should not really care about. I think you should revive the KIP thread with the proposed updates and give people a chance to comment on it. Also the KIP should cover the compatibility impact that is behind this update.

vvcephei: Thanks @ijuma .

I do agree with you. In the vote thread, others made the case that the behavior change is not ok. If you disagree with the conclusion of the vote thread and want to argue that it actually is ok for `Consumer#poll` to return when there are only metadata updates, I'm happy to reconsider the decision.

I didn't update the text of the KIP yet because I suspected that this PR might inspire a new round of discussion.

vvcephei:
Replced by https://github.com/apache/kafka/pull/10137

PR #10098, opened 2021-02-10
KAFKA-12321 the comparison function for uuid type should be 'equals' …

chia7712:
issue: https://issues.apache.org/jira/browse/KAFKA-12321

### current comparison
```java
if (this.taggedUuid != Uuid.fromString("H3KKO4NTRPaCWtEmm3vW7A"))
```

### fixed comparison
```java
if (!this.taggedUuid.equals(Uuid.fromString("H3KKO4NTRPaCWtEmm3vW7A")))
```

### DIFF
--- a/clients/src/test/java/org/apache/kafka/common/message/MessageTest.java
+++ b/clients/src/test/java/org/apache/kafka/common/message/MessageTest.java
@@ -49,6 +49,7 @@ import org.apache.kafka.common.protocol.ApiKeys;
 import org.apache.kafka.common.protocol.ByteBufferAccessor;
 import org.apache.kafka.common.protocol.Errors;
 import org.apache.kafka.common.protocol.Message;
+import org.apache.kafka.common.protocol.MessageUtil;
 import org.apache.kafka.common.protocol.ObjectSerializationCache;
 import org.apache.kafka.common.protocol.types.RawTaggedField;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


clients/src/test/java/org/apache/kafka/common/message/MessageTest.java
```diff
@@ -749,6 +750,13 @@ public void testProduceResponseVersions() throws Exception {
         }
     }
 
+    @Test
+    public void defaultValueShouldBeWritable() {
+        for (short version = SimpleExampleMessageData.LOWEST_SUPPORTED_VERSION; version <= SimpleExampleMessageData.HIGHEST_SUPPORTED_VERSION; ++version) {
+            MessageUtil.toByteBuffer(new SimpleExampleMessageData(), version);
```
ijuma, line 756: This fails without the change in this PR?

ijuma (approved):
Good catch, LGTM assuming the question below is true. We should also backport it to older branches.

clients/src/test/java/org/apache/kafka/common/message/MessageTest.java
chia7712: > This fails without the change in this PR?

yep. The generated code is shown below.
```java
            if (this.taggedUuid != Uuid.fromString("H3KKO4NTRPaCWtEmm3vW7A")) {
                throw new UnsupportedVersionException("Attempted to write a non-default taggedUuid at version " + _version);
            }
```

chia7712:
push to trunk, 2.8, 2.7 and 2.6

PR #10101, opened 2021-02-10
Add ClientQuotaMetadataManager for processing QuotaRecord 

mumrah:
This PR brings in the new broker metadata processor for handling QuotaRecord-s coming from the metadata log. Also included is a new cache class to allow for fast lookups of quotas on the broker for handling DescribeClientQuotaRequest. 

See the original PR here https://github.com/confluentinc/kafka/pull/477

### DIFF
--- a/core/src/main/scala/kafka/server/metadata/ClientQuotaCache.scala
+++ b/core/src/main/scala/kafka/server/metadata/ClientQuotaCache.scala
@@ -1,0 +1,284 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package kafka.server.metadata
+
+import kafka.utils.CoreUtils.{inReadLock, inWriteLock}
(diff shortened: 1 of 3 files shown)


core/src/main/scala/kafka/server/metadata/ClientQuotaCache.scala
```diff
@@ -0,0 +1,284 @@
...
+
+
+class ClientQuotaCache {
+  private type QuotaCacheIndex = mutable.HashMap[CacheIndexKey, mutable.HashSet[QuotaEntity]]
+
+  // A mapping of the quota entities to their quotas, for example:
+  // {
+  //   (user:alice) -> {consumer_byte_rate: 10000},
+  //   (user:alice,client:x) -> {consumer_byte_rate: 8000, producer_byte_rate: 8000}
+  // }
```
cmccabe: can we do this as JavaDoc (and maybe use a unicode arrow, while you're at it, to avoid the issues JavaDoc always has with brackets)?

core/src/main/scala/kafka/server/metadata/ClientQuotaMetadataManager.scala
```diff
@@ -0,0 +1,174 @@
...
+      case QuotaConfigs.CONSUMER_BYTE_RATE_OVERRIDE_CONFIG => quotaManagers.fetch
+      case QuotaConfigs.PRODUCER_BYTE_RATE_OVERRIDE_CONFIG => quotaManagers.produce
+      case QuotaConfigs.REQUEST_PERCENTAGE_OVERRIDE_CONFIG => quotaManagers.request
+      case QuotaConfigs.CONTROLLER_MUTATION_RATE_OVERRIDE_CONFIG => quotaManagers.controllerMutation
+      case _ =>
+        warn(s"Ignoring unexpected quota key ${quotaRecord.key()} for entity $quotaEntity")
+        return
+    }
+
+    quotaCache.updateQuotaCache(quotaEntity, quotaRecord.key, quotaRecord.value, quotaRecord.remove)
```
cmccabe: does it make sense to do this update at the end, in case there's an exception in between?  or is that too paranoid

mumrah: Actually that seems pretty reasonable. The quota managers could theoretically fail to accept and apply a quota for whatever reason. And in that case we probably don't want to cache the quota

cmccabe (approved):
LGTM

PR #10102, opened 2021-02-10
KAFKA-12272: Fix commit-interval metrics

mjsax:
Call for review @ableegoldman

Fixes a regression bug. Should be cherry-picked back to 2.6 branch.

### DIFF
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -776,10 +776,10 @@ public class StreamThread extends Thread {
 
                 log.debug("{} punctuators ran.", punctuated);
 
-                final long beforeCommitTs = now;
+                final long beforeCommitMs = now;
                 final int committed = maybeCommit();
                 totalCommittedSinceLastSummary += committed;
-                final long commitLatency = Math.max(now - beforeCommitTs, 0);
+                final long commitLatency = Math.max(now - beforeCommitMs, 0);
                 totalCommitLatency += commitLatency;
                 if (committed > 0) {
                     commitSensor.record(commitLatency / (double) committed, now);
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -776,9 +776,10 @@ public class StreamThread extends Thread {
 
                 log.debug("{} punctuators ran.", punctuated);
 
+                final long beforeCommitTs = now;
                 final int committed = maybeCommit();
                 totalCommittedSinceLastSummary += committed;
-                final long commitLatency = advanceNowAndComputeLatency();
+                final long commitLatency = Math.max(now - beforeCommitTs, 0);
                 totalCommitLatency += commitLatency;
                 if (committed > 0) {
                     commitSensor.record(commitLatency / (double) committed, now);
--- a/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
+++ b/streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
@@ -1021,7 +1021,7 @@ public class StreamThread extends Thread {
             if (committed == -1) {
                 log.debug("Unable to commit as we are in the middle of a rebalance, will try again when it completes.");
             } else {
-                advanceNowAndComputeLatency();
+                now = time.milliseconds();
                 lastCommitMs = now;
             }
         } else {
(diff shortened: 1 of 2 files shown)


streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -776,9 +776,10 @@ void runOnce() {
 
                 log.debug("{} punctuators ran.", punctuated);
 
+                final long beforeCommitTs = now;
```
ableegoldman: such a small nit that I don't think it's worth pushing a fix and restarting the build, but should this be `beforeCommitMs` rather than `Ts`?

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
```diff
@@ -776,9 +776,10 @@ void runOnce() {
...
                 final int committed = maybeCommit();
                 totalCommittedSinceLastSummary += committed;
-                final long commitLatency = advanceNowAndComputeLatency();
+                final long commitLatency = Math.max(now - beforeCommitTs, 0);
```
ableegoldman: Also kind of a nit, since technically this does work, but wouldn't it make more sense to just remove the `advanceNowAndComputeLatency` call in `maybeCommit`, and then just call `advancedNowAndComputeLatency` here as before? Otherwise we're just computing the latency inside `maybeCommit` for no reason, and throwing out the result.

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
mjsax: Happy to address.

ableegoldman (approved):
This looks good enough to me, but I think it could be cleaned up just a bit to avoid calling `advanceNowAndComputeLatency` just to advance now, but then throwing out the latency. But I'm ok with it as is for now, especially since we want to get this merged quickly

streams/src/main/java/org/apache/kafka/streams/processor/internals/StreamThread.java
mjsax: Well, we need to advance time _within_ `maybeCommit()` to make sure we compute `lastCommitMs` correctly (what was exactly the fix that introduced the commit-latency metric regression).

mjsax: We could replace `advancedNowAndComputeLatency` by `now = time.milliseconds()` in `maybeCommit()` though. -- In the original PR, we just called `advancedNowAndComputeLatency` to keep the patter of a single place in the code that does call `time.milliseconds`

ableegoldman: But currently we call `advanceNowAndComputeLatency` pretty much at the end of the `maybeCommit` method, so there shouldn't be a noticeable difference between advancing `now` at the end of `maybeCommit` vs advancing it immediately after `maybeCommit` returns, right?
I'm also ok with just advancing `now` inside `maybeCommit`. The main thing that felt off was just that we compute the latency for no reason inside `maybeCommit`, and ignore the result

ableegoldman: (and similarly move the `lastCommitMs = now` to after `maybeCommit` returns)

mjsax: > so there shouldn't be a noticeable difference between advancing now at the end of maybeCommit vs advancing it immediately after maybeCommit returns, right?

Well, not from a runtime point of view.

> (and similarly move the lastCommitMs = now to after maybeCommit returns)

Exactly: we need to first advance `now` and than update `lastCommitMs` -- but I would rather not update `lastCommitMs` outside of `maybeCommit`, because it breaks encapsulation?

> I'm also ok with just advancing now inside maybeCommit. The main thing that felt off was just that we compute the latency for no reason inside maybeCommit, and ignore the result

Yes, that is a little odd -- as I said, happy to change `advancedNowAndComputeLatency` to `now = time.milliseconds()` within `maybeCommit()`.

ableegoldman: > Yes, that is a little odd -- as I said, happy to change advancedNowAndComputeLatency to now = time.milliseconds() within maybeCommit().

Sounds good to me, let's do it

mjsax:
Thanks @ableegoldman -- updated the PR. If there are no further concerns raised, I'll merge after Jenkins passed and cherry-pick to `2.8` branch.

ableegoldman (approved):
LGTM! We just need to backport this to 2.8 right (edit: duh, didn't see your last comment)

mjsax:
Merged to `trunk` and cherry-picked to `2.8` branch.

PR #10105, opened 2021-02-10
MINOR: System tests for Raft-based metadata quorums

rondagostino:
Adds a new `sanity_checks/bounce_test.py` system test for a simple produce/bounce/produce series of events.  Augments this and other `sanity_checks` tests so they run all metadata quorum types: ZooKeeper, remote Raft, and co-located Raft.

Augments several `tests/client` and `tests/core` system tests to run for remote Raft-based metadata quorums in addition to ZooKeeper.  Co-located Raft controllers are not tested here as this configuration is presumed to work assuming `sanity_checks` as well as any unit/integration tests pass.

Augments a simple set of tests in `tests/connect`, `tests/streams`, and `tests/tools` to run for remote Raft-based metadata quorums in addition to ZooKeeper.

Adds missing `@cluster` annotations for dozens of system tests that were missing them -- this was causing these tests to grab the entire cluster of nodes and negatively impact the parallelism of the system test run.


rondagostino:
There are 6 failing tests across these 3 test classes:

**kafka.server.KafkaConfigTest**
```
KafkaConfigTest > testFromPropsInvalid() FAILED
    org.opentest4j.AssertionFailedError: Expected exception for property `broker.heartbeat.interval.ms` with invalid value `-1` was not thrown ==> Expected java.lang.Exception to be thrown, but nothing was thrown.
```

**AutoTopicCreationManagerTest**
```
AutoTopicCreationManagerTest > testCreateTxnTopic() FAILED
    Wanted but not invoked:
    brokerToControllerChannelManager.sendRequest(
        CreateTopicsRequestData(topics=[CreatableTopic(name='__transaction_state', numPartitions=2, replicationFactor=2, assignments=[], configs=[])], timeoutMs=100, validateOnly=false),
        <any kafka.server.ControllerRequestCompletionHandler>
    );
    -> at kafka.server.AutoTopicCreationManagerTest.testCreateTopic(AutoTopicCreationManagerTest.scala:119)
    Actually, there were zero interactions with this mock.

AutoTopicCreationManagerTest > testCreateOffsetTopic() FAILED
    Wanted but not invoked:
    brokerToControllerChannelManager.sendRequest(
        CreateTopicsRequestData(topics=[CreatableTopic(name='__consumer_offsets', numPartitions=2, replicationFactor=2, assignments=[], configs=[])], timeoutMs=100, validateOnly=false),
        <any kafka.server.ControllerRequestCompletionHandler>
    );
    -> at kafka.server.AutoTopicCreationManagerTest.testCreateTopic(AutoTopicCreationManagerTest.scala:119)
    Actually, there were zero interactions with this mock.

AutoTopicCreationManagerTest > testCreateNonInternalTopic() FAILED
    Wanted but not invoked:
    brokerToControllerChannelManager.sendRequest(
        CreateTopicsRequestData(topics=[CreatableTopic(name='topic', numPartitions=1, replicationFactor=1, assignments=[], configs=[])], timeoutMs=100, validateOnly=false),
        <any kafka.server.ControllerRequestCompletionHandler>
    );
    -> at kafka.server.AutoTopicCreationManagerTest.testCreateTopic(AutoTopicCreationManagerTest.scala:119)
    Actually, there were zero interactions with this mock.
```

**MetadataRequestWithForwardingTest**
```
MetadataRequestWithForwardingTest > testAutoCreateOfCollidingTopics() FAILED
    org.opentest4j.AssertionFailedError: expected: <Set(LEADER_NOT_AVAILABLE, INVALID_TOPIC_EXCEPTION)> but was: <Set(UNKNOWN_TOPIC_OR_PARTITION)>

MetadataRequestWithForwardingTest > testAutoTopicCreation() FAILED
    org.opentest4j.AssertionFailedError: expected: <LEADER_NOT_AVAILABLE> but was: <UNKNOWN_TOPIC_OR_PARTITION>
```

build.gradle
```diff
@@ -1029,6 +1029,7 @@ project(':metadata') {
     compile project(':clients')
     compile libs.jacksonDatabind
     compile libs.jacksonJDK8Datatypes
+    compile libs.metrics
```
ijuma: Let's file a JIRA to discuss whether the metadata module should use Kafka metrics or Yammer metrics.

rondagostino: > file a JIRA to discuss whether the metadata module should use Kafka metrics or Yammer metrics.

https://issues.apache.org/jira/browse/KAFKA-12348

cc: @cmccabe

rondagostino:
http://confluent-kafka-branch-builder-system-test-results.s3-us-west-2.amazonaws.com/2021-02-20--001.1613812911--rondagostino--kip500_full--f59453faa/report.html

Tests|Passes|Failures|Ignored|Time
940|558|185|197|359 minutes

Some observations.  First, adding the missing @cluster annotations really helped.  Running almost 750 tests in under 6 hours is a dramatic improvement.  Here’s the stats from a run 2 days ago:

Tests|Passes|Failures|Ignored|Time
729|427|105|197|388 minutes

We are running 743/532 = 40% more tests, but we are able to finish in 10% less time.  So that’s a big win.

We have 130 more tests passing and 80 more tests failing.  It is VERY difficult to find the signal amid so much noise.  A quick browse through our results shows lots of test failures that are not augmented for Raft at all, and many test that are augmented are failing for both ZK and Raft cases.

build.gradle
cmccabe: thanks. I think this will be a post-2.8 thing

tests/kafkatest/services/console_consumer.py
```diff
@@ -151,7 +151,10 @@ def prop_file(self, node):
     def start_cmd(self, node):
         """Return the start command appropriate for the given node."""
         args = self.args.copy()
-        args['zk_connect'] = self.kafka.zk_connect_setting()
+        if self.new_consumer:
+            args['broker_list'] = self.kafka.bootstrap_servers(self.security_config.security_protocol)
+        else:
+            args['zk_connect'] = self.kafka.zk_connect_setting()
```
cmccabe, line 156: This is a slight behavior change for the old consumer, right?  Previously we set both `broker_list` and `zk_connect`, but now we set only `zk_connect`. Can we keep setting both for now? (I agree we should skip setting zk_connect for the new consumer)

tests/kafkatest/services/performance/end_to_end_latency.py
```diff
@@ -74,12 +74,15 @@ def __init__(self, context, num_nodes, kafka, topic, num_records, compression_ty
     def start_cmd(self, node):
         args = self.args.copy()
         args.update({
-            'zk_connect': self.kafka.zk_connect_setting(),
             'bootstrap_servers': self.kafka.bootstrap_servers(self.security_config.security_protocol),
             'config_file': EndToEndLatencyService.CONFIG_FILE,
             'kafka_run_class': self.path.script("kafka-run-class.sh", node),
             'java_class_name': self.java_class_name()
         })
+        if node.version < V_0_9_0_0:
```
cmccabe: It would be good to have a function in `version.py` to reflect this

tests/kafkatest/tests/client/client_compatibility_features_test.py
```diff
@@ -107,7 +108,9 @@ def invoke_compatibility_program(self, features):
           self.logger.info("** Command failed.  See %s for log messages." % ssh_log_file)
           raise
 
-    @parametrize(broker_version=str(DEV_BRANCH))
+    @cluster(num_nodes=7)
+    @matrix(broker_version=[str(DEV_BRANCH)], metadata_quorum=quorum.all_non_upgrade)
+    @cluster(num_nodes=7)
```
cmccabe: It seems like `@cluster` is repeated twice here.  We only need it once, right?

tests/kafkatest/tests/client/client_compatibility_features_test.py
```diff
@@ -120,10 +123,9 @@ def invoke_compatibility_program(self, features):
     @parametrize(broker_version=str(LATEST_2_3))
     @parametrize(broker_version=str(LATEST_2_4))
     @parametrize(broker_version=str(LATEST_2_5))
-    @parametrize(broker_version=str(LATEST_2_6))
-    @parametrize(broker_version=str(LATEST_2_7))
```
cmccabe, line 124: hmm.  we don't want to drop 2.6 and 2.7 here, do we?

tests/kafkatest/tests/client/client_compatibility_produce_consume_test.py
```diff
@@ -46,13 +47,16 @@ def __init__(self, test_context):
...
+            self.zk.start()
 
     def min_cluster_size(self):
         # Override this since we're adding services outside of the constructor
         return super(ClientCompatibilityProduceConsumeTest, self).min_cluster_size() + self.num_producers + self.num_consumers
 
-    @parametrize(broker_version=str(DEV_BRANCH))
+    @cluster(num_nodes=9)
+    @matrix(broker_version=[str(DEV_BRANCH)], metadata_quorum=quorum.all_non_upgrade)
+    @cluster(num_nodes=9)
```
cmccabe: `@cluster` is repeated again

tests/kafkatest/tests/core/round_trip_fault_test.py
```diff
@@ -47,32 +55,48 @@ def __init__(self, test_context):
...
+            self.zk.stop()
 
-    def test_round_trip_workload(self):
+    def remote_quorum_nodes(self):
+        if quorum.for_test(self.test_context) == quorum.zk:
+            return self.zk.nodes
+        elif quorum.for_test(self.test_context) == quorum.remote_raft:
+            return self.kafka.controller_quorum.nodes
+        else: # co-located case, which we currently don't test but handle here for completeness in case we do test it
+            return []
```
cmccabe, line 75: should this throw an exception? Or does the current code actually work for this case?

cmccabe:
Good work spotting the missing `@cluster` annotations! That is quite a performance win. I wonder if we should disallow tests without this annotation in the future.

tests/kafkatest/services/performance/end_to_end_latency.py
rondagostino: > would be good to have a function in version.py to reflect [functionality unsupported below 0.9.0]

I'll add this function:

```
    def supports_bootstrap_server(self):
        return self >= V_0_9_0_0
```

I then made the changes all over since this 0.9.0 constant is checked in several places:

```
tests/kafkatest/services/console_consumer.py
tests/kafkatest/services/performance/consumer_performance.py
tests/kafkatest/services/performance/end_to_end_latency.py
tests/kafkatest/services/performance/producer_performance.py
tests/kafkatest/tests/core/upgrade_test.py
tests/kafkatest/tests/core/upgrade_test.py
```

We can revert if you think I went too far.

tests/kafkatest/tests/core/round_trip_fault_test.py
rondagostino: > throw an exception or does the current code actually work for this case?

The code always needs `<Kafka nodes>` + `<remote quorum nodes>`, where the latter are the ZooKeeper or remote Controller Quorum nodes.  If we were to run this test with co-located Raft Quorum controllers then those nodes would be accounted for because they are part of the Kafka nodes, so there is no need to explicitly add them.  So this code is correct in that it return an empty list for that case.  As was indicated in the comment, it's here just in case we ever decide we want to test it.

rondagostino:
> I wonder if we should disallow tests without `[the @cluster]` annotation in the future.

I think perhaps yes.  Is there a need to allow it?  A simple oversight generates a significant parallelism hit at this point -- and they compound quickly as evidenced here.

tests/kafkatest/services/performance/end_to_end_latency.py
cmccabe: thanks.  that sounds good!

can you name it something like `consumer_supports_bootstrap_server`?  There are a bunch of other bootstrap server functions (`acl_command_supports_bootstrap_server`, `topic_command_supports_bootstrap_server`, etc.) so it would be good to be clear

tests/kafkatest/tests/core/round_trip_fault_test.py
cmccabe: sounds good

tests/kafkatest/services/performance/end_to_end_latency.py
rondagostino: > name it something like consumer_supports_bootstrap_server... bunch of other bootstrap server functions

Good point.  I renamed it `consumer_supports_bootstrap_server()` as you suggested.

cmccabe (approved):
LGTM. Thanks, @rondagostino !

PR #10106, opened 2021-02-10
MINOR: add the MetaLogListener, LocalLogManager, and Controller interface.

cmccabe:
Add MetaLogListener, LocalLogManager, and related classes.  These
classes are used by the KIP-500 controller and broker to interface with the
Raft log.

Also add the Controller interface.  The implementation will be added in a separate PR.


metadata/src/main/java/org/apache/kafka/controller/Controller.java
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * Decommission a broker.
+     *
+     * @param brokerId      The broker id to decommission.
+     *
+     * @return              A future that is completed successfully hwne the broker is
+     *                      decommissioned, or if it is not registered in the first place.
+     */
+    CompletableFuture<Void> decommissionBroker(int brokerId);
```
rondagostino: `unregister` instead of `decommission` given the recent rename?

metadata/src/main/java/org/apache/kafka/controller/Controller.java
```diff
@@ -0,0 +1,180 @@
...
+import org.apache.kafka.metadata.FeatureMapAndEpoch;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+
+public interface Controller extends AutoCloseable {
+    /**
+     * Change partition ISRs.
```
mumrah: Should we spell out ISR? Though I suppose anyone working with the controller code should know what it stands for :)

metadata/src/main/java/org/apache/kafka/controller/ResultOrError.java
```diff
@@ -0,0 +1,76 @@
...
+
+package org.apache.kafka.controller;
+
+import org.apache.kafka.common.protocol.Errors;
+import org.apache.kafka.common.requests.ApiError;
+
+import java.util.Objects;
+
+
+class ResultOrError<T> {
```
mumrah, line 26: Should we have methods like `isError` and `isResult` on this class?

metadata/src/main/java/org/apache/kafka/metadata/BrokerRegistration.java
```diff
@@ -0,0 +1,153 @@
...
+        this.id = id;
+        this.epoch = epoch;
+        this.incarnationId = incarnationId;
+        Map<String, Endpoint> listenersMap = new HashMap<>();
+        for (Endpoint endpoint : listeners) {
+            listenersMap.put(endpoint.listenerName().get(), endpoint);
+        }
+        this.listeners = Collections.unmodifiableMap(listenersMap);
+        Objects.requireNonNull(supportedFeatures);
+        this.supportedFeatures = supportedFeatures;
```
mumrah: Defensive copy of supportedFeatures?

metadata/src/main/java/org/apache/kafka/metadata/BrokerRegistration.java
```diff
@@ -0,0 +1,153 @@
...
+        Objects.requireNonNull(rack);
+        this.rack = rack;
+        this.fenced = fenced;
+    }
+
+    public BrokerRegistration(int id,
```
mumrah, line 65: In the other constructor we have some nice preconditions and defensive copies, should we do the same here? Or maybe this constructor is meant to be private to `cloneWithFencing` below

metadata/src/main/java/org/apache/kafka/metalog/metalog/LocalLogManager.java
```diff
@@ -0,0 +1,378 @@
...
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.TreeMap;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Collectors;
+
+/**
+ * The LocalLogManager is a test implementation that relies on the contents of memory.
```
mumrah: Javadoc says this is a test implementation, but it's in the main sources. Is this class used by any production code? Maybe the metadata shell? If so, we should probably not say it's a "test" thing

metadata/src/main/java/org/apache/kafka/metalog/metalog/MetaLogManager.java
```diff
@@ -0,0 +1,79 @@
...
+     * The write will be scheduled to happen at some time in the future.  There is no
+     * error return or exception thrown if the write fails.  Instead, the listener may
+     * regard the write as successful if and only if the MetaLogManager reaches the given
+     * index before renouncing its leadership.  The listener should determine this by
+     * monitoring the committed indexes.
+     *
+     * @param epoch         The controller epoch.
+     * @param batch         The batch of messages to write.
+     *
+     * @return              The index of the message.
```
mumrah: Is this index the same as the metadata log offset?

metadata/src/test/java/org/apache/kafka/metadata/BrokerRegistrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void testToString() {
+        assertEquals("BrokerRegistration(id=1, epoch=0, " +
+            "incarnationId=3MfdxWlNSn2UDYsmDP1pYg, listeners=[Endpoint(" +
+            "listenerName='INTERNAL', securityProtocol=PLAINTEXT, " +
+            "host='localhost', port=9091)], supportedFeatures={foo: 1-2}, " +
+            "rack=Optional.empty, fenced=false)",
+            REGISTRATIONS.get(1).toString());
+    }
+}
```
mumrah: nit: newline

metadata/src/main/java/org/apache/kafka/metalog/metalog/LocalLogManager.java
cmccabe: Good question.  It was here because this code existed before the Raft integration.  Let's just move it into the test directory.

metadata/src/main/java/org/apache/kafka/metadata/BrokerRegistration.java
cmccabe: there aren't that many broker registrations, so it's fine to just do defensive copies in both places.  I will add that

metadata/src/main/java/org/apache/kafka/metadata/BrokerRegistration.java
cmccabe: ok

metadata/src/main/java/org/apache/kafka/metalog/metalog/MetaLogManager.java
cmccabe: Oh, this is another historical artifact, back when "index" was separate from "offset".

Right now it should just be replaced with "offset."  I will do that

mumrah (approved):
LGTM!

PR #10107, opened 2021-02-11
MINOR: Improve confusing admin client shutdown logging

hachikuji:
If the admin client is shutdown with some unfinished calls, we see messages such as the following in the log:
```
2021-02-09 11:08:05.964 DEBUG [AdminClient clientId=adminclient-1] Call(callName=fetchMetadata, deadlineMs=1612843805378) timed out at 9223372036854775807 after 1 attempt(s)
```
The problem is that we are using passing `Long.MaxValue` as the current time in `Call.fail` in order to ensure the call is timed out and we are discarding the original cause. The patch fixes the problem by setting `aborted=true` instead and preserving the original exception message.

### DIFF
--- a/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
+++ b/clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
@@ -657,7 +657,7 @@ public class KafkaAdminClient extends AdminClient {
                 // Wait for the thread to be joined.
                 thread.join(waitTimeMs);
             }
-            log.info("Kafka admin client closed.");
+            log.debug("Kafka admin client closed.");
         } catch (InterruptedException e) {
             log.debug("Interrupted while joining I/O thread", e);
             Thread.currentThread().interrupt();
@@ -756,9 +756,9 @@ public class KafkaAdminClient extends AdminClient {
             return curNode;
         }
 
-        void abortAndFail(Throwable throwable) {
+        void abortAndFail(TimeoutException timeoutException) {
             this.aborted = true;
-            fail(time.milliseconds(), throwable);
+            fail(time.milliseconds(), timeoutException);
         }
 
         /**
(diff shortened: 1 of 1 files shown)


clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -756,6 +756,11 @@ protected Node curNode() {
             return curNode;
         }
 
+        void abortAndFail(Throwable throwable) {
```
chia7712: How about changing the input type from `Throwable` to `TimeoutException`? The exception passed to this method should NOT be wrapped by `TimeoutException`. Using explicit type (`TimeoutException`) can prevent us from passing incorrect exception in the future.

chia7712 (approved):
@hachikuji Thanks for this patch. I had seen this weird error message many times :)

one small question is left but it does not obstruct this patch from merging.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
```diff
@@ -657,7 +657,7 @@ public void close(Duration timeout) {
                 // Wait for the thread to be joined.
                 thread.join(waitTimeMs);
             }
-            log.debug("Kafka admin client closed.");
+            log.info("Kafka admin client closed.");
```
dajac: I wonder if changing this log level is really necessary. What's the rational behind it?

dajac (approved):
Left a minor comment. LGTM otherwise.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
hachikuji: Makes sense.

clients/src/main/java/org/apache/kafka/clients/admin/KafkaAdminClient.java
hachikuji: I think it is helpful to get at least one log message at info level in shutdown. Really at the start of shutdown is best since it triggers logic which can be difficult to explain if you don't know shutdown is taking place. Anyway, let me remove this change here and I can open a separate patch so that we can do this consistently in all the clients.

chia7712 (approved):
+1 to latest commit

PR #10110, opened 2021-02-11
KAFKA-12297 - Make MockProducer return RecordMetadata with values as per contract

aadubey:
This is a simple change to MockProducer as per request in KAFKA-12297.
MockProducer currently returns a null RecordMetadata on Exception. The fix will make MockProducer return the right value as per specification.

This only impacts clients which use send with a custom callback and try to then use the RecordMetadata inspite of getting an exception. This should mostly impact customer unit and integration tests as the mock end point was never intended for use in a real Kafka cluster.


omkreddy (approved):
@aadubey Thanks for the PR. LGTM.

PR #10111, opened 2021-02-11
MINOR: Add BrokerMetadataListener

mumrah:
Replaces #10090 

This adds BrokerMetadataListener which is responsible for processing metadata records received by the broker when running in Raft mode.

This also moves some classes that were added to the wrong folder in trunk


cmccabe (approved):
LGTM

